PR #6753, opened 2020-10-05
Do not wait Sync Triggers on a function creation. Fixes #6669

### LINKED ISSUE #6669: Deployment of Functions using ARM template gives 'Bad Request' Internal Server Error
When trying to publish a Azure Functions app and two functions using ARM template `BadRequest (500)` appears during creation of second function.

We have a race when creating functions using "[Microsoft.Web/sites/functions](https://docs.microsoft.com/en-us/azure/templates/microsoft.web/2018-02-01/sites/functions)" thought the ARM template. The thing is that creating the function app + functions in the ARM template isn't atomic operation, the function host restarts during ARM deployment when the function app is created and PUT call to create second functions can be failed due function host restart. As result the ARM deployment is failed but function app + functions are created and work fine. 

ARM templates for reproducing:
[Repro2.txt](https://github.com/Azure/azure-functions-host/files/5247446/Repro2.txt)
[Repro1.txt](https://github.com/Azure/azure-functions-host/files/5247447/Repro1.txt)

Internal [incident](https://portal.microsofticm.com/imp/v3/incidents/details/205483736/home).

alrod:

resolves #6669

We are waiting for 5 secs after a restart and then we dispose the scopes needed for web requests completions:
https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.WebHost/DependencyInjection/ScopedResolver.cs#L34

So if an ARM template has several functions defined we will end up with the exception "DryIoc.ContainerException: 'Container is disposed and should not be used: Container is disposed". as Sync Trigger can't be executed in parallel

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/Management/WebFunctionsManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/WebFunctionsManager.cs
@@ -146,7 +146,8 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
             }
 
             // we need to sync triggers if config changed, or the files changed
-            await _functionsSyncManager.TrySyncTriggersAsync();
+            // we do not want to wait until TrySyncTriggersAsync is finished, there can be several function creations at once
+            _ = Task.Factory.StartNew(() => _functionsSyncManager.TrySyncTriggersAsync());
 
             (var success, var functionMetadataResult) = await TryGetFunction(name, request);
 


fabiocav (dismissed):
Haven't looked at the tests, but seems like something is wrong with them. Can you please double check?

src/WebJobs.Script.WebHost/Management/WebFunctionsManager.cs
```diff
@@ -146,7 +146,8 @@ await functionMetadata
             }
 
             // we need to sync triggers if config changed, or the files changed
-            await _functionsSyncManager.TrySyncTriggersAsync();
+            // we do not want to wait until TrySyncTriggersAsync is finished, there can be several function creations at once
```
mathewc, line 149: This doesn't seem right. Why does this fix the issue? With this, we'd be doing things inconsistently - all the other places (including other places in this file) we await.

fabiocav: I provided some feedback on this in person. The ideal approach is to await, but suspend the restart until this call completes.

mathewc: If the underlying issue is disposal of scopes, then this doesn't seem specific to SyncTriggers at all. Any other async operation invoked by a service in the DI container can run into similar issues?

alrod: The fix is for a CRI, the customer deploys few functions files inside ARM template. As we allow only one SyncTriggers call in once (~3 secs) the deployment take to much time and ends up with the dryloc container disposed exception. We do not rely on  SyncTriggers result in the function creation flow so the fire-end-forget SyncTrigger calls fix this particular issue.

>Any other async operation invoked by a service in the DI container can run into similar issues?

Yes, if we restarting the host and the remaining operation takes > 5 secs.

>The ideal approach is to await, but suspend the restart until this call completes.

We use file watchers for the restart and suspending restart will be a bit tricky. Do you have an idea how to achive suspension of the restart?

fabiocav: The above is not entirely accurate. In this particular case, the only reason the scope is being disposed is because this API logic is specifically triggering a restart. That would not happen in other async operations and is not an issue there.

fabiocav: @alrod the logic is all driven by the file monitoring service. If you enhance the service to suspend notifications, the restart wouldn't be triggered by the operations taking place during the API call. You'd be required to explicitly restart the host at the end, though.

alrod: As discussed with @fabiocav offline we are OK with a practical fix now, but ideally the proper fix is adding new methods to `FileMonitorService` which will insrease/descrease count of threads requested for the host restart suspending and restart the host only if the count = 0

mathewc: Since we're not doing a release this sprint, we have time to fix this properly right? Also, whatever we do, we'd want to do it consistently as I mentioned above for Delete as well, where we're also doing a sync.

alrod: Send the PR with "proper" fix:
https://github.com/Azure/azure-functions-host/pull/6799/files

alrod:
stale

PR #6764, opened 2020-10-07
A couple logging performance improvements + Benchmarks

mathewc:
Making a couple minor logging performance improvements, and adding a BenchmarksDotNet project that we can start using for benchmarking. As a sample, I added a benchmark for this change. Here's the before/after results:

| Method |     Mean |    Error |   StdDev |   Median |
|------- |---------:|---------:|---------:|---------:|
| Filter | 473.3 ns | 14.60 ns | 42.13 ns | 458.4 ns |

| Method |     Mean |    Error |   StdDev |   Median |
|------- |---------:|---------:|---------:|---------:|
| Filter | 45.65 ns | 1.301 ns | 3.733 ns | 44.18 ns |

So over a 10x improvement. Now this filtering is a small part of the total request pipeline, so the impact will be minimal, however it's just something I noticed showing up in profiles.

Going forward, we can add more benchmarks for performance critical hotpath methods.

### DIFF
--- a/src/WebJobs.Script.WebHost/Middleware/FunctionInvocationMiddleware.cs
+++ b/src/WebJobs.Script.WebHost/Middleware/FunctionInvocationMiddleware.cs
@@ -122,7 +122,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Middleware
                 // Add the request to the logging scope. This allows the App Insights logger to
                 // record details about the request.
                 ILoggerFactory loggerFactory = context.RequestServices.GetService<ILoggerFactory>();
-                ILogger logger = loggerFactory.CreateLogger(LogCategories.CreateFunctionCategory(functionExecution.Descriptor.Name));
+                ILogger logger = loggerFactory.CreateLogger(functionExecution.Descriptor.LogCategory);
                 var scopeState = new Dictionary<string, object>()
                 {
                     [ScriptConstants.LoggerHttpRequest] = context.Request,
--- a/src/WebJobs.Script/Extensions/ScriptLoggingBuilderExtensions.cs
+++ b/src/WebJobs.Script/Extensions/ScriptLoggingBuilderExtensions.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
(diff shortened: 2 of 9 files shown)


src/WebJobs.Script.WebHost/Middleware/FunctionInvocationMiddleware.cs
```diff
@@ -122,7 +122,7 @@ private async Task<IActionResult> GetResultAsync(HttpContext context, IFunctionE
                 // Add the request to the logging scope. This allows the App Insights logger to
                 // record details about the request.
                 ILoggerFactory loggerFactory = context.RequestServices.GetService<ILoggerFactory>();
-                ILogger logger = loggerFactory.CreateLogger(LogCategories.CreateFunctionCategory(functionExecution.Descriptor.Name));
```
mathewc, line 125: No need to format this category string anew on each invocation

src/WebJobs.Script/Extensions/ScriptLoggingBuilderExtensions.cs
```diff
@@ -26,9 +29,14 @@ public static ILoggingBuilder AddDefaultWebJobsFilters(this ILoggingBuilder buil
...
-        private static bool Filter(string category, LogLevel actualLevel, LogLevel minLevel)
+        internal static bool Filter(string category, LogLevel actualLevel, LogLevel minLevel)
+        {
+            return actualLevel >= minLevel && IsFiltered(category);
+        }
+
+        private static bool IsFiltered(string category)
         {
-            return actualLevel >= minLevel && ScriptConstants.SystemLogCategoryPrefixes.Where(p => category.StartsWith(p)).Any();
+            return _filteredCategoryCache.GetOrAdd(category, c => ScriptConstants.SystemLogCategoryPrefixes.Where(p => category.StartsWith(p)).Any());
```
mathewc, line 39: Cache, to avoid doing a startswith string comparison for each category prefix (currently there are 4) on every log statement.

paulbatum (approved):
Looks good to me. I agree its worthwhile that to add microbenchmarks as we make small optimizations like this. Just one thing - the PR template includes a bunch of checks that we should include on every PR. I think all of them are no-ops in this case but I'd like to get in the habit of keeping those checks in every PR we file.

PR #6765, opened 2020-10-07
Add environment variable for AzureMonitorDiagnosticLogger dynamic enablement

mathewc:
### Description

Defines an environment variable that enables AzureMonitor logging categories to be flowed to the host. The AzureMontitor logger is relatively expensive, so we should only register it when needed. DWAS will be updated to flow the log categories to us via this new environment variable. For backwards compatability, if the env var isn't present, we keep the current behavior (logger present).

TFS work item for Antares side: https://msazure.visualstudio.com/One/_workitems/edit/8521884

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.WebHost/DependencyInjection/DependencyValidator/DependencyValidator.cs
+++ b/src/WebJobs.Script.WebHost/DependencyInjection/DependencyValidator/DependencyValidator.cs
@@ -57,7 +57,6 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.DependencyInjection
                 .OptionalExternal("Microsoft.Extensions.Diagnostics.HealthChecks.HealthCheckPublisherHostedService", "Microsoft.Extensions.Diagnostics.HealthChecks", "adb9793829ddae60"); // Popularly-registered by Health Check Monitor.
 
             expected.ExpectSubcollection<ILoggerProvider>()
-                .Expect<AzureMonitorDiagnosticLoggerProvider>()
                 .Expect<FunctionFileLoggerProvider>()
                 .Expect<HostFileLoggerProvider>()
                 .Expect<SystemLoggerProvider>();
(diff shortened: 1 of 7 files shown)


src/WebJobs.Script.WebHost/DependencyInjection/DependencyValidator/DependencyValidator.cs
```diff
@@ -53,11 +53,11 @@ private static ExpectedDependencyBuilder CreateExpectedDependencies()
                 .Optional<FuncAppFileProvisioningService>() // Used by powershell.
                 .Optional<JobHostService>() // Missing when host is offline.
                 .Optional<FunctionsSyncService>() // Conditionally registered.
+                .Optional<AzureMonitorDiagnosticLoggerProvider>() // Conditionally registered.
```
brettsam: You can just remove this completely. The `ExpectSubcollection` below means -- we require these services, but customers (or us) can add additional services without failing the check. So just by removing the `.Expect<AzureMonitorDiagnosticLoggerProvider>()`, that puts this into an "optional" category.

src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
```diff
@@ -57,6 +57,7 @@ public static class EnvironmentSettingNames
         public const string RoleInstanceId = "RoleInstanceId";
         public const string HealthPingEnabled = "WEBSITE_FUNCTIONS_HEALTH_PING_ENABLED";
         public const string TestDataCapEnabled = "WEBSITE_FUNCTIONS_TESTDATA_CAP_ENABLED";
+        public const string AzureMonitorEnabled = "WEBSITE_FUNCTIONS_AZUREMONITOR_ENABLED";
```
brettsam: I'd recommend a change to this. Today we only have one log category (FunctionAppLogs), but we've already been discussing adding a second (something like FunctionInvocationResults -- that has just the success/fail results from all invocations, minus all the trace logs). 

That turns this from an on/off switch to a collection of categories, depending on how you surface this from DWAS. So I think this should change to be something like `WEBSITE_FUNCTIONS_AZUREMONITOR_CATEGORIES`. And today, if "FunctionAppLogs" is there, we add the provider.

brettsam (approved):

PR #6772, opened 2020-10-09
Add debug logs in FunctionRpcService

pragnagopa:
Follow up from a CRI. We need debug logs in `FunctionRpcService` to figure out if an `invocation` is handed off to language worker or not.

### DIFF
--- a/src/WebJobs.Script.Grpc/Server/FunctionRpcService.cs
+++ b/src/WebJobs.Script.Grpc/Server/FunctionRpcService.cs
@@ -60,7 +60,7 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
                             {
                                 if (evt.MessageType == MsgType.InvocationRequest)
                                 {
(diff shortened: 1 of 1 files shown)


mathewc:
We changed these to Trace logs as part of performance work in https://github.com/Azure/azure-functions-host/pull/5897. In the case you were looking at, where did the issue lay? Wondering if there isn't a way for us to only write error logs to catch that case, rather than adding these logs back on all requests.

pragnagopa:
@mathewc - these are the only logs  we have in gRpc service - without these logs per invocation - CRI investigation become harder as it will be difficult to figure out when an invocation was handed off to language worker.

pragnagopa:
Closing this PR. Logs coming from https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs#L415 should be enough.

PR #6781, opened 2020-10-12
Minor metrics improvements for concurrent access

Scooletz:
This PR introduces a few minor fixes related to the `Metrics` area of the host:

1. stopping a StopWatch is checked now
1. `QueuedEvents` no longer modifies existing entries. They are recreated to ensure that potential multiple writers don't perform dirty writes.
1. `FlushMetrics` separates obtaining a list of keys from removing them. With this is, if a new metric is reported between capturing a key and removing it it won't be lost. With previous approach using `.Clear()` there was a potential for loosing a metric.

A few remarks:

- I put changes into separate commits to make it easier to review or potentially create separate PRs.
- I'm new to this repo. I'm trying to add some minor improvements/fixes learning about the host implementation at the same time.

### DIFF
--- a/src/WebJobs.Script.WebHost/Diagnostics/MetricsEventManager.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/MetricsEventManager.cs
@@ -80,8 +80,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
             SystemMetricEvent evt = eventHandle as SystemMetricEvent;
             if (evt != null)
             {
-                long latencyMS = 0;
-               
+                long latencyMS;
                 if (evt.StopWatch != null)
                 {
                     evt.StopWatch.Stop();
@@ -115,17 +114,15 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
                             Data = evt.Data
                         };
                     },
-                    (name, evtToUpdate) =>
+                    (name, existing) =>
                     {
-                        // Aggregate into the existing event
-                        // While we'll be performing an aggregation later,
-                        // we retain the count so weighted averages can be performed
-                        evtToUpdate.Maximum = Math.Max(evtToUpdate.Maximum, latencyMS);
-                        evtToUpdate.Minimum = Math.Min(evtToUpdate.Minimum, latencyMS);
-                        evtToUpdate.Average += latencyMS;  // the average is calculated later - for now we sum
-                        evtToUpdate.Count++;
-
-                        return evtToUpdate;
+                        return new SystemMetricEvent
+                        {
+                            Maximum = Math.Max(existing.Maximum, latencyMS),
+                            Minimum = Math.Min(existing.Minimum, latencyMS),
+                            Average = existing.Average + latencyMS,
+                            Count = existing.Count + 1
+                        };
                     });
             }
         }
--- a/src/WebJobs.Script.WebHost/Diagnostics/MetricsEventManager.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/MetricsEventManager.cs
@@ -118,10 +118,13 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
                     {
                         return new SystemMetricEvent
                         {
+                            FunctionName = existing.FunctionName,
+                            EventName = existing.EventName,
                             Maximum = Math.Max(existing.Maximum, latencyMS),
                             Minimum = Math.Min(existing.Minimum, latencyMS),
                             Average = existing.Average + latencyMS,
-                            Count = existing.Count + 1
+                            Count = existing.Count + 1,
+                            Data = existing.Data
                         };
                     });
             }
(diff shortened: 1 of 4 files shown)


src/WebJobs.Script.WebHost/Diagnostics/MetricsEventManager.cs
```diff
@@ -114,17 +114,15 @@ public void EndEvent(object eventHandle)
                             Data = evt.Data
                         };
                     },
-                    (name, evtToUpdate) =>
+                    (name, existing) =>
```
mathewc, line 117: You're pointing out that the update delegate isn't run under a lock and can therefore run concurrently. Does the change below actually improve things? E.g. in a case where 2 updates run concurrently, your code will create a new value from the "existing" value passed in, but they'll both get the same existing value, so it's still last one wins?

src/WebJobs.Script.WebHost/Diagnostics/MetricsEventManager.cs
```diff
@@ -231,13 +232,19 @@ public void Flush()
 
         private void FlushMetrics()
         {
-            if (QueuedEvents.Count == 0)
+            if (QueuedEvents.IsEmpty)
```
mathewc, line 241: Count takes a lock, whereas in https://github.com/drewnoakes/corefx/commit/67e2891b64edd510e4245dc86434f50859776472 a fastpath check was added. So this looks like a good change :)

src/WebJobs.Script.WebHost/Diagnostics/MetricsEventManager.cs
```diff
@@ -80,10 +80,10 @@ public void EndEvent(object eventHandle)
             SystemMetricEvent evt = eventHandle as SystemMetricEvent;
             if (evt != null)
             {
-                long latencyMS = 0;
-                evt.StopWatch.Stop();
+                long latencyMS;
                 if (evt.StopWatch != null)
                 {
+                    evt.StopWatch.Stop();
```
mathewc, line 86: Because StopWatch is never null this won't have any real effect because we always set it when creating SystemMetricEvents above - you're just improving the logic because we're already checking for null?

src/WebJobs.Script.WebHost/Diagnostics/MetricsEventManager.cs
Scooletz: This is correct, as the updated value is calculated without the lock. This means that ++ can be run by two threads at the same time with no barriers (`volatile` or `Interlocked`) and no locking. After this change, when 2 updates are run concurrently, only 1 will succeed in this loop https://github.com/dotnet/runtime/blob/12299e2ed8e926ce20c3af6f49de39ee0d17e71c/src/libraries/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentDictionary.cs#L1253-L1272 and the other will spin again, recalculate the value and try to update again. This is change is about correctness in reporting, not performance (for sure it will alloc at least one object per update).

src/WebJobs.Script.WebHost/Diagnostics/MetricsEventManager.cs
Scooletz: This was my point. There's a check, let it check what it checks 😅

src/WebJobs.Script.WebHost/Diagnostics/MetricsEventManager.cs
mathewc: I'm going to run these changes using our [Crank](https://github.com/dotnet/crank) setup to measure the perf impact.

Scooletz: 👍 For sure there will be some, this allocates at least one instance per `AddOrUpdate`. 

If we want to make it really fast and correct at the same time, I'll need to kill this PR and issue another one. I think this would result in a bigger set of changes. In the past I did the metrics/monitoring protocol stuff for https://github.com/Particular/ServiceControl.Monitoring.Data 

Anyway, looking forward to hearing more! If we could correct it first and then consider whether or not more performance is required it'd be awesome.

src/WebJobs.Script.WebHost/Diagnostics/MetricsEventManager.cs
```diff
@@ -150,12 +148,15 @@ public void LogEvent(string eventName, string functionName = null, string data =
...
                     };
                 },
-                (name, evtToUpdate) =>
+                (name, existing) =>
                 {
-                    // update the existing event
-                    evtToUpdate.Count++;
-
-                    return evtToUpdate;
+                    return new SystemMetricEvent
```
mathewc, line 156: When perf testing your changes I was getting failures. The load test failed due to failed requests. I just added the full-ci label to this PR to force a test run, and I expect some tests to fail. One issue is this code - it needs to copy just the FunctionName,EventName,Data,Count properties as in the create statement above. This isn't an aggregate event, it's a "count" event based on event name, so needs these properties set. There are other test failures as well - recommend you run the Metrics unit tests locally to find them all.

mathewc: Even after experimenting myself and fixing things up so tests pass on my box https://github.com/Azure/azure-functions-host/tree/metrics-improvements under load these changes aren't working (request failures).

Scooletz: Oh I see, sorry for wasting your cycles @mathewc Will run it locally and push fixes soon.

Scooletz:
I pushed the changes and adjusted some of `MetricsEventManagerTests` as they were using the `initialEvent` which is no longer valid after the changes.

fabiocav:
@mathewc @Scooletz is this still something we want to get in?

Scooletz:
This PR is 5 months old and has a conflict. I think I don't have spare cycles now to resolve the conflict and load the context. I'm of with closing it.

PR #6782, opened 2020-10-12
Make bindings parallel

### LINKED ISSUE #6775: Load bindings in parallel
The bindings are loaded in series with the current code.

Make them load in parallel.

goiri:

resolves #6775

### Pull request checklist

### Additional information

The bindings are processed serially. We could save some time processing the bindings in parallel.

The test to validate this was a Python function with 3 input bindings and 2 output bindings.
Each binding was less than 1KB and the access times in my machine for the inputs were around 2-3 milliseconds.
In the serial case, I was seeing ~7ms for processing all the inputs and with this change, it went under 3 ms.
For the output bindings, the times per binding were ~10ms per binding which ended up adding around 20ms.
With this approach the time went down to ~10ms for both outputs.

Note that a similar optimization also works in the Python side and we will open a separate issue for that side.

### DIFF
--- a/src/WebJobs.Script/Description/Workers/WorkerFunctionInvoker.cs
+++ b/src/WebJobs.Script/Description/Workers/WorkerFunctionInvoker.cs
@@ -125,6 +125,7 @@ namespace Microsoft.Azure.WebJobs.Script.Description
         private async Task<(string name, DataType type, object value)[]> BindInputsAsync(Binder binder)
         {
             var bindingTasks = _inputBindings
+                .AsParallel()
                 .Where(binding => !binding.Metadata.IsTrigger)
                 .Select(async (binding) =>
                 {
@@ -152,7 +153,7 @@ namespace Microsoft.Azure.WebJobs.Script.Description
 
             _handleScriptReturnValue(result);
 
-            var outputBindingTasks = _outputBindings.Select(async binding =>
+            var outputBindingTasks = _outputBindings.AsParallel().Select(async binding =>
             {
                 // apply the value to the binding
                 if (result.Outputs.TryGetValue(binding.Metadata.Name, out object value) && value != null)


src/WebJobs.Script/Description/Workers/WorkerFunctionInvoker.cs
```diff
@@ -152,7 +153,7 @@ private async Task BindOutputsAsync(object input, Binder binder, ScriptInvocatio
 
             _handleScriptReturnValue(result);
 
-            var outputBindingTasks = _outputBindings.Select(async binding =>
+            var outputBindingTasks = _outputBindings.AsParallel().Select(async binding =>
```
pragnagopa, line 156: I do not think we have E2E tests as part of PR builds that large number of multiple inputs/outputs. Can you please update the PR description with the tests you have run and any perf improvements this changes would bring in?

src/WebJobs.Script/Description/Workers/WorkerFunctionInvoker.cs
```diff
@@ -125,6 +125,7 @@ private async Task DelayUntilFunctionDispatcherInitializedOrShutdown()
         private async Task<(string name, DataType type, object value)[]> BindInputsAsync(Binder binder)
         {
             var bindingTasks = _inputBindings
+                .AsParallel()
```
pragnagopa, line 128: Wondering if it might be safer to enforce `.WithDegreeOfParallelism` to defined number.

src/WebJobs.Script/Description/Workers/WorkerFunctionInvoker.cs
pragnagopa: cc @brettsam / @mathewc  - if they have any concerns processing multiple output bindings in parallel

pragnagopa (commented):
Added minor comments. Thanks!

src/WebJobs.Script/Description/Workers/WorkerFunctionInvoker.cs
goiri: That would make sense, I have tested with 3 input bindings and output bindings.
Do you have any preference for the maximum degree of parallelism?

src/WebJobs.Script/Description/Workers/WorkerFunctionInvoker.cs
goiri: I added the description of the test application I used for validating this change.
My tests show that the behavior we get with this change is pretty much the optimal one would expect.
The more bindings we have, the more functions benefit from this.
However, as mentioned in the other comment we may want to limit the maximum number of parallel requests.

mathewc: We already bind the input/output bindings in parallel - that's what the below Task.WhenAll calls do. So I don't see why we need any changes here?


This is for Functions out of proc. Interestingly, the core SDK has always bound inputs/outputs serially (input binding code [here](https://github.com/Azure/azure-webjobs-sdk/blob/6e2e5695ced990ae32b01d8108777bd87c7b7f22/src/Microsoft.Azure.WebJobs.Host/Triggers/TriggeredFunctionBinding.cs#L83), output binding code [here](https://github.com/Azure/azure-webjobs-sdk/blob/b05b4466ed67de8331dd940e1262b5efb43c9f65/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs#L933)).

goiri: @mathewc from the code, that's what I thought too.
However, when running this and timing it, I see the bindings being done sequentially.
It must be some weird behavior with the await/async.

mathewc: Should be straight forward to write a unit test that verifies that these bindings are executed in parallel. The Linq query returns an enumerable of tasks that may have started sequentially, but they'll complete at different times based on what they do I expect. I'd like to see a test proving that this isn't happening, since it's hard to believe :) We use Task.WhenAll all over the codebase for parallel execution.

goiri: I was as skeptical as you :)
We tested adding {{await Task.Delay(1000)}} but in this case, it does run in parallel.
The issue comes when the code inside runs without stop (e.g., the binding one).
I am not very familiar with the unit testing infrastructure for the bindings, do you have any suggestion?
I can also do a screen share session with the example app and the timing.

mathewc: That means this code is already doing what it should. For a binding that behaves well and goes async for any IO (as all the existing bindings I know about do), everything will be done in parallel. What specific bindings are you having issues with? You're saying that you're using a binding that blocks the thread? Is this a custom binding?

goiri: This is happening with the Azure Blob binding.
For blocking I was referring anything that does not do Delay().
I can add instrumentation to the code and show the timings before and after.

fabiocav:
@mathewc @goiri this PR hasn't been active. Is it still relevant?

goiri:
@fabiocav thanks for the ping.
This improves a very narrow case that shaves a couple of milliseconds with lots of small bindings.
It is hard to justify the change so I am fine closing it.

PR #6789, opened 2020-10-14
Fix Flaky Validate_HostLogs

### LINKED ISSUE #6521: Fix Flaky Validate_HostLogs Test
Microsoft.Azure.WebJobs.Script.Tests.ApplicationInsights.ApplicationInsightsNodeEndToEndTests.Validate_HostLogs

has been failing in devops intermittently lately but passes locally. It was disabled in the meantime but needs to be looked at to improve consistency. This is in v2 and v3.

yojagad:

resolves #6521 .

### Pull request checklist

### Additional information

This seems to be possibly due to some other test Host holding the lease on the blob. Most recent example of this https://azfunc.visualstudio.com/Azure%20Functions/_build/results?buildId=8572&view=logs&j=c68c283c-e579-5115-c911-5c3ac7f5af09&t=188f054b-4c24-59d9-f813-b2324ef559fd

This is also evident from storage logs below where you can see that an entity keeps trying to acquire lease on fvaz123231-46172260/host but keeps failing with "LeaseAlreadyPresent". There are also frequent lease renewals on this same blob

(navigate to 2020/10/13/1900 000000.log)
https://ms.portal.azure.com/#blade/Microsoft_Azure_Storage/ContainerMenuBlade/overview/storageAccountId/%2Fsubscriptions%2Fca01ee66-e03f-4d4f-abd8-ea53e5d4f226%2FresourceGroups%2Fazure-functions-host-ci-5%2Fproviders%2FMicrosoft.Storage%2FstorageAccounts%2Fazurefunctionshostci5/path/%24logs/etag/%220x8D7986F34C2D428%22/defaultEncryptionScope/%24account-encryption-key/denyEncryptionScopeOverride//defaultId//publicAccessVal/none

Overriding the IHostIdProvider with a customProvider so that a there's a unique blob for this test just might fix the issue we have here. @brettsam Let me know if there's a better fix for this that I may have missed, please.

### DIFF
--- a/test/WebJobs.Script.Tests.Integration/Controllers/WarmupScenarios.cs
+++ b/test/WebJobs.Script.Tests.Integration/Controllers/WarmupScenarios.cs
@@ -6,14 +6,13 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Net.Http;
-using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Xunit;
 
 namespace Microsoft.Azure.WebJobs.Script.Tests.Integration.Controllers
 {
-    public class WarmupScenarios
+    public class WarmupScenarios : IDisposable
     {
         private static readonly TimeSpan SemaphoreWaitTimeout = TimeSpan.FromSeconds(30);
 
@@ -85,6 +84,11 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.Integration.Controllers
             // Better be successful
             Assert.True(response.IsSuccessStatusCode, "Warmup endpoint did not return a success status. " +
                 $"Instead found {response.StatusCode}");
+        }
+
+        public void Dispose()
+        {
+            _testHost.Dispose();
         }
 
         private class PausingScriptHostBuilder : IScriptHostBuilder
--- a/test/WebJobs.Script.Tests.Integration/Controllers/WarmupScenarios.cs
+++ b/test/WebJobs.Script.Tests.Integration/Controllers/WarmupScenarios.cs
@@ -88,7 +88,7 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.Integration.Controllers
 
         public void Dispose()
         {
-            _testHost.Dispose();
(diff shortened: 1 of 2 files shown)


yojagad:
Pinging for reviews.

brettsam:
In your example above -- where did the id `fvaz123231-46172260` come from? It looks like the logs from 10/13 are gone -- is there any time this happened more recently?

yojagad:
The build I referenced has these logs along with the additional log that I added to confirm this
https://azfunc.visualstudio.com/Azure%20Functions/_build/results?buildId=8572&view=logs&j=c68c283c-e579-5115-c911-5c3ac7f5af09&t=188f054b-4c24-59d9-f813-b2324ef559fd . Here's a screenshot from the build too
image

But it looks like the storage logs for this are gone :( And no, the test has been disabled and it took me 3 days of running builds to capture that storage log :( Fortunately, it looks like I had an offline copy of logs from 10/13, so uploading that here
[000000.log](https://github.com/Azure/azure-functions-host/files/5447906/000000.log)

test/WebJobs.Script.Tests.Integration/Controllers/WarmupScenarios.cs
```diff
@@ -87,6 +86,11 @@ public async Task TestWarmupEndPoint_WhenHostStarts()
                 $"Instead found {response.StatusCode}");
         }
 
+        public void Dispose()
+        {
+            _testHost.Dispose();
```
brettsam: make this `_testHost?.Dispose();`, just in case something weird happens and testhost isn't created yet.

brettsam (dismissed):

brettsam (approved):

yojagad:
Discussed with Brett offline. Thought it might be a good idea to first check whether WarmupScenarios not disposing the _testFunctionHost was causing the side effect of the test holding onto the lease.

Made the change and reenabled the test. If it's found that the test starts failing again, then we could potentially bring back the change I originally had (ie., to override the HostIdProvider).

PR #6792, opened 2020-10-14
Shared memory data transfer between Functions Host and out-of-proc workers

### LINKED ISSUE #6791: Shared memory data transfer between Functions host and out-of-proc workers for binary data

    - A `GUID` is generated
    - `MemoryMappedFileName`

##### Sequence of Events
13) Done

#### Additional context
image

gohar94:

resolves #6791

### Pull request checklist

### Additional information

Additional PR information


gohar94:
cc: @goiri 

Opening a draft PR for now. You can skim through whenever.

pragnagopa:
cc @fabiocav 

Assigned to myself for review

pragnagopa:
cc @ankitkumarr  FYI - as he will be working on some changes in this area

fabiocav:
@gohar94 thank you! will be spending some time reviewing this ASAP

gohar94:
@fabiocav Thanks a lot. I would suggest going over the [issue](https://github.com/Azure/azure-functions-host/issues/6791) for now which has details on the protocol etc. Once I have the code in a decent shape, then I will change this PR from draft to published and you can review it in detail then. :)

src/WebJobs.Script/Workers/Rpc/SharedMemoryDataTransfer/MemoryMappedFileHandler/FileOperators/FileAccessor.cs
```diff
@@ -0,0 +1,331 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Script.Grpc
+{
+    /// <summary>
+    /// Class to perform access operations related to <see cref="MemoryMappedFile"/> like creating
+    /// new ones, opening existing ones, deleting etc.
+    /// </summary>
+    internal class FileAccessor
+    {
```
pragnagopa: Create a new type: `SharedMemoryFile`  - this should contain all the methods you exposed in : `FileAccessor` `FileReader` `FileWriter` - as the purpose of all these is to do CRUD operations on the MemoryMappedFile

src/WebJobs.Script/Workers/Rpc/SharedMemoryDataTransfer/MemoryMappedFileHandler/FileOperators/FileAccessor.cs
```diff
@@ -0,0 +1,331 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using System.IO.MemoryMappedFiles;
+using System.Runtime.InteropServices;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Script.Grpc
```
pragnagopa: nit: Fix namespace

src/WebJobs.Script/Workers/Rpc/SharedMemoryDataTransfer/MemoryMappedFileHandler/MemoryMappedFileConstants.cs
```diff
@@ -0,0 +1,94 @@
...
+
+using System.Collections.Generic;
+using System.IO.MemoryMappedFiles;
+
+namespace Microsoft.Azure.WebJobs.Script.Grpc
+{
+    /// <summary>
+    /// Constants for the <see cref="MemoryMappedFile"/>.
+    /// </summary>
+    public static class MemoryMappedFileConstants
```
pragnagopa: nit: Move each public type to its own file

src/WebJobs.Script/Workers/Rpc/SharedMemoryDataTransfer/MemoryMappedFileHandler/ResponseTypes/ResponseByteArray.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.MemStore.MemoryMappedFileHandler.ResponseTypes
+{
+    /// <inheritdoc/>
+    public class ResponseByteArray : ResponseBase
```
pragnagopa: All the response types can be simplified into conversion methods - extension methods.

src/WebJobs.Script/Workers/Rpc/SharedMemoryDataTransfer/SharedMemoryManager.cs
```diff
@@ -0,0 +1,101 @@
...
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.IO.MemoryMappedFiles;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Microsoft.MemStore.MemoryMappedFileHandler.ResponseTypes;
+
+namespace Microsoft.Azure.WebJobs.Script.Grpc
+{
+    internal class SharedMemoryManager
```
pragnagopa: New type `SharedMemoryFile` should handle methods exposed here as well.

src/WebJobs.Script/Workers/Rpc/SharedMemoryDataTransfer/SharedMemoryManager.cs
```diff
@@ -0,0 +1,101 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections;
```
pragnagopa: Move contents from folder: `SharedMemoryDataTransfer` up one level - under workers - as this can used by non-Rpc workers as well.

pragnagopa (commented):
Provided initial feedback around abstractions and code layout. Thanks!

src/WebJobs.Script/Workers/Rpc/SharedMemoryDataTransfer/SharedMemoryManager.cs
gohar94: I want to keep this separate as this is going to act as a basic collection/manager of `SharedMemoryFile`s. It holds the references (to keep the allocated memory) until it is used (after the invocation is over) and then can free the `SharedMemoryFile`s and the allocated memory.

src/WebJobs.Script/Description/Workers/ScriptInvocationContext.cs
```diff
@@ -31,5 +32,13 @@ public class ScriptInvocationContext
...
 
         public System.Threading.ExecutionContext AsyncExecutionContext { get; set; }
+
+        /// <summary>
+        /// Gets or sets the list of names of <see cref="MemoryMappedFile"/> that were allocated to
+        /// transfer data to the worker process for this invcation.
+        /// These are tracked here so that once the invocation is complete, the resources can be
+        /// freed.
+        /// </summary>
+        public IList<string> SharedMemoryResources { get; set; }
```
pragnagopa: This list is created and deleted with in the worker channel.  No need to expose this property here

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -35,6 +37,7 @@ internal class GrpcWorkerChannel : IRpcWorkerChannel, IDisposable
         private readonly string _runtime;
         private readonly IEnvironment _environment;
         private readonly IOptionsMonitor<ScriptApplicationHostOptions> _applicationHostOptions;
+        private readonly SharedMemoryManager _sharedMemoryManager;
```
pragnagopa: Create an interface - `IMemoryMappedFileManager` - Move all the logic that maintains and deletes list of memorymapped files into concrete implementation - `MemoryMappedFileManager`

src/WebJobs.Script/Workers/SharedMemoryDataTransfer/SharedMemoryFile.cs
```diff
@@ -0,0 +1,599 @@
...
+
+        public static SharedMemoryFile Create(ILogger logger, string mapName, long size)
+        {
+            if (TryCreate(logger, mapName, size, out MemoryMappedFile mmf))
+            {
+                return new SharedMemoryFile(logger, mapName, mmf);
+            }
+            else
+            {
+                throw new Exception($"Cannot create MemoryMappedFile {mapName} with {size} bytes");
```
pragnagopa: do not throw new exception through out this file. Let the caller handle the error case.

src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs
```diff
@@ -36,11 +38,18 @@ public static object ToObject(this TypedData typedData)
                 case RpcDataType.Json:
                     return JsonConvert.DeserializeObject(typedData.Json, _datetimeSerializerSettings);
                 case RpcDataType.Http:
-                    return GrpcMessageExtensionUtilities.ConvertFromHttpMessageToExpando(typedData.Http);
+                    return GrpcMessageExtensionUtilities.ConvertFromHttpMessageToExpando(typedData.Http, sharedMemoryManager);
                 case RpcDataType.Int:
                     return typedData.Int;
                 case RpcDataType.Double:
                     return typedData.Double;
+                case RpcDataType.SharedMemoryData:
```
pragnagopa: Can you take a look at the proto file changes I requested?  Moving new type: memory mapped data under `parameterbinding` will simplify this code.

src/WebJobs.Script.Grpc/Extensions/ScriptInvocationContextExtensions.cs
```diff
@@ -33,9 +35,16 @@ public static async Task<InvocationRequest> ToRpcInvocationRequest(this ScriptIn
                 TypedData rpcValue = null;
                 if (input.val == null || !rpcValueCache.TryGetValue(input.val, out rpcValue))
                 {
-                    rpcValue = await input.val.ToRpc(logger, capabilities);
+                    // TODO gochaudh: Check config before using SharedMemoryManager
+                    rpcValue = await input.val.ToRpc(logger, capabilities, sharedMemoryManager);
```
pragnagopa: - check `capabilities` if shared memory data feature is supported
   - If yes, enforce minimum size of the payload - 1MB?
       -  if size > 1MB , create memory map
        - if creating memory map fails - fall back to existing code path

pragnagopa: Also, add check for environment variable - `Functions_Worker_SharedMemoryData` is not `false` -- this is to ensure users have ability to turn off the feature via AppSettings.

src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs
```diff
@@ -36,11 +38,18 @@ public static object ToObject(this TypedData typedData)
...
+                    var sharedMemData = typedData.SharedMemoryData;
+                    string mmfName = sharedMemData.MemoryMappedFileName;
+                    long offset = sharedMemData.Offset;
+                    long count = sharedMemData.Count;
+                    string type = sharedMemData.Type;
+                    return sharedMemoryManager.TryGetAsync(mmfName, offset, count).GetAwaiter().GetResult();
```
pragnagopa: Moving this under ParameterBinding - will avoid making this a blocking call as well.

pragnagopa (commented):
Added few more comments.

src/WebJobs.Script/Description/Workers/ScriptInvocationContext.cs
gohar94: The reason for keeping this list here is so that all `SharedMemoryFile`s created for a particular invocation can be passed from `SendInvocationRequest` (where they are created) to `InvocationResponse` (where they are freed.) I don't see a better way to return all these from `ToRpcInvocationRequest` (which is called from `SendInvocationRequest`) to track this list in `GrpcWorkerChannel`.

The alternate I can think of is to have a method in `SharedMemoryManager` to add a mapping between invocation ID and list of `SharedMemoryFile` names (something like `SharedMemoryManager::TrackSharedMemoryFilesForInvocation(string invocationId, IList<string> mapNames)`) which can be called from inside `ToRpcInvocationRequest` where the `SharedMemoryFile`s are created and we already know the invocation ID and also have the `SharedMemoryManager`. Then `GrpcWorkerChannel` can free these from inside `InvocationResponse` by calling another method on `SharedMemoryManager` (something like `SharedMemoryManager::FreeSharedMemoryFilesForInvocation(string invocationId)`).

Thoughts?

pragnagopa: yes, that is what I am referring to . Define an interface and move the logic that keeps a record of invocationId and memory maps created for that invocation into it. This isolates how a channel manages memory maps into an interface. Worker channels that choose to use memory maps can use this interface.

src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs
```diff
@@ -13,6 +13,8 @@
 using Microsoft.Azure.WebJobs.Script.Extensions;
 using Microsoft.Azure.WebJobs.Script.Grpc.Messages;
 using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
+using Microsoft.Azure.WebJobs.Script.Workers.SharedMemoryDataTransfer;
+using Microsoft.CodeAnalysis.VisualBasic.Syntax;
```
pragnagopa: nit: undo changes in this file.

src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageExtensionUtilities.cs
```diff
@@ -7,6 +7,7 @@
 using System.Linq;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Azure.WebJobs.Script.Grpc.Messages;
+using Microsoft.Azure.WebJobs.Script.Workers.SharedMemoryDataTransfer;
```
pragnagopa: nit: undo changes in this file.

src/WebJobs.Script/Description/Workers/ScriptInvocationContext.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System.Collections.Generic;
+using System.IO.MemoryMappedFiles;
```
pragnagopa: nit: undo changes in this file.

src/WebJobs.Script/Workers/SharedMemoryDataTransfer/ISharedMemoryManager.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        Task<SharedMemoryMetadata> PutBytesAsync(byte[] content);
+
+        Task<SharedMemoryMetadata> PutObjectAsync(object input);
+
+        Task<byte[]> GetBytesAsync(string mapName, long offset, long count);
+
+        bool IsSupported(object input);
+
+        void TrackSharedMemoryMapForInvocation(string invocationId, string mapName);
```
pragnagopa: Rename to : `AddSharedMemoryMapForInvocation`

src/WebJobs.Script/Workers/SharedMemoryDataTransfer/ISharedMemoryManager.cs
```diff
@@ -0,0 +1,24 @@
...
+
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.Script.Workers.SharedMemoryDataTransfer
+{
+    public interface ISharedMemoryManager
+    {
+        Task<SharedMemoryMetadata> PutBytesAsync(byte[] content);
+
+        Task<SharedMemoryMetadata> PutObjectAsync(object input);
```
pragnagopa, line 10: For now, if you are adding support only for `byte[]` - remove `PutObjectAsync`

src/WebJobs.Script/Workers/SharedMemoryDataTransfer/ISharedMemoryManager.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        Task<byte[]> GetBytesAsync(string mapName, long offset, long count);
+
+        bool IsSupported(object input);
+
+        void TrackSharedMemoryMapForInvocation(string invocationId, string mapName);
+
+        bool TryFreeAllResourcesForInvocation(string invocationId);
```
pragnagopa: nit: Rename to `TryFreeMemoryMapsForInvocation` - this follows similar naming convention as Add

src/WebJobs.Script/Workers/SharedMemoryDataTransfer/ISharedMemoryManager.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        bool TryFree(string mapName);
```
pragnagopa: nit: Rename to `TryFreeMemoryMap`

src/WebJobs.Script/Workers/SharedMemoryDataTransfer/SharedMemoryMap.cs
```diff
@@ -0,0 +1,599 @@
...
+
+using System;
+using System.IO;
+using System.IO.MemoryMappedFiles;
+using System.Runtime.InteropServices;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Script.Workers.SharedMemoryDataTransfer
+{
```
pragnagopa, line 12: As discussed, Create a base abstract class and create OS specific implementations

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -595,5 +634,40 @@ public bool TryFailExecutions(Exception workerException)
             }
             return true;
         }
+
+        private static bool IsSharedMemoryDataTransferEnabled(ILogger logger, IEnvironment environment, GrpcCapabilities capabilities)
```
pragnagopa: change to 
```
internal static bool IsSharedMemoryDataTransferEnabled(ILogger logger, IEnvironment environment, GrpcCapabilities capabilities)
        {
            // 1) Check if the environment variable (AppSetting) has this feature enabled
            string envVal = environment.GetEnvironmentVariable(RpcWorkerConstants.FunctionsWorkerSharedMemoryDataTransferSettingName);
            if (string.IsNullOrEmpty(envVal))
            {
                return false;
            }

            bool envValEnabled = false;
            if (bool.TryParse(envVal, out bool boolResult))
            {
                // Check if value was specified as a bool (true/false)
                envValEnabled = boolResult;
            }
            else if (int.TryParse(envVal, out int intResult))
            {
                // Check if value was specified as an int (1/0)
                envValEnabled = intResult == 1;
            }
            if (!envValEnabled)
            {
                return false;
            }

            // Check if the worker supports this feature.
            bool capabilityEnabled = !string.IsNullOrEmpty(capabilities.GetCapabilityState(RpcWorkerConstants.SharedMemoryDataTransfer));
            logger.LogTrace($"IsSharedMemoryDataTransferEnabled: {capabilityEnabled}");
            return capabilityEnabled;
        }
```

Add unit tests.

pragnagopa (commented):
Thanks for the updates. Overall layout looks good. Mostly added more nits. Please ping here when ready for more review.

pragnagopa:
@fabiocav - @gohar94  addressed first round of feedback. Please take a look when you get a chance.

src/WebJobs.Script.Grpc/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
```diff
@@ -343,8 +384,13 @@ message ParameterBinding {
   // Name for the binding
   string name = 1;
 
+  // Note: Either *data* or *rpc_shared_memory* field will be populated
+
   // Data for the binding
   TypedData data = 2;
```
goiri, line 391: Should we use the oneof semantics?

gohar94: @pragnagopa suspects that it may not be backward compatible (and hence break other language workers that we don't modify yet) - I will test that theory out later and make a final call on this then.

src/WebJobs.Script/Workers/SharedMemoryDataTransfer/ISharedMemoryManager.cs
gohar94: The reason I am keeping this is because the caller (`ToRpcInvocationRequest` --> `SharedMemoryDataTransfer` --> `PutObjectAsync`) is dealing with `object` so the checks for which type the `object` is and how to deal with it would be cleaner if kept inside `ISharedMemoryManager`.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -595,5 +634,36 @@ public bool TryFailExecutions(Exception workerException)
             }
             return true;
         }
+
+        internal static bool IsSharedMemoryDataTransferEnabled(ILogger logger, IEnvironment environment, GrpcCapabilities capabilities)
+        {
+            // Check if the environment variable (AppSetting) has this feature enabled
+            string envVal = environment.GetEnvironmentVariable(RpcWorkerConstants.FunctionsWorkerSharedMemoryDataTransferSettingName);
```
mhoeger, line 669: I'm not a fan of exposing what feel like implementation details to customers through app settings - is this a temporary change largely for testing and tweaking or is this the way of the future? Does shared memory data have a negative impact on perf if enabled by default for all sizes and is there a way we could avoid customers having to take actions / a dependency on this app setting?

src/WebJobs.Script/Workers/SharedMemoryDataTransfer/SharedMemoryManager.cs
```diff
@@ -0,0 +1,213 @@
...
+            else
+            {
+                _logger.LogWarning($"Cannot find SharedMemoryMap: {mapName}");
+                return false;
+            }
+        }
+
+        public bool IsSupported(object input)
+        {
+            if (input is byte[] arr)
```
mhoeger, line 177: Maybe out of scope, but I think a lot of folks would be happy to see this optimization on HTTP request bodies too (`input is HttpRequest request`)

mhoeger (commented):
Super excited for these changes!! A few comments / questions

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
pragnagopa: You are right! This is temporary. To give us a chance to test in prod and also select few customers who are willing to try. Idea is to default to shared memory after one or two releases.

src/WebJobs.Script/Workers/SharedMemoryDataTransfer/SharedMemoryManager.cs
pragnagopa: Good call out. Same here. For now scoping the change to binary data. Once we have initial implementation checked in and tested in prod - we can expand to HttpTriggers that send large data in the body - @gohar94 can you please open tracking issue for this.

gohar94: Done #6832

src/WebJobs.Script.Grpc/Extensions/ScriptInvocationContextExtensions.cs
```diff
@@ -9,13 +9,61 @@
...
 using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
+using Microsoft.Azure.WebJobs.Script.Workers.SharedMemoryDataTransfer;
 using Microsoft.Extensions.Logging;
 
 namespace Microsoft.Azure.WebJobs.Script.Grpc
 {
     internal static class ScriptInvocationContextExtensions
     {
-        public static async Task<InvocationRequest> ToRpcInvocationRequest(this ScriptInvocationContext context, ILogger logger, GrpcCapabilities capabilities)
+        private static async Task<RpcSharedMemory> SharedMemoryDataTransferAsync(ILogger logger, string invocationId, object value, ISharedMemoryManager sharedMemoryManager)
```
pragnagopa: rename to : `ToRpcSharedMemory` , Move `ToRpcSharedMemory` and any helpers to their own extensions class `SharedMemoryDataConversions`

pragnagopa: Mark this `internal` to allow for unit test

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -410,6 +416,24 @@ internal async Task SendInvocationRequest(ScriptInvocationContext context)
...
 
+        private object GetBindingData(ParameterBinding binding)
+        {
+            if (binding.RpcSharedMemory != null)
+            {
+                // Data was transferred by the worker using shared memory
+                string mapName = binding.RpcSharedMemory.Name;
+                long offset = binding.RpcSharedMemory.Offset;
+                long count = binding.RpcSharedMemory.Count;
+                // TODO gochaudh: InvokeResponse (caller of this method) seems like it can't be made async - confirm?
```
pragnagopa: You need to tweak the `toDictionary`  to take in an `async` value selector

src/WebJobs.Script/Workers/SharedMemoryDataTransfer/ISharedMemoryManager.cs
pragnagopa: in that case, remove `PutBytesAsync`

src/WebJobs.Script/Workers/SharedMemoryDataTransfer/ISharedMemoryManager.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    public interface ISharedMemoryManager
+    {
+        Task<SharedMemoryMetadata> PutBytesAsync(byte[] content);
+
+        Task<SharedMemoryMetadata> PutObjectAsync(object input);
+
+        Task<byte[]> GetBytesAsync(string mapName, long offset, long count);
+
+        bool IsSupported(object input);
```
pragnagopa, line 14: nit: Rename to `IsDatatypeSupported` also change the input to `Type`

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -420,14 +444,29 @@ internal void InvokeResponse(InvocationResponse invokeResponse)
...
-                        .ToDictionary(binding => binding.Name, binding => binding.Data.ToObject());
+                        .ToDictionary(binding => binding.Name, binding => GetBindingData(binding));
 
                     var result = new ScriptInvocationResult()
                     {
                         Outputs = bindingsDictionary,
                         Return = invokeResponse?.ReturnValue?.ToObject()
                     };
                     context.ResultSource.SetResult(result);
+
```
pragnagopa, line 474: move freeing and sending `closeSharedMemoryResourcesRequest` to a separate method. Also this method needs to be called only if shared memory is used by an invocation. Check if `SharedMemorymanager` is tracking this invocation before calling free.

pragnagopa: Also, Rename `closeSharedMemoryResourcesRequest` to `FreeSharedMemoryResourcesRequest` ?

pragnagopa (commented):
Added few more comments.  After addressing these comments:
- Close this PR and create PR from a branch on this repo -  to get basic validation with PR builds
- Please start adding unit tests.

I will spend more cycles on reviewing implemention for `SharedMemoryMap` and `MemoryMapFileAccessor`

src/WebJobs.Script/Workers/SharedMemoryDataTransfer/ISharedMemoryManager.cs
gohar94: This will not just be checking the type. It will check things like if min size exceeds the threshold or not (and maybe other things in the future) so keeping it generic and passing the full object.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
gohar94: Technically, it may not be freeing. It only closes the reference it has open. When we add some caching logic to this, we will potentially use this to make the worker close its references (but not free anything) since the host will have references to the same memory map (which we want to keep alive.)

I am going to change the message a little and call it `CloseSharedMemoryMaps` and pass it a list of maps we want to close.

gohar94:
Abandoning this one and moving to #6836

PR #6799, opened 2020-10-15
Blocking a restart from FileMonitoringService

### LINKED ISSUE #6669: Deployment of Functions using ARM template gives 'Bad Request' Internal Server Error

alrod:
resolves #6669

The fix discussed here:
https://github.com/Azure/azure-functions-host/pull/6753#discussion_r503459887

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/Controllers/FunctionsController.cs
+++ b/src/WebJobs.Script.WebHost/Controllers/FunctionsController.cs
@@ -65,14 +65,20 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Controllers
         [HttpPut]
         [Route("admin/functions/{name}")]
         [Authorize(Policy = PolicyNames.AdminAuthLevel)]
-        public async Task<IActionResult> CreateOrUpdate(string name, [FromBody] FunctionMetadataResponse functionMetadata)
+        public async Task<IActionResult> CreateOrUpdate(string name, [FromBody] FunctionMetadataResponse functionMetadata, [FromServices] IFileMonitoringService fileMonitoringService)
         {
             if (!Utility.IsValidFunctionName(name))
             {
                 return BadRequest($"{name} is not a valid function name");
             }
 
-            (var success, var configChanged, var functionMetadataResponse) = await _functionsManager.CreateOrUpdate(name, functionMetadata, Request);
+            bool success, configChanged;
+            FunctionMetadataResponse functionMetadataResponse;
+            using (fileMonitoringService.SuspendRestart())
+            {
+                (success, configChanged, functionMetadataResponse) = await _functionsManager.CreateOrUpdate(name, functionMetadata, Request);
+            }
+            await fileMonitoringService.ScheduleRestartAsync(false);
 
             if (success)
             {
@@ -159,7 +165,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Controllers
         [HttpDelete]
         [Route("admin/functions/{name}")]
         [Authorize(Policy = PolicyNames.AdminAuthLevel)]
-        public async Task<IActionResult> Delete(string name)
+        public async Task<IActionResult> Delete(string name, [FromServices] IFileMonitoringService fileMonitoringService)
         {
             (var found, var function) = await _functionsManager.TryGetFunction(name, Request);
             if (!found)
@@ -167,7 +173,13 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Controllers
                 return NotFound();
             }
 
-            (var deleted, var error) = await _functionsManager.TryDeleteFunction(function);
+            bool deleted;
+            string error;
+            using (fileMonitoringService?.SuspendRestart())
+            {
+                (deleted, error) = await _functionsManager.TryDeleteFunction(function);
+            }
+            await fileMonitoringService?.ScheduleRestartAsync(false);
 
             if (deleted)
             {
--- a/src/WebJobs.Script.WebHost/FileMonitoringService.cs
+++ b/src/WebJobs.Script.WebHost/FileMonitoringService.cs
@@ -21,7 +21,7 @@ using IApplicationLifetime = Microsoft.AspNetCore.Hosting.IApplicationLifetime;
 
 namespace Microsoft.Azure.WebJobs.Script.WebHost
 {
-    public class FileMonitoringService : IHostedService, IDisposable
+    public class FileMonitoringService : IFileMonitoringService, IDisposable
     {
         private readonly ScriptJobHostOptions _scriptOptions;
         private readonly IScriptEventManager _eventManager;
@@ -39,10 +39,11 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
         private AutoRecoveringFileSystemWatcher _diagnosticModeFileWatcher;
         private FileWatcherEventSource _fileEventSource;
         private bool _shutdownScheduled;
-        private int _restartRequested;
+        private long _restartRequested;
         private bool _disposed = false;
         private bool _watchersStopped = false;
         private object _stopWatchersLock = new object();
+        private long _suspensionRequestsCount = 0;
 
         public FileMonitoringService(IOptions<ScriptJobHostOptions> scriptOptions, ILoggerFactory loggerFactory, IScriptEventManager eventManager, IApplicationLifetime applicationLifetime, IScriptHostManager scriptHostManager, IEnvironment environment)
         {
@@ -96,6 +97,41 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
         {
             StopFileWatchers();
             return Task.CompletedTask;
+        }
+
+        public IDisposable SuspendRestart()
+        {
+            return new SuspendRestartRequest(this);
+        }
+
+        public void ResumeRestart()
+        {
+            Interlocked.Decrement(ref _suspensionRequestsCount);
+            _typedLogger.LogDebug($"Resuming restart. ({_suspensionRequestsCount} requests).");
+        }
+
+        public async Task ScheduleRestartAsync(bool shutdown)
+        {
+            if (shutdown)
+            {
+                _shutdownScheduled = true;
+            }
+
+            if (Interlocked.Read(ref _suspensionRequestsCount) > 0)
+            {
+                _logger.LogDebug("Restart requested while currently suspended. Ignoring request.");
+            }
+            else
+            {
+                if (shutdown || _shutdownScheduled)
+                {
+                    _shutdown();
+                }
+                else
+                {
+                    await _restart();
+                }
+            }
         }
 
         /// <summary>
@@ -300,19 +336,6 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
             return string.Empty;
         }
 
-        private async Task ScheduleRestartAsync(bool shutdown)
-        {
-            if (shutdown)
-            {
-                _shutdownScheduled = true;
-                _shutdown();
-            }
-            else
-            {
-                await _restart();
-            }
-        }
-
         private Task RestartAsync()
         {
             if (!_shutdownScheduled && Interlocked.Exchange(ref _restartRequested, 1) == 0)
@@ -367,6 +390,29 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
                         File.Delete(path);
                     }
                 }, maxRetries: 3, retryInterval: TimeSpan.FromSeconds(1));
+            }
+        }
+
+        private class SuspendRestartRequest : IDisposable
+        {
+            private FileMonitoringService _fileMonitoringService;
+            private bool _disposed = false;
+
+            public SuspendRestartRequest(FileMonitoringService fileMonitoringService)
+            {
+                _fileMonitoringService = fileMonitoringService;
+                Interlocked.Increment(ref _fileMonitoringService._suspensionRequestsCount);
+                _fileMonitoringService._typedLogger.LogDebug($"Entering restart suspension scope. ({_fileMonitoringService._suspensionRequestsCount} requests).");
+            }
+
+            public void Dispose()
+            {
+                if (!_disposed)
+                {
+                    Interlocked.Decrement(ref _fileMonitoringService._suspensionRequestsCount);
+                    _fileMonitoringService._typedLogger.LogDebug($"Exiting restart suspension scope. ({_fileMonitoringService._suspensionRequestsCount} requests).");
+                    _disposed = true;
+                }
             }
         }
     }
--- a/src/WebJobs.Script.WebHost/IFileMonitoringService.cs
+++ b/src/WebJobs.Script.WebHost/IFileMonitoringService.cs
@@ -1,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Hosting;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost
+{
+    public interface IFileMonitoringService : IHostedService
+    {
+        IDisposable SuspendRestart();
+
+        void ResumeRestart();
+
+        Task ScheduleRestartAsync(bool shutdown);
+    }
+}
--- a/src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
+++ b/src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using System.Linq;
 using Microsoft.AspNetCore.Mvc.ApplicationParts;
 using Microsoft.Azure.WebJobs.Host.Config;
 using Microsoft.Azure.WebJobs.Host.Executors;
@@ -130,7 +131,8 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
                     services.AddSingleton<IEventCollectorProvider, FunctionInstanceLogCollectorProvider>();
 
                     // Hosted services
-                    services.TryAddEnumerable(ServiceDescriptor.Singleton<IHostedService, FileMonitoringService>());
+                    services.AddSingleton<IFileMonitoringService, FileMonitoringService>();
+                    services.TryAddEnumerable(ServiceDescriptor.Singleton<IHostedService, IFileMonitoringService>(p => p.GetService<IFileMonitoringService>()));
 
                     ConfigureRegisteredBuilders(services, rootServiceProvider);
                 });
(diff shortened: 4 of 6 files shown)


src/WebJobs.Script.WebHost/Controllers/FunctionsController.cs
```diff
@@ -65,22 +66,32 @@ public async Task<IActionResult> Get(string name)
         [HttpPut]
         [Route("admin/functions/{name}")]
         [Authorize(Policy = PolicyNames.AdminAuthLevel)]
-        public async Task<IActionResult> CreateOrUpdate(string name, [FromBody] FunctionMetadataResponse functionMetadata)
+        public async Task<IActionResult> CreateOrUpdate(string name, [FromBody] FunctionMetadataResponse functionMetadata, [FromServices] IEnumerable<IHostedService> hostedServices)
         {
-            if (!Utility.IsValidFunctionName(name))
+            FileMonitoringService fileMonitoringService = (FileMonitoringService)hostedServices.SingleOrDefault(x => x is FileMonitoringService);
```
mathewc: This will never be null, right? Assuming so, rather than doing a manual lookup like this, can't you just take FileMonitoringService as a ctor DI param?

src/WebJobs.Script.WebHost/Controllers/FunctionsController.cs
```diff
@@ -65,22 +66,32 @@ public async Task<IActionResult> Get(string name)
...
+            try
             {
-                return BadRequest($"{name} is not a valid function name");
-            }
+                fileMonitoringService?.BlockRestartOrShutdown();
```
mathewc: As a programming model, can we shoot for the below? The service can return a disposable that will handle everything internally. This will make it cleaner and reduce mistakes, keeping the logic centralized.

```csharp
using (_fileMonitoringService.SuspendRestart())
{
     ...
}
```

src/WebJobs.Script.WebHost/FileMonitoringService.cs
```diff
@@ -86,6 +87,26 @@ internal ImmutableArray<string> GetDirectorySnapshot()
             return ImmutableArray<string>.Empty;
         }
 
+        public void BlockRestartOrShutdown()
+        {
+            lock (_cde)
+            {
+                if (_cde.CurrentCount == 0)
```
mathewc: Countdownevent public members are threadsafe, so you should be able to simplify this code to just a single line _cde.AddCount()? Seems to me that if CurrentCount is 0, we don't need to reset, but to just call AddCount?

src/WebJobs.Script.WebHost/FileMonitoringService.cs
```diff
@@ -43,6 +43,7 @@ public class FileMonitoringService : IHostedService, IDisposable
         private bool _disposed = false;
         private bool _watchersStopped = false;
         private object _stopWatchersLock = new object();
+        private CountdownEvent _cde = new CountdownEvent(0);
```
mathewc: nit: _restartCountdownEvent

src/WebJobs.Script.WebHost/FileMonitoringService.cs
```diff
@@ -302,6 +323,9 @@ internal static string GetRelativeDirectory(string path, string scriptRoot)
 
         private async Task ScheduleRestartAsync(bool shutdown)
         {
+            // will return when the count reaches 0 or after 1 minute
+            _cde.Wait(60000);
```
mathewc: We really don't want to block a thread on this. Unfortunately there isn't an async version of this class. Might be simpler for us to to implement this ourselves via a simple syncronized counter, and use of our Utility.DelayAsync helper, to delay (without blocking) until the counter reaches 0.

src/WebJobs.Script.WebHost/Controllers/FunctionsController.cs
fabiocav: Yes. That was the original idea I had proposed. This method can return an `IDisposable` that we can use in this scope. Also, since this is with the file monitoring service, a better and more appropriate name would be `SuspendNotifications`

fabiocav: We should also not ignore and continue here if the service is null.

src/WebJobs.Script.WebHost/FileMonitoringService.cs
fabiocav: I think this is really because of the usage pattern here. Though the members are thread safe, checking the property and subsequently calling method isn't. 

I think we can simplify this with a count that is atomically incremented/decremented when the notifications are suspended and resumed. Then the notification handler can notify conditional on that count.

src/WebJobs.Script.WebHost/FileMonitoringService.cs
fabiocav: There's no need to wait here. If we reach this and notifications are suspended, something didn't work as expected. We don't want to block threads and have them all fire here when the notifications are resumed.

If restart/notifications are suspended, we should just return here.

src/WebJobs.Script.WebHost/Controllers/FunctionsController.cs
```diff
@@ -65,22 +66,32 @@ public async Task<IActionResult> Get(string name)
         [HttpPut]
         [Route("admin/functions/{name}")]
         [Authorize(Policy = PolicyNames.AdminAuthLevel)]
-        public async Task<IActionResult> CreateOrUpdate(string name, [FromBody] FunctionMetadataResponse functionMetadata)
+        public async Task<IActionResult> CreateOrUpdate(string name, [FromBody] FunctionMetadataResponse functionMetadata, [FromServices] IEnumerable<IHostedService> hostedServices)
```
fabiocav: Instead of taking the entire collection of hosted services, let's introduce an interface for the file monitoring service and take a dependency on that instead.

fabiocav (changes requested):
The approach here is definitely better than the previous.

Left a few comments, but in addition to what is in the code, let's make sure we add a good amount of logging so we have traces for conditions like:

1 - Notifications are suspended
2 - Notifications are resumed
3 - Restart is triggered when notifications are suspended
4 - Resume is called bringing the suspension count to a negative number (error condition)
5 - Other logging for the new logic

fabiocav:
Can you please also go through the PR checklist and check what is appropriate?

src/WebJobs.Script.WebHost/Controllers/FunctionsController.cs
alrod: we still need `FileMonitoringService` added as `IHostedService` to be a background task. I tried couple of ways:

1.  
```
                    services.AddSingleton<FileMonitoringService>();
                    services.AddSingleton<IHostedService, FileMonitoringService>(serviceProvider =>
                    {
                        return serviceProvider.GetService<FileMonitoringService>();
                    });
```
in this case we need to remove the line from DependencyValidator as FileMonitoringService as IHostedService will be resolved letter.
https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.WebHost/DependencyInjection/DependencyValidator/DependencyValidator.cs#L48
2. Add a helper class - BackgroundServiceStarter:
https://stackoverflow.com/questions/51254053/how-to-inject-a-reference-to-a-specific-ihostedservice-implementation

So I decided to extract the service from` IEnumerable<IHostedService>`. Please advise which direction should we go.

src/WebJobs.Script.WebHost/Controllers/FunctionsController.cs
alrod: we still need FileMonitoringService added as IHostedService to be a background task. See the comment above.

fabiocav: Comment where? The service would be registered as both, a hosted service, for background processing, as well as a file monitoring service.

src/WebJobs.Script.WebHost/Controllers/FunctionsController.cs
fabiocav: Oh I see the comment (GitHub put it at the bottom)

In addition to the resolution as a hosted service, you can add it as an `IFileMonitoringService` as well.

src/WebJobs.Script.WebHost/FileMonitoringService.cs
alrod: Calling `Reset ` is required before using the `CountdownEvent`, otherwise we will get `ArgumentOutOfRangeException `- signalCount is less than or equal to 0. Also Reset is not threadsafe - https://docs.microsoft.com/en-us/dotnet/api/system.threading.countdownevent?view=netcore-3.1#thread-safety

fabiocav: We shouldn't need to use a CountDownEvent here....

src/WebJobs.Script.WebHost/FileMonitoringService.cs
alrod: @fabiocav, I do not quiet understand "notifications are suspended". We are processing a file change, initiate restart flow and wait right before the restart until counter is 0. Can you please clarify?

alrod: @fabiocav, does "suspending notification" mean Initialze/Stop FileWatchers?

mathewc: You shouldn't have to actually stop filewatchers. You can just track if a restart was requested during the suspension, and if it was when we resume, you'd invoke it then.

src/WebJobs.Script.WebHost/IFileMonitoringService.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.Hosting;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost
+{
+    public interface IFileMonitoringService : IHostedService
```
fabiocav, line 10: This should also expose an API to resume the restart monitoring, even if that's provided by the disposable result.

src/WebJobs.Script.WebHost/FileMonitoringService.cs
```diff
@@ -369,5 +382,41 @@ internal static async Task SetAppOfflineState(string rootPath, bool offline)
...
+        private class SuspendRestartRequest : IDisposable
+        {
+            private FileMonitoringService _fileMonitoringService;
+
+            public SuspendRestartRequest(FileMonitoringService fileMonitoringService)
+            {
+                _fileMonitoringService = fileMonitoringService;
+                if (Interlocked.Increment(ref _fileMonitoringService._suspensionRequestsCount) == 1)
+                {
+                    _fileMonitoringService._typedLogger.LogDebug("Restart is suspended");
```
fabiocav: Can we update the log message to something like `"Entering restart suspension scope. ({0} requests)."`, showing the number of suspension requests/subscribers

fabiocav: We can move this logic to the FileMonitoringService, so this type just calls methods on it.

src/WebJobs.Script.WebHost/FileMonitoringService.cs
```diff
@@ -369,5 +382,41 @@ internal static async Task SetAppOfflineState(string rootPath, bool offline)
...
+            public void Dispose()
+            {
+                if (Interlocked.Decrement(ref _fileMonitoringService._suspensionRequestsCount) == 0)
+                {
+                    _fileMonitoringService._typedLogger.LogDebug("Restart is resumed");
+                    if (Interlocked.Read(ref _fileMonitoringService._restartRequested) == 1)
+                    {
+                        _fileMonitoringService._typedLogger.LogDebug("Initiate force restart as it was triggerd during the suspension");
+                        // Blocking call as we can't use async call in Dispose method
+                        _fileMonitoringService._scriptHostManager.RestartHostAsync().GetAwaiter().GetResult();
```
fabiocav: This differs a bit from the behavior we had discussed. Would the caller expect a restart to happen if something would have triggered it while it was suspended? I think this would come as a surprise and would also limit the caller's choice as to when/if a restart should happen. 

I think we should just resume watching and not restart. It would be up to the caller to do so.

src/WebJobs.Script.WebHost/FileMonitoringService.cs
```diff
@@ -369,5 +382,41 @@ internal static async Task SetAppOfflineState(string rootPath, bool offline)
...
+                    }
+                }
+                else
+                {
+                    // Error case
+                    if (_fileMonitoringService._suspensionRequestsCount < 0)
```
fabiocav: Should we even allow that?

src/WebJobs.Script.WebHost/FileMonitoringService.cs
```diff
@@ -317,7 +323,14 @@ private Task RestartAsync()
         {
             if (!_shutdownScheduled && Interlocked.Exchange(ref _restartRequested, 1) == 0)
             {
-                return _scriptHostManager.RestartHostAsync();
+                if (Interlocked.Read(ref _suspensionRequestsCount) == 0)
```
fabiocav: Looks like we're still processing notifications and just preventing the restart. Should we stop the notifications instead, like we had discussed?

alrod: In case we suspend all notifications we loose all logs for the files/directories changes.

src/WebJobs.Script.WebHost/FileMonitoringService.cs
alrod: I will remove the code in next iteration and decrease `suspensionRequestCount` only if it > 0

src/WebJobs.Script.WebHost/FileMonitoringService.cs
alrod: this make sense so caller code will be changed to:
```
using(fileMonitoringService?.SuspendRestart()) // increment suspensionRequestCount
{
..
} // decrement suspensionRequestCount
await fileMonitoringService?.TryRestartAsyc()
```

`TryRestartAsync` will restart only if `suspensionRequestCount== 0`

src/WebJobs.Script.WebHost/FileMonitoringService.cs
fabiocav: I see. Makes sense.

src/WebJobs.Script.WebHost/FileMonitoringService.cs
```diff
@@ -317,7 +323,14 @@ private Task RestartAsync()
...
+                {
+                    return _scriptHostManager.RestartHostAsync();
+                }
+                else
+                {
+                    _logger.LogDebug("Restart is triggered when notifications are suspended");
```
fabiocav: nit: The message should be consistent with what we call the API, in the current case, if the API is called `SuspendRestart` the message should be something like: `Restart requested while currently suspended. Ignoring request.`

src/WebJobs.Script.WebHost/IFileMonitoringService.cs
alrod: I do not know how we are going to use the method assuming we can call `Dispose` from the code. Should it just decrease `suspensionRequestCount `or set it to 0?

alrod: I added `ResumeRestart` method

fabiocav (commented):
The changes look better, but I'm curious; do we not have to prevent shutdowns here as well? If one of the files is a DLL or something that would trigger a hard process shutdown, the current changes don't handle that, correct?

src/WebJobs.Script.WebHost/FileMonitoringService.cs
```diff
@@ -98,6 +99,41 @@ public Task StopAsync(CancellationToken cancellationToken)
...
+                _shutdownScheduled = true;
+            }
+
+            if (Interlocked.Read(ref _suspensionRequestsCount) > 0)
+            {
+                _logger.LogDebug("Restart requested while currently suspended. Ignoring request.");
+            }
+            else
+            {
+                if (shutdown || _shutdownScheduled)
```
fabiocav, line 126: Why is the logic being updated to trigger a shutdown if `_shutdownScheduled` is true? If anything, we should avoid duplicate calls here, correct?

fabiocav: Shutdown scheduled should essentially short circuit all of this.

src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
```diff
@@ -131,6 +132,7 @@ public static class WebScriptHostBuilderExtension
 
                     // Hosted services
                     services.TryAddEnumerable(ServiceDescriptor.Singleton<IHostedService, FileMonitoringService>());
+                    services.AddSingleton<IFileMonitoringService, FileMonitoringService>(serviceProvider => (FileMonitoringService)serviceProvider.GetServices<IHostedService>().FirstOrDefault(x => x is FileMonitoringService));
```
fabiocav: The resolution here doesn't have to iterate over the hosted services. You can simply retrieve the registered `FileMonitoringService` instance in both cases (otherwise, we're just pushing the lookup cost up)

src/WebJobs.Script.WebHost/FileMonitoringService.cs
alrod: this is for scenario:
1. SuspendRestart (shutdown)
2. SuspenRestart (restart)
3. ResumeRestart (shutdown)
4. ResumeRestart (restart)

Shutdown must be performed after step 4.

src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
alrod: I suppose we you are talking about the code:

```
services.TryAddSingleton<IHostedService>(p => (IHostedService)p.GetService<IFileMonitoringService>());
```
I this case we need to remove:
https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.WebHost/DependencyInjection/DependencyValidator/DependencyValidator.cs#L48

Are we ok with that?

fabiocav: We can adjust the validation as needed. The goal here is to register the `FileMonitoringService` as an `IFileMonitoringService` and that same instance as a hosted service, resolving the registered instance directly.

fabiocav (dismissed):

fabiocav (approved):

PR #6803, opened 2020-10-16
Enforce ordering of functions in synctriggers payload

### LINKED ISSUE #6686: Enforce ordering of functions in synctriggers payload
EnumerateDirectories() doesnt guarantee ordering. Since the list is used compute hash this can lead to false positives in hash comparison causing unnecessary synctrigger calls

https://github.com/Azure/azure-functions-host/blob/bbb3d257dd3480f5a94b5b2776d1d71aa8c25fb3/src/WebJobs.Script/Host/FunctionMetadataProvider.cs#L44

https://github.com/Azure/azure-functions-host/blob/621df2c50b5ab96dbd403d173161a4d7820d02c0/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs#L200

balag0:
Background synctriggers task compares hashes of current and last known synctriggers payload to avoid unnecessary /synctriggers call to update the site metadata. Currently the ordering of functions in Synctriggers payload is non deterministic since it relies EnumerateDirectories() which is implemented differently on Linux and Windows. This leads to unnecessary updates to the site metadata if the function app has more than 1 function even though the triggers themselves haven't been updated. Fix is to sort the functions before calculating the hash.

Fixes https://github.com/Azure/azure-functions-host/issues/6686

resolves https://github.com/Azure/azure-functions-host/issues/6686

### Pull request checklist

    * [X ] Otherwise: Backport tracked by issue/PR https://github.com/Azure/azure-functions-host/issues/6686
* [X ] I have added all required tests (Unit tests, E2E tests)

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script/Host/FunctionMetadataManager.cs
+++ b/src/WebJobs.Script/Host/FunctionMetadataManager.cs
@@ -157,7 +157,7 @@ namespace Microsoft.Azure.WebJobs.Script
                 Errors = _functionErrors.Where(kvp => functionsAllowList.Any(functionName => functionName.Equals(kvp.Key, StringComparison.CurrentCultureIgnoreCase))).ToImmutableDictionary(kvp => kvp.Key, kvp => kvp.Value.ToImmutableArray());
             }
 
-            return functionMetadataList.ToImmutableArray();
+            return functionMetadataList.OrderBy(f => f.Name, StringComparer.OrdinalIgnoreCase).ToImmutableArray();
         }
 
         internal bool IsScriptFileDetermined(FunctionMetadata functionMetadata)
--- a/test/WebJobs.Script.Tests.Integration/Management/FunctionsSyncManagerTests.cs
+++ b/test/WebJobs.Script.Tests.Integration/Management/FunctionsSyncManagerTests.cs
@@ -715,8 +715,8 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.Managment
                         return new[]
                         {
                             Path.Combine(rootPath, "bin"),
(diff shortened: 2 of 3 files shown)


src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -267,7 +267,7 @@ internal async Task<CloudBlockBlob> GetHashBlobAsync()
         public async Task<SyncTriggersPayload> GetSyncTriggersPayload()
```
balag0: Note: Since GetSyncTriggersPayload() doesn't expose the new sortFunctionsMetadata param, triggers will be sorted for both background synctriggers and the explicit api call even though the background task is the one the issue was originally opened for.

src/WebJobs.Script/Extensions/FunctionMetadataNameComparer.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using FunctionMetadata = Microsoft.Azure.WebJobs.Script.Description.FunctionMetadata;
+
+namespace Microsoft.Azure.WebJobs.Script.Extensions
+{
+    public class FunctionMetadataNameComparer : IComparer<FunctionMetadata>
```
mathewc: Recommend you just remove this class and simply sort via `functionMetadataList.OrderBy(p => p.Name).ToImmutableArray();`

mathewc (approved):

Issue #6704, opened 2020-09-25
[Rpc][MultipleWorkers] Add check for memory consumption before spawning new processes


pragnagopa:
cc @AnatoliB @vrdmr 

This is related to https://github.com/Azure/azure-functions-host/issues/6671

Add check for memory consumption before spawning new worker processes.

fabiocav:
@pragnagopa could you share more details about:

- Issues this would help address (any evidence of crashes OOM exceptions from creating additional worker processes?)
- Details on what that check would validate before launching the process

AnatoliB:
> * Issues this would help address (any evidence of crashes OOM exceptions from creating additional worker processes?)

@fabiocav We may not be getting enough signals on these crashes today because:
- By default, this feature is turned off: https://github.com/Azure/azure-functions-host/issues/7499
- By default, FUNCTIONS_WORKER_PROCESS_COUNT is 1

As a result, only users who decided to experiment with FUNCTIONS_WORKER_PROCESS_COUNT will run into these crashes. These users already expect potential troubles from increasing FUNCTIONS_WORKER_PROCESS_COUNT, so they probably just decrease the value instead of complaining to us. Enabling https://github.com/Azure/azure-functions-host/issues/7499 by default (which we intend to do eventually) will lead to creating more processes automatically, so we should anticipate OOM to happen more often.

/cc: @alrod

pragnagopa:
@fabiocav / @liliankasem  - removed "Needs:Author Feedback" label for this issue. Thanks @AnatoliB  for adding details.

liliankasem:
Is it worth closing this in favour of #7499, or is 7499 acting as an epic we can link issues to?

fabiocav:
Based on our discussion this week, I'll say yes. This should be tracked by the `Intelligent concurrency for language workers` issue you've linked (#7499).

liliankasem:
@fabiocav I had a sync with Anatoli and Alexey about this and here's the TLDR;

- Intelligent concurrency as a feature has already been implemented and is in preview (Alexey will update #7499 with the current progress)
- The team is waiting for feedback to understand how many customers will be impacted by this through CRIs/customer complaints
- Anatoli does not believe we can GA without the memory checks being implemented, so either way regardless of Intelligent concurrency, this should be implemented
- There are plans to eventually do this but it is low priority at the moment for their team as they are waiting for customer feedback to roll in

[...]

PR #6806, opened 2020-10-16
[Rpc]Add processOptions in worker.config

### LINKED ISSUE #6671: Add support in worker.config to specify number of worker process count or opt-in to set worker count to number of core available on the VM
This came up while investigating some CRIs. The current default is one no matter how many cores are available on the VM.
And customers should be aware of the FUNCTIONS_WORKER_PROCESS_COUNT app setting to be able to increase their number of language workers for their specific scenarios.

For better throughput, the default should be set to the number of cores available. For Consumption  this will still be one but this will be different for Premium and App Service plans.

pragnagopa:

resolves #6671 

### Pull request checklist

### Additional information

Additional PR information

Added support for customize multiple worker processes in `worker.config.json`

```json
{
  "description": {
    "language": "node",
    "extensions": [ ".js" ],
    "defaultExecutablePath": "node",
    "defaultWorkerPath": "dist/src/nodejsWorker.js"
  },
  "processOptions": {
    "processCount": 5,
    "maxProcessCount": 12,
    "processStartUpInterval": "00:00:05"
  }
}
```

If `processOptions` section is not specified,  following defaults will be applied:

`processCount` :  1
`maxProcessCount` : 10
`processStartUpInterval` : 00:00:10

### DIFF
--- a/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
+++ b/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
@@ -28,8 +28,6 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
         private readonly IRpcWorkerChannelFactory _rpcWorkerChannelFactory;
         private readonly IEnvironment _environment;
         private readonly IApplicationLifetime _applicationLifetime;
-        private readonly int _debounceSeconds = 10;
-        private readonly int _maxAllowedProcessCount = 10;
         private readonly TimeSpan _shutdownTimeout = TimeSpan.FromSeconds(10);
         private readonly TimeSpan thresholdBetweenRestarts = TimeSpan.FromMinutes(WorkerConstants.WorkerRestartErrorIntervalThresholdInMinutes);
 
@@ -50,6 +48,7 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
         private IEnumerable<FunctionMetadata> _functions;
         private ConcurrentStack<WorkerErrorEvent> _languageWorkerErrors = new ConcurrentStack<WorkerErrorEvent>();
         private CancellationTokenSource _processStartCancellationToken = new CancellationTokenSource();
+        private int _debouceMilliSeconds = (int)TimeSpan.FromSeconds(10).TotalMilliseconds;
 
         public RpcFunctionInvocationDispatcher(IOptions<ScriptJobHostOptions> scriptHostOptions,
             IMetricsLogger metricsLogger,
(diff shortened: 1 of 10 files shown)


pragnagopa:
if any of you have suggestions for better section name for `processOptions` - please let me know.

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -50,6 +48,7 @@ internal class RpcFunctionInvocationDispatcher : IFunctionInvocationDispatcher
         private IEnumerable<FunctionMetadata> _functions;
         private ConcurrentStack<WorkerErrorEvent> _languageWorkerErrors = new ConcurrentStack<WorkerErrorEvent>();
         private CancellationTokenSource _processStartCancellationToken = new CancellationTokenSource();
+        private int _debouceMilliSeconds = (int)TimeSpan.FromSeconds(10).TotalMilliseconds;
```
AnatoliB: Typo: "debouNce"

AnatoliB:
I have a concern that does not directly block this specific PR, but let's keep this in mind before workers actually start opting in. We previously talked about adding more safeguards before spawning more processes by default, such as checking memory consumption. I think we should introduce this check _before_ workers start opting in. Without this check, opting in cannot be a per-worker or per-language decision: if an app consumes a lot of memory, this have little to do with the language choice in many cases. Are we going to have this check soon?

pragnagopa:
> checking memory consumption. I think we should introduce this check before workers start opting in. Without this check, opting in cannot be a per-worker or per-language decision: if an app consumes a lot of memory, this have little to do with the language choice in many cases. Are we going to have this check soon?

Tracking issue: https://github.com/Azure/azure-functions-host/issues/6704 . But I do not have an ETA yet.

Yes, I agree it makes sense to wait for the check to be added before workers can opt-in. I believe this PR would allow @vrdmr to run perf tests customizing number of workers and debouce interval. I would like get feedback from @vrdmr before prioritizing issue #6704

AnatoliB (approved):

vrdmr:
> Yes, I agree it makes sense to wait for the check to be added before workers can opt-in. I believe this PR would allow @vrdmr to run perf tests customizing the number of workers and debounce intervals. I would like to get feedback from @vrdmr before prioritizing issue #6704

Thanks, @pragnagopa for the changes. This definitely helps me to do tests. The current configuration still requires hardcoding the `processCount` and we cannot choose based on the # cores on the node. Is there a way would you recommend to generate the process counts based on the number of cores?

pragnagopa:
Missed adding this as part of PR description. `processOptions` exposes `SetWorkerCountToNumberOfCpuCores`  property. Setting this `true` will set the number of workers to cpu cores.

Here a sample worker.config.json
```json
{
  "description": {
    "language": "node",
    "extensions": [ ".js" ],
    "defaultExecutablePath": "node",
    "defaultWorkerPath": "dist/src/nodejsWorker.js"
  },
  "processOptions": {
    "setWorkerCountToNumberOfCpuCores": true,
    "processStartUpInterval": "00:00:05"
  }
}
```

vrdmr:
Awesome! 👍 

Thanks a lot, @pragnagopa for the change. Cannot wait to try this out now.

PR #6808, opened 2020-10-17
Improve functions startup time when warmup triggers are provided.

safihamid:
This PR will move the /admin/warmup call completely within the functions execution pipeline rather than being a separate controller
and will avoid all the JIT cost that incur during the first real function execution even when warmup triggers are present.

With this PR and when a warmup trigger function was provided, the cold start time of the first real function execution went down from ~600 msec to ~80 sec on my stamp on a large worker.

resolves #issue_for_this_pr

### Pull request checklist

### Additional information
This PR is dependent on https://github.com/Azure/azure-webjobs-sdk-extensions/pull/668 to be merged first


src/WebJobs.Script.WebHost/Middleware/FunctionInvocationMiddleware.cs
```diff
@@ -186,7 +186,7 @@ private async Task<bool> AuthenticateAndAuthorizeAsync(HttpContext context, Func
 
         internal static bool RequiresAuthz(HttpRequest request, FunctionDescriptor descriptor)
         {
-            if (descriptor.Metadata.IsProxy())
+            if (descriptor.Metadata.IsProxy() || descriptor.IsWarmupFunction())
```
mathewc, line 194: @ankitkumarr Warmup functions are always anonymous? Users can't configure auth level in function.json?

src/WebJobs.Script.WebHost/WebScriptHostHttpRoutesManager.cs
```diff
@@ -87,6 +96,17 @@ public void InitializeHttpFunctionRoutes(IScriptJobHost host)
...
 
+            if (warmupRouteBuilder != null)
+            {
+                // Adding the default admin/warmup route when no warmup function is present
+                if (warmupRouteBuilder.Count == 0)
+                {
+                    warmupRouteBuilder.MapFunctionRoute(string.Empty, "warmup", string.Empty);
+                }
+                IRouter warmupRouter = warmupRouteBuilder.Build();
+                _router.AddCustomRoutes(warmupRouter);
```
mathewc: As mentioned in my comment in PR https://github.com/Azure/azure-webjobs-sdk-extensions/pull/668, I'm concerned that this is a breaking change to the public IWebJobsRouter interface

src/WebJobs.Script.WebHost/Middleware/FunctionInvocationMiddleware.cs
safihamid: yes they are anonymous. and they are only being called by App Service FEs and Data roles. They are not external http triggers.

src/WebJobs.Script.WebHost/WebScriptHostHttpRoutesManager.cs
safihamid: Isn't this ok to update the interface under a new version? here is another PR which we updated this interface in the past.
https://github.com/Azure/azure-webjobs-sdk-extensions/pull/443

 I am not sure if there is any cleaner way to add these warmup routes to the routes collection and not have them get mixed up with external http trigger and proxies routes. please take a look at the tests, we should not mix these admin/warmup calls with api/warmup http trigger or proxies if customers happen to have them or with any warmup function which is not a warmup trigger function.

mathewc: It's safe to break in a new major version. That commit you link to was 2 years ago when the extension was still in beta I believe and didn't have its own nuget package, so we hadn't yet released the 3.0 major version with it. These Warmup functions are actual functions, implemented by the user, so why can't we add these via the existing AddFunctionRoutes API?

Specifically, you can create your own route builder like you do above, and after building, rather than call AddCustomRoutes, just make a second call _router.AddFunctionRoutes(route, null) to add these. This way we can revert the AddCustomRoutes change.

src/WebJobs.Script.WebHost/Routing/ScriptRouteHandler.cs
```diff
@@ -37,8 +42,25 @@ public async Task InvokeAsync(HttpContext context, string functionName)
                 ProxyFunctionExecutor proxyFunctionExecutor = new ProxyFunctionExecutor(_scriptHost);
                 context.Items.TryAdd(ScriptConstants.AzureProxyFunctionExecutorKey, proxyFunctionExecutor);
             }
+            else if (_isWarmup)
+            {
+                // warmup function will get executed just once for the process.
+                if (Interlocked.CompareExchange(ref _warmupExecuted, 1, 0) != 0)
```
mathewc, line 50: You're maintaining the logic that existed in the controller method I see. @ankitkumarr, why do we need to enforce a single execution? I'd say if someone calls again for some reason, we just invoke again.

src/WebJobs.Script.WebHost/Features/FunctionExecutionFeature.cs
```diff
@@ -62,10 +62,18 @@ public async Task ExecuteAsync(HttpRequest request, CancellationToken cancellati
             }
 
             var sw = Stopwatch.StartNew();
-            var arguments = new Dictionary<string, object>()
+
+            var arguments = new Dictionary<string, object>();
+            if (_descriptor.IsWarmupFunction())
+            {
+                ParameterDescriptor inputParameter = _descriptor.Parameters.First(p => p.IsTrigger);
```
mathewc: `_descriptor.TriggerParameter.Name` like we use below doesn't work for warmup functions?

src/WebJobs.Script.WebHost/Routing/ScriptRouteHandler.cs
safihamid: @mathewc again nobody can call warmup functions from outside, they are not http triggers. the idea is user adds all initialization of heavy objects and what not in this warmup function and it only gets called ONCE for each process. Also FEs and Data roles will call this /admin/warmup end point to make sure host is up and running. FE by default will call it every few minutes as part of AlwaysOn and we need to make sure the call is very light and returns quickly except the very first time that we do initializations. This behavior of warmup functions is documented as well.

src/WebJobs.Script.WebHost/WebScriptHostHttpRoutesManager.cs
safihamid: I see, yes makes sense now. sent a PR to revert changes to the interface. will bring that here then.

src/WebJobs.Script.WebHost/Routing/ScriptRouteHandler.cs
mathewc: @ankitkumarr - when customers develop/test their warmup function, they presumably use portal/cli to invoke/test the function. If the existing behavior is to only run it once, they can only run/debug once, then they have to restart their app to try again? Seems artificial to me.

safihamid: Not sure if anyone has ever complained about invoke/testing of this but if it's ok to break the existing behavior, maybe we could allow this only once if environment is app service and otherwise remove this requirement for cli scenarios and local testing and debugging. would still fail for portal but works for cli.

src/WebJobs.Script.WebHost/Features/FunctionExecutionFeature.cs
safihamid: _descriptor.TriggerParameter.Name is not the issue here. it's the type.
basically this line adds HttpRequest object but we need the WamupContext object here.

            else
            {
                arguments.Add(_descriptor.TriggerParameter.Name, request);
            }

src/WebJobs.Script.WebHost/Routing/ScriptRouteHandler.cs
ankitkumarr: > why do we need to enforce a single execution? I'd say if someone calls again for some reason, we just invoke again.

What Hamid said and also when we built this, there was an issue where sometimes platform called `/admin/warmup` multiple times during startup. To avoid users incurring costs for execution and avoiding duplicate calls, we added this.

> when customers develop/test their warmup function, they presumably use portal/cli to invoke/test the function. If the existing behavior is to only run it once, they can only run/debug once, then they have to restart their app to try again? Seems artificial to me.

Yeah that makes sense. Right now, you can't even execute once outside of Azure as we have a check to only run when in Elastic Premium Azure environment. I think that we can expose a way to run locally and execute it. Probably can take that separately (in a different PR) as that would be a separate discussion.

src/WebJobs.Script.WebHost/Features/FunctionExecutionFeature.cs
mathewc: My point was that it seems to get the name you can just use `_descriptor.TriggerParameter.Name` since that is the cached value you're re-looking up it seems.

safihamid: oh yes, updated it, thanks

ankitkumarr (approved):

PR #6818, opened 2020-10-20
Fixing MetricsEventManager memory leak

### LINKED ISSUE #6813: MetricEventManager - Memory Leak
During investigations of memory issues on one of our azure functions applications after taking memory dumps we found that the memory usage was dominated by a huge list (4 million+) of MetricStartEvent within a list of _runningFunctions on the Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics.MetricEventManager class.  Further investigations/memory dumps showed that other applications and other environments also have high memory usage by the same class and list of events.  Our production applications are currently in a cycle of memory usage build up until they error and restart.

We attempted to recreate the issue in our dev environment and found that sending repeated invalid requests that cause SQL errors did increase the memory usage in the same area.  Our code handles the errors and exits the function invocations gracefully.  It's not understood whether it has to be an error that causes the issue or just a heavy load on the functions host runtime.

Debugging the Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics.MetricEventManager locally we were unable to reproduce the issue.

It seems Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics.MetricEventManager has had significant changes in this area over the last few months.

This code below is pretty new and is meant to purge the completed functions from the list of "FunctionStartedEvent".  I suspect that somehow some functions are not being set to "Completed" and therefore never get purged.  This function is called on a timer and also when a function completes.

```
            private void RaiseFunctionMetricEvents()
            {
                if (_runningFunctions.Count == 0)
                {
                    return;
                }

                // We only need to raise events here for functions that aren't completed.
                // Events are raised immediately for completed functions elsewhere.
                var runningFunctions = new List<FunctionStartedEvent>();
                lock (_runningFunctionsSyncLock)
                {
                    // effectively we're pruning all the completed invocations here
                    runningFunctions = _runningFunctions = _runningFunctions.Where(p => !p.Completed).ToList();
                }

                // we calculate concurrency here based on count, since these events are raised
                // on a background thread, so we want the actual count for this interval, not
                // the current count.
                var currentTime = DateTime.UtcNow;
                foreach (var runningFunction in runningFunctions)
                {
                    RaiseFunctionMetricEvent(runningFunction, runningFunctions.Count, currentTime);
                }
            }
```

- Timestamp: n/a
- Function App version: 4.8.0 / 1.7.1
- Function App name: buhler-prod-we-topology-management-v4 / buhler-prod-we-ingestion-pipeline
- Function name(s) (as appropriate): Multiple /  OnIoTHubMessage
- Invocation ID: n/a
- Region: West Europe

#### Repro steps

_Topology Management_

1. Send high load of requests that cause errors persisting to SQL via entity framework for a prolonged period of time
2. Take a memory dump and review the memory usage
3. Repeat steps 1 & 2 and review to see further increase in memory usage

#### Expected behavior

_runningFunctions list should only ever contain the number of actual running functions

#### Actual behavior

_runningFunctions list builds up to a large number of FunctionStartEvent until it causes out of memory exceptions

#### Known workarounds

The host restarts itself when memory usage become too high and the application runs again until the process repeats.  Restart of the application clears the list of FunctionStartedEvent.

#### Related information 

* c#
* https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.WebHost/Diagnostics/MetricsEventManager.cs
* Topology manager uses mainly Http Triggers and Timers.  Ingestion pipeline uses an IoT Hub/event hub binding.
* Memory dumps available on request
* Memory usage history can be seen in application insights workbooks

mathewc:

Resolves https://github.com/Azure/azure-functions-host/issues/6813

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/Diagnostics/MetricsEventManager.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/MetricsEventManager.cs
@@ -7,7 +7,6 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Threading;
-using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Script.Configuration;
 using Microsoft.Azure.WebJobs.Script.Description;
 using Microsoft.Azure.WebJobs.Script.Diagnostics;
@@ -45,7 +44,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
             // Initialize the periodic log flush timer
             _metricsFlushTimer = new Timer(TimerFlush, null, metricsFlushIntervalMS, metricsFlushIntervalMS);
 
-            _functionActivityTracker = new FunctionActivityTracker(_appServiceOptions, _eventGenerator, metricsPublisher, linuxContainerActivityPublisher, _functionActivityFlushIntervalSeconds);
+            _functionActivityTracker = new FunctionActivityTracker(_appServiceOptions, _eventGenerator, metricsPublisher, linuxContainerActivityPublisher, _functionActivityFlushIntervalSeconds, _logger);
         }
 
         /// <summary>
@@ -316,7 +315,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 
                     if (_functionActivityTracker != null)
                     {
-                        _functionActivityTracker.StopEtwTaskAndRaiseFinishedEvent();
+                        _functionActivityTracker.StopTimerAndRaiseFinishedEvent();
                         _functionActivityTracker.Dispose();
                     }
                 }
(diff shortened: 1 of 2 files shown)


src/WebJobs.Script.WebHost/Diagnostics/MetricsEventManager.cs
```diff
@@ -474,20 +473,23 @@ private void RaiseFunctionMetricEvents()
 
                 // We only need to raise events here for functions that aren't completed.
                 // Events are raised immediately for completed functions elsewhere.
-                var runningFunctions = new List<FunctionStartedEvent>();
+                FunctionStartedEvent[] runningFunctionsSnapshot = null;
                 lock (_runningFunctionsSyncLock)
                 {
                     // effectively we're pruning all the completed invocations here
-                    runningFunctions = _runningFunctions = _runningFunctions.Where(p => !p.Completed).ToList();
```
mathewc, line 481: The cause of the issue was here - the runningFunctions local was intended to be a snapshot not a reference to _runningFunctions. Once outside the lock, _runningFunctions can be modified on other threads as functions are executed, which can cause failures below while we're enumerating. This results in "collection was modified" exceptions that caused the activity task to fail silently stop running (background task exception).

src/WebJobs.Script.WebHost/Diagnostics/MetricsEventManager.cs
```diff
@@ -367,43 +370,37 @@ internal FunctionActivityTracker(IOptionsMonitor<AppServiceOptions> appServiceOp
...
-                                }
-                                else
-                                {
-                                    currentSecond = currentSecond + 1;
-                                }
-
-                                await Task.Delay(TimeSpan.FromSeconds(1), _etwTaskCancellationSource.Token);
-                            }
-                        }
-                        catch (TaskCanceledException)
```
mathewc, line 400: Unhandled exceptions that weren't of this type cause this task to fail silently, meaning the metrics pump would then no longer be running. That leads to a memory leak because _runningFunctions and _functionMetricsQueue are no longer being cleared.

src/WebJobs.Script.WebHost/Diagnostics/MetricsEventManager.cs
```diff
@@ -450,9 +449,11 @@ internal void FunctionCompleted(FunctionStartedEvent startedEvent)
...
 
-            internal void StopEtwTaskAndRaiseFinishedEvent()
+            internal void StopTimerAndRaiseFinishedEvent()
             {
-                _etwTaskCancellationSource.Cancel();
+                // stop the timer if it has been started
+                _activityTimer?.Change(Timeout.Infinite, Timeout.Infinite);
+
                 RaiseMetricsPerFunctionEvent();
             }
```
brettsam, line 458: Is there any interesting race condition here where we're in the middle of `TimerFlush` when we call Stop and then we exit before `TimerFlush` is complete?

This is mostly a brainstorming question -- I don't see any issue but wanted to see if you did.

src/WebJobs.Script.WebHost/Diagnostics/MetricsEventManager.cs
```diff
@@ -367,43 +372,37 @@ internal FunctionActivityTracker(IOptionsMonitor<AppServiceOptions> appServiceOp
                     _metricsPublisher = metricsPublisher;
                 }
 
-                StartActivityTimer();
+                _activityFlushCounter = _functionActivityFlushInterval;
+                _activityTimer = new Timer(TimerFlush, null, _activityTimerIntervalMS, _activityTimerIntervalMS);
```
brettsam, line 376: Any idea why this wasn't a Timer to begin with?

brettsam (approved):
Looks good! Just a couple of questions.

src/WebJobs.Script.WebHost/Diagnostics/MetricsEventManager.cs
mathewc: Not sure - this is how Ashish originally coded this

src/WebJobs.Script.WebHost/Diagnostics/MetricsEventManager.cs
mathewc: I don't believe so. The TimerFlush will be on a background thread, so will complete. Anyhow, it'll behave as it did before - I've just added better error handling.

absolutebandit:
Hi, is there any update on the release date for this, we're still experiencing errors on our services and the chiefs are asking?
Thanks :)

mathewc:
Please see the release announcement for version 3.0.14916 here: https://github.com/Azure/app-service-announcements-discussions/issues/207. Fix is rolling out now, and should be complete by EOD Wednesday 10/28.

PR #6820, opened 2020-10-20
[Rpc]Raise worker error event if function load response is delayed

### LINKED ISSUE #6819: [Rpc]Enforce timeout on funciton load response
cc @brettsam @yojagad @mhoeger 

This is a follow up from a CRI. 

- Host starts up OK
- Node worker start up OK
- Host send function load requests to node worker and in parallel accepts any invocation requests while functions are being loaded by the worker
- Logs indicate a number of function started but the invocations are not dispatched to node worker as host did not receive any function load responses
- Function executions start timing out in the host after 5mins

Issue:
- Worker errors are not raised as we do not enforce a timeout on function load response.
- Functions host continues to accept invocation requests and never dispatches invocation to the worker.

Proposal: 
- Enforce timeout on function load request - timeout same as `functionTimeout`  - if a function does not load within `functionTimeout` - invocation is guaranteed to fail.

pragnagopa:

resolves #6819

### Pull request checklist

### Additional information

Additional PR information


yojagad:
>             }

I'm slightly confused, and let me know if I'm missing anything here please. If Function was never loaded by the worker, wouldn't it come to this line and throw InvalidOperationException ? As long as they keep sending invocations before it's loaded, it would keep throwing this exception I'd imagine. 

---
Refers to: src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs:278 in 42fd6ac. [](commit_id = 42fd6ac77920aaaabe4b4ae4539b6a908991ba49, deletion_comment = True)

yojagad:
>                 throw new InvalidOperationException($"Function:{invocationContext.FunctionMetadata.Name} is not loaded by the language worker: {rpcWorkerChannel.Id}");

I'm slightly confused, and let me know if I'm missing anything here please. If Function was never loaded by the worker, wouldn't it come to this line and throw InvalidOperationException ? As long as they keep sending invocations before it's loaded, it would keep throwing this exception I'd imagine.  https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs#L277

pragnagopa:
> I'm slightly confused, and let me know if I'm missing anything here please. If Function was never loaded by the worker, wouldn't it come to this line and throw InvalidOperationException ? As long as they keep sending invocations before it's loaded, it would keep throwing this exception I'd imagine. 

Take a look at following line:
https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs#L273

Note that we posting to a [BuferBlock](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.dataflow.bufferblock-1?view=netcore-3.1) - this is precisely to allow host to accept invocations while worker is still loading functions.

In this case, functions were never loaded by the worker and the executions started timing out but the host was never shutdown.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -510,6 +520,16 @@ internal void HandleWorkerInitError(Exception exc)
             PublishWorkerErrorEvent(exc);
         }
 
+        internal void HandleWorkerFunctionLoadError(Exception exc)
+        {
+            _workerChannelLogger.LogError(exc, "Did not receive function load response within functionTimeout.");
```
mhoeger: Nit: this log message only applies to the timeout case, not the valid worker function load error case

yojagad:
Probably a bigger design discussion then, should we accept invocations at all (even in buffer) not knowing whether the load would be successful ?

pragnagopa:
> Probably a bigger design discussion then, should we accept invocations at all (even in buffer) not knowing whether the load would be successful ?

This is a corner error case scenario. For happy path scenarios - doing in parallel helps - especially with cold start. And yes, we should think about re-design here if we see more more of these.

mhoeger:
@yojagad - I think we would need a fundamental change to the WebJobs architecture to accomplish that though right? i.e. what we were doing in Brasilia?

yojagad:
@mhoeger For the bigger redesign yes, but for shorter term I was thinking along the lines of deferring https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs#247 to when the FunctionLoad is complete. This way it would just throw InvalidOperationException that I pointed to if a customer tries to invoke before Load is complete. But that probably still requires a bit more thought and discussion to understand if it has other impacts, hmm..

yojagad (dismissed):

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
pragnagopa: good point. Fixed error message. Exception should have more details.

mhoeger (dismissed):
lgtm!

yojagad (approved):

PR #6823, opened 2020-10-21
Added timeout for processWaitForExit

### LINKED ISSUE #6754: Handle language worker shutdown failures during worker restart
This issue is tracking the fix for one of the CRIs where language worker gets stuck while restarting on function invocation timeouts.  @yojagad pointed to this line - https://github.com/Azure/azure-functions-host/blob/v2.x/src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs#L192 which might be the issue.

yojagad:

resolves #6754

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs
+++ b/src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs
@@ -21,6 +21,7 @@ namespace Microsoft.Azure.WebJobs.Script.Workers
         private readonly IWorkerConsoleLogSource _consoleLogSource;
         private readonly IScriptEventManager _eventManager;
         private readonly IMetricsLogger _metricsLogger;
+        private readonly int processExitTimeoutInMilliseconds = 60000;
 
         private Process _process;
         private bool _useStdErrorStreamForErrorsOnly;
@@ -200,7 +201,12 @@ namespace Microsoft.Azure.WebJobs.Script.Workers
                     if (!_process.HasExited)
                     {
                         _process.Kill();
-                        _process.WaitForExit();
+                        _process.WaitForExit(processExitTimeoutInMilliseconds);
+
+                        if (!_process.HasExited)
+                        {
+                            throw new TimeoutException($"Worker process has not exited despite waiting for {processExitTimeoutInMilliseconds} ms");
+                        }
                     }
                     _process.Dispose();
                 }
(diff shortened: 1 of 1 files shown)


src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs
```diff
@@ -21,6 +21,7 @@ internal abstract class WorkerProcess : IWorkerProcess, IDisposable
         private readonly IWorkerConsoleLogSource _consoleLogSource;
         private readonly IScriptEventManager _eventManager;
         private readonly IMetricsLogger _metricsLogger;
+        private readonly int processExitTimeoutInMilliseconds = 60000;
```
yojagad: >private readonly int processExitTimeoutInMilliseconds = 60000; [](start = 8, length = 62)

Added an arbitrary timeout here. Would love to get feedback if you had something else in mind

src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs
```diff
@@ -200,7 +201,12 @@ public void Dispose()
                     if (!_process.HasExited)
                     {
                         _process.Kill();
-                        _process.WaitForExit();
+                        _process.WaitForExit(processExitTimeoutInMilliseconds);
```
yojagad: > _process.WaitForExit(processExitTimeoutInMilliseconds); [](start = 23, length = 56)

Unfortunately, this and other _process methods are non-virtual so was not immediately able to add tests for these. Not unless we do some kind of refactor here where we have some kind of an abstraction and have WorkerProcess take dependency on that abstraction of process instead of directly on the Process.

pragnagopa: We have good E2E test coverage for this. No need to add more tests.

src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs
```diff
@@ -200,7 +201,12 @@ public void Dispose()
...
+
+                        if (!_process.HasExited)
+                        {
```
pragnagopa, line 205: do not throw.  `WaitForExit` returns a bool . Check the result. If `false`, log warning and call `process.Kill()`

pragnagopa (changes requested):
We should call `WaitForExit` first then `Kill`.

From documentation: https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.kill?view=netcore-3.1#System_Diagnostics_Process_Kill

```
The Kill method executes asynchronously. After calling the Kill method, call the WaitForExit method to wait for the process to exit, or check the HasExited property to determine if the process has exited.
```

src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs
pragnagopa: 1 minute might be too long.  lets change this to 1sec

pragnagopa (approved):

PR #6840, opened 2020-10-23
Performance improvements

mathewc:
Some minor performance improvements. Crank runs below (5 iterations for each, each iteration 15 seconds) show small improvements from this in RPS/Latency (about 4%).

Before:

| load                   |           |
| ---------------------- | --------- |
| CPU Usage (%)          | 13        |
| Raw CPU Usage (%)      | 26.53     |
| Working Set (MB)       | 32        |
| Build Time (ms)        | 6,217     |
| Start Time (ms)        | 1,161     |
| Published Size (KB)    | 68,222    |
| First Request (ms)     | 261       |
| Requests               | 55,567    |
| Bad responses          | 0         |
| Mean latency (us)      | 69,216    |
| Max latency (us)       | 2,043,404 |
| Requests/sec           | 3,712     |
| Requests/sec (max)     | 13,430    |
| Read throughput (MB/s) | 0.55      |

After:

| load                   |           |
| ---------------------- | --------- |
| CPU Usage (%)          | 12        |
| Raw CPU Usage (%)      | 23.62     |
| Working Set (MB)       | 32        |
| Build Time (ms)        | 5,775     |
| Start Time (ms)        | 1,077     |
| Published Size (KB)    | 68,222    |
| First Request (ms)     | 246       |
| Requests               | 57,893    |
| Bad responses          | 0         |
| Mean latency (us)      | 66,422    |
| Max latency (us)       | 1,997,583 |
| Requests/sec           | 3,873     |
| Requests/sec (max)     | 18,451    |
| Read throughput (MB/s) | 0.58      |

resolves #issue_for_this_pr

### Pull request checklist

### Additional information

Additional PR information


src/WebJobs.Script.WebHost/Diagnostics/FunctionInstanceLogger.cs
```diff
@@ -123,6 +125,23 @@ private void StartFunction(FunctionInstanceLogEntry item)
             }
         }
 
+        internal object LogInvocationMetrics(FunctionMetadata metadata)
+        {
+            // log events for each of the binding types used
+            foreach (var binding in metadata.Bindings)
+            {
+                string eventName = _bindingMetricEventNames.GetOrAdd(binding, (existing) =>
```
mathewc, line 133: Caching these metric event names on this hotpath. No need to reformat these on each invocation for each binding.

src/WebJobs.Script.WebHost/Extensions/IExternalScopeProviderExtensions.cs
```diff
@@ -8,22 +8,24 @@ namespace Microsoft.Extensions.Logging
 {
     internal static class IExternalScopeProviderExtensions
     {
-        public static IDictionary<string, object> GetScopeDictionary(this IExternalScopeProvider scopeProvider)
+        public static IDictionary<string, object> GetScopeDictionaryOrNull(this IExternalScopeProvider scopeProvider)
         {
-            var dictionary = new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase);
+            IDictionary<string, object> result = null;
```
mathewc, line 13: Much of the time there are no scopes, so we don't need to create a dictionary.

src/WebJobs.Script.WebHost/Middleware/FunctionInvocationMiddleware.cs
```diff
@@ -125,6 +125,11 @@ private async Task<IActionResult> GetResultAsync(HttpContext context, IFunctionE
                     var applicationLifetime = context.RequestServices.GetService<IApplicationLifetime>();
                     CancellationToken cancellationToken = applicationLifetime != null ? applicationLifetime.ApplicationStopping : CancellationToken.None;
                     await functionExecution.ExecuteAsync(context.Request, cancellationToken);
+
+                    if (context.Items.TryGetValue(ScriptConstants.AzureFunctionsDuplicateHttpHeadersKey, out object value))
```
mathewc, line 129: This simple check doesn't need to be a first class middleware component, with the overhead associated with that.

src/WebJobs.Script.WebHost/Middleware/JobHostPipelineMiddleware.cs
```diff
@@ -14,16 +14,18 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Middleware
...
+        private readonly IJobHostMiddlewarePipeline _middleware;
 
-        public JobHostPipelineMiddleware(RequestDelegate next)
+        public JobHostPipelineMiddleware(RequestDelegate next, IJobHostMiddlewarePipeline middleware)
         {
             _next = next;
+            _middleware = middleware;
         }
 
-        public Task Invoke(HttpContext httpContext, IJobHostMiddlewarePipeline middleware)
```
mathewc, line 23: Don't need to DI resolve this for each invocation

src/WebJobs.Script.WebHost/Middleware/ScriptHostRequestServiceProviderMiddleware.cs
```diff
@@ -12,19 +12,19 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Middleware
...
         }
 
-        public Task Invoke(HttpContext httpContext, WebJobsScriptHostService manager)
+        public Task Invoke(HttpContext httpContext)
         {
-            if (manager.Services is IServiceScopeFactory scopedServiceProvider)
+            if (_scriptHostService.Services is IServiceScopeFactory scopedServiceProvider)
             {
-                var features = httpContext.Features;
-                var servicesFeature = features.Get<IServiceProvidersFeature>();
```
mathewc, line 26: We were doing this service lookup, but not using the value

src/WebJobs.Script.WebHost/Middleware/ScriptHostRequestServiceProviderMiddleware.cs
```diff
@@ -12,19 +12,19 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Middleware
     public class ScriptHostRequestServiceProviderMiddleware
     {
         private readonly RequestDelegate _next;
+        private readonly WebJobsScriptHostService _scriptHostService;
 
-        public ScriptHostRequestServiceProviderMiddleware(RequestDelegate next)
+        public ScriptHostRequestServiceProviderMiddleware(RequestDelegate next, WebJobsScriptHostService scriptHostService)
         {
             _next = next;
+            _scriptHostService = scriptHostService;
```
brettsam, line 20: Curious -- is there a perf improvement by moving this to the constructor?

src/WebJobs.Script/Utility.cs
```diff
@@ -824,6 +826,13 @@ public static void LogAutorestGeneratedJsonIfExists(string rootScriptPath, ILogg
             // If we dont find the .autorest_generated.json in the function app, we just don't log anything.
         }
 
+        public static void AccumulateDuplicateHeader(HttpContext httpContext, string headerName)
+        {
+            // Add duplicate http header to HttpContext.Items. This will be logged in ResponseContextItemsCheckMiddleware
```
brettsam: Nit -- I think you just removed this middleware in the comment :-)

brettsam (dismissed):

src/WebJobs.Script.WebHost/Middleware/ScriptHostRequestServiceProviderMiddleware.cs
mathewc: Well, doing the DI resolution on Invoke incurs a DI resolution per invocation

src/WebJobs.Script/Utility.cs
mathewc: Will fix, thanks

src/WebJobs.Script.WebHost/Middleware/FunctionInvocationMiddleware.cs
fabiocav: Would be nice to sill maintain this logic outside of this middleware, though. How big of an impact was this having?

src/WebJobs.Script.WebHost/Routing/ScriptRouteHandler.cs
```diff
@@ -30,19 +33,22 @@ public ScriptRouteHandler(ILoggerFactory loggerFactory, IScriptJobHost scriptHos
...
 
-            var descriptor = _scriptHost.Functions.FirstOrDefault(f => string.Equals(f.Name, functionName));
+            var descriptor = _functionMap.GetOrAdd(functionName, (name) =>
+            {
+                return _scriptHost.Functions.FirstOrDefault(p => string.Equals(p.Name, name, StringComparison.OrdinalIgnoreCase));
+            });
             var executionFeature = new FunctionExecutionFeature(_scriptHost, descriptor, _environment, _loggerFactory);
             context.Features.Set<IFunctionExecutionFeature>(executionFeature);
 
-            await Task.CompletedTask;
```
fabiocav, line 45: ?!?!

src/WebJobs.Script.WebHost/Middleware/FunctionInvocationMiddleware.cs
mathewc: Not a big impact really - this was just about reducing the number of pieces of middleware we register in the pipeline. For simple post invocation things like this, why not just include them in the invoke middleware?

Looking at performance profiles, the we dispatch through a large number of middleware components, each involving async state transitions/overhead. Granted, some early experiments I've done with middleware compression didn't yield large improvements, but its an area I want to investigate more.

fabiocav: Makes sense, but I just want us to be careful about separation of concerns and making sure we don't end up with a bloated piece of middleware that does more than it should, is difficult to maintain and ends up being touched more than it should as that would lead to regressions. We should be able to optimize existing pieces of middleware to ensure we're applying good patterns to avoid going async when not needed, but in many cases, even if there's a small gain from combining logic, if that leads to code that is more prone to issues, I think we should keep them separate.

Having them as separate pieces also enables us to more easily test the middleware surface, which also gives us another layer of safety.

PR #6842, opened 2020-10-27
[pack] Update grpc to use ASP.NET managed version

### LINKED ISSUE #5449: Migrate gRPC to the new ASP.NET Core v3 gRPC

ankitkumarr:

resolves https://github.com/Azure/azure-functions-host/issues/5449

### Pull request checklist

### Additional information

Part 2 of https://github.com/Azure/azure-functions-host/issues/5237.
Reviving this work https://github.com/Azure/azure-functions-host/pull/5428.

**I need to run extensive tests for this, including perf runs to make sure things are ok. In the meanwhile, figured it'd be good to get code feedback.**


src/WebJobs.Script.Grpc/WebJobs.Script.Grpc.csproj
```diff
@@ -19,14 +20,13 @@
...
+    <PackageReference Include="Grpc.Tools" Version="2.32.0">
       <PrivateAssets>All</PrivateAssets>
     </PackageReference>
-    <PackageReference Include="Microsoft.Extensions.Logging" Version="2.2.0" />
-    <PackageReference Include="Microsoft.Extensions.Configuration" Version="2.2.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="3.1.9" />
+    <PackageReference Include="Microsoft.Extensions.Configuration" Version="3.1.9" />
     <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
     <PackageReference Include="StyleCop.Analyzers" Version="1.1.0-beta004">
       <PrivateAssets>all</PrivateAssets>
```
ankitkumarr, line 31: I will update our protobuf generation logic to use https://docs.microsoft.com/en-us/aspnet/core/grpc/dotnet-grpc?view=aspnetcore-3.1#add-references available in the managed grpc library, instead of trunning `generate_protos` scripts.

```csharp
<Protobuf Include="Protos\greet.proto" GrpcServices="Server" />
```

ankitkumarr:
The two tests that are failing are --
1. ColdStart_JitFailuresTest: I may need to capture a new jittrace, and I will do that once my changes are finalized.
2. DependencyTests.Verify_DepsJsonChanges: It says dependencies have changed (yes they have), and I need to talk to @fabiocav or @brettsam for next steps.

Other than that, everything else so far seems good.

src/WebJobs.Script.Grpc/Server/GrpcServer.cs
```diff
@@ -2,70 +2,61 @@
...
         {
-            ChannelOption maxReceiveMessageLength = new ChannelOption(ChannelOptions.MaxReceiveMessageLength, MaxMessageLengthBytes);
-            ChannelOption maxSendMessageLength = new ChannelOption(ChannelOptions.MaxSendMessageLength, MaxMessageLengthBytes);
-            ChannelOption[] grpcChannelOptions = { maxReceiveMessageLength, maxSendMessageLength };
-            _server = new Server(grpcChannelOptions)
-            {
-                Services = { FunctionRpc.BindService(serviceImpl) },
-                Ports = { new ServerPort("127.0.0.1", ServerPort.PickUnused, ServerCredentials.Insecure) }
-            };
+            _port = WorkerUtilities.GetUnusedTcpPort();
```
pragnagopa, line 22: This helper was added for `HttpWorker` - so there is no in built way to get a random unused port?

src/WebJobs.Script.Grpc/Server/GrpcServer.cs
```diff
@@ -2,70 +2,61 @@
...
 
+        public ValueTask DisposeAsync()
+        {
+            return Dispose(true);
+        }
+
         public void Dispose()
         {
-            Dispose(true);
+            Dispose(true).GetAwaiter().GetResult();
```
pragnagopa: As this is an application level service dispose would be called as part for application shutdown. Can we avoid this blocking call?

src/WebJobs.Script.Grpc/Startup.cs
```diff
@@ -0,0 +1,39 @@
...
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+namespace Microsoft.Azure.WebJobs.Script.Grpc
+{
+    internal class Startup
+    {
+        private const int MaxMessageLengthBytes = 128 * 1024 * 1024;
```
pragnagopa: This needs to ` public const int MaxMessageLengthBytes = int.MaxValue;`  chaning this size will cause apps that send large to fail. Can you please add unit test so that we dont accidentally update the size in future?

src/WebJobs.Script.Grpc/WebJobs.Script.Grpc.csproj
pragnagopa: Will you be sending a separate PR for removing generate_protos script?

pragnagopa (changes requested):
Thanks! Added few comments.

src/WebJobs.Script.Grpc/Startup.cs
ankitkumarr: Oh sorry, I must have used an outdated version to port. I will update this.

ankitkumarr: Updated and added a test

src/WebJobs.Script.Grpc/Server/GrpcServer.cs
ankitkumarr: I changed it to Dispose in the background, similar to what it was in the previous version.

src/WebJobs.Script.Grpc/Server/GrpcServer.cs
ankitkumarr: I didn't find a reliable way for us to get a port and record that without significant changes. Kestrel has a way to declare that it can pick up any unused port, but for us to be keep track of that becomes more involved, and I suspect there could be races if we need the port number before the server started.

Is there a problem reusing that helper?

src/WebJobs.Script.Grpc/WebJobs.Script.Grpc.csproj
ankitkumarr: I updated this PR with the change. The only reason I hadn't because I haven't figured out a proper way to specify the output directory for specific files after conversion. For example, the way we have `Messages/DotNet/FunctionRpc.cs` and `Messages/DotNet/FunctionRpcGrpc.cs` I couldn't get that with the new target way.

The Target has an option for `OutputDir` but that causes issues with other ".rpc" helper files. It doesn't have something like `OutputFiles` similar to `Outputs` that we use today.

With the current change, the converted files don't show up in `Messages/DotNet`. Is there a reason we need that? Or is it just for convenience? If we do, then I can revert my target change and take that separately to not block this work.

src/WebJobs.Script.WebHost/WebJobs.Script.WebHost.csproj
```diff
@@ -73,6 +73,8 @@
     <PackageReference Include="Microsoft.Azure.WebJobs.Logging" Version="4.0.1" />
     <PackageReference Include="Microsoft.Azure.WebSites.DataProtection" Version="2.1.91-alpha" />
     <PackageReference Include="Microsoft.VisualStudio.Web.CodeGeneration.Design" Version="2.2.3" />
+    <!-- The Grpc.Core.dll and Grpc.Core.Api.dll were removed and required to maintain backward compatibility -->
+    <PackageReference Include="Grpc.Core" Version="2.27.0" />
```
ankitkumarr, line 77: @brettsam, fyi

src/WebJobs.Script.Grpc/Server/GrpcServer.cs
pragnagopa: Just making sure we were utilizing sth built in. This is OK!

src/WebJobs.Script.Grpc/WebJobs.Script.Grpc.csproj
pragnagopa: Folder structure is OK. As long as namespace remains the same this change is OK. Thanks.

pragnagopa:
Can you please see if logic here: https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.Grpc/Server/FunctionRpcService.cs#L34-L111 can be simplified?  Code samples here: https://www.stevejgordon.co.uk/server-streaming-with-grpc-in-asp-dotnet-core show a cleaner approach

pragnagopa (approved):

brettsam:
We talked about this in-person -- but how risky do we think this is? Is it worth hiding behind a feature flag at first before rolling it out everywhere? @fabiocav -- your thoughts?

pragnagopa:
@ankitkumarr is testing the changes in Azure via private site extension.  Test did run into sandbox errors. We can follow up offline about this.

As long as E2E tests that are part of the PR builds, a simple perf test in Azure and OGFs are OK, this is a safe change. It might be overkill to add a feature flag for this.

fabiocav:
Closing stale PR. If this work is still relevant, can you please rebase, reopen and engage reviewers? Thank you!

PR #6848, opened 2020-10-27
Remove hardcoded placeholder mode for linux consumption

divyagandhisethi:

resolves #issue_for_this_pr

### Pull request checklist

    * [X] Otherwise: Backport tracked by issue/PR #6847 

### Additional information

Enabling placeholders app setting should not be hardcoded inside the functions host. The control plane will add the app setting from now on to enable placeholders if this container is for a placeholder pool. If this is a pinned linux consumption container, then placeholders will not be enabled from the control plane side.

### DIFF
--- a/src/WebJobs.Script.WebHost/Program.cs
+++ b/src/WebJobs.Script.WebHost/Program.cs
@@ -87,8 +87,6 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
         {
             if (SystemEnvironment.Instance.IsLinuxConsumption())
             {
-                // Linux containers always start out in placeholder mode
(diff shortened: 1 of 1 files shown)


balag0:
Since the end result is the same whether the app setting is set in the image / from control plane, why do we prefer one over the other?

divyagandhisethi:
> 
> 
> Since the end result is the same whether the app setting is set in the image / from control plane, why do we prefer one over the other?

Because currently placeholder mode is enabled for pinned containers as well, which leads to other issues like longer specialization and cold start time for pinned apps.

balag0 (dismissed):

Hazhzeng (approved):

PR #6849, opened 2020-10-28
Added methods to skip cleaning of output assemblies to ExtensionsMetadataGenerator.

### LINKED ISSUE #5894: [ExtensionsMetadataGenerator] Improve configuration for cleaning build output
Today we have an "all or nothing" approach to cleaning your build output and disabling this involves a (currently) undocumented MSBuild setting: `_FunctionsSkipCleanOutput`. We're seeing a handful of customers run into this -- they move forward to a newer version of an assembly that we're cleaning away.

This often fails with `Could not load file or assembly...` exceptions during runtime.

I think we need to:
1. Make this setting not as "internal" (remove the underscore) and document this... I'd assume here: https://docs.microsoft.com/en-us/azure/azure-functions/functions-develop-vs
2. How fancy we can get with the cleanup and still have it be maintainable? Can we only delete if the version is to an exact match to what we have in https://github.com/Azure/azure-functions-host/blob/dev/tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator/runtimeassemblies.txt?
3. If we can't do this smart enough with number 2, allow you use MSBuild items to selectively exclude or include assemblies. That way you can still benefit from the cleaning but get things running.

Some related issues that have come up (I'm resolving all of those to this one):
- https://github.com/Azure/azure-functions-vs-build-sdk/issues/397
- https://github.com/Azure/azure-functions-vs-build-sdk/issues/412
- https://github.com/Azure/Azure-Functions/issues/1518#issuecomment-615432601
- https://github.com/Azure/Azure-Functions/issues/1525

gzuber:

Resolves #5894 

### Pull request checklist

    * [x] Otherwise: Documentation issue linked to PR (https://github.com/MicrosoftDocs/azure-docs/pull/65127)

### Additional information

This adds a new property that users can specify in their csproj to define how ExtensionsMetadataGenerator handles output clean up.

They can add the following:
```xml
  <ItemGroup>
    <FunctionsPreservedDependencies Include="Microsoft.AspNetCore.Http.dll" />
    <FunctionsPreservedDependencies Include="Microsoft.AspNetCore.Http.Extensions.dll" />
    <FunctionsPreservedDependencies Include="Microsoft.AspNetCore.Http.Features.dll" />
  </ItemGroup>
```
For every assembly/file in the build output they don't want deleted during the cleaning step.

Notes about the PR:
- I've made a PR for documenting this feature, but I'm unsure if it's really in the right place and would appreciate a review of the content/wording: https://github.com/MicrosoftDocs/azure-docs/pull/65127
- I'm open to suggestions on the name of these settings. I thought `FunctionsPreserveOutputAssembly` might be better, but went with what I did to match `FunctionsSkipCleanOutput`. I kept `FunctionsSkipCleanOutput` because I know some customers use and are familiar with `_FunctionsSkipCleanOutput` but that might not be a strong enough reason to keep it if we could come up with a better one.
- I looked into adding tests. My idea was to create a new test function app project (TestProject_SkipClean for instance) with some FunctionsSkipCleanOutputAssembly defined. Then traverse the file system to make sure those assemblies are still in the build output. However, I'm not sure how I can specify that the project should use the ExtensionMetadataGenerator from the project and not the one from `Microsoft.NET.Sdk.Functions`. I could use some help with this.
- I'm not sure if I should add to the Host's release notes. Isn't this on its own release cadence that's not tied to the host?

### DIFF
--- a/tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator/BuildTasks/RemoveRuntimeDependencies.cs
+++ b/tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator/BuildTasks/RemoveRuntimeDependencies.cs
@@ -21,8 +21,17 @@ namespace Microsoft.Azure.WebJobs.Script.ExtensionsMetadataGenerator.BuildTasks
         [Required]
         public string OutputPath { get; set; }
 
+        [Required]
+        public ITaskItem[] IgnoreFiles { get; set; }
+
         public override bool Execute()
         {
+            HashSet<string> ignoreFilesSet = new HashSet<string>();
+            foreach (ITaskItem item in IgnoreFiles)
+            {
+                ignoreFilesSet.Add(item.ItemSpec);
+            }
+
             Assembly assembly = typeof(RemoveRuntimeDependencies).Assembly;
             using (Stream resource = assembly.GetManifestResourceStream(assembly.GetName().Name + ".runtimeassemblies.txt"))
             using (var reader = new StreamReader(resource))
@@ -32,7 +41,7 @@ namespace Microsoft.Azure.WebJobs.Script.ExtensionsMetadataGenerator.BuildTasks
                 {
                     string fileName = Path.Combine(OutputPath, assemblyName);
 
-                    if (File.Exists(fileName))
+                    if (File.Exists(fileName) && !ignoreFilesSet.Contains(assemblyName))
                     {
                         File.Delete(fileName);
                     }
--- a/tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator/Targets/Microsoft.Azure.WebJobs.Script.ExtensionsMetadataGenerator.targets
+++ b/tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator/Targets/Microsoft.Azure.WebJobs.Script.ExtensionsMetadataGenerator.targets
@@ -21,11 +21,11 @@
              AssemblyFile="$(_FunctionsExtensionsTaskAssemblyFullPath)"/>
   
   <Target Name="_FunctionsBuildCleanOutput" AfterTargets="_GenerateFunctionsExtensionsMetadataPostBuild" Condition="$(_FunctionsSkipCleanOutput) != 'true'" >
-    <RemoveRuntimeDependencies OutputPath="$(TargetDir)bin"/>
+    <RemoveRuntimeDependencies OutputPath="$(TargetDir)bin" IgnoreFiles="@(FunctionsPreservedDependencies)"/>
   </Target>
 
   <Target Name="_FunctionsPublishCleanOutput" AfterTargets="_GenerateFunctionsExtensionsMetadataPostPublish" Condition="$(_FunctionsSkipCleanOutput) != 'true'" >
-    <RemoveRuntimeDependencies OutputPath="$(PublishDir)bin"/>
+    <RemoveRuntimeDependencies OutputPath="$(PublishDir)bin" IgnoreFiles="@(FunctionsPreservedDependencies)"/>
   </Target>
   
   <UsingTask TaskName="GenerateFunctionsExtensionsMetadata"


brettsam:
Adding Fabio as well to take a look. To answer some questions: 
1. I think that's a good place for them. @ggailey777 may have some feedback if not.
2. I agree the names seem pretty verbose. @fabiocav may have some recommendations. I'm not sure if we have any other publicly settable properties like this so there may not be precedent for naming.
3. You may want to add these to [my set of end-to-end tests](https://github.com/brettsam/FunctionsSdkE2ETests) -- we run these before every release against staging packages. For example, [here](https://github.com/brettsam/FunctionsSdkE2ETests/tree/master/DirectRef/DirectRefEMG) is a project that directly references ExtensionsMetadataGenerator (EMG for short). There's a [common file](https://github.com/brettsam/FunctionsSdkE2ETests/blob/master/SdkVersion.props) where you can update the EMG version to whichever you want to test and run through tests that way. Feel free to ping me with questions and I can walk you through it.
4. No, we'll add this to the release notes when we release EMG. [Example](https://github.com/Azure/azure-functions-host/releases/tag/emg-v1.2.0).

tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator/BuildTasks/RemoveRuntimeDependencies.cs
```diff
@@ -21,8 +21,17 @@ public class RemoveRuntimeDependencies : Task
         [Required]
         public string OutputPath { get; set; }
 
+        [Required]
+        public ITaskItem[] FilesToSkip { get; set; }
```
fabiocav: How about `IgnoreFiles` ?

tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator/Targets/Microsoft.Azure.WebJobs.Script.ExtensionsMetadataGenerator.targets
```diff
@@ -21,11 +22,11 @@
              AssemblyFile="$(_FunctionsExtensionsTaskAssemblyFullPath)"/>
   
   <Target Name="_FunctionsBuildCleanOutput" AfterTargets="_GenerateFunctionsExtensionsMetadataPostBuild" Condition="$(_FunctionsSkipCleanOutput) != 'true'" >
-    <RemoveRuntimeDependencies OutputPath="$(TargetDir)bin"/>
+    <RemoveRuntimeDependencies OutputPath="$(TargetDir)bin" FilesToSkip="@(FunctionsSkipCleanOutputAssembly)"/>
```
fabiocav: `FunctionsSkipCleanOutputAssembly`-> `FunctionsPreservedDependencies`

tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator/Targets/Microsoft.Azure.WebJobs.Script.ExtensionsMetadataGenerator.targets
```diff
@@ -9,6 +9,7 @@
     <_FunctionsExtensionsDir Condition="$(_IsFunctionsSdkBuild) == 'true'">$(_FunctionsExtensionsDir)bin</_FunctionsExtensionsDir>
     <_ExtensionsMetadataGeneratorTargetsImported>true</_ExtensionsMetadataGeneratorTargetsImported>
     <IsPackable>false</IsPackable>
+    <_FunctionsSkipCleanOutput Condition="$(FunctionsSkipCleanOutput) == 'true' Or $(_FunctionsSkipCleanOutput) == 'true'">true</_FunctionsSkipCleanOutput>
```
fabiocav: After chatting with @gzuber , let's not expose this as a public property/option right now.

brettsam (approved):

fabiocav:
@gzuber let's coordinate a NuGet release for this. We should be able to do that at any point

sonic1981:
did this get released?

brettsam:
@sonic1981 -- yes, see details here: https://github.com/Azure/azure-functions-host/releases/tag/emg-v1.2.1. At the moment, you'll need to explicitly add a package reference to this version of the package (it's usually brought in as a dependency).

sonic1981:
Great thanks. Is there a list of what the "assemblies that are shared with the functions runtime" are anywhere?

sonic1981:
I still don't really understand how I'm supposed to use this functionality. I'd argue the [docs are not really very clear](https://docs.microsoft.com/en-us/azure/azure-functions/functions-develop-vs#configure-your-build-output-settings). 

Given the web of dependencies that get installed by Nuget and the lack of visibility in Azure of what dll's are available by default and what version they are,, etc it seems a big ask to somehow figure out which ones I want to clean and which I don't. I'm continuing to use the blunt hammer of `FunctionsSkipCleanOutput` because it just works.

gzuber:
Hi @sonic1981 sorry for the delayed response. That's a good point, the documentation isn't specific about the dlls functions is bringing in which can be frustrating. If it helps, we keep a list [here](https://github.com/Azure/azure-functions-host/blob/dev/tools/ExtensionsMetadataGenerator/test/ExtensionsMetadataGeneratorTests/ExistingRuntimeAssemblies.txt) that might help you (or at least give you a starting point to) determine which dlls you need to preserve for your project.

I believe that in general, when a customer needs to use this feature it's because they have a very specific requirement e.g. needing to lock themselves to a specific version of a dll. Or they've ran into an exception because the version of a dll they need can't be found, so they're aware of the specific dll that's causing the issue and this feature allows them to fix it without losing much efficiency. `FunctionsSkipCleanOutput` is still around for situations where that doesn't apply.

PR #6867, opened 2020-11-04
Fixing http throttle enablement (#6548)

### LINKED ISSUE #6548: Setting maxConcurrentRequests to 1 does not work
@jaylorch and I have tried to run an application that is CPU-bound on Azure Functions, and found that setting `maxConcurrentRequests` to 1 does not limit the number of concurrently running requests to 1. 

Tagging @pragnagopa to get the right folks for this issue.

- Timestamp: 2020-08-21T21:13:05.3540926Z
- Function App version: 3
- Invocation ID: dfc21e2b-f960-4a47-bb47-a41eca3dea93
- Region: East US

#### Repro steps

Use a hosts.json like
```
{
    "version": "2.0",
    "extensions": {
        "http": {
            "routePrefix": "api",
            "maxOutstandingRequests": 200,
            "maxConcurrentRequests": 1,
            "dynamicThrottlesEnabled": true
        }
    }
}
```
and use code like the example below. Then, make a big batch of concurrent requests (e.g. make 20 reqs in parallel).

#### Expected Behavior

Since the hosts.json has `maxConcurrentRequests` set to 1, Azure Functions should only process one request on an instance at a time.

#### Actual behavior

Azure Functions attempts to concurrently process requests on one instance, which causes our function to run out of memory and results in slow response times (because many invocations of our CPU bound program are being run concurrently). In our function, each request should use a few hundred MB of memory and take not more than 10-20 seconds to complete.

#### Known workarounds

We could try to detect when our function is being run concurrently, and force our code to wait for a currently-executing function to finish, but this would mean we get billed for the time spent waiting and might prevent Azure Functions' scaling logic from recognizing that we require more instances to process our requests.

#### Example code

<details>
<summary>Example code</summary>

```csharp

  public static class ComputeBoundFunction
  {
    static bool AlreadyRunning = false;

    [FunctionName("ComputeBoundFunction")]
    public static IActionResult Run(
        [HttpTrigger(AuthorizationLevel.Anonymous, "post", Route = null)] HttpRequest req,
        ILogger log, ExecutionContext executionContext)
    {
      if (AlreadyRunning) {
        var result = new ObjectResult("Already running on this instance!");
        result.StatusCode = StatusCodes.Status418ImATeapot; // HTTP 418 means concurrent request
        return result;
      }
      AlreadyRunning = true;

      // Run compute-bound process here;

      // Can simulate running time to see concurrent requests by un-commenting
      // Thread.Sleep(5000);

      AlreadyRunning = false;
      return new OkObjectResult("Successfully finished");
    }
  }

<details>

### PR DESCRIPTION
**Pull request by mathewc**

Resolves https://github.com/Azure/azure-functions-host/issues/6548

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/Middleware/HttpThrottleMiddleware.cs
+++ b/src/WebJobs.Script.WebHost/Middleware/HttpThrottleMiddleware.cs
@@ -8,6 +8,7 @@ using Microsoft.AspNetCore.Http;
 using Microsoft.Azure.WebJobs.Extensions.Http;
 using Microsoft.Azure.WebJobs.Script.Diagnostics;
 using Microsoft.Azure.WebJobs.Script.Scale;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 
@@ -28,9 +29,15 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Middleware
             _logger = loggerFactory?.CreateLogger("Host." + ScriptConstants.TraceSourceHttpThrottleMiddleware);
         }
 
-        public static bool ShouldEnable(HttpOptions options)
+        public static bool ShouldEnable(IServiceProvider serviceProvider)
         {
-            return HttpRequestQueue.IsEnabled(options) || options.DynamicThrottlesEnabled;
+            var scriptHostManager = serviceProvider?.GetService<IScriptHostManager>();
+            if (scriptHostManager != null && Utility.TryGetHostService(scriptHostManager, out IOptions<HttpOptions> options))
+            {
+                return HttpRequestQueue.IsEnabled(options.Value) || options.Value.DynamicThrottlesEnabled;
+            }
+
+            return false;
         }
 
         public async Task Invoke(HttpContext httpContext, IOptions<HttpOptions> httpOptions, HttpRequestQueue requestQueue, HostPerformanceManager performanceManager, IMetricsLogger metricsLogger)
--- a/src/WebJobs.Script.WebHost/WebJobsApplicationBuilderExtension.cs
+++ b/src/WebJobs.Script.WebHost/WebJobsApplicationBuilderExtension.cs
@@ -25,7 +25,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
             IEnvironment environment = builder.ApplicationServices.GetService<IEnvironment>() ?? SystemEnvironment.Instance;
             IOptionsMonitor<StandbyOptions> standbyOptions = builder.ApplicationServices.GetService<IOptionsMonitor<StandbyOptions>>();
             IOptionsMonitor<HttpBodyControlOptions> httpBodyControlOptions = builder.ApplicationServices.GetService<IOptionsMonitor<HttpBodyControlOptions>>();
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: src/WebJobs.Script.WebHost/WebJobsApplicationBuilderExtension.cs [hunk 1]
```diff
@@ -71,10 +70,7 @@ public static IApplicationBuilder UseWebJobsScriptHost(this IApplicationBuilder
             {
                 config.UseMiddleware<HomepageMiddleware>();
             });
-            builder.UseWhen(context => HttpThrottleMiddleware.ShouldEnable(httpOptions.CurrentValue) && !context.Request.IsAdminRequest(), config =>
```
mathewc, line 74: Unlike some of the other options types being used to conditionally register other middleware in this method, HttpOptions is ScriptHost scoped, so it can't be used here. I.e. on startup when this code runs, it has the default value, and will never receive the value read from host.json.

This was regressed recently in https://github.com/Azure/azure-functions-host/commit/826c0b520c1e81d2f3106d08711c08f1bce4aa95. That change yielded about a 7% RPS improvement in a Hello World scenario, and the changes here retain that.

test/WebJobs.Script.Tests/Middleware/HttpThrottleMiddlewareTests.cs
```diff
@@ -108,10 +131,11 @@ public async Task Invoke_NoThrottle_DispatchesDirectly()
             Assert.Equal(HttpStatusCode.Accepted, (HttpStatusCode)httpContext.Response.StatusCode);
         }
 
-        [Fact]
-        public async Task Invoke_MaxParallelism_RequestsAreThrottled()
+        [Theory]
+        [InlineData(3)]
+        [InlineData(1)]
```
mathewc, line 138: Verfied serial invocation with this fix. Updating this test to verify ActionBlock behavior with MaxConcurrentRequests set to 1

src/WebJobs.Script.WebHost/Middleware/HttpThrottleMiddleware.cs
```diff
@@ -28,9 +29,11 @@ public HttpThrottleMiddleware(RequestDelegate next, ILoggerFactory loggerFactory
             _logger = loggerFactory?.CreateLogger("Host." + ScriptConstants.TraceSourceHttpThrottleMiddleware);
         }
 
-        public static bool ShouldEnable(HttpOptions options)
+        public static bool ShouldEnable(IScriptHostManager scriptHostManager)
         {
-            return HttpRequestQueue.IsEnabled(options) || options.DynamicThrottlesEnabled;
+            var options = (scriptHostManager as IServiceProvider)?.GetService<IOptions<HttpOptions>>();
```
brettsam: I believe there's a chance you'll hit a race and get an `ObjectDisposedException` here if the host is stopping or restarting. I know elsewhere we try/catch for this explicitly. I think we'll want to do this here too.

brettsam (commented):
One comment.... but looks good.

src/WebJobs.Script.WebHost/Middleware/HttpThrottleMiddleware.cs
mathewc: Added an extension method and used it in a few places where we're already using this pattern

mathewc: Seeing some test failures - investigating

brettsam (dismissed):

PR #6870, opened 2020-11-05
Suppressing execution context flow on create/update/delete

alrod:


src/WebJobs.Script.WebHost/Controllers/FunctionsController.cs
```diff
@@ -74,11 +74,19 @@ public async Task<IActionResult> CreateOrUpdate(string name, [FromBody] Function
 
             bool success, configChanged;
             FunctionMetadataResponse functionMetadataResponse;
-            using (fileMonitoringService.SuspendRestart())
+            using (fileMonitoringService?.SuspendRestart())
```
fabiocav: Have you seen instances where this was null?

src/WebJobs.Script.WebHost/Controllers/FunctionsController.cs
```diff
@@ -74,11 +74,19 @@ public async Task<IActionResult> CreateOrUpdate(string name, [FromBody] Function
...
             {
                 (success, configChanged, functionMetadataResponse) = await _functionsManager.CreateOrUpdate(name, functionMetadata, Request);
             }
-            await fileMonitoringService.ScheduleRestartAsync(false);
+            Task restartTask = null;
```
fabiocav: Can you please introduce a method to encapsulate this pattern so we don't have to repeat it?

src/WebJobs.Script.WebHost/Controllers/FunctionsController.cs
alrod: No, actually. Will remove "?".

fabiocav (approved):

PR #6883, opened 2020-11-11
Sequential JobHost restart if config set

mhoeger:
Here's one way we can achieve non-overlapping workers. Help with name of config would be appreciated! Also debating whether we should actually just do this off of `_environment.IsCoreTools()` since I don't think this is the cleanest solution and I don't feel great about exposing what I feel is a bit of a hacky capability to users. Maybe better to tie to core tools and make the assumption that core tools == local development?

Another option is to move all language worker channel management to the WebHost level and get rid of the JobHost worker channel manager, but in that case we would also need to re-do how we notify a WebHost service to restart workers on a file change (or other JobHost restart). However, this solution would be a lot more refactoring and more impactful. This should be a targeted fix.

Will need to add `AzureFunctionsJobHost__SequentialRestart = true` to core tools

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
+++ b/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
@@ -12,9 +12,11 @@ using Microsoft.ApplicationInsights.AspNetCore.Extensions;
 using Microsoft.ApplicationInsights.Extensibility;
 using Microsoft.ApplicationInsights.Extensibility.Implementation.ApplicationId;
 using Microsoft.Azure.WebJobs.Logging;
+using Microsoft.Azure.WebJobs.Script.Configuration;
 using Microsoft.Azure.WebJobs.Script.Diagnostics;
 using Microsoft.Azure.WebJobs.Script.Scale;
 using Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics.Extensions;
+using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
@@ -34,6 +36,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
         private readonly IMetricsLogger _metricsLogger;
         private readonly HostPerformanceManager _performanceManager;
         private readonly IOptions<HostHealthMonitorOptions> _healthMonitorOptions;
+        private readonly IConfiguration _config;
         private readonly SlidingWindow<bool> _healthCheckWindow;
         private readonly Timer _hostHealthCheckTimer;
         private readonly SemaphoreSlim _hostStartSemaphore = new SemaphoreSlim(1, 1);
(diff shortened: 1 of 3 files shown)


pragnagopa:
> Maybe better to tie to core tools and make the assumption that core tools == local development

I am in favor of relying on CoreTools !

> Another option is to move all language worker channel management to the WebHost level and get rid of the JobHost worker channel manager, but in that case we would also need to re-do how we notify a WebHost service to restart workers on a file change (or other JobHost restart).

I have concerns with this option. We do want tie the lifetime / scope of language worker process to JobHost scope. Moving to webhost scope makes sense if we move to hot reloads.

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -461,8 +466,23 @@ public async Task RestartHostAsync(CancellationToken cancellationToken)
...
                     using (var activeOperation = ScriptHostStartupOperation.Create(cancellationToken, _logger))
                     {
-                        Task startTask = UnsynchronizedStartHostAsync(activeOperation);
-                        Task stopTask = Orphan(previousHost, cancellationToken);
+                        bool.TryParse(_config.GetSection(ConfigurationSectionNames.SequentialJobHostRestart).Value, out bool enforceSequentialOrder);
+                        Task startTask, stopTask;
+
+                        // If we are running in development mode with core tools, do not overlap the restarts.
+                        // Overlapping restarts are problematic when language worker processes are listening
+                        // to the same debug port
```
pragnagopa, line 473: +1 for  doing this only when running in core tools.  I believe we can add E2E tests in Core Tools for this change?

fabiocav: I thought we were specifically tying to decouple this from core tools. What is the down side of the current implementation? I can see this being valuable in other scenarios as well, and it doesn't prevent us from automatically setting this from core tools and having tests there.

fabiocav: Added a comment, the the port conflict is just one example of a resource that cannot be shared between processes. Others do apply. 

I would make the configuration name more intuitive (e.g. `DisableOverlappedRecycles`) or something along those lines.

pragnagopa: So far the issue only came up with in the context of local debugging. I cannot think of other scenarios where this flag would be useful outside of core tools. Do you have other examples where this might be useful?

fabiocav: This could happen in any scenario where a resource contention might exist. From ports (like what we're seeing with the worker) to file locking issues (one example of that are some of the file locking issues that lead us to disable overlapped recycles in Antares). It's also good to keep in mind that, although we're only seeing this impact local debugging operations with _our_ scenarios, it would be entirely feasible that similar issues could impact customer workloads as well. 

This approach keeps the implementation generic, targeting a class of issues.

pragnagopa: Ok.  I see this flag is useful outside the context of language worker processes! Thanks.

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -461,8 +466,23 @@ public async Task RestartHostAsync(CancellationToken cancellationToken)
 
                     using (var activeOperation = ScriptHostStartupOperation.Create(cancellationToken, _logger))
                     {
-                        Task startTask = UnsynchronizedStartHostAsync(activeOperation);
-                        Task stopTask = Orphan(previousHost, cancellationToken);
+                        bool.TryParse(_config.GetSection(ConfigurationSectionNames.SequentialJobHostRestart).Value, out bool enforceSequentialOrder);
```
pragnagopa: Move the check to a internal method and add unit test

pragnagopa (commented):
Just one comment on adding a test. Thanks. Simpler changes than I had imagined!

pragnagopa (dismissed):
Please add reference to this PR in the Core-Tools that sets  `SequentialJobHostRestart`

pragnagopa (approved):

PR #6896, opened 2020-11-12
[RpcSharedMemory]Update proto file to v1.4.0-protofile

### LINKED ISSUE #6791: Shared memory data transfer between Functions host and out-of-proc workers for binary data

### LINKED ISSUE #6897: Protobuf changes to support shared memory data transfer
Protobuf data types to support shared memory data transfer between host and out-of-process workers where applicable.

n/a

n/a

#### Additional context
Related to #6791

gohar94:

Adds protobuf data types and messages needed to enable shared memory data transfer between host and out-of-process workers.

resolves #6897
related to https://github.com/Azure/azure-functions-host/issues/6791

### Pull request checklist

NOTE: Existing tests are enough to verify the changes in this PR.

### DIFF
--- a/src/WebJobs.Script.Grpc/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
+++ b/src/WebJobs.Script.Grpc/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
@@ -71,6 +71,10 @@ message StreamingMessage {
     FunctionEnvironmentReloadRequest function_environment_reload_request = 25;
 
     FunctionEnvironmentReloadResponse function_environment_reload_response = 26;
+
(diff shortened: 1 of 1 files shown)


goiri (dismissed):

pragnagopa:
Please open a separate issue for this.
Just created a release:  https://github.com/Azure/azure-functions-language-worker-protobuf/releases/tag/v1.4.0-protofile . See instructions here: https://github.com/Azure/azure-functions-language-worker-protobuf#pulling-updates

pragnagopa (approved):

goiri (approved):

PR #6900, opened 2020-11-14
Adding sync trigger endpoint for Kubernetes managed K8SE environment.

SatishRanjan:
This PR enables the functionality for the function host to call the build service endpoint running in the managed kubernetes K8SE environment to sync the triggers for KEDA scaling to work as expected. 

### Pull request checklist


src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -558,6 +563,10 @@ internal HttpRequestMessage BuildSetTriggersRequest()
 
             var hostname = _hostNameProvider.Value;
             var url = $"{protocol}://{hostname}/operations/settriggers";
+            if (_environment.IsKubernetesManagedHosting())
```
mathewc, line 558: nit: can you make this an if/else, so we don't form the url above only to overwrite it?

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -558,6 +563,10 @@ internal HttpRequestMessage BuildSetTriggersRequest()
...
+            {
+                url = $"http://{ManagedKubernetesBuildServiceName}.{ManagedKubernetesBuildServiceNamespace}.svc.cluster.local:{ManagedKubernetesBuildServicePort}/operations/settriggers";
```
mathewc, line 560: What is the auth model for this request? In SetTriggers below, we're sending a token that the FE uses for authentication. Is this request also handled by the FE?

SatishRanjan: Hi @mathewc - in this case request goes to the build service (kudulite) running in the same Kubernetes cluster as the function host, and the encryption key for the function app will also be accessible to build service, which is used to decrypt the token and pass the authentication successfully. The function app access token is created/updated through the geo-master control plane code as a Kubernetes secret and build service will have access to read the secret. So, this flow in Kubernetes doesn't hit the Antares front end. Hope this helps.

mathewc: Ok, so the token is actually required (not ignored) in your scenario - that's all I was trying to establish.

SatishRanjan: Got it, yes that's right @mathewc.

mathewc (dismissed):

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -291,8 +291,17 @@ public static bool IsAppService(this IEnvironment environment)
...
         {
-            return !string.IsNullOrEmpty(environment.GetEnvironmentVariable(KubernetesServiceHost))
+            try
+            {
+                return !string.IsNullOrEmpty(environment.GetEnvironmentVariable(KubernetesServiceHost))
                 && !string.IsNullOrEmpty(environment.GetEnvironmentVariable(PodNamespace));
+            }
+            catch
+            {
+                //if there's an error reading the environment variable then return false
```
mathewc: I'd rather you just fix the test to mock things appropriately. Let's not have test code here.

SatishRanjan:
@mathewc - could you please take a look again? I've resolved your comments.

mathewc (dismissed):

pragnagopa:
@SatishRanjan  - JIT Trace is updated. Will merge the PR once the build is green.

pragnagopa (approved):

SatishRanjan:
Thanks a ton @pragnagopa :)

PR #6913, opened 2020-11-19
Fixing metrics Timespan persistence

mathewc:
Currently we're not actually persisting Timespan types when writing table metrics. This currently affects QueueTrigger/ServiceBusTrigger since they both have Timespan properties for QueueTime.

Currently our TableEntityConverter skips persistence of property types it doesn't know how to convert, meaning Timespans weren't being written.

### Pull request checklist

    * [x] Otherwise: Backport tracked by issue/PR https://github.com/Azure/azure-functions-host/pull/6915

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/Scale/TableEntityConverter.cs
+++ b/src/WebJobs.Script.WebHost/Scale/TableEntityConverter.cs
@@ -96,6 +96,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Scale
                     entityProperty = new EntityProperty(value.ToObject<long>());
                     return true;
                 case JTokenType.String:
+                case JTokenType.TimeSpan:
                     entityProperty = new EntityProperty(value.ToObject<string>());
                     return true;
                 default:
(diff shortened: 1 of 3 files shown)


src/WebJobs.Script.WebHost/Scale/TableEntityConverter.cs
```diff
@@ -96,6 +96,7 @@ public static bool TryGetEntityProperty(JProperty property, out EntityProperty e
                     entityProperty = new EntityProperty(value.ToObject<long>());
                     return true;
                 case JTokenType.String:
+                case JTokenType.TimeSpan:
```
mathewc, line 99: We recently moved to the latest version of Azure Storage SDK, and ideally we'd be taking advantage of [TableEntity.Flatten](https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.cosmos.table.tableentity.flatten?view=azure-dotnet) which was newly added, rather than this custom conversion code we had to write.

To limit the scope of the changes for this fix, I won't make those changes here, since we have to be careful. I've logged https://github.com/Azure/azure-functions-host/issues/6914 for that.

soninaren (approved):

PR #6930, opened 2020-11-23
[RPC]Lock process restarts on WorkerError

### LINKED ISSUE #6919: Allow only one restart worker channel operation at a time
cc @brettsam @yojagad 

If a function app is using multiple worker processes configured via app setting `FUNCTIONS_WORKER_COUNT` -
 Language worker processes are launched 10secs apart:

https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs#L142-L149

If multiple language worker channels need to be restarted at the same time - for example - function invocations timed out across various channels - restarts need to be staggered as well to avoid resource contention:

https://github.com/Azure/azure-functions-host/blob/068209f79864805d210ba5df1e05ca7b92033b9d/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs#L383

pragnagopa:

resolves #6919

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
+++ b/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
@@ -29,6 +29,7 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
         private readonly IEnvironment _environment;
         private readonly IApplicationLifetime _applicationLifetime;
         private readonly TimeSpan _shutdownTimeout = TimeSpan.FromSeconds(10);
+        private readonly TimeSpan _restartWait = TimeSpan.FromSeconds(10);
         private readonly TimeSpan thresholdBetweenRestarts = TimeSpan.FromMinutes(WorkerConstants.WorkerRestartErrorIntervalThresholdInMinutes);
 
         private IScriptEventManager _eventManager;
@@ -49,6 +50,7 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
         private ConcurrentStack<WorkerErrorEvent> _languageWorkerErrors = new ConcurrentStack<WorkerErrorEvent>();
         private CancellationTokenSource _processStartCancellationToken = new CancellationTokenSource();
         private int _debounceMilliSeconds = (int)TimeSpan.FromSeconds(10).TotalMilliseconds;
+        private SemaphoreSlim _restartWorkerProcessSLock = new SemaphoreSlim(1, 1);
 
         public RpcFunctionInvocationDispatcher(IOptions<ScriptJobHostOptions> scriptHostOptions,
             IMetricsLogger metricsLogger,
@@ -380,7 +382,18 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
         {
             if (_languageWorkerErrors.Count < ErrorEventsThreshold)
             {
-                await InitializeJobhostLanguageWorkerChannelAsync(_languageWorkerErrors.Count);
+                try
+                {
+                    // Issue only one restart at a time.
+                    await _restartWorkerProcessSLock.WaitAsync();
+                    await InitializeJobhostLanguageWorkerChannelAsync(_languageWorkerErrors.Count);
+                }
+                finally
+                {
+                    // Wait before releasing the lock to give time for the process to startup and initialize.
+                    await Task.Delay(_restartWait);
+                    _restartWorkerProcessSLock.Release();
+                }
             }
             else if (_jobHostLanguageWorkerChannelManager.GetChannels().Count() == 0)
             {
@@ -417,6 +430,7 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
                 _jobHostLanguageWorkerChannelManager.ShutdownChannels();
                 State = FunctionInvocationDispatcherState.Disposed;
                 _disposed = true;
+                _restartWorkerProcessSLock.Dispose();
             }
         }
 
--- a/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
+++ b/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
@@ -384,7 +384,7 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
 
         private async Task RestartWorkerChannel(string runtime)
         {
-            if (!_disposing || !_disposed)
+            if (_disposing)
             {
                 return;
             }
--- a/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
+++ b/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
@@ -50,7 +50,6 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
         private IEnumerable<FunctionMetadata> _functions;
         private ConcurrentStack<WorkerErrorEvent> _languageWorkerErrors = new ConcurrentStack<WorkerErrorEvent>();
         private CancellationTokenSource _processStartCancellationToken = new CancellationTokenSource();
-        private CancellationTokenSource _disposeToken = new CancellationTokenSource();
         private int _debounceMilliSeconds = (int)TimeSpan.FromSeconds(10).TotalMilliseconds;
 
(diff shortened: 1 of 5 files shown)


src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -376,11 +378,25 @@ internal bool ShouldRestartWorkerChannel(string runtime, bool isWebHostChannel,
...
+                    await _restartWorkerProcessSLock.WaitAsync();
+                    await InitializeJobhostLanguageWorkerChannelAsync(_languageWorkerErrors.Count);
+                }
+                finally
+                {
+                    if (!_disposing)
+                    {
+                        // Wait before releasing the lock to give time for the process to startup and initialize.
+                        await Task.Delay(_restartWait);
+                        _restartWorkerProcessSLock.Release();
```
yojagad, line 411: Can we also do a Release() inside the TaskContinuation within InitializeJobhostLanguageWorkerChannelAsync ? This way if initialization process startup completes faster than _restartWait, another thread can still enter. (I'm assuming calling Release() twice doesn't throw an exception here)

pragnagopa: Intentionally blocking to give more time for the process that is initialized to process invocations before starting new processes. This is similar to debouce we have in place during initialization

yojagad: I was assuming Task.Delay() is giving time for this (https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs#L116) to complete. Is it an additional time on top of that ?

pragnagopa: Yes, intentional delay before releasing the lock is to enforce delay between process start up operations similar to what we have here: https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs#L142-L149

yojagad: Okay, got it . Thanks for clarifying that !

yojagad (dismissed):

yojagad:
Pinging @pragnagopa since I wasn't able to add her to the review.

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -376,11 +382,29 @@ internal bool ShouldRestartWorkerChannel(string runtime, bool isWebHostChannel,
...
+                try
+                {
+                    // Issue only one restart at a time.
+                    await _restartWorkerProcessSLock.WaitAsync();
+                    await InitializeJobhostLanguageWorkerChannelAsync(_languageWorkerErrors.Count);
+                }
+                finally
+                {
+                    // Wait before releasing the lock to give time for the process to startup and initialize.
+                    await Task.Delay(_restartWait);
```
pragnagopa: Changes look OK!  Instead of exposing disposed, it would help to pass cancellationToken to [Task.Delay](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.delay?view=netcore-3.1#System_Threading_Tasks_Task_Delay_System_TimeSpan_System_Threading_CancellationToken_)

Can you give it a try and see if setting the cancellation token in "Dispose" will do the trick?

yojagad: I think Task.Delay with a cancellationToken would throw an exception if task is canceled before timeout... How about TaskCompletionSource<<bool>> disposeTask that gets set (and therefore runs to completion) when dispose is invoked? Other conditions can check to see if this task ran to completion or not... and `await Task.Delay(_restartWait)` can become `await Task.WhenAny(Task.Delay(_restartWait), disposeTask)` followed by a check below to see if disposeTask's status is completed?

pragnagopa: Take a look at discussion [here ](https://stackoverflow.com/questions/20509158/taskcanceledexception-when-calling-task-delay-with-a-cancellationtoken-in-an-key) . Try the following:

```
Task.Delay(_restartWait).ContinueWith(tsk => {
   //code to run after the delay goes here
}, cancellationToken.Token);

```

yojagad: The other reason I suggested the above is, we can't dispose that _dispose cancellation token like we do for `_processStartCancellationToken` given that something in finally block is expecting it. Anyway, trying it out now..

yojagad: Anyway, that still seemed to throw an exception so reverted it.

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -376,11 +383,33 @@ internal bool ShouldRestartWorkerChannel(string runtime, bool isWebHostChannel,
...
+                {
+                    // Wait before releasing the lock to give time for the process to startup and initialize.
+                    try
+                    {
+                        await Task.Delay(_restartWait, _disposeToken.Token);
+                        _restartWorkerProcessSLock.Release();
+                    }
+                    catch (TaskCanceledException ex)
+                    {
+                        _logger.LogError($"Restart wait task canceled possibly because host instance is being disposed/recycled", ex);
```
pragnagopa: if this does work...no need to log error...ignoring is OK!

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -310,7 +313,7 @@ internal async Task<IEnumerable<IRpcWorkerChannel>> GetInitializedWorkerChannels
 
         public async void WorkerError(WorkerErrorEvent workerError)
         {
-            if (!_disposing)
+            if (!_disposing && !_disposed)
             {
```
pragnagopa, line 317: shouldn't this be ``` || ``` ?

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -310,7 +313,7 @@ internal async Task<IEnumerable<IRpcWorkerChannel>> GetInitializedWorkerChannels
 
         public async void WorkerError(WorkerErrorEvent workerError)
         {
-            if (!_disposing)
+            if (!_disposing && !_disposed)
```
pragnagopa: nit: lets use similar pattern to below ... just a bit easier to read
```
if (_disposing || _disposed)
            {
                return;
            }
```

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -328,7 +331,7 @@ public async void WorkerError(WorkerErrorEvent workerError)
 
         public async void WorkerRestart(WorkerRestartEvent workerRestart)
         {
-            if (!_disposing)
+            if (!_disposing && !_disposed)
```
pragnagopa: same here. change to 
```
if (_disposing || _disposed)
            {
                return;
            }
```

yojagad: Updated the pattern here and in other places too

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
yojagad: Removed log.

pragnagopa (commented):
Approved :) I cannot click approve here as I started the PR.

yojagad (approved):

Issue #6955, opened 2020-12-01
Give Language Workers way to specify default function file names


mhoeger:
Example: "index.js"
This would let us remove this hard-coding:
https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Host/FunctionMetadataProvider.cs#L245

pragnagopa:
@vrdmr  / @Hazhzeng - can you please point to code how default file name is handled in python worker if script file is not specified?

pragnagopa:
@mhoeger - I believe this issue needs to be tracked in the nodejs worker repo

mhoeger:
@pragnagopa - I don't think worker.config.json exposes this capability, my understanding of Python is that it adds the scriptFile property explicitly to the templates: https://github.com/Azure/azure-functions-templates/blob/0b0cdcadeed8860adf14aafe3e81b7d60992547b/Functions.Templates/Templates/BlobTrigger-Python/function.json#L2

https://docs.microsoft.com/en-us/azure/azure-functions/functions-reference-python#programming-model

For Node.js though, I think we need to specify default file name through worker.config.json

vrdmr:
@pragnagopa - As @mhoeger pointed out - The [`scriptFile`](https://docs.microsoft.com/en-us/azure/azure-functions/functions-reference-python#alternate-entry-point) property is explicitly added to function.json in the templates. We normally default it to `__init__.py`.

vrdmr:
@mhoeger - Could the `scriptFile` help your case?

mhoeger:
@vrdmr - Probably a good idea to add that explicitly to templates (making it easier for people to change), but this can't be the solution because it's a significant breaking change to existing apps

vrdmr:
Ah - yes. You'll need a combination of the two - If scriptFile doesn't exist, then send the default file name.

pragnagopa:
We should consider schema / values specified in `worker.config.json` - to have significance both on the host end and worker end
In this case, default script file if not specified in the function.json can be handled by the worker as host does not need to know this value at all. Thoughts?

mhoeger:
I think there still needs to be a host change? :/ I was running into this exception:
```
Unable to determine the primary function script.Make sure atleast one script file is present.Try renaming your entry point script to 'run' or alternatively you can specify the name of the entry point script explicitly by adding a 'scriptFile' property to your function metadata.
```
From here:
[...]

PR #6956, opened 2020-12-01
Update Node.js Worker to 2.1.0

### LINKED ISSUE #344: Add support for ES Modules
Allow Node 14 functions to use ESM by supporting `.mjs` files.

A function would look something like this: 

```js
// index.mjs
export default async function (context, req) {
    context.res = {
        body: process.version
    };
};
```

mhoeger:
Change also to support index.mjs in the simplest way possible. This is being done to make sure ES Modules make it into this next deployment, while saving the engineering improvement to do this the right, extensible way for a later item this sprint: https://github.com/Azure/azure-functions-host/issues/6955 . **Will remove hard-coding soon!**

Node.js Worker 2.1.0: https://github.com/Azure/azure-functions-nodejs-worker/releases/tag/v2.1.0

resolves https://github.com/Azure/azure-functions-nodejs-worker/issues/344

### Pull request checklist

    * [X] Otherwise: I've added my notes to `release_notes.md`

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -4,7 +4,7 @@
 -->
 - Update Python Worker Version to [1.1.10](https://github.com/Azure/azure-functions-python-worker/releases/tag/1.1.10)
 - Configure host.json to use workflow when creating a default host.json and app is identified as a logic app. (#6810)
-
+- Update Node.js Worker Version to [2.1.0](https://github.com/Azure/azure-functions-nodejs-worker/releases/tag/v2.1.0)
 
 **Release sprint:** Sprint 89
 [ [bugs](https://github.com/Azure/azure-functions-host/issues?q=is%3Aissue+milestone%3A%22Functions+Sprint+89%22+label%3Abug+is%3Aclosed) | [features](https://github.com/Azure/azure-functions-host/issues?q=is%3Aissue+milestone%3A%22Functions+Sprint+89%22+label%3Afeature+is%3Aclosed) ]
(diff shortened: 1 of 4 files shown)


pragnagopa:
Please add link to node js worker release notes in the PR description

pragnagopa:
Can you please check if  the changes in the PR can be handled in nodejs worker i.e if script file is empty or not set in function load request - > default to `index.js` or `index.mjs`

mhoeger:
@pragnagopa - Tested and if script file is empty, this change works to find `index.mjs` (or `index.js`)! The case of a script file not being set in function load request doesn't happen because of the validation here: 
https://github.com/Azure/azure-functions-host/blob/c198263bdcf7539c0765fdea6a8ec82b93c90a70/src/WebJobs.Script/Host/FunctionMetadataManager.cs#L167-L173

pragnagopa:
Thanks @mhoeger  for the the follow up.
So the check you pointed out can be removed : https://github.com/Azure/azure-functions-host/blob/c198263bdcf7539c0765fdea6a8ec82b93c90a70/src/WebJobs.Script/Host/FunctionMetadataManager.cs#L167-L173  if NodeJs worker adds this check as part of function load response? - Will you be handling this as part of the follow up issue #6955 ?

pragnagopa (approved):

mhoeger:
@pragnagopa - I think we'll need that check still since it's relevant for C# :\ What I was thinking of doing was changing that check from being hard-coded to a value from the worker config. But yes was planning on exploring solutions in #6955 !

Issue #6973, opened 2020-12-09
Make MetricEvent.Completed more robust


mathewc:
The check performed to determine whether an event is complete is based on checking whether the Duration (as set when the event ends) is equal to the default TimeSpan. When an event ends, generally Duration is set to DateTime.UtcNow - Timestamp. However for events that happen quickly (within the resolution of the system clock, e.g. 15ms) these times can be equal, resulting in Completed returning false, when it should return true. We should change this event class so completion is explicitly set as a boolean on event completion and not computed.

Based on how this Completed property is used, it can cause metering issues.

The fact that DateTime.UtcNow can return the same value on subsequent calls can be reproed via the below code. Note that the repro is system dependent.

```csharp
for (int i = 0; i < 1000000; i++)
{
    var t1 = DateTime.UtcNow;
    var delta = DateTime.UtcNow - t1;
    if (delta == default(TimeSpan))
    {
        throw new Exception("Kaboom!");
    }
}
```

PR #6974, opened 2020-12-09
Make MetricEvent.Completed more robust (#6973)

mathewc:
Fix for https://github.com/Azure/azure-functions-host/issues/6973.

### DIFF
--- a/src/WebJobs.Script/Diagnostics/MetricEvent.cs
+++ b/src/WebJobs.Script/Diagnostics/MetricEvent.cs
@@ -14,7 +14,7 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics
 
         public Stopwatch StopWatch { get; set; }
 
-        public TimeSpan Duration { get; set; }
+        public TimeSpan Duration { get; private set; }
 
         public string Data { get; set; }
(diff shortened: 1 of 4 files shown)


src/WebJobs.Script/Diagnostics/MetricEvent.cs
```diff
@@ -8,26 +8,45 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics
...
+        {
+            get
+            {
+                return _duration;
+            }
+
+            set
+            {
+                _duration = value;
+                _completed = true;
```
mathewc: To minimize changes, I'm maintaining the existing behavior where setting Duration leads to the event being completed. We could also have made things more explicit by requiring Completed to be explicitly set, but I think it's fine leaving it inferred.

fabiocav: I really think this would be a lot clearer with a `Complete(TimeSpan duration)` method instead of a settable property. Can we please make that change?

paulbatum (approved):

fabiocav (approved):

PR #6975, opened 2020-12-10
[AzureMonitor]User logs for Linux Dedicated/EP Sku

### LINKED ISSUE #6887: Linux Dedicated/Premium: Write console logs to stdout
Part of #6215. 

Currently, user logs don't appear in Azure Monitor for function apps running on Linux dedicated and premium plans. By piping user logs to stdout, we can take advantage of existing infrastructure that writes stdout to AzMon. This is already done for Linux Consumption, so we should be able to use the same logic.

Right now stdout output goes to the AppServiceConsoleLogs table. Ideally, all output will go to a single, better-named table (FunctionsLogs) with filters for the user to separate specific streams, ex. Durable Functions specific output. Additional work required on Antares.

kashimiz:

resolves #6887

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
@@ -10,11 +10,16 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
     {
         private readonly LinuxAppServiceFileLoggerFactory _loggerFactory;
         private readonly HostNameProvider _hostNameProvider;
+        private readonly bool _consoleEnabled = true;
 
         public LinuxAppServiceEventGenerator(LinuxAppServiceFileLoggerFactory loggerFactory, HostNameProvider hostNameProvider)
         {
             _loggerFactory = loggerFactory;
             _hostNameProvider = hostNameProvider ?? throw new ArgumentNullException(nameof(hostNameProvider));
+            if (Environment.GetEnvironmentVariable(EnvironmentSettingNames.ConsoleLoggingDisabled) == "1")
+            {
+                _consoleEnabled = false;
+            }
         }
 
         public static string TraceEventRegex { get; } = $"(?<Level>[0-6]),(?<SubscriptionId>[^,]*),(?<HostName>[^,]*),(?<AppName>[^,]*),(?<FunctionName>[^,]*),(?<EventName>[^,]*),(?<Source>[^,]*),\"(?<Details>.*)\",\"(?<Summary>.*)\",(?<HostVersion>[^,]*),(?<EventTimestamp>[^,]+),(?<ExceptionType>[^,]*),\"(?<ExceptionMessage>.*)\",(?<FunctionInvocationId>[^,]*),(?<HostInstanceId>[^,]*),(?<ActivityId>[^,\"]*)";
@@ -69,8 +74,17 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
             logger.Log(evt);
         }
 
+        private void ConsoleWriter(string evt)
+        {
+            if (_consoleEnabled)
+            {
+                Console.WriteLine(evt);
+            }
+        }
+
         public override void LogAzureMonitorDiagnosticLogEvent(LogLevel level, string resourceId, string operationName, string category, string regionName, string properties)
         {
+            ConsoleWriter($"{ScriptConstants.LinuxAzureMonitorEventStreamName} {(int)ToEventLevel(level)},{resourceId},{operationName},{category},{regionName},{NormalizeString(properties.Replace("'", string.Empty))},{"CONTAINER_NAME_ALT"},{"TENANT_ID"},{DateTime.UtcNow.ToString()}");
         }
 
         public static void LogUnhandledException(Exception e)
--- a/src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
@@ -25,7 +25,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 
         public static string DetailsEventRegex { get; } = $"(?<AppName>[^,]*),(?<FunctionName>[^,]*),\"(?<InputBindings>.*)\",\"(?<OutputBindings>.*)\",(?<ScriptType>[^,]*),(?<IsDisabled>[0|1])";
 
-        public static string AzureMonitorEventRegex { get; } = $"{ScriptConstants.LinuxAzureMonitorEventStreamName} (?<Level>[0-6]),(?<ResourceId>[^,]*),(?<OperationName>[^,]*),(?<Category>[^,]*),(?<RegionName>[^,]*),\"(?<Properties>[^,]*)\",(?<ContainerName>[^,\"]*),(?<TenantId>[^,\"]*),(?<EventTimestamp>[^,]+)";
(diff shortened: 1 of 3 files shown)


src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
```diff
@@ -69,8 +74,17 @@ private static void WriteEvent(LinuxAppServiceFileLogger logger, string evt)
             logger.Log(evt);
         }
 
+        private void ConsoleWriter(string evt)
```
fabiocav: nit: method name -> `WriteEvent`?

Argument name to something fully spelled out (e.g. eventPayload, etc.)

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
```diff
@@ -10,11 +10,16 @@ internal class LinuxAppServiceEventGenerator : LinuxEventGenerator
     {
         private readonly LinuxAppServiceFileLoggerFactory _loggerFactory;
         private readonly HostNameProvider _hostNameProvider;
+        private readonly bool _consoleEnabled = true;
 
         public LinuxAppServiceEventGenerator(LinuxAppServiceFileLoggerFactory loggerFactory, HostNameProvider hostNameProvider)
         {
             _loggerFactory = loggerFactory;
             _hostNameProvider = hostNameProvider ?? throw new ArgumentNullException(nameof(hostNameProvider));
+            if (Environment.GetEnvironmentVariable(EnvironmentSettingNames.ConsoleLoggingDisabled) == "1")
```
fabiocav: Is this check expected on Linux dedicated?

fabiocav:
Can you please update the PR description? We can wait to review until this moves out of draft.

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
balag0: Some background on this appsetting: It was added for investigating performance impact of different logging mechanisms in consumption. So not needed in dedicated.

pragnagopa:
cc @CooperLink

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
```diff
@@ -8,13 +8,20 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 {
     internal class LinuxAppServiceEventGenerator : LinuxEventGenerator
     {
+        private readonly IEnvironment _environment;
+        private readonly Action<string> _writeEvent;
         private readonly LinuxAppServiceFileLoggerFactory _loggerFactory;
         private readonly HostNameProvider _hostNameProvider;
+        private string _tenantId;
```
divyagandhisethi: Why are these fields needed?

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
```diff
@@ -8,13 +8,20 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
...
+        private string _containerName;
 
-        public LinuxAppServiceEventGenerator(LinuxAppServiceFileLoggerFactory loggerFactory, HostNameProvider hostNameProvider)
+        public LinuxAppServiceEventGenerator(IEnvironment environment, LinuxAppServiceFileLoggerFactory loggerFactory, HostNameProvider hostNameProvider, Action<string> writeEvent = null)
         {
+            _writeEvent = writeEvent ?? WriteEvent;
+            _environment = environment;
             _loggerFactory = loggerFactory;
             _hostNameProvider = hostNameProvider ?? throw new ArgumentNullException(nameof(hostNameProvider));
+            _containerName = _environment.GetEnvironmentVariable(EnvironmentSettingNames.ContainerName)?.ToUpperInvariant();
```
divyagandhisethi: Looks like this was lifted from the LinuxContainerEventGenerator. We don't set these environment variables for Linux App Service.

divyagandhisethi (changes requested):

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
```diff
@@ -69,8 +90,14 @@ private static void WriteEvent(LinuxAppServiceFileLogger logger, string evt)
             logger.Log(evt);
         }
 
+        private void WriteEvent(string eventPayload)
+        {
+            Console.WriteLine(eventPayload);
```
pragnagopa, line 78: @brettsam - is this needed if  `AzureFunctionsJobHost__Logging__Console__IsEnabled` is set to `true` ?

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
kashimiz: I wasn't sure if these fields were applicable to logs from apps running on Linux dedicated and/or premium. I'll remove them, as well as the corresponding placeholders from the log statement.

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
brettsam: Good question -- the Console Logger will write out logs with extra "stuff" around the message itself (like `Info:`). I'm assuming that Antares will use a regex to identify these logs -- do we know if that would match?

kashimiz: @brettsam - I assume that the behavior of ConsoleLogger would be the same on the Linux Consumption plan, as in [LinuxContainerEventGenerator](https://github.com/Azure/azure-functions-host/blob/14adf6d336db5da0feb0221cc4fe0a40bc14edb0/src/WebJobs.Script.WebHost/Diagnostics/LinuxContainerEventGenerator.cs)?
@pragnagopa or @divyagandhisethi - do you know how Antares handles the stdout for Linux Consumption?

balag0: I think the log format depends on the logging driver used ( the one used in Antares might not be the same as what is used on Linux Consumption but I think they are the same currently also Antares never sees any of the raw logs from Linux Consumption as it is handled elsewhere)

In this case as long as the eventual log format emitted by the containers are parsed correctly by the log parsers used in Antares everything should be good. If a particular log entry doesnt match the format the parsers expect then it will be dropped and wont reach AzureMonitor. Testing this on a Linux worker will help confirm/clarify things greatly :)

pragnagopa: Ideally we do not have to write to console directly. I will track this as a separate issue to remove `cosole.writeline` across linux skus and set `AzureFunctionsJobHost__Logging__Console__IsEnabled` and test them out. resolving this comment.

pragnagopa:
@kashimiz  - can you please resolve conversations if already addressed?

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
kashimiz: If those fields aren't populated, they're not. I think I was hewing a little too closely to what was being done in LinuxContainerEventGenerator. I'll remove them.

kashimiz:
@divyagandhisethi - requested changes have been made. Could you take a look?

divyagandhisethi:
> 
> 
> @divyagandhisethi - requested changes have been made. Could you take a look?

Hi @kashimiz . I am sorry I don't see your new commits after the last iteration I reviewed. Can you share the commit id you pushed?

divyagandhisethi (dismissed):

pragnagopa:
@kashimiz - can you please take a look at the Unit test failures?

pragnagopa (approved):

PR #7006, opened 2020-12-22
Update specialization logic  for linux consumption containers running on pinned host versions

divyagandhisethi:

When an app is running on a pinned mode, we do not start in placeholder mode, since the container is already assigned to the customer app at the time of startup. This change updates our specialization logic to account for pinned containers running in non placeholder mode.

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.WebHost/Management/InstanceManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/InstanceManager.cs
@@ -100,7 +100,13 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
         {
             if (!_webHostEnvironment.InStandbyMode)
             {
-                _logger.LogError("Assign called while host is not in placeholder mode");
+                // This is only true when specializing pinned containers.
+                _logger.LogInformation("Assign called while host is not in placeholder mode.");
+            }
+
+            if (_environment.IsContainerReady())
+            {
+                _logger.LogError("Assign called while container is marked as specialized.");
                 return false;
             }
 
--- a/test/WebJobs.Script.Tests.Integration/Management/InstanceManagerTests.cs
+++ b/test/WebJobs.Script.Tests.Integration/Management/InstanceManagerTests.cs
@@ -102,15 +102,16 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.Managment
                 p => Assert.StartsWith("Applying 1 app setting(s)", p),
                 p => Assert.StartsWith("Triggering specialization", p));
 
-            // calling again should return false, since we're no longer
-            // in placeholder mode
+            // calling again should return false, since we have 
+            // already marked the container as specialized.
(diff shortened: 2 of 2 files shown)


src/WebJobs.Script.WebHost/Management/InstanceManager.cs
```diff
@@ -100,7 +100,13 @@ public bool StartAssignment(HostAssignmentContext context)
         {
             if (!_webHostEnvironment.InStandbyMode)
             {
-                _logger.LogError("Assign called while host is not in placeholder mode");
+                // This is only true when specializing pinned containers.
```
balag0, line 103: Is this change to address some other changes elsewhere or is this fixing an existing issue?
It seems strange to continue specialization when the container is not in placeholder mode. Can you please describe the scenario..

divyagandhisethi:
@balag0 Please look at this.

balag0:
Could you please link to the issue this PR is fixing. Just trying to understand the scenario.

divyagandhisethi:
Fixes #7029

divyagandhisethi:
> 
> 
> @balag0 Please look at this.

I have created a new issue for this, let me know if you have any other questions.

src/WebJobs.Script.WebHost/Management/InstanceManager.cs
balag0: Just curious, can we instead set WEBSITE_PLACEHOLDER_MODE to 1 at container creation like it used to be? (I dont remember why we made that change) Will something break if we do that? since it seems like this change is needed as a result of a prev change and if we can just remove both changes that would be good as well.

Since this is intended only for custom containers, can we use
!string.isnullorempty(env[CONTAINER_START_CONTEXT]) &&  _environment.IsContainerReady()

divyagandhisethi: Ok I will revert my change to disable placeholder mode for custom containers. You can make the change to reduce cold start for pinned containers as you see fit.

divyagandhisethi:
Abandoning this PR.

src/WebJobs.Script.WebHost/Management/InstanceManager.cs
balag0: Sorry. I don't understand the reference to cold start. could you explain please.
The issue you linked to this PR https://github.com/Azure/azure-functions-host/issues/7029 doesnt seem related to cold start either. 
Are you saying abandoing this PR will increase cold start?

divyagandhisethi: Can you please look at https://github.com/Azure/azure-functions-host/pull/6848#issuecomment-717415907

balag0: Okay. I don't see any workitems for improving cold start for pinned apps. Could you open one please so we dont miss this. Thanks

PR #7068, opened 2021-01-18
[Linux consumption] Notify health events to mesh service process

balag0:
Notifies mesh service process of fatal errors during specialization. These events are used to monitor overall container health.

related to https://github.com/Azure/azure-functions-host/issues/5222

### Pull request checklist

    * [X ] Otherwise: Backport tracked by issue/PR https://github.com/Azure/azure-functions-host/issues/5222

### Additional information

Additional PR information
Corresponding server side changes
https://github.com/Azure/azure-functions-docker-private/pull/131

### DIFF
--- a/src/WebJobs.Script.WebHost/Management/IMeshServiceClient.cs
+++ b/src/WebJobs.Script.WebHost/Management/IMeshServiceClient.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Script.WebHost.Models;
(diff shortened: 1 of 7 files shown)


src/WebJobs.Script.WebHost/Management/IMeshServiceClient.cs
```diff
@@ -16,5 +17,7 @@ public interface IMeshServiceClient
         Task MountFuse(string type, string filePath, string scriptPath);
 
         Task PublishContainerActivity(IEnumerable<ContainerFunctionExecutionActivity> activities);
+
+        Task NotifyHealthEvent(ContainerHealthEventType healthEventType, Type source, string details);
```
mathewc, line 21: I think you should generalize this to PublishEvent so it can be used for other event types in the future. Then you'd have ContainerEventType, ContainerEvent, etc.

balag0: I spent some time thinking about this but couldn't come up with other event types we want to publish :)
So leaning towards just leaving this as is for now and generalize the api later if required. 
Any other events are you thinking of? The primary purpose for publishing the health event is to take remedial action at runtime vs just for logging purposes.

mathewc (approved):

PR #7084, opened 2021-01-21
Stop popping up console windows during programmatic builds

shyamnamboodiripad:
Changes GenerateFunctionsExtensionsMetadata MSBuild task to avoid popping up console window for dotnet.exe during programmatic builds launched from within processes that don't have a console window.

If the launching process has an associated console window (e.g. when MSBuild is launched from command line), CreateNoWindow = false has no effect. However, if the launching process does not have an allocated console window (which is what happens in the case of the Live Unit Testing feature in VS) then setting CreateNoWindow = false ends up popping up console window for the child dotnet.exe process.

This is particularly problematic for Live Unit Testing. Live Unit Testing constantly performs builds in the background as the user changes their code in the VS editor. Without this fix, any code change that requires to rebuild an Azure Functions project in the user's solution results in popping up the above console window. The repeated pop ups end up impeding the user's ability to type in the editor.

Since standard input and output are already being redirected, the popup console window was not particularly helpful anyways. I suspect the original intention of this code may have been to set CreateNoWindow = true and that it may have been set to false inadvertently.

### Pull request checklist

### DIFF
--- a/tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator/BuildTasks/GenerateFunctionsExtensionsMetadata.cs
+++ b/tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator/BuildTasks/GenerateFunctionsExtensionsMetadata.cs
@@ -37,7 +37,7 @@ namespace ExtensionsMetadataGenerator.BuildTasks
             var info = new ProcessStartInfo
             {
                 UseShellExecute = false,
-                CreateNoWindow = false,
+                CreateNoWindow = true,
                 RedirectStandardError = true,
                 RedirectStandardOutput = true,
                 WorkingDirectory = Path.Combine(Path.GetDirectoryName(taskAssembly.Location), "..", "netstandard2.0", "generator"),


shyamnamboodiripad:
FYI @vijayrkn This is the issue we discussed in email. I am not sure if the branch I am targeting is the correct one. Would appreciate any guidance :)

vijayrkn:
@shyamnamboodiripad - I haven't worked in this repo much but I would assume v3.x

@brettsam - Can you please help ?

brettsam (commented):
This certainly looks like the right fix here (see how we do this in FunctionGenerator here: https://github.com/Azure/azure-functions-vs-build-sdk/blob/main/src/Microsoft.NET.Sdk.Functions.MSBuild/Tasks/GenerateFunctions.cs#L111). 

@fabiocav -- any reason why this was set to `false` originally?

brettsam (approved):

shyamnamboodiripad:
@brettsam Any ideas why the PR checks are not complete yet? Do I need any special permissions to trigger those jobs?

shyamnamboodiripad:
Thank you @brettsam @fabiocav! 

I am not sure about the code flow for the dev branch, but could you please tell me when the fix will be available in the product? Will the fix be available to users who create a new Azure Functions project in the upcoming VS 2019 16.9 Preview 4 release?

fabiocav:
The generator is released out-of-band, so we'll schedule a release at the end of the current sprint (ending next week)

shyamnamboodiripad:
@fabiocav Thanks!  Could you please confirm if this fix is released at this point? I am assuming the fix will be part of the `Microsoft.NET.Sdk.Functions` nuget package. If yes, which version of this package would contain the fix? I see that the latest version on nuget.org is currently 3.0.11 (and looks like this was released last November).

brettsam:
@shyamnamboodiripad -- we can get this out early next week. It'll be included in the new version of `Microsoft.Azure.WebJobs.Script.ExtensionsMetadataGenerator` (which is then included in `Microsoft.NET.Sdk.Functions`).

shyamnamboodiripad:
@brettsam Any updates on the nuget publish? Thanks!

shyamnamboodiripad:
@brettsam @fabiocav Any updates on the nuget publish? I noticed that the [`Microsoft.Azure.WebJobs.Script.ExtensionsMetadataGenerator`](https://www.nuget.org/packages/Microsoft.Azure.WebJobs.Script.ExtensionsMetadataGenerator) package has not been updated recently (latest available version is still `1.2.1`).

We are seeing more reports of this issue such as this feedback ticket https://developercommunity.visualstudio.com/t/When-using-live-unit-testing-console-win/1363141. Since the issue is already fixed, it would be great if you could please publish the nuget :) Thanks!

shyamnamboodiripad:
@brettsam @fabiocav I see a new version `1.2.2` was published for [`Microsoft.Azure.WebJobs.Script.ExtensionsMetadataGenerator`](https://www.nuget.org/packages/Microsoft.Azure.WebJobs.Script.ExtensionsMetadataGenerator) today and based on a quick test it seems to contain the current fix. Thank you for publishing.

brettsam:
Great, thanks for confirming!

PR #7085, opened 2021-01-22
Skip external config validation if RuntimeDrivenScaling is enabled in Elastic Premium

### LINKED ISSUE #6542: Allow External Startup to update the config when Runtime Scale Monitoring is enabled

I would like to be able to use Key Vault to manage secure secrets in applications using Premium Plan and VNET Integration / Private Endpoint.

In the current version (3.0.14287.0), the host fails to start with the following error, even if Runtime Scale Monitoring is enabled

>Microsoft.Azure.WebJobs.Script: The Functions scale controller may not scale the following functions correctly because some configuration values were modified in an external startup class. Function 'Function2' uses the modified key(s): DefaultCosmosConnection

I understand that the Scale Controller is not restricted in the Function App with Runtime Scale Monitoring enabled, so please change the behavior to the same as when you use the App Service Plan.

https://github.com/Azure/azure-functions-host/blob/dcdd1c251fb041d6828d7b2cdcad99a311518cbe/src/WebJobs.Script/ScriptHostBuilderExtensions.cs#L165

The App Service Key Vault Reference cannot be restricted using Service Endpoint / Private Endpoint, so I want to use External Startup.

soninaren:
Function host was throwing a host initialization exception when an configuration from external source was used in a trigger binding. The validation / exception was invalid when Runtime driven scaling was enabled for function app was hosted on elastic premium hosting plan. The change here is skips the validation for this specific case.

resolves #6542 

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -11,6 +11,7 @@
 - Updated [System.Data.SqlClient to 4.8.2](https://www.nuget.org/packages/System.Data.SqlClient/4.8.2)
 - Added direct refereces to [System.IO.Pipes](https://www.nuget.org/packages/System.IO.Pipes/4.3.0)  and [System.Threading.Overlapped](https://www.nuget.org/packages/System.Threading.Overlapped/4.3.0) to ensure System.Data.SqlClient package update does not impact unification 
 - Updated Java Worker Version to [1.8.2-SNAPSHOT](https://github.com/Azure/azure-functions-java-worker/releases/tag/1.8.2-SNAPSHOT)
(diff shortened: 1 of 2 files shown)


brettsam (commented):
Were you able to test this out before/after in a private site extension?

soninaren:
> Were you able to test this out before/after in a private site extension?

Yes, i tested with private site extension.

brettsam (approved):

PR #7086, opened 2021-01-22
Restart only if needed

### LINKED ISSUE #7089: Call restart only if FileMonitoringService scheduled a restart in FunctionController
We do not need explicitly call restart on function create/update/delete. Instead we need to call restart only if FileMonitoringService got an event for restart(appropriate file change).

alrod:
Resolves #7089

We do not need explicitly call restart on function create/update/delete. Instead we need to call restart only if FileMonitoringService got an event for restart(appropriate file change).

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.WebHost/Controllers/FunctionsController.cs
+++ b/src/WebJobs.Script.WebHost/Controllers/FunctionsController.cs
@@ -74,11 +74,10 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Controllers
 
             bool success, configChanged;
             FunctionMetadataResponse functionMetadataResponse;
-            using (fileMonitoringService.SuspendRestart())
+            using (fileMonitoringService.SuspendRestart(true))
             {
                 (success, configChanged, functionMetadataResponse) = await _functionsManager.CreateOrUpdate(name, functionMetadata, Request);
             }
-            await ScheduleRestartAsync(fileMonitoringService);
 
             if (success)
             {
@@ -175,11 +174,10 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Controllers
 
             bool deleted;
             string error;
-            using (fileMonitoringService.SuspendRestart())
+            using (fileMonitoringService.SuspendRestart(true))
             {
                 (deleted, error) = await _functionsManager.TryDeleteFunction(function);
             }
-            await ScheduleRestartAsync(fileMonitoringService);
 
             if (deleted)
             {
@@ -219,16 +217,6 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Controllers
             {
                 FileDownloadName = (System.Environment.GetEnvironmentVariable("WEBSITE_SITE_NAME") ?? "functions") + ".zip"
             };
-        }
-
-        private async Task ScheduleRestartAsync(IFileMonitoringService fileMonitoringService)
-        {
-            Task restartTask = null;
-            using (System.Threading.ExecutionContext.SuppressFlow())
-            {
-                restartTask = fileMonitoringService.ScheduleRestartAsync(false);
-            }
-            await restartTask;
         }
     }
 }
--- a/src/WebJobs.Script.WebHost/FileMonitoringService.cs
+++ b/src/WebJobs.Script.WebHost/FileMonitoringService.cs
@@ -38,6 +38,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
         private AutoRecoveringFileSystemWatcher _debugModeFileWatcher;
         private AutoRecoveringFileSystemWatcher _diagnosticModeFileWatcher;
         private FileWatcherEventSource _fileEventSource;
+        private bool _restartScheduled;
         private bool _shutdownScheduled;
         private long _restartRequested;
         private bool _disposed = false;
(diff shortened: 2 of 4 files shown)


src/WebJobs.Script.WebHost/IFileMonitoringService.cs
```diff
@@ -13,6 +13,6 @@ public interface IFileMonitoringService : IHostedService
 
         void ResumeRestart();
 
-        Task ScheduleRestartAsync(bool shutdown);
+        Task RestartIfScheduledAsync();
```
mathewc: First, it looks like the ResumeRestart method above is never called anywhere. Can't it be removed? Next, rather than adding this new method, how about just having SuspendRestart take a bool autoResume parameter that when true, will do this "restart if scheduled" logic internally? Likely it'll be the case that anytime you suspend, you want to "restart if scheduled" after, so why not bake it into the disposal pattern to prevent errors and simplify things?

src/WebJobs.Script.WebHost/Controllers/FunctionsController.cs
```diff
@@ -226,7 +226,7 @@ private async Task ScheduleRestartAsync(IFileMonitoringService fileMonitoringSer
             Task restartTask = null;
             using (System.Threading.ExecutionContext.SuppressFlow())
             {
-                restartTask = fileMonitoringService.ScheduleRestartAsync(false);
+                restartTask = fileMonitoringService.RestartIfScheduledAsync();
```
mathewc: The containing method name no longer matches the logic here, which is a bit confusing. However, maybe this method can just go away, as per suggestion below.

pragnagopa:
Please link to an issue or open an issue for this PR to help keep track of issues that needed hotfix in release pipeline

alrod:
> Please link to an issue or open an issue for this PR to help keep track of issues that needed hotfix in release pipeline

Related issue #7089

src/WebJobs.Script.WebHost/IFileMonitoringService.cs
alrod: Regarding ResumeRestart there was the discussion about the method:
https://github.com/Azure/azure-functions-host/pull/6799#discussion_r509661565

Regarding autoResume parameter
it will require to call async from sync Dispose() what it's not good.
There is DisposeAsync() in C# 8.0 but we still use C# 7.2.

pragnagopa:
@gzuber - Please prepare hotfix after this PR is merged.

src/WebJobs.Script.WebHost/FileMonitoringService.cs
```diff
@@ -99,31 +100,41 @@ public Task StopAsync(CancellationToken cancellationToken)
...
-            return new SuspendRestartRequest(this);
+            return new SuspendRestartRequest(this, autoRestart);
         }
 
-        public void ResumeRestart()
+        private async Task RestartIfScheduledAsync()
         {
-            Interlocked.Decrement(ref _suspensionRequestsCount);
-            _typedLogger.LogDebug($"Resuming restart. ({_suspensionRequestsCount} requests).");
+            if (_restartScheduled)
```
mathewc, line 110: Seems like we don't need the complexity of this new boolean. You can just use the existing _suspensionRequestsCount and in the Disposal, if this is going from 1 to 0 you know there were suspended restarts that can then be executed if autoResume is true.

src/WebJobs.Script.WebHost/FileMonitoringService.cs
```diff
@@ -411,6 +424,15 @@ public void Dispose()
                 {
                     Interlocked.Decrement(ref _fileMonitoringService._suspensionRequestsCount);
                     _fileMonitoringService._typedLogger.LogDebug($"Exiting restart suspension scope. ({_fileMonitoringService._suspensionRequestsCount} requests).");
+                    if (_autoResume)
```
mathewc, line 429: _suspensionRequestsCount might still be > 0 here right? Seems like you need to also check this value and only restart if it's now zero. E.g. you could have nested using blocks.

src/WebJobs.Script.WebHost/FileMonitoringService.cs
```diff
@@ -411,6 +424,15 @@ public void Dispose()
...
+                    {
+                        // Suppressing execution context as it can cause scope container disposing issue
+                        using (System.Threading.ExecutionContext.SuppressFlow())
+                        {
+                            // Attempting to restart in the background
+                            Task.Run(async () => await _fileMonitoringService.RestartIfScheduledAsync());
```
mathewc: Let's just remove the RestartIfScheduledAsync method and inline the work here. For logging, I'd just add a single log statement like "Resuming scheduled restart"

src/WebJobs.Script.WebHost/FileMonitoringService.cs
mathewc: Please add a test for that scenario - nested restart disposal blocks

src/WebJobs.Script.WebHost/FileMonitoringService.cs
alrod: We need to restart only if there was a file change for which a restart is needed. Checking _suspensionRequestsCount does not work, there can be a PUT which does not change the files.

src/WebJobs.Script.WebHost/FileMonitoringService.cs
alrod: We will check later `_suspensionRequestsCount` here:
https://github.com/Azure/azure-functions-host/blob/alrod/fixrestart/src/WebJobs.Script.WebHost/FileMonitoringService.cs#L131

src/WebJobs.Script.WebHost/FileMonitoringService.cs
alrod: I still think we need `RestartIfScheduledAsync` if we want to check if there was a file change which needs a restart.

src/WebJobs.Script.WebHost/FileMonitoringService.cs
alrod: Added the test for nested disposal blocks:
https://github.com/Azure/azure-functions-host/pull/7086/commits/8aefaff68c037a41b059959844baef47db59aa8c

mathewc (dismissed):

mathewc (approved):

PR #6566, opened 2020-08-27
[pack] FunctionAssemblyLoadContext: handling multiple assemblies with dependencies on different versions of runtime assemblies

brettsam:
resolves #6555 

Description of the scenario is here: https://github.com/Azure/azure-functions-host/issues/6555#issuecomment-681157384

I'm submitting this now to get a build created. But we'll wait until @fabiocav returns from vacation and can provide a review before merging. I'll also see if I can add tests (but this is a tough scenario). This will all happen next sprint.

* [x] My changes **do not** require documentation changes
    * [x] Otherwise: Documentation issue linked to PR
* [ ] My changes **should not** be added to the release notes for the next release
    * [ ] Otherwise: I've added my notes to `release_notes.md`
* [ ] My changes **do not** need to be backported to a previous version
    * [ ] Otherwise: Backport tracked by issue/PR #issue_or_pr
* [ ] I have added all required tests (Unit tests, E2E tests)

PR #7113, opened 2021-01-30
AssemblyLoadContext fix

brettsam:
Ready for review. Here's a comment from the PR that explains the whole scenario. In #6566 I added the logic to adjust our dependency loading based on the deps file. The scenario that was missing (and this fixes) is 3(b)(ii)

From a user standpoint, this is typically happening when there is a reference to version `2.2.+` of something like `Microsoft.Extensions.Hosting.Abstractions` or `Microsoft.Extensions.Configuration.Abstractions`. Until this hotfix is applied, the fix is to:
1. Pin back to 3.0.15815.
2. Update these references to 2.1 and re-deploy.

```
There are several possible scenarios:
  1. The assembly was found and the policy evaluator succeeded.
     - Return the assembly.

  2. The assembly was not found (meaning the policy evaluator wasn't able to run).
     - Return null as there's not much else we can do. This will fail and come back to this context
       through the fallback logic for another attempt. This is likely an error case so let it fail.

  3. The assembly was found but the policy evaluator failed.
     a. We did not adjust the requested assembly name via the deps file.
        - This means that the DefaultLoadContext is looking for the correct version. Return null and let
         it handle the load attempt.
     b. We adjusted the requested assembly name via the deps file. If we return null the DefaultLoadContext would attempt to
        load the original assembly version, which may be incorrect if we had to adjust it forward past the runtime's version.
        i.  The adjusted assembly name is higher than the runtime's version.
            - Do not trust the DefaultLoadContext to handle this as it may resolve to the runtime's assembly. Instead,
              call LoadCore() to ensure the adjusted assembly is loaded.
        ii. The adjusted assembly name is still lower than or equal to the runtime's version (this likely means
            a lower major version).
            - Return null and let the DefaultLoadContext handle it. It may come back here due to fallback logic, but
              ultimately it will unify on the runtime version.
```

### DIFF
--- a/src/WebJobs.Script/Description/DotNet/FunctionAssemblyLoadContext.cs
+++ b/src/WebJobs.Script/Description/DotNet/FunctionAssemblyLoadContext.cs
@@ -213,7 +213,7 @@ namespace Microsoft.Azure.WebJobs.Script.Description
             if (TryGetRuntimeAssembly(assemblyName, out ScriptRuntimeAssembly scriptRuntimeAssembly))
             {
                 // There are several possible scenarios:
-                //  1. The assembly was found and the policy evaluator succeeeded.
+                //  1. The assembly was found and the policy evaluator succeeded.
                 //     - Return the assembly.
                 //
                 //  2. The assembly was not found (meaning the policy evaluator wasn't able to run).
--- a/src/WebJobs.Script/Description/DotNet/FunctionAssemblyLoadContext.cs
+++ b/src/WebJobs.Script/Description/DotNet/FunctionAssemblyLoadContext.cs
@@ -224,7 +224,8 @@ namespace Microsoft.Azure.WebJobs.Script.Description
                 //     a. We did not adjust the requested assembly name via the deps file.
                 //        - This means that the DefaultLoadContext is looking for the correct version. Return null and let
                 //          it handle the load attempt.
-                //     b. We adjusted the requested assembly name via the deps file.
+                //     b. We adjusted the requested assembly name via the deps file. If we return null the DefaultLoadContext would attempt to
+                //        load the original assembly version, which may be incorrect if we had to adjust it forward past the runtime's version.
                 //        i.  The adjusted assembly name is higher than the runtime's version.
                 //            - Do not trust the DefaultLoadContext to handle this as it may resolve to the runtime's assembly. Instead,
                 //              call LoadCore() to ensure the adjusted assembly is loaded.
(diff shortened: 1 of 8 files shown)


src/WebJobs.Script/Description/DotNet/FunctionAssemblyLoadContext.cs
```diff
@@ -234,6 +269,12 @@ protected override Assembly Load(AssemblyName assemblyName)
             return LoadCore(assemblyName);
         }
 
+        internal bool IsHigherThan(AssemblyName name1, AssemblyName name2)
```
fabiocav: Why a custom method here when we should be able to rely on the version comparison? Are you explicitly trying to ignore the other elements?

src/WebJobs.Script/Description/DotNet/FunctionAssemblyLoadContext.cs
```diff
@@ -212,16 +212,51 @@ protected override Assembly Load(AssemblyName assemblyName)
...
+                else
+                {
+                    if (assembly == null || !isNameAdjusted)
+                    {
+                        // Scenario 2 and 3a.
+                        return null;
+                    }
 
-                return assembly;
+                    AssemblyName runtimeAssemblyName = assembly.GetName();
```
fabiocav: Assembly name should probably go through the assembly name cache

src/WebJobs.Script/Description/DotNet/FunctionAssemblyLoadContext.cs
```diff
@@ -257,26 +298,31 @@ private bool TryAdjustRuntimeAssemblyFromDepsFile(AssemblyName assemblyName, out
...
 
-        private Assembly LoadRuntimeAssembly(AssemblyName assemblyName, ScriptRuntimeAssembly scriptRuntimeAssembly)
+        /// <summary>
+        /// Loads the runtime's version of this assembly and runs it through the appropriate policy evaluator.
+        /// </summary>
+        /// <param name="assemblyName">The assembly name to load.</param>
+        /// <param name="scriptRuntimeAssembly">The policy evaluator details.</param>
+        /// <param name="runtimeAssembly">The runtime's assembly.</param>
+        /// <returns>true if the policy evaluation succeeded, otherwise, false.</returns>
+        private bool ShouldUseRuntimeAssembly(AssemblyName assemblyName, ScriptRuntimeAssembly scriptRuntimeAssembly, out Assembly runtimeAssembly)
```
fabiocav: nit: Change to `TryLoadRuntimeAssembly`

src/WebJobs.Script/Description/DotNet/FunctionAssemblyLoadContext.cs
brettsam: No... just a brain fart. I think I copied it from our custom logic we have for the policy and went with it.

fabiocav (approved):

PR #7129, opened 2021-02-04
draft change for the hostId on keda issue

### LINKED ISSUE #7128: Timer Trigger trigged on each pod on kubernetes with KEDA.
Found this issue.

https://github.com/kedacore/keda/discussions/1508

#### Repro steps

Create a timer trigger and deploy the kuberenetes with min-replica as 2. then configure `AzureWebJobsStorage` to point a Storage Account. 

```
func init --docker-only --csharp
func new   // chose timer trigger
// config local.settings.json 
func kubernetes deploy --name timger-func --registry tsuyoshiushio --min-replicas 2
```

#### Expected behavior

Only one container runs timer trigger. 

#### Actual behavior

Both containers run timer trigger.

#### Known workarounds

Create a single container. 

#### Related information 

The root cause is, the functions host looks at hostname to create a lease blob.
In case of KEDA, the host name is container name.  We need to create a PR for solving this. 

https://github.com/Azure/azure-functions-host/blob/ecaafa21d276f99ed5bfe105e6798bd82320ba66/src/WebJobs.Script/Host/ScriptHostIdProvider.cs#L63

image

image

TsuyoshiUshio:
I started this PR for discussing the design.  

The issue is, on the KEDA environment, it refers to hostname on a container. That will be a pod name on kubernetes. 
When you scale out, on each pod, the hostid is different for the each pod. 
That cause issue for example timer trigger. (Timer trigger make it singleton even if scale out. It uses [leas blob](https://github.com/Azure/azure-webjobs-sdk-extensions/wiki/TimerTrigger#singleton-locks), it create a different lease lock blob container for each hostid. 

My current design is simply watch the `KUBERNETES_SERVICE_HOST` environment variables, if there is, refer to the ` WEBSITE_SITE_NAME` as the host name.  Customer need to set `WEBSITE_SITE_NAME` as an environment variables. 

However, I'm not sure if this design is harmful or not. For example we can introduce a special Environment Variable to avoid the side effect on existing Kubernetes hosting pod instead of `KUBERNETES_SERVICE_HOST`.
Once design is fixed, I'll write a test and closing this PR. 

resolves https://github.com/Azure/azure-functions-host/issues/7128

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script/Environment/EnvironmentExtensions.cs
+++ b/src/WebJobs.Script/Environment/EnvironmentExtensions.cs
@@ -296,6 +296,16 @@ namespace Microsoft.Azure.WebJobs.Script
         }
 
         /// <summary>
+        /// Gets a value indicating whether the application is running in Kubernetes
+        /// </summary>
+        /// <param name="environment">The environment to verify</param>
+        /// <returns><see cref="true"/> If running in a Kubernetes Azure App Service; otherwise, false.</returns>
+        public static bool IsKubernetesHosting(this IEnvironment environment)
+        {
+            return !string.IsNullOrEmpty(environment.GetEnvironmentVariable(KubernetesServiceHost));
+        }
+
+        /// <summary>
         /// Gets a value that uniquely identifies the site and slot.
         /// </summary>
         public static string GetAzureWebsiteUniqueSlotName(this IEnvironment environment)
@@ -310,6 +320,14 @@ namespace Microsoft.Azure.WebJobs.Script
             }
 
             return name?.ToLowerInvariant();
+        }
+
+        /// <summary>
+        /// Gets a value from AzureWebsiteName environment variables.
+        /// </summary>
+        public static string GetKubernetesHostname(this IEnvironment environment)
+        {
+            return environment.GetEnvironmentVariable(AzureWebsiteName);
         }
 
         /// <summary>
--- a/src/WebJobs.Script/Host/ScriptHostIdProvider.cs
+++ b/src/WebJobs.Script/Host/ScriptHostIdProvider.cs
@@ -52,6 +52,11 @@ namespace Microsoft.Azure.WebJobs.Script
                 string hostName = environment.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebsiteHostName);
                 hostId = hostName?.Replace(".azurewebsites.net", string.Empty);
             }
+            else if (environment.IsKubernetesHosting())
+            {
+                // The hostid is delivered from the AzureWebsiteName.
+                hostId = environment.GetKubernetesHostname() ?? Environment.MachineName;
+            }
             else
             {
                 // When running locally, derive a stable host ID from machine name


src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -312,6 +322,14 @@ public static string GetAzureWebsiteUniqueSlotName(this IEnvironment environment
             return name?.ToLowerInvariant();
         }
 
+        /// <summary>
+        /// Gets a value from AzureWebsiteName environment variables.
+        /// </summary>
+        public static string GetKubernetesHostname(this IEnvironment environment)
+        {
+            return environment.GetEnvironmentVariable(AzureWebsiteName);
```
ahmelsayed, line 330: We don't add this currently in core-tools, right?

TsuyoshiUshio: yeah. we don't have it.  current environment variables when I deploy with `func kubernetes deploy` is 

```
root@timger-func-68bc867dd8-gvhx8:/# env
WEBSITE_SITE_NAME=singletonTimer
KUBERNETES_SERVICE_PORT_HTTPS=443
KUBERNETES_SERVICE_PORT=443
HOSTNAME=timger-func-68bc867dd8-gvhx8
AzureFunctionsJobHost__Logging__Console__IsEnabled=true
ASPNETCORE_URLS=http://+:80
PWD=/
AzureWebJobsScriptRoot=/home/site/wwwroot
FUNCTIONS_WORKER_RUNTIME=dotnet
HOME=/home
KUBERNETES_PORT_443_TCP=tcp://10.0.0.1:443
AzureFunctionsJobHost__functions__0=TimerTrigger
AzureWebJobsStorage=MY_CONNECTION_STRING
TERM=xterm
HOST_VERSION=3.0.15277
SHLVL=1
KUBERNETES_PORT_443_TCP_PROTO=tcp
DOTNET_RUNNING_IN_CONTAINER=true
DOTNET_USE_POLLING_FILE_WATCHER=true
KUBERNETES_PORT_443_TCP_ADDR=10.0.0.1
KUBERNETES_SERVICE_HOST=10.0.0.1
KUBERNETES_PORT=tcp://10.0.0.1:443
KUBERNETES_PORT_443_TCP_PORT=443
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
_=/usr/bin/env
``

fabiocav:
Instead of introducing a new variable, wouldn't using the existing variable to explicitly set the host ID suffice here?

TsuyoshiUshio:
Hi @fabiocav  Thank you for the review. 
`WEBSITE_SITE_NAME` is already exists. https://github.com/Azure/azure-functions-host/blob/75755512135afebcbb6c717b624718f649116b42/src/WebJobs.Script/Environment/EnvironmentSettingNames.cs#L10
and `KUBERNETES_SERVICE_HOST` is identified. https://github.com/Azure/azure-functions-host/blob/75755512135afebcbb6c717b624718f649116b42/src/WebJobs.Script/Environment/EnvironmentSettingNames.cs#L96

The point I thought was,  for the `WEBSITE_SITE_NAME` might be good, however,  `KUBERNETES_SERVICE_HOST` is injected by Kubernetes. So that it could possible the Azure Functions Host change the behavior on k8s after introducing this change. 
If it is a bug and expected behavior is having the same hostId among the same function app in scale (in case of kuberentes the same replica set), we can just use `KUBERNETES_SERVICE_HOST` to introduce this fix.  Am I understand correctly?

balag0:
I think @fabiocav might be referring to AzureFunctionsWebHost__hostid
https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-timer?tabs=csharp#function-apps-sharing-storage

TsuyoshiUshio:
Thanks @balag0  and @fabiocav 
I'm close this PR and issue. It works.

PR #7138, opened 2021-02-06
Fixes slow start of multiple worker processes

### LINKED ISSUE #7136: Worker processes do not start in linear time
The functions host adds worker processes more slowly than intended and as documented. Here are the details provided by @cgillum - 

```
According to this documentation, I should expect to wait 100 seconds before all 10 python worker processes are available.
https://docs.microsoft.com/en-us/azure/azure-functions/functions-app-settings#functions_worker_process_count

What I actually observed is that we take 10 seconds for the second python process to be created and 10 more seconds for each additional python worker process. Here is my Functions log output which illustrates what I mean: 

[2021-02-03T21:33:57.717] Worker process started and initialized.
[2021-02-03T21:34:07.667] Worker process started and initialized.
[2021-02-03T21:34:27.649] Worker process started and initialized.
[2021-02-03T21:34:57.652] Worker process started and initialized.
...
[2021-02-03T21:41:27.695] Worker process started and initialized.

It took 390 seconds before all 10 python worker processes came up.
```

paulbatum:

resolves #7136 - with this fix workers should be started as originally intended and documented

### Pull request checklist

### Additional information

This is the least invasive version of the fix - just stops the debounce calls from being chained. However its not clear to me why this code is using debounce in the first place - the result of the debounce call is never shared, so no invocations will ever be prevented. The code might be clearer with the following implementation, let me know if you'd prefer this.

```csharp
        private void StartWorkerProcesses(int startIndex, Action startAction)
        {
            for (var count = startIndex; count < _maxProcessCount; count++)
            {
                Task.Delay(count * _debounceMilliSeconds).ContinueWith(t => startAction());
            }
        }
```

Other thing to look at is the new unit test - it uses a stopwatch to make sure that the workers start up in a reasonable time. The test might be flaky if it has to run in extremely slow environments. Let me know if you'd like a different approach used here.

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -12,6 +12,7 @@
 - Added direct refereces to [System.IO.Pipes](https://www.nuget.org/packages/System.IO.Pipes/4.3.0)  and [System.Threading.Overlapped](https://www.nuget.org/packages/System.Threading.Overlapped/4.3.0) to ensure System.Data.SqlClient package update does not impact unification 
 - Updated Java Worker Version to [1.8.2-SNAPSHOT](https://github.com/Azure/azure-functions-java-worker/releases/tag/1.8.2-SNAPSHOT)
 - Skip external config validation if RuntimeDrivenScaling is enabled in Elastic Premium sku (#6542)
+- Fixed issue where multiple worker processes are started slower than expected
 
 **Release sprint:** Sprint 89, 90, 91
 [ [bugs](https://github.com/Azure/azure-functions-host/issues?q=is%3Aissue+milestone%3A%22Functions+Sprint+89%22+label%3Abug+is%3Aclosed) | [features](https://github.com/Azure/azure-functions-host/issues?q=is%3Aissue+milestone%3A%22Functions+Sprint+89%22+label%3Afeature+is%3Aclosed) ]
--- a/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
+++ b/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
@@ -146,8 +146,8 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
         {
             for (var count = startIndex; count < _maxProcessCount; count++)
             {
-                startAction = startAction.Debounce(_processStartCancellationToken.Token, count * _debounceMilliSeconds);
-                startAction();
+                var debouncedAction = startAction.Debounce(_processStartCancellationToken.Token, count * _debounceMilliSeconds);
+                debouncedAction();
             }
         }
 
--- a/test/WebJobs.Script.Tests/Workers/Rpc/RpcFunctionInvocationDispatcherTests.cs
+++ b/test/WebJobs.Script.Tests/Workers/Rpc/RpcFunctionInvocationDispatcherTests.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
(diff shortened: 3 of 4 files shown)


release_notes.md
```diff
@@ -12,6 +12,7 @@
 - Added direct refereces to [System.IO.Pipes](https://www.nuget.org/packages/System.IO.Pipes/4.3.0)  and [System.Threading.Overlapped](https://www.nuget.org/packages/System.Threading.Overlapped/4.3.0) to ensure System.Data.SqlClient package update does not impact unification 
 - Updated Java Worker Version to [1.8.2-SNAPSHOT](https://github.com/Azure/azure-functions-java-worker/releases/tag/1.8.2-SNAPSHOT)
 - Skip external config validation if RuntimeDrivenScaling is enabled in Elastic Premium sku (#6542)
+- Fixed issue where multiple worker processes are started slower than expected
```
pragnagopa, line 15: nit: Add link to Issue/PR

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -146,8 +146,8 @@ private void StartWorkerProcesses(int startIndex, Action startAction)
         {
             for (var count = startIndex; count < _maxProcessCount; count++)
             {
-                startAction = startAction.Debounce(_processStartCancellationToken.Token, count * _debounceMilliSeconds);
-                startAction();
+                var debouncedAction = startAction.Debounce(_processStartCancellationToken.Token, count * _debounceMilliSeconds);
```
mathewc, line 149: You're right that Debounce likely shouldn't be used here. Debounce is just for actions that may be called very rapidly that we want to wait for a quiet period to execute. I agree with you that we should remove it and use a simpler approach.

What is the actual intent here? Is it to simply stagger the calls but still allow them to overlap? Or are you trying to also prevent overlap, so we only start initializing a new one after awaiting the last one? If the latter, then you could have something like:

```csharp
private void StartWorkerProcesses(int count, Func<Task> startAction)
{
    Task.Run(async () =>
    {
        for (int i = 0; i < count; i++)
        {
            await Task.Delay((i + 1) * 10000);
            await startAction();
        }
    });
}
```

mathewc: Another oddity with the 2 actions passed into this start method is that they're declared `async void`. These aren't event handlers, so violate the [guidelines](https://docs.microsoft.com/en-us/archive/msdn-magazine/2013/march/async-await-best-practices-in-asynchronous-programming#avoid-async-void). They should return tasks, and callers can handle the Tasks as needed, with appropriate error handling, etc.

test/WebJobs.Script.Tests/Workers/Rpc/RpcFunctionInvocationDispatcherTests.cs
```diff
@@ -44,6 +45,40 @@ public async Task GetWorkerStatusesAsync_ReturnsExpectedResult()
...
+            stopwatch.Stop();
+
+            // The first worker is created immediately.
+            // So if we are waiting for 3 more workers to spin up and startup time is 2 seconds, then we expect it to take at least 3*2=6 seconds.
+            // If we waited 6+2=8 seconds and the workers are still not started, then we fail.
+            TimeSpan actual = stopwatch.Elapsed;
+            TimeSpan minimum = (expectedProcessCount - 1) * startupInterval;
+            TimeSpan maximum = minimum + startupInterval;
+
+            Assert.True(actual >= minimum, $"Workers should have taken at least {minimum} to start but took {actual}");
```
mathewc, line 78: I share your concern that this may end up being flaky. You might consider instead just unit testing the method you're modifying. You could pass in a dummy action, capture the invocation times and do a simple check on all the start times to verify expected delays/staggering (deltas >= delay time). That would remove a lot of the variability.

test/WebJobs.Script.Tests/Workers/Rpc/RpcFunctionInvocationDispatcherTests.cs
```diff
@@ -44,6 +45,40 @@ public async Task GetWorkerStatusesAsync_ReturnsExpectedResult()
             }
         }
 
+        [Fact]
+        public async Task WorkersStartInLinearTime()
+        {
```
pragnagopa, line 50: Nit: Instead of adding a new test - update `Starting_MultipleJobhostChannels_Succeeds` to verify the time?

pragnagopa (commented):
Added minor nits

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
pragnagopa: Yes, we should simplify the code here.  Original intent was simply to stagger the calls. But an improvement would be to wait for the process startup and initialization which is handled here `InitializeJobhostLanguageWorkerChannelAsync` to complete before starting a new process

fabiocav: Agree with Mathew's comment above. `Debounce` is being misused here and we should use a different mechanism.

test/WebJobs.Script.Tests/Workers/Rpc/RpcFunctionInvocationDispatcherTests.cs
fabiocav: Is that test validating other things as well? It's nice to have specialized tests if that is that case, as it is a lot easier to troubleshoot the issue when there's a failure. Maintenance is simpler as well.

test/WebJobs.Script.Tests/Workers/Rpc/RpcFunctionInvocationDispatcherTests.cs
fabiocav: Agreed. These kinds of tests are notoriously flaky and sensitive to environment differences.

test/WebJobs.Script.Tests/Workers/Rpc/RpcFunctionInvocationDispatcherTests.cs
```diff
@@ -44,6 +45,40 @@ public async Task GetWorkerStatusesAsync_ReturnsExpectedResult()
...
+            TimeSpan startupInterval = TimeSpan.FromSeconds(2);
+            int expectedProcessCount = 4;
+            RpcFunctionInvocationDispatcher functionDispatcher = GetTestFunctionDispatcher(expectedProcessCount, processStartupInterval: startupInterval);
+
+            Stopwatch stopwatch = Stopwatch.StartNew();
+            await functionDispatcher.InitializeAsync(GetTestFunctionsList(RpcWorkerConstants.NodeLanguageWorkerName));
+
+            await TestHelpers.Await(() =>
+            {
+                var currentChannelCount = functionDispatcher.JobHostLanguageWorkerChannelManager.GetChannels().Count();
```
fabiocav, line 60: nit: Seeing this in a few different spots. Can you please update to the actual type instead of `var` on method return assignments?

paulbatum:
Thanks for the feedback all. Sounds like there is an appetite for a larger overhaul, instead of just a targeted fix. I'll send an updated PR early next week.

paulbatum:
closing this as @alrod has some larger changes he wants to make in this area

PR #7144, opened 2021-02-10
[pack] Fix functions APIs to return config if no function.json

### LINKED ISSUE #7146: Return proper "admin/functions" response for IFunctionProvider functions
During a call to get function metadata `<app-name>/functions/<function-name>`, the `config` is returned empty if no physical function.json file is present. Source -- https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.WebHost/Extensions/FunctionMetadataExtensions.cs#L137

We need to fix that, and add a fallback logic to get FunctionMetadata from `IFunctionMetadataManager`, if a `function.json` is not present.

In addition, we should also not return `config_href` and `script_root_href` when that's the case. As there is no physcial file there.

This is crucial to support such functions without "function.json" (.NET 5) in Portal, and show proper metadata.

ankitkumarr:

resolves https://github.com/Azure/azure-functions-host/issues/7146

### Pull request checklist

### Additional information

Adding a fallback logic for getting `Config` object as part of FunctionMetadataResponse. In case, a physical file is not present, we will rely on `FunctionMetadata`'s config. This fixes the API for functions that do not have `function.json` such as .NET 5 functions.

Additionally, if ever the function directory and files are not present, we should not send VFS links in the response.

cc: @mathewc as fyi

### DIFF
--- a/src/WebJobs.Script.WebHost/Extensions/FunctionMetadataExtensions.cs
+++ b/src/WebJobs.Script.WebHost/Extensions/FunctionMetadataExtensions.cs
@@ -112,11 +112,11 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Extensions
 
         private static string GetFunctionPathOrNull(string scriptRoot, string functionName)
         {
-            var possiblePath = Path.Combine(scriptRoot, functionName);
+            var functionPath = Path.Combine(scriptRoot, functionName);
 
-            if (FileUtility.DirectoryExists(possiblePath))
+            if (FileUtility.DirectoryExists(functionPath))
             {
-                return possiblePath;
+                return functionPath;
             }
 
             return null;
@@ -126,11 +126,11 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Extensions
         {
             if (!string.IsNullOrEmpty(functionPath))
             {
-                var possiblePath = Path.Combine(functionPath, ScriptConstants.FunctionMetadataFileName);
+                var metadataPath = Path.Combine(functionPath, ScriptConstants.FunctionMetadataFileName);
 
-                if (FileUtility.FileExists(possiblePath))
+                if (FileUtility.FileExists(metadataPath))
                 {
-                    return possiblePath;
+                    return metadataPath;
                 }
             }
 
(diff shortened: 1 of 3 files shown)


src/WebJobs.Script.WebHost/Extensions/FunctionMetadataExtensions.cs
```diff
@@ -45,6 +44,16 @@ public static async Task<FunctionMetadataResponse> ToFunctionMetadataResponse(th
...
             };
 
+            if (!string.IsNullOrEmpty(functionPath))
+            {
+                response.ScriptRootPathHref = VirtualFileSystem.FilePathToVfsUri(functionPath, baseUrl, hostOptions, isDirectory: true);
+            }
+
+            if (!string.IsNullOrEmpty(functionMetadataFilePath))
+            {
+                response.ConfigHref = VirtualFileSystem.FilePathToVfsUri(functionMetadataFilePath, baseUrl, hostOptions);
```
brettsam, line 54: What are these hrefs used for? Are they okay to be null?

fabiocav: Unless something changed since the initial design, those should be OK as null

src/WebJobs.Script.WebHost/Extensions/FunctionMetadataExtensions.cs
```diff
@@ -112,35 +104,50 @@ private static async Task<JObject> GetRegularFunctionTrigger(FunctionMetadata fu
...
+
+        private static string GetFunctionPathOrNull(string scriptRoot, string functionName)
         {
-            if (functionMetadata.Bindings == null)
+            var possiblePath = Path.Combine(scriptRoot, functionName);
+
+            if (FileUtility.DirectoryExists(possiblePath))
             {
-                return null;
+                return possiblePath;
```
fabiocav: nit: `functionPath` here would be fine.

src/WebJobs.Script.WebHost/Extensions/FunctionMetadataExtensions.cs
```diff
@@ -112,35 +104,50 @@ private static async Task<JObject> GetRegularFunctionTrigger(FunctionMetadata fu
...
+        }
+
+        private static string GetMetadataPathOrNull(string functionPath)
+        {
+            if (!string.IsNullOrEmpty(functionPath))
             {
-                JObject rawBinding = binding.Raw;
-                var type = (string)rawBinding["type"];
-                if (type != null && type.EndsWith("Trigger", StringComparison.OrdinalIgnoreCase))
+                var possiblePath = Path.Combine(functionPath, ScriptConstants.FunctionMetadataFileName);
```
fabiocav: same as the above about the naming.

src/WebJobs.Script.WebHost/Extensions/FunctionMetadataExtensions.cs
ankitkumarr: These APIs are used by Portal to get the files for "Code + Test". Right now, it generates a VFS link that's invalid and throws error when called. So, I am making these nulls and working with Portal so they can handle the values.

I will confirm with Portal before merging, but this shouldn't have a negative impact.

fabiocav (approved):

PR #7148, opened 2021-02-10
[pack] Fix keys APIs for functions without function.json

### LINKED ISSUE #7147: Fix keys APIs for IFunctionProvider functions (without "function.json")
The `/keys` endpoints do not seem to work for functions without "function.json". This is because `KeysController` relies on physical `function.json` to figure out if function name is accurate or not. We need to use `IFunctionMetadataManager` for such operations.

ankitkumarr:

resolves https://github.com/Azure/azure-functions-host/issues/7147

### Pull request checklist

### Additional information

The PR changes the code to rely on `IFunctionMetadataManager` as the source of truth for functions. The fallback logic is to check in the filesystem to avoid any edge cases for backward compatibility.

### DIFF
--- a/src/WebJobs.Script.WebHost/Controllers/KeysController.cs
+++ b/src/WebJobs.Script.WebHost/Controllers/KeysController.cs
@@ -31,15 +31,17 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Controllers
         private readonly IOptions<ScriptApplicationHostOptions> _applicationOptions;
         private readonly IFileSystem _fileSystem;
         private readonly IFunctionsSyncManager _functionsSyncManager;
+        private readonly IFunctionMetadataManager _functionMetadataManager;
         private readonly IScriptHostManager _hostManager;
 
-        public KeysController(IOptions<ScriptApplicationHostOptions> applicationOptions, ISecretManagerProvider secretManagerProvider, ILoggerFactory loggerFactory, IFileSystem fileSystem, IFunctionsSyncManager functionsSyncManager, IScriptHostManager hostManager)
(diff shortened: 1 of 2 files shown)


brettsam (dismissed):

src/WebJobs.Script.WebHost/Controllers/KeysController.cs
```diff
@@ -285,9 +287,13 @@ internal bool IsBuiltInSystemKeyName(string keyName)
 
         private bool IsFunction(string functionName)
         {
-            string json = null;
+            if (_functionMetadataManager.TryGetFunctionMetadata(functionName, out _))
+            {
+                return true;
+            }
```
brettsam, line 294: I'm wondering if you should flip this to have less impact. Try to read the FunctionConfig first, and then, if false, try this as a fallback? 

I don't know what could break from this, but it's definitely a behavior change like this if the function.json is present.

fabiocav: Agreed. The previous behavior will actually attempt to parse the file. Placing this here does have a behavior impact.

mathewc: Admin APIs like this have to work even if the host isn't running (i.e. keys/functions APIs aren't marked RequiresRunningHost). Doesn't the metadata manager require host startup to initialize codeless extension providers? Or is it all independent of the host running. Should verify this.

ankitkumarr: Yeah I will flip this. @mathewc, metadata manager doesn't rely on Host running. When Host starts, it reinitializes the list of functions it has (including codeless ones). But it shouldn't block on anything like that.

mathewc: Ok

fabiocav (approved):

PR #7151, opened 2021-02-10
[pack] Updated Webjobs SDKs

azfuncgh:


yojagad:
@fabiocav Just checking.. Is there something else that needs to be done here ?

mathewc:
@yojagad @fabiocav I had proposed that we take a new version of the SDK + AI logger packages, since the version we're taking here have a bug which is addressed by https://github.com/Azure/azure-webjobs-sdk/pull/2674.

yojagad:
If someone can publish a new version of webjobs sdk, other packages, I can update this PR with the same and then we can go from there?

mathewc:
New packages published. You're good to go with this PR now.

yojagad:
Thanks @fabiocav !!

yojagad (approved):

fabiocav (approved):

PR #7198, opened 2021-03-02
Adding Functions Host Storage to use WebJobs.Host.Storage with Track2…

karshinlin:

Migrates Functions Host to use Track2 storage SDKs through WebJobs.Host.Storage package

### Pull request checklist

### Additional information

Additional PR information


src/WebJobs.Script.WebHost/WebJobs.Script.WebHost.csproj
```diff
@@ -60,15 +60,17 @@
...
     <PackageReference Include="DotNetTI.BreakingChangeAnalysis" Version="1.1.0-preview" />
     <PackageReference Include="Microsoft.AspNetCore.Authentication.JwtBearer" Version="3.1.0" />
     <PackageReference Include="Microsoft.AspNetCore.Mvc.NewtonsoftJson" Version="3.1.0" />
     <PackageReference Include="Microsoft.Azure.AppService.Middleware.Functions" Version="1.4.0" />
     <PackageReference Include="Microsoft.Azure.AppService.Proxy.Client" Version="2.0.11020001-fabe022e" />
     <PackageReference Include="Microsoft.Azure.Services.AppAuthentication" Version="1.0.3" />
     <PackageReference Include="Microsoft.Azure.Storage.File" Version="11.1.7" />
-    <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.27" />
-    <PackageReference Include="Microsoft.Azure.WebJobs.Host.Storage" Version="4.0.1" />
+    <PackageReference Include="Microsoft.Azure.WebJobs" Version="5.0.0" />
```
fabiocav: Submitting this comment only as this is a foundational thing we need to follow up on before we can look at the other changes. We can't have a major version change in the host without impacting compatibility. I'll reach out so we can discuss offline, as I think this is being impacted by the layering upstream.

fabiocav (changes requested):

src/WebJobs.Script.WebHost/BreakingChangeAnalysis/BlobChangeAnalysisStateProvider.cs
```diff
@@ -19,63 +21,84 @@ internal sealed class BlobChangeAnalysisStateProvider : IChangeAnalysisStateProv
...
+            if (handle != null && handle is BlobClient blobClient)
             {
-                if (!await blob.ExistsAsync())
+                try
                 {
-                    await blob.UploadTextAsync(string.Empty);
-                }
+                    if (!await blobClient.ExistsAsync())
+                    {
+                        using (Stream stream = new MemoryStream(Encoding.UTF8.GetBytes(string.Empty)))
```
pakrym:
```suggestion
                        using (Stream stream = new MemoryStream())
```

src/WebJobs.Script.WebHost/Security/KeyManagement/BlobStorageSecretsRepository.cs
```diff
@@ -65,15 +68,17 @@ public override bool IsEncryptionSupported
             }
         }
 
-        protected virtual CloudBlobContainer CreateBlobContainer(string connectionString)
+        protected virtual BlobContainerClient CreateBlobContainerClient(string connectionString)
```
pakrym: The parameter name is the connection string and it seemed to have been used as a connection string in a previous version of the code. Are we sure we are doing the right thing here?

src/WebJobs.Script.WebHost/Security/KeyManagement/BlobStorageSecretsRepository.cs
```diff
@@ -151,17 +160,23 @@ public override async Task<string[]> GetSecretSnapshots(ScriptSecretsType type,
...
 
-            BlobResultSegment segmentResult;
+            var blobList = new List<BlobItem>();
             try
             {
-                segmentResult = await Container.ListBlobsSegmentedAsync(string.Format("{0}/{1}", _secretsBlobPath, prefix.ToLowerInvariant()), null);
+                var segmentResult = Container.GetBlobsAsync(prefix: string.Format("{0}/{1}", _secretsBlobPath, prefix.ToLowerInvariant()));
+                var asyncEnumerator = segmentResult.GetAsyncEnumerator();
+
+                while (await asyncEnumerator.MoveNextAsync())
```
pakrym: await foreach?

src/WebJobs.Script.WebHost/Security/KeyManagement/BlobStorageSecretsRepository.cs
```diff
@@ -151,17 +160,23 @@ public override async Task<string[]> GetSecretSnapshots(ScriptSecretsType type,
...
+                    blobList.Add(asyncEnumerator.Current);
+                }
             }
             catch (Exception e)
             {
                 LogErrorMessage("list");
                 throw e;
             }
-            return segmentResult.Results.Select(x => x.Uri.ToString()).ToArray();
+            return blobList.Select(x => x.Name).ToArray();
```
pakrym: Can we add names to the list in the first place?

src/WebJobs.Script.WebHost/WebJobs.Script.WebHost.csproj
```diff
@@ -60,15 +60,17 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Azure.Core" Version="1.9.0" />
```
pakrym: You shouldn't need a direct core dependency, storage would bring one for you. You shouldn't need Azure.Storage.Blobs, as well, it would be provided by Microsoft.Azure.WebJobs.Host.Storage

src/WebJobs.Script.WebHost/Security/KeyManagement/BlobStorageSecretsRepository.cs
karshinlin: Yes. BlobStorageSecretsRepository is called in 2 places:
- BlobStorageSasSecretsRepository inherits from it, but overrides the CreateBlobContainerClient() method to create a container from a SasUri
- DefaultSecretManagerProvider creates a BlobStorageSecretsRespository object and now passes in a Connection name instead of a connection string.

src/WebJobs.Script.WebHost/Security/KeyManagement/BlobStorageSecretsRepository.cs
karshinlin: Ideally, we would use await foreach here but it's not supported in the C# version that functions host targets (async streams added in C# 8.0). 
https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/storage/Azure.Storage.Blobs/AzureStorageNetMigrationV12.md#flat-listing

pakrym: You can set the lang version in the project file (https://www.meziantou.net/using-foreach-with-ienumerator-t-iasyncenumerator-t-in-csharp-9.htm) that's what we do for our libraries even though we target .net standard

src/WebJobs.Script.WebHost/Security/KeyManagement/BlobStorageSecretsRepository.cs
karshinlin: The functionality of the secrets manager relies on the actual blob name. The path and extension are stripped out when using this return value of blob names. In the ReadAsync call, the path is built again. Changing from uri -> blobName shouldn't cause problems here.

pakrym: Sorry, I need to work on the clarity of my comments. What I was saying is that we can change the list to be a list of string and do ` blobList.Add(asyncEnumerator.Current.Name);`. Instead of transforming it before returning.

karshinlin: Ah, good call. Will do.

src/WebJobs.Script/Host/DistributedLockManagers/BlobLeaseDistributedLockManager.cs
```diff
@@ -0,0 +1,385 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Script
+{
+    /// <summary>
+    /// Provides a CloudBlob lease-based implementation of the <see cref="IDistributedLockManager"/> service for singleton locking.
+    /// This class can be overridden to control where the container comes from.
+    /// The default derived implementation is <see cref="CloudBlobContainerDistributedLockManager"/> which is container based.
+    /// Hosts can provide a derived implementation to leverage the accountName and allow different hosts to share.
+    /// </summary>
```
paulbatum, line 25: A comment highlighting that this code has been moved from Host.Storage in the SDK repo would be useful (anyone trying to find history for this file will hit a dead-end if they don't realize this)

src/WebJobs.Script/StorageProvider/AzureStorageProvider.cs
```diff
@@ -0,0 +1,93 @@
...
+        /// <returns>successful client creation</returns>
+        public virtual bool TryGetBlobServiceClientFromConnectionString(out BlobServiceClient client, string connectionString = null)
+        {
+            try
+            {
+                var connectionStringToUse = connectionString ?? _configuration.GetWebJobsConnectionString(ConnectionStringNames.Storage);
+                _blobServiceClientProvider.TryGetFromConnectionString(connectionString, out client);
+                return true;
+            }
+            catch
```
paulbatum: This code confuses me a little - you're saying TryGetFromConnectionString can throw exceptions because parsing the connection string fails? I would expect it to return false - otherwise it doesn't behave how we'd typically expect for a "TryFoo" method. Any reason you're not checking the return value? Can the underlying implementation use StorageConnectionString.TryParse?

src/WebJobs.Script/StorageProvider/AzureStorageProvider.cs
```diff
@@ -0,0 +1,93 @@
...
+        /// <param name="connection">Name of the connection to use</param>
+        /// <returns>successful client creation</returns>
+        public virtual bool TryGetBlobServiceClientFromConnection(out BlobServiceClient client, string connection = null)
+        {
+            try
+            {
+                client = _blobServiceClientProvider.Get(connection);
+                return true;
+            }
+            catch
```
paulbatum: similar to above, why not _blobServiceClientProvider.TryGet(..)?

src/WebJobs.Script/StorageProvider/IAzureStorageProvider.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Azure.Storage.Blobs;
+
+namespace Microsoft.Azure.WebJobs.Script
+{
+    public interface IAzureStorageProvider
+    {
+        bool TryGetBlobServiceClientFromConnectionString(out BlobServiceClient client, string connectionString);
```
paulbatum, line 14: The fact that both of these methods take a string param while the only difference in the method name is "string" makes potential for subtle bugs where the wrong method is called. Did you consider introducing some sort of wrapper type for the connection scenario, just to help make the semantic difference between these methods more pronounced/clear?

src/WebJobs.Script/Timer/AzureStorageScheduleMonitor.cs
```diff
@@ -0,0 +1,147 @@
...
+using Microsoft.Azure.WebJobs.Logging;
+using Microsoft.Azure.WebJobs.Script;
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Timers
+{
+    /// <summary>
+    /// <see cref="ScheduleMonitor"/> that stores schedule information in blob storage.
+    /// </summary>
```
paulbatum, line 23: This is another move, this time from the timer extension right? Comment pointing to the original file would help.

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -296,12 +300,49 @@ public static IHostBuilder AddScriptHostCore(this IHostBuilder builder, ScriptAp
...
+                {
+                    return new BlobLeaseDistributedLockManager(loggerFactory, azureStorageProvider);
+                }
+            }
+            catch (Exception ex)
+            {
+                var str = ex.ToString();
+            }
+
+            return new InMemoryDistributedLockManager();
```
paulbatum: Is this fallback to the inmemory manager just for testing? Is it possible we'll hit this codepath in prod, and the fact that its silently falling back will be confusing? Should we emit a log entry about the fallback?

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -296,12 +300,49 @@ public static IHostBuilder AddScriptHostCore(this IHostBuilder builder, ScriptAp
...
+                var container = azureStorageProvider.GetBlobContainerClient();
+
+                if (container != null)
+                {
+                    return new BlobLeaseDistributedLockManager(loggerFactory, azureStorageProvider);
+                }
+            }
+            catch (Exception ex)
+            {
+                var str = ex.ToString();
```
paulbatum: I don't understand this exception handler, it looks like it will swallow anything and no logs will be emitted

src/WebJobs.Script.WebHost/BreakingChangeAnalysis/BlobChangeAnalysisStateProvider.cs
```diff
@@ -19,63 +20,84 @@ internal sealed class BlobChangeAnalysisStateProvider : IChangeAnalysisStateProv
         private readonly ILogger _logger;
         private readonly IConfiguration _configuration;
         private readonly IHostIdProvider _hostIdProvider;
+        private readonly IAzureStorageProvider _azureStorageProvider;
 
-        public BlobChangeAnalysisStateProvider(IConfiguration configuration, IHostIdProvider hostIdProvider, ILogger<BlobChangeAnalysisStateProvider> logger)
+        public BlobChangeAnalysisStateProvider(IConfiguration configuration, IHostIdProvider hostIdProvider, ILogger<BlobChangeAnalysisStateProvider> logger, IAzureStorageProvider azureStorageProvider)
```
paulbatum, line 25: @fabiocav Any reason for keeping this around, now that V2->V3 is done?

test/WebJobs.Script.Tests.Integration/Storage/StorageEndToEndTests.cs
```diff
@@ -116,6 +116,26 @@ public async Task CanUseStorageV11Types()
             }, userMessageCallback: _fixture.Host.GetLog);
         }
 
+        [Fact]
+        public async Task CanUseStorageV12Types()
+        {
+            // The point of this test is to verify back compat behavior that a user can still #r and use storage v11 types in CSX,
```
paulbatum: comment needs to be updated to say v12 instead of v11.

test/WebJobs.Script.Tests.Integration/Storage/StorageEndToEndTests.cs
```diff
@@ -116,6 +116,26 @@ public async Task CanUseStorageV11Types()
             }, userMessageCallback: _fixture.Host.GetLog);
         }
 
+        [Fact]
+        public async Task CanUseStorageV12Types()
```
paulbatum: @fabiocav do we want this behavior? I seem to recall that we discussed this a while ago and we didn't want to introduce any net-new storage types that can be used from csx without installing a storage extension... that we just added the V9 and V11 tests because our hands were tied. But I might be missing some key details.

test/WebJobs.Script.Tests.Integration/TestScripts/CSharp/StorageReferenceV12/run.csx
```diff
@@ -0,0 +1,11 @@
+﻿// Use storage v12
+#r "Azure.Storage.Blobs"
+
+using Azure.Storage.Blobs;
+
+public static void Run(string input, TraceWriter log)
+{
+    // it is enough to just reference the type - as long as compilation is successful, we're good
+    BlobClient blobClient;
```
paulbatum, line 9: As per my comment in the EndToEndTests, I'm not sure this is desired behavior.

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
karshinlin: This is intended behavior for cases where Blob storage isn't available (i.e. a placeholder that is warmed up will use this InMemoryDistributedLockManager. Don't think we should log here because it is intended and may just cause more noise.

src/WebJobs.Script/Host/DistributedLockManagers/BlobLeaseDistributedLockManager.cs
karshinlin: Good call. Will put this comment here.

src/WebJobs.Script/StorageProvider/AzureStorageProvider.cs
karshinlin: Ah. I overlooked when I went back and forth deciding where the actual try/catch logic should live. I've updated to just use the logic in StorageClientProvider in webjobs. This makes these methods in AzureStorageProvider a lot simpler.

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
karshinlin: Leftover from debugging :)

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -296,12 +300,48 @@ public static IHostBuilder AddScriptHostCore(this IHostBuilder builder, ScriptAp
...
+            services.AddSingleton<IDistributedLockManager>(provider => AddLockManager(provider));
+        }
+
+        public static IDistributedLockManager AddLockManager(IServiceProvider provider)
+        {
+            try
+            {
+                var azureStorageProvider = provider.GetRequiredService<IAzureStorageProvider>();
+                var loggerFactory = provider.GetRequiredService<ILoggerFactory>();
+                var container = azureStorageProvider.GetBlobContainerClient();
```
mathewc: Not clear to me why we're creating this client here - it's not being used below? It's just a validation check? Looks like internally inside BlobLeaseDistributedLockManager you're doing the same call again. Why not just pass the client in on construction?

karshinlin: This follows the similar logic in WebJobs to check if Blob storage is available before registering an implementation of IDistributedLockManager. Creating the client here gives us this validation to choose BlobLeaseDistributedLockManager; or otherwise just use In-Memory Lock.

It is only 1 additional GetContainerClient() call (which I don't think results in a network call) whenever the ScriptHost is recreated. To cut down on this call, I could save the reference in the GetContainerClient() if you think that'll help.

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
paulbatum: Okay do you know what exceptions to check for here? ArgumentException maybe? If you can scope down the catch block to just handling the expected exception type I'd feel more comfortable with this.

mathewc: Still not clear to me that we need to pull this type into host. SDK will use it's version of InMemoryDistributedLockManager IFF we haven't provided our own implementation, so seems we can just rely on that?

karshinlin: In order to determine if Blob Storage is available, we need access to the IAzureStorageProvider service. When the scriptHost has a parent scope (i.e. WebHost), IAzureStorageProvider is not available yet when these services are registered, so I used the AddSingleton<TService>(this IServiceCollection services, Func<IServiceProvider, TService> implementationFactory) so that the factory can create the implementation at runtime when it is needed. This factory requires an implementation to be returned, so I brought in InMemoryDistributedLockManger. 

Do let me know if there's a way around this.

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
mathewc: Also, the method can be private since only used above?

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -296,12 +300,48 @@ public static IHostBuilder AddScriptHostCore(this IHostBuilder builder, ScriptAp
...
                 services.TryAddSingleton<FunctionsScaleManager>();
+
+                services.AddHostOverrides();
             });
 
             RegisterFileProvisioningService(builder);
             return builder;
         }
 
+        public static void AddHostOverrides(this IServiceCollection services)
```
mathewc: Since this is only needed above and by tests, it could be internal rather than pubic.

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -285,6 +288,7 @@ public static IHostBuilder AddScriptHostCore(this IHostBuilder builder, ScriptAp
                 if (!applicationHostOptions.HasParentScope)
                 {
                     AddCommonServices(services);
+                    services.AddAzureStorageProvider();
```
mathewc, line 291: Previously at this layer we took no dependencies on Azure Storage, the config model being that customers add services/extensions they need on startup.  Functions WebHost calls AddScriptHostCore to register these core services, and is the one that brings in the additional storage dependencies. We should be able to remove this and let WebHost add the storage dependencies?

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -296,12 +300,48 @@ public static IHostBuilder AddScriptHostCore(this IHostBuilder builder, ScriptAp
                     services.TryAddEnumerable(ServiceDescriptor.Singleton<IHostedService, FunctionsScaleMonitorService>());
                 }
                 services.TryAddSingleton<FunctionsScaleManager>();
+
+                services.AddHostOverrides();
```
mathewc: This is adding storage specific implementations so doesn't it rely on the above call to AddAzureStorageProvider happening? That call above is conditional, which means that this could be called without the other required services being registered?

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -210,9 +211,13 @@ internal async Task<string> CheckHashAsync(CloudBlockBlob hashBlob, string conte
 
                 // get the last hash value if present
                 string lastHash = null;
-                if (await hashBlob.ExistsAsync())
+                if (await hashBlobClient.ExistsAsync())
                 {
-                    lastHash = await hashBlob.DownloadTextAsync();
+                    var downloadResponse = await hashBlobClient.DownloadAsync();
+                    using (StreamReader reader = new StreamReader(downloadResponse.Value.Content))
```
mathewc, line 217: I'm not familiar with the new v12 client APIs. Can Value ever be null here?

karshinlin: If there was an error with the download, the DownloadAsync call would throw Azure's RequestFailedException. Here we always expect the response payload to have a Value.

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
karshinlin: We already call AddAzureStorageProvider in the WebHost root builder. The conditional is meant to avoid registering two instances of these services.

test/WebJobs.Script.Tests.Integration/Storage/StorageEndToEndTests.cs
karshinlin: You're right. We'll have to make sure csx can't use these types. I'll keep the test around but make sure these v12 types aren't usable.

src/WebJobs.Script/StorageProvider/IAzureStorageProvider.cs
karshinlin: Personally I think the verboseness of the method is enough to distinguish between using a connection vs. connection string. Adding a wrapper to hold a single property may be adding too much complexity. Current tests that rely on AzureStorageProvider would fail if one is used over the other. 
In addition, I have the TryGetBlobServiceClientFromConnectionString as a convenience method that isn't used anywhere right now (since we implemented this secretless approach).

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
```diff
@@ -179,6 +179,9 @@ public static void AddWebJobsScriptHost(this IServiceCollection services, IConfi
 
             services.TryAddSingleton<IDependencyValidator, DependencyValidator>();
             services.TryAddSingleton<IJobHostMiddlewarePipeline>(s => DefaultMiddlewarePipeline.Empty);
+
+            // Add AzureStorageProvider to WebHost (also needed for ScriptHost)
+            services.AddAzureStorageProvider();
```
karshinlin, line 184: Let's review this service registration to make sure the behavior is as intended. With this work, all blob operations go through IAzureStorageProvider, which internally uses the IConfiguration. We have two scopes (Root WebHost and ScriptHost) at play here and the used IConfiguration object in IAzureStorageProvider depends on where we register this service. 
ScriptHost adds more Configuration builders (i.e. HostJson) - will IAzureStorageProvider need a dependency on these extra configurations? Right now, it's enough to register IAzureStorageProvider at WebHost layer since Environment Variables are included.

paulbatum (dismissed):
Latest changes address my major points of feedback. Please get one more approver before merging.

src/WebJobs.Script.WebHost/BreakingChangeAnalysis/BlobChangeAnalysisStateProvider.cs
```diff
@@ -19,63 +20,84 @@ internal sealed class BlobChangeAnalysisStateProvider : IChangeAnalysisStateProv
...
+            }
 
-            return new ChangeAnalysisState(lastAnalysisTime, reference);
+            return new ChangeAnalysisState(lastAnalysisTime, blobClient);
         }
 
         public async Task SetTimestampAsync(DateTimeOffset timestamp, object handle, CancellationToken cancellationToken)
         {
-            if (handle is CloudBlockBlob blob)
+            if (handle != null && handle is BlobClient blobClient)
```
pakrym:
```suggestion
            if (handle is BlobClient blobClient)
```

pakrym: `is` handles the `null` check

src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
```diff
@@ -154,6 +155,35 @@ private static void ConfigureRegisteredBuilders<TBuilder>(TBuilder builder, ISer
...
+            try
+            {
+                IServiceProvider provider;
+                if (applicationHostOptions.HasParentScope)
+                {
+                    provider = applicationHostOptions.RootServiceProvider;
+                }
+                else
+                {
+                    provider = services.BuildServiceProvider();
```
pakrym: I might be missing some context here but building a temporary ServiceProvider is not a good idea in 98.5% of cases I've seen. Can we loop over the `services` and see if there are any `IAzureStorageProvider` registrations?

src/WebJobs.Script/Host/DistributedLockManagers/BlobLeaseDistributedLockManager.cs
```diff
@@ -0,0 +1,383 @@
...
+using Azure.Storage.Blobs.Models;
+using Azure.Storage.Blobs.Specialized;
+using Microsoft.Azure.WebJobs.Host;
+using Microsoft.Azure.WebJobs.Logging;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Script
+{
+    /// <summary>
+    /// This is an overriden implementation based off the StorageBaseDistributedLockManager in Microsoft.Azure.WebJobs.Host.Storage package.
```
mathewc, line 23: In a remarks section here, can we label this TEMP to make it clear that this is a stop-gap solution until we release the actual storage v12 version of Microsoft.Azure.WebJobs.Host.Storage, when this will go away?

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -414,6 +419,19 @@ internal static ExtensionBundleOptions GetExtensionBundleOptions(IConfiguration
             return options;
         }
 
+        internal static void AddHostTimerOverride(this IServiceCollection services)
```
mathewc: nit: AddTimerScheduleMonitorOverride?

src/WebJobs.Script/Timer/AzureStorageScheduleMonitor.cs
```diff
@@ -0,0 +1,148 @@
...
+using Microsoft.Azure.WebJobs.Host.Executors;
+using Microsoft.Azure.WebJobs.Logging;
+using Microsoft.Azure.WebJobs.Script;
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Timers
+{
+    /// <summary>
+    /// This is an overriden implementation based off the StorageScheduleMonitor in Microsoft.Azure.WebJobs.Extensions package.
```
mathewc, line 21: In a remarks section here, can we label this TEMP to make it clear that this is a stop-gap solution until we release the actual storage v12 version of Microsoft.Azure.WebJobs.Host.Storage, when this will go away?

Same comment for any other types we've copied in temporarily. I've created a tracking work item https://github.com/Azure/azure-webjobs-sdk/issues/2710 that we can point these TEMP comments to.

test/WebJobs.Script.Tests.Integration/Host/SingletonTests/SingletonEndToEndTests.cs
```diff
@@ -0,0 +1,906 @@
...
+using Microsoft.Azure.Storage.Blob;
+using Microsoft.Azure.Storage.Queue;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using Xunit;
+using Microsoft.Azure.WebJobs.Host;
+
+namespace Microsoft.Azure.WebJobs.Script.Tests.Integration.Host.SingletonTests
+{
+    // Borrowed from WebJobs.Host.EndToEndTests
```
mathewc: Same comment for all the tests that were copied in - mark them TEMP and link to the issue

src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
karshinlin: So we need the actual service IAzureStorageProvider instantiated in order to check if Blob storage is available. We always register IAzureStorageProvider regardless, but here we don't want to add the BlobLeaseDistributedLockManager if blob storage isn't available (InMemoryDistributedLockManager already registered by webjobs).

Do you know what the concern is with building this ServiceProvider (i.e. any known issues)?

test/WebJobs.Script.Tests.Integration/Host/SingletonTests/SingletonEndToEndTests.cs
karshinlin: Added "TODO: TEMP - implementation should be moved" to all these temp files.

src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
pakrym: >Do you know what the concern is with building this ServiceProvider (i.e. any known issues)?

Yeah, for one it's pretty expensive and might slow down the startup. The other is around resource management, the created service provider needs to be disposed.

src/WebJobs.Script.WebHost/BreakingChangeAnalysis/BlobChangeAnalysisStateProvider.cs
fabiocav: We can remove this (as discussed). Since the work is already done, let's keep it so we don't increase the scope.

src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
fabiocav: Agree. Provided some feedback in person here.

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -187,6 +189,8 @@ public static IHostBuilder AddScriptHostCore(this IHostBuilder builder, ScriptAp
                 .AddManualTrigger()
                 .AddWarmup();
 
+                webJobsBuilder.Services.AddTimerScheduleMonitorOverride();
```
fabiocav: As discussed, this method needs a better name that follows the convention.

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -440,6 +461,33 @@ private static void AddProcessRegistry(IServiceCollection services)
...
         }
 
+        private static IDistributedLockManager GetBlobLockManagerOverride(IServiceProvider provider, ScriptApplicationHostOptions applicationHostOptions)
+        {
+            try
+            {
+                //if (applicationHostOptions.HasParentScope)
+                //{
+                //    provider = applicationHostOptions.RootServiceProvider;
+                //}
```
fabiocav: Remove comment

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -440,6 +461,33 @@ private static void AddProcessRegistry(IServiceCollection services)
...
+
+                var azureStorageProvider = provider.GetRequiredService<IAzureStorageProvider>();
+                var container = azureStorageProvider.GetBlobContainerClient();
+
+                if (container != null)
+                {
+                    var loggerFactory = provider.GetRequiredService<ILoggerFactory>();
+                    return new BlobLeaseDistributedLockManager(loggerFactory, azureStorageProvider);
+                }
+            }
```
fabiocav, line 472: Are you expecting exceptions in this case? We shouldn't be swallowing it here. Are the operations being introduced above expensive?

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -440,6 +461,33 @@ private static void AddProcessRegistry(IServiceCollection services)
             }
         }
 
+        private static IDistributedLockManager GetBlobLockManagerOverride(IServiceProvider provider, ScriptApplicationHostOptions applicationHostOptions)
```
fabiocav: You should be able to inline this method above. Otherwise; rename to something like `GetBlobLockManager`

src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
```diff
@@ -52,8 +53,6 @@ public static class WebScriptHostBuilderExtension
                 })
                 .AddScriptHost(webHostOptions, configLoggerFactory, metricsLogger, webJobsBuilder =>
                 {
-                    webJobsBuilder.AddAzureStorageCoreServices();
```
fabiocav, line 55: Is the scope of this registration changing? Just want to make sure we validate the other services without this registration here. Was it moved elsewhere?

fabiocav (dismissed):
Approving to initiate the build, but we may need to follow up on comments prior to release.

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
karshinlin: This logic could throw an InvalidOperationException (if no valid connection), UriFormatException (if the sas container override is not properly formatted), or ArgumentException (validation logic in creating a BlobContainerClient at the sdk level). Thought the default behavior should be to register an InMemoryDistributedLockManager in this case.

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
karshinlin: Renamed in preference of the method over inline block.

src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
karshinlin: Didn't come across anything in my manual validation but yes, we are changing the scope. When there is a WebHost, it is registered in this parent - if not, registered in the ScriptHost container.

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
fabiocav: All of the above indicate an improperly configured app and this behavior will hide the issue from the user, applying a fallback that would likely confuse them and potentially cause the configuration issue to go unnoticed. Can you please update this so we're not hiding the problem and swallowing the exception?

src/WebJobs.Script/StorageProvider/AzureStorageProvider.cs
```diff
@@ -61,7 +64,9 @@ public virtual BlobContainerClient GetBlobContainerClient()
 
             if (!TryGetBlobServiceClientFromConnection(out BlobServiceClient blobServiceClient, ConnectionStringNames.Storage))
             {
-                throw new InvalidOperationException($"Could not create BlobServiceClient to obtain the BlobContainerClient using Connection: {ConnectionStringNames.Storage}");
+                var ex = new InvalidOperationException($"Could not create BlobServiceClient to obtain the BlobContainerClient using Connection: {ConnectionStringNames.Storage}");
+                _logger.LogError(ex, "Invalid blob storage connection configuration. Could not create BlobServiceClient.");
+                throw ex;
```
fabiocav, line 69: Change `throw ex;` -> `throw;`

PR #7208, opened 2021-03-04
Remove hardcoded placeholder mode for Linux consumption containers

### LINKED ISSUE #7215: Pinned linux consumption containers should not start in placeholder mode.
Currently we hardcode linux consumption containers to always start in placeholder mode. Antares now dynamically sets the placeholder mode if the container is part of the placeholder pool. So we can remove the hardcoded placeholder mode.

Specialization logic also has hardcoded assumption that the container will only be specialized in placeholder mode. This part needs to be updated as well.

divyagandhisethi:
Fixes #7215 

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.WebHost/Management/InstanceManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/InstanceManager.cs
@@ -103,7 +103,11 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
             if (!_webHostEnvironment.InStandbyMode)
             {
                 // This is only true when specializing pinned containers.
-                _logger.LogInformation("Assign called while host is not in placeholder mode.");
+                if (!context.Environment.TryGetValue(EnvironmentSettingNames.ContainerStartContext, out string startContext))
+                {
+                    _logger.LogError("Assign called while host is not in placeholder mode and start context is not present.");
+                    return false;
+                }
             }
 
             if (_environment.IsContainerReady())
--- a/src/WebJobs.Script.WebHost/Management/InstanceManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/InstanceManager.cs
@@ -102,7 +102,13 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
         {
             if (!_webHostEnvironment.InStandbyMode)
             {
-                _logger.LogError("Assign called while host is not in placeholder mode");
+                // This is only true when specializing pinned containers.
+                _logger.LogInformation("Assign called while host is not in placeholder mode.");
+            }
+
+            if (_environment.IsContainerReady())
+            {
+                _logger.LogError("Assign called while container is marked as specialized.");
                 return false;
             }
 
--- a/src/WebJobs.Script.WebHost/Program.cs
+++ b/src/WebJobs.Script.WebHost/Program.cs
@@ -87,8 +87,6 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
         {
             if (SystemEnvironment.Instance.IsLinuxConsumption())
             {
-                // Linux containers always start out in placeholder mode
-                SystemEnvironment.Instance.SetEnvironmentVariable(EnvironmentSettingNames.AzureWebsitePlaceholderMode, "1");
                 AppDomain.CurrentDomain.UnhandledException += CurrentDomainOnUnhandledExceptionInLinuxConsumption;
             }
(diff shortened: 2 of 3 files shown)


src/WebJobs.Script.WebHost/Management/InstanceManager.cs
```diff
@@ -102,7 +102,13 @@ public bool StartAssignment(HostAssignmentContext context)
         {
             if (!_webHostEnvironment.InStandbyMode)
             {
-                _logger.LogError("Assign called while host is not in placeholder mode");
+                // This is only true when specializing pinned containers.
+                _logger.LogInformation("Assign called while host is not in placeholder mode.");
```
pragnagopa: Change to LogDebug - Debug is the default log level for Kusto

pragnagopa (commented):
Please open an issue and link to the PR. Also, please add details in the issue how this change will not impact current behavior as platform will set EnvironmentSettingNames.AzureWebsitePlaceholderMode to 1

src/WebJobs.Script.WebHost/Management/InstanceManager.cs
```diff
@@ -102,7 +102,13 @@ public bool StartAssignment(HostAssignmentContext context)
...
+            }
+
+            if (_environment.IsContainerReady())
```
balag0, line 113: Add && !string.IsNullOrEmpty(CONTAINER_START_CONTEXT)?
basically make Line 105 explicit and retain current validations as is for non-pinned containers.

pragnagopa: @balag0 - as discussed offline , my understanding was that we can safely remove hardcoding here as platform explicitly set placeholder mode to 1. Why should this be special cased for pinned apps only?

balag0: Correct, but it is not set to 1 always. it is only done for pinned apps, so we need special handling here correspondingly.

1) Current
Non-pinned apps , pinned apps - assign can be called only if container is in placeholder mode. This is a problem for pinned apps

2) With this change
a) Pinned apps - Assign can be called in non-placeholder mode as well. This is needed.
b) Non-pinned apps - Assign can be called in non-placeholder mode as well. 

2b is not needed currently. if Assign does get called for non-pinned apps it is a bug somewhere and should be blocked. Right?

Something like
```
if (!_webHostEnvironment.InStandbyMode)
{
    _logger.LogError("Assign called while host is not in placeholder mode");
    // This is only true when specializing pinned containers.
	if (string.NullOrEmpty(CONTAINER_START_CONTEXT))
    {
		return false;
	}
}
```

divyagandhisethi:
Can someone please approve this?

src/WebJobs.Script.WebHost/Management/InstanceManager.cs
```diff
@@ -102,7 +102,17 @@ public bool StartAssignment(HostAssignmentContext context)
         {
             if (!_webHostEnvironment.InStandbyMode)
             {
-                _logger.LogError("Assign called while host is not in placeholder mode");
+                // This is only true when specializing pinned containers.
+                if (!context.Environment.TryGetValue(EnvironmentSettingNames.ContainerStartContext, out string startContext))
```
yojagad, line 106: >if (!context.Environment.TryGetValue(EnvironmentSettingNames.ContainerStartContext, out string startContext)) [](start = 16, length = 109)

Do we need to handle scenario where Env setting is present but startContext can still be null or empty? (return false in that case, right?)

yojagad (approved):
Added minor comment. Might be a non-issue.

src/WebJobs.Script.WebHost/Management/InstanceManager.cs
balag0, line 106: nit: for safety, add (|| string.IsNullEmpty(startContext)) in case something changes later on the platform side?

balag0 (approved):

src/WebJobs.Script.WebHost/Management/InstanceManager.cs
divyagandhisethi: If start context is null or empty, that means it's a placeholder pool container.

divyagandhisethi: I think we should continue failing in that case, because we should never be setting a null/empty env variable from the platform side.

PR #7226, opened 2021-03-15
Fixes slow start of multiple worker processes (resolves #7136)

### LINKED ISSUE #7136: Worker processes do not start in linear time

alrod:
resolves #7136 

### Pull request checklist

### Additional information

Additional PR information
This is continuation of the PR:
https://github.com/Azure/azure-functions-host/pull/7138

### DIFF
--- a/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
+++ b/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
@@ -147,15 +147,8 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
                 for (var count = startIndex; count < _maxProcessCount
                     && !_processStartCancellationToken.IsCancellationRequested; count++)
                 {
-                    // Start a new worker right away if there is no one
-                    if (startIndex > 0)
-                    {
-                        await Task.Delay(_processStartupInterval);
-                    }
-                    if (!_processStartCancellationToken.IsCancellationRequested)
-                    {
-                        await startAction();
-                    }
+                    await startAction();
+                    await Task.Delay(_processStartupInterval);
                 }
             }, _processStartCancellationToken.Token);
         }
--- a/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
+++ b/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
@@ -51,7 +51,7 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
         private ConcurrentStack<WorkerErrorEvent> _languageWorkerErrors = new ConcurrentStack<WorkerErrorEvent>();
         private CancellationTokenSource _processStartCancellationToken = new CancellationTokenSource();
         private CancellationTokenSource _disposeToken = new CancellationTokenSource();
-        private int _processStartupInterval = (int)TimeSpan.FromSeconds(10).TotalMilliseconds;
+        private int _processStartupInterval;
 
         public RpcFunctionInvocationDispatcher(IOptions<ScriptJobHostOptions> scriptHostOptions,
             IMetricsLogger metricsLogger,
@@ -141,14 +141,26 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
 
         private void StartWorkerProcesses(int startIndex, Func<Task> startAction)
         {
-            // Starting the worker processes in a different thread to avoid delaying in the host initialization
             Task.Run(async () =>
             {
                 for (var count = startIndex; count < _maxProcessCount
                     && !_processStartCancellationToken.IsCancellationRequested; count++)
                 {
-                    await startAction();
-                    await Task.Delay(_processStartupInterval);
+                    try
+                    {
+                        await startAction();
+                        await Task.Delay(_processStartupInterval);
+                    }
+                    catch (Exception ex)
+                    {
+                        _logger.LogError(ex, $"Failed to start a new language worker for runtime: {_workerRuntime}.");
+                    }
+                }
+
+                if (State != FunctionInvocationDispatcherState.Initialized)
+                {
+                    _logger.LogError("FunctionInvocationDispatcher was not initialized. Shutting down and proactively recycling the Functions Host to recover.");
+                    _applicationLifetime.StopApplication();
                 }
             }, _processStartCancellationToken.Token);
         }
@@ -215,7 +227,7 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
                         }
                     }
                     StartWorkerProcesses(webhostLanguageWorkerChannels.Count(), InitializeWebhostLanguageWorkerChannel);
-                    if (webhostLanguageWorkerChannels.Count() > 0)
+                    if (webhostLanguageWorkerChannels.Any())
                     {
                         SetFunctionDispatcherStateToInitializedAndLog();
                     }
--- a/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
+++ b/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
@@ -51,7 +51,7 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
         private ConcurrentStack<WorkerErrorEvent> _languageWorkerErrors = new ConcurrentStack<WorkerErrorEvent>();
         private CancellationTokenSource _processStartCancellationToken = new CancellationTokenSource();
         private CancellationTokenSource _disposeToken = new CancellationTokenSource();
-        private int _debounceMilliSeconds = (int)TimeSpan.FromSeconds(10).TotalMilliseconds;
+        private int _processStartupInterval = (int)TimeSpan.FromSeconds(10).TotalMilliseconds;
 
         public RpcFunctionInvocationDispatcher(IOptions<ScriptJobHostOptions> scriptHostOptions,
             IMetricsLogger metricsLogger,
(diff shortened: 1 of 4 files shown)


src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -209,17 +217,18 @@ public async Task InitializeAsync(IEnumerable<FunctionMetadata> functions, Cance
...
+                    if (webhostLanguageWorkerChannels.Count() > 0)
+                    {
+                        SetFunctionDispatcherStateToInitializedAndLog();
+                    }
                 }
                 else
                 {
-                    await InitializeJobhostLanguageWorkerChannelAsync(0);
-                    StartWorkerProcesses(1, InitializeJobhostLanguageWorkerChannelAsync);
+                    StartWorkerProcesses(0, InitializeJobhostLanguageWorkerChannelAsync);
```
alrod, line 231: This can potentially improve cold start for non-placeholder scenarios as we are not waiting for the process to start.

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -142,13 +139,25 @@ internal async void ShutdownWebhostLanguageWorkerChannels()
...
+            Task.Run(async () =>
             {
-                startAction = startAction.Debounce(_processStartCancellationToken.Token, count * _debounceMilliSeconds);
-                startAction();
-            }
+                for (var count = startIndex; count < _maxProcessCount
+                    && !_processStartCancellationToken.IsCancellationRequested; count++)
+                {
+                    // Start a new worker right away if there is no one
+                    if (startIndex > 0)
```
paulbatum: If you invert the order of these two operations - call the action first, then do task.delay, I think you can remove this startindex check.

paulbatum (changes requested):
Can you add a new test case that fails without this fix? My previous PR had a unit test that confirmed the behavior change, though there was some feedback on the PR to use a different testing approach.

alrod:
> Can you add a new test case that fails without this fix? My previous PR had a unit test that confirmed the behavior change, though there was some feedback on the PR to use a different testing approach.

I added some new code to existing tests as the tests already covers multiple language workers start up.

test/WebJobs.Script.Tests/Workers/Rpc/RpcFunctionInvocationDispatcherTests.cs
```diff
@@ -562,5 +569,21 @@ private IEnumerable<FunctionMetadata> GetTestFunctionsList(string runtime)
...
+            var startTimestamps = _testLogger.GetLogMessages().Where(x => x.FormattedMessage
+                .Contains("RegisterFunctions called")).Select(x => x.Timestamp).ToList();
+            if (ignoreFirstStart)
+            {
+                startTimestamps.RemoveAt(0);
+            }
+            for (int i = 1; i < startTimestamps.Count(); i++)
+            {
+                var diff = startTimestamps[i] - startTimestamps[i - 1];
+                Assert.True(diff > TimeSpan.FromSeconds(10) && diff < TimeSpan.FromSeconds(15));
```
paulbatum: I would consider making the hardcoded values (10 seconds and 15 seconds) parameters, so that readers of the process start unit tests have an idea of what VerifyStartIntervals is doing without having to read through the implementation.

alrod: Fixed.

paulbatum (dismissed):
Would be good to get one other approver before merging.

pragnagopa:
> Would be good to get one other approver before merging.

Looking into this!

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -51,7 +51,7 @@ internal class RpcFunctionInvocationDispatcher : IFunctionInvocationDispatcher
         private ConcurrentStack<WorkerErrorEvent> _languageWorkerErrors = new ConcurrentStack<WorkerErrorEvent>();
         private CancellationTokenSource _processStartCancellationToken = new CancellationTokenSource();
         private CancellationTokenSource _disposeToken = new CancellationTokenSource();
-        private int _debounceMilliSeconds = (int)TimeSpan.FromSeconds(10).TotalMilliseconds;
+        private int _processStartupInterval = (int)TimeSpan.FromSeconds(10).TotalMilliseconds;
```
fabiocav: Just make this constant with the proper comment/calculation.

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -142,13 +139,18 @@ internal async void ShutdownWebhostLanguageWorkerChannels()
             await _webHostLanguageWorkerChannelManager?.ShutdownChannelsAsync();
         }
 
-        private void StartWorkerProcesses(int startIndex, Action startAction)
+        private void StartWorkerProcesses(int startIndex, Func<Task> startAction)
         {
-            for (var count = startIndex; count < _maxProcessCount; count++)
+            // Starting the worker processes in a different thread to avoid delaying in the host initialization
```
fabiocav: This task is unobserved. Please add a test inducing a failure in the code executed in the lambda and ensure that is properly handled. As-is, potential exceptions here will go unhandled in the background.

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -209,17 +211,18 @@ public async Task InitializeAsync(IEnumerable<FunctionMetadata> functions, Cance
                             {
                                 _logger.LogWarning(ex, "Removing errored webhost language worker channel for runtime: {workerRuntime} workerId:{workerId}", _workerRuntime, workerId);
                                 await _webHostLanguageWorkerChannelManager.ShutdownChannelIfExistsAsync(_workerRuntime, workerId, ex);
-                                InitializeWebhostLanguageWorkerChannel();
                             }
                         }
                     }
                     StartWorkerProcesses(webhostLanguageWorkerChannels.Count(), InitializeWebhostLanguageWorkerChannel);
-                    SetFunctionDispatcherStateToInitializedAndLog();
+                    if (webhostLanguageWorkerChannels.Count() > 0)
```
fabiocav: No need to iterate. Just replace with `Any()`

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
fabiocav: nit: also update the comment as this is just scheduling a new `Task`.

fabiocav (changes requested):

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
pragnagopa: FYI - Failure to start worker process is handled here: https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs#L136-L147 - we have test coverage for this

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
pragnagopa: For the first process - we do want to block. Launching multiple  processes can be "fire and forget".
As good way to test this E2E - locally - Set up a queue with 100+ messages then start the functions runtime - check if executions succeed.

pragnagopa (commented):
Added one comment. Rest looks good. Thanks!

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
fabiocav: Can you clarify the comment above? Is that additional information to share that a class of errors is properly handled and surfaced, or are you suggesting we leave this as-is?

The code you've pointed to handles a number of known potential exceptions, but it does not make what we have here safe. There are a of calls up to that point that could introduce a failure that would end up being thrown in the background, and that's if we only think about the current state of the code. As some of those code paths evolve in the future, chances of exposing the fragility here just keep growing.

Any concerns with being defensive here?

pragnagopa: Sorry. I should have provided a complete response. The code I referred to above is an example of how to induce worker start up errors. Yes, if there is a way to log Task errors - we should

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
alrod: I tested this with 100+ messages and all executions were succeed.
In old code we also set up `RpcFunctionInvocationDispatcher.State=FunctionInvocationDispatcherState.Initialized`; state in the non-blocking code:
https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs#L119

pragnagopa: Thanks for confirming!

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
alrod: can't make this constant as we set the value in workerConfig. I removed the assignment as it's calculated on InitializeAsync call.

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
alrod: Added the logging and a test.

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -142,13 +139,30 @@ internal async void ShutdownWebhostLanguageWorkerChannels()
...
+                        await startAction();
+                        await Task.Delay(_processStartupInterval);
+                    }
+                    catch (Exception ex)
+                    {
+                        _logger.LogError(ex, $"Failed to start a new language worker for runtime: {_workerRuntime}.");
+                    }
+                }
+
+                if (State != FunctionInvocationDispatcherState.Initialized)
```
pragnagopa: Checking initialized state is handled here: https://github.com/Azure/azure-functions-host/blob/1927f5fd736931d2a6407b84d35927f2404eb8dc/src/WebJobs.Script/Description/Workers/WorkerFunctionInvoker.cs#L109-L123 

Safe to remove the check here.

alrod: The check is removed,

pragnagopa (approved):

fabiocav (approved):

PR #7239, opened 2021-03-20
Get shared memory directories (in Unix) from AppSettings 

### LINKED ISSUE #6791: Shared memory data transfer between Functions host and out-of-proc workers for binary data

gohar94:
Provides ability to specify custom paths for use with shared memory in Unix. If not provided, the default (`/dev/shm`) is used.
This can be helpful when running on macOS, for example, where the user needs to create a volume manually before it can be used for shared memory data transfer and `/dev/shm` is not allowed in macOS.

resolves #issue_for_this_pr

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.Grpc/Extensions/RpcSharedMemoryDataExtensions.cs
+++ b/src/WebJobs.Script.Grpc/Extensions/RpcSharedMemoryDataExtensions.cs
@@ -13,6 +13,11 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc.Extensions
     {
         internal static async Task<RpcSharedMemory> ToRpcSharedMemoryAsync(this object value, ILogger logger, string invocationId, ISharedMemoryManager sharedMemoryManager)
         {
+            if (value == null)
+            {
+                return new RpcSharedMemory();
+            }
+
             if (!sharedMemoryManager.IsSupported(value))
             {
                 return null;
--- a/src/WebJobs.Script/Workers/SharedMemoryDataTransfer/MemoryMappedFileAccessorUnix.cs
+++ b/src/WebJobs.Script/Workers/SharedMemoryDataTransfer/MemoryMappedFileAccessorUnix.cs
@@ -5,7 +5,9 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.IO.MemoryMappedFiles;
+using System.Linq;
 using System.Runtime.InteropServices;
+using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
 using Microsoft.Extensions.Logging;
 
 namespace Microsoft.Azure.WebJobs.Script.Workers.SharedMemoryDataTransfer
@@ -15,14 +17,21 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.SharedMemoryDataTransfer
     /// </summary>
     public class MemoryMappedFileAccessorUnix : MemoryMappedFileAccessor
     {
-        public MemoryMappedFileAccessorUnix(ILogger<MemoryMappedFileAccessor> logger) : base(logger)
+        private IEnvironment _environment;
+
+        public MemoryMappedFileAccessorUnix(ILogger<MemoryMappedFileAccessor> logger, IEnvironment environment) : base(logger)
         {
             ValidatePlatform(new List<OSPlatform>()
             {
                 OSPlatform.Linux,
                 OSPlatform.OSX
             });
+
+            _environment = environment;
+            ValidDirectories = GetValidDirectories();
         }
+
+        internal List<string> ValidDirectories { get; private set; }
 
         public override bool TryCreate(string mapName, long size, out MemoryMappedFile mmf)
         {
@@ -128,6 +137,64 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.SharedMemoryDataTransfer
         }
 
         /// <summary>
+        /// Checks if a list of directories is specified in AppSettings to create <see cref="MemoryMappedFile"/>.
+        /// If one is specified, returns that list. Otherwise returns the default list.
+        /// </summary>
+        /// <returns>List of paths of directories where <see cref="MemoryMappedFile"/> are allowed to be created.</returns>
+        private List<string> GetAllowedDirectories()
+        {
+            string envVal = _environment.GetEnvironmentVariable(RpcWorkerConstants.FunctionsUnixSharedMemoryDirectories);
+            if (string.IsNullOrEmpty(envVal))
+            {
+                return SharedMemoryConstants.TempDirs;
+            }
+
+            return envVal.Split(',').ToList();
+        }
+
+        /// <summary>
+        /// From a list of allowed directories where <see cref="MemoryMappedFile"/> can be created, it will return
+        /// a list of those that are valid (i.e. exist, or have been successfully created).
+        /// </summary>
+        /// <returns>List of paths of directories where <see cref="MemoryMappedFile"/> can be created.</returns>
+        private List<string> GetValidDirectories()
+        {
+            List<string> allowedDirectories = GetAllowedDirectories();
+            List<string> validDirectories = new List<string>();
+
+            foreach (string directory in allowedDirectories)
+            {
+                string path = Path.Combine(directory, SharedMemoryConstants.TempDirSuffix);
+                if (Directory.Exists(path))
+                {
+                    Logger.LogInformation("Found directory for shared memory usage: {Directory}", path);
+                    validDirectories.Add(path);
+                }
+                else
+                {
+                    try
+                    {
+                        DirectoryInfo info = Directory.CreateDirectory(path);
+                        if (info.Exists)
+                        {
+                            validDirectories.Add(path);
+                        }
+                        else
+                        {
+                            Logger.LogWarning("Cannot create directory for shared memory usage: {Directory}", path);
+                        }
+                    }
+                    catch (Exception exception)
+                    {
+                        Logger.LogWarning(exception, "Cannot create directory for shared memory usage: {Directory}", path);
+                    }
+                }
+            }
+
+            return validDirectories;
+        }
+
+        /// <summary>
         /// Get the path of the file to store a <see cref="MemoryMappedFile"/>.
         /// We first try to mount it in memory-mounted directories (e.g. /dev/shm/).
         /// </summary>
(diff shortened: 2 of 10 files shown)


src/WebJobs.Script.Grpc/Extensions/RpcSharedMemoryDataExtensions.cs
```diff
@@ -13,6 +13,11 @@ internal static class RpcSharedMemoryDataExtensions
     {
         internal static async Task<RpcSharedMemory> ToRpcSharedMemoryAsync(this object value, ILogger logger, string invocationId, ISharedMemoryManager sharedMemoryManager)
         {
+            if (value == null)
```
goiri, line 16: This is an optimization but curious on which situation you found this?

src/WebJobs.Script/Workers/SharedMemoryDataTransfer/MemoryMappedFileAccessorUnix.cs
```diff
@@ -127,6 +136,64 @@ public override void Delete(string mapName, MemoryMappedFile mmf)
...
+        /// <returns>List of paths of directories where <see cref="MemoryMappedFile"/> are allowed to be created.</returns>
+        private List<string> GetAllowedDirectories()
+        {
+            string envVal = _environment.GetEnvironmentVariable(RpcWorkerConstants.FunctionsUnixSharedMemoryDirectories);
+            if (string.IsNullOrEmpty(envVal))
+            {
+                return SharedMemoryConstants.TempDirs;
+            }
+
+            return envVal.Split(',').ToList();
```
goiri, line 152: Do we want to sanitize empty entries?
Do we want to log something like in the other?

goiri: I see the logging is in the following, I guess that also covers the empty/non-valid entries.

src/WebJobs.Script/Workers/SharedMemoryDataTransfer/MemoryMappedFileAccessorUnix.cs
```diff
@@ -127,6 +136,64 @@ public override void Delete(string mapName, MemoryMappedFile mmf)
...
+        /// </summary>
+        /// <returns>List of paths of directories where <see cref="MemoryMappedFile"/> can be created.</returns>
+        private List<string> GetValidDirectories()
+        {
+            List<string> allowedDirectories = GetAllowedDirectories();
+            List<string> validDirectories = new List<string>();
+
+            foreach (string directory in allowedDirectories)
+            {
+                string path = Path.Combine(directory, SharedMemoryConstants.TempDirSuffix);
```
goiri, line 167: Should this "TempDirSuffix" also be an environment variable?

test/WebJobs.Script.Tests/Workers/MemoryMappedFileAccessorTests.cs
```diff
@@ -17,22 +18,45 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.Workers
...
             {
-                _mapAccessor = new MemoryMappedFileAccessorUnix(logger);
+                _mapAccessor = new MemoryMappedFileAccessorUnix(logger, _testEnvironment);
             }
         }
 
+        [Fact]
+        public void Use_AppSetting_Directory_Unix()
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
goiri, line 49: There's no cleaner way to ignore on Windows?

src/WebJobs.Script/Workers/SharedMemoryDataTransfer/MemoryMappedFileAccessorUnix.cs
gohar94: Umm, I think this is fine. That will be too many configurable pieces and I don't see a use for that. The user can set the main path and inside that we create a subdirectory by this name.

test/WebJobs.Script.Tests/Workers/MemoryMappedFileAccessorTests.cs
gohar94: I was disappointed to see that people are recommending doing [this](https://stackoverflow.com/a/4421941/3132415) as the solution. I went with keeping it simple, but I agree, it's not ideal.

src/WebJobs.Script.Grpc/Extensions/RpcSharedMemoryDataExtensions.cs
gohar94: When you try to access a blob that does not exist, for example.

gohar94: Doing what the corresponding `ToRpc` method [does](https://github.com/Azure/azure-functions-host/blob/1927f5fd736931d2a6407b84d35927f2404eb8dc/src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs#L57).

test/WebJobs.Script.Tests/Workers/MemoryMappedFileAccessorTests.cs
vrdmr: Just FYI - there is a helper method (PlatformHelper.IsWindows()) that is present - the actual implementation of the method is the same as yours, but I would recommend using that.

goiri (dismissed):

test/WebJobs.Script.Tests/Workers/MemoryMappedFileAccessorTests.cs
vrdmr: Just checking - isn't there a skip option for the test?

src/WebJobs.Script/Workers/SharedMemoryDataTransfer/MemoryMappedFileAccessorUnix.cs
```diff
@@ -127,6 +136,64 @@ public override void Delete(string mapName, MemoryMappedFile mmf)
...
+                if (Directory.Exists(path))
+                {
+                    Logger.LogInformation("Found directory for shared memory usage: {Directory}", path);
```
pragnagopa: change this log to Trace level

src/WebJobs.Script/Workers/SharedMemoryDataTransfer/MemoryMappedFileAccessorUnix.cs
```diff
@@ -127,6 +136,64 @@ public override void Delete(string mapName, MemoryMappedFile mmf)
...
+                        }
+                    }
+                    catch (Exception exception)
+                    {
+                        Logger.LogWarning(exception, "Cannot create directory for shared memory usage: {Directory}", path);
+                    }
+                }
+            }
+
+            return validDirectories;
```
pragnagopa, line 204: Log debug - value of `validDirectories`  before returning

src/WebJobs.Script/Workers/SharedMemoryDataTransfer/MemoryMappedFileAccessorUnix.cs
```diff
@@ -127,6 +136,64 @@ public override void Delete(string mapName, MemoryMappedFile mmf)
             }
         }
 
+        /// <summary>
+        /// Checks if a list of directories is specified in AppSettings to create <see cref="MemoryMappedFile"/>.
+        /// If one is specified, returns that list. Otherwise returns the default list.
+        /// </summary>
+        /// <returns>List of paths of directories where <see cref="MemoryMappedFile"/> are allowed to be created.</returns>
+        private List<string> GetAllowedDirectories()
```
pragnagopa: internal and add unit tests

src/WebJobs.Script/Workers/SharedMemoryDataTransfer/MemoryMappedFileAccessorUnix.cs
```diff
@@ -127,6 +136,64 @@ public override void Delete(string mapName, MemoryMappedFile mmf)
...
+
+            return envVal.Split(',').ToList();
+        }
+
+        /// <summary>
+        /// From a list of allowed directories where <see cref="MemoryMappedFile"/> can be created, it will return
+        /// a list of those that are valid (i.e. exist, or have been successfully created).
+        /// </summary>
+        /// <returns>List of paths of directories where <see cref="MemoryMappedFile"/> can be created.</returns>
+        private List<string> GetValidDirectories()
```
pragnagopa: internal and add unit tests

pragnagopa (commented):
Added minor comments

src/WebJobs.Script/Workers/SharedMemoryDataTransfer/MemoryMappedFileAccessorUnix.cs
```diff
@@ -127,6 +136,74 @@ public override void Delete(string mapName, MemoryMappedFile mmf)
...
+            foreach (string directory in allowedDirectories)
+            {
+                string path = Path.Combine(directory, SharedMemoryConstants.TempDirSuffix);
+                if (Directory.Exists(path))
+                {
+                    // If the directory already exists (maybe from a previous run of the host) then clean it up and start afresh
+                    Logger.LogTrace("Found directory for shared memory usage: {Directory}", path);
+                    try
+                    {
+                        Directory.Delete(path);
```
goiri: Add a comment explaining why we do this.

src/WebJobs.Script/Workers/SharedMemoryDataTransfer/MemoryMappedFileAccessorUnix.cs
```diff
@@ -127,6 +136,74 @@ public override void Delete(string mapName, MemoryMappedFile mmf)
...
+                        Logger.LogWarning("Cannot create directory for shared memory usage: {Directory}", path);
+                    }
+                }
+                catch (Exception exception)
+                {
+                    Logger.LogWarning(exception, "Cannot create directory for shared memory usage: {Directory}", path);
+                }
+            }
+
+            Logger.LogDebug("Valid directories for shared memory usage: {Directories}", validDirectories);
```
pragnagopa: fix log - `string.Join(",", validDirectories)`

src/WebJobs.Script/Workers/SharedMemoryDataTransfer/MemoryMappedFileAccessorUnix.cs
```diff
@@ -127,6 +136,74 @@ public override void Delete(string mapName, MemoryMappedFile mmf)
...
+                {
+                    DirectoryInfo info = Directory.CreateDirectory(path);
+                    if (info.Exists)
+                    {
+                        validDirectories.Add(path);
+                        Logger.LogTrace("Created directory for shared memory usage: {Directory}", path);
+                    }
+                    else
+                    {
+                        Logger.LogWarning("Cannot create directory for shared memory usage: {Directory}", path);
```
pragnagopa: be more explicit in logging to distinguish failure vs no exists- `Directory for shared memory usage: {Directory}, does not exist`

pragnagopa:
@fabiocav - flagging here that We should have build pipeline for Linux that runs unit tests. Opened https://github.com/Azure/azure-functions-host/issues/7250 to track it.

pragnagopa (approved):

Issue #7188, opened 2021-02-25
Add Inflight invocations completed status on the host


pragnagopa:
This is a follow up for DrainMode feature

Calling drain mode API allows for a graceful shutdown of the host to complete inflight invocations and stops listening for new events from trigger sources.

To handle this in LinuxConsumption more effectively, host needs to expose an API that platform can poll to figure out if all the inflight invocations are completed. Without this information a host instance needs to stay in drain mode for a specified amount of time - even though all the inflight invocations completed.  This leads to capacity crunch.

cc @fabiocav @soninaren @divyagandhisethi @balag0

pragnagopa:
@divyagandhisethi - Can DataRole determine state of inflight invocations by looking at functionExecution status?

soninaren:
We do provide customers with cancellation tokens to all functions including inflight and new invocations.  Any new invocations that get started after drain mode has been activated should be handled by the customer. Functions already inflight are provided ample time for them to complete execution.

soninaren:
Also the proposed API may not be helpful because the customer could pre-fetch a 1000 messages. Each can take up to 10 minute to finish execution and user could also configure the function app to process one function at a time.

pragnagopa:
> Functions already inflight are provided ample time for them to complete execution.

Can you please expand on what ample time means? My understanding is there is a fixed grace period that we allow for drain mode.

> Also the proposed API may not be helpful because the customer could pre-fetch a 1000 messages. Each can take up to 10 minute to finish execution and user could also configure the function app to process one function at a time.

Ask here is for platform to be able to detect and optimize cases where all invocations after drain mode is called complete in < 10mins . This will ensure a host instance (container) is not waiting for the whole 10mins before being teared down.

soninaren:
> Can you please expand on what ample time means? My understanding is there is a fixed grace period that we allow for drain mode.

By ample time I meant `time >= default timeout period` for the respective sku. So for consumption it would be >=10 mins
 
> Ask here is for platform to be able to detect and optimize cases where all invocations after drain mode is called complete in < 10mins . This will ensure a host instance (container) is not waiting for the whole 10mins before being teared down.

The ask makes sense and has clear advantages. We would have to investigate this further to understand performance impact as this would require keeping track of all invocations across all extensions.

soninaren:
tagging @mathewc as he is working on a design that will likely provide this information.

pragnagopa:
cc @alrod @AnatoliB fyi

PR #7262, opened 2021-04-01
Call EnableDrainModeAsync on ApplicationStopping

### LINKED ISSUE #7332: Handle SigTerm with grace period to shutdown when running in Kubernetes
The way we scale down in Kubernetes is via the KEDA scaler. When KEDA issues a scale down command, it sends the SIGTERM signal to the functions runtime pod. After that if no action is taken, it will kill the pod in 30 seconds (by default) after it issued SIGTERM

Each pod in Kubernetes is expected to handle this SIGTERM notification from KEDA in it's own way to handle graceful shutdown
Without a grace period for shutdown, function invocations will be killed everytime KEDA scales down.

ahmelsayed:
Call `EnableDrainModeAsync()` during `ApplicationStopping`

Resolves #7332 

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
+++ b/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
@@ -700,7 +700,8 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
                     if (drainModeManager != null)
                     {
                         _logger.LogDebug("Application Stopping: enabling drain mode");
-                        drainModeManager.EnableDrainModeAsync(CancellationToken.None).Wait();
+                        drainModeManager.EnableDrainModeAsync(CancellationToken.None);
+                        Thread.Sleep(TimeSpan.FromMinutes(10));
                     }
                     _logger.LogDebug("Draining completed");
                 }
--- a/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
+++ b/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
@@ -694,13 +694,14 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
             _applicationLifetime.ApplicationStopping.Register(() =>
             {
                 Interlocked.Exchange(ref _applicationStopping, 1);
-                if (_environment.IsDrainOnApplicationStopping())
+                if (_environment.DrainOnApplicationStoppingEnabled())
                 {
                     var drainModeManager = _host?.Services.GetService<IDrainModeManager>();
                     if (drainModeManager != null)
                     {
                         _logger.LogDebug("Application Stopping: initiate drain mode");
                         drainModeManager.EnableDrainModeAsync(CancellationToken.None);
+                        // Workaround until https://github.com/Azure/azure-functions-host/issues/7188 is addressed.
                         Thread.Sleep(TimeSpan.FromMinutes(10));
                     }
                 }
--- a/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
+++ b/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
@@ -11,6 +11,7 @@ using Microsoft.ApplicationInsights.AspNetCore;
 using Microsoft.ApplicationInsights.AspNetCore.Extensions;
 using Microsoft.ApplicationInsights.Extensibility;
 using Microsoft.ApplicationInsights.Extensibility.Implementation.ApplicationId;
+using Microsoft.Azure.WebJobs.Host;
 using Microsoft.Azure.WebJobs.Logging;
 using Microsoft.Azure.WebJobs.Script.Configuration;
 using Microsoft.Azure.WebJobs.Script.Diagnostics;
@@ -693,6 +694,16 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
             _applicationLifetime.ApplicationStopping.Register(() =>
             {
                 Interlocked.Exchange(ref _applicationStopping, 1);
+                if (_environment.IsDrainOnApplicationStopping())
+                {
+                    var drainModeManager = _host?.Services.GetService<IDrainModeManager>();
+                    if (drainModeManager != null)
+                    {
+                        _logger.LogInformation("Application Stopping: enabling drain mode");
+                        drainModeManager.EnableDrainModeAsync(CancellationToken.None).Wait();
+                    }
+                    _logger.LogInformation("Draining completed");
+                }
             });
 
             _applicationLifetime.ApplicationStopped.Register(() =>
--- a/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
+++ b/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
@@ -699,10 +699,10 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
                     var drainModeManager = _host?.Services.GetService<IDrainModeManager>();
                     if (drainModeManager != null)
                     {
-                        _logger.LogInformation("Application Stopping: enabling drain mode");
+                        _logger.LogDebug("Application Stopping: enabling drain mode");
                         drainModeManager.EnableDrainModeAsync(CancellationToken.None).Wait();
                     }
-                    _logger.LogInformation("Draining completed");
+                    _logger.LogDebug("Draining completed");
                 }
             });
 
--- a/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
+++ b/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
@@ -699,11 +699,10 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
                     var drainModeManager = _host?.Services.GetService<IDrainModeManager>();
                     if (drainModeManager != null)
                     {
-                        _logger.LogDebug("Application Stopping: enabling drain mode");
+                        _logger.LogDebug("Application Stopping: initiate drain mode");
                         drainModeManager.EnableDrainModeAsync(CancellationToken.None);
                         Thread.Sleep(TimeSpan.FromMinutes(10));
                     }
-                    _logger.LogDebug("Draining completed");
                 }
             });
 
--- a/src/WebJobs.Script/Environment/EnvironmentExtensions.cs
+++ b/src/WebJobs.Script/Environment/EnvironmentExtensions.cs
@@ -475,7 +475,7 @@ namespace Microsoft.Azure.WebJobs.Script
                 StringComparison.OrdinalIgnoreCase);
         }
 
-        public static bool IsDrainOnApplicationStopping(this IEnvironment environment)
+        public static bool DrainOnApplicationStoppingEnabled(this IEnvironment environment)
         {
             return !string.IsNullOrEmpty(environment.GetEnvironmentVariable(KubernetesServiceHost)) ||
                 (bool.TryParse(environment.GetEnvironmentVariable(DrainOnApplicationStopping), out bool v) && v);
--- a/src/WebJobs.Script/Environment/EnvironmentExtensions.cs
+++ b/src/WebJobs.Script/Environment/EnvironmentExtensions.cs
@@ -474,5 +474,10 @@ namespace Microsoft.Azure.WebJobs.Script
             return string.Equals(environment.GetFunctionsWorkerRuntime(), RpcWorkerConstants.PowerShellLanguageWorkerName,
                 StringComparison.OrdinalIgnoreCase);
         }
+        public static bool IsDrainOnApplicationStopping(this IEnvironment environment)
+        {
+            return !string.IsNullOrEmpty(environment.GetEnvironmentVariable(KubernetesServiceHost)) ||
+                (bool.TryParse(environment.GetEnvironmentVariable(DrainOnApplicationStopping), out bool v) && v);
+        }
     }
 }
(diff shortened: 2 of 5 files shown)


src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -693,6 +694,16 @@ private void RegisterApplicationLifetimeEvents()
...
                 Interlocked.Exchange(ref _applicationStopping, 1);
+                if (_environment.IsDrainOnApplicationStopping())
+                {
+                    var drainModeManager = _host?.Services.GetService<IDrainModeManager>();
+                    if (drainModeManager != null)
+                    {
+                        _logger.LogInformation("Application Stopping: enabling drain mode");
+                        drainModeManager.EnableDrainModeAsync(CancellationToken.None).Wait();
+                    }
+                    _logger.LogInformation("Draining completed");
```
pragnagopa: nit: LogDebug - Generally we use debug level logs for system logs.

src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
```diff
@@ -117,5 +117,7 @@ public static class EnvironmentSettingNames
         public const string AntaresComputerName = "COMPUTERNAME";
 
         public const string AppKind = "APP_KIND";
+
+        public const string DrainOnApplicationStopping = "DRAIN_ON_APP_STOPPING";
```
pragnagopa: prefix with `FUNCTIONS_` - `FUNCTIONS_ENABLE_DRAIN_ON_APP_STOPPING`

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -459,5 +459,11 @@ public static string GetStorageSuffix(this IEnvironment environment)
                     return CloudConstants.AzureStorageSuffix;
             }
         }
+
+        public static bool IsDrainOnApplicationStopping(this IEnvironment environment)
+        {
+            return !string.IsNullOrEmpty(environment.GetEnvironmentVariable(KubernetesServiceHost)) ||
```
pragnagopa, line 480: Add unit tests here : https://github.com/Azure/azure-functions-host/blob/dev/test/WebJobs.Script.Tests/Extensions/EnvironmentExtensionsTests.cs

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -693,6 +694,16 @@ private void RegisterApplicationLifetimeEvents()
             _applicationLifetime.ApplicationStopping.Register(() =>
             {
                 Interlocked.Exchange(ref _applicationStopping, 1);
+                if (_environment.IsDrainOnApplicationStopping())
+                {
+                    var drainModeManager = _host?.Services.GetService<IDrainModeManager>();
+                    if (drainModeManager != null)
+                    {
+                        _logger.LogInformation("Application Stopping: enabling drain mode");
+                        drainModeManager.EnableDrainModeAsync(CancellationToken.None).Wait();
```
pragnagopa: Is wait needed? If not, use similar approach to: https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.WebHost/Controllers/HostController.cs#L115

Also, if wait is needed - pass in timeout

pragnagopa (commented):
Added few comments. Thanks!

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
ahmelsayed: Yes, a synchronous wait is needed here, because otherwise .NET assumes that the process is ready to shutdown. I'm not sure what timeout to set since the original ask was to enable a function to run for 8 or 10 hours to completion. If we add a default timeout, then I think it'll need to be configurable somewhere to allow those scenarios. Should that go in host.json?

ahmelsayed: btw, in Kubernetes that timeout is managed externally. Kubernetes will sent the process a SIGTERM, which invokes `ApplicationStopping`, then waits for a set time (default 30 seconds, but configurable) then it'll send a SIGKILL, which will immediately invoke `ApplicationStopped` regardless of the wait here.

pragnagopa: Tagging @divyagandhisethi - as Kubernetes scenario is similar to how drain mode is handled in Linux Consumption -  `Drain` api is invoked on the host which puts instance in `DrainMode` and then after a pre-configured timeout - `SIGKILL` is sent which terminates the container.

Can we take similar approach here and  call `Drain` API?

divyagandhisethi: Yes, interested to know how we will determine the timeout between SIGTERM and SIGKILL. I think the same solution as https://github.com/Azure/azure-functions-host/issues/7188 is needed here - an endpoint on the host which tells us when all function invocations have been drained and it is safe to kill the instance at that point.

pragnagopa: @ahmelsayed - In Kubernetes -  Is it possible to explicitly call `Drain`  api on `SIGTERM` ? If yes, we can focus on looking into design for #7188

ahmelsayed: Yes, that's what this PR is doing. `SIGTERM` on the process causes .NET to fire `ApplicationStopping`, which in this PR I call drainModeManager.EnableDrainModeAsync()

pragnagopa: > Yes, that's what this PR is doing.

Sorry - I should have been  more explicit - can we call `Drain` API outside of the host - similar to other skus - Platform makes a call  to Drain API

ahmelsayed: oh sorry I misunderstood. We could add a `preStop` hook on the container that looks like this 

```yaml
lifecycle:
  preStop:
    exec:
      command: ["/bin/sh", "-c", "command to drain the host"]
# or
lifecycle:
  preStop:
    httpGet:
      path: "/admin/host/..."
```
This will execute a command, or sent a request to the container right before it sends SIGTERM. 

There are a couple of downsides to this:
1 - every functions user on kubernetes will need to know how to do this specifically for function deployment. We can auto do it for managed hostings.
2- Auth on the API. We could have a command line utility in the functions container that knows how to call that API on the host.

soninaren: Is this one way transaction? Is there a chance this container would be reused / activated. Also how would it impact other hosting platform App service plan, windows consumption. My main concern is that with indefinite wait it might create orphan hosts. Also we have work in progress for a resume API, So we can quickly start processing events incase there is an inflow of new events. In that case even though drain mode is initiated the same w3wp process might be used to quickly instantiate a new host and this change could potentially create orphan hosts.

pragnagopa: Thanks! Let's do the call in the host then.

pragnagopa (dismissed):
We should continue discussion to resolve #7188 to ensure there is a consistent path to figuring out when to call SIGKILL

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
pragnagopa: Another question: if container will not be terminated until sigkill is called, do we still need synchronous wait?

ahmelsayed: > Is this one way transaction? Is there a chance this container would be reused / activated.

In kubernetes (or .NET core on linux in general) it is one way. A process/container that recieves SIGTERM is expected to be terminated. It'll not be reused. 

This might affect the behavior on Windows, hence it's only enabled by default for Kubernetes and with a flag. 

> Another question: if container will not be terminated until sigkill is called, do we still need synchronous wait?

Yes, sigkill is a last option if the application doesn't terminate on sigterm. The application is expected to terminate on sigterm, but it can trap the signal perform clean up/graceful shutdown. it can't stop sigkill as that will force the process to exit from the OS.

fabiocav:
@soninaren can you please also take a look at this?

pragnagopa:
@soninaren - Gentle ping on this for review.

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -693,6 +694,16 @@ private void RegisterApplicationLifetimeEvents()
...
                 Interlocked.Exchange(ref _applicationStopping, 1);
+                if (_environment.IsDrainOnApplicationStopping())
+                {
+                    var drainModeManager = _host?.Services.GetService<IDrainModeManager>();
+                    if (drainModeManager != null)
+                    {
+                        _logger.LogDebug("Application Stopping: enabling drain mode");
+                        drainModeManager.EnableDrainModeAsync(CancellationToken.None).Wait();
+                    }
+                    _logger.LogDebug("Draining completed");
```
soninaren: This log statement is misleading. At this point all we have done is called stop on all listeners. Which puts the function host in "drain mode". This should say something like "Drain mode enabled"

pragnagopa:
@ahmelsayed  - please address the unresolved comments and update the PR - also need to look at the failing tests

fabiocav:
@ahmelsayed is this ready for review? It has been flagged as a requirement for sprint 100, but looks like there are some unresolved comments

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -693,6 +694,17 @@ private void RegisterApplicationLifetimeEvents()
             _applicationLifetime.ApplicationStopping.Register(() =>
             {
                 Interlocked.Exchange(ref _applicationStopping, 1);
+                if (_environment.IsDrainOnApplicationStopping())
+                {
+                    var drainModeManager = _host?.Services.GetService<IDrainModeManager>();
+                    if (drainModeManager != null)
+                    {
+                        _logger.LogDebug("Application Stopping: enabling drain mode");
+                        drainModeManager.EnableDrainModeAsync(CancellationToken.None);
```
fabiocav, line 703: I'm assuming this call means "initiate draining/drain mode", is that correct?

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -693,6 +694,17 @@ private void RegisterApplicationLifetimeEvents()
...
+                        Thread.Sleep(TimeSpan.FromMinutes(10));
```
fabiocav, line 705: Do we need to sleep for 10 mins here regardless of the work being done in drain mode?

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -693,6 +694,17 @@ private void RegisterApplicationLifetimeEvents()
             _applicationLifetime.ApplicationStopping.Register(() =>
             {
                 Interlocked.Exchange(ref _applicationStopping, 1);
+                if (_environment.IsDrainOnApplicationStopping())
+                {
+                    var drainModeManager = _host?.Services.GetService<IDrainModeManager>();
+                    if (drainModeManager != null)
+                    {
+                        _logger.LogDebug("Application Stopping: enabling drain mode");
```
mathewc: I don't have context on the reason for this change - I didn't see a linked issue or anything? When the application is stopping, all hosted services are also stopping, which will stop the ScriptHost, which will stop listeners and do regular shutdown, etc. already naturally as we've always done. Is there some problem we've identified with that existing behavior?

In WebJobs outside of functions, a customer will similarly call JobHost.StopAsync to shut things down, and they don't know anything about or use IDrainModeManager.

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
ahmelsayed: This is supposed to be a stop gab solution until https://github.com/Azure/azure-functions-host/issues/7188 is done. it used to be 

```csharp
drainModeManager.EnableDrainModeAsync(CancellationToken.None).Wait();
```

as I assumed that task will finish when all the listeners have stopped and all functions finished, but @soninaren explained that this won't work (works for storage queues, but not other triggers and I happen to just test with queues) 

The sleep here is intended as a timeout from the time we receive SIGTERM, giving any function that started at least 10 minutes to finish. It's not a great solution, but should match consumption limit.

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
ahmelsayed: @mathewc this PR is related to the sync up we had on waiting on all inprogress function executions before shutting down the host. In Kubernetes that's orchestrated using SIGTERM (ApplicationStopping event in .NET)

ahmelsayed:
@pragnagopa I addressed the comment. Waiting for the CI to check the test

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -693,6 +694,16 @@ private void RegisterApplicationLifetimeEvents()
             _applicationLifetime.ApplicationStopping.Register(() =>
             {
                 Interlocked.Exchange(ref _applicationStopping, 1);
+                if (_environment.IsDrainOnApplicationStopping())
```
mathewc: nit: I'd rename this to DrainOnApplicationStoppingEnabled()

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
mathewc: Can you mark this code TEMP then, with a pointer to the work item https://github.com/Azure/azure-functions-host/issues/7188?

mathewc (approved):

PR #7270, opened 2021-04-04
Fixed ExtensionsController PUT request

### LINKED ISSUE #7266: Incorrect route registration may impact function routes
The following route will impact functions handling `PUT` requests in the latest host version due to a route precedence change:

https://github.com/Azure/azure-functions-host/blob/9c2ce4c108b08724b623df088d9ed4d232af116a/src/WebJobs.Script.WebHost/Controllers/ExtensionsController.cs#L62

The action must be updated to handle the correct route.

**Workaround**: Pin functions runtime version to : 3.0.15406 ([instructions](https://docs.microsoft.com/en-us/azure/azure-functions/set-runtime-version?tabs=portal#view-and-update-the-current-runtime-version))

yojagad:

resolves #7266

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/Controllers/ExtensionsController.cs
+++ b/src/WebJobs.Script.WebHost/Controllers/ExtensionsController.cs
@@ -60,8 +60,8 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Controllers
         public Task<IActionResult> Post([FromBody] ExtensionPackageReferenceWithActions package) => InstallExtension(package);
 
         [HttpPut]
-        [Route("admin/host/extensions/{id}")]
-        public Task<IActionResult> Put(int id, [FromBody]ExtensionPackageReferenceWithActions package) => InstallExtension(package);
+        [Route("admin/host/extensions")]
+        public Task<IActionResult> Put([FromBody]ExtensionPackageReferenceWithActions package) => InstallExtension(package);
 
         [HttpDelete]
         [Route("admin/host/extensions/{id}")]
--- a/src/WebJobs.Script.WebHost/Controllers/ExtensionsController.cs
+++ b/src/WebJobs.Script.WebHost/Controllers/ExtensionsController.cs
@@ -57,10 +57,10 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Controllers
 
         [HttpPost]
         [Route("admin/host/extensions")]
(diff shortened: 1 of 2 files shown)


src/WebJobs.Script.WebHost/Controllers/ExtensionsController.cs
```diff
@@ -57,10 +57,10 @@ public async Task<IActionResult> Get()
 
         [HttpPost]
         [Route("admin/host/extensions")]
-        public Task<IActionResult> Post([FromBody]ExtensionPackageReferenceWithActions package) => InstallExtension(package);
+        public Task<IActionResult> Post([FromBody] ExtensionPackageReferenceWithActions package) => InstallExtension(package);
 
-        [HttpPut("{id}")]
-        [Route("admin/host/extensions")]
+        [HttpPut]
+        [Route("admin/host/extensions/{id}")]
```
yojagad: @soninaren I vaguely remember portal using either this or the `POST` API above to install extensions, right? Do you know if this change will affect that in any way ?

yojagad: This `id` is not even being used by `InstallExtension`. Okay to remove? `ExtensionPackageReferenceWithActions` has a separate `id` being fed as input and that is what is used by `InstallExtension`

mathewc: Portal URI for extensions is [here](https://github.com/Azure/azure-functions-ux/blob/a0eec3ef46e5185a3c3539a327b83993354d4d59/client/src/app/shared/services/function-app.service.ts#L975). It's using the POST route which doesn't take an ID. However, if you query prod, you'll see some calls to the PUT route with an ID. Not clear if those are just testing.

However, we need to be sure that even though we're not using this ID for PUT, we have to make sure that we don't break the contract. Does the current route only match if an ID is specified? Or does it match both with/without? Need to test before changes and verify we don't change this.

yojagad: Confirmed that the above API can be called by either 

`PUT http://<appName>.azurewebsites.net/123`
or 
`PUT http://<appName>.azurewebsites.net/admin/host/extensions`

From logs, I don't see anything that's using the second API except the ones I've been testing. I just left the second API there and updated tests accordingly.

safihamid (approved):
:shipit:

mathewc (approved):

PR #7280, opened 2021-04-08
Caching blobs in shared memory for faster I/O

### LINKED ISSUE #7310: Caching of function bindings (input/output) for faster I/O
When a function accesses a remote object via bindings (e.g. blob from storage) or produces an output object, we should cache the object so that it can be served from memory the next time it is requested to improve performance (latency/throughput of fetching the object from remote storage).

Using the shared memory architecture implemented earlier to exchange data between the host and out-of-proc language workers (#6791), we can maintain a caching layer on top. Instead of freeing the shared memory resources after single-use, the caching layer can keep data around and provide a mechanism for the storage extension to query that before going to remote storage to retrieve an object. This will reduce the time taken to retrieve objects from remote storage, reduce traffic to upstream storage servers and perform better (perhaps more predictable) performance to the application.

n/a

#### Additional context
PRs for the caching layer:
#7280 
azure/azure-webjobs-sdk#2692
azure/azure-sdk-for-net#20209
azure/azure-functions-python-worker#844
#7757

PRs for shared memory communication between host and workers (completed):
#6791
#6836 
azure/azure-functions-python-worker#816

gohar94:

resolves #7310 

### Pull request checklist

### Additional information

Additional PR information


src/WebJobs.Script.Grpc/Extensions/ScriptInvocationContextExtensions.cs
```diff
@@ -149,6 +156,33 @@ private static bool ShouldSkipBindingData(KeyValuePair<string, object> bindingDa
...
+        /// <returns><see cref="true"/> if a conversion took place, <see cref="false"/> otherwise.</returns>
+        private static bool TryConvertObjectIfNeeded(object inputVal, out object convertedVal)
+        {
+            convertedVal = null;
+
+            if (inputVal is CacheableObjectStream stream)
+            {
+                using (MemoryStream ms = new MemoryStream())
+                {
+                    stream.CopyTo(ms);
```
goiri: This was a double operation right? First copying to memory and then copy to an array.
Did we have a more efficient way of just copying to the final array?

src/WebJobs.Script/Workers/FunctionDataCache/FunctionDataCache.cs
```diff
@@ -0,0 +1,329 @@
...
+        /// ensure the entry is safe to be removed.
+        /// </summary>
+        /// <param name="cacheKey">The <see cref="FunctionDataCacheKey"/> corresponding to the entry to be removed.</param>
+        /// <returns><see cref="true"/> if the entry was successfully removed, <see cref="false"/> if not.</returns>
+        public bool TryRemove(FunctionDataCacheKey cacheKey)
+        {
+            lock (_lock)
+            {
+                // TODO remove this Get and check if we can do remove with *out*
+                if (!TryGet(cacheKey, isIncrementActiveReference: false, out SharedMemoryMetadata sharedMemoryMeta))
```
ankitkumarr, line 167: I think calling `TryGet` here would deadlock as well. Because it won't be able to aquire a `_lock`

src/WebJobs.Script/Workers/FunctionDataCache/FunctionDataCache.cs
```diff
@@ -0,0 +1,329 @@
...
+                }
+            }
+
+            return true;
+        }
+
+        /// <summary>
+        /// Adds the given key to the end of the LRU list.
+        /// This key is the most recently used key.
+        /// Assumes that the key is not already present.
```
ankitkumarr: "Assumes that the key is not already present" is this true? As the function seems to take care of the scenario where `cacheKey` may already be in the `LRUList`

src/WebJobs.Script/Workers/FunctionDataCache/FunctionDataCache.cs
```diff
@@ -0,0 +1,331 @@
...
+                    if (_localCache.ContainsKey(cacheKey))
+                    {
+                        // Key already exists in the local cache; do not overwrite and don't
+                        // delete the existing data.
+                        isFailure = false;
+                        return false;
+                    }
+
+                    long bytesRequired = sharedMemoryMeta.Count;
+                    if (!EvictUntilCapacityAvailable(bytesRequired))
```
ankitkumarr, line 91: `EvictUntilCapacityAvailable(...)` calls `EvictOne()` until capacity is available, which calls `TryRemove(...)` which also requires the `_lock`. I think this may deadlock when cache is full.

You could probably do the same thing you did with `DecrementActiveReference` and `DecrementActiveReferenceCore` i.e. have lock in the public method, while logic in the private unsafe method.

src/WebJobs.Script/Workers/FunctionDataCache/FunctionDataCache.cs
```diff
@@ -0,0 +1,331 @@
...
+        /// <summary>
+        /// Gets the list maintaining the order of use of <see cref="FunctionDataCacheKey"/> in an LRU fashion.
+        /// The last element of the list is the most recently used item.
+        /// The first element of the list is the least recently used item.
+        /// </summary>
+        internal LinkedList<FunctionDataCacheKey> LRUList { get; private set; }
+
+        internal Dictionary<FunctionDataCacheKey, long> ActiveReferences { get; private set; }
+
+        public bool TryPut(FunctionDataCacheKey cacheKey, SharedMemoryMetadata sharedMemoryMeta, bool isIncrementActiveReference, bool isDeleteOnFailure)
```
ankitkumarr, line 72: Given that this would be a blocking call as each caller needs a lock, and if cache is full, we evict unbounded until enough space is available, I am wondering if there could be cases where --
1. First several (> 160,000) items cached are super small (Eg - 100 Bytes each) occupying almost 16 MB of the cache limit.
2. Another function runs that tries to cache say 1 file worth 15 MB.
3. In this case, we'd be blocking all executions until we clear about 150,000ish items in the cache.

This may not be a major issue as eviction seems cheap. But wanted to call out. Perhaps, if we see such issues, we could restrict the number of evictions per `TryPut`.

src/WebJobs.Script/Workers/FunctionDataCache/FunctionDataCache.cs
gohar94: This is a re-entrant lock, and since it's on the same object in both methods, it won't deadlock. 
However, for readability I can see if wrapping it into private methods is easier to understand.

src/WebJobs.Script/Workers/FunctionDataCache/FunctionDataCache.cs
gohar94: Same as above - it's re-entrant.

src/WebJobs.Script/Workers/FunctionDataCache/FunctionDataCache.cs
ankitkumarr: Ah you are right! I didn't know that the `lock` block makes it re-entrant. Thanks! That clears it.

src/WebJobs.Script/Workers/FunctionDataCache/FunctionDataCache.cs
gohar94: Good catch - removing this comment line.

src/WebJobs.Script/Workers/FunctionDataCache/FunctionDataCache.cs
gohar94: That's a fair point - although from the blob access analysis that we did earlier from production data, it seems a bit like an edge case. But yes, I will add logging here and once we roll this out, it would be good to see the number of evictions we have to do on average.

src/WebJobs.Script/Host/ScriptHost.cs
```diff
@@ -314,6 +322,32 @@ public async Task InitializeAsync(CancellationToken cancellationToken = default)
...
         }
 
+        /// <summary>
+        /// Checks if the conditions to use <see cref="IFunctionDataCache"/> are met (if a valid implementation was found at runtime,
+        /// if the setting was enabled, the app is using out-of-proc languages which communicate with the host over shared memory).
+        /// </summary>
+        /// <returns><see cref="true"/> if <see cref="IFunctionDataCache"/> can be used, <see cref="false"/> otherwise.</returns>
+        private bool GetIsFunctionDataCacheEnabled()
+        {
+            if (Utility.IsDotNetLanguageFunction(_workerRuntime) ||
```
soninaren, line 338: Nit: Can convert all of the if statements into a single return statement with `||`

goiri (dismissed):

src/WebJobs.Script.Grpc/Extensions/RpcSharedMemoryDataExtensions.cs
```diff
@@ -23,45 +24,124 @@ internal static async Task<RpcSharedMemory> ToRpcSharedMemoryAsync(this object v
                 return null;
             }
 
-            // Put the content into shared memory and get the name of the shared memory map written to
-            SharedMemoryMetadata putResponse = await sharedMemoryManager.PutObjectAsync(value);
-            if (putResponse == null)
+            SharedMemoryMetadata sharedMemoryMeta;
```
soninaren: nit: sharedMemoryMetadata

test/WebJobs.Script.Tests/Configuration/ScriptHostOptionsSetupTests.cs
```diff
@@ -4,9 +4,14 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.InteropServices;
```
soninaren: Change to this file can be reverted

src/WebJobs.Script/Workers/SharedMemoryDataTransfer/ISharedMemoryManager.cs
```diff
@@ -57,5 +57,13 @@ public interface ISharedMemoryManager : IDisposable
...
         /// <returns><see cref="true"/> if the shared memory map was freed, <see cref="false"/> otherwise.</returns>
         bool TryFreeSharedMemoryMap(string mapName);
+
+        /// <summary>
+        /// Track a particular shared memory map.
+        /// Used to hold a reference to a shared memory map so the OS does not clean it up.
+        /// </summary>
+        /// <param name="mapName">Name of the shared memory map.</param>
+        /// </summary>
+        bool TryTrackSharedMemoryMap(string mapName);
```
soninaren, line 67: Nit: `TryOperation()` style of method by convention has a out parameter that holds the result of the operation.  For example TryGetValue or TryParse.

src/WebJobs.Script/Host/ScriptHost.cs
```diff
@@ -314,6 +322,32 @@ public async Task InitializeAsync(CancellationToken cancellationToken = default)
...
+            {
+                return false;
+            }
+
+            if (_functionDataCache == null)
+            {
+                return false;
+            }
+
+            if (!_functionDataCache.IsEnabled)
```
soninaren: Can do this instead  `_functionDataCache.IsEnabled;`

src/WebJobs.Script.Grpc/Extensions/RpcSharedMemoryDataExtensions.cs
```diff
@@ -74,15 +154,16 @@ internal static async Task<object> ToObjectAsync(this RpcSharedMemory sharedMem,
...
-                return RpcDataType.Bytes;
-            }
-            else if (value is string)
+            switch (dataType)
             {
-                return RpcDataType.String;
+                case DataType.String:
+                    return RpcDataType.String;
+                case DataType.Binary:
+                    return RpcDataType.Bytes;
```
soninaren, line 164: can omit this return statement

src/WebJobs.Script/Workers/FunctionDataCache/FunctionDataCache.cs
```diff
@@ -0,0 +1,385 @@
...
+        /// <see cref="FunctionDataCacheConstants.FunctionDataCacheDefaultMaximumSizeBytes"/> is used.
+        /// </summary>
+        /// <param name="environment">Environment from which to get the configuration from.</param>
+        /// <returns>Number of bytes of capacity that the cache can have.</returns>
+        private long GetMaximumCapacityBytes(IEnvironment environment)
+        {
+            string capacityVal = environment.GetEnvironmentVariable(FunctionDataCacheConstants.FunctionDataCacheMaximumSizeBytesSettingName);
+            if (!string.IsNullOrEmpty(capacityVal))
+            {
+                if (long.TryParse(capacityVal, out long capacity))
```
soninaren: can remove this if by adding `&&` to line 256

soninaren (commented):
Some minor changes. Looks good otherwise

src/WebJobs.Script/Workers/SharedMemoryDataTransfer/ISharedMemoryManager.cs
gohar94: I agree but that's when the operation needs to return something. The primary reason for `Try*` is to tell that the method does not throw and returns a `bool` to indicate success/failure. In this case there is nothing to return but failure is a possibility and we don't want to throw, hence the use of `Try` in the name.

src/WebJobs.Script.Grpc/Extensions/RpcSharedMemoryDataExtensions.cs
```diff
@@ -23,45 +24,124 @@ internal static async Task<RpcSharedMemory> ToRpcSharedMemoryAsync(this object v
...
-            SharedMemoryMetadata putResponse = await sharedMemoryManager.PutObjectAsync(value);
-            if (putResponse == null)
+            SharedMemoryMetadata sharedMemoryMetadata;
+            bool needToFreeAfterInvocation = true;
+
+            // Check if the cache is being used or not.
+            // The binding extension will only hand out ICacheAwareReadObject if the FunctionDataCache is available and enabled.
+            if (value is ICacheAwareReadObject obj)
+            {
+                if (obj.IsCacheHit)
```
goiri, line 34: This is starting to look like a very long function.
Can we extract this into a function that is easier to read?

src/WebJobs.Script.Grpc/Extensions/RpcSharedMemoryDataExtensions.cs
```diff
@@ -23,45 +24,124 @@ internal static async Task<RpcSharedMemory> ToRpcSharedMemoryAsync(this object v
...
+                    case RpcDataType.Bytes:
+                    case RpcDataType.String:
+                        // This is where the SharedMemoryManager will hold a reference to the memory map
+                        if (sharedMemoryManager.TryTrackSharedMemoryMap(mapName))
+                        {
+                            return await sharedMemoryManager.GetObjectAsync(mapName, offset, count, typeof(SharedMemoryObject));
+                        }
+                        break;
+                    default:
+                        logger.LogError("Unsupported shared memory data type: {SharedMemDataType} with FunctionDataCache for invocation id: {Id}", sharedMem.Type, invocationId);
```
goiri, line 138: Do we have a test covering this?

src/WebJobs.Script.Grpc/Extensions/RpcSharedMemoryDataExtensions.cs
```diff
@@ -23,45 +24,124 @@ internal static async Task<RpcSharedMemory> ToRpcSharedMemoryAsync(this object v
...
+            if (isFunctionDataCacheEnabled)
+            {
+                switch (sharedMem.Type)
+                {
+                    case RpcDataType.Bytes:
+                    case RpcDataType.String:
+                        // This is where the SharedMemoryManager will hold a reference to the memory map
+                        if (sharedMemoryManager.TryTrackSharedMemoryMap(mapName))
+                        {
+                            return await sharedMemoryManager.GetObjectAsync(mapName, offset, count, typeof(SharedMemoryObject));
```
goiri, line 134: Should we log debug or this is too much info?

src/WebJobs.Script.Grpc/Extensions/RpcSharedMemoryDataExtensions.cs
```diff
@@ -74,18 +154,19 @@ internal static async Task<object> ToObjectAsync(this RpcSharedMemory sharedMem,
             }
         }
 
-        private static RpcDataType? GetRpcDataType(object value)
+        private static RpcDataType? GetRpcDataType(DataType dataType)
```
goiri, line 157: Should we make this part of RpcDataType or DataType?

src/WebJobs.Script.Grpc/Extensions/ScriptInvocationContextExtensions.cs
goiri: It may make sense to implement our own function that does all this.

src/WebJobs.Script/Binding/ExtensionBinding.cs
```diff
@@ -42,6 +44,28 @@ public override Collection<CustomAttributeBuilder> GetCustomAttributes(Type para
             return attributeBuilders;
         }
 
+        private async Task<bool> TryPrepareAttributesAndBindCacheAwareAsync(BindingContext context, FileAccess access)
```
goiri, line 51: add a doc explaining the purpose

src/WebJobs.Script/Binding/ExtensionBinding.cs
```diff
@@ -42,6 +44,28 @@ public override Collection<CustomAttributeBuilder> GetCustomAttributes(Type para
...
+        {
+            if (access == FileAccess.Write)
+            {
+                if (context.Value is SharedMemoryObject sharedMemoryObj)
+                {
+                    // First copy the attributes and then add a specific attribute (mapName) for the particular invocation.
+                    var currentAttributes = Attributes.ToList();
+                    currentAttributes.Add(new SharedMemoryAttribute(sharedMemoryObj.MemoryMapName, sharedMemoryObj.Count));
```
goiri: Extract this into variables so it is easier to read.

src/WebJobs.Script/Binding/FunctionBinding.cs
```diff
@@ -238,6 +238,45 @@ internal static async Task BindStreamAsync(BindingContext context, FileAccess ac
...
+                    throw new NotSupportedException($"Cannot perform cache-aware binding of write object with type: {context.Value.GetType()}");
+                }
+
+                context.Value = obj;
+            }
+            else if (access == FileAccess.Read)
+            {
+                ICacheAwareReadObject obj = await context.Binder.BindAsync<ICacheAwareReadObject>(context.Attributes);
+
+                // We get a binding to the object. It could either be read from the cache or from storage.
```
goiri, line 273: This kind of explanations could go into a doc for the whole BindCacheAwareAsync function.

src/WebJobs.Script/Binding/FunctionBinding.cs
```diff
@@ -247,6 +286,7 @@ public static void ConvertValueToStream(object value, Stream stream)
                 // to the stream
                 byte[] bytes = null;
                 Type type = value.GetType();
+
```
goiri: Avoid

src/WebJobs.Script/Workers/FunctionDataCache/FunctionDataCacheConstants.cs
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Script.Workers.FunctionDataCache
+{
+    internal static class FunctionDataCacheConstants
+    {
+        public const string FunctionDataCacheEnabledSettingName = "FUNCTIONS_DATA_CACHE_ENABLED";
+        public const string FunctionDataCacheMaximumSizeBytesSettingName = "FUNCTIONS_DATA_CACHE_MAXIMUM_SIZE_BYTES";
+
+        public const long FunctionDataCacheDefaultMaximumSizeBytes = 16 * 1024 * 1024; // 16 MB
```
goiri, line 15: Explain why 16MB?

src/WebJobs.Script/Workers/SharedMemoryDataTransfer/SharedMemoryMap.cs
```diff
@@ -32,20 +32,15 @@ public class SharedMemoryMap : IDisposable
...
-            {
-                throw new ArgumentNullException(nameof(memoryMappedFile));
-            }
-
             if (string.IsNullOrEmpty(mapName))
             {
                 throw new ArgumentException(nameof(mapName));
             }
 
+            _memoryMappedFile = memoryMappedFile ?? throw new ArgumentNullException(nameof(memoryMappedFile));
```
goiri, line 40: I kind of prefer the old check and assignment.

src/WebJobs.Script.Grpc/Extensions/RpcSharedMemoryDataExtensions.cs
gohar94: We need to be very conservative with log outputs so I'd skip this one.

src/WebJobs.Script.Grpc/Extensions/RpcSharedMemoryDataExtensions.cs
gohar94: `DataType` is an enum so can't do it for that. `RpcDataType` is a protobuf class so we will need to have an `*Extensions` C# class just for this conversion; overkill I think.

src/WebJobs.Script.Grpc/Extensions/RpcSharedMemoryDataExtensions.cs
gohar94: It does a bunch of operations which are kind of interleaved so it will be hard and not too clean - we have discussed this the last time we did a review of this too, IIRC. Let's keep it as is.

src/WebJobs.Script/Workers/SharedMemoryDataTransfer/SharedMemoryMap.cs
gohar94: This is the "C#" way. :)

goiri (dismissed):

soninaren (approved):

soninaren (dismissed):
Would need to update the webjobs sdk package reference before we can merge this.

gohar94:
> Would need to update the webjobs sdk package reference before we can merge this.

Correct - we need a green build here. Do we need to bump up versions in webjobs-sdk?

soninaren (approved):

PR #7295, opened 2021-04-15
Fix check for IsFileSystemReadOnly

### LINKED ISSUE #7293: Fix check for IsFileSystemReadOnly
Today, [this check](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Config/HostJsonFileConfigurationSource.cs#L260) for IsFileSystemReadOnly will always return true if `SCM_RUN_FROM_PACKAGE` app setting is present. This app setting is always present for Linux Consumption deployments, even if it's not used, so there can be a situation where we are not using `SCM_RUN_FROM_PACKAGE` on Linux Consumption (e.g. portal deployment/editing) and want the filesystem to be writable, so the logic for the check needs to be updated.

Impact:
1) Extension bundles are not loaded because the host cannot modify host.json and write the appropriate bundle.
2) File watching is disabled if the filesystem is read-only, causing slower function app updates.

pgombar:

Update logic to determine if the filesystem is read-only. It's not sufficient to check if `SCM_RUN_FROM_PACKAGE` is set, the URI also needs to point to an existing blob to deduce it's a zip deployment (therefore read-only).

The changes in this PR are:
1) Move the `IsFileSystemReadOnly` to `ScriptApplicationHostOptions` from `EnvironmentExtensions`
2) Add a special layer of options setup only used in Linux Consumption, `LinuxScriptApplicationHostOptionsSetup` which contains the logic to check for blob existence

resolves #7293

### Pull request checklist

### Additional information

Additional PR information


pgombar:
Hi @fabiocav @brettsam, would appreciate your input if this is the right design approach. The requirement is to keep some state (`CacheIfBlobExists()` that is called from `IsFileSystemReadOnly()`) that needs to be accessible from both `WebHost` and `ScriptHost`. For that, I choose to add a new service, `FileSystemManager` that I registered as a common service.

Please let me know if you think there's a better/neater way to achieve the same! Thanks!

pragnagopa:
@fabiocav  / @brettsam  - this PR updates how we check for `IsFileSystemReadOnly` - this is used in multiple places. 

Do you recommend exposing  `IsFileSystemReadOnly` as property in `ScriptApplicationHostOptions` similar to https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Config/ScriptApplicationHostOptions.cs#L15 ?

src/WebJobs.Script.WebHost/Configuration/ScriptApplicationHostOptionsSetup.cs
```diff
@@ -61,6 +69,69 @@ public void Configure(string name, ScriptApplicationHostOptions options)
...
+        {
+            var blobExists = _cloudBlockBlobHelperService.BlobExists(_environment.GetEnvironmentVariable(ScmRunFromPackage), EnvironmentSettingNames.ScmRunFromPackage, _logger).GetAwaiter().GetResult();
+            _logger.LogInformation($"Checked if ${EnvironmentSettingNames.ScmRunFromPackage} points to an existing blob: ${blobExists}");
+            return blobExists;
+        }
+
+        private bool IsZipDeployment(ScriptApplicationHostOptions options)
+        {
+            // If the app is using the new app settings for zip deployment, we don't need to check further, it must be a zip deployment.
+            bool usesNewZipDeployAppSettings = IsValidZipSetting(_environment.GetEnvironmentVariable(AzureWebsiteZipDeployment)) ||
```
balag0: >usesNewZipDeployAppSettings [](http://example.com/codeflow?start=17&length=27)

runFromPkgConfigured

(these settings are not newer than scm_run_..)

src/WebJobs.Script.WebHost/Configuration/ScriptApplicationHostOptionsSetup.cs
```diff
@@ -61,6 +69,69 @@ public void Configure(string name, ScriptApplicationHostOptions options)
                 options.IsSelfHost = options.IsSelfHost;
                 options.IsStandbyConfiguration = true;
             }
+
+            options.AreZipDeploymentAppSettingsValid = ValidateZipDeploymentAppSettings();
+            options.ScmRunFromPackageBlobExists = BlobExists();
```
balag0: >BlobExists(); [](http://example.com/codeflow?start=50&length=13)

Since BlobExists() needs to be called only if SCM_RUN_FROM_PACKAGE is present, Set it to false here and update later on.

src/WebJobs.Script.WebHost/Configuration/ScriptApplicationHostOptionsSetup.cs
```diff
@@ -61,6 +69,69 @@ public void Configure(string name, ScriptApplicationHostOptions options)
...
+            // If not, this is not a zip deployment. If yes, we still need to check if the blob exists,
+            // since this app setting is always created for Linux Consumption, even when it is not used.
+            // In portal deployment, SCM_RUN_FROM_PACKAGE will be set, but we will be using Azure Files instead.
+            if (!IsValidZipUrl(_environment.GetEnvironmentVariable(ScmRunFromPackage)))
+            {
+                return false;
+            }
+
+            // If SCM_RUN_FROM_PACKAGE is a valid app setting and Azure Files are not being used, it must be a zip deployment.
+            if (string.IsNullOrEmpty(_environment.GetEnvironmentVariable(AzureFilesConnectionString)) &&
```
balag0: this check seems new. does it matter if AzureFilesConnectionString is set or not when deciding if the app is using ZipDeployment?

Will this work?

bool IsZipDeployment(bool scmRunFromPackageBlobExists)
{
scmRunFromPackageBlobExists = false;
bool runFromPkgConfigured = IsValidZipSetting(_environment.GetEnvironmentVariable(AzureWebsiteZipDeployment)) ||
                IsValidZipSetting(_environment.GetEnvironmentVariable(AzureWebsiteAltZipDeployment)) ||
                IsValidZipSetting(_environment.GetEnvironmentVariable(AzureWebsiteRunFromPackage));

   if (runFromPkgConfigured) {return true;}

bool scmRunFromPkgConfigured = IsValidZipSetting(_environment.GetEnvironmentVariable(ScmRunFromPackage)) && BlobExists(_environment.GetEnvironmentVariable(ScmRunFromPackage);

scmRunFromPackageBlobExists = scmRunFromPkgConfigured;

return scmRunFromPkgConfigured;
}

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -169,7 +169,7 @@ public async Task StartAsync(CancellationToken cancellationToken)
         private void CheckFileSystem()
         {
             // Shutdown if RunFromZipFailed
-            if (_environment.IsZipDeployment(validate: false))
+            if (_applicationHostOptions.CurrentValue.AreZipDeploymentAppSettingsValid)
```
balag0: I think checking if the appsettings are not empty can be done in _environment. We can then remove AreZipDeploymentAppSettingsValid from ScriptApplicationHostOptions

src/WebJobs.Script.WebHost/Configuration/ScriptApplicationHostOptionsSetup.cs
```diff
@@ -61,6 +69,69 @@ public void Configure(string name, ScriptApplicationHostOptions options)
...
+                string.IsNullOrEmpty(_environment.GetEnvironmentVariable(AzureFilesContentShare)))
+            {
+                return true;
+            }
+
+            // SCM_RUN_FROM_PACKAGE is set, as well as Azure Files app settings, so we need to check if we are actually using the zip blob.
+            return options.ScmRunFromPackageBlobExists;
+        }
+
+        private bool ValidateZipDeploymentAppSettings()
```
balag0: ValidateZipDeploymentAppSettings can be moved to Environment since it is just checking the settings are not empty?

pgombar: Good point, will do that.

src/WebJobs.Script.WebHost/Configuration/ScriptApplicationHostOptionsSetup.cs
pgombar: From my understanding, the assignment on L74 happens whenever we call configuration bind on options, which is on startup. So, if I set it to false here, it will remain false until it is actually updated by the codepath that calls `IsZipDeployment` or `IsFileSystemReadOnly`.

However, in `RunFromPackageContext#IsRunFromPackage` we directly check against `options.ScmRunFromPackageBlobExists`. That means we implicitly rely on having a `IsZipDeployment` / `IsFileSystemReadOnly` call before specialization (otherwise it will just always be false), which I'm not sure happens today.

src/WebJobs.Script.WebHost/Configuration/ScriptApplicationHostOptionsSetup.cs
pgombar: > does it matter if AzureFilesConnectionString is set or not when deciding if the app is using ZipDeployment?

You have a point, it should not matter. My check is weaker than yours; I say it's zip deployment if there's absence of other app settings (Azure Files), I don't even bother checking if the blob exists (which we should, just to be safe).

I'll update the check to be explicit like yours; only if the app setting `SCM_RUN_FROM_PACKAGE` is valid *and* the blob exists, it means we are using zip deployment.

src/WebJobs.Script.WebHost/Configuration/ScriptApplicationHostOptionsSetup.cs
pgombar: Maybe I can delay the assignment using a Lazy bool? cc @brettsam can you remind me if this was one of the possibilities you suggested?

src/WebJobs.Script.WebHost/Models/RunFromPackageContext.cs
```diff
@@ -31,14 +26,19 @@ public RunFromPackageContext(string envVarName, string url, long? packageContent
...
+            return (IsScmRunFromPackage() && ScmRunFromPackageBlobExists(options, logger)) || (!IsScmRunFromPackage() && !string.IsNullOrEmpty(Url) && Url != "1");
         }
 
-        public async Task<bool> IsRunFromPackage(ILogger logger)
+        private bool ScmRunFromPackageBlobExists(ScriptApplicationHostOptions options, ILogger logger)
         {
-            return (IsScmRunFromPackage() && await _runFromPackageCloudBlockBlobService.BlobExists(Url, EnvironmentVariableName, logger)) ||
-                   (!IsScmRunFromPackage() && !string.IsNullOrEmpty(Url) && Url != "1");
+            var blobExists = options.ScmRunFromPackageBlobExists;
+            logger.LogInformation($"Checking if {EnvironmentSettingNames.ScmRunFromPackage} points to an existing blob: {blobExists}");
```
pragnagopa: change to logDebug

src/WebJobs.Script/Config/ScriptApplicationHostOptions.cs
```diff
@@ -31,5 +31,11 @@ public class ScriptApplicationHostOptions
         public IServiceProvider RootServiceProvider { get; set; }
 
         public bool IsStandbyConfiguration { get; internal set; }
+
+        public bool IsFileSystemReadOnly { get; set; }
+
+        public bool IsZipDeployment { get; set; }
```
pragnagopa: `IsZipDeployment` is only used in `ScriptApplicationHostOptionsSetup` - no need to expose it here. Calculate directly in `ScriptApplicationHostOptionsSetup`

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -78,34 +78,12 @@ public static bool IsRemoteDebuggingEnabled(this IEnvironment environment)
             return !string.IsNullOrEmpty(environment.GetEnvironmentVariable(RemoteDebuggingPort));
         }
 
-        public static bool IsZipDeployment(this IEnvironment environment, bool validate = true)
+        public static bool AreZipDeploymentAppSettingsValid(this IEnvironment environment)
```
pragnagopa: rename to `ZipDeploymentAppSettingsExist`

test/WebJobs.Script.Tests/ManagedDependencies/ManagedDependencyOptionsSetupTest.cs
```diff
@@ -1,14 +1,11 @@
...
 
-using System;
 using System.IO;
-using System.Linq;
 using Microsoft.Azure.WebJobs.Script.Configuration;
 using Microsoft.Azure.WebJobs.Script.ManagedDependencies;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
-using Microsoft.Extensions.Options;
 using Microsoft.WebJobs.Script.Tests;
```
pragnagopa: undo to scope the PR changes

test/WebJobs.Script.Tests/FileProvisioning/FuncAppFileProvisioningServiceTests.cs
```diff
@@ -8,7 +8,6 @@
 using Microsoft.Azure.WebJobs.Script.FileProvisioning;
 using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
 using Microsoft.Extensions.Logging;
-using Microsoft.Extensions.Logging.Abstractions;
 using Microsoft.Extensions.Options;
```
pragnagopa: undo to scope the PR changes

pragnagopa (commented):
Added few comments

src/WebJobs.Script/CloudBlockBlobHelperService.cs
```diff
@@ -4,19 +4,22 @@
...
         {
-            return await BlobExistsAsync(url, environmentVariableName, logger);
+            return await BlobExistsAsync(url);
         }
 
-        private static async Task<bool> BlobExistsAsync(string url, string environmentVariableName, ILogger logger)
+        private static async Task<bool> BlobExistsAsync(string url)
         {
+            if (string.IsNullOrEmpty(url))
+            {
```
pragnagopa: move this check to public method

src/WebJobs.Script/CloudBlockBlobHelperService.cs
```diff
@@ -4,19 +4,22 @@
...
+namespace Microsoft.Azure.WebJobs.Script
 {
-    public class RunFromPackageCloudBlockBlobService
+    public class CloudBlockBlobHelperService
     {
-        public virtual async Task<bool> BlobExists(string url, string environmentVariableName, ILogger logger)
+        public virtual async Task<bool> BlobExists(string url)
         {
-            return await BlobExistsAsync(url, environmentVariableName, logger);
+            return await BlobExistsAsync(url);
```
pragnagopa: change to `return BlobExistsAsync`

pragnagopa (commented):
Add test in `StandbyManagerE2ETests_Linux.cs` - to verify isFileSystemReadOnly options are refreshed after specialization.

pragnagopa:
@balag0 - Do we have a good way to verify changes E2E using mesh image and triggering specialization?

src/WebJobs.Script/Config/ScriptApplicationHostOptions.cs
pgombar: Thanks, good catch. Due to the refactoring, `IsZipDeployment` doesn't need to be exposed anymore. Updated!

src/WebJobs.Script.WebHost/Configuration/ScriptApplicationHostOptionsSetup.cs
balag0: Sorry I meant still doing in the same method, but not doing it before we check if the app is using WEBSITE_RUN_FROM_PACKAGE or not.

something like

Instead of 
options.ScmRunFromPackageBlobExists = BlobExists();
options.IsZipDeployment = IsZipDeployment(options);

change to

options.IsZipDeployment = IsZipDeployment(options, options.ScmRunFromPackageBlobExists);

```
bool IsZipDeployment(bool scmRunFromPackageBlobExists)
{
scmRunFromPackageBlobExists = false;
bool runFromPkgConfigured = IsValidZipSetting(_environment.GetEnvironmentVariable(AzureWebsiteZipDeployment)) ||
                IsValidZipSetting(_environment.GetEnvironmentVariable(AzureWebsiteAltZipDeployment)) ||
                IsValidZipSetting(_environment.GetEnvironmentVariable(AzureWebsiteRunFromPackage));

   if (runFromPkgConfigured) {return true;}

bool scmRunFromPkgConfigured = IsValidZipSetting(_environment.GetEnvironmentVariable(ScmRunFromPackage)) && BlobExists(_environment.GetEnvironmentVariable(ScmRunFromPackage);

scmRunFromPackageBlobExists = scmRunFromPkgConfigured;

return scmRunFromPkgConfigured;
}
```

Basically BloblExists()  needs to be called only if runFromPkgConfigured is false. Doesn't necessarily have to use the same pattern :)

pgombar: Thanks for clarifying, I understand now what you meant. Agreed, this is better!

src/WebJobs.Script/CloudBlockBlobHelperService.cs
pgombar: Sorry, not sure I understand your comment?

src/WebJobs.Script/CloudBlockBlobHelperService.cs
pgombar: BlobExists is async so without await this results in a compiler error

pragnagopa: remove async from the method signature :)

src/WebJobs.Script/CloudBlockBlobHelperService.cs
pgombar: Oh, you mean the public method above? Sure, let me do that.

src/WebJobs.Script/CloudBlockBlobHelperService.cs
pgombar: But it is still an async method? We have:
`await Utility.InvokeWithRetriesAsync(async () => (...)`

pragnagopa: change to :

```cs
public virtual Task<bool> BlobExists(string url)
        {
            if (string.IsNullOrEmpty(url))
            {
                return Task.FromResult(false);
            }
            return BlobExistsAsync(url);
        }
```

Its good to avoid `async .. await` when result is not needed and is simple used to return the result.

test/WebJobs.Script.Tests/ManagedDependencies/ManagedDependencyOptionsSetupTest.cs
```diff
@@ -1,9 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
-using System;
 using System.IO;
-using System.Linq;
 using Microsoft.Azure.WebJobs.Script.Configuration;
```
pragnagopa: nit: undo edits in this file

pragnagopa (dismissed):
💯

src/WebJobs.Script/CloudBlockBlobHelperService.cs
pgombar: Didn't know that, thanks for letting me know! Updated.

pragnagopa (dismissed):

src/WebJobs.Script.WebHost/Configuration/ScriptApplicationHostOptionsSetup.cs
```diff
@@ -61,6 +67,44 @@ public void Configure(string name, ScriptApplicationHostOptions options)
                 options.IsSelfHost = options.IsSelfHost;
                 options.IsStandbyConfiguration = true;
             }
+
+            options.IsFileSystemReadOnly = IsZipDeployment(options);
+        }
+
+        private bool BlobExists()
+        {
+            return _cloudBlockBlobHelperService.BlobExists(_environment.GetEnvironmentVariable(ScmRunFromPackage)).GetAwaiter().GetResult();
```
mathewc: We shouldn't be doing sync over async like this. I don't understand why we need to be checking for actual blob existence like this, particularly on startup which will affect coldstart.

Why isn't the existing EnvironmentSettings.IsZipDeployment sufficient? We should be able to do this as a static environment check as we do currently, perhaps with some small tweaks if necessary to the existing function.

pragnagopa: Copying from PR description:

**It's not sufficient to check if SCM_RUN_FROM_PACKAGE is set, the URI also needs to point to an existing blob to deduce it's a zip deployment (therefore read-only).** 

This is an unfortunate limitation as of today - as this app setting is needed for Kudu deployments.

pgombar: >  particularly on startup which will affect coldstart.

Some things to note:
1) In prod today, we already have a check for blob existence during specialization. It happens in [this call](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.WebHost/Management/InstanceManager.cs#L305).
2) The newly added check will only happen once and be cached, so the check in 1) won't be calling storage again.
3) The call for actual blob existence only happens if `SCM_RUN_FROM_PACKAGE` is present, which is true for Linux Consumption.

mathewc: That check only happens during Linux consumption specialization, but the change you're making here will be for all platforms. Problems with the above call aside, I still don't see why what we have now isn't sufficient. The existing IsZipDeployment logic will only return true if the env var is present AND it's a URI format or 1, which is correct. Can you update the bug with details on when the existing check doesn't work? I.e. what value do the settings have.

balag0: @mathewc BlobExists() will only be called if
IsValidZipSetting(_environment.GetEnvironmentVariable(ScmRunFromPackage) is true

since this will be true only on Linux consumption we are still be retaining the old behavior

On Linux consumption ScmRunFromPackage will always point to an external url. But preference is given to WEBSITE_RUN_FROM_PACAKGE if present.

So currently:
BlobExists() will be called only on LinuxConsumption and only if WEBSITE_RUN_FROM_PACAKGE is empty

With this change as well I think we have the same behavior.

mathewc: I see. I was unaware we had that logic already and that you're moving things around/fixing. Let's at least make the call sync here then, to avoid the sync over async ani-pattern which has caused us problems in the past (more info [here](https://devblogs.microsoft.com/pfxteam/should-i-expose-synchronous-wrappers-for-asynchronous-methods/)).

src/WebJobs.Script.WebHost/Configuration/ScriptApplicationHostOptionsSetup.cs
```diff
@@ -61,6 +67,44 @@ public void Configure(string name, ScriptApplicationHostOptions options)
...
+        }
+
+        private bool IsZipDeployment(ScriptApplicationHostOptions options)
+        {
+            // If the app is using app settings for run from package, we don't need to check further, it must be a zip deployment.
+            bool runFromPkgConfigured = IsValidZipSetting(_environment.GetEnvironmentVariable(AzureWebsiteZipDeployment)) ||
```
mathewc: Can we avoid duplicating these settings here, perhaps by leveraging you refactored utility method, so we could call _environment.IsRunFromZipOrPackage() here so the logic is centralized? I'd leave the IsValidZipSetting helper and checks in the Utility so IsRunFromZipOrPackage does those checks as it does now.

src/WebJobs.Script.WebHost/Models/RunFromPackageContext.cs
```diff
@@ -31,14 +26,19 @@ public RunFromPackageContext(string envVarName, string url, long? packageContent
...
+            return (IsScmRunFromPackage() && ScmRunFromPackageBlobExists(options, logger)) || (!IsScmRunFromPackage() && !string.IsNullOrEmpty(Url) && Url != "1");
         }
 
-        public async Task<bool> IsRunFromPackage(ILogger logger)
+        private bool ScmRunFromPackageBlobExists(ScriptApplicationHostOptions options, ILogger logger)
         {
-            return (IsScmRunFromPackage() && await _runFromPackageCloudBlockBlobService.BlobExists(Url, EnvironmentVariableName, logger)) ||
-                   (!IsScmRunFromPackage() && !string.IsNullOrEmpty(Url) && Url != "1");
+            var blobExists = options.ScmRunFromPackageBlobExists;
+            logger.LogDebug($"Checking if {EnvironmentSettingNames.ScmRunFromPackage} points to an existing blob: {blobExists}");
```
mathewc: nit: I wonder if we shouldn't have this log where we're actually doing the existence check, since it seems to imply we're doing it at this point when we're not.

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -169,7 +169,7 @@ public async Task StartAsync(CancellationToken cancellationToken)
         private void CheckFileSystem()
         {
             // Shutdown if RunFromZipFailed
-            if (_environment.IsZipDeployment(validate: false))
+            if (_environment.ZipDeploymentAppSettingsExist())
```
mathewc, line 172: As noted above, to avoid impacting existing usages like this, it'd be good if the IsRunFromZipOrPackage (or whatever we're renaming it to - maybe we leave the name as it was before?) continued to do the IsValidZipSetting checks that it did before

src/WebJobs.Script/Config/ScriptApplicationHostOptions.cs
```diff
@@ -31,5 +31,9 @@ public class ScriptApplicationHostOptions
         public IServiceProvider RootServiceProvider { get; set; }
 
         public bool IsStandbyConfiguration { get; internal set; }
+
+        public bool IsFileSystemReadOnly { get; set; }
+
+        public bool ScmRunFromPackageBlobExists { get; set; }
```
mathewc: I wonder if we shouldn't rename this to IsSCMRunFromPackage? Whe this is set, we've checked that the app setting is configured and the blob exists, so we know at that point that it's a valid SCM RFP scenario?

src/WebJobs.Script.WebHost/Models/RunFromPackageContext.cs
```diff
@@ -31,14 +26,19 @@ public RunFromPackageContext(string envVarName, string url, long? packageContent
...
         public bool IsScmRunFromPackage()
         {
-            return string.Equals(EnvironmentVariableName, EnvironmentSettingNames.ScmRunFromPackage,
-                        StringComparison.OrdinalIgnoreCase);
+            return string.Equals(EnvironmentVariableName, EnvironmentSettingNames.ScmRunFromPackage, StringComparison.OrdinalIgnoreCase);
+        }
+
+        public bool IsRunFromPackage(ScriptApplicationHostOptions options, ILogger logger)
+        {
+            return (IsScmRunFromPackage() && ScmRunFromPackageBlobExists(options, logger)) || (!IsScmRunFromPackage() && !string.IsNullOrEmpty(Url) && Url != "1");
```
mathewc, line 34: Can't we skip the IsScmRunFromPackage checks here. When you set the new options member you already check for settings and blob existence. Seems like this logic can be simplified.

src/WebJobs.Script/CloudBlockBlobHelperService.cs
```diff
@@ -25,18 +28,12 @@ private static async Task<bool> BlobExistsAsync(string url, string environmentVa
                     CloudBlockBlob blob = new CloudBlockBlob(new Uri(url));
```
brettsam: I think everything accessing this is synchronous -- can we use `Exists()` here and make everything sync so we don't do any `GetAwaiter()` stuff?

src/WebJobs.Script/CloudBlockBlobHelperService.cs
```diff
@@ -25,18 +28,12 @@ private static async Task<bool> BlobExistsAsync(string url, string environmentVa
...
                     exists = await blob.ExistsAsync();
                 }
-                catch (Exception e)
+                catch (Exception)
```
mathewc: We should probably improve this logic. As it stands, if the URL is invalid or doesn't point to a blob, we'll still do retries (e.g. if new Uri(url) throws). Instead, I think the intent is to retry only for transient errors arising from the Existence check. So we should only do retries if the latter part throws.

brettsam (commented):
This looks good to me after @mathewc comments are addressed.

src/WebJobs.Script.WebHost/Models/RunFromPackageContext.cs
pgombar: I agree that the message is misleading and could be improved. I decided to log the value of blob existence here because adding the logger to `ScriptApplicationHostOptionsSetup` wasn't an option; it introduced recursive dependencies.

Let me refactor this class by simplifying the checks like you're suggesting in the comment above and improve the log message as well.

src/WebJobs.Script.WebHost/Models/RunFromPackageContext.cs
pgombar: I think you're right, they're duplicate checks. cc @balag0 just to confirm there's no issues with this approach.

The new method can just be:
```
public bool IsRunFromPackage(ScriptApplicationHostOptions options, ILogger logger)
{
            var blobExists = options.ScmRunFromPackageBlobExists;
            logger.LogDebug($"{EnvironmentSettingNames.ScmRunFromPackage} points to an existing blob: {blobExists}");
}
```

pgombar: Ah, as I typed this comment I realized this is not sufficient. `RunFromPackageContext` includes other types of run from package, like from these app settings:
```
public const string AzureWebsiteZipDeployment = "WEBSITE_USE_ZIP";
public const string AzureWebsiteAltZipDeployment = "WEBSITE_RUN_FROM_ZIP";
public const string AzureWebsiteRunFromPackage = "WEBSITE_RUN_FROM_PACKAGE";
```

Which is why we have this part of the check: `|| (!IsScmRunFromPackage() && !string.IsNullOrEmpty(Url) && Url != "1");`

src/WebJobs.Script/Config/ScriptApplicationHostOptions.cs
pgombar: That's correct and a better name for it, thanks! Updating.

src/WebJobs.Script.WebHost/Models/RunFromPackageContext.cs
pgombar: I've updated the message to omit the "Checking if" part. Please let me know if you have a better idea of where/how we can log when we're checking if the blob exists.

src/WebJobs.Script.WebHost/Configuration/ScriptApplicationHostOptionsSetup.cs
pgombar: Before my changes, `_environment.IsZipDeployment(bool validate = true)` had two checks.
1) If `validate=true` then we utilize `IsValidZipSetting` to check the app settings.
2) If `validate=false` (this happens only in a single call in `WebJobsScriptHostService`) then we utilize `!string.IsNullOrEmpty` to check the app settings.

My change refactored 1) into this method in Options and 2) remained in environment. They are different checks, so there's no code duplication.

Can you please clarify what you are suggesting? I'm not sure I understand what you want to live in environment as opposed to here. Thanks :)

src/WebJobs.Script/CloudBlockBlobHelperService.cs
pgombar: Good catch, updating.

src/WebJobs.Script.WebHost/Configuration/ScriptApplicationHostOptionsSetup.cs
pgombar: Changed `BlobExists` to sync, thanks for the suggestions.

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
pgombar: Please note that, before my change, the call was IsZipDeployment(**validate: false**) which meant that the check was doing !string.IsNullOrEmpty, not IsValidZipSetting

balag0 (dismissed):
:shipit:

src/WebJobs.Script/CloudBlockBlobHelperService.cs
```diff
@@ -0,0 +1,47 @@
...
+                    {
+                        return blob.Exists();
+                    }
+                    catch (Exception ex) when (!ex.IsFatal())
+                    {
+                        if (++attempt > 2)
+                        {
+                            throw;
+                        }
+                        Thread.Sleep(TimeSpan.FromSeconds(0.3));
```
pgombar: @mathewc by switching this check to sync I am using Thread.Sleep to sleep between retries. Any concerns with this or suggestions around it?

Before the review, this was async and the caller had `GetAwaiter().GetResult()` because we can't make it async all the way to Startup.

fabiocav:
@pgombar a reminder, just in case. Can you please make sure you squash the commits prior to merging? Thanks!

src/WebJobs.Script.WebHost/Configuration/ScriptApplicationHostOptionsSetup.cs
mathewc: I'm just trying to avoid having two disjointed blocks of code where we're maintaining the list of 3 setting we have to check. We centralized that in one place to ensure when we modify, it's all done in one place so people do it correctly. So I'm suggesting to retain the previous helper as it was, with any modifications you need to make for ScmRunFromPackage.

fabiocav:
Haven't had a chance to review this in detail, but given the code being touched, I wanted to make sure this was validated during specialization and you have the expected behavior with placeholder/specialized sites

src/WebJobs.Script.WebHost/Models/RunFromPackageContext.cs
mathewc: I was just wondering why your check couldn't be like:

```csharp
return options.IsScmRunFromPackage || (!IsScmRunFromPackage() && !string.IsNullOrEmpty(Url) && Url != "1");
```

Anyhow, if Bala is happy with the logic, I'm ok I guess

src/WebJobs.Script/CloudBlockBlobHelperService.cs
```diff
@@ -0,0 +1,47 @@
...
+                    }
+                }
+            }
+            catch (Exception)
```
fabiocav: we're not doing anything with this exception. Let's remove the try catch here.

src/WebJobs.Script.WebHost/Models/RunFromPackageContext.cs
balag0: Yes. this should also work and is better. I misunderstood the comment earlier and I just wanted to make sure we retain the 
(!IsScmRunFromPackage() && !string.IsNullOrEmpty(Url) && Url != "1") part of it :)

src/WebJobs.Script.WebHost/Configuration/ScriptApplicationHostOptionsSetup.cs
pgombar: > I'm just trying to avoid having two disjointed blocks of code where we're maintaining the list of 3 setting we have to check. 

Actually, there is a slight difference in the checks in the previous helper in Environment and the checks I added here. The difference is that, here, we omit checking for `!string.IsNullOrEmpty(environment.GetEnvironmentVariable(ScmRunFromPackage))`, so I can't use the previous helper here.

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
pgombar: Answered above, the previous check doesn't fit with the new requirements.

fabiocav (approved):

src/WebJobs.Script.WebHost/Configuration/LinuxScriptApplicationHostOptionsSetup.cs
```diff
@@ -0,0 +1,92 @@
...
+                    {
+                        return blob.Exists();
+                    }
+                    catch (Exception ex) when (!ex.IsFatal())
+                    {
+                        if (++attempt > 2)
+                        {
+                            return false;
+                        }
+                        Thread.Sleep(TimeSpan.FromSeconds(0.3));
```
pragnagopa, line 74: decided to keep Thread.Sleep?

pragnagopa (approved):
LGTM!

@fabiocav  / @brettsam  / @mathewc

src/WebJobs.Script.WebHost/Configuration/LinuxScriptApplicationHostOptionsSetup.cs
pgombar: There wasn't any strong push against it, so I'd prefer keeping it to minimize changes from the current behaviour.

pgombar:
Done with validation for specialization. With these changes, the filesystem read-only check is fixed!

PR #7306, opened 2021-04-21
Defer Script Host disposal during cold start

### LINKED ISSUE #7305: Defer Script Host disposal during cold start
One of the steps during specialization is to restart host. This restart involves starting a new host and disposing the old host.
https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs#L483

Disposing old host impacts cold starts at higher percentiles under some specific ordering of code execution.

balag0:
Defer Script Host disposal during cold start

One of the steps during specialization is to restart host. This restart involves starting a new host and disposing the old host.

From profiling, disposal of the old host is the most expensive part in the restart operation. On a single core machine, while the restart is in progress, function invocation will be delayed during cold start. On most cold starts the host dispose seems to be getting scheduled only after cold start has completed so this doesn't impact cold start at lower percentiles. But if the ordering of execution happened to be reversed, the host dispose will increase cold start.

Fix is to defer the Host disposal by 5 seconds. The change reduces P90 cold start by 60%.

resolves https://github.com/Azure/azure-functions-host/issues/7305

### Pull request checklist

* [ X] My changes **do not** need to be backported to a previous version

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
+++ b/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
@@ -656,6 +656,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
         /// <param name="instance">The <see cref="IHost"/> instance to remove</param>
         private async Task Orphan(IHost instance, CancellationToken cancellationToken = default)
         {
+            var isStandbyHost = false;
             try
             {
                 var scriptHost = (ScriptHost)instance?.Services.GetService<ScriptHost>();
@@ -663,6 +664,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
                 {
                     scriptHost.HostInitializing -= OnHostInitializing;
                     scriptHost.HostInitialized -= OnHostInitialized;
+                    isStandbyHost = scriptHost.IsStandbyHost;
                 }
             }
             catch (ObjectDisposedException)
@@ -684,7 +686,32 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
                 if (instance != null)
                 {
                     GetHostLogger(instance).LogDebug("Disposing ScriptHost.");
-                    instance.Dispose();
+
+                    if (isStandbyHost && !_scriptWebHostEnvironment.InStandbyMode)
+                    {
+                        // For cold start reasons delay disposing script host if specializing out of placeholder mode
+                        Utility.ExecuteAfterColdStartDelay(_environment, () =>
+                        {
+                            try
+                            {
+                                GetHostLogger(instance).LogDebug("Starting Standby ScriptHost dispose");
+                                instance.Dispose();
+                                _logger.LogDebug("Standby ScriptHost disposed");
+                            }
+                            catch (Exception e)
+                            {
+                                _logger.LogError(e, "Failed to dispose Standby ScriptHost instance");
+                                throw;
+                            }
+                        }, cancellationToken);
+
+                        GetHostLogger(instance).LogDebug("Standby ScriptHost marked for disposal");
+                    }
+                    else
+                    {
+                        instance.Dispose();
+                        GetHostLogger(instance).LogDebug("ScriptHost disposed");
+                    }
                 }
             }
         }
--- a/src/WebJobs.Script/Config/ScriptJobHostOptionsSetup.cs
+++ b/src/WebJobs.Script/Config/ScriptJobHostOptionsSetup.cs
@@ -8,7 +8,7 @@ using Microsoft.Extensions.Options;
 
 namespace Microsoft.Azure.WebJobs.Script.Configuration
 {
-    internal class ScriptHostOptionsSetup : IConfigureOptions<ScriptJobHostOptions>
+    internal class ScriptJobHostOptionsSetup : IConfigureOptions<ScriptJobHostOptions>
     {
         private readonly IConfiguration _configuration;
         private readonly IEnvironment _environment;
@@ -19,7 +19,7 @@ namespace Microsoft.Azure.WebJobs.Script.Configuration
         internal static readonly TimeSpan MaxFunctionTimeoutDynamic = TimeSpan.FromMinutes(10);
         internal static readonly TimeSpan DefaultFunctionTimeout = TimeSpan.FromMinutes(30);
 
-        public ScriptHostOptionsSetup(IConfiguration configuration, IEnvironment environment, IOptions<ScriptApplicationHostOptions> applicationHostOptions)
+        public ScriptJobHostOptionsSetup(IConfiguration configuration, IEnvironment environment, IOptions<ScriptApplicationHostOptions> applicationHostOptions)
         {
             _configuration = configuration;
             _environment = environment;
@@ -72,6 +72,7 @@ namespace Microsoft.Azure.WebJobs.Script.Configuration
             options.IsSelfHost = webHostOptions.IsSelfHost;
             options.TestDataPath = webHostOptions.TestDataPath;
             options.IsFileSystemReadOnly = webHostOptions.IsFileSystemReadOnly;
+            options.IsStandbyConfiguration = webHostOptions.IsStandbyConfiguration;
         }
 
         private void ConfigureFunctionTimeout(ScriptJobHostOptions options)
--- a/src/WebJobs.Script/Config/ScriptJobHostOptions.cs
+++ b/src/WebJobs.Script/Config/ScriptJobHostOptions.cs
@@ -117,5 +117,10 @@ namespace Microsoft.Azure.WebJobs.Script
         /// Gets or sets a value indicating whether the filesystem is read-only.
         /// </summary>
         public bool IsFileSystemReadOnly { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether the ScriptHost is in standby mode.
+        /// </summary>
+        public bool IsStandbyConfiguration { get; set; }
     }
 }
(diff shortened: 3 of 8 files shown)


src/WebJobs.Script.WebHost/Helpers/ScheduledDisposer.cs
```diff
@@ -0,0 +1,139 @@
...
+            {
+                timer?.Change(_delayTimeMs, Timeout.Infinite);
+            }
+            catch (ObjectDisposedException)
+            {
+                // might race with dispose
+            }
+            catch (Exception e)
+            {
+                _logger.LogError(e, $"{nameof(ScheduleTimer)}");
```
fabiocav: Let's not swallow this exception here. This might well be something we can't recover from.

src/WebJobs.Script.WebHost/Helpers/SemaphoreHelper.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Threading;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Helpers
+{
+    public class SemaphoreHelper
+    {
+        public virtual bool Wait(SemaphoreSlim semaphore, int millisecondsTimeout)
```
fabiocav: Can you clarify why this class is needed? Is this purely to avoid registering a semaphore?

fabiocav (commented):
This change could have unintended side effects and it introduces enough of a behavior change that it could impact customers. Have you considered those? Was there a consideration on having this pushed to 4.0 instead?

src/WebJobs.Script.WebHost/Helpers/SemaphoreHelper.cs
balag0: Its to make unit testing easier.

src/WebJobs.Script.WebHost/Helpers/ScheduledDisposer.cs
balag0: Good point. Will fix this.

balag0:
> This change could have unintended side effects and it introduces enough of a behavior change that it could impact customers. Have you considered those? Was there a consideration on having this pushed to 4.0 instead?

Thanks for taking a look @fabiocav 
Yes, that is what I am trying to make sure with this PR :). Could you explain a bit more on how this impacts customers. I am not very familiar with this code and looking at the way things are handled currently (fire and forget) it doesn't seem like we are enforcing any ordering with this task.  Also the exact ordering being enforced by this change is already happening in prod every now and then. So not clear on what the behavior change is :)

balag0:
Updated the PR to use ExecuteAfterColdStartDelay thanks to @safihamid's suggestion

balag0:
> > This change could have unintended side effects and it introduces enough of a behavior change that it could impact customers. Have you considered those? Was there a consideration on having this pushed to 4.0 instead?
> 
> Thanks for taking a look @fabiocav
> Yes, that is what I am trying to make sure with this PR :). Could you explain a bit more on how this impacts customers. I am not very familiar with this code and looking at the way things are handled currently (fire and forget) it doesn't seem like we are enforcing any ordering with this task. Also the exact ordering being enforced by this change is already happening in prod every now and then. So not clear on what the behavior change is :)

@fabiocav @brettsam I have updated the PR to make the changes simpler. Please do take a look and let me know if (and how) this will have any customer impact :)

src/WebJobs.Script.WebHost/Helpers/SemaphoreHelper.cs
balag0: not used anymore.

src/WebJobs.Script.WebHost/Helpers/ScheduledDisposer.cs
balag0: removed this code

src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
```diff
@@ -134,6 +135,9 @@ public static class WebScriptHostBuilderExtension
                     services.AddSingleton<IFileMonitoringService, FileMonitoringService>();
                     services.TryAddEnumerable(ServiceDescriptor.Singleton<IHostedService, IFileMonitoringService>(p => p.GetService<IFileMonitoringService>()));
 
+                    // Record Standby state when the Host was created
+                    services.AddSingleton(new ScriptHostStandbyStateProvider(standbyOptions.InStandbyMode));
```
balag0: Also considered injecting ScriptHostStandbyStateProvider to ScriptHost and exposing this with a getter on ScriptHost. But this seemed like a smaller changelist since this is not actually needed in ScriptHost and some of the tests were complicated to write since it is not easy to mock  ScriptHost.

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -683,7 +691,31 @@ private async Task Orphan(IHost instance, CancellationToken cancellationToken =
...
-                    instance.Dispose();
+
+                    if (isStandbyHost && !_scriptWebHostEnvironment.InStandbyMode)
+                    {
+                        // For cold start reasons delay disposing script host if specializing out of placeholder mode
+                        Utility.ExecuteAfterColdStartDelay(_environment, () =>
+                        {
+                            try
+                            {
+                                _logger.LogDebug("Starting ScripHost dispose");
```
brettsam: Should these calls be using `GetHostLogger()`?

brettsam: Also, can we make these two messages specifically mention that they're disposing the standby host? That'll be easier to diagnose things in kusto.

src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
brettsam: Rather than introduce a new service, could you add an `IsStandbyConfiguration` property to `ScriptJobHostOptions`? In the `ScriptHostOptionsSetup` (which should be renamed to match the options) -- it could copy the value directly from the `ScriptApplicationHostOptions`, which already has an `IsStandbyConfiguration` property.

brettsam (commented):
One refactoring comment that may clean up the PR a bit.

src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
balag0: done.

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
balag0: fixed.

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -684,7 +686,31 @@ private async Task Orphan(IHost instance, CancellationToken cancellationToken =
...
+                        {
+                            try
+                            {
+                                GetHostLogger(instance).LogDebug("Starting Standby ScriptHost dispose");
+                                instance.Dispose();
+                                _logger.LogDebug("Standby ScriptHost disposed");
+                            }
+                            catch (Exception e)
+                            {
+                                _logger.LogError(e, "Failed to dispose Standby ScriptHost instance");
```
pragnagopa, line 703: is try...catch needed? Ok to keep running even this disposal fails?

pragnagopa (commented):
Changes LGTM! Scoped to cold start request only! 

@fabiocav / @brettsam  - any other concerns?

src/WebJobs.Script/Config/ScriptJobHostOptionsSetup.cs
```diff
@@ -8,7 +8,7 @@
 
 namespace Microsoft.Azure.WebJobs.Script.Configuration
 {
-    internal class ScriptHostOptionsSetup : IConfigureOptions<ScriptJobHostOptions>
+    internal class ScriptJobHostOptionsSetup : IConfigureOptions<ScriptJobHostOptions>
```
brettsam, line 11: Thank you! :-)

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
brettsam: i think it's okay to catch and log -- but we should re-throw the exception to match the previous behavior (of not catching).

brettsam (commented):
Just 1 minor comment about that try/catch.

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
balag0: Thanks. Yes didnt mean to change that behavior.

src/WebJobs.Script/Config/ScriptJobHostOptionsSetup.cs
balag0: :)

pragnagopa (approved):

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
balag0: yes. didn't mean to change that behavior. Fixed.

src/WebJobs.Script/Config/ScriptJobHostOptionsSetup.cs
balag0: :)

brettsam (approved):

PR #7308, opened 2021-04-21
[pack] some artifact cleanup

brettsam:
I wanted to remove as many references to versions as I could to make moving to v4 (and v5, v6, etc) more straight-forward.

I looked through the devops pipelines that I knew about to see if anything depended on these artifact names and I didn't find anything. I assumed that there wouldn't be any hard dependencies since the names would change every build. But I want to see if @yojagad, @sidkri, @madelinegordon, @Francisco-Gamino had any objections.

*For example, the "Deploy to myget" step of the release pipeline uses paths like `$(System.DefaultWorkingDirectory)/_Azure.azure-functions-host/*/Microsoft.Azure.WebJobs.Script.3.*.nupkg`, which will continue to work.*

Here's what the change looks like:

Current (https://azfunc.visualstudio.com/Azure%20Functions/_build/results?buildId=12712&view=artifacts&pathAsName=false&type=publishedArtifacts):
image

This change (https://azfunc.visualstudio.com/Azure%20Functions/_build/results?buildId=12732&view=artifacts&pathAsName=false&type=publishedArtifacts):
image

### DIFF
--- a/build/build-extensions.ps1
+++ b/build/build-extensions.ps1
@@ -1,7 +1,7 @@
 param (
   [string]$buildNumber = "0",
   [string]$extensionVersion = "3.0.$buildNumber",
-  [string]$v2CompatibleExtensionVersion = "2.1.$buildNumber",
+  [string]$v2CompatibleExtensionVersion = "2.1.$buildNumber",  
   [string]$suffix = "",
   [string]$commitHash = "N/A",
   [string]$hashesForHardlinksFile = "hashesForHardlinks.txt"
@@ -15,7 +15,7 @@ $extensionVersionNoSuffix = $extensionVersion
 $v2CompatibleExtensionVersionNoSuffix = $v2CompatibleExtensionVersion
 
 if ($hasSuffix) {
-  $extensionVersion = "$extensionVersion-$suffix"
+  $extensionVersion = "$extensionVersion-$suffix"  
   $v2CompatibleExtensionVersion = "$v2CompatibleExtensionVersion-$suffix"
 }
 
@@ -44,11 +44,6 @@ function BuildRuntime([string] $targetRid, [bool] $isSelfContained) {
     
     $publishTarget = "$buildOutput\publish\$targetRid"
     $symbolsTarget = "$buildOutput\symbols\$targetRid"
-    
-    if ($isSelfContained) {
-        $publishTarget = "$publishTarget.self-contained"
-        $symbolsTarget = "$symbolsTarget.self-contained"
-    }
 
     $suffixCmd = ""
     if ($hasSuffix) {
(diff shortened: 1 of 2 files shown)


Francisco-Gamino:
Hello @brettsam -- Thank you for the heads up. 

From the Host build point of view, things will continue to work. However, for the Core Tools build (for integration tests), this is the list of packages that get update before we build: 
 
https://github.com/Azure/azure-functions-integration-tests/blob/dev/integrationTestsBuild/V3/CoreToolsBuild.json
```
[
"Microsoft.Azure.WebJobs.Script",
"Microsoft.Azure.WebJobs.Script.Grpc",
"Microsoft.Azure.WebJobs.Script.WebHost",
"Microsoft.Azure.Functions.PowerShellWorker.PS7",
"Microsoft.Azure.Functions.PythonWorker"
] 
```
The new package names will look like this `Microsoft.Azure.<HostPackageName>.<version>`. So, as long as the command below is able to find the new packages, we should be ok (https://github.com/Azure/azure-functions-core-tools/blob/v3.x/build/BuildSteps.cs#L68)

```
NuGet list {package} -Source {AzureFunctionsPreReleaseFeedName} -prerelease
```

Once we have all the language worker onboarded, we plan to move to use this command `dotnet list [<PROJECT>|<SOLUTION>] package ` to build the list of packages to update.

build/build-extensions.ps1
```diff
@@ -135,48 +133,57 @@ function CreateSiteExtensions() {
...
     Write-Host "Done copying. Elapsed: $($stopwatch.Elapsed)"
     Write-Host "======================================"
     Write-Host ""
 
-    Write-Host "Generating $hashesForHardlinksFile"
     Write-Host "======================================"
+    Write-Host "Generating $hashesForHardlinksFile"    
     WriteHashesFile $siteExtensionPath/$extensionVersionNoSuffix
-    Write-Host "Done generating $hashesForHardlinksFile"
+    Write-Host "Done generating $siteExtensionPath/$extensionVersionNoSuffix"
```
sidkri: Was this intentional?

sidkri (dismissed):

build/build-extensions.ps1
brettsam: Yeah, i like logging the full path rather than just the file name like it was previously

build/build-extensions.ps1
```diff
@@ -44,11 +44,6 @@ function BuildRuntime([string] $targetRid, [bool] $isSelfContained) {
     
     $publishTarget = "$buildOutput\publish\$targetRid"
     $symbolsTarget = "$buildOutput\symbols\$targetRid"
-    
-    if ($isSelfContained) {
-        $publishTarget = "$publishTarget.self-contained"
-        $symbolsTarget = "$symbolsTarget.self-contained"
-    }
```
brettsam, line 51: Removed this as it wasn't necessary.

build/build-extensions.ps1
sidkri: Got it.  But 149 should be changed too then so it matches "Generating .... Done generating".
Suggest mentioning the context in the log message as the path alone does not communicate what is happening.
Either
Write-Host "Generating $siteExtensionPath/$extensionVersionNoSuffix/$hashesForHardlinksFile"
...
Write-Host "Done generating $siteExtensionPath/$extensionVersionNoSuffix/$hashesForHardlinksFile"

OR 

Write-Host "Generating hashes for hard links"
...
Write-Host "Done generating hashes for hard links into "$siteExtensionPath/$extensionVersionNoSuffix"

yojagad (dismissed):
:shipit:

build/build-extensions.ps1
brettsam: i didn't see that on 149 -- you're right. Let me update!

yojagad (approved):

PR #7329, opened 2021-04-29
Add test for no content type in HTTP trigger

### LINKED ISSUE #6398: Add tests for more HttpTrigger scenarios
We recently had a regression where HttpTrigger functions were throwing exceptions if buffering related operations were done by the function code and "Content Typee" was not set to "application/json".

More info here -- https://github.com/Azure/azure-webjobs-sdk-extensions/issues/655

We need to add tests for these type of scenarios to avoid such issues.

ankitkumarr:

resolves https://github.com/Azure/azure-functions-host/issues/6398

We may want to consider adding similar tests for no content-type in language workers. I will open an issue for Node. But at least this by covers the gap to avoid regressions like that.

### Pull request checklist

### DIFF
--- a/test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SamplesTestHelpers.cs
+++ b/test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SamplesTestHelpers.cs
@@ -28,7 +28,19 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.EndToEnd
             string masterKey = await fixture.Host.GetMasterKeyAsync();
             uri = $"api/{functionName}?code={masterKey}&name=Mathew";
             request = new HttpRequestMessage(HttpMethod.Get, uri);
+            request.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("text/plain"));
+            response = await fixture.Host.HttpClient.SendAsync(request);
             Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+            Assert.Equal("text/plain", response.Content.Headers.ContentType.MediaType);
+            Assert.Equal("Hello Mathew", body);
+
+            // verify request suceeds without any content type set
+            uri = $"api/{functionName}?code={functionKey}&name=Mathew";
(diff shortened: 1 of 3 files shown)


test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SamplesTestHelpers.cs
```diff
@@ -28,7 +28,19 @@ public static async Task InvokeAndValidateHttpTrigger(EndToEndTestFixture fixtur
             string masterKey = await fixture.Host.GetMasterKeyAsync();
             uri = $"api/{functionName}?code={masterKey}&name=Mathew";
             request = new HttpRequestMessage(HttpMethod.Get, uri);
+            request.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("text/plain"));
+            response = await fixture.Host.HttpClient.SendAsync(request);
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+            Assert.Equal("text/plain", response.Content.Headers.ContentType.MediaType);
+            Assert.Equal("Hello Mathew", body);
```
ankitkumarr, line 35: This test was missing assertions.

mathewc: You're right - this helper intended to make the same request with the master key but never sent the request! Let's just fix this helper by sending the request and doing the same validations as above, and do your content type tests in a separate test. This helper is used in many places to send/validate requests, so we shouldn't be adding a bunch of combination testing in here.

ankitkumarr: That's fair! The reason I added here is that the callers for this function are the E2E test anytime they want to test Http scenario. To me it made sense that any time we test Http, we should also cover the absence of content types.
But I see your point to separate this out and intentionally add tests for content types that only targets the scenario.

Ok, I will make that update. Your suggestion makes more sense design wise.

mathewc (approved):

PR #7363, opened 2021-05-11
Removing noisy log for failure to get blob container

karshinlin:

resolves #issue_for_this_pr

### Pull request checklist

* [X ] My changes **should not** be added to the release notes for the next release

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script/StorageProvider/AzureStorageProvider.cs
+++ b/src/WebJobs.Script/StorageProvider/AzureStorageProvider.cs
@@ -65,7 +65,6 @@ namespace Microsoft.Azure.WebJobs.Script
             if (!TryGetBlobServiceClientFromConnection(out BlobServiceClient blobServiceClient, ConnectionStringNames.Storage))
             {
                 var ex = new InvalidOperationException($"Could not create BlobServiceClient to obtain the BlobContainerClient using Connection: {ConnectionStringNames.Storage}");
-                _logger.LogError(ex, "Invalid blob storage connection configuration. Could not create BlobServiceClient.");
                 throw ex;
             }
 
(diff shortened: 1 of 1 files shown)


src/WebJobs.Script/StorageProvider/AzureStorageProvider.cs
```diff
@@ -65,7 +65,6 @@ public virtual BlobContainerClient GetBlobContainerClient()
             if (!TryGetBlobServiceClientFromConnection(out BlobServiceClient blobServiceClient, ConnectionStringNames.Storage))
             {
                 var ex = new InvalidOperationException($"Could not create BlobServiceClient to obtain the BlobContainerClient using Connection: {ConnectionStringNames.Storage}");
```
alrod: "Could not create BlobServiceClient to obtain the BlobContainerClient using Connection:" will still be logged as error, right?

karshinlin: Since this is thrown, it's up to the caller to handle the exception. If it needs to be caught and logged, it will be up to the caller. In places where it is used today, the exception handling is not changed (whether it should be caught and eaten or not).

alrod: I meant you removed one exception but another one will be still logged
image

@balag0, do we also need to remove another one?

balag0: I don't completely understand the scenario to comment on the intent behind these logs being exposed to customers :) It may be that we think these error messages are useful to customers.

I think the bigger point is - It is okay for function apps to not have AzureWebJobsStorage app setting set. Emitting these error logs due to the lack of that appsetting will generate support cases. If we were to start emitting these logs for an app that is not seeing error messages currently we should document what this error means what action is expected from customers.

karshinlin: https://github.com/Azure/azure-functions-core-tools/blob/dev/src/Azure.Functions.Cli/Diagnostics/ColoredConsoleLogger.cs#L91

Looks like the core tools splits up the exception into multiple lines. Removing this log statement here should remove it from the cli tools. Working on validating this.

alrod (approved):

Issue #1234, opened 2017-02-24
Add in documentation that Node.js global vars may be keeped between deployments


yvele:
Specially usefull information when using Webpack to [drastically reduce Azure Function cold start](https://github.com/Azure/azure-webjobs-sdk-script/issues/298#issuecomment-281779696).

I understand the global vars sharing when I tried to use `babel-polyfill` and got this error:

> Error: only one instance of babel-polyfill is allowed

[Babel polyfill](https://babeljs.io/docs/usage/polyfill/) is using global vars and although I was importing it only once (in my `webpack.config`), I constantly get this error, and I finally had to make a little wrapper around my Azure Function handle to make sure `babel-polyfill` is not already loaded in global vars before requiring it.

I suggest adding a global vars warning in the documentation (especially with all Node.js developers going to make a webpack bundle to [reduce Azure Function cold start](https://github.com/Azure/azure-webjobs-sdk-script/issues/298#issuecomment-281779696).

This issue is kind of related to https://github.com/Azure/azure-webjobs-sdk-script/issues/298#issuecomment-281779696 and attempts to make Azure Function work with Webpack.

lindydonna:
@christopheranderson @yochay Shouldn't this be moved to the Functions repo, since it's about documentation?

christopheranderson:
@lindydonna - yes this should be moved

lindydonna:
@christopheranderson Could you file a new issue in the docs repo to track this?  You have more context on the doc requirement.

paulbatum:
This issue was moved to Azure/Azure-Functions#368

PR #7386, opened 2021-05-19
Add host.json for sync trigger payload

### LINKED ISSUE #7288: Adding Durable storageProvider.type to SyncTriggers payload
Currently the host does syncTriggers to update the system status of the triggers using this code 
https://github.com/Azure/azure-functions-host/blob/6b95141c6463d8815567bba913346dc780c40630/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs#L509-L535

But as far as I can tell, the type of the storage provider isn't taken into account. We will need that information in the payload somewhere if the type is `MicrosoftSQL`

Include the storage type somewhere in the payload. 

We could assume `MicrosoftSQL` for envorinments where it should be the right choice, but it may not be entirely correct or lead to issues.  

/cc @mathewc @cgillum

TsuyoshiUshio:
Under the Kubernetes Environment, We need to `host.json` payload for SyncTrigger scenario for transforming Durable Functions and LogicApp scenario. 

I currently testing this branch not finished, so that I make it Draft, However, this change should requires discussion with us. I just simply add the `host.json` payload on the current sync trigger payload. 

CC @pragnagopa , @ahmelsayed 

resolves https://github.com/Azure/azure-functions-host/issues/7288 https://github.com/Azure/azure-functions-host/issues/7382

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
@@ -304,10 +304,11 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
             var functionDetails = await WebFunctionsManager.GetFunctionMetadataResponse(listableFunctions, hostOptions, _hostNameProvider);
             result.Add("functions", new JArray(functionDetails.Select(p => JObject.FromObject(p))));
 
+            // TODO discuss more about the proper refactoring for GetHostJsonExtensionsAsync()
             // Add host.json to the payload
             if (_environment.IsKubernetesManagedHosting())
             {
-                result.Add("host.json", GetHostJson());
+                result.Add("extensions", await GetHostJsonExtensionsAsync());
             }
 
             // Add functions secrets to the payload
@@ -352,7 +353,9 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
             }
 
             string json = JsonConvert.SerializeObject(result);
-            if (json.Length > ScriptConstants.MaxTriggersStringLength)
+
+            // TODO Avoiding max request size limit operartion for Kubernetes Environment
+            if (json.Length > ScriptConstants.MaxTriggersStringLength && !_environment.IsKubernetesManagedHosting())
             {
                 // The settriggers call to the FE enforces a max request size
                 // limit. If we're over limit, revert to the minimal triggers
@@ -372,7 +375,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
             };
         }
 
-        internal Task<JObject> GetHostJsonAsync()
+        internal async Task<JObject> GetHostJsonExtensionsAsync()
         {
             var defaultJObject = new JObject();
             var hostOptions = _applicationHostOptions.CurrentValue.ToHostOptions();
@@ -381,7 +384,6 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
             {
                 _logger.LogInformation("Reading host.json for SyncTrigger.");
                 return JObject.Parse(await FileUtility.ReadAsync(hostJsonPath));
-
             }
             catch (JsonException ex)
             {
--- a/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
@@ -386,7 +386,6 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
             {
                 try
                 {
-                    _logger.LogInformation("Reading host.json for SyncTrigger.");
                     var hostJson = JObject.Parse(await FileUtility.ReadAsync(hostJsonPath));
                     if (hostJson.TryGetValue("extensions", out JToken token))
                     {
--- a/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
@@ -383,7 +383,15 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
             try
             {
                 _logger.LogInformation("Reading host.json for SyncTrigger.");
-                return JObject.Parse(await FileUtility.ReadAsync(hostJsonPath));
+                var hostJson = JObject.Parse(await FileUtility.ReadAsync(hostJsonPath));
+                if (hostJson.TryGetValue("extensions", out JToken token))
+                {
+                    return (JObject)token;
+                }
+                else
+                {
+                    return defaultJObject;
+                }
             }
             catch (JsonException ex)
             {
--- a/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
@@ -308,7 +308,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
             // Add the host.json extensions to the payload
             if (_environment.IsKubernetesManagedHosting())
             {
-                JObject extensionsPayload = await GetHostJsonExtensionsAsync();
+                JObject extensionsPayload = await GetHostJsonExtensionsAsync(_applicationHostOptions, _logger);
                 if (extensionsPayload != null)
                 {
                     result.Add("extensions", extensionsPayload);
@@ -378,9 +378,9 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
             };
         }
 
-        internal async Task<JObject> GetHostJsonExtensionsAsync()
+        internal static async Task<JObject> GetHostJsonExtensionsAsync(IOptionsMonitor<ScriptApplicationHostOptions> applicationHostOptions, ILogger logger)
         {
-            var hostOptions = _applicationHostOptions.CurrentValue.ToHostOptions();
+            var hostOptions = applicationHostOptions.CurrentValue.ToHostOptions();
             string hostJsonPath = Path.Combine(hostOptions.RootScriptPath, ScriptConstants.HostMetadataFileName);
             if (FileUtility.FileExists(hostJsonPath))
             {
@@ -398,7 +398,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
                 }
                 catch (JsonException ex)
                 {
-                    _logger.LogWarning($"Unable to parse host configuration file '{hostJsonPath}'. : {ex}");
+                    logger.LogWarning($"Unable to parse host configuration file '{hostJsonPath}'. : {ex}");
                     return null;
                 }
             }
--- a/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
@@ -382,12 +382,12 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
             }
             catch (JsonException ex)
             {
-                _logger.LogWarning($"Unable to parse host configuration file '{scriptPath}'. : {ex}");
+                _logger.LogWarning($"Unable to parse host configuration file '{hostFilePath}'. : {ex}");
                 return defaultJObject;
             }
             catch (Exception ex) when (ex is FileNotFoundException || ex is DirectoryNotFoundException)
             {
-                _logger.LogWarning($"Unable to find host configuration file '{scriptPath}'. : {ex}");
+                _logger.LogWarning($"Unable to find host configuration file '{hostFilePath}'. : {ex}");
                 return defaultJObject;
             }
         }
--- a/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
@@ -304,8 +304,11 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
             var functionDetails = await WebFunctionsManager.GetFunctionMetadataResponse(listableFunctions, hostOptions, _hostNameProvider);
             result.Add("functions", new JArray(functionDetails.Select(p => JObject.FromObject(p))));
 
-            // Add host.json to the pyaload
-            result.Add("host.json", GetHostJson());
+            // Add host.json to the payload
+            if (_environment.IsKubernetesManagedHosting())
+            {
+                result.Add("host.json", GetHostJson());
+            }
 
             // Add functions secrets to the payload
             // Only secret types we own/control can we cache directly
--- a/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
@@ -372,25 +372,25 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
             };
         }
 
-        internal JObject GetHostJson()
+        internal Task<JObject> GetHostJsonAsync()
         {
             var defaultJObject = new JObject();
-            var scriptPath = _applicationHostOptions.CurrentValue?.ScriptPath;
-            string hostFilePath = Path.Combine(scriptPath, ScriptConstants.HostMetadataFileName);
+            var hostOptions = _applicationHostOptions.CurrentValue.ToHostOptions();
+            string hostJsonPath = Path.Combine(hostOptions.RootScriptPath, ScriptConstants.HostMetadataFileName);
             try
             {
                 _logger.LogInformation("Reading host.json for SyncTrigger.");
-                string json = File.ReadAllText(scriptPath);
-                return JObject.Parse(json);
+                return JObject.Parse(await FileUtility.ReadAsync(hostJsonPath));
+
             }
             catch (JsonException ex)
             {
-                _logger.LogWarning($"Unable to parse host configuration file '{hostFilePath}'. : {ex}");
+                _logger.LogWarning($"Unable to parse host configuration file '{hostJsonPath}'. : {ex}");
                 return defaultJObject;
             }
             catch (Exception ex) when (ex is FileNotFoundException || ex is DirectoryNotFoundException)
             {
-                _logger.LogWarning($"Unable to find host configuration file '{hostFilePath}'. : {ex}");
+                _logger.LogWarning($"Unable to find host configuration file '{hostJsonPath}'. : {ex}");
                 return defaultJObject;
             }
         }
--- a/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
@@ -387,7 +387,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
             }
             catch (Exception ex) when (ex is FileNotFoundException || ex is DirectoryNotFoundException)
             {
-                _logger.LogWarning($"Unable to find host configuration file '{scriptPath}'. : {ex.Message} : {ex.StackTrace}");
+                _logger.LogWarning($"Unable to find host configuration file '{scriptPath}'. : {ex}");
                 return defaultJObject;
             }
         }
--- a/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
@@ -304,11 +304,15 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
             var functionDetails = await WebFunctionsManager.GetFunctionMetadataResponse(listableFunctions, hostOptions, _hostNameProvider);
             result.Add("functions", new JArray(functionDetails.Select(p => JObject.FromObject(p))));
 
-            // TODO discuss more about the proper refactoring for GetHostJsonExtensionsAsync()
-            // Add host.json to the payload
+            // TEMP: refactor this code to properly add extensions in all scenario(#7394)
+            // Add the host.json extensions to the payload
             if (_environment.IsKubernetesManagedHosting())
             {
-                result.Add("extensions", await GetHostJsonExtensionsAsync());
+                JObject extensionsPayload = await GetHostJsonExtensionsAsync();
+                if (extensionsPayload != null)
+                {
+                    result.Add("extensions", extensionsPayload);
+                }
             }
 
             // Add functions secrets to the payload
@@ -354,7 +358,6 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
 
             string json = JsonConvert.SerializeObject(result);
 
-            // TODO Avoiding max request size limit operartion for Kubernetes Environment
             if (json.Length > ScriptConstants.MaxTriggersStringLength && !_environment.IsKubernetesManagedHosting())
             {
                 // The settriggers call to the FE enforces a max request size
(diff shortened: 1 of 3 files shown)


src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -366,6 +369,29 @@ public async Task<SyncTriggersPayload> GetSyncTriggersPayload()
...
+                return JObject.Parse(json);
+            }
+            catch (JsonException ex)
+            {
+                _logger.LogWarning($"Unable to parse host configuration file '{scriptPath}'. : {ex.Message} : {ex.StackTrace}");
+                return defaultJObject;
+            }
+            catch (Exception ex) when (ex is FileNotFoundException || ex is DirectoryNotFoundException)
+            {
+                _logger.LogWarning($"Unable to find host configuration file '{scriptPath}'. : {ex.Message} : {ex.StackTrace}");
```
cgillum: Why do you separate `ex.Message` from `ex.StackTrace`? Isn't it better to just trace the full exception details?
```suggestion
                _logger.LogWarning($"Unable to find host configuration file '{scriptPath}'. : {ex}");
```

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -366,6 +369,29 @@ public async Task<SyncTriggersPayload> GetSyncTriggersPayload()
...
+                return defaultJObject;
```
cgillum: Just return a new JObject here and remove line 374.
```suggestion
                return new JObject();
```

cgillum (dismissed):
This looks good to me, but let's also get a sign-off from @mathewc

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
TsuyoshiUshio: I think we have one more place to need to do new JObject, so that I though to remove the duplication.

TsuyoshiUshio: I update the PR according to your practice. How about this? @cgillum 
It looks CI fails because of the testing for SyncTrigger. I'll check the tests and test with Kuberentes environment and remove draft tomorrow morning.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -304,6 +304,9 @@ public async Task<SyncTriggersPayload> GetSyncTriggersPayload()
             var functionDetails = await WebFunctionsManager.GetFunctionMetadataResponse(listableFunctions, hostOptions, _hostNameProvider);
             result.Add("functions", new JArray(functionDetails.Select(p => JObject.FromObject(p))));
 
+            // Add host.json to the pyaload
+            result.Add("host.json", GetHostJson());
```
mathewc: Note that the extended format { triggers: [], functions: [], secrets: {} } is best effort, and if the payload is oversize based on what FE can handle, we revert to the old triggers array format. This is fine because existing usages of the additional info are able to handle this - the extra data just enables optimizations.

However, in your case, it seems the host config level trigger info you want to add isn't optional? I.e. what what happen in your scenarios if due to size the payload reverted to minimal format?

mathewc: Also I worry about dumping the entire host.json file contents. Aren't we really only interested in host level extensions config?

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -304,6 +304,12 @@ public async Task<SyncTriggersPayload> GetSyncTriggersPayload()
             var functionDetails = await WebFunctionsManager.GetFunctionMetadataResponse(listableFunctions, hostOptions, _hostNameProvider);
             result.Add("functions", new JArray(functionDetails.Select(p => JObject.FromObject(p))));
 
+            // Add host.json to the payload
+            if (_environment.IsKubernetesManagedHosting())
+            {
+                result.Add("host.json", GetHostJson());
```
mathewc: See the comment I left in the previous iteration. Looks like a revision you posted blew the comment away.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
TsuyoshiUshio: Thank you @mathewc for the review. I didn't know the restriction. 
I updated the PR. We deicide to narrow it down the scope. Our usage is, under the Kubernetes Environment, The Sync Trigger will post request to the Kudu Lite API. (As far as I know, It doesn't limit the payload size).  https://github.com/Azure-App-Service/KuduLite/blob/feature/k8se/Kudu.Services/Function/FunctionController.cs#L43 In this case, looks no problem? I'm not sure which part is required for Durable/Logic Apps usage. @cgillum  Do you know which part is required?

cgillum: @mathewc It sounds like you're saying we can't rely on this logic to actually work? What controls whether we use the extended format or the "old triggers array format"? Also, what is the maximum size allowed by the FE? Indeed, the host.json metadata is required for our scenarios. We could probably trim it down to just the ["extensions" section](https://docs.microsoft.com/en-us/azure/azure-functions/functions-host-json#sample-hostjson-file), though, if you're worried about too much host.json bloat.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -304,6 +304,12 @@ public async Task<SyncTriggersPayload> GetSyncTriggersPayload()
             var functionDetails = await WebFunctionsManager.GetFunctionMetadataResponse(listableFunctions, hostOptions, _hostNameProvider);
             result.Add("functions", new JArray(functionDetails.Select(p => JObject.FromObject(p))));
 
+            // Add host.json to the payload
+            if (_environment.IsKubernetesManagedHosting())
```
divyagandhisethi, line 309: Can we add a unit test specifically for Kubernetes scenario?

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
mathewc: The code [here](https://github.com/Azure/azure-functions-host/blob/d831b969dcd3c8ed6a6949b82329b0db7ababf12/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs#L349) enforces the FE size limit. What we'd have to do is redefine what the "minimal format" is, and rather than just returning the triggers [] in such cases, we'd return `{ triggers: [], extensions: [] }`, scoping it so that we're not dumping the entire host.json in there, but only the extensions config section.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
TsuyoshiUshio: Thank you for you advice! Clever. Done.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
mathewc: Also note that there are still some legacy codepaths in Kudu (e.g. [here](https://github.com/projectkudu/kudu/blob/0fbbb59fd1151f61daee544715e1326a8632be76/Kudu.Core/Helpers/PostDeploymentHelper.cs#L266) that will end up doing an "old format" sync). @ankitkumarr did the work in Kudu to forward old Kudu synctriggers calls to the Functions runtime to get the new format, but if that fails for any reason, Kudu falls back to it's previous logic where the sync formatting was done by Kudu. So any consumers of this synctriggers payload (ScaleController, Geo ARM APIs, etc.) need to handle both formats, and they do currently.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
cgillum, line 309: Is there a strong reason to limit this to Kubernetes? I'd like to also have this change for Antares too.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -366,6 +372,29 @@ public async Task<SyncTriggersPayload> GetSyncTriggersPayload()
             };
         }
 
+        internal JObject GetHostJson()
```
mathewc: There's already logic in ReadDurableTaskConfig that is reading host.json. Let's not duplicate this code, instead refactoring to share the existing logic.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
mathewc: Let's not make this Kubernetes specific, since part of our larger goal is to address https://github.com/Azure/azure-functions-host/issues/7288. Instead, I think we should add just the host.json "extensions" section here, and below if we revert to minimal format based on size, we redefine the minimal format as `{ triggers: [], extensions: {} }` ensuring this data is present in that case as well.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
TsuyoshiUshio: Hi @cgillum 
This is mainly time pressure. Since we need to finish today with testing, 
Scoping it that is not effect existing scenario. Also save time for fixing E2E fails. 
We can create a issue to tracking make this global.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
TsuyoshiUshio: What do you think, @cgillum ? Is it unblock the Durable for Kubernetes Environment?

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
TsuyoshiUshio: Thank you  @mathewc , @cgillum @ankitkumarr 
Since I create this PR for unblock the Durable usage,  Is this looks good? or any idea/discussion?

1. limit the payload adding `extensions` part only for kubernetes environmnet.
2. Only the kubernetes environmnet, disable the MaxTriggersStringLength (it is 204800)
3. Create an issue for make SyincTrigger with host.json payload in global for keep on discussing it. 

Only thing, what I don't know is the [code](https://github.com/projectkudu/kudu/blob/0fbbb59fd1151f61daee544715e1326a8632be76/Kudu.Core/Helpers/PostDeploymentHelper.cs#L266) that is done by @ankitkumarr  is releated kubernets environmnet. We are working on this [branch](https://github.com/Azure-App-Service/KuduLite/tree/feature/k8se)

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
pragnagopa: We need to scope the changes for Kubernetes Env as we will be taking commit as hotfix and skip OGFs

@TsuyoshiUshio  - Please open a separate issue to remove the check for Kubernetes env after we cut the hotfix release - to ensure change go through full functions release validation

cgillum: Makes sense. Thanks!

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
cgillum: Since this PR is more of an "emergency" PR and we're opening a second one for the "proper" fix, I suggest we defer refactoring for the later PR (let's add a note so that we don't forget).

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -366,6 +372,29 @@ public async Task<SyncTriggersPayload> GetSyncTriggersPayload()
...
+        {
+            var defaultJObject = new JObject();
+            var scriptPath = _applicationHostOptions.CurrentValue?.ScriptPath;
```
mathewc: Don't need the ? null check, since you're not handling the null if it happened on the next line anyways. Existing ReadDurableTaskConfig code doesn't check for null either.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -366,6 +372,29 @@ public async Task<SyncTriggersPayload> GetSyncTriggersPayload()
...
+            string hostFilePath = Path.Combine(scriptPath, ScriptConstants.HostMetadataFileName);
+            try
+            {
+                _logger.LogInformation("Reading host.json for SyncTrigger.");
+                string json = File.ReadAllText(scriptPath);
```
mathewc: Recommend using FileUtility.ReadAsync as the other code does, which will enable you to mock the file access as other tests do

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
TsuyoshiUshio: Thanks.  I change the logic to use the same as Durable.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
TsuyoshiUshio: Thanks. I didn't know that. I changed it.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
TsuyoshiUshio: Thank you @mathewc  
Done. 
Could you have a look?

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
TsuyoshiUshio: I added some TODO's try not to forget.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -366,6 +375,36 @@ public async Task<SyncTriggersPayload> GetSyncTriggersPayload()
...
+                {
+                    return defaultJObject;
+                }
+            }
+            catch (JsonException ex)
+            {
+                _logger.LogWarning($"Unable to parse host configuration file '{hostJsonPath}'. : {ex}");
+                return defaultJObject;
+            }
+            catch (Exception ex) when (ex is FileNotFoundException || ex is DirectoryNotFoundException)
```
mathewc: Remove this file not found warning completely. Presence of the host.json file isn't mandatory so we don't want to create unnecessary warnings. See the other code which does a FileUtility.FileExists(hostJsonPath) check. 

Also, you should just return null if no host.json file or no extensions node found in the file, and have the calling code only add the extensions property when needed.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -346,7 +353,9 @@ public async Task<SyncTriggersPayload> GetSyncTriggersPayload()
             }
 
             string json = JsonConvert.SerializeObject(result);
-            if (json.Length > ScriptConstants.MaxTriggersStringLength)
+
+            // TODO Avoiding max request size limit operartion for Kubernetes Environment
```
mathewc: Recommend removing this TODO, since likely we'll never want the limit to apply in Kubernetes cases, since there is no FE limit to deal with?

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -304,6 +304,13 @@ public async Task<SyncTriggersPayload> GetSyncTriggersPayload()
             var functionDetails = await WebFunctionsManager.GetFunctionMetadataResponse(listableFunctions, hostOptions, _hostNameProvider);
             result.Add("functions", new JArray(functionDetails.Select(p => JObject.FromObject(p))));
 
+            // TODO discuss more about the proper refactoring for GetHostJsonExtensionsAsync()
```
mathewc: Please create a new issue in this repo for removing/fixing this code, and link to that issue here. E.g. single comment like "TEMP: refactor this code to properly add extensions in all scenarios (#1234)". We can then include all the details in the issue.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
TsuyoshiUshio: Done!

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
TsuyoshiUshio: Thank you for the review. I added issue, and update the comment as you requested! 
https://github.com/Azure/azure-functions-host/issues/7394

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
TsuyoshiUshio: I introduce the change as you requested. Thank you for the review!

TsuyoshiUshio:
Hi @mathewc 
I made change as you requested. Thank you for the quick response and great review!

TsuyoshiUshio:
I'm testing this PR E2E once it finished, I'll share it.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -366,6 +378,35 @@ public async Task<SyncTriggersPayload> GetSyncTriggersPayload()
...
 
+        internal async Task<JObject> GetHostJsonExtensionsAsync()
+        {
+            var hostOptions = _applicationHostOptions.CurrentValue.ToHostOptions();
+            string hostJsonPath = Path.Combine(hostOptions.RootScriptPath, ScriptConstants.HostMetadataFileName);
+            if (FileUtility.FileExists(hostJsonPath))
+            {
+                try
+                {
+                    _logger.LogInformation("Reading host.json for SyncTrigger.");
```
mathewc: Please remove this log statement - it'll just be noise.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
mathewc: Looks good. I just had one more comment that I added on removing the unnecessary log statement. They we can get this checked in after you add a test for the scenario - I believe you're in the process of adding one?

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
pragnagopa: If this needed for debugging - please change to Trace level

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -366,6 +378,35 @@ public async Task<SyncTriggersPayload> GetSyncTriggersPayload()
             };
         }
 
+        internal async Task<JObject> GetHostJsonExtensionsAsync()
+        {
+            var hostOptions = _applicationHostOptions.CurrentValue.ToHostOptions();
```
pragnagopa: take `_applicationHostOptions` as input param to make easier to add unit test

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
mathewc: I'd argue that its not needed at all. Note that on the other existing path for Durable where we're reading host.json, we don't log anything. I'd just remove it :)

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
TsuyoshiUshio: Thank you @mathewc 
I tested E2E, so that code is good for now, I'm planning add unit testing for it. 
Do we need E2E testing as well? If so I'd happy to do that.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
TsuyoshiUshio: Ok, I removed it.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
pragnagopa: For the scope of the changes here unit test is enough! Thanks!

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
TsuyoshiUshio: Done.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
TsuyoshiUshio: I added an Unit Test. Looks good?

pragnagopa: I don’t see updates to unit tests. Missed a commit?

TsuyoshiUshio: Hi @pragnagopa 
Sorry, I missed. Now added.

test/WebJobs.Script.Tests/Managment/FunctionsSyncManagerTests.cs
```diff
@@ -0,0 +1,103 @@
...
+            monitor.Setup(x => x.CurrentValue).Returns(options);
+            var json = await FunctionsSyncManager.GetHostJsonExtensionsAsync(monitor.Object, logger.Object);
+            if (isNull)
+            {
+                Assert.Null(json);
+            }
+            else
+            {
+                Assert.Equal(value, json["queues"]?["maxPollingInterval"]);
+            }
```
pragnagopa, line 100: Add try… finally . Move test directory clean up into finally

TsuyoshiUshio: That's a good idea.

TsuyoshiUshio: Done.

pragnagopa (approved):

test/WebJobs.Script.Tests/Managment/FunctionsSyncManagerTests.cs
```diff
@@ -0,0 +1,110 @@
...
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Script.WebHost.Management;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Moq;
+using Xunit;
+
+namespace Microsoft.Azure.WebJobs.Script.Tests.Managment
+{
+    public class FunctionsSyncManagerTests
```
mathewc, line 17: Why did you add a new file for this test - we already have an existing test suite here https://github.com/Azure/azure-functions-host/blob/6cef5808f90e3a9d3437209ab128361852120597/test/WebJobs.Script.Tests.Integration/Management/FunctionsSyncManagerTests.cs#L31.

In the upcoming PR, let's fix this. Also, the test doesn't really test the feature/changes you added - ensuring that the actual sync triggers payload contains the expected content.

pragnagopa: Thanks @mathewc  - captured notes here: https://github.com/Azure/azure-functions-host/issues/7394

PR #7395, opened 2021-05-20
calling log statement while in placeholder mode

### LINKED ISSUE #7242: Microsoft.Azure.WebJobs.Script.Diagnostics.Extensions.LoggerExtension logging during cold start
This call https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs#L111 is invoking bunch of logging and JIT methods during specialization as shown below. Inc% is ~7% of the time spent in Functions process in VMSS stamp.
Need to call something similar during placeholder to avoid all those initializations during specialization or remove the method altogether?
 

Name | Inc % | Inc
-- | -- | --
\|\|    \|     \|     \|    \|\| \|+   microsoft.azure.webjobs.script!Microsoft.Azure.WebJobs.Script.ScriptHostBuilderExtensions+<>c__DisplayClass7_3.<AddScriptHostCore>b__8(Microsoft.Azure.WebJobs.Hosting.IWebJobsStartupTypeLocator) | 9.4 | 28
\|\|    \|     \|     \|    \|\| \|\|+   microsoft.azure.webjobs.host!WebJobsBuilderExtensions.UseExternalConfigurationStartup | 9.4 | 28
\|\|    \|     \|     \|    \|\| \|\| + system.private.corelib!System.Lazy`1[System.__Canon].get_Value() | 9.4 | 28
\|\|    \|     \|     \|    \|\| \|\|  +   system.private.corelib!System.Lazy`1[System.__Canon].CreateValue() | 9.4 | 28
\|\|    \|     \|     \|    \|\| \|\|   + system.private.corelib!System.Lazy`1[System.__Canon].ExecutionAndPublication(System.LazyHelper,   Boolean) | 9.4 | 28
\|\|    \|     \|     \|    \|\| \|\|    +   system.private.corelib!System.Lazy`1[System.__Canon].ViaFactory(System.Threading.LazyThreadSafetyMode) | 9.4 | 28
\|\|    \|     \|     \|    \|\| \|\|     +   microsoft.azure.webjobs.script!Microsoft.Azure.WebJobs.Script.DependencyInjection.ScriptStartupTypeLocator.<.ctor>b__7_0() | 9.4 | 28
\|\|    \|     \|     \|    \|\| \|\|      +   system.private.corelib.il!System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1[System.__Canon].Start(!!0&) | 9.4 | 28
\|\|    \|     \|     \|    \|\| \|\|       +   system.private.corelib.il!AsyncMethodBuilderCore.Start | 9.4 | 28
\|\|    \|     \|     \|    \|\| \|\|          +   microsoft.azure.webjobs.script!Microsoft.Azure.WebJobs.Script.DependencyInjection.ScriptStartupTypeLocator+<GetExtensionsStartupTypesAsync>d__11.MoveNext() | 9.4 | 28
\|\|    \|     \|     \|    \|\| \|\|           + microsoft.azure.webjobs.script!LoggerExtension.ScriptStartNotLoadingExtensionBundle | 7.7 | 23
\|\|    \|     \|     \|    \|\| \|\|           \|+   microsoft.extensions.logging.abstractions.il!Microsoft.Extensions.Logging.LoggerMessage+<>c__DisplayClass8_0`4[System.__Canon,System.Boolean,System.Boolean,System.Boolean].<Define>b__1(class   Microsoft.Extensions.Logging.ILogger,!0,!1,!2,!3,class System.Exception) | 7.4 | 22
\|\|    \|     \|     \|    \|\| \|\|           \|\|+   microsoft.extensions.logging.abstractions.il!Microsoft.Extensions.Logging.LoggerMessage+<>c__DisplayClass8_0`4[System.__Canon,System.Boolean,System.Boolean,System.Boolean].<Define>g__Log\|0(class   Microsoft.Extensions.Logging.ILogger,!0,!1,!2,!3,class System.Exception) | 7.1 | 21
\|\|    \|     \|     \|    \|\| \|\|           \|\|\|+   microsoft.extensions.logging.abstractions.il!Microsoft.Extensions.Logging.Logger`1[System.__Canon].Microsoft.Extensions.Logging.ILogger.Log(value   class Microsoft.Extensions.Logging.LogLevel,value class   Microsoft.Extensions.Logging.EventId,!!0,class System.Exception,class   System.Func`3) | 6.4 | 19
\|\|    \|     \|     \|    \|\| \|\|           \|\|\|\|+   microsoft.extensions.logging.il!Logger.Log | 5.7 | 17
\|\|    \|     \|     \|    \|\| \|\|           \|\|\|\|\|+ microsoft.extensions.logging.il!Microsoft.Extensions.Logging.Logger.<Log>g__LoggerLog\|12_0(value   class Microsoft.Extensions.Logging.LogLevel,value class   Microsoft.Extensions.Logging.EventId,class   Microsoft.Extensions.Logging.ILogger,class System.Exception,class System.Func`3,class   System.Collections.Generic.List`1&,!!0&) | 5.4 | 16
\|\|    \|     \|     \|    \|\| \|\|           \|\|\|\|\|\|+   microsoft.azure.webjobs.script.webhost.il!SystemLogger.Log | 3.0 | 9
\|\|    \|     \|     \|    \|\| \|\|           \|\|\|\|\|\|\|+   microsoft.extensions.logging.abstractions.il!Microsoft.Extensions.Logging.LoggerMessage+LogValues`4+<GetEnumerator>d__13[System.__Canon,System.Boolean,System.Boolean,System.Boolean].MoveNext() | 1.7 | 5

soninaren:
Placing a log statement to be called while the function app is in placeholder mode to avoid jitting cost during specialization

resolves #7242

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
+++ b/src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
@@ -95,6 +95,12 @@ namespace Microsoft.Azure.WebJobs.Script.DependencyInjection
                     _logger.ScriptStartUpErrorLoadingExtensionBundle();
                     return new Type[0];
                 }
+
+                if (SystemEnvironment.Instance.IsPlaceholderModeEnabled())
+                {
+                    // Calling this log statement in the placeholder mode to avoid jitting during specializtion
+                    _logger.ScriptStartNotLoadingExtensionBundle(extensionsPath, bundleConfigured, isPrecompiledFunctionApp, isLegacyExtensionBundle);
+                }
                 _logger.ScriptStartUpLoadingExtensionBundle(extensionsPath);
             }
             else
--- a/src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
+++ b/src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
@@ -87,6 +87,14 @@ namespace Microsoft.Azure.WebJobs.Script.DependencyInjection
             }
 
             bool isLegacyExtensionBundle = _extensionBundleManager.IsLegacyExtensionBundle();
+
+            if (SystemEnvironment.Instance.IsPlaceholderModeEnabled())
+            {
+                // Do not move this.
+                // Calling this log statement in the placeholder mode to avoid jitting during specializtion
+                _logger.ScriptStartNotLoadingExtensionBundle("WARMUP_LOG_ONLY", bundleConfigured, isPrecompiledFunctionApp, isLegacyExtensionBundle);
+            }
+
             if (bundleConfigured && (!isPrecompiledFunctionApp || _extensionBundleManager.IsLegacyExtensionBundle()))
             {
                 extensionsPath = await _extensionBundleManager.GetExtensionBundleBinPathAsync();
@@ -94,12 +102,6 @@ namespace Microsoft.Azure.WebJobs.Script.DependencyInjection
                 {
                     _logger.ScriptStartUpErrorLoadingExtensionBundle();
                     return new Type[0];
-                }
-
-                if (SystemEnvironment.Instance.IsPlaceholderModeEnabled())
-                {
-                    // Calling this log statement in the placeholder mode to avoid jitting during specializtion
-                    _logger.ScriptStartNotLoadingExtensionBundle(extensionsPath, bundleConfigured, isPrecompiledFunctionApp, isLegacyExtensionBundle);
                 }
                 _logger.ScriptStartUpLoadingExtensionBundle(extensionsPath);
             }


fabiocav (dismissed):
This looks good to me, but it would be good for @safihamid to look and let us know whether he has any additional concerns.

safihamid (commented):
Shouldn't this get called outside of the 'if' statement to make sure it always gets called during placeholder mode?
Not sure if this condition will be true during placeholder mode:


 if (bundleConfigured &amp;&amp; (!isPrecompiledFunctionApp || _extensionBundleManager.IsLegacyExtensionBundle()))            {



Keep in mind, we see this issue for .Net apps during specialization which have no extension bundles.

brettsam:
@safihamid -- it already gets called in the `else` so I think it's covered in both cases now.

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
```diff
@@ -95,6 +95,12 @@ public async Task<IEnumerable<Type>> GetExtensionsStartupTypesAsync()
                     _logger.ScriptStartUpErrorLoadingExtensionBundle();
                     return new Type[0];
                 }
+
+                if (SystemEnvironment.Instance.IsPlaceholderModeEnabled())
+                {
+                    // Calling this log statement in the placeholder mode to avoid jitting during specializtion
+                    _logger.ScriptStartNotLoadingExtensionBundle(extensionsPath, bundleConfigured, isPrecompiledFunctionApp, isLegacyExtensionBundle);
```
brettsam: Can we do something to make it super-obvious that this isn't an accurate log? Just thinking:

```
_logger.ScriptStartNotLoadingExtensionBundle("WARMUP_LOG_ONLY", bundleConfigured, isPrecompiledFunctionApp, isLegacyExtensionBundle);
```

safihamid:
@brettsam the call for else part is actually what caused all these JIT calls.  Else part happens during specialization for .Net apps.

brettsam:
In what placeholder scenario would this call get missed now? With this change, it's in both the `if` and the `else`, so it should always be called no matter what in placeholder mode, right?

It certainly doesn't hurt to move it outside the `if`, but I think his change covers it.

safihamid:
Yes makes sense, I just want to make sure this new code path does get called during placeholder. Is bundleConfigured set true for dotnet placeholders?
Need to make sure this whole method gets called during placeholders. If it doesn't then doesn't matter where we add the log either in if or else part. It's not easy to track the call stack on my phone right now 🙂

The else part was getting called during specialization for dotnet apps.

fabiocav:
@safihamid has a good point about logic that is conditional on the arguments passed in. It's hard to get from the call site, but there's indeed a possibility we'd be calling the method and not executing the code we expect, principally if someone inadvertently changes the logic in the method.

Can we at least have a test that simulates placeholder mode and ensures this is getting called? That will avoid potential regressions.

soninaren:
@safihamid , @fabiocav, @brettsam 
I can comment on the over all logic, when the else part gets and why is this issue only surfacing now.

1. We currently have bundles configured in placeholder mode. Specifically version we have configured is `[2.*, 3.0.0)`. This configuration is a non-legacy bundle. 
2. This until very recently it was a legacy bundle `[1.*, 2.0.0)`. This might be why we are seeing it just now.
3. For dotnet apps that have bundle version configured to `[2.*, 3.0.0)` the else block will trigger.
4. The `If` block will trigger during the placeholder mode because `!isPrecompiledFunctionApp` would be true. 

Just to be on the safe side I will move the logic in the outside if block and put a comment on it saying that `do not move this` along with the existing comment. Will also update the path text to `WARM_UP_ONLY` as suggested by @brettsam

soninaren:
> @safihamid , @fabiocav, @brettsam
> I can comment on the over all logic, when the else part gets and why is this issue only surfacing now.
> 
> 1. We currently have bundles configured in placeholder mode. Specifically version we have configured is `[2.*, 3.0.0)`. This configuration is a non-legacy bundle.
> 2. This until very recently it was a legacy bundle `[1.*, 2.0.0)`. This might be why we are seeing it just now.
> 3. For dotnet apps that have bundle version configured to `[2.*, 3.0.0)` the else block will trigger.
> 4. The `If` block will trigger during the placeholder mode because `!isPrecompiledFunctionApp` would be true.
> 
> Just to be on the safe side I will move the logic in the outside if block and put a comment on it saying that `do not move this` along with the existing comment. Will also update the path text to `WARM_UP_ONLY` as suggested by @brettsam

@safihamid , @fabiocav, @brettsam
Please let me know if this sounds good. I will go ahead and make the change.

safihamid:
@soninaren thanks for the explanation. yes moving it outside the if block sounds good to me.

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
soninaren: done

soninaren:
Have updated the PR as discussed.

safihamid (approved):

PR #7410, opened 2021-05-26
[pack] updating all packages to preview4

brettsam:


test/WebJobs.Script.Tests.Integration/Middleware/AllowSynchronousIOMiddlewareTests.cs
```diff
@@ -15,7 +15,7 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.Integration.Middleware
 {
     public class AllowSynchronousIOMiddlewareTests
     {
-        [Fact]
+        [Fact(Skip = "Seems very difficult to trigger this failure now.")]
```
brettsam, line 18: We have a test that does a synchronous read of the request and we expect it to fail (just to confirm that our following tests are actually doing what they're supposed to). This stopped working when I moved to .NET 6.0.

It looks like a [fix in the TestServer](https://github.com/dotnet/aspnetcore/pull/24107) now writes the Content-Length automatically (since 5.0 rc1)

Because of this, the HttpTrigger now sees this as [readable and calls `EnableBuffering()`](https://github.com/brettsam/azure-webjobs-sdk-extensions/blob/ac8226ba786fe041cf215889d83880961bef93b0/src/WebJobs.Extensions.Http/Extensions/HttpRequestExtensions.cs#L28), which swaps out the Request body stream for [one that does not throw an exception with sync reads](https://github.com/dotnet/aspnetcore/blob/52eff90fbcfca39b7eb58baad597df6a99a542b0/src/Http/Http/src/Internal/BufferingHelper.cs#L24). This is the behavior that happens in prod -- but now we're seeing it here b/c the TestServer was fixed.

I tried a bunch of other stuff and was unable to throw this exception from within a function. Even this didn't work in the TestServer: https://github.com/Azure/azure-functions-host/issues/7064

I'm skipping this now -- I'm not sure how valuable this is going forward but we can revisit.

ankitkumarr: Wondering if we need a follow-up issue to either remove this test or fix / modify this test.

ankitkumarr (approved):

PR #7412, opened 2021-05-27
[Linux Consumption] Use standard JsonSerializer for publishing Functi…

### LINKED ISSUE #7409: Possible FunctionExecutionStatus deserialization errors
related to https://github.com/Azure/azure-functions-docker-private/issues/175

We had one CRI, where the FunctionExecutionStatus on functionhost and metering service didn't match each other. Add a count of inprogress execution event count to allow clients to identify and take remedial action.

balag0:
…onExecutionStatus

This is to isolate metrics publisher from overrides to global json serializer settings from function app code.

resolves https://github.com/Azure/azure-functions-host/issues/7409

### Pull request checklist

* [ X] My changes **do not** require documentation changes
* [ X] My changes **do not** need to be backported to a previous version

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/Management/MeshServiceClient.cs
+++ b/src/WebJobs.Script.WebHost/Management/MeshServiceClient.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using System.Net.Http;
 using System.Threading.Tasks;
@@ -19,9 +20,12 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
         private const string BindMountOperation = "bind-mount";
         public const string SquashFsOperation = "squashfs";
         public const string ZipOperation = "zip";
+        public const string AddFES = "add-fes";
         private readonly HttpClient _client;
         private readonly ILogger _logger;
         private readonly IEnvironment _environment;
+
+        private static readonly JsonSerializer Serializer = JsonSerializer.Create(new JsonSerializerSettings());
 
         public MeshServiceClient(HttpClient client, IEnvironment environment, ILogger<MeshServiceClient> logger)
         {
@@ -92,7 +96,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
                 Source = source.ToString()
             };
 
-            var healthEventString = JsonConvert.SerializeObject(healthEvent);
+            var healthEventString = Serialize(healthEvent);
 
             _logger.LogInformation($"Posting health event {healthEventString}");
 
@@ -126,8 +130,8 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
 
             var operation = new[]
             {
-                new KeyValuePair<string, string>(Operation, "add-fes"),
-                new KeyValuePair<string, string>("content", JsonConvert.SerializeObject(activities)),
+                new KeyValuePair<string, string>(Operation, AddFES),
+                new KeyValuePair<string, string>("content", Serialize(activities)),
             };
 
             var response = await SendAsync(operation);
@@ -142,6 +146,18 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
                 new FormUrlEncodedContent(formData));
             _logger.LogDebug($"Mesh response {res.StatusCode}");
             return res;
+        }
+
+        private static string Serialize<T>(T o)
+        {
+            string serialized;
+            using (var stringWriter = new StringWriter())
+            {
+                Serializer.Serialize(stringWriter, o);
+                serialized = stringWriter.ToString();
+            }
+
+            return serialized;
         }
     }
 }
--- a/test/WebJobs.Script.Tests.Integration/Management/MeshServiceClientTests.cs
+++ b/test/WebJobs.Script.Tests.Integration/Management/MeshServiceClientTests.cs
@@ -55,7 +55,7 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.Integration.Management
         {
             var formData = request.Content.ReadAsFormDataAsync().Result;
             if (string.Equals(MeshInitUri, request.RequestUri.AbsoluteUri) &&
-                string.Equals("add-fes", formData["operation"]))
+                string.Equals(MeshServiceClient.AddFES, formData["operation"]))
             {
                 var activityContent = formData["content"];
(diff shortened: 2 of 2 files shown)


balag0:
Replaces https://github.com/Azure/azure-functions-host/pull/7411

src/WebJobs.Script.WebHost/Management/MeshServiceClient.cs
```diff
@@ -126,8 +130,8 @@ private async Task PublishActivities(IEnumerable<ContainerFunctionExecutionActiv
 
             var operation = new[]
             {
-                new KeyValuePair<string, string>(Operation, "add-fes"),
-                new KeyValuePair<string, string>("content", JsonConvert.SerializeObject(activities)),
+                new KeyValuePair<string, string>(Operation, AddFES),
+                new KeyValuePair<string, string>("content", Serialize(activities)),
```
mathewc, line 134: You could consider simplifying here by just using the static JsonConvert.Serialize(object, JsonSerializerSettings) overload, and using a statically defined set of settings.

test/WebJobs.Script.Tests.Integration/Management/MeshServiceClientTests.cs
```diff
@@ -55,7 +55,7 @@ private static bool IsPublishExecutionStatusRequest(HttpRequestMessage request,
         {
             var formData = request.Content.ReadAsFormDataAsync().Result;
             if (string.Equals(MeshInitUri, request.RequestUri.AbsoluteUri) &&
-                string.Equals("add-fes", formData["operation"]))
+                string.Equals(MeshServiceClient.AddFES, formData["operation"]))
```
mathewc, line 58: I believe your Deserialize call is also exposed to this issue. Should probably specify explicit settings on that call as well.

test/WebJobs.Script.Tests.Integration/Management/MeshServiceClientTests.cs
```diff
@@ -209,5 +209,77 @@ public async Task CreatesBindMount()
                 ItExpr.Is<HttpRequestMessage>(r => IsCreateBindMountRequest(r, sourcePath, targetPath)),
                 ItExpr.IsAny<CancellationToken>());
         }
+
+        [Fact]
+        public async Task UsesCustomJsonSerializer()
```
mathewc: This test really isn't testing the actual production code change you're making (that when no serializer specified, we're using a set of explicit settings rather than taking global defaults. To do so, you'd use a try/finally block, and have the test repro the actual issue by capturing the current global settings, modifying them for the test to change enum behavior, and verify those changes have no effect. Then restore the previous global settings in the finally block.

I.e. you want a test that would fail if the code started using global default settings.

You can then remove the client ctor serializer param.

src/WebJobs.Script.WebHost/Management/MeshServiceClient.cs
balag0: I tried that initially but it doesnt seem to work unless we use JsonSerializer.Create.

https://www.newtonsoft.com/json/help/html/M_Newtonsoft_Json_JsonSerializer_Create_1.htm

Also noticed similar pattern in https://github.com/Azure/azure-functions-host/blob/cbf9ad1b107c52531975f622d7f50f63ab31e592/src/WebJobs.Script/Timer/AzureStorageScheduleMonitor.cs#L44

brettsam: I agree with @mathewc -- I think you can make this a smaller change with https://www.newtonsoft.com/json/help/html/M_Newtonsoft_Json_JsonConvert_SerializeObject_5.htm.

```
private static readonly JsonSerializer _serializerSettings = new JsonSerializerSettings();
```

and then...

```
new KeyValuePair<string, string>("content", JsonConvert.SerializeObject(activities, _serializerSettings)),
```

will that work? I think that's typically the pattern we use.

balag0: No. That doesn't seem to work and based on my searching around JsonSerializer.Create seems to be the only way to avoid using the global settings

From https://www.newtonsoft.com/json/help/html/M_Newtonsoft_Json_JsonSerializer_Create_1.htm
Creates a new JsonSerializer instance using the specified JsonSerializerSettings. The JsonSerializer **will not use** default settings from DefaultSettings.

brettsam: If you pass settings to JsonConvert.SerializeObject() as the second param, it'll use those instead of global:

From: https://www.newtonsoft.com/json/help/html/M_Newtonsoft_Json_JsonConvert_SerializeObject_5.htm
The JsonSerializerSettings used to serialize the object. If this is null, default serialization settings will be used.

Are you saying that this still uses the global settings?

balag0: Yes. It still uses the global one.

pgombar (approved):

src/WebJobs.Script.WebHost/Management/MeshServiceClient.cs
fabiocav: Sounds like we might have a different issue somewhere. I'm also concerned about positioning this as an option that would unblock the modification of global serialization settings, as there are other code paths that rely on that as well.

balag0: Yeah there could be other places that needs similar fix as well as they are using global settings. This PR addresses one of those known issues that is resulting in CRIs. 

(Not trying to solve the problem of unblocking the modification of global serialization settings at all here :) This is much smaller in scope.)

fabiocav: I'm concerned this would lead to a game of whack-a-mole. We'll address this just to have something down the road fail for the same reason and end up sprinkling similar code throughout the host. 

We should be crisp about this not being a supported pattern and work on a proper solution that would address the root of the problem.

fabiocav: To be clear, I agree that this needs to be addressed, but we need to have a more generic approach we can leverage from the different code paths relying on serializer settings.

balag0: I agree with that. But this could be a tactical fix? Do you have any alternative fixes we can try instead? The goal here is to reduce the CRIs resulting from this particular feature being broken.

brettsam (approved):

PR #7424, opened 2021-06-01
[pack] Reacting to ScriptHost configuration changes for AzureStorageProvider

karshinlin:

Resolves issue with custom configuration sources set by App not accessible by AzureStorageProvider. Added an event handler for AzureStorageProvider to update the used IConfiguration with that of the active JobHost.

### Pull request checklist

### Additional information

Additional PR information


src/WebJobs.Script/StorageProvider/IAzureStorageProvider.cs
```diff
@@ -13,8 +14,8 @@ public interface IAzureStorageProvider
     {
         bool TryGetBlobServiceClientFromConnectionString(out BlobServiceClient client, string connectionString);
 
-        bool TryGetBlobServiceClientFromConnection(out BlobServiceClient client, string connection);
+        bool TryGetBlobServiceClientFromConnection(out BlobServiceClient client, string connection, IConfiguration configurationOverride = null);
 
-        BlobContainerClient GetBlobContainerClient();
+        BlobContainerClient GetBlobContainerClient(IConfiguration configurationOverride = null);
```
brettsam, line 19: Should this be a required param, or is sending null okay in some cases? Looking at FunctionsSyncManager, for instance...

src/WebJobs.Script/StorageProvider/AzureStorageProvider.cs
```diff
@@ -18,16 +19,23 @@ namespace Microsoft.Azure.WebJobs.Script
...
             _configuration = configuration;
+            _serviceProvider = scriptHostManager as IServiceProvider;
             _storageOptions = options;
-            _logger = logger;
+
+            // Every time script host is re-intializing, we also need to re-initialize
+            // services that change with the scope of the script host.
+            scriptHostManager.HostInitializing += (s, e) =>
+            {
+                InitializeServices();
```
brettsam, line 37: This means that no WebHost-level service should be caching any of the blob clients created by this, right? B/c they can change whenever there's a new host? Which services were we worried about for that?

Not sure if I like that long-term, but as a fast, pragmatic fix, it should be okay.

src/WebJobs.Script/StorageProvider/IAzureStorageProvider.cs
karshinlin: So in most cases where IAzureStorageProvider is used, the IConfiguration will update with that of the active JobHost -- meaning this parameter isn't needed and null is okay. There's a case when the active JobHost isn't started yet but we need it's configuration (deciding between blob lock and in-memory lock manager during ScriptHost service initialization) when we need to pass the configuration as a parameter. 

FunctionsSyncManager is in the context of a controller which already has the right IConfiguration I believe (but the added eventhandler logic takes care of this too). FunctionsSyncManager relies on the BlobStorageSecretsRepository through dependency injection though, which is still using the WebHost configuration and has trouble accessing storage when customers don't set AzureWebJobsStorage AppSetting (in previous versions too).

src/WebJobs.Script/StorageProvider/AzureStorageProvider.cs
karshinlin: Services that use IAzureStorageProvider:
ScriptHost:
- BlobLeaseDistributedLockManager
- AzureScheduleStorageMonitor (caches blobContainerClient)

WebHost:
- BlobChangeAnalysisStateProvider (about to be removed)
- FunctionsSyncManager (caches blobContainerClient)
- BlobStorageSecretsRepository (caches blobClient)

We do need to consider the effects of caching a client if the IConfiguration changes. Do ScriptHost services get disposed and replaced when ScriptHost gets restarted? When does the outer WebHost get restarted?

karshinlin:
Note that a custom configuration sourcing AzureWebJobsStorage will still have problems with BlobStorageSecretsRepository (used by FunctionsSyncManager). This problem exists in 3.0.15584.0 and this change.

src/WebJobs.Script/StorageProvider/AzureStorageProvider.cs
brettsam: Yes, ScriptHost services are disposed and re-created when it restarts. The outer WebHost only disposes when the process exits.

brettsam (dismissed):
I'm good with this now, knowing that we'll have some refactoring coming in the future that means we can potentially revisit the fragility of how this works at WebHost/JobHost level.

Still want @fabiocav to weigh in

Hazhzeng:
HotFix branch **release/3.0-hotfix** is now even with **release/3.0**.
Please create a PR targeting **release/3.0-hotfix** as well once this PR got merged.

src/WebJobs.Script/StorageProvider/AzureStorageProvider.cs
```diff
@@ -49,25 +57,30 @@ public virtual bool TryGetBlobServiceClientFromConnectionString(out BlobServiceC
...
             {
                 throw new InvalidOperationException($"Could not create BlobServiceClient to obtain the BlobContainerClient using Connection: {ConnectionStringNames.Storage}");
             }
 
             return blobServiceClient.GetBlobContainerClient(ScriptConstants.AzureWebJobsHostsContainerName);
         }
+
+        private void InitializeServices()
+        {
+            _configuration = _serviceProvider.GetService<IConfiguration>();
```
fabiocav, line 83: This might benefit from deferral. No change required for now, but wanted to flag.

src/WebJobs.Script/StorageProvider/IAzureStorageProvider.cs
fabiocav: From the `FunctionsSyncManager` you should aways have access to `IConfiguration`, but that could come from different scopes depending on the scenario, but this simply means the the right instance will be passed to you automatically. 

Being defensive and correctly logging unexpected conditions (which is what Brett is concerned about) is advisable.

src/WebJobs.Script/StorageProvider/StorageClientProvider.cs
```diff
@@ -96,27 +95,29 @@ public virtual bool TryGetFromConnectionString(string connectionString, out TCli
...
+        public virtual TClient Get(string name, IConfiguration configuration)
         {
             if (string.IsNullOrWhiteSpace(name))
             {
                 name = ConnectionStringNames.Storage; // default
             }
 
             // $$$ Where does validation happen?
-            IConfigurationSection connectionSection = _configuration.GetWebJobsConnectionStringSection(name);
+            IConfigurationSection connectionSection = configuration.GetWebJobsConnectionStringSection(name);
```
fabiocav, line 120: This code should be more defensive. Even with the check that exists in `GetWebJobsConnectionStringSection` (actually indirectly in the method it ends up calling), not performing a `null` check here makes this a bit fragile. Please update to perform a check.

fabiocav (dismissed):
A few nits, but not blocking this for now. Just wanted to capture and we should address ASAP.

alrod:
@karshinlin,  Can you please add the fix to release notes?

alrod (approved):

eranjbarSuncor:
Will there be a similar fix for the Python function? Noticed that this version isn't available/can't be referenced by the Python function apps

PR #7448, opened 2021-06-10
Disposing of DependencyTrackingTelemetryModule before host is disposed

brettsam:
*Note -- I originally made a commit has the fix commented out so I can verify it fails in CI. It failed 100% of the time locally.*

This fixes the error:
```
The specified configuration does not have a telemetry channel. (Parameter 'configuration')
```

Tactical change to fix a race condition during disposal of the `DependencyTrackingTelemetryModule`. It must be disposed before `TelemetryConfiguration`. This change explicitly disposes it before disposing the host.

This issue is fixed in ApplicationInsights here -- https://github.com/microsoft/ApplicationInsights-dotnet/pull/2294 -- but this won't be released until 2.0.18. Once we pull that version, we can delete this fix (but keep the test).

### DIFF
--- a/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
+++ b/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
@@ -721,6 +721,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
         // A temporary fix until we are able to take App Insights 2.18.0. There is a potential
         // race during disposal where new DiagnosticListeners can throw exceptions after TelemetryConfiguration
         // is disposed, but before this module is disposed. This ensures the module disposes first.
+        // Tracking issue: https://github.com/Azure/azure-functions-host/issues/7450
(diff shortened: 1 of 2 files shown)


brettsam:
Related to #7446. 

/cc @cijothomas

brettsam:
Test failed as expected: https://azfunc.visualstudio.com/Azure%20Functions/_build/results?buildId=18651&view=ms.vss-test-web.build-test-results-tab&runId=1096486&resultId=100278&paneView=debug.

Fix is now uncommented.

cijothomas (dismissed):

fabiocav (dismissed):
The change looks good to me, but could we add a link to an issue tracking the permanent fix so we can reference that in the future when reviewing this code?

fabiocav (approved):

alejandromelis:
@brettsam @fabiocav 

When do you expect this workaround to be released?

PR #7476, opened 2021-06-18
[pack] Sourcing configuration from WebHost and Active JobHost for Storage operations

karshinlin:

Resolves issue of custom startups disrupting access to Storage connection strings depending on the scope of the service calling storage. 
- AzureStorageProvider keeps a configuration that maintains the WebHost configuration in addition to the Active JobHost's configuration. In a custom startup situation, AzureWebJobsStorage should exist in one of WebHost or JobHost (or both). 

### Pull request checklist

### Additional information

Additional PR information


src/WebJobs.Script/Config/ActiveHostConfigurationSource.cs
```diff
@@ -0,0 +1,72 @@
...
+        public IConfigurationProvider Build(IConfigurationBuilder builder)
+        {
+            return new ActiveHostConfigurationProvider(this);
+        }
+
+        private class ActiveHostConfigurationProvider : ConfigurationProvider
+        {
+            private readonly ActiveHostConfigurationSource _configurationSource;
+
+            public ActiveHostConfigurationProvider(ActiveHostConfigurationSource configurationSource)
```
brettsam: Can you pass the `ServiceProvider` here instead? It doesn't look like you need to access the `ConfigSource` at all.

src/WebJobs.Script/Config/ActiveHostConfigurationSource.cs
```diff
@@ -0,0 +1,72 @@
...
+                _configurationSource = configurationSource ?? throw new ArgumentNullException(nameof(configurationSource));
+            }
+
+            public override bool TryGet(string key, out string value)
+            {
+                if (_configurationSource?.ServiceProvider?.GetService(typeof(IConfiguration)) is IConfigurationRoot activeHostConfiguration)
+                {
+                    return
+                        (value = activeHostConfiguration.GetValue<string>(key)) != null ||
+                        (value = activeHostConfiguration.GetValue<string>(NormalizeKey(key))) != null;
```
brettsam: You shouldn't have to use `NormalizeKey()` -- this would only impact if you were reading directly from env vars.

src/WebJobs.Script/Config/ActiveHostConfigurationSource.cs
```diff
@@ -0,0 +1,72 @@
...
+                value = default;
+                return false;
+            }
+
+            public override IEnumerable<string> GetChildKeys(IEnumerable<string> earlierKeys, string parentPath)
+            {
+                var activeHostConfiguration = _configurationSource?.ServiceProvider?.GetService(typeof(IConfiguration)) as IConfigurationRoot;
+                var keys = new HashSet<string>();
+
+                if (activeHostConfiguration != null)
```
brettsam: I think you can use the pattern matching here like you did above.

`if (_configurationSource?.ServiceProvider?.GetService(typeof(IConfiguration)) is IConfigurationRoot activeHostConfiguration)`

src/WebJobs.Script.WebHost/Security/KeyManagement/DefaultSecretManagerProvider.cs
```diff
@@ -61,7 +61,7 @@ public sealed class DefaultSecretManagerProvider : ISecretManagerProvider
         internal ISecretsRepository CreateSecretsRepository()
         {
             string secretStorageType = Environment.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebJobsSecretStorageType);
-            var storageConnection = _configuration.GetWebJobsConnectionStringSection(ConnectionStringNames.Storage);
+            var storageConnection = _azureStorageProvider.Configuration.GetWebJobsConnectionStringSection(ConnectionStringNames.Storage);
```
fabiocav: Ideally, the storage provider wouldn't just expose the configuration and expect the logic to consume/validate is implemented by the consumer. We can eliminate the need to get configuration and expose the property by adding a method to check whether the provider is enabled (which, internally, would mean validating whether the connection string is setup) and replacing the `Exists` call below with that method instead.

src/WebJobs.Script/Config/ActiveHostConfigurationSource.cs
```diff
@@ -0,0 +1,72 @@
...
+namespace Microsoft.Azure.WebJobs.Script.Configuration
+{
+    public class ActiveHostConfigurationSource : IConfigurationSource
+    {
+        public ActiveHostConfigurationSource(IServiceProvider serviceProvider)
+        {
+            ServiceProvider = serviceProvider;
+        }
+
+        public IServiceProvider ServiceProvider { get; set; }
```
fabiocav: No need for a public property here.

src/WebJobs.Script/Config/ActiveHostConfigurationSource.cs
```diff
@@ -0,0 +1,72 @@
...
+
+        public IConfigurationProvider Build(IConfigurationBuilder builder)
+        {
+            return new ActiveHostConfigurationProvider(this);
```
fabiocav: With the current implementation, passing the service provider would be sufficient.

src/WebJobs.Script/Config/ActiveHostConfigurationSource.cs
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        private class ActiveHostConfigurationProvider : ConfigurationProvider
+        {
+            private readonly ActiveHostConfigurationSource _configurationSource;
+
+            public ActiveHostConfigurationProvider(ActiveHostConfigurationSource configurationSource)
+            {
+                _configurationSource = configurationSource ?? throw new ArgumentNullException(nameof(configurationSource));
```
fabiocav: Take the service provider here.

src/WebJobs.Script/Config/ActiveHostConfigurationSource.cs
```diff
@@ -0,0 +1,72 @@
...
+            }
+
+            public override bool TryGet(string key, out string value)
+            {
+                if (_configurationSource?.ServiceProvider?.GetService(typeof(IConfiguration)) is IConfigurationRoot activeHostConfiguration)
```
fabiocav: this call will be simplified when you take the service provider, and the null check will also not be needed.

src/WebJobs.Script/Config/ActiveHostConfigurationSource.cs
```diff
@@ -0,0 +1,72 @@
...
+                        (value = activeHostConfiguration.GetValue<string>(NormalizeKey(key))) != null;
+                }
+
+                value = default;
+                return false;
+            }
+
+            public override IEnumerable<string> GetChildKeys(IEnumerable<string> earlierKeys, string parentPath)
+            {
+                var activeHostConfiguration = _configurationSource?.ServiceProvider?.GetService(typeof(IConfiguration)) as IConfigurationRoot;
```
fabiocav: Same as above with this call.

src/WebJobs.Script/Config/ActiveHostConfigurationSource.cs
```diff
@@ -0,0 +1,72 @@
...
+                var keys = new HashSet<string>();
```
fabiocav: No need to create an instance here.

Instantiate if the configuration is not `null`, otherwise; return an `Enumerable.Empty<string>()`

src/WebJobs.Script/ScriptConstants.cs
```diff
@@ -112,6 +112,7 @@ public static class ScriptConstants
         public const string FeatureFlagEnableEnhancedScopes = "EnableEnhancedScopes";
         public const string FeatureFlagDisableDevInDebug = "DisableDevModeInDebug";
         public const string FeatureFlagEnableDiagnosticEventLogging = "EnableDiagnosticEventLogging";
+        public const string FeatureFlagDisableMergedWebHostScriptHostConfiguration = "DisableMergedConfiguration";
```
fabiocav, line 115: Can we also create a GitHub issue to remove this in a future release?

src/WebJobs.Script/StorageProvider/AzureStorageProvider.cs
```diff
@@ -18,16 +20,37 @@ namespace Microsoft.Azure.WebJobs.Script
...
             _storageOptions = options;
             _logger = logger;
+
+            if (FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagDisableMergedWebHostScriptHostConfiguration))
+            {
+                _configuration = configuration;
+            }
+            else
+            {
+                _configuration = new ConfigurationBuilder()
```
fabiocav, line 41: nit: indentation

src/WebJobs.Script/StorageProvider/AzureStorageProvider.cs
```diff
@@ -18,16 +20,37 @@ namespace Microsoft.Azure.WebJobs.Script
...
+                .AddConfiguration(configuration)
+                .Add(new ActiveHostConfigurationSource(_serviceProvider))
+                .Build();
+            }
+        }
+
+        public IConfiguration Configuration
```
fabiocav: As noted above, hopefully, after the suggested changes, this property won't be required.

src/WebJobs.Script/StorageProvider/IAzureStorageProvider.cs
```diff
@@ -11,6 +12,8 @@ namespace Microsoft.Azure.WebJobs.Script
     /// </summary>
     public interface IAzureStorageProvider
     {
+        IConfiguration Configuration { get; }
```
fabiocav: See comment about the property.

fabiocav (commented):
Initial round of feedback

brettsam (commented):
Some initial comments -- want to think through some more and will update later today.

src/WebJobs.Script/Config/ActiveHostConfigurationSource.cs
```diff
@@ -30,17 +31,41 @@ public ActiveHostConfigurationProvider(ActiveHostConfigurationSource configurati
...
+                    }
                 }
 
-                value = default;
-                return false;
+                keys.UnionWith(earlierKeys);
+                return keys;
+            }
+
+            private static string NormalizeKey(string key)
```
fabiocav: Can you describe the scenario you're trying to address by normalizing this? The limitation is already a behavior that is in place with the current configuration functionality, that behavior shouldn't change (i.e. a setting with a `:` in a different platform shouldn't suddenly start working).

src/WebJobs.Script/Config/ActiveHostConfigurationSource.cs
karshinlin: If `activeHostConfiguration == null`, shouldn't it still return earlierKeys? That's how I interpreted this: https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.iconfigurationprovider.getchildkeys?view=dotnet-plat-ext-5.0#Microsoft_Extensions_Configuration_IConfigurationProvider_GetChildKeys_System_Collections_Generic_IEnumerable_System_String__System_String_

src/WebJobs.Script/ScriptConstants.cs
karshinlin: https://github.com/Azure/azure-functions-host/issues/7483
Created issue.

src/WebJobs.Script/Config/ActiveHostConfigurationSource.cs
karshinlin: Removing NormalizeKey() here - as Brett mentioned this is only needed when reading from EnvironmentVariables directly and the EnvironmentVariableConfigurationProvider already handles both formats.

src/WebJobs.Script/StorageProvider/IAzureStorageProvider.cs
karshinlin: Removed this property and added ConnectionExists() method to AzureStorageProvider.

src/WebJobs.Script.WebHost/Security/KeyManagement/DefaultSecretManagerProvider.cs
karshinlin: Added ConnectionExists public method to IAzureStorageProvider that attempts to create the BlobServiceClient and returns true/false.

src/WebJobs.Script/Config/ActiveHostConfigurationSource.cs
```diff
@@ -0,0 +1,60 @@
...
+
+            public ActiveHostConfigurationProvider(IScriptHostManager scriptHostManager)
+            {
+                _scriptHostManager = scriptHostManager ?? throw new ArgumentNullException(nameof(scriptHostManager));
+                _scriptHostManager.ActiveHostChanged += ScriptHostManager_ActiveHostChanged;
+            }
+
+            public override void Load()
+            {
+                if ((_scriptHostManager as IServiceProvider)?.GetService(typeof(IConfiguration)) is IConfigurationRoot activeHostConfiguration)
```
fabiocav, line 37: Avoid the cast here by having your variable be of type `IServiceProvider`, as that is all this code needs.

src/WebJobs.Script/Config/ActiveHostConfigurationSource.cs
```diff
@@ -0,0 +1,60 @@
...
+                        }
+                    }
+
+                    _changeTokenRegistration?.Dispose();
+                    _changeTokenRegistration = activeHostConfiguration.GetReloadToken().RegisterChangeCallback(_ => Load(), null);
+                    OnReload();
+                }
+            }
+
+            private void ScriptHostManager_ActiveHostChanged(object sender, ActiveHostChangedEventArgs e)
```
fabiocav: nit: Method name -> Something like `HandleActiveHostChange`

src/WebJobs.Script/StorageProvider/StorageClientProvider.cs
```diff
@@ -177,12 +178,12 @@ protected virtual bool TryGetServiceUri(IConfiguration configuration, out Uri se
...
+                if ((accountName = configuration?.GetValue<string>("accountName")) != null)
                 {
                     serviceUri = FormatServiceUri(accountName);
                     return true;
                 }
-                else if ((uriStr = configuration.GetValue<string>(serviceUriConfig)) != null)
+                else if ((uriStr = configuration?.GetValue<string>(serviceUriConfig)) != null)
                 {
                     serviceUri = new Uri(uriStr);
                     return true;
```
fabiocav, line 189: Should we have a debug log here to indicate when we were unable to resolve the service URI?

karshinlin: We try to resolve the serviceUri every time we create a client so adding a log here would add a good bit of noise. If we should be able to resolve a serviceUri, but can't in this method, we'd run into an exception in Microsoft.Extensions.Azure which would be caught and logged in this file in the TryGet() method.

src/WebJobs.Script/Config/ActiveHostConfigurationSource.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            _scriptHostManager = scriptHostManager;
+        }
+
+        public IConfigurationProvider Build(IConfigurationBuilder builder)
+        {
+            return new ActiveHostConfigurationProvider(_scriptHostManager);
+        }
+
+        private class ActiveHostConfigurationProvider : ConfigurationProvider
```
brettsam: Also implement `IDisposable` here -- and in the `Dispose()` method, dispose the `_changeTokenRegistration` and also unregister `ActiveHostChanged` with `scriptHostManager.ActiveHostChanged -= HandleActiveHostChange`

src/WebJobs.Script/StorageProvider/AzureStorageProvider.cs
```diff
@@ -18,16 +20,37 @@ namespace Microsoft.Azure.WebJobs.Script
...
+                    .Build();
+            }
+        }
+
+        /// <summary>
+        /// Checks if a BlobServiceClient can be created (indicates the specified connection can be parsed)
+        /// </summary>
+        /// <param name="connection">connection to use for the BlobServiceClient</param>
+        /// <returns>true if successful; false otherwise</returns>
+        public bool ConnectionExists(string connection)
```
brettsam, line 53: Should this be renamed to something like `CanParseConnection()` if it's just checking the parse-ability?

test/WebJobs.Script.Tests.Integration/Storage/AzureStorageProviderTests.cs
```diff
@@ -0,0 +1,234 @@
...
+    public class AzureStorageProviderTests
+    {
+        private const string StorageConnection = "AzureWebJobsStorage";
+
+        [Fact]
+        public async Task TestAzureStorageProvider_ConnectionInWebHostConfiguration()
+        {
+            var testData = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase)
+            {
+                { "ConnectionStrings:AzureWebJobsStorage", Environment.GetEnvironmentVariable(StorageConnection) },
```
brettsam: Instead of grabbing this straight from `Environment` -- use `TestHelpers.GetTestConfiguration()` and pull it from there. That combines env vars with an optional test settings file (which is what I use locally) -- so it's more flexible, and what other tests do.

test/WebJobs.Script.Tests.Integration/Storage/AzureStorageProviderTests.cs
```diff
@@ -0,0 +1,234 @@
...
+            {
+                { "AzureWebJobsStorage", "" }
+            };
+
+            using (new TestScopedEnvironmentVariable(testData))
+            {
+                var webHostConfiguration = new ConfigurationBuilder()
+                    .AddEnvironmentVariables()
+                    .AddTestSettings()
+                    .Build();
```
brettsam: Nit -- You should be able to replace these lines (here and elsewhere) with `TestHelpers.GetTestConfiguration()`.

test/WebJobs.Script.Tests.Integration/Storage/AzureStorageProviderTests.cs
```diff
@@ -0,0 +1,234 @@
...
+
+                var inMemory = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase)
+                {
+                    { connectionName, $"DefaultEndpointsProtocol=https;AccountName=jobHostAccount;AccountKey={encodedString};EndpointSuffix=core.windows.net" },
+                };
+                var jobHostConfiguration = new ConfigurationBuilder()
+                    .AddInMemoryCollection(inMemory)
+                    .AddTestSettings()
+                    .Build();
```
brettsam: Do you want `AddTestSettings()` here? That adds a test settings file that we support for tests. If you don't have one, this test will pass, but if you do it may fail. May just want to remove `AddTestSettings()` completely here (and elsewhere that you use this pattern).

test/WebJobs.Script.Tests/Configuration/ActiveHostConfigurationSourceTests.cs
```diff
@@ -0,0 +1,106 @@
...
+                { "ConnectionStrings__ConnStr1", "str1" },
+                { "ConnectionStrings__ConnStr2", "str2" },
+                { "SectionA:SectionB", "middleValue" },
+                { "SectionA:SectionB:SectionC", "finalValue" }
+            };
+
+            using (new TestScopedEnvironmentVariable(testData))
+            {
+                var testActiveHostConfig = new ConfigurationBuilder()
+                    .AddEnvironmentVariables()
```
brettsam: Can you use `InMemory` for this instead of env vars? In general, we prefer that as it relies less upon what someone has set up on their machine. (then you don't need the TestScopedEnvVar either).

test/WebJobs.Script.Tests/Configuration/ActiveHostConfigurationSourceTests.cs
```diff
@@ -0,0 +1,106 @@
...
+using System.Collections.Generic;
+using Microsoft.Azure.WebJobs.Script.Configuration;
+using Microsoft.Azure.WebJobs.Script.Extensions;
+using Microsoft.Extensions.Configuration;
+using Moq;
+using Xunit;
+
+namespace Microsoft.Azure.WebJobs.Script.Tests.Configuration
+{
+    public class ActiveHostConfigurationSourceTests
```
brettsam, line 14: Nice tests! One more suggestion for these -- create a configuration with two InMemory dictionaries registered -- where the second one overrides values in the first, and ensure that it retrieves the correct one.

That would specifically be checking the code in `Load()` where you check for `ContainsKey`.

brettsam (commented):
Looking good! Another round of reviews....

src/WebJobs.Script/Config/ActiveHostConfigurationSource.cs
karshinlin: Since we make this ConfigurationProvider implement IDisposable, we'll need the scriptHostManager reference to unregister the event handler and dispose the change token.

test/WebJobs.Script.Tests.Integration/Storage/AzureStorageProviderTests.cs
karshinlin: I removed AddTestSettings in this file since we should only be using InMemoryCollection anyways.

src/WebJobs.Script/StorageProvider/AzureStorageProvider.cs
brettsam: Also, is there any way to parse this without creating a client? That may be a bunch of work during startup that we don't really need to do if we just care about whether a string can be parsed.

brettsam (commented):
Another comment...

brettsam (approved):

fabiocav (approved):

PR #7406, opened 2021-05-25
[pack]Dynamic Concurrency support

mathewc:
Implements the Dynamic Concurrency features in https://github.com/Azure/azure-functions-host/issues/7300. See supporting SDK PR [here](https://github.com/Azure/azure-webjobs-sdk/pull/2720). Track 2 Azure Storage Extension PR is [here](https://github.com/Azure/azure-sdk-for-net/pull/22283).

### Pull request checklist

* [x] My changes **do not** require documentation changes
    * [ ] Otherwise: Documentation issue linked to PR
* [x] My changes **should not** be added to the release notes for the next release
    * [ ] Otherwise: I've added my notes to `release_notes.md`
* [x] My changes **do not** need to be backported to a previous version
    * [ ] Otherwise: Backport tracked by issue/PR #issue_or_pr
* [x] I have added all required tests (Unit tests, E2E tests)

PR #7477, opened 2021-06-21
[pack] Language workers concurrency

alrod:
Feature details: #7499

The PR includes:
 * Monitoring “WorkerStatus” latency and adding a new worker depending on "WorkerStatus" latency.
 * Removing existing concurrency limits.
 
Feature is opt-in by an app setting and it's possible to modify concurrency manager parameters. Without enabling the feature everything works in the old way. Also the PR needs to be updated after merging "Dynamic concurrency" Mathew is working on:
https://github.com/Azure/azure-functions-host/pull/7406

### Pull request checklist

### Additional information

Additional PR information


src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -148,8 +155,48 @@ public async Task StartWorkerProcessAsync()
...
 
+            TimeSpan? workerStatusLatency = await GetWorkerStatusResponseAsync(null);
+            if (workerStatusLatency != null)
+            {
+                workerStatus.Latency = workerStatusLatency.Value;
+                _workerChannelLogger.LogDebug($"[HostMonitor] Worker status request took {workerStatusLatency.Value.TotalMilliseconds}ms");
+            }
+
+            // get the process stats for the worker
+            var workerProcessStats = _rpcWorkerProcess.GetStats();
```
mathewc: WorkerStatus.ProcessStats (including the ProcessStats class itself) are going away in https://github.com/Azure/azure-functions-host/pull/7406. All process monitoring for the host is now done in the WebJobs SDK, and child process stats are already logged there (code [here](https://github.com/Azure/azure-webjobs-sdk/blob/7bfcd07e520210c9cc092e6c1ac9413fc5c3d178/src/Microsoft.Azure.WebJobs.Host/Scale/DefaultHostProcessMonitor.cs#L103)). Those states are also feeding into ConcurrencyManager throttle decisions.

So you should remove any usages of ProcessStats. Likely, this PR will need to go in after mine - we'll need to coordinate.

src/WebJobs.Script/Workers/Rpc/RpcWorkerConcurrencyManager.cs
```diff
@@ -0,0 +1,243 @@
...
+                double proportion = (double)overloadedCount / _concurrencyOptions.Value.HistorySize;
+
+                return proportion >= _concurrencyOptions.Value.HistoryThreshold;
+            }
+            return false;
+        }
+
+        internal string LogWorkerState(WorkerDescription desc)
+        {
+            string formattedLoadHistory = string.Empty, formattedLatencyHistory = string.Empty;
```
mathewc: As mentioned above, all this process stats logging should be removed. It'll already be logged by the WebJobs SDK HostProcessMonitor peridically.

src/WebJobs.Script/Workers/Rpc/RpcWorkerConcurrencyManager.cs
```diff
@@ -0,0 +1,243 @@
...
+                    _logger.LogDebug(sb.ToString());
+                }
+            }
+
+            return result;
+        }
+
+        internal bool IsOverloaded(WorkerStatus status)
+        {
+            if (status.RpcWorkerStats.LatencyHistory.Count() >= _concurrencyOptions.Value.HistorySize)
```
mathewc: So only Latency is currently being used to determine whether to scale out workers. I think the right way to do this integration with the new ConcurrencyManager stuff would be to:

- implement the WorkerChannelThrottleProvider shell I provide in my host DC PR [here](https://github.com/Azure/azure-functions-host/pull/7406/files#diff-185f0f62183c68e8362f1bdcfe08241606f2917c7874cc740c27c6cda4f0023cR15)
- the throttle provider could own the timer that periodically pings all the channels (via existing IFunctionInvocationDispatcher.GetWorkerStatusesAsync). It can then maintain latency history per channel
- it can use these latencies to determine whether the throttle is Enabled. When it returns its ConcurrencyThrottleStatus it'll add its "WorkerChannel" name to the EnabledThrottles collection
- This RpcWorkerConcurrencyService can then take an instance of IConcurrencyThrottleManager on construction, allowing it to check whether the "WorkerChannel" throttle is enabled at any time.

The benefit of putting the latency/throttle logic in the throttle provider is that it's already plugged into ConcurrencyManager, so in addition to you being able to use it for worker scale out, when Enabled it'll also limit function concurrency. While the channels are overloaded we'll ramp down work, and CM and your new worker scale out logic will work together.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -32,7 +33,7 @@
 
 namespace Microsoft.Azure.WebJobs.Script.Grpc
 {
-    internal class GrpcWorkerChannel : IRpcWorkerChannel, IDisposable
+    internal class GrpcWorkerChannel : IRpcWorkerChannel, IDisposable, IRpcWorkerConcurrencyChannel
```
mathewc: What about HttpWorkerChannel? If you follow my suggestions below, things will work for both Grpc and Http workers, since the throttle provider goes through the IFunctionInvocationDispatcher interface which hides the underlying worker/channel implementation.

src/WebJobs.Script/Workers/Rpc/IRpcWorkerConcurrencyChannel.cs
```diff
@@ -0,0 +1,17 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
+{
+    internal interface IRpcWorkerConcurrencyChannel
```
mathewc: I don't think you need this new interface at all. Channels already implement GetWorkerStatusAsync and we should be able to return everything we need via WorkerStatus.

src/WebJobs.Script/Workers/Rpc/RpcWorkerConcurrencyManager.cs
```diff
@@ -0,0 +1,243 @@
...
+            {
+                int overloadedCount = status.RpcWorkerStats.LatencyHistory.Where(x => x.TotalMilliseconds >= _concurrencyOptions.Value.LatencyThreshold.TotalMilliseconds).Count();
```
mathewc: Any reason you chose to do the calculation this way, as opposed to just computing the average latency and seeing whether that is over the threshold?

src/WebJobs.Script/Config/RpcWorkerConcurrencyOptions.cs
```diff
@@ -0,0 +1,78 @@
...
+            // Setting deafault values
+            LatencyThreshold = TimeSpan.FromSeconds(1);
+            AdjustmentPeriod = TimeSpan.FromSeconds(10);
+            CheckInterval = TimeSpan.FromSeconds(1);
+            HistorySize = 10;
+            HistoryThreshold = 1F;
+        }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether language worker concurrancy
```
mathewc: nit: fix "concurrancy" spelling throughout

src/WebJobs.Script/Config/RpcWorkerConcurrencyOptions.cs
```diff
@@ -0,0 +1,78 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs.Script
+{
+    public class RpcWorkerConcurrencyOptions : Hosting.IOptionsFormatter
```
mathewc: This shouldn't be Rpc specific right, since we want these worker options to apply to Http workers as well? E.g. WorkerConcurrencyOptions?

src/WebJobs.Script/Config/RpcWorkerConcurrencyOptions.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        /// <summary>
+        /// Gets or sets interval to check language worker channel state.
+        /// </summary>
+        public TimeSpan CheckInterval { get; set; }
+
+        /// <summary>
+        /// Gets or sets the history size to store language workers channel states.
+        /// </summary>
+        public int HistorySize { get; set; }
```
mathewc: I don't think we should expose some of these publicly - probably more than we want users to have to think about. E.g. AdjustmentPeriod, CheckInterval, HistorySize, HistoryThreshold could all be internal for now I think. It's our job to determine the right numbers for these that work. For ConcurrencyManager, I made many of these non-configurable constants for now, e.g. [MinSampleCount](https://github.com/Azure/azure-webjobs-sdk/blob/7bfcd07e520210c9cc092e6c1ac9413fc5c3d178/src/Microsoft.Azure.WebJobs.Host/Scale/DefaultHostProcessMonitor.cs#L19), and many of the core concurrency values [here](https://github.com/Azure/azure-webjobs-sdk/blob/7bfcd07e520210c9cc092e6c1ac9413fc5c3d178/src/Microsoft.Azure.WebJobs.Host/Scale/ConcurrencyStatus.cs#L15).

src/WebJobs.Script/Config/RpcWorkerConcurrencyOptionsSetup.cs
```diff
@@ -0,0 +1,52 @@
...
+                        return;
+                    }
+
+                    options.Enabled = true;
+                    // Configure language worker concurrency options from IConfiguration
+                    _configuration.GetSection(nameof(RpcWorkerConcurrencyOptions)).Bind(options);
+
+                    if (options.MaxWorkerCount == 0)
+                    {
+                        int effectiveCores = (_environment.IsConsumptionSku() && !_environment.IsVMSS()) ? 1 : Environment.ProcessorCount;
```
mathewc: Use Utility.GetEffectiveCoresCount rather than copying this code

src/WebJobs.Script/Workers/Rpc/RpcWorkerConstants.cs
```diff
@@ -51,5 +51,12 @@ public static class RpcWorkerConstants
         public const string DotNetExecutableName = "dotnet";
         public const string DotNetExecutableNameWithExtension = DotNetExecutableName + ".exe";
         public const string DotNetFolderName = "dotnet";
+
+        // Concurrancy settings
+        public const string PythonTreadpoolThreadCount = "PYTHON_THREADPOOL_THREAD_COUNT";
+        public const string PSWorkerInProcConcurrencyUpperBound = "PSWorkerInProcConcurrencyUpperBound";
+        public const string DefaultConcurrecnyPython = "32";
```
mathewc: spelling

src/WebJobs.Script/Workers/ProcessManagement/DefaultWorkerProcessFactory.cs
```diff
@@ -89,5 +96,22 @@ internal string SanitizeExpandedArgument(string envExpandedString)
             }
             return envExpandedString;
         }
+
+        internal void RemoveConcurrencyLimits(ProcessStartInfo startInfo)
```
mathewc: Should this instead be named ApplyWorkerConcurrencyLimits?

mathewc: Also, perhaps rather than overriding these from the outside, we could instead flow via env var to the worker that dynamic concurrency is enabled and internally it can default as needed?

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -314,7 +320,7 @@ internal async Task<IEnumerable<IRpcWorkerChannel>> GetInitializedWorkerChannels
         {
             IEnumerable<IRpcWorkerChannel> workerChannels = await GetAllWorkerChannelsAsync();
             IEnumerable<IRpcWorkerChannel> initializedWorkers = workerChannels.Where(ch => ch.IsChannelReadyForInvocations());
-            if (initializedWorkers.Count() > _maxProcessCount)
+            if (initializedWorkers.Count() > _maxProcessCount && !_concurrencyOptions.Value.Enabled)
```
mathewc: Should you also be checking the options.MaxWorkerCount here? Probably where we initialize _maxProcessCount above we should be applying options.MaxWorkerCount when enabled? That way we're also configuring ErrorEventsThreshold properly as well.

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -398,13 +404,18 @@ internal bool ShouldRestartWorkerChannel(string runtime, bool isWebHostChannel,
             return string.Equals(_workerRuntime, runtime, StringComparison.InvariantCultureIgnoreCase) && (isWebHostChannel || isJobHostChannel);
         }
 
-        private async Task RestartWorkerChannel(string runtime)
+        internal async Task RestartWorkerChannel(string runtime)
```
mathewc: Strange that this is called "Restart" when it's actually just starting a new channel. Should we rename this to make it clearer?

src/WebJobs.Script/Workers/Rpc/RpcWorkerConcurrencyManager.cs
```diff
@@ -0,0 +1,243 @@
...
+        }
+
+        public async Task StartAsync(CancellationToken cancellationToken)
+        {
+            if (_concurrencyOptions.Value.Enabled)
+            {
+                // Delay monitoring
+                await Task.Delay(_concurrencyOptions.Value.AdjustmentPeriod);
+
+                _logger.LogDebug($"Staring language worker concurrency monitoring. Options: {_concurrencyOptions.Value.Format()}");
```
mathewc: spelling

src/WebJobs.Script/Workers/Rpc/RpcWorkerConcurrencyManager.cs
```diff
@@ -0,0 +1,243 @@
...
+        {
+            if (_concurrencyOptions.Value.Enabled && _timer != null)
+            {
+                _logger.LogDebug("Stoping language worker concurrency monitoring.");
+                _timer.Stop();
+            }
+            return Task.CompletedTask;
+        }
+
+        //public void Start()
```
mathewc: remove dead code

src/WebJobs.Script/Workers/Rpc/RpcWorkerConcurrencyManager.cs
```diff
@@ -0,0 +1,243 @@
...
+    internal class RpcWorkerConcurrencyManager : IHostedService, IDisposable
+    {
+        private readonly TimeSpan _logStateInterval = TimeSpan.FromSeconds(60);
+        private readonly IOptions<RpcWorkerConcurrencyOptions> _concurrencyOptions;
+        private readonly ILogger _logger;
+        private readonly IFunctionInvocationDispatcherFactory _functionInvocationDispatcherFactory;
+
+        private RpcFunctionInvocationDispatcher _dispatcher;
+        private System.Timers.Timer _timer;
+        private Stopwatch _addWorkeStopWatch = Stopwatch.StartNew();
```
mathewc: spelling

src/WebJobs.Script/Workers/Rpc/RpcWorkerConcurrencyManager.cs
```diff
@@ -0,0 +1,243 @@
...
+        private Stopwatch _logStateStopWatch = Stopwatch.StartNew();
+        private bool _disposed = false;
+        private object syncObj = new object();
```
mathewc: _syncLock

src/WebJobs.Script/Workers/Rpc/RpcWorkerConstants.cs
```diff
@@ -51,5 +51,12 @@ public static class RpcWorkerConstants
...
+        public const string DefaultConcurrencyPS = "100";
+        public const string FunctionWorkerConcurrencyEnabledSettingName = "FUNCTIONS_WORKER_CONCURRENCY_ENABLED";
```
mathewc: FUNCTIONS_WORKER_DYNAMIC_CONCURRENCY_ENABLED

src/WebJobs.Script/Workers/ProcessManagement/DefaultWorkerProcessFactory.cs
alrod: We decided that only function host decides when adjust the level of concurrency. @AnatoliB, what do you think?

src/WebJobs.Script/Config/RpcWorkerConcurrencyOptions.cs
alrod: These settings are exposed publicly for experimenting without redeploying, once we find the best values they will become internal.

src/WebJobs.Script/Workers/Rpc/RpcWorkerConcurrencyManager.cs
alrod: My idea was that one very long WorkerStatus call can affect overall average latency. Having oveloadedCount seems more reliable approach to me.

src/WebJobs.Script/Workers/ProcessManagement/DefaultWorkerProcessFactory.cs
AnatoliB: I like the the word "Remove" a bit more here. Conceptually, removing the limits is the actual intent. Technically, passing very high values is the way to achieve this goal, but this is just an implementation detail (simply because this is the only knob the workers have, there is no way to say "no limit").

We could also pass a boolean flag, like @mathewc is proposing. This will also work, but I can see two problems with this approach:

1. Worker concurrency will now be controlled by _two_ config values instead of one, and the relationship between them is not absolutely obvious. We would have to make sure every language worker (currently - at least Python and PowerShell) handles this pair properly.
2. In order to implement this logic, we'll have to rebuild and re-release every language worker. If we ever want to change this logic, we'll have to rebuild and re-release every language worker again.

With this PR, the situation remains very simple: every worker has a single and trivial concurrency limit knob, and the host is the only component making decisions, so all the relevant changes will be in the host. This will make experimentation a bit easier now, and this is the main reason I like it.

src/WebJobs.Script/Workers/Rpc/RpcWorkerConstants.cs
```diff
@@ -51,5 +51,12 @@ public static class RpcWorkerConstants
         public const string DotNetExecutableName = "dotnet";
         public const string DotNetExecutableNameWithExtension = DotNetExecutableName + ".exe";
         public const string DotNetFolderName = "dotnet";
+
+        // Concurrancy settings
+        public const string PythonTreadpoolThreadCount = "PYTHON_THREADPOOL_THREAD_COUNT";
+        public const string PSWorkerInProcConcurrencyUpperBound = "PSWorkerInProcConcurrencyUpperBound";
+        public const string DefaultConcurrecnyPython = "32";
+        public const string DefaultConcurrencyPS = "100";
```
AnatoliB: I really doubt we need separate values for Python and PowerShell. Why would we assume Python can handle less threads than PowerShell? I think the main idea of what we are currently doing about language worker concurrency is that we want to stop _assuming without evidence_. If the latency is not growing, we are not running out of memory, etc., let's keep increasing concurrency, because why not? Until we start executing functions, we simply don't have data for meaningful decisions per language, and we should stop pretending that we do. The limit here should not be something that we carefully design, this is just a very large number for an extra sanity check. So, I would suggest we keep a single number and increase it even more aggressively: 1000 or 10,000.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
alrod: I will update the PR once #7406 is merged.

src/WebJobs.Script/Workers/Rpc/RpcWorkerConcurrencyManager.cs
alrod: I will update the PR once #7406 is merged.

src/WebJobs.Script/Workers/Rpc/IRpcWorkerConcurrencyChannel.cs
alrod: Removed the interface.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
alrod: HttpWorkerChannel is included.

src/WebJobs.Script/Config/WorkerConcurrencyOptionsSetup.cs
```diff
@@ -0,0 +1,50 @@
...
+                        return;
+                    }
+
+                    options.Enabled = true;
+                    // Configure worker concurrency options from IConfiguration
+                    _configuration.GetSection(nameof(WorkerConcurrencyOptions)).Bind(options);
+
+                    if (options.MaxWorkerCount == 0)
+                    {
+                        options.MaxWorkerCount = (_environment.GetEffectiveCoresCount() * 2) + 2;
```
AnatoliB: I'm not convinced we should calculate a limit based on the number of cores. If the latency is not growing and we have enough memory, why can't we spawn more processes? This is especially important for sync functions in Python, which are frequently affected by GIL contention within a single process. In these cases, adding a new process is likely to resolve the bottleneck. If we worry about overwhelming the CPU with context switching, this should be taken care of by us monitoring the execution latency: from this perspective, adding a new process is not that different from adding a thread.

src/WebJobs.Script/Workers/Rpc/RpcWorkerConcurrencyManager.cs
alrod: The integration with `WorkerChannelThrottleProvider` is planned in a next PR for language worker concurrency.

src/WebJobs.Script/Workers/ProcessManagement/DefaultWorkerProcessFactory.cs
alrod: Having function host as only point for making worker concurrency decisions is better as it simplifies testing and we do not need to change the workers. We can change it in future if needed.

src/WebJobs.Script/Workers/Rpc/RpcWorkerConstants.cs
alrod: For python 3.8 and it takes max concurrency=32. I tried 33 and the concurrency resets to 1. I do not know who limits the concurrency python worker or python runtime. Need to chat with Varad/Roger about this.

For PS I will change it to 1000.

src/WebJobs.Script/Config/RpcWorkerConcurrencyOptions.cs
mathewc: I see - so you're saying that you won't check them in as public and this is just draft for your testing?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
alrod: This is updated.

src/WebJobs.Script/Workers/Rpc/RpcWorkerConcurrencyManager.cs
alrod: This is updated.

src/WebJobs.Script/Config/RpcWorkerConcurrencyOptions.cs
alrod: Fixed.

src/WebJobs.Script/Config/RpcWorkerConcurrencyOptions.cs
alrod: We want to test in production after the PR merge and have ability to change the settings. After the testing is done we can hardcode the values with fixed values.

src/WebJobs.Script/Workers/Rpc/RpcWorkerConstants.cs
alrod: 1. There was a bug in the Python worker, sync up with Varad about current max concurrency.
2. Most likely in future the host would be sending "concurrency capability" for the worker and the worker will set the internal concurrency himself.

After discussing offline with @AnatoliB we decided to leave it until the workers change.

src/WebJobs.Script/Config/WorkerConcurrencyOptionsSetup.cs
alrod: After discussing offline with @AnatoliB we decided to leave the formula as is and change in the future if necessary.

src/WebJobs.Script/Workers/Rpc/RpcWorkerConstants.cs
alrod: The fix is not in Python worker yet so we want to go with DefaultConcurrecnyPython=32 for now.

src/WebJobs.Script/Config/WorkerConcurrencyOptionsSetup.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        public WorkerConcurrencyOptionsSetup(IConfiguration configuration, IEnvironment environment)
+        {
+            _configuration = configuration;
+            _environment = environment;
+        }
+
+        public void Configure(WorkerConcurrencyOptions options)
+        {
+            if (bool.TryParse(_environment.GetEnvironmentVariable(RpcWorkerConstants.FunctionWorkerDynamicConcurrencyEnabledSettingName), out bool enabled))
```
mathewc: Why do we have both an app setting and a config option WorkerConcurrencyOptions.Enabled? Seems to me you can do away with the app setting and just have users enable via host.json, similar to what we do for Dynamic Concurrency. There's no app setting for DC either.

src/WebJobs.Script/Config/WorkerConcurrencyOptionsSetup.cs
```diff
@@ -0,0 +1,50 @@
...
+            {
+                // Do not enable convurrency if any concurrency settings are defined
```
mathewc: Why can't we allow worker dynamic concurrency (WDC) enablement to override any other concurrency related settings? That way, customers who may have already configured things can try WDC by enabling a flag, without having to remove all their existing settings. This enables flipping back and forth via a single flag.

mathewc: spelling "convurrency"

src/WebJobs.Script/Config/WorkerConcurrencyOptions.cs
```diff
@@ -0,0 +1,78 @@
...
+        /// <summary>
+        /// Gets or sets the latency threshold dictating when worker channel is overloaded.
+        /// </summary>
+        public TimeSpan LatencyThreshold { get; set; }
+
+        /// <summary>
+        /// Gets or sets the time when worker
+        /// channels will start to monitor again after adding a new worker.
+        /// </summary>
+        public TimeSpan AdjustmentPeriod { get; set; }
```
mathewc, line 33: I think you should keep AdjustmentPeriod, CheckInterval, HistorySize internal. These aren't things users should have to think about.

src/WebJobs.Script/Config/WorkerConcurrencyOptions.cs
```diff
@@ -0,0 +1,78 @@
...
+            // Setting deafault values
+            LatencyThreshold = TimeSpan.FromSeconds(1);
+            AdjustmentPeriod = TimeSpan.FromSeconds(10);
+            CheckInterval = TimeSpan.FromSeconds(1);
+            HistorySize = 10;
+            HistoryThreshold = 1F;
+        }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether worker concurrency
```
mathewc: I think in user doc we should refer to this as "worker dynamic concurrency" to make it clear that we're talking about dynamic adjustment of worker counts

src/WebJobs.Script/Config/WorkerConcurrencyOptions.cs
```diff
@@ -0,0 +1,78 @@
...
+        /// Gets or sets the history size to store workers channel states.
+        /// </summary>
+        public int HistorySize { get; set; }
+
+        /// <summary>
+        /// Gets or sets the history threshold.
+        /// E. g. value equal to 1.0 means all states in history should be
+        /// overloaded to consider worker state as overloaded.
+        /// </summary>
+        public float HistoryThreshold { get; set; }
```
mathewc: Trying to come up with a more descriptive name for this. Perhaps NewWorkerThreshold? I'd describe in this xml doc as "Gets or sets the threshold dictating when a new worker will be added. Value should be between 0 and 1 indicating the percentage of overloaded channel latency samples required to trigger a addition of a new worker".

mathewc: Also should add a range check on this

src/WebJobs.Script/Config/WorkerConcurrencyOptions.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        /// <summary>
+        /// Gets or sets the max count of workers.
+        /// It will be set depending on SKU if equal to 0.
+        /// </summary>
+        public int MaxWorkerCount { get; set; }
```
mathewc, line 57: Add a range check on this as well?

src/WebJobs.Script/Workers/FunctionInvocationDispatcherFactory.cs
```diff
@@ -30,7 +30,8 @@ internal class FunctionInvocationDispatcherFactory : IFunctionInvocationDispatch
             IWebHostRpcWorkerChannelManager webHostLanguageWorkerChannelManager,
             IJobHostRpcWorkerChannelManager jobHostLanguageWorkerChannelManager,
             IOptions<ManagedDependencyOptions> managedDependencyOptions,
-            IRpcFunctionInvocationDispatcherLoadBalancer functionDispatcherLoadBalancer)
+            IRpcFunctionInvocationDispatcherLoadBalancer functionDispatcherLoadBalancer,
+            IOptions<WorkerConcurrencyOptions> concurrencyOptions)
```
mathewc: nit: here and elsewhere, name the param workerConcurrencyOptions? To distinguish between ConcurrencyOptions, which we also use in places?

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -24,12 +24,14 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
     internal class RpcFunctionInvocationDispatcher : IFunctionInvocationDispatcher
     {
         private readonly IMetricsLogger _metricsLogger;
+        private readonly ILoggerFactory _loggerFactory;
         private readonly ILogger _logger;
         private readonly IRpcWorkerChannelFactory _rpcWorkerChannelFactory;
         private readonly IEnvironment _environment;
         private readonly IApplicationLifetime _applicationLifetime;
-        private readonly SemaphoreSlim _restartWorkerProcessSLock = new SemaphoreSlim(1, 1);
+        private readonly SemaphoreSlim _startWorkerProcessSLock = new SemaphoreSlim(1, 1);
```
mathewc: spelling "ProcessS"

src/WebJobs.Script/Config/WorkerConcurrencyOptions.cs
```diff
@@ -0,0 +1,78 @@
...
+            CheckInterval = TimeSpan.FromSeconds(1);
+            HistorySize = 10;
+            HistoryThreshold = 1F;
+        }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether worker concurrency
+        /// is enabled.
+        /// </summary>
+        public bool Enabled { get; set; }
```
mathewc: Rename to DynamicConcurrencyEnabled?

src/WebJobs.Script/Workers/Http/HttpFunctionInvocationDispatcher.cs
```diff
@@ -119,6 +124,12 @@ public async void WorkerRestart(HttpWorkerRestartEvent workerRestart)
             }
         }
 
+        public async Task StartWorkerChannel()
+        {
+            // currenly only one worker
+            await Task.CompletedTask;
```
mathewc: nit: remove async, and just return Task.CompletedTask

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -146,6 +152,11 @@ private void StartWorkerProcesses(int startIndex, Func<Task> startAction)
                 for (var count = startIndex; count < _maxProcessCount
                     && !_processStartCancellationToken.IsCancellationRequested; count++)
                 {
+                    if (_concurrencyOptions.Value.Enabled && count > 0)
+                    {
+                        // Make sure only one worker is started if concurancy is enabled
```
mathewc: spelling

mathewc: Also, I'd move this check after the start action below and before the delay since we don't need to delay. Consider breaking rather than returning, in case any logic is added after the loop in the future.

src/WebJobs.Script/Workers/Rpc/RpcWorkerConstants.cs
```diff
@@ -51,5 +51,12 @@ public static class RpcWorkerConstants
         public const string DotNetExecutableName = "dotnet";
         public const string DotNetExecutableNameWithExtension = DotNetExecutableName + ".exe";
         public const string DotNetFolderName = "dotnet";
+
+        // Concurrancy settings
```
mathewc: spelling

src/WebJobs.Script/Workers/Rpc/RpcWorkerConstants.cs
```diff
@@ -51,5 +51,12 @@ public static class RpcWorkerConstants
...
+        public const string PythonTreadpoolThreadCount = "PYTHON_THREADPOOL_THREAD_COUNT";
```
mathewc: spelling

src/WebJobs.Script/Workers/WorkerChannelMonitor.cs
```diff
@@ -0,0 +1,113 @@
...
+
+using System;
+using System.Collections.Generic;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.WebJobs.Script.Workers
+{
+    internal class WorkerChannelMonitor : IDisposable
+    {
+        private readonly List<TimeSpan> _workerStatusLatecyHistory = new List<TimeSpan>();
```
mathewc: spelling

src/WebJobs.Script/Workers/WorkerChannelMonitor.cs
```diff
@@ -0,0 +1,113 @@
...
+            }
+
+            try
+            {
+                WorkerStatus workerStatus = await _channel.GetWorkerStatusAsync();
+                AddSample(_workerStatusLatecyHistory, workerStatus.Latency);
+            }
+            catch
+            {
+                // Don't allow backround execptions to escape
```
mathewc: spelling

src/WebJobs.Script/Workers/WorkerConcurrencyManager.cs
```diff
@@ -0,0 +1,218 @@
...
+        private Stopwatch _logStateStopWatch = Stopwatch.StartNew();
+        private bool _disposed = false;
+
+        public WorkerConcurrencyManager(IFunctionInvocationDispatcherFactory functionInvocationDispatcherFactory,
+            IOptions<WorkerConcurrencyOptions> concurrencyOptions, ILoggerFactory loggerFactory)
+        {
+            _concurrencyOptions = concurrencyOptions ?? throw new ArgumentNullException(nameof(concurrencyOptions));
+            _functionInvocationDispatcherFactory = functionInvocationDispatcherFactory ?? throw new ArgumentNullException(nameof(functionInvocationDispatcherFactory));
+
+            _logger = loggerFactory?.CreateLogger<WorkerConcurrencyManager>();
```
mathewc: Wonder if you shouldn't log to the same LogCategories.Concurrency category that Dynamic Concurrency logs to?

src/WebJobs.Script/Workers/WorkerConcurrencyManager.cs
```diff
@@ -0,0 +1,218 @@
...
+        }
+
+        public async Task StartAsync(CancellationToken cancellationToken)
+        {
+            if (_concurrencyOptions.Value.Enabled)
+            {
+                // Delay monitoring
+                await Task.Delay(_concurrencyOptions.Value.AdjustmentPeriod);
```
mathewc: This will delay host startup and impact coldstart. Can't do this delay here. Shouldn't be needed though, since you're already checking in the timer callback?

src/WebJobs.Script/Workers/WorkerConcurrencyManager.cs
```diff
@@ -0,0 +1,218 @@
...
+
+                _functionInvocationDispatcher = _functionInvocationDispatcherFactory.GetFunctionDispatcher();
+                if (_functionInvocationDispatcher is HttpFunctionInvocationDispatcher)
+                {
+                    _logger.LogDebug($"Http worker concurrency is not supported.");
+                    return;
+                }
+
+                _logger.LogDebug($"Starting worker concurrency monitoring. Options: {_concurrencyOptions.Value.Format()}");
```
mathewc: Options logger already logs these options, so don't need to write them again here

src/WebJobs.Script/Workers/WorkerConcurrencyManager.cs
```diff
@@ -0,0 +1,218 @@
...
+        {
+            if (_concurrencyOptions.Value.Enabled)
+            {
+                // Delay monitoring
+                await Task.Delay(_concurrencyOptions.Value.AdjustmentPeriod);
+
+                _functionInvocationDispatcher = _functionInvocationDispatcherFactory.GetFunctionDispatcher();
+                if (_functionInvocationDispatcher is HttpFunctionInvocationDispatcher)
+                {
+                    _logger.LogDebug($"Http worker concurrency is not supported.");
```
mathewc: Should this instead be an error earlier during host build? Seems like we can identify this earlier and error out.

src/WebJobs.Script/Workers/WorkerConcurrencyManager.cs
mathewc: Also, maybe reword "Starting worker dynamic concurrency monitoring", adding "dynamic" here and elsewhere where you log this feature name.

src/WebJobs.Script/Workers/WorkerConcurrencyManager.cs
```diff
@@ -0,0 +1,218 @@
...
+                    AutoReset = false,
+                    Interval = _concurrencyOptions.Value.CheckInterval.TotalMilliseconds,
+                };
+
+                _timer.Elapsed += OnTimer;
+                _timer.Start();
+            }
+            else
+            {
+                _logger.LogDebug($"Language worker concurrency is disabled.");
```
mathewc: Consider removing this. Only need to log when the feature is enabled.

src/WebJobs.Script/Workers/WorkerConcurrencyManager.cs
```diff
@@ -0,0 +1,218 @@
...
+            }
+
+            await Task.CompletedTask;
+        }
+
+        public Task StopAsync(CancellationToken cancellationToken)
+        {
+            if (_concurrencyOptions.Value.Enabled && _timer != null)
+            {
+                _logger.LogDebug("Stoping language worker concurrency monitoring.");
```
mathewc: I'd just remove this log - doesn't add any value

src/WebJobs.Script/Workers/WorkerConcurrencyManager.cs
```diff
@@ -0,0 +1,218 @@
...
+                _timer.Stop();
+            }
+            return Task.CompletedTask;
+        }
+
+        internal async void OnTimer(object sender, System.Timers.ElapsedEventArgs e)
+        {
+            if (_disposed && _functionInvocationDispatcher == null)
```
mathewc: Shouldn't we return any time we're disposed? I.e. remove the dispatcher check.

src/WebJobs.Script/Workers/WorkerConcurrencyManager.cs
```diff
@@ -0,0 +1,218 @@
...
+            }
+            catch (Exception ex)
+            {
+                // don't allow background exceptions to escape
+                _logger.LogError(ex.ToString());
+            }
+            _timer.Start();
+        }
+
+        internal bool AddWorkerIfNeeded(IDictionary<string, WorkerStatus> workerStatuses, TimeSpan elaspsedFromLastAdding)
```
mathewc: This method doesn't add a worker. Either rename to something like NewWorkerIsRequired, or move all the logic above into this method so it does both the check and the addition.

src/WebJobs.Script/Workers/WorkerConcurrencyManager.cs
```diff
@@ -0,0 +1,218 @@
...
+    internal class WorkerConcurrencyManager : IHostedService, IDisposable
+    {
+        private readonly TimeSpan _logStateInterval = TimeSpan.FromSeconds(60);
+        private readonly IOptions<WorkerConcurrencyOptions> _concurrencyOptions;
+        private readonly ILogger _logger;
+        private readonly IFunctionInvocationDispatcherFactory _functionInvocationDispatcherFactory;
+
+        private IFunctionInvocationDispatcher _functionInvocationDispatcher;
+        private System.Timers.Timer _timer;
+        private Stopwatch _addWorkerStopWatch = Stopwatch.StartNew();
```
mathewc: nit: rename Stopwatch vars by fixing casing "Stopwatch"

src/WebJobs.Script/Workers/WorkerConcurrencyManager.cs
```diff
@@ -0,0 +1,218 @@
...
+            {
+                // don't allow background exceptions to escape
+                _logger.LogError(ex.ToString());
+            }
+            _timer.Start();
+        }
+
+        internal bool AddWorkerIfNeeded(IDictionary<string, WorkerStatus> workerStatuses, TimeSpan elaspsedFromLastAdding)
+        {
+            if (elaspsedFromLastAdding < _concurrencyOptions.Value.AdjustmentPeriod)
```
mathewc: spelling "elaspsed". Maybe rename to "timeSinceLastNewWorker"

src/WebJobs.Script/Workers/WorkerConcurrencyManager.cs
```diff
@@ -0,0 +1,218 @@
...
+            {
+                return false;
+            }
+
+            bool result = false;
+            if (workerStatuses.All(x => x.Value.IsReady))
+            {
+                // Check how many channels are oveloaded
```
mathewc: spelling

src/WebJobs.Script/Workers/WorkerConcurrencyManager.cs
```diff
@@ -0,0 +1,218 @@
...
+            {
+                int overloadedCount = status.WorkerStats.LatencyHistory.Where(x => x.TotalMilliseconds >= _concurrencyOptions.Value.LatencyThreshold.TotalMilliseconds).Count();
+                double proportion = (double)overloadedCount / _concurrencyOptions.Value.HistorySize;
+
+                return proportion >= _concurrencyOptions.Value.HistoryThreshold;
+            }
+            return false;
+        }
+
+        internal string LogWorkerState(WorkerDescription desc)
```
mathewc: rename to FormatWorkerDescription since you're not logging here. Better yet, make this a Format method on WorkerDescription itself.

src/WebJobs.Script/Workers/WorkerConcurrencyManager.cs
```diff
@@ -0,0 +1,218 @@
...
+                _disposed = true;
+            }
+        }
+
+        public void Dispose()
+        {
+            Dispose(true);
+        }
+
+        internal class WorkerDescription
```
mathewc: nit: perhaps rename to WorkerStatusDetails?

src/WebJobs.Script/Workers/WorkerConcurrencyManager.cs
```diff
@@ -0,0 +1,218 @@
...
+        {
+            public string WorkerId { get; set; }
+
+            public WorkerStatus WorkerStatus { get; set; }
+
+            public bool Overloaded { get; set; }
```
mathewc: nit: IsOverloaded

src/WebJobs.Script/Workers/WorkerStatus.cs
```diff
@@ -15,5 +16,15 @@ public class WorkerStatus
         /// Gets or sets the current latency for worker channel.
         /// </summary>
         public TimeSpan Latency { get; set; }
+
+        /// <summary>
+        /// Gets or sets the worker latency statistic.
```
mathewc: spelling. Also, this description is incorrect. It gets or sets the WorkerStats

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -173,6 +179,12 @@ public async Task<WorkerStatus> GetWorkerStatusAsync()
                 }
             }
 
+            if (_concurrencyOptions.Value.Enabled)
+            {
+                workerStatus.IsReady = IsChannelReadyForInvocations();
+                workerStatus.WorkerStats = _monitor.GetStats();
```
mathewc: You should remove the separate WorkerChannelMonitor and WorkerStats types. It'd be cleaner to just incorporate the latency history samples as part of WorkerStatus. That's what it was designed for and already has some latency info. GrpcWorkerChannel can just include the timer and sample the channel periodically. As it stands now, you have a weird interdependency between your monitor and the channel, where a call to this GetWorkerStatuses method is accessing monitor.GetStats to formulate the WorkerStatus, and internally the monitor is itself calling GetWorkerStatuses. That's convoluted.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -173,6 +179,12 @@ public async Task<WorkerStatus> GetWorkerStatusAsync()
                 }
             }
 
+            if (_concurrencyOptions.Value.Enabled)
+            {
+                workerStatus.IsReady = IsChannelReadyForInvocations();
```
mathewc: Should we set IsReady regardless of whether WDC is enabled? Doesn't seem tied to DC.

src/WebJobs.Script/Config/WorkerConcurrencyOptions.cs
alrod: Those settings will be internal, we want to expose them to give us more space for experiments after the change will be deployed to PROD.

src/WebJobs.Script/Config/WorkerConcurrencyOptionsSetup.cs
alrod: added 2 settings:
```
RemoveConcurrencyLimits 
DynamicConcurrencyEnabled
```

src/WebJobs.Script/Workers/WorkerConcurrencyManager.cs
alrod: good catch

src/WebJobs.Script/Workers/WorkerConcurrencyManager.cs
alrod: it would be helpful to have those logs to see how customers do not want to enable the WDC

src/WebJobs.Script/Config/WorkerConcurrencyOptionsSetup.cs
```diff
@@ -0,0 +1,42 @@
...
+        public WorkerConcurrencyOptionsSetup(IConfiguration configuration, IEnvironment environment)
+        {
+            _configuration = configuration;
+            _environment = environment;
+        }
+
+        public void Configure(WorkerConcurrencyOptions options)
+        {
+            if (string.IsNullOrEmpty(_environment.GetEnvironmentVariable(RpcWorkerConstants.FunctionsWorkerProcessCountSettingName))
+               && string.IsNullOrEmpty(_environment.GetEnvironmentVariable(RpcWorkerConstants.PythonThreadpoolThreadCount))
```
AnatoliB: Does this mean that the presence of either PythonThreadpoolThreadCount or PSWorkerInProcConcurrencyUpperBound app settings effectively turns the feature off? I think we should be able to preserve the feature while respecting these settings. I'm ok with merging this code as is, but let's take a note to discuss/fix this later.

Not sure about FunctionsWorkerProcessCountSettingName - this one probably should turn the feature off.

src/WebJobs.Script/Config/WorkerConcurrencyOptions.cs
```diff
@@ -0,0 +1,80 @@
...
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs.Script
+{
+    public class WorkerConcurrencyOptions : Hosting.IOptionsFormatter
+    {
+        public WorkerConcurrencyOptions()
+        {
+            // Setting deafault values
+            LatencyThreshold = TimeSpan.FromSeconds(1);
```
AnatoliB: Is this hardcoded and never changes? Any plans to adjust it dynamically?

test/WebJobs.Script.Tests.Integration/Rpc/WorkerConcurancyManagerEndToEndTests.cs
```diff
@@ -0,0 +1,86 @@
...
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Script.Config;
+using Microsoft.Azure.WebJobs.Script.Eventing;
+using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
+using Microsoft.Extensions.Options;
+using Xunit;
+
+namespace Microsoft.Azure.WebJobs.Script.Tests
+{
+    public class WorkerConcurancyManagerEndToEndTests : IClassFixture<WorkerConcurancyManagerEndToEndTests.TestFixture>
```
AnatoliB: "Concurrency"

test/WebJobs.Script.Tests/Workers/Rpc/WorkerConcurancyManagerTest.cs
```diff
@@ -0,0 +1,314 @@
...
+using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Microsoft.WebJobs.Script.Tests;
+using Moq;
+using Xunit;
+
+namespace Microsoft.Azure.WebJobs.Script.Tests.Workers
+{
+    public class WorkerConcurancyManagerTest
```
AnatoliB: "Concurrency"

src/WebJobs.Script/Config/WorkerConcurrencyOptionsSetup.cs
alrod: Added 2 flags:
```
RemoveConcurrencyLimits 
DynamicConcurrencyEnabled
```

src/WebJobs.Script/Config/WorkerConcurrencyOptions.cs
alrod: This can be changed in `host.json`

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
alrod: this check needs to be before the start action as we already started one process

src/WebJobs.Script/Config/WorkerConcurrencyOptions.cs
mathewc: I don't understand why you say they'll be internal if you're exposing them publicly now. We can't remove them later.

src/WebJobs.Script/Utility.cs
```diff
@@ -261,6 +262,11 @@ public static bool IsValidUserType(Type type)
             return data.ToDictionary(p => p.Key, p => p.Value != null ? p.Value.ToString() : null, StringComparer.OrdinalIgnoreCase);
         }
 
+        public static string GetValue(this StringDictionary dictionary, string key)
```
mathewc: nit: GetValueOrNull

src/WebJobs.Script/Workers/IWorkerChannel.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.Script.Workers
+{
+    public interface IWorkerChannel
+    {
+        string Id { get; }
```
mathewc, line 11: Are there any other common methods all channels have that should be included in this interface, e.g. StartWorkerProcessAsync?

src/WebJobs.Script/Workers/ProcessManagement/DefaultWorkerProcessFactory.cs
```diff
@@ -89,5 +100,21 @@ internal string SanitizeExpandedArgument(string envExpandedString)
             }
             return envExpandedString;
         }
+
+        internal void ApplyWorkerConcurrencyLimits(ProcessStartInfo startInfo)
+        {
+            // Set higher concurrency limits for Python and Powershell language workers for V4
```
mathewc: You mention this is only for v4 here, but this PR targets dev?

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
```diff
@@ -139,17 +140,13 @@ internal void AddProvider(string workerDir)
...
-                            Arguments = arguments
-                        };
                         var rpcWorkerconfig = new RpcWorkerConfig()
                         {
                             Description = workerDescription,
                             Arguments = arguments,
-                            CountOptions = workerProcessCount
+                            CountOptions = workerProcessCount,
                         };
+                        AddWorkerConcuurencyOptions(rpcWorkerconfig);
```
mathewc: spelling

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
```diff
@@ -237,5 +234,34 @@ internal bool ShouldAddWorkerConfig(string workerDescriptionLanguage)
...
+            var workerConcurrencySection = jobHostSection.GetSection(ConfigurationSectionNames.WorkerConcurrency);
+            if (workerConcurrencySection.Exists())
+            {
+                workerConcurrencySection.Bind(options);
+
+                if (options.DynamicConcurrencyEnabled)
+                {
+                    if (!string.IsNullOrEmpty(_environment.GetEnvironmentVariable(RpcWorkerConstants.FunctionsWorkerProcessCountSettingName)))
+                    {
+                        options.DynamicConcurrencyEnabled = false;
```
mathewc: Shouldn't DynamicConcurrencyEnabled override any explicit FunctionsWorkerProcessCount value? e.g. an existing customer may have that and other preexisting concurrency settings configured, and we want them to be able to simply flip a single switch and enable/disable WDC. Otherwise, they'll have to go in and remove settings to try out WDC?

test/WebJobs.Script.Tests/Workers/Rpc/RpcWorkerConfigFactoryTests.cs
```diff
@@ -248,5 +249,91 @@ public void GetWorkerProcessCount_ThrowsException_Tests()
             var resultEx3 = Assert.Throws<ArgumentOutOfRangeException>(() => rpcWorkerConfigFactory.GetWorkerProcessCount(workerConfig));
             Assert.Contains("The TimeSpan must not be negative", resultEx3.Message);
         }
+
+        [Theory]
+        [InlineData(null, null, false)]
+        [InlineData("1", "true", false)]
+        [InlineData(null, "true", true)]
+        public void AddWorkerConcuurencyOptions_Works_As_Expected_DynamicConcurrency(
```
mathewc: spelling

test/WebJobs.Script.Tests/Workers/Rpc/RpcWorkerConfigFactoryTests.cs
```diff
@@ -248,5 +249,91 @@ public void GetWorkerProcessCount_ThrowsException_Tests()
...
+            };
+            rpcWorkerConfigFactory.AddWorkerConcuurencyOptions(rpcWorkerConfig);
+
+            if (rpcWorkerConfig.ConcurrencyOptions != null)
+            {
+                Assert.Equal(rpcWorkerConfig.ConcurrencyOptions.DynamicConcurrencyEnabled, dynamicConcurrencyEnabledValueExpected);
+            }
+        }
+
+        //[Theory]
```
mathewc: remove or enable

test/WebJobs.Script.Tests/Workers/Rpc/WorkerConcurrencyManagerTest.cs
```diff
@@ -0,0 +1,512 @@
...
+                    new int[] { 150, 150, 150, 150, 150 },
+                    true,
+                    false,
+                    TimeSpan.FromSeconds(2000),
+                    false
+                }
+            };
+
+        [Fact]
+        public async Task Start_StartsTimer()
```
mathewc: Start_StartsTimer_WhenDynamicConcurrencyEnabled

test/WebJobs.Script.Tests/Workers/Rpc/WorkerConcurrencyManagerTest.cs
```diff
@@ -0,0 +1,512 @@
...
+
+            await TestHelpers.Await(() =>
+            {
+                var sratedLog = _loggerProvider.GetAllLogMessages().FirstOrDefault(x => x.FormattedMessage.StartsWith("Starting dynamic worker concurrency monitoring."));
+                return sratedLog != null;
+            }, pollingInterval: 1000, timeout: 10 * 1000);
+        }
+
+        [Fact]
+        public async Task Start_DoesNot_StartTimer()
```
mathewc: _WhenDynamicConcurrencyDisabled

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
alrod: No, it should be vise versa: if a customer already have FunctionsWorkerProcessCount specified we should not enable dynamic concurrency as the customer already have the number of workers configured. Yes, the customer need to remove FunctionsWorkerProcessCount in order to use WDC.

src/WebJobs.Script/Workers/ProcessManagement/DefaultWorkerProcessFactory.cs
alrod: it was copy pasted from: https://github.com/Azure/azure-functions-host/commit/d45892af1d049d066edb92b449fe131ffc062776
removed the comment

src/WebJobs.Script/Workers/IWorkerChannel.cs
alrod: No

`StartWorkerProcessAsync `has different signature:
```
IRpcWorkerChannel -  Task StartWorkerProcessAsync();
IHttpWorkerChannel -  Task StartWorkerProcessAsync(CancellationToken cancellationToken = default);
```

mathewc: We should be able to just converge on the second signature taking the cancellation token

src/WebJobs.Script/Workers/WorkerConcurrencyManager.cs
```diff
@@ -0,0 +1,225 @@
...
+                {
+                    await _functionInvocationDispatcher.StartWorkerChannel();
+                    _logger.LogDebug("New worker is added.");
+                    _addWorkerStopwatch.Restart();
+                }
+            }
+            catch (Exception ex)
+            {
+                // don't allow background exceptions to escape
+                _logger.LogError(ex.ToString());
```
mathewc: Recommend using the LogError(ex, "Error monitoring worker concurrency")

src/WebJobs.Script/Workers/WorkerConcurrencyManager.cs
```diff
@@ -0,0 +1,225 @@
...
+            }
+
+            try
+            {
+                var workerStatuses = await _functionInvocationDispatcher.GetWorkerStatusesAsync();
+
+                if (NewWorkerIsRequired(workerStatuses, _addWorkerStopwatch.Elapsed))
+                {
+                    await _functionInvocationDispatcher.StartWorkerChannel();
+                    _logger.LogDebug("New worker is added.");
```
mathewc: Should this be Info level? Seems like an important log.

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
```diff
@@ -139,18 +140,14 @@ internal void AddProvider(string workerDir)
...
+                        var rpcWorkerConfig = new RpcWorkerConfig()
                         {
                             Description = workerDescription,
                             Arguments = arguments,
-                            CountOptions = workerProcessCount
+                            CountOptions = workerProcessCount,
                         };
-                        _workerDescripionDictionary[workerDescription.Language] = rpcWorkerconfig;
+                        AddWorkerConcuurencyOptions(rpcWorkerConfig);
+                        _workerDescripionDictionary[workerDescription.Language] = rpcWorkerConfig;
```
mathewc: spelling

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -157,6 +161,11 @@ private void StartWorkerProcesses(int startIndex, Func<Task> startAction, bool i
                 for (var count = startIndex; count < _maxProcessCount
                     && !_processStartCancellationToken.IsCancellationRequested; count++)
                 {
+                    if (_workerConfig.ConcurrencyOptions != null && _workerConfig.ConcurrencyOptions.DynamicConcurrencyEnabled && count > 0)
+                    {
+                        // Make sure only one worker is started of concurrency is enabled
```
mathewc: "if dynamic concurrency is enabled"

src/WebJobs.Script/Workers/WorkerConcurrencyManager.cs
```diff
@@ -0,0 +1,225 @@
...
+                        IsOverloaded = overloaded
+                    });
+                }
+
+                int overloadedCount = descriptions.Where(x => x.IsOverloaded == true).Count();
+                if (overloadedCount > 0)
+                {
+                    if (workerStatuses.Count() < _workerConcurrencyOptions.MaxWorkerCount)
+                    {
+                        _logger.LogDebug($"Adding a new worker, overloaded workers = {overloadedCount}, initialized workers = {workerStatuses.Count()} ");
```
mathewc: The caller is already logging. I think this log should be removed, and update the other log if needed. This method doesn't add the worker, it just checks if one needs to be added

src/WebJobs.Script/Workers/WorkerConcurrencyManager.cs
```diff
@@ -0,0 +1,225 @@
...
+                if (result == true || _logStateStopWatch.Elapsed > _logStateInterval)
+                {
+                    StringBuilder sb = new StringBuilder();
+                    foreach (WorkerStatusDetails description in descriptions)
+                    {
+                        sb.Append(FormatWorkerDescription(description));
+                        sb.Append(Environment.NewLine);
+                    }
+                    _logStateStopWatch.Restart();
+                    _logger.LogInformation(sb.ToString());
```
mathewc: Should this be Debug level?

src/WebJobs.Script/Config/HostJsonFileConfigurationSource.cs
```diff
@@ -51,7 +51,7 @@ public class HostJsonFileConfigurationProvider : ConfigurationProvider
             {
                 "version", "functionTimeout", "retry", "functions", "http", "watchDirectories", "watchFiles", "queues", "serviceBus",
                 "eventHub", "singleton", "logging", "aggregator", "healthMonitor", "extensionBundle", "managedDependencies",
-                "customHandler", "httpWorker", "extensions"
+                "customHandler", "httpWorker", "extensions", "workerConcurrency"
```
mathewc: Can you also add "concurrency" to this list? I forgot to do that for DC.

src/WebJobs.Script/Config/WorkerConcurrencyOptions.cs
```diff
@@ -0,0 +1,74 @@
...
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs.Script
+{
+    public class WorkerConcurrencyOptions : Hosting.IOptionsFormatter
+    {
+        public WorkerConcurrencyOptions()
+        {
+            // Setting deafault values
```
mathewc: spelling

src/WebJobs.Script/Config/WorkerConcurrencyOptions.cs
```diff
@@ -0,0 +1,74 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.ComponentModel.DataAnnotations;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs.Script
+{
+    public class WorkerConcurrencyOptions : Hosting.IOptionsFormatter
```
mathewc: nit: can you mark this internal? Also rather than newing up instances in services, you should still be able to add this to DI via services.ConfigureOptions - it just doesn't get bound to host.json. Going this route, I'd remove the IOptionsFormatter implementation so the options aren't logged (since they're static and not meant for user config).

src/WebJobs.Script/Config/WorkerConcurrencyOptions.cs
```diff
@@ -0,0 +1,74 @@
...
+    {
+        public WorkerConcurrencyOptions()
+        {
+            // Setting deafault values
+            LatencyThreshold = TimeSpan.FromMilliseconds(100);
+            AdjustmentPeriod = TimeSpan.FromSeconds(10);
+            CheckInterval = TimeSpan.FromSeconds(1);
+            HistorySize = 10;
+            NewWorkerThreshold = 0.2F;
+            MaxWorkerCount = 10;
```
mathewc, line 21: Just so I understand - the max cap on dynamic workers is 10 when WDC is enabled?

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -28,8 +28,9 @@ internal class RpcFunctionInvocationDispatcher : IFunctionInvocationDispatcher
         private readonly IRpcWorkerChannelFactory _rpcWorkerChannelFactory;
         private readonly IEnvironment _environment;
         private readonly IApplicationLifetime _applicationLifetime;
-        private readonly SemaphoreSlim _restartWorkerProcessSLock = new SemaphoreSlim(1, 1);
+        private readonly SemaphoreSlim _startWorkerProcessLock = new SemaphoreSlim(1, 1);
         private readonly TimeSpan _thresholdBetweenRestarts = TimeSpan.FromMinutes(WorkerConstants.WorkerRestartErrorIntervalThresholdInMinutes);
+        private readonly WorkerConcurrencyOptions _workerConcurrencyOptions = new WorkerConcurrencyOptions();
```
mathewc: Recommend taking this via DI in the ctor which will facilitate tests

src/WebJobs.Script/Workers/WorkerConcurrencyManager.cs
```diff
@@ -0,0 +1,218 @@
...
+namespace Microsoft.Azure.WebJobs.Script.Workers
+{
+    internal class WorkerConcurrencyManager : IHostedService, IDisposable
+    {
+        private readonly TimeSpan _logStateInterval = TimeSpan.FromSeconds(60);
+        private readonly ILogger _logger;
+        private readonly IFunctionInvocationDispatcherFactory _functionInvocationDispatcherFactory;
+        private readonly IEnvironment _environment;
+
+        private WorkerConcurrencyOptions _workerConcurrencyOptions = new WorkerConcurrencyOptions();
```
mathewc: Recommend taking this via DI in the ctor which will facilitate tests

src/WebJobs.Script/Workers/WorkerConcurrencyManager.cs
```diff
@@ -0,0 +1,218 @@
...
+        public WorkerConcurrencyManager(IFunctionInvocationDispatcherFactory functionInvocationDispatcherFactory,
+            IEnvironment environment, ILoggerFactory loggerFactory)
+        {
+            _functionInvocationDispatcherFactory = functionInvocationDispatcherFactory ?? throw new ArgumentNullException(nameof(functionInvocationDispatcherFactory));
+            _environment = environment ?? throw new ArgumentNullException(nameof(environment));
+
+            _logger = loggerFactory?.CreateLogger(LogCategories.Concurrency);
+        }
+
+        public WorkerConcurrencyOptions WorkerConcurrencyOptions
```
mathewc: Once you're taking from DI, you won't need this setter in tests so it can be removed.

src/WebJobs.Script/Config/WorkerConcurrencyOptions.cs
```diff
@@ -0,0 +1,74 @@
...
+        /// Value should be between 0 and 1 indicating the percentage of overloaded channel latency samples required to trigger a addition of a new worker
+        /// </summary>
+        [Range(typeof(float), "0F", "1F", ErrorMessage = "Value for {0} must be between {1} and {2}.")]
+        public float NewWorkerThreshold { get; set; }
+
+        /// <summary>
+        /// Gets or sets the max count of workers.
+        /// It will be set depending on SKU if equal to 0.
+        /// </summary>
+        [Range(typeof(int), "0", "100", ErrorMessage = "Value for {0} must be between {1} and {2}.")]
```
mathewc: 0 is no longer a valid value, right? It's no longer used as a sentinel value?

src/WebJobs.Script/Config/WorkerConcurrencyOptions.cs
alrod: yes, we are starting with 10 as FUNCTIONS_WORKER_PROCESS_COUNT has 10 max(  https://docs.microsoft.com/en-us/azure/azure-functions/functions-app-settings#functions_worker_process_count)
 it can be changed in future.

src/WebJobs.Script/Config/WorkerConcurrencyOptions.cs
alrod: yes, changed to 1

src/WebJobs.Script/Config/WorkerConcurrencyOptions.cs
alrod: Can not make it internal as it's needed on a test as input.

Changed to:
```
                services.AddOptions<WorkerConcurrencyOptions>()
                    .Configure<IConfiguration>((o, c) =>
                    {
                        o = new WorkerConcurrencyOptions();
                    });
```

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -40,6 +41,8 @@ internal class GrpcWorkerChannel : IRpcWorkerChannel, IDisposable
         private readonly IEnvironment _environment;
         private readonly IOptionsMonitor<ScriptApplicationHostOptions> _applicationHostOptions;
         private readonly ISharedMemoryManager _sharedMemoryManager;
+        private readonly List<TimeSpan> _workerStatusLatencyHistory = new List<TimeSpan>();
+        private readonly WorkerConcurrencyOptions _workerConcurrencyOptions = new WorkerConcurrencyOptions();
```
mathewc: take via DI

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -287,6 +288,11 @@ public static IHostBuilder AddScriptHostCore(this IHostBuilder builder, ScriptAp
                          .GetSection(ConfigurationSectionNames.Scale)
                          .Bind(o);
                     });
+                services.AddOptions<WorkerConcurrencyOptions>()
```
mathewc: Doesn't just services.AddOptions<WorkerConcurrencyOptions>() work? Shouldn't need the separate .Configure call I believe. Just let DI create the instance.

src/WebJobs.Script/Config/WorkerConcurrencyOptions.cs
mathewc: Tests have internal access so that wouldn't be an issue, however the real reason you can't make it internal is because it's defined in WebJobs.Script but needed in WebJobs.Script.Grpc, so must be public.

mathewc (approved):

PR #7491, opened 2021-06-22
exposing configurable worker timeouts

brettsam:
Part 1 of fix for https://github.com/Azure/azure-functions-dotnet-worker/issues/434

When debugging, dotnet-isolated worker pauses while waiting for the debugger to attach. This happens before we've sent the StartStream message from the worker to the host. This means that you can debug startup, etc, before the grpc communication even begins.

This change allows the worker process and initialization timeouts to be configured so that we can wait much longer in these scenarios while debugging. 

There will be a follow-up in the CLI to extend these timeouts if `--dotet-isolated-debug` is passed.

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
@@ -132,7 +132,6 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
 
         public async Task StartWorkerProcessAsync()
         {
-            _workerConfig.ProcessStartupTimeout = TimeSpan.FromDays(1);
             _startSubscription = _inboundWorkerEvents.Where(msg => msg.MessageType == MsgType.StartStream)
                 .Timeout(_workerConfig.ProcessStartupTimeout)
                 .Take(1)
--- a/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
@@ -22,7 +22,6 @@ using Microsoft.Azure.WebJobs.Script.ManagedDependencies;
 using Microsoft.Azure.WebJobs.Script.Workers;
 using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
 using Microsoft.Azure.WebJobs.Script.Workers.SharedMemoryDataTransfer;
-using Microsoft.CodeAnalysis.VisualBasic.Syntax;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using static Microsoft.Azure.WebJobs.Script.Grpc.Messages.RpcLog.Types;
@@ -34,7 +33,6 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
 {
     internal class GrpcWorkerChannel : IRpcWorkerChannel, IDisposable
     {
-        private readonly TimeSpan workerInitTimeout = TimeSpan.FromSeconds(30);
         private readonly IScriptEventManager _eventManager;
         private readonly RpcWorkerConfig _workerConfig;
         private readonly string _runtime;
@@ -48,7 +46,6 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
         private WorkerInitResponse _initMessage;
         private string _workerId;
         private RpcWorkerChannelState _state;
-        private Queue<string> _processStdErrDataQueue = new Queue<string>(3);
         private IDictionary<string, Exception> _functionLoadErrors = new Dictionary<string, Exception>();
         private ConcurrentDictionary<string, ScriptInvocationContext> _executingInvocations = new ConcurrentDictionary<string, ScriptInvocationContext>();
         private IDictionary<string, BufferBlock<ScriptInvocationContext>> _functionInputBuffers = new ConcurrentDictionary<string, BufferBlock<ScriptInvocationContext>>();
@@ -135,8 +132,9 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
 
         public async Task StartWorkerProcessAsync()
         {
+            _workerConfig.ProcessStartupTimeout = TimeSpan.FromDays(1);
             _startSubscription = _inboundWorkerEvents.Where(msg => msg.MessageType == MsgType.StartStream)
-                .Timeout(TimeSpan.FromSeconds(WorkerConstants.ProcessStartTimeoutSeconds))
+                .Timeout(_workerConfig.ProcessStartupTimeout)
                 .Take(1)
                 .Subscribe(SendWorkerInitRequest, HandleWorkerStartStreamError);
 
@@ -192,7 +190,7 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
         {
             _workerChannelLogger.LogDebug("Worker Process started. Received StartStream message");
             _inboundWorkerEvents.Where(msg => msg.MessageType == MsgType.WorkerInitResponse)
-                .Timeout(workerInitTimeout)
+                .Timeout(_workerConfig.InitializationTimeout)
                 .Take(1)
                 .Subscribe(WorkerInitResponse, HandleWorkerInitError);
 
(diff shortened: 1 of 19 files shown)


brettsam:
Side benefit -- the unit test runtime went from 21 mins -> 9 mins!

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -174,6 +174,10 @@ public static IHostBuilder AddScriptHostCore(this IHostBuilder builder, ScriptAp
 
                     return NullHostedService.Instance;
                 });
+
+                // Wire this up early so that any early worker logs are guaranteed to be flushed if any other
+                // IHostedService has a slow startup.
+                services.AddSingleton<IHostedService, WorkerConsoleLogService>();
```
brettsam, line 180: This is a second issue with debugging. The log coming from the worker identifying the target process id was being written, but never flushed, if we were waiting for a debugger to attach *and* the host wasn't able to start up fully because of that. 

This happens, for example, when RunOnStartup is enabled for timer triggers.

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptions.cs
```diff
@@ -14,5 +16,7 @@ public class HttpWorkerOptions
         public int Port { get; set; }
 
         public bool EnableForwardingHttpRequest { get; set; }
+
+        public TimeSpan InitializationTimeout { get; set; } = TimeSpan.FromSeconds(30);
```
pragnagopa, line 20: should this be 60 secs?

src/WebJobs.Script/Host/ScriptHost.cs
```diff
@@ -509,7 +513,14 @@ private void AddFunctionDescriptors(IEnumerable<FunctionMetadata> functionMetada
             else
             {
                 _logger.AddingDescriptorProviderForLanguage(_workerRuntime);
-                _descriptorProviders.Add(new RpcFunctionDescriptorProvider(this, _workerRuntime, ScriptOptions, _bindingProviders, _functionDispatcher, _loggerFactory, _applicationLifetime));
+
+                var workerConfig = _languageWorkerOptions.Value.WorkerConfigs?.FirstOrDefault(c => c.Description.Language.Equals(_workerRuntime, StringComparison.OrdinalIgnoreCase));
```
fabiocav, line 517: nit: I think assigning an instance of `RpcWorkerConfig` if the result here is `null` would simplify the code.

fabiocav (commented):
Are you expecting to have a Core Tools update to pass more appropriate defaults for local dev?

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfig.cs
```diff
@@ -1,14 +1,46 @@
...
 
         public WorkerProcessArguments Arguments { get; set; }
 
         public WorkerProcessCountOptions CountOptions { get; set; }
+
+        /// <summary>
+        /// Gets or sets the process startup timeout. This is the time from when the process is
+        /// launched to when the StartStream message is received.
+        /// </summary>
+        public TimeSpan ProcessStartupTimeout { get; set; } = TimeSpan.FromSeconds(60);
```
pragnagopa: new properties should be added to `WorkerProcessCountOptions` --> Please see PR: https://github.com/Azure/azure-functions-host/pull/6806 for more details on implementation.

This needs to be renamed to `WorkerProcessOptions` --> this was specifically added to be able to configure various values related to process start up

pragnagopa (commented):
Please move new properties into WorkerProcessCountOptions

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -136,7 +133,7 @@ public bool IsChannelReadyForInvocations()
         public async Task StartWorkerProcessAsync()
         {
             _startSubscription = _inboundWorkerEvents.Where(msg => msg.MessageType == MsgType.StartStream)
-                .Timeout(TimeSpan.FromSeconds(WorkerConstants.ProcessStartTimeoutSeconds))
+                .Timeout(_workerConfig.CountOptions.ProcessStartupTimeout)
```
pragnagopa, line 136: rename `CountOptions` to `processOptions` or `startUpOptions` ?

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptions.cs
fabiocav: Prior to the changes, this was 30 seconds, correct? Is there a reason to increase?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
fabiocav: This is an existing property. If we want to refactor/rename, that should probably be done in a separate PR.

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptions.cs
pragnagopa: Also, let's make `InitializationTimeout` internal for now. Eventually, HttpWorkerOptions can rely on `processOptions` as well.

src/WebJobs.Script/Workers/ProcessManagement/WorkerProcessCountOptions.cs
```diff
@@ -7,6 +7,8 @@ namespace Microsoft.Azure.WebJobs.Script.Workers
 {
     public class WorkerProcessCountOptions
     {
```
pragnagopa, line 9: rename :)

pragnagopa (commented):
Added minor comments. Thanks!

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptions.cs
pragnagopa: > Prior to the changes, this was 30 seconds, correct? Is there a reason to increase?

Never mind. I misinterpreted this as process startup timeout. Initialization should be 30secs.  Value looks ok

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
fabiocav: @pragnagopa wouldn't that also be a potential breaking change? If this options type is used from configuration and binding, different sources might be relying on this property name, which would break if it changes.

src/WebJobs.Script/Host/ScriptHost.cs
brettsam: I changed it slightly below so we don't need to create a worker config just to drop it. I use an internal static field directly now.

src/WebJobs.Script/Workers/ProcessManagement/WorkerProcessCountOptions.cs
brettsam: Punting for a future PR -- also this may be a breaking change depending on how we parse the worker options json file.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
brettsam: Punting for a future discusssion.

fabiocav (approved):

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
pragnagopa: @fabiocav  / @brettsam  -  In json it is named correctly - https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Workers/WorkerConstants.cs#L30 -- so not a breaking change

pragnagopa:
@brettsam - please open an issue to rename to avoid confusion in future.

pragnagopa (approved):

PR #7500, opened 2021-06-24
Skip shared memory initialization for core tools

### LINKED ISSUE #2600: MacOsx upgrade from 3.0.3442 to 3.0.3477 introduces /dev/shm/AzureFunctions warning message
Version 3.0.3442 starts without any warning (yellow) messages
```
Azure Functions Core Tools
Core Tools Version:       3.0.3442 Commit hash: 6bfab24b2743f8421475d996402c398d2fe4a9e0  (64-bit)
```

Version 3.0.3477 now starts with Operation not permitted message
```
Azure Functions Core Tools
Core Tools Version:       3.0.3477 Commit hash: 5fbb9a76fc00e4168f2cc90d6ff0afe5373afc6d  (64-bit)
Function Runtime Version: 3.0.15584.0

[2021-05-21T21:42:28.276Z] Cannot create directory for shared memory usage: /dev/shm/AzureFunctions
[2021-05-21T21:42:28.276Z] System.IO.FileSystem: Access to the path '/dev/shm/AzureFunctions' is denied. Operation not permitted.
```

ankitkumarr:

resolves https://github.com/Azure/azure-functions-core-tools/issues/2600

### Pull request checklist


gohar94:
If I understand correctly, you're completely disabling shared memory when running with core tools. Wouldn't people want to test this feature for their apps locally in some cases? 
I was thinking we should disable this if it's running on Mac only and change the log level of the line where it emits a warning that /dev/shm could not be created to "debug" so that if it complains on Linux (maybe due to permissions etc.) even that doesn't generate noise unless the user is running in verbose mode.

fabiocav:
I'm in favor of keeping it enabled and warning, but we should keep the experience consistent for all OSes, including on Mac. Thoughts?

gohar94:
> I'm in favor of keeping it enabled and warning, but we should keep the experience consistent for all OSes, including on Mac. Thoughts?

Agreed. 
My vote would be to skip the initialization if the feature is not enabled, instead of `if (environment.IsCoreTools())` from @ankitkumarr's latest change. 
Something that checks for the feature like being done [here](https://github.com/Azure/azure-functions-host/blob/c51e7bf4f6820308ba2012f9c417f68a354ecfe1/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs#L720) - just make sure that either you re-use this or make a separate helper (maybe a static method in `SharedMemoryManager` class and use that in both places?)

ankitkumarr:
Thanks @gohar94, I addressed your feedback, and am using the environment extensions to check for is shared memory is enabled on the environment.

src/WebJobs.Script/Workers/SharedMemoryDataTransfer/NullSharedMemoryManager.cs
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+        public Task<SharedMemoryMetadata> PutObjectAsync(object input)
+        {
+            return null;
+        }
+
+        public bool TryFreeSharedMemoryMap(string mapName)
+        {
+            return true;
```
gohar94: Let's make this `false` just to be safe.

src/WebJobs.Script/Workers/SharedMemoryDataTransfer/NullSharedMemoryManager.cs
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+        public bool TryFreeSharedMemoryMapsForInvocation(string invocationId)
+        {
+            return true;
```
gohar94: Same as above - `true` -> `false`.

gohar94 (dismissed):
Looks good to me - just added a couple of minor comments.

src/WebJobs.Script/Workers/SharedMemoryDataTransfer/NullSharedMemoryManager.cs
```diff
@@ -0,0 +1,44 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.Script.Workers.SharedMemoryDataTransfer
+{
+    public class NullSharedMemoryManager : ISharedMemoryManager
```
gohar94, line 9: CI build is failing.
` 'NullSharedMemoryManager' does not implement interface member 'ISharedMemoryManager.TryTrackSharedMemoryMap(string)'`

Please implement that and try a local build to make sure.

gohar94 (changes requested):

src/WebJobs.Script/Workers/SharedMemoryDataTransfer/NullSharedMemoryManager.cs
gohar94: We should also check if things work fine in cases where we don't create a `FunctionDataCache` service - running the end to end tests might show that. 
Otherwise, we could implement a `NullFunctionDataCache` and make sure to return `IsEnabled` as `False`.

src/WebJobs.Script/Workers/FunctionDataCache/NullFunctionDataCache.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+        }
+
+        public bool TryGet(FunctionDataCacheKey cacheKey, bool isIncrementActiveReference, out SharedMemoryMetadata sharedMemoryMeta)
+        {
+            try
+            {
+                lock (_lock)
+                {
+                    _localCache.TryGetValue(cacheKey, out sharedMemoryMeta);
```
gohar94: Do we need to have a lock and even do this? Can't we just return `false` right away?

src/WebJobs.Script/Workers/SharedMemoryDataTransfer/MemoryMappedFileAccessorUnix.cs
```diff
@@ -17,7 +17,7 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.SharedMemoryDataTransfer
     /// </summary>
     public class MemoryMappedFileAccessorUnix : MemoryMappedFileAccessor
     {
-        private IEnvironment _environment;
+        private readonly IEnvironment _environment;
```
gohar94, line 20: Any reason for making this `readonly`? If so, can you also check if we need to do the same in `MemoryMappedFileAccessorWindows.cs`?

src/WebJobs.Script/Workers/FunctionDataCache/NullFunctionDataCache.cs
```diff
@@ -0,0 +1,57 @@
...
+using System.Collections.Generic;
+
+namespace Microsoft.Azure.WebJobs.Script.Workers.FunctionDataCache
+{
+    public class NullFunctionDataCache : IFunctionDataCache
+    {
+        /// <summary>
+        /// Lock to be used with any operations that access state of this cache.
+        /// </summary>
+        private readonly object _lock = new object();
```
gohar94: Why do we need this in this Null implementation?

gohar94 (changes requested):

gohar94:
Let's do a before/after test on a Mac to be sure this fixes it before checking this in. 
Thanks.

src/WebJobs.Script/Workers/SharedMemoryDataTransfer/MemoryMappedFileAccessorUnix.cs
michaelpeng36: I believe this is to prevent unintentionally setting any environmental variables, as the class should be using the `_environment` variable only to get a list of directories where it can create a memory-mapped file. Currently, no methods set environmental variables, but the `readonly` modifier may nonetheless be a useful safeguard for future modifications.

michaelpeng36: As for the `MemoryMappedFileAccessorWindows.cs`, we have no corresponding `_environment` variable because the `MemoryMappedFile.CreateNew` method by default creates a memory-mapped file that is not mapped to an existing file on the disk (see [here](https://docs.microsoft.com/en-us/dotnet/standard/io/memory-mapped-files)).

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -601,18 +601,21 @@ internal async Task InvokeResponse(InvocationResponse invokeResponse)
                 }
                 finally
                 {
-                    // Free memory allocated by the host (for input bindings) which was needed only for the duration of this invocation
-                    if (!_sharedMemoryManager.TryFreeSharedMemoryMapsForInvocation(invokeResponse.InvocationId))
+                    if (_environment.SupportsSharedMemoryTransfer())
```
gohar94: Checking from the `_environment` can be expensive if you are doing it for every single invocation. It checks env vars etc.
Use `_isSharedMemoryDataTransferEnabled` instead which is already properly set in this class.

gohar94 (changes requested):

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
```diff
@@ -134,17 +134,46 @@ public static void AddWebJobsScriptHost(this IServiceCollection services, IConfi
...
+                    if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                    {
+                        services.AddSingleton<IMemoryMappedFileAccessor, MemoryMappedFileAccessorWindows>();
+                    }
+                    else
+                    {
+                        services.AddSingleton<IMemoryMappedFileAccessor, MemoryMappedFileAccessorUnix>();
+                    }
+
+                    var fileaccessor = p.GetService<IMemoryMappedFileAccessor>();
```
ejizba, line 156: @michaelpeng36 I think I finally figured out how to debug the host on my mac and `fileaccessor` is being returned as null here. Admittedly I don't know how `AddSingleton` works, but I wonder if there's a timing/ordering issue? It seems to work if I move the `AddSingleton<IMemoryMappedFileAccessor` stuff ahead of `services.AddSingleton<ISharedMemoryManager`

gohar94: That's because IMemoryMappedFileAccessor is required when constructing ISharedMemoryManager. Order of these initializations matters.

fabiocav:
Closing stale PR

PR #7510, opened 2021-06-29
disable proxies in v4

brettsam:


src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -249,7 +248,6 @@ public static IHostBuilder AddScriptHostCore(this IHostBuilder builder, ScriptAp
                 services.AddSingleton<IFunctionInvocationDispatcherFactory, FunctionInvocationDispatcherFactory>();
                 services.AddSingleton<IScriptJobHost>(p => p.GetRequiredService<ScriptHost>());
                 services.AddSingleton<IJobHost>(p => p.GetRequiredService<ScriptHost>());
-                services.AddSingleton<IFunctionProvider, ProxyFunctionProvider>();
```
yojagad, line 252: This caught my eye. I'm not aware of the context/decision, but loader.io relies on proxies to validate the test apps before running the load tests. Something to keep in mind.

brettsam: Yes, I'll need to fix those apps; good point.

ankitkumarr (approved):

PR #7538, opened 2021-07-13
[pack] Fix BlobStorageSecretRepoError

### LINKED ISSUE #7537: Functions not working due to BlobStorageSecretRepoError

Starting from 7/6 with Functions runtime release 3.1.0, we introduced a change in the order of precedence for applying application configuration related to secrets, specifically AzureWebJobsStorage.
With 3.1.0, settings from custom sources take precedence over app settings, which didn’t use to be the case. Given this breaking change in behaviour, customers apps might stop working when an invalid AzureWebJobsStorage URI value is given from custom sources.

#### Repro steps

Deploy a function app in one of the following regions: Canary, West Central US, North Central US, or West US to get version 3.1.0. With app creation, AzureWebJobsStorage app setting will be set to a valid blob. Deploy a HTTP trigger function and modify the code to inject a custom value for AzureWebJobsStorage and set it to an invalid value. Function app and function invocations will stop working.

#### Expected behavior

The expected behavior would be to not take the custom source as the ground truth for app settings, so the app would continue working after the faulty custom source was introduced.

#### Actual behavior

The function app stops working because the AzureWebJobsStorage value is being read from custom source and is invalid.

#### Known workarounds

1)	Validate and fix the value for AzureWebJobsStorage (in ALL configuration sources). Requires app restart.
2)	Add the following feature flag as an app setting, which prevents the new logic to be used. Requires app restart:
a.	App setting name: AzureWebJobsFeatureFlags
b.	App setting value: DisableMergedConfiguration

yojagad:

resolves #7537

Tests before and after the fix

Before the fix

https://wawswus.kusto.windows.net/wawsprod
FunctionsLogs
| where PreciseTimeStamp between (datetime(2021-07-13 04:00:00.1869752)..datetime(2021-07-13 06:00:00.1869752))
| where AppName has "functiondevyojagad"
| where HostVersion == "3.1.1.0"
| where Summary == "There was an error performing a read operation on the Blob Storage Secret Repository. Please ensure the \'AzureWebJobsStorage\' connection string is valid."

After the fix

https://wawswus.kusto.windows.net/wawsprod
FunctionsLogs
| where PreciseTimeStamp between (datetime(2021-07-13 04:00:00.1869752)..datetime(2021-07-13 06:00:00.1869752))
| where AppName has "functiondevyojagad"
| where HostVersion == "3.2.0.0"
| where Summary == "There was an error performing a read operation on the Blob Storage Secret Repository. Please ensure the \'AzureWebJobsStorage\' connection string is valid."

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script/StorageProvider/AzureStorageProvider.cs
+++ b/src/WebJobs.Script/StorageProvider/AzureStorageProvider.cs
@@ -39,8 +39,8 @@ namespace Microsoft.Azure.WebJobs.Script
             {
                 _ = scriptHostManager ?? throw new ArgumentNullException(nameof(scriptHostManager));
                 _configuration = new ConfigurationBuilder()
-                    .AddConfiguration(configuration)
                     .Add(new ActiveHostConfigurationSource(scriptHostManager))
+                    .AddConfiguration(configuration)
                     .Build();
             }
         }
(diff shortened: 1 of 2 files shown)


pgombar (approved):
lgtm

fabiocav (approved):
The change looks good, but @brettsam can you look at the test and perhaps run through the other scenarios with @yojagad? In 4.0, this logic needs to remain with the order of precedence we had prior to the change.

PR #7557, opened 2021-07-20
Addprofiler

jingwenw15:
### added exploratory profiling endpoint to hostcontroller.cs using diagnostics.netcore.client package 
modified httptrigger function to be anonymous 

(no tests added)


jingwenw15:
So so sorry, please ignore - this was an accident!

PR #7576, opened 2021-08-03
[pack] Modify dispatcher being set to initialized

### LINKED ISSUE #7577: Fix FunctionDispatcherState being set to initialized inside StartWorkerProcesses
Following the fix here https://github.com/Azure/azure-functions-host/issues/7565 , initialization of WebhostLanguageWorkerChannel can happen outside of `StartWorkerProcesses` as well. FunctionDispatcherState needs to be set to initialized in that case as well.

yojagad:

resolves #7577

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
+++ b/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
@@ -139,6 +139,16 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
             await _webHostLanguageWorkerChannelManager?.ShutdownChannelsAsync();
         }
 
+        private void SetDispatcherStateToInitialized(Dictionary<string, TaskCompletionSource<IRpcWorkerChannel>> webhostLanguageWorkerChannel = null)
+        {
+            if (State != FunctionInvocationDispatcherState.Initialized
+                && webhostLanguageWorkerChannel != null
+                && webhostLanguageWorkerChannel.Any())
+            {
+                SetFunctionDispatcherStateToInitializedAndLog();
+            }
+        }
+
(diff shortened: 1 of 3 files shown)


src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -165,6 +173,13 @@ private void StartWorkerProcesses(int startIndex, Func<Task> startAction, bool i
                         _logger.LogError(ex, $"Failed to start a new language worker for runtime: {_workerRuntime}.");
                     }
                 }
+
+                // It is necessary that webhostLanguageWorkerChannel.Any() happens in this thread since 'startAction()' above can modify this collection.
+                // WebhostLanguageWorkerChannel can be initialized and process started up outside of StartWorkerProcesses as well, hence the statement here.
+                if (initializeDispatcher)
+                {
+                    SetDispatcherStateToInitialized(webhostLanguageWorkerChannel);
```
pragnagopa, line 181: Can you please see how we could have caught this issue with unit tests or E2E tests we have in the host repo?

yojagad: I did take a look at it. I was initially under the impression that the test here https://github.com/Azure/azure-functions-host/blob/9bdb40b2d517f32c5052956ba8c9cf662ea36a9e/test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SamplesEndToEndTests_Java.cs#L37 would have caught this.

Apparently that test seems to have always returned 500 status code but since we don't test for the status code, we never caught this (java process never starts up but that could be because there's no JAVA_HOME set)... I'll see if I can find a way to fix it.

yojagad: @pragnagopa Fixed the Java E2E test and added a couple more assertions. That test should catch this now.

pragnagopa (approved):

PR #7581, opened 2021-08-04
Minimum extension version enforcement

paulbatum:
First iteration of https://github.com/Azure/azure-functions-host/issues/7645 which implements proposal https://github.com/Azure/Azure-Functions/issues/1987 - Require a minimum version for supported extensions.

Here is an example error thrown when using the 1.x bundle:

`Microsoft.Azure.WebJobs.Script.HostInitializationException: 'Referenced bundle 'Microsoft.Azure.Functions.ExtensionBundle' of version '1.8.1' does not meet the required minimum version of '2.2.0'. Update your extension bundle reference in host.json to reference '2.2.0' or later. For more information see https://aka.ms/func-min-bundle-versions.'`

And an example when using storage extension 3.x:

`Microsoft.Azure.WebJobs.Script.HostInitializationException: 'ExtensionStartupType 'AzureStorageWebJobsStartup' from assembly 'Microsoft.Azure.WebJobs.Extensions.Storage, Version=3.0.10.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35' does not meet the required minimum version of '4.0.4.0'. Update your NuGet package reference for 'Microsoft.Azure.WebJobs.Extensions.Storage' to '4.0.4' or later. For more information see https://aka.ms/func-min-extension-versions.'`

Obviously I need to write some tests, but I am keen to get feedback on this overall approach before I start doing that.

### Pull request checklist

* [x] Fill out extensionrequirements.json to cover all the extensions we want to enforce a minimum version for
* [x] Verify that 2.x bundle is up to date with these requirements (the upcoming 2.7.0 bundle is)
* [x] Emit a log entry that is easier to check for in a detector than the currently thrown exception
* [x] Validate these errors appear in a readable manner in core tools (they appear OK but it can be improved https://github.com/Azure/azure-functions-core-tools/issues/2714)
    * [x] Otherwise: https://github.com/MicrosoftDocs/azure-docs/issues/81253


src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
```diff
@@ -182,6 +187,8 @@ public async Task<IEnumerable<Type>> GetExtensionsStartupTypesAsync()
                         continue;
                     }
 
+                    ValidateExtensionRequirements(extensionType);
```
brettsam: Rather than throwing the first time we find a problem, it may be nicer to collect all the issues and then throw a nice message later with all the details.

brettsam:
The approach looks good to me. Just one minor comment for now...

brettsam (commented):
This approach looks good to me!

paulbatum:
Current  behavior in core tools is not ideal - an ArgumentNullException is thrown for the `GetRequiredService<IScriptJobHost>` step:

image

paulbatum:
I need some input on how we want this to work exactly - is throwing an exception the right way to block startup? If so, it looks like there might be some code changes needed in core tools to handle it more gracefully. Alternatively, we could skip loading the extensions that fail to validate. This would let the host startup in a partially running state - the functions that depend on the skipped extensions would fail to load. This might be useful if we eventually want these errors to be retrievable over the host errors API that the portal will be calling in the future.

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
```diff
@@ -73,6 +75,9 @@ public async Task<IEnumerable<Type>> GetExtensionsStartupTypesAsync()
 
             if (bundleConfigured)
             {
+                var bundleDetails = await _extensionBundleManager.GetExtensionBundleDetails();
```
fabiocav: nit: `var` -> Type name

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
```diff
@@ -220,6 +227,89 @@ private ExtensionReference[] ParseExtensions(string metadataFilePath)
...
+                _logger.MinimumExtensionVersionNotSatisfied(extensionType.Name, extensionType.Assembly.FullName, minimumVersion, requirement.PackageName, requirement.MinimumPackageVersion);
+                string requirementNotMetError = $"ExtensionStartupType {extensionType.Name} from assembly '{extensionType.Assembly.FullName}' does not meet the required minimum version of {minimumVersion}. Update your NuGet package reference for {requirement.PackageName} to {requirement.MinimumPackageVersion} or later.";
+                errors.Add(requirementNotMetError);
+            }
+
+            foreach (var extensionType in startupTypes)
+            {
+                if (_extensionRequirements.ExtensionRequirementsByStartupType.ContainsKey(extensionType.Name))
+                {
+                    ExtensionStartupTypeRequirement requirement = _extensionRequirements.ExtensionRequirementsByStartupType[extensionType.Name];
```
fabiocav: Replace with:
``` csharp
if (_extensionRequirements.ExtensionRequirementsByStartupType.TryGetValue(extensionType.Name, out ExtensionStartupTypeRequirement requirement))
```

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
```diff
@@ -220,6 +227,89 @@ private ExtensionReference[] ParseExtensions(string metadataFilePath)
             }
         }
 
+        private void ValidateBundleRequirements(ExtensionBundleDetails bundleDetails)
+        {
+            if (_extensionRequirements.BundleRequirementsByBundleId.ContainsKey(bundleDetails.Id))
+            {
+                BundleRequirement requirement = _extensionRequirements.BundleRequirementsByBundleId[bundleDetails.Id];
```
fabiocav: Replace with the `TryGetValue` call (same as above)

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
```diff
@@ -220,6 +227,89 @@ private ExtensionReference[] ParseExtensions(string metadataFilePath)
...
+
+            foreach (var extensionType in startupTypes)
+            {
+                if (_extensionRequirements.ExtensionRequirementsByStartupType.ContainsKey(extensionType.Name))
+                {
+                    ExtensionStartupTypeRequirement requirement = _extensionRequirements.ExtensionRequirementsByStartupType[extensionType.Name];
+                    Version minimumAssemblyVersion = new Version(requirement.MinimumAssemblyVersion);
+
+                    Version extensionAssemblyVersion = extensionType.Assembly.GetName().Version;
+                    string extensionAssemblySimpleName = extensionType.Assembly.GetName().Name;
```
fabiocav: Call `GetName()` once and reuse the `AssemblyName` result to get `Version` and `Name`

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
```diff
@@ -220,6 +227,89 @@ private ExtensionReference[] ParseExtensions(string metadataFilePath)
...
+                        continue;
+                    }
+
+                    if (requirement.MinimumAssemblyFileVersion != null)
+                    {
+                        Version minimumAssemblyFileVersion = new Version(requirement.MinimumAssemblyFileVersion);
+
+                        try
+                        {
+                            Version extensionAssemblyFileVersion = new Version(System.Diagnostics.FileVersionInfo.GetVersionInfo(extensionType.Assembly.Location).FileVersion);
```
fabiocav, line 279: This might be an expensive call in the startup path. Do we want to support file versions? I'll look at the file, but I'm assuming this would be for a small number of scenarios, correct?

fabiocav: No concrete scenarios today, but in dynamic assembly scenarios (and some load scenarios like loading from bytes), `Location` will not be populated as expected, so it would be good to guard against that.

src/WebJobs.Script/extensionrequirements.json
```diff
@@ -0,0 +1,68 @@
...
+{
+  "bundles": [
+    {
+      "id": "Microsoft.Azure.Functions.ExtensionBundle",
+      "minimumVersion": "2.7.0"
+    }
+  ],
+  "types": [
+    {
+      "name": "AzureStorageWebJobsStartup",
```
fabiocav, line 11: Wondering if it would be a good idea to update all the extensions (create issues) with a comment on their startup type letting folks know they shouldn't be renamed  🤔

paulbatum: Since we are just protecting against older extensions, I am not too worried about this. Yes this implementation would get more complicated if the types were renamed and we were doing similar enforcement in another major version (say V5).. but if that does end up becoming an issue, there are other strategies we could use - we could ignore the startup type and just use its assembly, for example. I went with basing it on startup types because it does provide a nice workaround if we ever need to circumvent this behavior...

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
paulbatum: We need to support it to provide the right check for durable. Notice that I coded this path to only activate if a file version specifier is present - durable is the only one with that specifier, and so we will only run through this codepath if the durable extension is present.

paulbatum: I did add a guard for the location though

src/WebJobs.Script/extensionrequirements.json
fabiocav: Make sense.

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
fabiocav: I did notice the check for the file version and did see that Durable is the only one with that set, but I do worry about the impact as it will be present in a large number of apps.

fabiocav (approved):
For the skipped tests, should we have an issue to track?

paulbatum:
Durable webhook tests were skipped as part of v4 upgrade  https://github.com/Azure/azure-functions-host/issues/7679
Functions V4 - Implement better handling for out of date extensions https://github.com/Azure/azure-functions-core-tools/issues/2714

PR #7621, opened 2021-08-18
Set dispatcher state to initialized only if there is at least one proper channel

### LINKED ISSUE #7590: Erroneous dispatcher state when webhostLanguageWorkerProcess doesn't come up
Language worker processes can fail to come up during runtime startup.

An example of successful process startup is 
image

An example of unsuccessful process startup is 
image

In the second scenario mentioned, currently the dispatcher is **still** set to `initialized` thus, requests being routed to erroneous channels. 

The original design is for the Dispatcher to wait until a certain timeout period and then shutdown the runtime if it does not (https://github.com/Azure/azure-functions-host/blob/2520fdb09fbaf7518925f8aaea618dd9b95de7ba/src/WebJobs.Script/Description/Workers/WorkerFunctionInvoker.cs#L63) .

This workitem is to ensure that `dispatcher` state is not set to `initialized` in case of language worker process startup errors thus conforming to the original, intended design.

cc: @pragnagopa @AnatoliB @alrod @amamounelsayed @fabiocav

yojagad:

resolves #7590 

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
+++ b/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
@@ -141,9 +141,10 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
 
         private void SetDispatcherStateToInitialized(Dictionary<string, TaskCompletionSource<IRpcWorkerChannel>> webhostLanguageWorkerChannel = null)
         {
+            // RanToCompletion indicates successful process startup
             if (State != FunctionInvocationDispatcherState.Initialized
                 && webhostLanguageWorkerChannel != null
-                && webhostLanguageWorkerChannel.Any())
+                && webhostLanguageWorkerChannel.Where(a => a.Value.Task.Status == TaskStatus.RanToCompletion).Any())
             {
                 SetFunctionDispatcherStateToInitializedAndLog();
             }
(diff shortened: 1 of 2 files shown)


pragnagopa:
Thanks for the PR @yojagad ! Looking at the issue description

>  the Dispatcher can still be set to initialized state even if WebhostlanguageLanguageWorkerChannel does not exist thus leading to requests attempting to be routed to non-existing language workers.

The fix looks right to me. Can you explain the caveat with this fix regarding

> This means there will still be at least one request routed before shutting down the host.

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -141,9 +141,10 @@ internal async void ShutdownWebhostLanguageWorkerChannels()
 
         private void SetDispatcherStateToInitialized(Dictionary<string, TaskCompletionSource<IRpcWorkerChannel>> webhostLanguageWorkerChannel = null)
         {
+            // RanToCompletion indicates successful process startup
             if (State != FunctionInvocationDispatcherState.Initialized
                 && webhostLanguageWorkerChannel != null
-                && webhostLanguageWorkerChannel.Any())
+                && webhostLanguageWorkerChannel.Where(a => a.Value.Task.Status == TaskStatus.RanToCompletion).Any())
```
pragnagopa, line 147: Please add unit test in `RpcFunctionInvocationDispatcherTests.cs`

yojagad:
`The fix looks right to me. Can you explain the caveat with this fix regarding`

The line here explains the caveat https://github.com/Azure/azure-functions-host/blob/2520fdb09fbaf7518925f8aaea618dd9b95de7ba/src/WebJobs.Script/Description/Workers/WorkerFunctionInvoker.cs#L63

Before this fix, the shutdown was never initiated because dispatcher is set to `Initialized` , so `InvokeCore` would just keep routing requests. With this fix, it does shutdown after waiting.

I'm just explicitly mentioning that it still takes at least one request before shutdown is triggered ((which seems to be the original design anyway from looking at `WorkerFunctionInvoker`)

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
yojagad: Added test for both success and failure scenarios (the state should remain as `initializing`)

yojagad:
@pragnagopa Updated description in the original issue. Let me know if I missed anything else, please.

pragnagopa (approved):

PR #7653, opened 2021-09-07
Ensure CheckFileSystem runs on specialization (#7648)

### LINKED ISSUE #7648: FileSystem error check needs to be performed during specialization restart as well
The file system check we perform [here](https://github.com/Azure/azure-functions-host/blob/ed814c569434d65abcc22a169ac2d95f1ef79c32/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs#L150) is only done when the WebJobsScriptHostService hosted service initially starts. That means when placeholders are used, the check is only performed for the placeholder, and when the app specializes it's not performed again. That's incorrect - when specializing, the app will be receiving it's app content and it's then that the check should be performed.

At a minimum, the file system check should also be performed [here](https://github.com/Azure/azure-functions-host/blob/ed814c569434d65abcc22a169ac2d95f1ef79c32/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs#L426) on restart. Or perhaps the check should be moved to the common UnsynchronizedStartHostAsync method so it's called in all cases at a single point.

To test the fix, create an app in placeholder mode wit the "FAILED TO INITIALIZE RUN FROM PACKAGE.txt" in place and trigger specialization, and verify that the app shuts down.

mathewc:
Resolves https://github.com/Azure/azure-functions-host/issues/7648

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
+++ b/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
@@ -147,7 +147,6 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
 
         public async Task StartAsync(CancellationToken cancellationToken)
         {
-            CheckFileSystem();
             if (ShutdownRequested)
             {
                 return;
@@ -176,9 +175,10 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
 
         private void CheckFileSystem()
         {
-            // Shutdown if RunFromZipFailed
             if (_environment.ZipDeploymentAppSettingsExist())
             {
+                // Check for marker file indicating a zip package failure, and if found stop the application.
+                // We never want to run with an incorrect file system.
                 string path = Path.Combine(_applicationHostOptions.CurrentValue.ScriptPath, ScriptConstants.RunFromPackageFailedFileName);
                 if (File.Exists(path))
                 {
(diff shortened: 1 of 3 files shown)


src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -222,6 +222,12 @@ private void CheckFileSystem()
         /// </summary>
         private async Task UnsynchronizedStartHostAsync(ScriptHostStartupOperation activeOperation, int attemptCount = 0, JobHostStartupMode startupMode = JobHostStartupMode.Normal)
         {
+            CheckFileSystem();
```
mathewc, line 225: Chose to centralize this, rather than add a CheckFileSystem call to RestartHostAsync which also would have worked.

test/WebJobs.Script.Tests.Shared/TestHelpers.cs
```diff
@@ -42,6 +43,22 @@ public static string FunctionsTestDirectory
...
+                throw new ArgumentNullException("waitHandle");
+            }
+
+            var tcs = new TaskCompletionSource<bool>();
+            var rwh = ThreadPool.RegisterWaitForSingleObject(waitHandle,
+                delegate { tcs.TrySetResult(true); }, null, -1, true);
+            var t = tcs.Task;
+            t.ContinueWith((antecedent) => rwh.Unregister(null));
+
+            return t;
```
fabiocav, line 58: Did you mean to return the task returned by the `ContinueWith` call above? Any other continuations should come after that one.

mathewc: I'm using code from official documentation here: https://docs.microsoft.com/en-us/dotnet/standard/asynchronous-programming-patterns/interop-with-other-asynchronous-patterns-and-types?redirectedfrom=MSDN#from-wait-handles-to-tap. Agree with you that that code should return the continuation. Making that change.

brettsam (approved):

PR #7663, opened 2021-09-11
Update key generation to provide fixed signatures and encoded checksums (for improved security tooling).

### LINKED ISSUE #7662: Consider updating key generation to increase identifiability (and glean security value from improved tooling)
Inadvertent plaintext exposure of secrets is a significant industry problem. Many modern security models are therefore updating their key generation logic to encode fixed signatures and checksums in secret data. This approach allows various scanners (such as Microsoft Defender and [GitHub Token Scanning](https://github.blog/2018-10-17-behind-the-scenes-of-github-token-scanning/), to efficiently detect/block exposures.

Without improvement of this kind, tools must either generate large #'s of false positives (which prevent meaningful engineer review and security response) or favor approaches that generate false negatives (which leaves projects exposed to security risk).

Encode a fixed signature (expressed in the base64-encoded secret) in Azure function secrets. Encode a checksum in the final four bytes of the decoded data. Provide both mechanisms without compromising required security (bits of entropy) in the generated secret. 

GitHub's post on its [new authentication format](https://github.blog/2021-04-05-behind-githubs-new-authentication-token-formats/) provides a good description of approach. Note that Azure DevOps, NuGet, and NPM also have implemented similar technique to provide additional security value.

Some providers have exclusively utilized a fixed signature (often a prefix) in the encoded value. This approach is subject to collisions, however, if there is a single (unrelated) data generator that duplicates the signature.

michaelcfanning:
Resolves #7662

### Pull request checklist


liliankasem (commented):
Thanks for contributing to the repo! This looks pretty good to me, only notes I have are around using magic numbers in the SecretManager.cs - I think assigning a lot of those numbers as readonly constants will improve readability and understanding around the code base 

e.g. `int padding = (int)((keyLengthInBytes - 7) * 8) % 6;` <-- what do the numbers 7 8 and 6 represent etc.

Hesitant to approve this without further review from senior members of the team as I might be missing some important context here. Tagging @fabiocav, @brettsam & @mathewc to give this a look over when they have a minute.

michaelcfanning:
Thanks for the review! It's been a while since I tabled this for the holidays. :) I will update the PR and look at your note. Portions of this contribution have now shipped as a NuGet package, [Microsoft.Security.Utilities](https://www.nuget.org/packages/Microsoft.Security.Utilities), so I intend to switch to using that (reducing the size of this change). One final question: does your resource provider have a feature flag/conditional enablement mechanism I should provide in order to help with staged deployment? Thanks for everything!

---
In reply to: [813828758](https://github.com/Azure/azure-functions-host/pull/7663#issuecomment-813828758) [](http://example.com/codeflow?ancestors=813828758,813828758)

Issue #7539, opened 2021-07-13
Microsoft.Azure.WebJobs.Script.WebHost.DependencyInjection.JobHostServiceProvider.GetRequiredService returns null;


qsdfplkj:
Using seviceProvider.GetRequiredService that supports ISupportRequiredService that is Microsoft.Azure.WebJobs.Script.WebHost.DependencyInjection.JobHostServiceProvide returns a null on a interface that is not registered in the serviceProvider.

i'm using 3.1.16 of the microsoft.extensions.dependecyinjection package.

v-bbalaiagar:
Hi @qsdfplkj, Thank you for your feedback! We will check for the possibilities internally and update you with the findings.

v-bbalaiagar:
Hi @qsdfplkj, We would like to have more insights on the scenario to understand it better. Could you please provide us the invocation id of the error, time stamp and region to get the logs and check the cause. Kindly provide the repro steps for us to clone the issue and get the error.

qsdfplkj:
Just create a function app and use the following

    public class INotImplemented
    {
    }

    public class Function1
    {
        public Function1(IServiceProvider serviceProvider)
        {
            // This should throw an error but it doesn't it just returns null;
            INotImplemented required = serviceProvider.GetRequiredService<INotImplemented>();
        }

        [FunctionName("Function1")]
        public void Run([TimerTrigger("*/1 * * * * *")] TimerInfo myTimer, ILogger log)
        {
            log.LogInformation($"C# Timer trigger function executed at: {DateTime.Now}");

        }
    }

v-bbalaiagar:
Hi @qsdfplkj, Thank you for the feedback, we will reproduce this issue and update as appropriate.

v-bbalaiagar:
Hi @qsdfplkj, Apologies for the delayed response, I have tried to reproduce the mentioned scenario. Below are the repro steps:

1.Created function project with Timer trigger function.

![image](https://user-images.githubusercontent.com/74654421/128323051-4579a70e-71e5-409b-baee-06cf5cb335e4.png)

When I debug the app, The function get executed successfully as below and it doesn't returns null.

![image](https://user-images.githubusercontent.com/74654421/128324513-1e0f8e76-41e7-44e5-b4b5-c508f1967c06.png)

Let us know if I am missing a step here.

qsdfplkj:
[...]

PR #7664, opened 2021-09-12
Add a test to expose issue #7539

qsdfplkj:
GetRequiredService_OnJobHost_ShouldThrowAnInvalidOperationExceptionWhenAServiceCannotBeResolved

Adding a test to expose this issue https://github.com/Azure/azure-functions-host/issues/7539

GetRequiredService is supposed to throw when there is an issue but instead returns null.

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/test/WebJobs.Script.Tests/DependencyInjection/JobHostServiceProviderTests.cs
+++ b/test/WebJobs.Script.Tests/DependencyInjection/JobHostServiceProviderTests.cs
@@ -88,6 +88,18 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.DependencyInjection
             // Disposing of the JobHost service provider should not dispose of root container services
             Assert.False(rootService.Disposed);
         }
+        
+        [Fact]
+        public void GetRequiredService_OnJobHost_ShouldThrowAnInvalidOperationExceptionWhenAServiceCannotBeResolved()
+        {
+            var services = new ServiceCollection();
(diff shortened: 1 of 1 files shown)


liliankasem:
@qsdfplkj Is this draft still being worked on? Can you add more details to the PR description about what this is for and why? trying to gauge if we should close this or if we're working to merge

qsdfplkj:
Yes well... I tried to show what the issue is.... 

This is a draft for a test (I don't know how to run them) please have a look at the attached issue where I did point out exactly what the issue is. I cannot reproduce this because I don't have access to the source of the JobHostServiceProvider. But the JobHostServiceProvider implements ISupportRequiredService and therefor should throw an exception if a service can't be resolved. It should not return null. **GetRequiredService_OnJobHost_ShouldThrowAnInvalidOperationExceptionWhenAServiceCannotBeResolved**

Please reed my comments on the related issue. It points out the exact line where a null is returned.
I've discussed this at https://github.com/dotnet/runtime/pull/58685 and they are right this JobHostServiceProvider  should throw an error according to the documentation on the ISupportRequiredService.

https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.isupportrequiredservice.getrequiredservice?view=dotnet-plat-ext-6.0#returns

How to get this testcase in? Is it correct? This test case should fail so someone can solve it.

liliankasem:
Issue will be addressed at a later date. Any future work around this issue should contain a unit test so we will not be checking in a unit test by itself.

PR #7682, opened 2021-09-20
HostID Collision detection

mathewc:
Adding logic to detect host ID collisions in shared storage accounts. In Functions v3 this will log a warning when detected, and in v4 it will be an error.

Breaking change proposal for v4: https://github.com/Azure/Azure-Functions/issues/2049

### DIFF
--- a/src/WebJobs.Script/Host/HostIdValidator.cs
+++ b/src/WebJobs.Script/Host/HostIdValidator.cs
@@ -1,0 +1,201 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using System.Threading.Tasks;
+using Azure;
+using Azure.Storage.Blobs;
+using Microsoft.Azure.WebJobs.Script.Properties;
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+using IApplicationLifetime = Microsoft.AspNetCore.Hosting.IApplicationLifetime;
+
+namespace Microsoft.Azure.WebJobs.Script
+{
+    /// <summary>
+    /// Used to perform Host ID validation checks, ensuring that when hosts are sharing
+    /// a storage account, their computed IDs don't collide.
+    /// </summary>
+    /// <remarks>
+    /// <see cref="ScriptHostIdProvider"/> computes a Host ID and truncates it if needed to
+    /// ensure it's under length limits. For two different Function Apps, this can result in
+    /// both apps resolving to the same Host ID. This can cause problems if those apps share
+    /// a storage account. This class helps detect/prevent such cases.
+    /// </remarks>
+    public class HostIdValidator
+    {
+        public const string BlobPathFormat = "ids/usage/{0}";
+        private const LogLevel DefaultLevel = LogLevel.Warning;
+
+        private readonly IEnvironment _environment;
+        private readonly IAzureStorageProvider _storageProvider;
+        private readonly IApplicationLifetime _applicationLifetime;
+        private readonly HostNameProvider _hostNameProvider;
+        private readonly ILogger _logger;
+
+        private readonly object _syncLock = new object();
+        private bool _validationScheduled;
+
+        public HostIdValidator(IEnvironment environment, IAzureStorageProvider storageProvider, IApplicationLifetime applicationLifetime,
+            HostNameProvider hostNameProvider, ILogger<HostIdValidator> logger)
+        {
+            _environment = environment;
+            _storageProvider = storageProvider;
+            _applicationLifetime = applicationLifetime;
+            _hostNameProvider = hostNameProvider;
+            _logger = logger;
+        }
+
+        internal bool ValidationScheduled => _validationScheduled;
+
+        public virtual void ScheduleValidation(string hostId)
+        {
+            lock (_syncLock)
+            {
+                if (!_validationScheduled)
+                {
+                    // Schedule the validation to run asynchronously after a delay. This delay ensures
+                    // we're not impacting coldstart.
+                    Utility.ExecuteAfterColdStartDelay(_environment, () => Task.Run(() => ValidateHostIdUsageAsync(hostId)));
+                    _validationScheduled = true;
+                }
+            }
+        }
+
+        internal async Task ValidateHostIdUsageAsync(string hostId)
+        {
+            try
+            {
+                if (!_storageProvider.ConnectionExists(ConnectionStringNames.Storage))
+                {
+                    return;
+                }
+
+                HostIdInfo hostIdInfo = await ReadHostIdInfoAsync(hostId);
+
+                if (hostIdInfo != null)
+                {
+                    // an existing record exists for this host ID
+                    CheckForCollision(hostId, hostIdInfo);
+                }
+                else
+                {
+                    // no existing record, so write one, claiming this host ID for this host name
+                    // in this storage account
+                    hostIdInfo = new HostIdInfo
+                    {
+                        Hostname = _hostNameProvider.Value
+                    };
+                    await WriteHostIdAsync(hostId, hostIdInfo);
+                }
+            }
+            catch (Exception ex)
+            {
+                // best effort - log error and continue
+                _logger.LogError(ex, "Error validating host ID usage.");
+            }
+        }
+
+        private void CheckForCollision(string hostId, HostIdInfo hostIdInfo)
+        {
+            // verify the host name is the same as our host name
+            if (!string.Equals(_hostNameProvider.Value, hostIdInfo.Hostname, StringComparison.OrdinalIgnoreCase))
+            {
+                HandleCollision(hostId);
+            }
+        }
+
+        private void HandleCollision(string hostId)
+        {
+            // see if the user has specified a level, otherwise default
+            string value = _environment.GetEnvironmentVariable(EnvironmentSettingNames.FunctionsHostIdCheckLevel);
+            if (!Enum.TryParse<LogLevel>(value, out LogLevel level))
+            {
+                level = DefaultLevel;
+            }
+
+            string message = string.Format(Resources.HostIdCollisionFormat, hostId);
+            if (level == LogLevel.Error)
+            {
+                _logger.LogError(message);
+                _applicationLifetime.StopApplication();
+            }
+            else
+            {
+                // we only allow Warning/Error levels to be specified, so anything other than
+                // Error is treated as warning
+                _logger.LogWarning(message);
+            }
+        }
+
+        internal async Task WriteHostIdAsync(string hostId, HostIdInfo hostIdInfo)
+        {
+            try
+            {
+                var containerClient = _storageProvider.GetBlobContainerClient();
+                string blobPath = string.Format(BlobPathFormat, hostId);
+                BlobClient blobClient = containerClient.GetBlobClient(blobPath);
+                BinaryData data = BinaryData.FromObjectAsJson(hostIdInfo);
+                await blobClient.UploadAsync(data);
+
+                _logger.LogDebug($"Host ID record written (ID:{hostId}, HostName:{hostIdInfo.Hostname})");
+            }
+            catch (RequestFailedException rfex) when (rfex.Status == 409)
+            {
+                // Another instance wrote the blob between the time when we initially
+                // checked and when we attempted to write. Read the blob and validate it.
+                hostIdInfo = await ReadHostIdInfoAsync(hostId);
+                if (hostIdInfo != null)
+                {
+                    CheckForCollision(hostId, hostIdInfo);
+                }
+            }
+            catch (Exception ex)
+            {
+                // best effort
+                _logger.LogError(ex, "Error writing host ID info");
+            }
+        }
+
+        internal async Task<HostIdInfo> ReadHostIdInfoAsync(string hostId)
+        {
+            HostIdInfo hostIdInfo = null;
+
+            try
+            {
+                // check storage to see if a record already exists for this host ID
+                var containerClient = _storageProvider.GetBlobContainerClient();
+                string blobPath = string.Format(BlobPathFormat, hostId);
+                BlobClient blobClient = containerClient.GetBlobClient(blobPath);
+                var downloadResponse = await blobClient.DownloadAsync();
+                string content;
+                using (StreamReader reader = new StreamReader(downloadResponse.Value.Content))
+                {
+                    content = reader.ReadToEnd();
+                }
+
+                if (!string.IsNullOrEmpty(content))
+                {
+                    hostIdInfo = JsonConvert.DeserializeObject<HostIdInfo>(content);
+                }
+            }
+            catch (RequestFailedException exception) when (exception.Status == 404)
+            {
+                // no record stored for this host ID
+            }
+            catch (Exception ex)
+            {
+                // best effort
+                _logger.LogError(ex, "Error reading host ID info");
+            }
+
+            return hostIdInfo;
+        }
+
+        internal class HostIdInfo
+        {
+            public string Hostname { get; set; }
+        }
+    }
+}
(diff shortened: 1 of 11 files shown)


src/WebJobs.Script/Host/HostIdValidator.cs
```diff
@@ -0,0 +1,200 @@
...
+    /// <remarks>
+    /// <see cref="ScriptHostIdProvider"/> computes a Host ID and truncates it if needed to
+    /// ensure it's under length limits. For two different Function Apps, this can result in
+    /// both apps resolving to the same Host ID. This can cause problems if those apps share
+    /// a storage account. This class helps detect/prevent such cases.
+    /// </remarks>
+    public class HostIdValidator
+    {
+        public const string BlobPathFormat = "ids/usage/{0}";
+        private const LogLevel DefaultLevel = LogLevel.Warning;
```
mathewc, line 29: When this is merged to the v4 branch, we simply have to update this to default Error level.

src/WebJobs.Script/Properties/Resources.resx
```diff
@@ -164,4 +164,7 @@
   IsStopwatchHighResolution: {3}
 }}</value>
   </data>
+  <data name="HostIdCollisionFormat" xml:space="preserve">
+    <value>A collision for Host ID '{0}' was detected in the configured storage account. For more information, see https://aka.ms/functions-hostid-collision.</value>
```
mathewc, line 168: This aka.ms link doesn't exist yet. We'll need to add it.

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -328,6 +328,8 @@ public static void AddCommonServices(IServiceCollection services)
             // to be careful with caching, etc. E.g. these services will get
             // initially created in placeholder mode and live on through the
             // specialized app.
+            services.AddSingleton<HostNameProvider>();
```
mathewc, line 331: My changes work when running the host normally, but some CI tests are failing for some reason seemingly due to the service changes made. Our tests have some complicated logic where they build the host in a custom way not using production builder codepaths. I'm debugging this.

src/WebJobs.Script/Host/HostIdValidator.cs
```diff
@@ -0,0 +1,198 @@
...
+            if (value != null && Enum.TryParse<LogLevel>(value, out LogLevel parsedLevel))
+            {
+                level = parsedLevel;
+            }
+
+            string message = string.Format(Resources.HostIdCollisionFormat, hostId);
+            if (level == LogLevel.Error)
+            {
+                _logger.LogError(message);
+                _applicationLifetime.StopApplication();
```
paulbatum, line 122: Just curious about the timing for this. Is it possible that some listeners would have started and some function executions are in progress by the time this StopApplication call is made? (I get that we don't want to block HTTP executions on this check to avoid cold start impact)

mathewc: Yes - to avoid coldstart impact we don't block startup on the check. So we're optimizing for apps with no issues - if an app does have a collision then yes, the hard failure will occur 10s after startup.

src/WebJobs.Script/Host/HostIdValidator.cs
```diff
@@ -0,0 +1,198 @@
...
+        public virtual void ScheduleValidation(string hostId, int validationDelaySeconds = 10)
+        {
+            lock (_syncLock)
+            {
+                if (!_validationScheduled)
+                {
+                    // Schedule the validation to run asynchronously after a delay. This delay ensures
+                    // we're not impacting coldstart, and also gives time for the primary host to be
+                    // identified.
+                    Utility.ExecuteAfterDelay(() => Task.Run(() => ValidateHostIdUsageAsync(hostId)), TimeSpan.FromSeconds(validationDelaySeconds));
```
fabiocav: Is there a reason for the `Task.Run` here?

src/WebJobs.Script/Host/HostIdValidator.cs
```diff
@@ -0,0 +1,198 @@
...
+        internal async Task ValidateHostIdUsageAsync(string hostId)
+        {
+            try
+            {
+                if (!_storageProvider.ConnectionExists(ConnectionStringNames.Storage))
+                {
+                    return;
+                }
+
+                var primaryHostStateProvider = _serviceProvider.GetScriptHostServiceOrNull<IPrimaryHostStateProvider>();
```
fabiocav: Should we move `IPrimaryHostStateProvider` to become an explicit dependency of this type? Getting this from the provider has some negative side-effects, including additional testing complexity.

src/WebJobs.Script/Host/HostIdValidator.cs
```diff
@@ -0,0 +1,198 @@
...
+                    return;
+                }
+
+                HostIdInfo hostIdInfo = await ReadHostIdInfoAsync(hostId);
+
+                if (hostIdInfo != null)
+                {
+                    // an existing record exists for this host ID
+                    // verify the host name is the same as our host name
+                    if (string.Compare(_hostNameProvider.Value, hostIdInfo.Hostname, StringComparison.OrdinalIgnoreCase) != 0)
```
fabiocav: nit: `string.Compare` -> `string.Equals`

src/WebJobs.Script/Host/HostIdValidator.cs
```diff
@@ -0,0 +1,198 @@
...
+
+        private void HandleCollision(string hostId)
+        {
+            // see if the user has specified a level, otherwise default
+            LogLevel level = DefaultLevel;
+            string value = _environment.GetEnvironmentVariable(EnvironmentSettingNames.FunctionsHostIdCheckLevel);
+            if (value != null && Enum.TryParse<LogLevel>(value, out LogLevel parsedLevel))
+            {
+                level = parsedLevel;
+            }
```
fabiocav:
```suggestion
            string value = _environment.GetEnvironmentVariable(EnvironmentSettingNames.FunctionsHostIdCheckLevel);
            if (!Enum.TryParse<LogLevel>(value, out LogLevel level))
            {
                level = DefaultLevel;
            }
```

src/WebJobs.Script/Host/HostIdValidator.cs
```diff
@@ -0,0 +1,198 @@
...
+
+        internal async Task WriteHostIdAsync(string hostId, HostIdInfo hostIdInfo)
+        {
+            try
+            {
+                var containerClient = _storageProvider.GetBlobContainerClient();
+                string blobPath = string.Format(BlobPathFormat, hostId);
+                BlobClient blobClient = containerClient.GetBlobClient(blobPath);
+                BinaryData data = BinaryData.FromObjectAsJson(hostIdInfo);
+                await blobClient.UploadAsync(data);
```
fabiocav, line 140: This overload will overwrite an existing blob if one exists, so that would cause a problem if another host had "claimed" the blob/storage. We should update this to use one of the overloads that throws if the blob exists (this would occur if there's a race condition and the blob was created by another app between the check and this write operation)

src/WebJobs.Script/Host/HostIdValidator.cs
```diff
@@ -0,0 +1,198 @@
...
+            }
+            catch (Exception ex)
```
fabiocav, line 154: After updating the overload to avoid overwriting the blob, we should handle the storage exception if the blob write fails and evaluate whether we need to shut this instance down (yes if the blob was written by another site, but we could repeat the check)

src/WebJobs.Script/Properties/Resources.resx
fabiocav: Do we have the docs page setup (or in progress)?

src/WebJobs.Script/Host/HostIdValidator.cs
```diff
@@ -0,0 +1,198 @@
...
+        private readonly IServiceProvider _serviceProvider;
+        private readonly IApplicationLifetime _applicationLifetime;
+        private readonly HostNameProvider _hostNameProvider;
+        private readonly ILogger _logger;
+
+        private readonly object _syncLock = new object();
+        private bool _validationScheduled;
+
+        public HostIdValidator(IEnvironment environment, IAzureStorageProvider storageProvider, IApplicationLifetime applicationLifetime,
+            HostNameProvider hostNameProvider, ILoggerFactory loggerFactory, IServiceProvider serviceProvider)
```
brettsam: Inject an `ILogger<HostIdValidator>` here instead of the factory (trying to get away from those custom categories we started with).

src/WebJobs.Script/Host/HostIdValidator.cs
```diff
@@ -0,0 +1,198 @@
...
+                // best effort
+                _logger.LogError(ex, "Error reading host ID info");
+            }
+
+            return hostIdInfo;
+        }
+
+        internal class HostIdInfo
+        {
+            public string Hostname { get; set; }
```
brettsam, line 198: Could this also include the untruncated host name so we can present a specific log to the customer about which app has claimed this account?

src/WebJobs.Script/Host/HostIdValidator.cs
```diff
@@ -0,0 +1,198 @@
...
+            else
+            {
+                // we only allow Warning/Error levels to be specified, so anything other than
+                // Error is treated as warning
+                _logger.LogWarning(message);
+            }
+        }
+
+        internal async Task WriteHostIdAsync(string hostId, HostIdInfo hostIdInfo)
+        {
```
brettsam, line 133: General question -- should we put a few more debug logs throughout here so we can track this should we need to? Examples -- when we write the blob, when we successfully validate? Mostly to ensure this is running as expected.

src/WebJobs.Script/Host/HostIdValidator.cs
mathewc: I need to kick off an asynchronous background task that isn't awaited. I could remove the Task.Run, and add inline #pragmas to ignore the unawaited task warnings, but this seems preferable.

src/WebJobs.Script/Host/HostIdValidator.cs
mathewc: The reason the code is like this is because IPrimaryHostStateProvider is a job host level service, registered [here](https://github.com/Azure/azure-webjobs-sdk/blob/206e194e03d82a101a6b5dffb7b4c3904730801d/src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsServiceCollectionExtensions.cs#L90), whereas ScriptHostIdProvider and HostNameProvider are WebHost level services. It can't be taken as a ctor param unfortunately.

src/WebJobs.Script/Host/HostIdValidator.cs
mathewc: No, the default for overwrite is false. This update will fail if the blob already exists.

src/WebJobs.Script/Host/HostIdValidator.cs
mathewc: The hostname isn't truncated, only the host ID is.

src/WebJobs.Script/Properties/Resources.resx
mathewc: No not yet. I think this can go in and the doc can come online later. Who owns the doc process - I've drafted up some content.

src/WebJobs.Script/Host/HostIdValidator.cs
mathewc: I added a debug log below when we write the host ID record. Not sure we want to log when we successfully validate - that might just be noise?

src/WebJobs.Script/Host/HostIdValidator.cs
mathewc: Done

src/WebJobs.Script/Host/HostIdValidator.cs
paulbatum: Got it. I am a little worried about side effects, like a large number of messages being written to poison queues, because the app is in a restart loop due to this error. For HTTP there is nothing we can/should do because we dont want to impact cold start. But for retriable event sources like queues, the impact of terminating the app 10 seconds after start could be a problem. I think it might be better to block listeners on this check, though not sure if that is going to be straightforward..

mathewc: Http processing for the host is gated on all hosted services starting. The hosted service WebJobsScriptHostService is responsible for starting the job host (and starting the listeners) and the host is started [here](https://github.com/Azure/azure-functions-host/blob/4764c25d885cdfb1d84af3b80861b181ce658faa/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs#L292). If we were to change that service to only start the JobHost _after_ the host ID check it would ensure we only start listeners after the ID is validated, however all apps would then be paying a coldstart cost (both for http, as well as background triggers since their startup is delayed). Given the lengths we've went to recently to remove all such storage calls off the startup path, I don't think we want to regress this.

A compromise/improvement I can make is to remove the IsPrimary check which will allow me to reduce the delay on this check to our standard coldstart delay of 5s, which minimizes how long the host is allowed to run before the issue is discovered. That does mean all instances will be doing the check, but that's not a huge problem.

paulbatum: yeah I agree this can't be part of the http startup path.

For now I'm going to assume we just don't have the code structured in a way that can allow a sequence where 1) http listener starts, 2) checks run, 3) other listeners start.. I get that we only have a limited amount of time to put this check in and we can't be engaging in larger refactorings right now..

I do not think the 5 seconds vs 10 seconds will make much of a difference. Either way its enough time for us to pull a bunch of messages of queues, start processing user code, and then terminate. 

Even with this potential problem, I still think the change is worth it. Just wanted to make sure we talked it through.

paulbatum: Actually wait a minute..your comment about IsPrimary makes me wonder.. why are we checking that? Lets say the customer has configured the app to run on an app service plan with 3 instances.. are you saying the primary will shutdown, but the other two instances will run?

mathewc: Ha good point - yeah you're right :) IsPrimary made sense when this was just an error log, but once we moved to host shutdown, clearly we need shutdown to happen on all instances. I'll remove the IsPrimary check.

src/WebJobs.Script/Host/HostIdValidator.cs
mathewc: Removed the IsPrimary check

brettsam (approved):

src/WebJobs.Script/Host/HostIdValidator.cs
fabiocav: Are you sure? According to the docs, that isn't the case with this overload (it's inconsistent with some others): https://docs.microsoft.com/en-us/dotnet/api/azure.storage.blobs.blobclient.uploadasync?view=azure-dotnet#Azure_Storage_Blobs_BlobClient_UploadAsync_System_BinaryData_

It's possible the docs are wrong, it looks like this will overwrite an existing blob.

src/WebJobs.Script/Host/HostIdValidator.cs
fabiocav: With the language version we're using, we should be able to use a discard, right?

src/WebJobs.Script/Properties/Resources.resx
mathewc: I've created the aka.ms link and have pointed it to a wiki page that is now in progress.

src/WebJobs.Script/Host/HostIdValidator.cs
mathewc: Yes I'm sure. It results in a 409. Please give it a try.

fabiocav: I believe you :) Looks like just bad docs.

PR #7683, opened 2021-09-20
Deprecating Microsoft.Azure.KeyVault in favour of Azure.Identity

### LINKED ISSUE #7055: Add a SecretStorageProvider for Key Vault that uses Azure.Identity
The existing Key Vault provider makes use of a library that is considered deprecated. 

`Azure.Identity` is the better choice for a dependency here, and it allows us to collapse the configuration, relying just on a single connection name. `Microsoft.Extensions.Azure` might be the best way to bring this in. This aligns with general design for referencing identities. User-assigned support is notable as still TBD from a configuration perspective, but should be handled by that connection abstraction.

Note that this must be a new provider within the host 3.0 context, as it would be a breaking change against the existing provider. In a future major version, the existing provider could be dropped in favor of this one.

liliankasem:

Resolves #7055 
Breaking change proposal: https://github.com/Azure/Azure-Functions/issues/2048

The existing Key Vault provider makes use of a library that is considered deprecated, this PR aims to update the Key Vault Secrets provider to use the `Azure.Identity` and  `Azure.Security.KeyVault.Secrets` libraries instead.

### Validated scenarios

Scenarios to validate both without a vnet and with a key vault inside a vnet.

- [x] Run hosted using AzureWebJobsSecretStorageType=keyvault, without AzureWebJobsSecretStorageKeyVaultUri; validate we handle the error gracefully and present a reasonable message
    - 🛠 `Failed to create mykey: Encountered an error (InternalServerError) from host runtime.`
    - Address fix as part of #7687 

### System-assigned managed identity

- [x] Run hosted using AzureWebJobsSecretStorageKeyVaultUri, without a system-assigned identity existing for the app; validate we handle the error gracefully and present a reasonable message
    - `Failed to create mykey: error creating or updating function key`
- [x] Run hosted using AzureWebJobsSecretStorageKeyVaultUri, with a system-assigned identity on the app that does not have permissions to the vault; validate we handle the error gracefully and present a reasonable message
    - `Failed to create mykey: Encountered an error (Forbidden) from extensions API.`
- [x] Run hosted using AzureWebJobsSecretStorageKeyVaultUri, with a system-assigned identity on the app that does have permissions to the vault; validate successful
    - `Successfully created mykey`

### User-assigned managed identity

- [x] Run hosted using AzureWebJobsSecretStorageKeyVaultUri and AzureWebJobsSecretStorageKeyVaultClientId, without the corresponding identity being assigned to the app; validate we handle the error gracefully and present a reasonable message
    - `Failed to create mykey: error creating or updating function key`
- [x] Run hosted using AzureWebJobsSecretStorageKeyVaultUri and AzureWebJobsSecretStorageKeyVaultClientId, with the corresponding user-assigned identity on the app but lacking permissions to the vault; validate we handle the error gracefully and present a reasonable message
    - `Failed to create mykey: Encountered an error (Forbidden) from extensions API.`
- [x] Run hosted using AzureWebJobsSecretStorageKeyVaultUri and AzureWebJobsSecretStorageKeyVaultClientId, with a corresponding user-assigned identity on the app that does have permissions to the vault; validate successful
    - `Successfully created mykey`

> Adding the ClientId with system-assigned MI enabled = success case

### App registration service principal

- [x] Run hosted using AzureWebJobsSecretStorageKeyVaultUri, AzureWebJobsSecretStorageKeyVaultClientId and AzureWebJobsSecretStorageKeyVaultTenantId, without AzureWebJobsSecretStorageKeyVaultClientSecret; validate we handle the error gracefully and present a reasonable message
    - `Failed to create mykey: Error creating or updating function key.`
- [x] Run hosted using AzureWebJobsSecretStorageKeyVaultUri, AzureWebJobsSecretStorageKeyVaultTenantId, AzureWebJobsSecretStorageKeyVaultClientId and AzureWebJobsSecretStorageKeyVaultClientSecret, with a service principal that does not have permissions to the vault; validate we handle the error gracefully and present a reasonable message
    - `Failed to create mykey: Encountered an error (Forbidden) from extensions API.`
- [x] Run hosted using AzureWebJobsSecretStorageKeyVaultUri, AzureWebJobsSecretStorageKeyVaultTenantId, AzureWebJobsSecretStorageKeyVaultClientId and AzureWebJobsSecretStorageKeyVaultClientSecret, with a service principal that does have permissions to the vault; validate successful
    - `Successfully created mykey`

### Pull request checklist


src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
```diff
@@ -328,7 +328,7 @@ public async Task<bool> DeleteSecretAsync(string secretName, string keyScope, Sc
                 return secrets;
             }, secretsFactory).ContinueWith(t =>
             {
-                if (t.IsFaulted && t.Exception.InnerException is KeyVaultErrorException)
+                if (t.IsFaulted && t.Exception.InnerException is RequestFailedException)
```
liliankasem, line 331: I think this is the equivalent in the new library but I'm not 100% sure, will need to confirm through testing as it is not well documented

src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
```diff
@@ -36,8 +36,11 @@ public static class EnvironmentSettingNames
...
         public const string FunctionsV2CompatibilityModeKey = "FUNCTIONS_V2_COMPATIBILITY_MODE";
         public const string EasyAuthEnabled = "WEBSITE_AUTH_ENABLED";
-        public const string AzureWebJobsSecretStorageKeyVaultName = "AzureWebJobsSecretStorageKeyVaultName";
-        public const string AzureWebJobsSecretStorageKeyVaultConnectionString = "AzureWebJobsSecretStorageKeyVaultConnectionString";
+
+        // You can auth through environment variables, do we want to use them?
+        // https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/identity/Azure.Identity#environment-variables
+        public const string AzureWebJobsSecretStorageKeyVaultUri = "AzureWebJobsSecretStorageKeyVaultUri";
+        public const string AzureWebJobsSecretStorageKeyVaultClientId = "AzureWebJobsSecretStorageKeyVaultClientId";
         public const string AzureWebsiteArmCacheEnabled = "WEBSITE_FUNCTIONS_ARMCACHE_ENABLED";
```
liliankasem, line 43: As noted in the comment, `DefaultAzureCredential` can be configured with environment variables that are [predefined](https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/identity/Azure.Identity#environment-variables), do we want to use those at all?

test/WebJobs.Script.Tests/Security/KeyVaultSecretsRepositoryTests.cs
```diff
@@ -61,34 +64,36 @@ public void FunctionKeys(string functionName, string secretName)
 
         [Theory]
         [MemberData(nameof(FindSecretsDataProvider.TestCases), MemberType = typeof(FindSecretsDataProvider))]
-        public void FindSecrets(Func<SecretItem, bool> comparison, List<string> expectedMatches)
+        public async Task FindSecrets(Func<SecretProperties, bool> comparison, List<string> expectedMatches)
         {
-            List<IEnumerable<SecretItem>> secretsPages = new List<IEnumerable<SecretItem>>()
+            // The type 'IAsyncEnumerable<T>' exists in both 'System.Interactive.Async, Version=3.2.0.0, Culture=neutral,
+            // PublicKeyToken=94bc3704cddfc263' and 'System.Runtime, Version=6.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
+            AsyncPageable<SecretProperties> secretsPages = new List<IAsyncEnumerable<SecretProperties>>
```
liliankasem: Trying to import 'IAsyncEnumerable' gives an error:

```csharp
The type 'IAsyncEnumerable<T>' exists in both 'System.Interactive.Async, Version=3.2.0.0, Culture=neutral,
PublicKeyToken=94bc3704cddfc263' and 'System.Runtime, Version=6.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
```

I've tried importing directly through the full namespace and I'm not having much luck. It's been a minute since I've been in the csharp world, if someone knows what's up, let me know! :D

src/WebJobs.Script.WebHost/Security/KeyManagement/KeyVaultSecretsRepository.cs
```diff
@@ -197,52 +193,39 @@ private async Task<ScriptSecrets> ReadFunctionSecrets(string functionName)
...
-                functionSecrets.Keys.Add(SecretBundleToKey(item, prefix));
+                KeyVaultSecret item = task.Result;
+                functionSecrets.Keys.Add(KeyVaultSecretToKey(item, prefix));
             }
 
             return functionSecrets;
         }
 
-        public static async Task<List<IEnumerable<SecretItem>>> GetKeyVaultSecretsPagesAsync(KeyVaultClient keyVaultClient, string keyVaultBaseUrl)
+        public static AsyncPageable<SecretProperties> GetKeyVaultSecretsPagesAsync(SecretClient keyVaultClient)
```
liliankasem, line 205: This one I have been debating a bit. I think taking advantage of the new mechanism with IAsyncEnumerable is the right way to go, but I want to make sure it doesn't have any serious impact to move away from `List<IEnumerable`

src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
liliankasem: It sounds like what we should be doing is using `ManagedIdentityCredential` so that we can use the system managed identity, with the option to provide a client id and use the user managed identity

mattchenderson:
Just addressing questions from the first comment...

For 1 - replacement is fine based on the scope change for the work

For 5 - I think the parameterless constructor should be used by default. The presence of particular config (proposal worth discussion which I can detail more after a proper review: https://github.com/Azure/azure-functions-host/issues/7055#issuecomment-921301997) can tell us to modify things for the user-assigned mechanism. You can kind of see that logic at play in the way Microsoft.Extensions.Azure does things: https://github.com/Azure/azure-sdk-for-net/blob/d0d1a52edd41cf95daf05a89d3dc370334c58412/sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientFactory.cs#L86

Basically they derive the choice from the config format, surfacing a few different options, but it defaults to the parameterless DefaultAzureCredential constructor.

src/WebJobs.Script.WebHost/Security/KeyManagement/DefaultSecretManagerProvider.cs
```diff
@@ -69,9 +69,9 @@ internal ISecretsRepository CreateSecretsRepository()
             else if (secretStorageType != null && secretStorageType.Equals("keyvault", StringComparison.OrdinalIgnoreCase))
             {
                 string azureWebJobsSecretStorageKeyVaultName = Environment.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebJobsSecretStorageKeyVaultName);
-                string azureWebJobsSecretStorageKeyVaultConnectionString = Environment.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebJobsSecretStorageKeyVaultConnectionString);
+                string azureWebJobsSecretStorageKeyVaultUserManagedClientId = Environment.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebJobsSecretStorageKeyVaultUserManagedClientId);
```
mattchenderson: I think we could reasonably name this `azureWebJobsSecretStorageKeyVaultClientId` just to simplify slightly.

src/WebJobs.Script.WebHost/Security/KeyManagement/KeyVaultSecretsRepository.cs
```diff
@@ -29,24 +24,29 @@ public class KeyVaultSecretsRepository : BaseSecretsRepository
...
                 throw new ArgumentNullException(nameof(secretsSentinelFilePath));
             }
 
-            _vaultName = vaultName ?? throw new ArgumentNullException(nameof(vaultName));
+            if (string.IsNullOrEmpty(vaultName))
+            {
+                throw new ArgumentNullException(nameof(vaultName));
+            }
+
+            Uri vaultUri = new Uri($"https://{vaultName}{_environment.GetVaultSuffix()}");
```
mattchenderson: Part of me wonders if we should be asking for `vaultUri` directly instead of asking for the name and constructing the URI.

src/WebJobs.Script.WebHost/Security/KeyManagement/KeyVaultSecretsRepository.cs
```diff
@@ -29,24 +24,29 @@ public class KeyVaultSecretsRepository : BaseSecretsRepository
...
 
-            _keyVaultClient = new Lazy<KeyVaultClient>(() =>
+            _secretClient = new Lazy<SecretClient>(() =>
             {
-                AzureServiceTokenProvider azureServiceTokenProvider = string.IsNullOrEmpty(connectionString) ? new AzureServiceTokenProvider()
-                    : new AzureServiceTokenProvider(connectionString);
-                return new KeyVaultClient(new KeyVaultClient.AuthenticationCallback(azureServiceTokenProvider.KeyVaultTokenCallback));
+                ManagedIdentityCredential credential = string.IsNullOrEmpty(clientId) ? new ManagedIdentityCredential()
```
mattchenderson: I worry about defaulting here to ManagedIdentityCredential, as that wouldn't work locally, technically. Managed Identity is purely a hosted platform feature. DefaultAzureCredential would allow for local development using the developer's identity in a consistent way.

mattchenderson (commented):
I think my only comments are about the surface area, which I'm still thinking about a bit. I will sketch out what that looks like in a bit more detail and provide it as a general comment on the PR.

src/WebJobs.Script.WebHost/Security/KeyManagement/DefaultSecretManagerProvider.cs
liliankasem: Agree, the shorter the better :)

src/WebJobs.Script.WebHost/Security/KeyManagement/KeyVaultSecretsRepository.cs
liliankasem: I did this originally, but then I was thinking it might be better to keep the changes to a minimal? Either way works, I don't have a strong preference

src/WebJobs.Script.WebHost/Security/KeyManagement/KeyVaultSecretsRepository.cs
liliankasem: Great point, I'll use `DefaultAzureCredential` as the default

mattchenderson:
Ok so I have talked myself out of the idea that we would use a "connection" style construction. Especially since this is projected as an environment setting more than a part of the checked-in payload. The indirection is clunky, and it's not like we would have gained anything from that - there really are just the two fields to worry with.

So I think `AzureWebJobsSecretStorageKeyVaultName` as required. and `AzureWebJobsSecretStorageKeyVaultClientId` (modified name per review comment) as optional would be sufficient. The structure of the logic you have in there should be sufficient, and I would just want to see DefaultAzureCredential used to enable the local development story.

With that said, we actually don't _do_ key-based validations in a local context, do we? I might be completely off base. If there isn't an actual local story, then ManagedIdentityCredential is 100% correct.

I am still musing over replacing `AzureWebJobsSecretStorageKeyVaultName`  with `AzureWebJobsSecretStorageKeyVaultUri`, but that is a fairly minor thing. We _could_ present them as alternatives and make the requirement to have one of them, and not both, but I think that introduces unnecessary complexity. So we'll need to make the call on one or the other.

Some basic validation scenarios:

- (Only if local could matter for some reason) Run locally using `AzureWebJobsSecretStorageKeyVaultName` , without giving yourself the right permissions to the vault
- (Only if local could matter for some reason) Run locally using `AzureWebJobsSecretStorageKeyVaultName`, having granted proper permissions
- Run hosted using `AzureWebJobsSecretStorageKeyVaultName`, without a system-assigned identity existing for the app; validate we handle the error gracefully and present a reasonable message
- Run hosted using `AzureWebJobsSecretStorageKeyVaultName`, with a system-assigned identity on the app that does not have permissions to the vault; validate we handle the error gracefully and present a reasonable message
- Run hosted using `AzureWebJobsSecretStorageKeyVaultName`, with a system-assigned identity on the app that does have permissions to the vault
- Run hosted using `AzureWebJobsSecretStorageKeyVaultName` and `AzureWebJobsSecretStorageKeyVaultClientId`, without the corresponding identity being assigned to the app; validate we handle the error gracefully and present a reasonable message
- Run hosted using `AzureWebJobsSecretStorageKeyVaultName` and `AzureWebJobsSecretStorageKeyVaultClientId`, with the corresponding user-assigned identity on the app but lacking permissions to the vault; validate we handle the error gracefully and present a reasonable message
- Run hosted using `AzureWebJobsSecretStorageKeyVaultName` and `AzureWebJobsSecretStorageKeyVaultClientId`, with a corresponding user-assigned identity on the app that does have permissions to the vault
- We should also check for network-restricted vaults as well - again, mostly concerned with graceful handling and error messages here.

src/WebJobs.Script.WebHost/Security/KeyManagement/KeyVaultSecretsRepository.cs
mattchenderson: I think I was wrong here. Local doesn't make sense since in local mode, we don't actually even use the secrets.

liliankasem: Oh, should I stick to what I have then? edit: back to ManagedIdentityCredential it is

src/WebJobs.Script.WebHost/Security/KeyManagement/KeyVaultSecretsRepository.cs
```diff
@@ -29,24 +24,29 @@ public class KeyVaultSecretsRepository : BaseSecretsRepository
...
-        private readonly Lazy<KeyVaultClient> _keyVaultClient;
-        private readonly string _vaultName;
+        private readonly Lazy<SecretClient> _secretClient;
         private readonly IEnvironment _environment;
 
-        public KeyVaultSecretsRepository(string secretsSentinelFilePath, string vaultName, string connectionString, ILogger logger, IEnvironment environment) : base(secretsSentinelFilePath, logger, environment)
+        public KeyVaultSecretsRepository(string secretsSentinelFilePath, string vaultName, string clientId, ILogger logger, IEnvironment environment) : base(secretsSentinelFilePath, logger, environment)
         {
-            if (secretsSentinelFilePath == null)
+            if (string.IsNullOrEmpty(secretsSentinelFilePath))
```
fabiocav, line 33: nit: with the validation/check change, the correct exception type should be thrown (i.e. `ArgumentNullException` for `null` arguments, `ArgumentException` for empty string in this case).

src/WebJobs.Script.WebHost/Security/KeyManagement/KeyVaultSecretsRepository.cs
liliankasem: Using vaultUri

azure-pipelines.yml
```diff
@@ -338,8 +338,8 @@ jobs:
       ConnectionStrings__CosmosDB: $(CosmosDB)
       AzureWebJobsEventHubSender: $(EventHub)
       AzureWebJobsEventHubReceiver: $(EventHub)
-      AzureWebJobsSecretStorageKeyVaultConnectionString: $(KeyVaultConnectionString)
-      AzureWebJobsSecretStorageKeyVaultName: $(KeyVaultName)
+      AzureWebJobsSecretStorageKeyVaultClientId: $(KeyVaultClientId)
+      AzureWebJobsSecretStorageKeyVaultUri: $(KeyVaultUri)
```
liliankasem, line 341: I don't know where `KeyVaultName` is coming from here, is it a pipeline variable? This needs to also be updated to `KeyVaultUri`

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
liliankasem: Seems to be equivalent based on the validation testing

test/WebJobs.Script.Tests/Security/KeyVaultSecretsRepositoryTests.cs
liliankasem: Resolved by removing `System.Interactive.Async` - no longer needed in v4

azure-pipelines.yml
liliankasem: Found this in the KV, added

.vscode/launch.json
```diff
@@ -20,14 +20,15 @@
...
+        "ASPNETCORE_ENVIRONMENT": "Development",
+        "AZURE_FUNCTIONS_ENVIRONMENT": "Development"
       },
       "sourceFileMap": {
         "/Views": "${workspaceFolder}/Views"
       }
     },
     {
-      "name": ".NET Core Attach",
+      "name": ".NET Core Attach ",
```
liliankasem: nit: remove extra whitespace

src/WebJobs.Script.WebHost/Security/KeyManagement/KeyVaultSecretsRepository.cs
```diff
@@ -29,24 +25,29 @@ public class KeyVaultSecretsRepository : BaseSecretsRepository
...
+
+            Uri keyVaultUri = new Uri(vaultUri);
 
-            _keyVaultClient = new Lazy<KeyVaultClient>(() =>
+            _secretClient = new Lazy<SecretClient>(() =>
             {
-                AzureServiceTokenProvider azureServiceTokenProvider = string.IsNullOrEmpty(connectionString) ? new AzureServiceTokenProvider()
-                    : new AzureServiceTokenProvider(connectionString);
-                return new KeyVaultClient(new KeyVaultClient.AuthenticationCallback(azureServiceTokenProvider.KeyVaultTokenCallback));
+                TokenCredential credential = string.IsNullOrEmpty(clientId) ? new DefaultAzureCredential()
```
liliankasem: So, for the non-E2E integration tests we need to use EnvironmentCredentials as we do not have a MSI setup for int tests. But we also want to enable the system managed identity scenario here. I think using `DefaultAzureCredential` is too broad, this opens us up to enabling all of the credential scenarios to users and I don't think we want that.

One suggestion is if it's a test, we use environment, otherwise we default to system managed identity. If there is a client Id provided, we use a user managed identity.

However if we decide to enable the environment scenario for customers too, then I think we should write our own chained credential  (`ChainedCredential `) so not to expose all the options.

src/WebJobs.Script/runtimeassemblies.json
```diff
@@ -6,10 +6,6 @@
       "name": "Azure.Core",
       "resolutionPolicy": "private"
     },
-    {
-      "name": "Azure.Identity",
```
liliankasem, line 10: I think maybe bring this back?

test/WebJobs.Script.Tests.Integration/Host/SecretsRepositoryTests.cs
```diff
@@ -374,13 +374,15 @@ public class Fixture : IDisposable
...
+                Environment = new TestEnvironment();
+
                 var configuration = TestHelpers.GetTestConfiguration();
                 BlobConnectionString = configuration.GetWebJobsConnectionString(ConnectionStringNames.Storage);
-                KeyVaultConnectionString = configuration.GetWebJobsConnectionString(EnvironmentSettingNames.AzureWebJobsSecretStorageKeyVaultConnectionString);
-                KeyVaultName = configuration.GetWebJobsConnectionString(EnvironmentSettingNames.AzureWebJobsSecretStorageKeyVaultName);
-                AzureServiceTokenProvider azureServiceTokenProvider = new AzureServiceTokenProvider(KeyVaultConnectionString);
-                KeyVaultClient = new KeyVaultClient(new KeyVaultClient.AuthenticationCallback(azureServiceTokenProvider.KeyVaultTokenCallback));
-                Environment = new TestEnvironment();
+                KeyVaultClientId = configuration.GetWebJobsConnectionString(EnvironmentSettingNames.AzureWebJobsSecretStorageKeyVaultClientId);
```
liliankasem: remove this as we don't use MSI for integration testing

src/WebJobs.Script.WebHost/Security/KeyManagement/KeyVaultSecretsRepository.cs
liliankasem: Update: after offline discussion the team has decided to go with `DefaultAzureCredential` as there are no real security concerns about enabling all options and the chained approach will lead to more code the team needs to potentially maintain

madelinegordon (dismissed):

src/WebJobs.Script.WebHost/Security/KeyManagement/DefaultSecretManagerProvider.cs
```diff
@@ -68,10 +68,15 @@ internal ISecretsRepository CreateSecretsRepository()
             }
             else if (secretStorageType != null && secretStorageType.Equals("keyvault", StringComparison.OrdinalIgnoreCase))
             {
-                string azureWebJobsSecretStorageKeyVaultName = Environment.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebJobsSecretStorageKeyVaultName);
-                string azureWebJobsSecretStorageKeyVaultConnectionString = Environment.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebJobsSecretStorageKeyVaultConnectionString);
+                string azureWebJobsSecretStorageKeyVaultUri = Environment.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebJobsSecretStorageKeyVaultUri);
```
fabiocav: I realizer you're following the previously established pattern here, but it's odd this is going directly to `Environment` instead. Let's update this to use the `_environment` variable, which makes this a lot simpler to test.

src/WebJobs.Script.WebHost/Security/KeyManagement/KeyVaultSecretsRepository.cs
```diff
@@ -29,24 +25,25 @@ public class KeyVaultSecretsRepository : BaseSecretsRepository
...
+        public KeyVaultSecretsRepository(string secretsSentinelFilePath, string vaultUri, string clientId, string clientSecret, string tenantId, ILogger logger, IEnvironment environment) : base(secretsSentinelFilePath, logger, environment)
         {
-            if (secretsSentinelFilePath == null)
+            if (string.IsNullOrEmpty(secretsSentinelFilePath))
             {
                 throw new ArgumentNullException(nameof(secretsSentinelFilePath));
             }
 
-            _vaultName = vaultName ?? throw new ArgumentNullException(nameof(vaultName));
+            Uri keyVaultUri = string.IsNullOrEmpty(vaultUri) ? throw new ArgumentNullException(nameof(vaultUri)) : new Uri(vaultUri);
```
fabiocav: Same comment about the argument validation and exception type thrown.

src/WebJobs.Script.WebHost/Security/KeyManagement/KeyVaultSecretsRepository.cs
```diff
@@ -29,24 +25,25 @@ public class KeyVaultSecretsRepository : BaseSecretsRepository
...
 
-            _keyVaultClient = new Lazy<KeyVaultClient>(() =>
+            _secretClient = new Lazy<SecretClient>(() =>
             {
-                AzureServiceTokenProvider azureServiceTokenProvider = string.IsNullOrEmpty(connectionString) ? new AzureServiceTokenProvider()
-                    : new AzureServiceTokenProvider(connectionString);
-                return new KeyVaultClient(new KeyVaultClient.AuthenticationCallback(azureServiceTokenProvider.KeyVaultTokenCallback));
+                // If clientSecret and tenantId are provided, use ClientSecret credential; otherwise use managed identity
+                TokenCredential credential = !string.IsNullOrEmpty(clientSecret) && !string.IsNullOrEmpty(tenantId) ? new ClientSecretCredential(tenantId, clientId, clientSecret)
```
fabiocav: nit: Can you wrap `?` and `:` to improve readability here?

src/WebJobs.Script/runtimeassemblies.json
```diff
@@ -34,10 +34,6 @@
       "name": "Microsoft.Identity.Client.Extensions.Msal",
       "resolutionPolicy": "private"
     },
```
fabiocav, line 40: What new dependencies are being brought in? Some of those might already be present, but this file should be updated with any new dependencies to make sure the host is not sharing them

liliankasem: Within the WebHost `Azure.Identity` and `Azure.Security.KeyVault.Secrets` were added. The identity dependency is already in this file (I notice that it is already imported in `WebJobs.Script.csproj` , so the only change would be to add the `Azure.Security.KeyVault.Secrets` dependency here?

Does it matter that these were added to `WebJobs.Script.WebHost` and not `WebJobs.Script`?

src/WebJobs.Script.WebHost/Security/KeyManagement/KeyVaultSecretsRepository.cs
```diff
@@ -197,52 +195,39 @@ private async Task<ScriptSecrets> ReadFunctionSecrets(string functionName)
...
-                functionSecrets.Keys.Add(SecretBundleToKey(item, prefix));
+                KeyVaultSecret item = task.Result;
+                functionSecrets.Keys.Add(KeyVaultSecretToKey(item, prefix));
             }
 
             return functionSecrets;
         }
 
-        public static async Task<List<IEnumerable<SecretItem>>> GetKeyVaultSecretsPagesAsync(KeyVaultClient keyVaultClient, string keyVaultBaseUrl)
+        public static AsyncPageable<SecretProperties> GetKeyVaultSecretsPagesAsync(SecretClient secretClient)
```
brettsam, line 207: This is a nit-picky question, but is this method name correct? I don't know the KeyVault Secret API -- but before it returned `SecretItem` -- now it returns `SecretProperties`. Are these the actual Secrets being returned?

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
brettsam: This also jumped out at me. It seems like the RequestFailedException gives us more information than before -- should we be mapping the status of that InnerException to the proper OperationResult? i.e. should we also map 409 -> Conflict; etc?

I think it'd make me feel a smidge better if we wrote out an error log statement here with the actual Status from the `RequestFailedException` -- then we can tell if there's anything better we could be doing in the future?

brettsam:
I had a couple of small comments...

src/WebJobs.Script/runtimeassemblies.json
fabiocav: No, ultimately, the Web Host will have all the dependencies, direct or transitive.

What you're describing aligns with what I expected. `Azure.Identity` was brought in with the secretless work. `Azure.Security.KeyVault.Secrets` should be added.

src/WebJobs.Script.WebHost/Security/KeyManagement/KeyVaultSecretsRepository.cs
liliankasem: Yeah they're essentially the same thing, both contain the name of the secret so you can get it through `GetSecretAsync(item.Name)` where "item" is the SecretProperties.

Happy to change it to `GetKeyVaultSecretPropertiesPagesAsync` if that makes sense?

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
liliankasem: discussed offline, leaving as is as there is clear logging on what the issue is

src/WebJobs.Script.WebHost/Security/KeyManagement/KeyVaultSecretsRepository.cs
brettsam: Nah, that's okay -- just wanted to understand better (I saw the method went from a bunch of lines down to one, so I took a closer look :))

brettsam (approved):

liliankasem:
# [7055](https://github.com/Azure/azure-functions-host/issues/7055) Deprecated auth dependency

https://azure.microsoft.com/en-us/blog/simplifying-security-for-serverless-and-web-apps-with-azure-functions-and-app-service/

The goal in general is to remove the AppAuthentication library and the older Key Vault library (if we’re even using it).
Currently, we take a dependency on app auth library that is now deprecated, but we should be using the identity library instead. We also want unified behavior across all identity based items in our projects and codebase.

> Customer quote: *A better option could be to store this keys into a Key Vault as backing store – which is a great feature of Azure functions, but I’ve found 
   few people are aware of this due to lack of documentations*

- Options:
  - [ ] Remove the old Key Vault provider for v4 and bring it back post GA as an extension
  - [ ] Add the new library as a dependency without actually using it, then update code post-GA
  - [x] Remove old lib from dependencies, add in the new lib and implement the code changes to migrate

## Questions

- If we swap to the new library, we do not want to keep shipping the old library. What will be the impact of this on users?

    - A: We don't currently have a way of telling how many users are using this feature, but there is work in progress to try and get those figures through adding logs. As for taking a dependency on the previous KV library, this will break code for users. They will need to either update to the Azure.Identity library, or import the old library directly themselves within their Function as it will not longer be packaged up with the host.

- For the credentials provider, do we want to use the default, which will try the environment variables first, then tries managed identity and goes up the chain of other providers?

    - A: After several design discussions, the team landed on using `ManagedIdentityCredential` and `ClientSecretCredential`. The benefit of this is that we can lock in to the key scenarios we want to support (system managed identity, user managed identity, and app registration service principal) without having to worry about the user/password and certificate scenarios. It provides users with a managed identity path and a non-managed identity path to authenticating with their key vault. An added benefit of chaining the two token credential methods we care about, over say the `EnvironmentCredential`, is that we can define our own app setting environment variables and ensure they align with existing app setting prefixes for key vault configuration.

## Code Change

We're using the KeyVault library to instantiate a new KeyVaultClient so that we can manage a secrets repository.

There is a [migration guide](https://docs.microsoft.com/en-us/dotnet/api/overview/azure/app-auth-migration) we can use to make this change.

1. Remove `<PackageReference Include="Microsoft.Azure.KeyVault" Version="3.0.3" />` from WebJobs.Script.WebHost.csproj
    - This package is used in these files and their corresponding tests:
        - SecretManager.cs
        - KeyVaultSecretsRepository.cs
        - KubernetesSecretsRepository.cs
1. Add a new project dependency for Azure.Identity (with Azure.Security.KeyVault.Secrets)
1. Update code to create a new KeyVaultClient equivalent
    - Using the `ManagedIdentityCredential` and `ClientSecretCredential`
1. Update tests to use new library
    - Mostly this is the unit and integration tests
1. Update the settings and environment variables

    azure-functions-host/src/WebJobs.Script.WebHost/Security/KeyManagement/DefaultSecretManagerProvider.cs

    ```csharp
    # Update to add keyVaultUri, clientId, clientSecret, and tenantId
    string azureWebJobsSecretStorageKeyVaultName = Environment.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebJobsSecretStorageKeyVaultName); 
    string azureWebJobsSecretStorageKeyVaultConnectionString = Environment.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebJobsSecretStorageKeyVaultConnectionString); 
    ```

## [Breaking changes] App settings

In v3, you can use app settings to set the Key Vault name and connection string to configure Azure Key Vault integration with your function app. These app setting have been replaced by new app settings that do the same thing.

Old

| App Setting |	Value |
|-------------|-------|
| AzureWebJobsSecretStorageType | keyvault |
| AzureWebJobsSecretStorageKeyVaultName | Key Vault name> |
| [Optional] AzureWebJobsSecretStorageKeyVaultConnectionString | Key Vault connection string |

New

### System-assigned managed identity

- Enabled system managed identity inside your Functions App
- Give the created identity Key Vault secret permissions through access policies

| App Setting |	Value |
|-------------|-------|
| AzureWebJobsSecretStorageType | keyvault |
| AzureWebJobsSecretStorageKeyVaultUri | Key Vault URI |

### User-assigned managed identity

- Create a user-assigned managed identity
- Assign the user managed identity to your Functions App
- Give the created identity Key Vault secret permissions through access policies

| App Setting |	Value |
|-------------|-------|
| AzureWebJobsSecretStorageType | keyvault |
| AzureWebJobsSecretStorageKeyVaultUri | Key Vault URI |
| AzureWebJobsSecretStorageKeyVaultClientId | User managed identity client ID |

### App registration

- Create an AAD app registration
- Give the created identity Key Vault secret permissions through access policies

| App Setting |	Value |
|-------------|-------|
| AzureWebJobsSecretStorageType | keyvault |
| AzureWebJobsSecretStorageKeyVaultUri | Key Vault URI |
| AzureWebJobsSecretStorageKeyVaultTenantId | App registration tenant ID |
| AzureWebJobsSecretStorageKeyVaultClientId | App registration client ID |
| AzureWebJobsSecretStorageKeyVaultClientSecret | App registration client secret |

## Behaviour changes

The core behaviour of how we are using the key vault library to manage secrets has not changed. The only real change has been the auth provider approach we use which is either managed identity or an app registration service principal, where previously you could provide a connection string to your Key Vault instance.

Other notable changes in behaviour:

- Users provide the full Uri of the key vault instead of just the name
- Use of `AsyncPageable` over List<IEnumerable>

## Validation

### Base Steps

- Created a new Azure Function App
- Added `AzureWebJobsSecretStorageType=keyvault` app setting

- [] Run hosted using AzureWebJobsSecretStorageType=keyvault, without AzureWebJobsSecretStorageKeyVaultUri; validate we handle the error gracefully and present a reasonable message
    - 🛠 `Failed to create mykey: Encountered an error (InternalServerError) from host runtime.` - under the covers this is `System.ArgumentNullException : Value cannot be null.`
    - TODO: surface clear error to users for missing arguments 

### System-assigned managed identity

- [x] Run hosted using AzureWebJobsSecretStorageKeyVaultUri, without a system-assigned identity existing for the app; validate we handle the error gracefully and present a reasonable message
    - `Failed to create mykey: error creating or updating function key`
- [x] Run hosted using AzureWebJobsSecretStorageKeyVaultUri, with a system-assigned identity on the app that does not have permissions to the vault; validate we handle the error gracefully and present a reasonable message
    - `Failed to create mykey: Encountered an error (Forbidden) from extensions API.`
- [x] Run hosted using AzureWebJobsSecretStorageKeyVaultUri, with a system-assigned identity on the app that does have permissions to the vault; validate successful
    - `Successfully created mykey`

### User-assigned managed identity

- [x] Run hosted using AzureWebJobsSecretStorageKeyVaultUri and AzureWebJobsSecretStorageKeyVaultClientId, without the corresponding identity being assigned to the app; validate we handle the error gracefully and present a reasonable message
    - `Failed to create mykey: error creating or updating function key`
- [x] Run hosted using AzureWebJobsSecretStorageKeyVaultUri and AzureWebJobsSecretStorageKeyVaultClientId, with the corresponding user-assigned identity on the app but lacking permissions to the vault; validate we handle the error gracefully and present a reasonable message
    - `Failed to create mykey: Encountered an error (Forbidden) from extensions API.`
- [x] Run hosted using AzureWebJobsSecretStorageKeyVaultUri and AzureWebJobsSecretStorageKeyVaultClientId, with a corresponding user-assigned identity on the app that does have permissions to the vault; validate successful
    - `Successfully created mykey`

> Adding the ClientId with system-assigned MI enabled = success case

### App registration service principal

- [x] Run hosted using AzureWebJobsSecretStorageKeyVaultUri, AzureWebJobsSecretStorageKeyVaultClientId and AzureWebJobsSecretStorageKeyVaultTenantId, without AzureWebJobsSecretStorageKeyVaultClientSecret; validate we handle the error gracefully and present a reasonable message
    - `Failed to create mykey: Error creating or updating function key.`
- [x] Run hosted using AzureWebJobsSecretStorageKeyVaultUri, AzureWebJobsSecretStorageKeyVaultTenantId, AzureWebJobsSecretStorageKeyVaultClientId and AzureWebJobsSecretStorageKeyVaultClientSecret, with a service principal that does not have permissions to the vault; validate we handle the error gracefully and present a reasonable message
    - `Failed to create mykey: Encountered an error (Forbidden) from extensions API.`
- [x] Run hosted using AzureWebJobsSecretStorageKeyVaultUri, AzureWebJobsSecretStorageKeyVaultTenantId, AzureWebJobsSecretStorageKeyVaultClientId and AzureWebJobsSecretStorageKeyVaultClientSecret, with a service principal that does have permissions to the vault; validate successful
    - `Successfully created mykey`

Issue #7667, opened 2021-09-14
Improve error message when multiple incompatible trigger bindings are specified


IanKemp:
#### What problem would the feature you're requesting solve? Please describe.
A co-worker accidentally defined a function to have multiple triggers:

```csharp
        public async Task<int> ProcessBlob(
          [QueueTrigger("QueueName", Connection = "ConnectionStringName")] string blobQueue,
          [ActivityTrigger] Orchestration_Context inputs,       
          ILogger log)
```

which results in host start failure, and this rather unhelpful error message:

```ps
A host error has occurred during startup operation '83b24a80-e9fa-477d-a85a-463ab000bb74'.
System.Linq: Sequence contains more than one matching element.
Value cannot be null. (Parameter 'provider')
```

Running `func.exe` manually with `CLI_DEBUG=1` set beforehand, I was able to view the full stack trace, which was actually helpful and led me to [ExternalConfigurationStartupValidator.cs](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/DependencyInjection/ExternalConfigurationStartupValidator.cs#L46). With that it became clear that the issue was with trigger bindings, which allowed me to figure out the above mistake that was made.

#### Describe the solution you'd like
Change the linked code from:

```csharp
                var trigger = function.Bindings.SingleOrDefault(b => b.IsTrigger);
```

to

```csharp
                var triggers = function.Bindings.Where(b => b.IsTrigger);
                if (triggers.Count() > 1)
                {
                    throw new InvalidOperationException($"Function \"{function.Name}\" has multiple triggers defined; this is not permitted");
                }

                var trigger = triggers.SingleOrDefault();
```

I don't believe that doing `Where()`/`Count()` here is particularly problematic from a performance standpoint, as this code is executed on host startup, which (should) happen relatively infrequently and only once during the host lifetime.

#### Describe alternatives you've considered
Painful debugging.

v-anvari:
Thank you for your feedback! We will investigate this scenario and update you with the possibilities.

mathewc:
@IanKemp thanks for taking the time to report this issue. PR https://github.com/Azure/azure-functions-host/pull/7685 addresses this.

IanKemp:
Thanks for the super fast turnaround @mathewc !

PR #7685, opened 2021-09-21
Fix ExternalConfigurationStartupValidator bug (#7667)

mathewc:
Addresses https://github.com/Azure/azure-functions-host/issues/7667. Full validation of triggers happens later in the startup process during indexing [here](https://github.com/Azure/azure-webjobs-sdk/blob/11b99659a7f4fdd62bf1e0370cfd459cf5739f80/src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs#L183) and a proper error message will be logged then. For the purposes of this issue, we just need to be sure the config validator doesn't prematurely choke on this.

### DIFF
--- a/src/WebJobs.Script/Description/FunctionDescriptorProvider.cs
+++ b/src/WebJobs.Script/Description/FunctionDescriptorProvider.cs
@@ -178,13 +178,6 @@ namespace Microsoft.Azure.WebJobs.Script.Description
 
         protected internal virtual void ValidateFunction(FunctionMetadata functionMetadata)
         {
-            // Functions must have a trigger binding
-            var triggerMetadata = functionMetadata.InputBindings.FirstOrDefault(p => p.IsTrigger);
-            if (triggerMetadata == null)
-            {
-                throw new InvalidOperationException("No trigger binding specified. A function must have a trigger input binding.");
-            }
-
             HashSet<string> names = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
             foreach (var binding in functionMetadata.Bindings)
             {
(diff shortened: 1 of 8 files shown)


src/WebJobs.Script/Description/FunctionDescriptorProvider.cs
```diff
@@ -213,6 +220,11 @@ protected internal virtual void ValidateBinding(BindingMetadata bindingMetadata)
             {
                 throw new ArgumentException($"{ScriptConstants.SystemReturnParameterBindingName} bindings must specify a direction of 'out'.");
             }
+
+            if (string.IsNullOrEmpty(bindingMetadata.Type))
```
mathewc: In writing the tests, it turned out that some of the test bindings weren't specifying a Type, causing errors. E.g. in the new trigger validations above, IsTrigger would throw for bindings that didn't specify a Type, since it relies on parsing that value.

src/WebJobs.Script/Description/FunctionDescriptorProvider.cs
```diff
@@ -200,6 +193,20 @@ protected internal virtual void ValidateFunction(FunctionMetadata functionMetada
                     names.Add(binding.Name);
                 }
             }
+
+            // Verify there aren't multiple triggers defined
+            var triggers = functionMetadata.InputBindings.Where(p => p.IsTrigger).ToArray();
+            if (triggers.Length > 1)
+            {
+                throw new InvalidOperationException($"Multiple trigger bindings defined. A function can only have a single trigger binding.");
```
mathewc, line 201: To get a proper error message, we need to do this validation here, otherwise we get a "sequence contains more than one matching element" [here](https://github.com/Azure/azure-functions-host/blob/8596a27c5a3cf1c26546418fe8ac6bcf132f808b/src/WebJobs.Script/Description/FunctionDescriptor.cs#L39) where it's expected that there's only ever a single trigger. So we need to ensure we never create an invalid FunctionDescriptor in the first place.

src/WebJobs.Script/Description/FunctionDescriptorProvider.cs
brettsam: To be clear, this scenario already threw an exception, correct? This is just a better exception? 

Just trying to avoid something now throwing that was accidentally "working" before.

brettsam (commented):
Had one question...

src/WebJobs.Script/Description/FunctionDescriptorProvider.cs
mathewc: Yes - as you can see in the code [here](https://github.com/Azure/azure-functions-host/blob/8596a27c5a3cf1c26546418fe8ac6bcf132f808b/src/WebJobs.Script.Abstractions/Description/Binding/BindingMetadata.cs#L56) for IsTrigger which is called above, we assume Type is always set. This was resulting in a null ref exception before, now a better message.

brettsam (dismissed):

PR #7691, opened 2021-09-23
Improve error handling for GetScriptHostServiceOrNull

mathewc:
When the host is being disposed, this service provider helper can throw due to objects/containers being disposed.

### DIFF
--- a/src/WebJobs.Script/Extensions/IServiceProviderExtensions.cs
+++ b/src/WebJobs.Script/Extensions/IServiceProviderExtensions.cs
@@ -21,11 +21,19 @@ namespace Microsoft.Extensions.DependencyInjection
                 throw new ArgumentNullException(nameof(serviceProvider));
             }
 
-            var hostManager = serviceProvider.GetService<IScriptHostManager>();
-            if (Utility.TryGetHostService(hostManager, out TService service))
+            try
             {
-                return service;
+                var hostManager = serviceProvider.GetService<IScriptHostManager>();
+                if (Utility.TryGetHostService(hostManager, out TService service))
+                {
+                    return service;
+                }
             }
+            catch
+            {
+                // can get exceptions if the host is being disposed
+            }
+
             return null;
         }
     }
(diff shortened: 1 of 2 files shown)


src/WebJobs.Script/Extensions/IServiceProviderExtensions.cs
```diff
@@ -21,11 +21,19 @@ public static class IServiceProviderExtensions
...
-                return service;
+                var hostManager = serviceProvider.GetService<IScriptHostManager>();
+                if (Utility.TryGetHostService(hostManager, out TService service))
+                {
+                    return service;
+                }
             }
+            catch
+            {
+                // can get exceptions if the host is being disposed
```
mathewc, line 34: Applying the same error handling that Utility.TryGetHostService uses [here](https://github.com/Azure/azure-functions-host/blob/405a25dc064d3cc85f5c6419361db87c610386a1/src/WebJobs.Script/Utility.cs#L62)

alrod (approved):

src/WebJobs.Script/Extensions/IServiceProviderExtensions.cs
```diff
@@ -21,11 +21,19 @@ public static class IServiceProviderExtensions
...
+            try
             {
-                return service;
+                var hostManager = serviceProvider.GetService<IScriptHostManager>();
+                if (Utility.TryGetHostService(hostManager, out TService service))
+                {
+                    return service;
+                }
             }
+            catch
```
alrod, line 32: do we need to catch only ObjectDisposedException

mathewc: I had done that originally, but decided to use the same handling that we do elsewhere. Also, another issue with trying to handle specific types is the DryIoc.ContainerException is defined in WebHost which isn't accessible here, so I'd have to catch the base InvalidOperation type. Rather than do that I think what I have here is better.

PR #7694, opened 2021-09-24
throwing exception for invalid services

brettsam:
Fixes https://github.com/Azure/azure-functions-host/issues/7677

implementation of breaking change: https://github.com/Azure/Azure-Functions/issues/2045

I validated this in a private site extension and the behavior is as expected: `Function host is not running.` and this banner in the portal:
image


mathewc (dismissed):

mathewc:
Hopefully the test failures aren't too hard to resolve - might be due to enforcing this check now with some tests attempting to register/override services?

brettsam:
Okay tests are all passing now -- I just had to replace the default validation in a couple places. @mathewc can I get a final signoff?

mathewc (approved):

PR #7702, opened 2021-09-27
Adding new major version for Host.Storage and WebJobs.Extensions

karshinlin:

- Adds new major versions for WebJobs.Host.Storage and WebJobs.Extensions packages.
- Adds reference to Microsoft.Azure.WebJobs.Extensions.Timers.Storage for v12 based ScheduleStorageMonitor
- Removes host implementations for BlobServiceClientProvider and StorageClientProvider (moved to WebJobs.Host.Storage)
  - Removes all IDistributedLockManager implementation and Blob-based DynamicConcurrencyRepository implementation (in WebJobs.Host.Storage)
- Removes host implementation for AzureStorageScheduleMonitor (moved to new WebJobs.Extensions.Timers.Storage)

### Pull request checklist

### Additional information

Additional PR information


src/WebJobs.Script.WebHost/ContainerManagement/LinuxContainerInitializationHostService.cs
```diff
@@ -115,15 +115,16 @@ public virtual async Task<string> Read(string uri)
             // But the blob pointed to by the uri will not exist until the container is specialized.
             // When the blob doesn't exist it just means the container is waiting for specialization.
             // Don't treat this as a failure.
-            var cloudBlockBlob = new CloudBlockBlob(new Uri(uri));
+            var blobClient = new BlobClient(new Uri(uri));
 
-            var blobRequestOptions = new BlobRequestOptions
+            if (await blobClient.ExistsAsync())
```
mathewc: You removed the retry specification here with no replacement? This was added to address issues we were seeing in prod.

mathewc:
There were additional changes made in https://github.com/Azure/azure-functions-host/commit/cbf9ad1b107c52531975f622d7f50f63ab31e592 that can be rolled back that this PR doesn't address currently. Recommend looking through that previous PR and rolling back those items. E.g. I think the changes to azure-pipelines.yml were for ported tests, and that can be removed?

src/WebJobs.Script/StorageProvider/HostAzureStorageProvider.cs
```diff
@@ -18,17 +17,15 @@ namespace Microsoft.Azure.WebJobs.Script
     /// This serves as a placeholder for a concrete implementation of a storage abstraction.
     /// This StorageProvider provides a wrapper to align all uses of storage by the Functions Host
     /// </summary>
-    internal class AzureStorageProvider : IAzureStorageProvider
+    internal class HostAzureStorageProvider : AzureStorageProvider
```
mathewc: Review/fix the comments above, and detail why we have to override the base here.

src/WebJobs.Script.WebHost/ContainerManagement/LinuxContainerInitializationHostService.cs
karshinlin: Do you have more details on these issues seen in prod? I'm not aware of an equivalent RetryPolicy for v12 storage. These methods removed the BlobRequestOptions. The closest thing I'm seeing is related to http headers for the BlobClient to add [BlobRequestConditions ](https://docs.microsoft.com/en-us/dotnet/api/azure.storage.blobs.models.blobrequestconditions?view=azure-dotnet)which don't have a similar RetryPolicy

mathewc: @balag0 made this addition originally and can comment more on whether it's still needed and what we should do

karshinlin: Added back the same retry logic.

karshinlin:
Note that the dependency test is skipped right now for v4 branch, but eventually we need to remove all instances of v12 storage in deps.json except Azure Files to prevent adding this dependency in the future:
 https://github.com/Azure/azure-functions-host/blob/da6a35dc3406ef663d4fac0e22ce87caddbfa766/test/WebJobs.Script.Tests/DependencyTests.cs#L39-L40

src/WebJobs.Script/StorageProvider/HostAzureStorageProvider.cs
```diff
@@ -0,0 +1,87 @@
...
+
+            _blobServiceClientProvider = new BlobServiceClientProvider(componentFactory, logForwarder);
+
+            if (FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagDisableMergedWebHostScriptHostConfiguration))
+            {
+                Configuration = configuration;
+            }
+            else
+            {
+                _ = scriptHostManager ?? throw new ArgumentNullException(nameof(scriptHostManager));
```
mathewc: nit: just do an if check here for this. Also, I recommend just doing null checks for all the params at the top of the ctor, since none should be null, right?

src/WebJobs.Script.WebHost/Security/KeyManagement/BlobStorageSasSecretsRepository.cs
```diff
@@ -13,7 +14,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
     /// </summary>
     public sealed class BlobStorageSasSecretsRepository : BlobStorageSecretsRepository
     {
-        public BlobStorageSasSecretsRepository(string secretSentinelDirectoryPath, string containerSasUri, string siteSlotName, ILogger logger, IEnvironment environment, IAzureStorageProvider azureStorageProvider)
+        public BlobStorageSasSecretsRepository(string secretSentinelDirectoryPath, string containerSasUri, string siteSlotName, ILogger logger, IEnvironment environment, IAzureBlobStorageProvider azureStorageProvider)
```
mathewc: nit: update the param names/fields to reflect the new naming. E.g. blobStorageProvider, or azureBlobStorageProvider

src/WebJobs.Script.WebHost/Security/KeyManagement/DefaultSecretManagerProvider.cs
```diff
@@ -82,7 +84,7 @@ internal ISecretsRepository CreateSecretsRepository()
                 string siteSlotName = _environment.GetAzureWebsiteUniqueSlotName() ?? _hostIdProvider.GetHostIdAsync(CancellationToken.None).GetAwaiter().GetResult();
                 return new BlobStorageSasSecretsRepository(Path.Combine(_options.CurrentValue.SecretsPath, "Sentinels"), secretStorageSas, siteSlotName, _loggerFactory.CreateLogger<BlobStorageSasSecretsRepository>(), _environment, _azureStorageProvider);
             }
-            else if (_azureStorageProvider.ConnectionExists(ConnectionStringNames.Storage))
+            else if ((connectionSection = (_azureStorageProvider as HostAzureStorageProvider).Configuration.GetWebJobsConnectionSection(ConnectionStringNames.Storage)) != null && connectionSection.Exists())
```
mathewc: Recommend guarding against nulls here, e.g. (_asp as HostAzureStorageProvider)?.Configuration?. Better yet, to make this more readable, just add a helper method to this file StorageConnectionExists(_azureStorageProvider)

src/WebJobs.Script.WebHost/WebJobs.Script.WebHost.csproj
```diff
@@ -79,8 +79,8 @@
     <PackageReference Include="Microsoft.Azure.Services.AppAuthentication" Version="1.0.3" />
     <PackageReference Include="Microsoft.Azure.Storage.File" Version="11.1.7" />
     <PackageReference Include="Microsoft.Azure.KeyVault" Version="3.0.3" />
-    <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.31-11877" />
-    <PackageReference Include="Microsoft.Azure.WebJobs.Host.Storage" Version="4.0.4-11874" />
+    <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.32" />
```
mathewc: As noted in the SDK PR, I think we're still in the 3.0.31-xxx release so we shouldn't need to bump minor

src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
```diff
@@ -52,6 +53,10 @@ public static class WebScriptHostBuilderExtension
                 })
                 .AddScriptHost(webHostOptions, configLoggerFactory, metricsLogger, webJobsBuilder =>
                 {
+                    // Adds necessary Azure-based services to the ScriptHost, which will use the host-provided IAzureBlobStorageProvider
```
mathewc: Would like @Fabio @brettsam to look at this one. Shouldn't scoping handle this?

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -185,11 +184,10 @@ public static IHostBuilder AddScriptHostCore(this IHostBuilder builder, ScriptAp
                 })
                 .AddHttp()
                 .AddTimers()
+                .AddTimersStorage()
```
mathewc: Can you use your new .AddTimersWithStorage method here?

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -379,17 +370,10 @@ internal static ExtensionBundleOptions GetExtensionBundleOptions(IConfiguration
...
-            // Custom implementation that the Host overrides
-            services.AddSingleton<ScheduleMonitor, AzureStorageScheduleMonitor>();
-        }
-
-        internal static void AddAzureStorageProvider(this IServiceCollection services)
-        {
-            services.AddAzureStorageCoreServices();
-            services.TryAddSingleton<IAzureStorageProvider, AzureStorageProvider>();
-            services.AddAzureStorageBlobs();
+            services.AddAzureStorageCoreServices(); // We just need the JobHostInternalStorageOptions service
```
mathewc: this comment confuses me - in Functions host we need all of the storage core services, right?

src/WebJobs.Script/StorageProvider/HostAzureStorageProvider.cs
mathewc: nit: just do an if check here for this. Also, I recommend just doing null checks for all the params at the top of the ctor, since none should be null, right?

src/WebJobs.Script/WebJobs.Script.csproj
```diff
@@ -39,12 +39,12 @@
...
   <ItemGroup>
     <!-- Dependencies needed for Storage Providers -->
-    <PackageReference Include="Azure.Core" Version="1.17.0" />
+    <PackageReference Include="Azure.Core" Version="1.19.0" />
     <PackageReference Include="Azure.Identity" Version="1.4.1" />
     <PackageReference Include="Azure.Storage.Blobs" Version="12.9.0" />
-    <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.31-11877" />
-    <PackageReference Include="Microsoft.Azure.WebJobs.Host.Storage" Version="4.0.4-11874" />
-    <PackageReference Include="Microsoft.Extensions.Azure" Version="1.1.0-beta.2" />
+    <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.32" />
```
mathewc: Where is this package version coming from? I don't see it published to nuget/myget. I think this should be 3.0.31-xxxx

src/WebJobs.Script/WebJobs.Script.csproj
```diff
@@ -54,8 +54,9 @@
     <PackageReference Include="Microsoft.Azure.Functions.NodeJsWorker" Version="2.1.2" />
     <PackageReference Include="Microsoft.Azure.Functions.PowerShellWorker.PS7.0" Version="4.0.1049" />
     <PackageReference Include="Microsoft.Azure.Functions.PowerShellWorker.PS7.2" Version="4.0.988" />
-    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions" Version="4.0.4-10859" />
-    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.Http" Version="3.1.1-10859" />
+	<PackageReference Include="Microsoft.Azure.WebJobs.Extensions" Version="5.0.0-beta.1" />
+    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.Http" Version="3.1.2" />
```
mathewc: Needs to be updated based on package version fixes

test/WebJobs.Script.Tests.Integration/Host/SecretsRepositoryTests.cs
```diff
@@ -406,7 +407,7 @@ public Fixture()
 
             public ILoggerProvider LoggerProvider { get; private set; }
 
-            public IAzureStorageProvider AzureStorageProvider { get; private set; }
+            public IAzureBlobStorageProvider AzureStorageProvider { get; private set; }
```
mathewc: nit: rename property

test/WebJobs.Script.Tests.Integration/Storage/AzureStorageProviderTests.cs
```diff
@@ -213,14 +137,14 @@ private async Task VerifyContainerClientAvailable(BlobContainerClient client)
             }
         }
 
-        private static IAzureStorageProvider GetAzureStorageProvider(IConfiguration webHostConfiguration, IConfiguration jobHostConfiguration, JobHostInternalStorageOptions storageOptions = null)
+        private static IAzureBlobStorageProvider GetAzureStorageProvider(IConfiguration webHostConfiguration, IConfiguration jobHostConfiguration, JobHostInternalStorageOptions storageOptions = null)
```
mathewc: fix naming throughout (e.g. GetAzureBlobStorageProvider)

src/WebJobs.Script/WebJobs.Script.csproj
karshinlin: You're right - added the temporary dev version.

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
karshinlin: It's a stale comment. I'll remove it.

src/WebJobs.Script.WebHost/Security/KeyManagement/DefaultSecretManagerProvider.cs
```diff
@@ -93,17 +93,17 @@ internal ISecretsRepository CreateSecretsRepository()
                                                                  siteSlotName,
                                                                  _loggerFactory.CreateLogger<BlobStorageSasSecretsRepository>(),
                                                                  _environment,
-                                                                 _azureStorageProvider);
+                                                                 _azureBlobStorageProvider);
             }
-            else if (_azureStorageProvider.ConnectionExists(ConnectionStringNames.Storage))
+            else if (StorageConnectionExists())
```
karshinlin: Instead of doing a naïve check for the storage connection, should this call `_azureBlobStorageProvider.TryCreateHostingBlobContainerClient(out _)` to see if the storage connection can be used correctly? This is the approach in HostIdValidator (changed in this PR).

src/WebJobs.Script/Host/HostIdValidator.cs
```diff
@@ -133,7 +134,11 @@ internal async Task WriteHostIdAsync(string hostId, HostIdInfo hostIdInfo)
         {
             try
             {
-                var containerClient = _storageProvider.GetBlobContainerClient();
+                if (!_azureBlobStorageProvider.TryCreateHostingBlobContainerClient(out var containerClient))
+                {
+                    throw new InvalidOperationException($"Could not create Hosting BlobContainerClient in {nameof(WriteHostIdAsync)}");
```
karshinlin: Just a note that these methods (WriteHostIdAsync and ReadHostIdInfoAsync are not called directly except for tests, so this throws should not be triggered. ValidateHostIdUsageAsync will silently fail if _azureBlobStorageProvider.TryCreateHostingBlobContainerClient() fails.

src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
```diff
@@ -54,6 +55,10 @@ public static class WebScriptHostBuilderExtension
...
+                    webJobsBuilder.Services.AddSingleton<IAzureBlobStorageProvider>(rootServiceProvider.GetService<IAzureBlobStorageProvider>());
```
karshinlin: swap these two lines

src/WebJobs.Script/Host/HostIdValidator.cs
```diff
@@ -29,19 +30,19 @@ public class HostIdValidator
...
+        private readonly IAzureBlobStorageProvider _azureBlobStorageProvider;
         private readonly IApplicationLifetime _applicationLifetime;
         private readonly HostNameProvider _hostNameProvider;
         private readonly ILogger _logger;
 
         private readonly object _syncLock = new object();
         private bool _validationScheduled;
 
-        public HostIdValidator(IEnvironment environment, IAzureStorageProvider storageProvider, IApplicationLifetime applicationLifetime,
+        public HostIdValidator(IEnvironment environment, IAzureBlobStorageProvider azureBtorageProvider, IApplicationLifetime applicationLifetime,
```
mathewc: Fix spelling

src/WebJobs.Script/Host/HostIdValidator.cs
mathewc: They are called directly a few lines above. What you should do is capture the client created on line 71 and pass that into these method calls, so we don't have to duplicate the calls. It also solves your error problem here.

test/WebJobs.Script.Tests.Integration/Host/HostIdValidatorTests.cs
```diff
@@ -42,16 +45,20 @@ public HostIdValidatorTests()
...
-            _blobContainerClient = azureStorageProvider.GetBlobContainerClient();
+            var azureBlobStorageProvider = TestHelpers.GetAzureBlobStorageProvider(config);
+            azureBlobStorageProvider.TryCreateHostingBlobContainerClient(out _blobContainerClient);
 
-            var mockStorageProvider = new Mock<IAzureStorageProvider>(MockBehavior.Strict);
-            mockStorageProvider.Setup(p => p.ConnectionExists(ConnectionStringNames.Storage)).Returns(() => _storageConfigured);
-            mockStorageProvider.Setup(p => p.GetBlobContainerClient()).Returns(_blobContainerClient);
+            var mockBlobStorageProvider = new Mock<IAzureBlobStorageProvider>(MockBehavior.Strict);
+            mockBlobStorageProvider.Setup(p => p.TryCreateHostingBlobContainerClient(out It.Ref<BlobContainerClient>.IsAny))
+                .Callback((ref BlobContainerClient containerClient) =>
```
mathewc: You're doing this just to moq the out param? I think Moq supports this more naturally: https://stackoverflow.com/questions/1068095/assigning-out-ref-parameters-in-moq.

test/WebJobs.Script.Tests.Integration/Host/HostIdValidatorTests.cs
```diff
@@ -110,6 +116,8 @@ public async Task ValidateHostIdUsageAsync_ExistingHostIdInfoMatches_Succeeds()
         [Fact]
         public async Task ValidateHostIdUsageAsync_Collision_WarningLevel_Logs()
         {
+            _storageConfigured = true;
```
mathewc: Shouldn't have to set this to true here and below since the ctor defaults to true already.

src/WebJobs.Script.WebHost/Security/KeyManagement/DefaultSecretManagerProvider.cs
mathewc: Sure I guess - things will fail pretty much right away anyways since the repository will attempt to create a client

src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
mathewc: Also, based on our discussions, we're going to need good comments on this code since the scoping issues are tricky and future maintainers may not understand

test/WebJobs.Script.Tests.Integration/Storage/AzureBlobStorageProviderTests.cs
```diff
@@ -105,88 +106,11 @@ public void TestAzureStorageProvider_JobHostConfigurationWinsConflict(string con
...
-            var azureStorageProvider = GetAzureStorageProvider(webHostConfiguration, jobHostConfiguration);
-            Assert.True(azureStorageProvider.TryGetBlobServiceClientFromConnection(out BlobServiceClient client, "Storage1"));
+            var azureBlobStorageProvider = GetAzureBlobStorageProvider(webHostConfiguration, jobHostConfiguration);
+            Assert.True(azureBlobStorageProvider.TryCreateBlobServiceClientFromConnection("Storage1", out BlobServiceClient client));
             Assert.Equal("webHostAccount", client.AccountName, ignoreCase: true);
         }
 
-        [Fact]
-        public void TestAzureStorageProvider_ConnectionExistsWorksProperly()
-        {
```
mathewc, line 115: All the checks this test was doing are no longer needed or are covered by the other tests? I get that the method doesn't exist anymore, but the scenarios might still

.nuget/NuGet.Config
```diff
@@ -3,6 +3,7 @@
   <packageSources>
     <add key="nuget.org" value="https://www.nuget.org/api/v2/" />
     <add key="azure_app_service" value="https://www.myget.org/F/azure-appservice/api/v2" />
+    <add key="azure_app_service_v3" value="https://www.myget.org/F/azure-appservice/api/v3" />
```
karshinlin: @mathewc Is this what you were referring to? Do we need to remove the line above for the v2 endpoint?

src/WebJobs.Script.WebHost/Security/KeyManagement/DefaultSecretManagerProvider.cs
karshinlin: Right, things will fail right away when we do use the SecretManager, but the problem is that we decide on the provider in this class here. If it's a broken storage connection, we may want to use throw the InvalidOperationException immediately rather than creating a SecretManager and then always fail to use it. I will change this as proposed.

test/WebJobs.Script.Tests.Integration/Storage/AzureBlobStorageProviderTests.cs
karshinlin: I figure the Host-provided implementation's responsibility is to provide the right BlobServiceClient or BlobContainerClient. How it provides this is an internal implementation detail and shouldn't be tightly coupled with the tests in my opinion if we aren't exposing a ConnectionExists method. The other tests already cover resolving a connection and verifies it can create the right client (the real behavior and expectation of this class).

.nuget/NuGet.Config
mathewc: Yes - just change the existing azure_app_service to the v3 url. We don't need both.

.nuget/NuGet.Config
```diff
@@ -3,6 +3,7 @@
   <packageSources>
     <add key="nuget.org" value="https://www.nuget.org/api/v2/" />
     <add key="azure_app_service" value="https://www.myget.org/F/azure-appservice/api/v2" />
+    <add key="azure_app_service_v3" value="https://www.myget.org/F/azure-appservice/api/v3/index.json" />
```
mathewc: remove this line and just update the existing value

mathewc (dismissed):
Just the couple things to fix, otherwise looks good to me. Get Fabio's final sign off on the service registration/layering changes.

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -289,16 +287,8 @@ public static IHostBuilder AddScriptHostCore(this IHostBuilder builder, ScriptAp
                 if (!applicationHostOptions.HasParentScope)
                 {
                     AddCommonServices(services);
-                    services.AddAzureStorageProvider();
```
karshinlin, line 292: @fabiocav, @brettsam - We shouldn't need this line anyways, right? I recall from when we first added this line that it wasn't necessary since there's no situation we don't have a parent scope (i.e. we will always register the host IAzureBlobStorageProvider in the WebHost).

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
```diff
@@ -280,6 +282,18 @@ private static void AddLinuxContainerServices(this IServiceCollection services)
...
 
+        internal static void AddAzureBlobStorageProvider(this IServiceCollection services)
+        {
+            // Adds necessary Azure services to create clients
+            services.AddAzureClientsCore();
+
+            // HostAzureBlobStorageProvider depends on JobHostInternalStorageOptions to support ability to provide a SAS blob container as the Hosting container.
+            // This is registered in WebJobs.Host.Storage, but since IAzureBlobStorageProvider needs to be accessible in the WebHost layer,
+            // we need to register the JobHostInternalStorageOptions in the WebHost layer too, using the merged configuration implemention in ActiveHostWebJobsOptionsSetup.
+            services.ConfigureOptionsWithChangeTokenSource<JobHostInternalStorageOptions, ActiveHostWebJobsOptionsSetup<JobHostInternalStorageOptions>, SpecializationChangeTokenSource<JobHostInternalStorageOptions>>();
```
karshinlin, line 293: @fabiocav - please take a look at this service registration. I allows JobHostInternalStorageOptions to survive specialization and use the merged configuration defined in ActiveHostWebJobsOptionsSetup. The test for this is here:
https://github.com/Azure/azure-functions-host/pull/7702/files#diff-e53672caedc0b4f5530b82d535089f37d27db2eed50ee88cb33fa22c13e60dddR456

brettsam:
For posterity, here's the failing depsjson test output:

```

Previous file: C:\git\dotnet_6\azure-functions-host\test\WebJobs.Script.Tests\bin\Debug\net6.0\Microsoft.Azure.WebJobs.Script.WebHost.deps.json
New file:      C:\git\dotnet_6\azure-functions-host\src\WebJobs.Script.WebHost\bin\Debug\net6.0\Microsoft.Azure.WebJobs.Script.WebHost.deps.json

  Changed:
    - Azure.Core.dll: 1.17.0.0/1.1700.21.40204 -> 1.19.0.0/1.1900.21.45105
    - Microsoft.Extensions.Azure.dll: 1.1.0.0/1.100.21.10801 -> 1.1.1.0/1.100.121.45201

  Removed:
    - Microsoft.Azure.Storage.Blob.dll: 11.1.7.0/11.1.7.0

  Added:
    - Microsoft.Azure.WebJobs.Extensions.Timers.Storage.dll: 1.0.0.0/1.0.0.0
```

src/WebJobs.Script/runtimeassemblies.json
```diff
@@ -439,6 +435,10 @@
       "name": "Microsoft.Azure.WebJobs.Extensions.Http",
       "resolutionPolicy": "minorMatchOrLower"
     },
+    {
+      "name": "Microsoft.Azure.WebJobs.Extensions.Timers.Storage",
+      "resolutionPolicy": "private"
```
brettsam, line 440: @fabiocav FYI -- I had Kar set this as private.

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
brettsam: At one point we also had a non-WebHost setup that did not always have a parent scope. I'm not sure that's necessary anymore (maybe only in tests where we test the ScriptHost directly).

Can you move this to AddCommonServices() so it works the same as other services that are in that same situation? Or does that throw things off?

brettsam (approved):

src/WebJobs.Script/runtimeassemblies.json
fabiocav: 👍🏼

PR #7705, opened 2021-09-28
removing proxy code

brettsam:
I previously disabled proxies with #7510 by ignoring `proxies.json`. This removes all proxy references, code, and tests.


src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -355,7 +355,7 @@ internal FunctionLoadRequest GetFunctionLoadRequest(FunctionMetadata metadata, M
                     Directory = metadata.FunctionDirectory ?? string.Empty,
                     EntryPoint = metadata.EntryPoint ?? string.Empty,
                     ScriptFile = metadata.ScriptFile ?? string.Empty,
-                    IsProxy = metadata.IsProxy()
+                    IsProxy = false
```
liliankasem, line 358: Just curious, why do we need to keep `IsProxy` around if we're getting rid of it? Is there an instance where this could be true?

liliankasem (dismissed):
looks good to me!

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
brettsam: I think we need to keep this b/c it's part of the protobuf contract (which we share between all versions of the host). I didn't want to pull it out at the risk of breaking workers.

mathewc (dismissed):

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
liliankasem: That makes a lot of sense, thanks :)

mathewc (dismissed):

liliankasem (approved):

Issue #7414, opened 2021-05-27
Analyzer: Suggested HttpClient Usage Pattern


kashimiz:
**Conditions To Trigger Analyzer**
Within a [Function]-decorated method, where a new HttpClient is instantiated..

```csharp
using System;
using System.IO;
using System.Threading.Tasks;
using Microsoft.AspNetCore.Mvc;
using Microsoft.Azure.WebJobs;
using Microsoft.Azure.WebJobs.Extensions.Http;
using Microsoft.AspNetCore.Http;
using Microsoft.Extensions.Logging;
using Newtonsoft.Json;
using System.Net.Http;
 
namespace TestManageConnectionApp
{
    public static class HttpWrong
    {
        [FunctionName("TestHTTPWrong")]
        public static async Task<IActionResult> Run(
            [HttpTrigger(AuthorizationLevel.Function, "get", "post", Route = null)] HttpRequest req,
            ILogger log)
        {
            log.LogInformation("C# HTTP trigger function processed a request.");
 
            var testUri = new Uri("https://reqbin.com/echo");
 
            var testHTTP = new HttpClient();
 
            string responseBody = await testHTTP.GetStringAsync(testUri);
 
            log.LogInformation("Here is the response body{0}", responseBody);
 
            string responseMessage = "This is me saying Hello";
 
            return new OkObjectResult(responseMessage);
        }
    }
}
```

**Diagnostic Level**
Warning

**How to fix violations**
1. Move HttpClient instantiation to a static instance defined outside of the [Function] (see example):

```csharp
namespace TestManageConnectionApp
{
    public static class TestHTTPRight
    {
 
        static readonly HttpClient client = new HttpClient();
        [FunctionName("TestHTTPRight")]
        public static async Task<IActionResult> Run(
            [HttpTrigger(AuthorizationLevel.Anonymous, "get", "post", Route = null)] HttpRequest req,
            ILogger log)
        {
            log.LogInformation("C# HTTP trigger function processed a request.");
 
            var testUri = new Uri("https://reqbin.com/echo");
 
            string responseBody = await client.GetStringAsync(testUri);
 
            log.LogInformation("Here is the response body{0}", responseBody);
 
            string responseMessage = "This is me saying Hello";
 
            return new OkObjectResult(responseMessage);
 
 
        }
    }
}
```

2. Inject HttpClient with DI (see example):

```csharp
using Microsoft.Azure.Functions.Extensions.DependencyInjection;
[...]

PR #7718, opened 2021-09-29
Added suggested HttpClient usage analyzer.

kashimiz:

Addresses #7414

### Pull request checklist

    * [X] Otherwise: https://github.com/MicrosoftDocs/azure-docs-pr/pull/174020

### Additional information

This implementation will raise the diagnostic, but I would like to have some more discussion about the code fix(es) themselves.

Currently, using reflection to get the full assembly name of `System.Net.Http.HttpClient` - however, might want to change this to a string for performance reasons, since analyzers should be fast, and reflection can be an expensive operation. Thoughts?

### DIFF
--- a/src/WebJobs.Script.Analyzers/AvoidNonStaticHttpClientAnalyzer.cs
+++ b/src/WebJobs.Script.Analyzers/AvoidNonStaticHttpClientAnalyzer.cs
@@ -1,0 +1,53 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.Diagnostics;
+using System.Collections.Immutable;
+using System.Net.Http;
+
+namespace Microsoft.Azure.Functions.Analyzers
+{
+    /// <summary>
+    /// AZF0002: Use static HttpClient
+    /// 
+    /// Cause:
+    /// An local declaration that happens inside a Function method, instantiates an HttpClient
+    /// </summary>
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class AvoidNonStaticHttpClientAnalyzer : DiagnosticAnalyzer
+    {
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get { return ImmutableArray.Create(DiagnosticDescriptors.AvoidNonStaticHttpClient); } }
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+            context.EnableConcurrentExecution();
+
+            context.RegisterSyntaxNodeAction(AnalyzeNode, SyntaxKind.ObjectCreationExpression);
+        }
+
+        private static void AnalyzeNode(SyntaxNodeAnalysisContext context)
+        {
+            // check if object is created within function method
+            var containingSymbol = context.ContainingSymbol as IMethodSymbol;
+            if (containingSymbol is null || !containingSymbol.IsFunction(context.Compilation))
+            {
+                return;
+            }
+
+            // check if constructor is HttpClient
+            var httpClientTypeSymbol = context.SemanticModel.Compilation.GetTypeByMetadataName(typeof(HttpClient).FullName);
+            var nodeTypeSymbol = context.SemanticModel.GetTypeInfo(context.Node, context.CancellationToken).ConvertedType;
+
+            if (!httpClientTypeSymbol.IsAssignableFrom(nodeTypeSymbol))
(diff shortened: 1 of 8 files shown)


liliankasem (dismissed):
looks good to me :)

src/WebJobs.Script.Analyzers/AvoidNonStaticHttpClientAnalyzer.cs
```diff
@@ -0,0 +1,53 @@
...
+            context.EnableConcurrentExecution();
+
+            context.RegisterSyntaxNodeAction(AnalyzeNode, SyntaxKind.ObjectCreationExpression);
+        }
+
+        private static void AnalyzeNode(SyntaxNodeAnalysisContext context)
+        {
+            // check if object is created within function method
+            var containingSymbol = context.ContainingSymbol as IMethodSymbol;
+            if (containingSymbol is null || !containingSymbol.IsFunction(context.Compilation))
```
kshyju, line 35: Question: If an HttpClient instance is created inside a private method which is called from the function, should we flag that as well ?

src/WebJobs.Script.Analyzers/DiagnosticDescriptors.cs
```diff
@@ -21,5 +21,11 @@ private static DiagnosticDescriptor Create(string id, string title, string messa
                 messageFormat: "Async void can lead to unexpected behavior. Return Task instead.",
                 category: Constants.DiagnosticsCategories.Usage,
                 severity: DiagnosticSeverity.Error);
+
+        public static DiagnosticDescriptor AvoidNonStaticHttpClient { get; }
+            = Create(id: "AZF0002", title: "Avoid non-static HttpClient usage",
+                messageFormat: "Reuse HttpClient instances to avoid holding more HTTP connections than necessary.", // TODO: fixup for code fixes?
```
kshyju: Are we recommending them to use static http client or should we be suggesting them to use `IHttpClientFactory` with DI ?

kshyju (approved):

src/WebJobs.Script.Analyzers/DiagnosticDescriptors.cs
brettsam: I think we should push for `IHttpClientFactory` usage.

src/WebJobs.Script.Analyzers/AvoidNonStaticHttpClientAnalyzer.cs
brettsam: I think so, yes.

src/WebJobs.Script.Analyzers/DiagnosticDescriptors.cs
kashimiz: The concern with skipping straight to DI/`IHttpClientFactory` usage is that users might not be familiar with DI, or have their system wired for it. The [associated docs](https://github.com/MicrosoftDocs/azure-docs-pr/pull/174020/files#diff-377d0612b7b033b4e7d855bea8766b05cb5729c775f23d6c21758d0b8c4cd5b6) offer both approaches.

@brettsam, @kshyju, are you suggesting we a) direct users *only* towards DI/`IHttpClientFactory` in both docs and diagnostic message, or b) change the wording of the diagnostic to suggest both approaches?

src/WebJobs.Script.Analyzers/AvoidNonStaticHttpClientAnalyzer.cs
kashimiz: I brought up whether we should include this pattern with Fabio, and his concern was that we might end up flagging third-party library methods that instantiate an HttpClient, which the user would have no control over. 

We might be able to restrict the diagnostic getting raised to cases where the method is within the same [workspace or solution](https://docs.microsoft.com/en-us/dotnet/csharp/roslyn-sdk/work-with-workspace#workspace) as the calling function. However, I suggest we might want to leave this as a later improvement. @fabiocav, thoughts?

src/WebJobs.Script.Analyzers/DiagnosticDescriptors.cs
fabiocav: I think we should suggest reuse in a broader way and use an aka.ms link with more thorough documentation outlining the options.

kshyju (dismissed):

src/WebJobs.Script.Analyzers/AvoidNonStaticHttpClientAnalyzer.cs
kashimiz: After offline discussion with Fabio, tracking this as a later improvement.

src/WebJobs.Script.Analyzers/DiagnosticDescriptors.cs
kashimiz: Updated. Feedback on diagnostic category (Performance? Usage?) and aka.ms link format appreciated.

kshyju (approved):

PR #7720, opened 2021-09-29
removing Grpc.Core dependency

brettsam:
Now that our AspNetCore.Grpc implementation has made it to v4.x, we can remove the older Grpc.Core dependency and the corresponding server implementation.


src/WebJobs.Script.Grpc/WebJobs.Script.Grpc.csproj
```diff
@@ -20,11 +20,7 @@
   </PropertyGroup>
   
   <ItemGroup>
-    <PackageReference Include="Grpc.AspNetCore" Version="2.32.0" />
-    <PackageReference Include="Grpc.Core" Version="2.32.0" />
-    <PackageReference Include="Grpc.Tools" Version="2.32.0">
-      <PrivateAssets>All</PrivateAssets>
-    </PackageReference>
+    <PackageReference Include="Grpc.AspNetCore" Version="2.39.0" />
```
brettsam, line 23: `Grpc.Tools` is brought in automatically with `Grpc.AspNetCore` -- and doesn't look like the `<PrivateAssets>All</PrivateAssets>` makes a difference in output.

soninaren (approved):

liliankasem (approved):

PR #7722, opened 2021-09-29
Logging resolved secret repository name

fabiocav:
Minor enhancement to log resolved secret storage providers


src/WebJobs.Script.WebHost/Security/KeyManagement/BaseSecretsRepository.cs
```diff
@@ -50,6 +50,8 @@ public BaseSecretsRepository(string secretsSentinelFilePath, ILogger logger, IEn
 
         public abstract bool IsEncryptionSupported { get; }
 
+        public abstract string Name { get; }
```
fabiocav, line 53: Introducing a new property here to enable the provider to define this and avoid having to log/resolve everything based on the type name only (also avoid the calls to `GetType`  that would be required)

brettsam (dismissed):

mathewc:
Would be good to have a simple test verrifying the new log statement. We have an existing SecretManagerProviderTests suite.

Issue #7724, opened 2021-09-30
Functions V4 should throw exception if Application Insights extension is misconfigured


gzuber:
If the customer is expecting to use application insights (has connection string or instrumentation key set) but doesn't have the extension installed (on azure, not when running locally), the host should fail to start.

We should also have a general warning when developing locally considering how many apps this impacts.

ladeak:
In V4 (NET6 function) I get an error when I try to run it in azure (previously I did not get this error):
```
Functions runtime error
System.Private.CoreLib: Could not load file or assembly 'Microsoft.ApplicationInsights, Version=2.12.2.43657, Culture=neutral, PublicKeyToken=31bf3856ad364e35'. The system cannot find the file specified.
```
and also shows Runtime version : Error.

Installing this package with my app does not solve the issue. How can I figure out and fix the issue?

gzuber:
Hi @ladeak would you be able to share you csproj file? I mainly need to see which packages you're referencing. Feel free to email it to me directly if that's easier. Thanks!

ladeak:
Sure:

```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <TargetFramework>net6.0</TargetFramework>
    <AzureFunctionsVersion>v4</AzureFunctionsVersion>
    <LangVersion>preview</LangVersion>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="LaDeak.JsonMergePatch.Http" Version="1.0.0" />
    <PackageReference Include="LaDeak.JsonMergePatch.SourceGenerator" Version="1.0.0" />
    <PackageReference Include="Microsoft.AspNetCore.Authentication.JwtBearer" Version="5.0.10" />
    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask" Version="2.5.1" />
    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.EventHubs" Version="4.2.0" />
    <PackageReference Include="Microsoft.Azure.Functions.Extensions" Version="1.1.0" />
    <PackageReference Include="Microsoft.NET.Sdk.Functions" Version="3.0.13" />
    <PackageReference Include="Microsoft.Azure.Devices" Version="1.35.0" />
    <PackageReference Include="System.Memory" Version="4.5.4" />
  </ItemGroup>
  <ItemGroup>
    <ProjectReference Include="..\IoTProcessorCore\IoTProcessorCore.csproj" />
[...]

PR #7725, opened 2021-09-30
Added support for Application Insights in ExtensionBundle filtering. Added warning/errors if Application Insights is misconfigured.

### LINKED ISSUE #7723: Extension Bundles filtering logic should support new special case "_"
Extension Bundles added a new special-case binding with value "_" for the application insights extension can be recognized and treated differently in V4, and not installed in V3. The filtering logic in the host needs to recognize this and act upon it.

gzuber:

resolves #7723 and #7724

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
+++ b/src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
@@ -357,7 +357,7 @@ namespace Microsoft.Azure.WebJobs.Script.DependencyInjection
 
             if (_isSelfHost)
             {
-                _logger.LogDebug("In order to use Application Insights in Azure Functions V4 and above, you must install the Application Insights Extension.");
+                _logger.LogWarning("In order to use Application Insights in Azure Functions V4 and above, you must install the Application Insights Extension.");
             }
             else
             {
--- a/src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
+++ b/src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
@@ -27,23 +27,33 @@ namespace Microsoft.Azure.WebJobs.Script.DependencyInjection
     public class ScriptStartupTypeLocator : IWebJobsStartupTypeLocator
     {
         private readonly string _rootScriptPath;
+        private readonly IEnvironment _environment;
         private readonly ILogger _logger;
         private readonly IExtensionBundleManager _extensionBundleManager;
         private readonly IFunctionMetadataManager _functionMetadataManager;
         private readonly IMetricsLogger _metricsLogger;
+        private readonly bool _isSelfHost;
         private readonly Lazy<IEnumerable<Type>> _startupTypes;
 
         private static readonly ExtensionRequirementsInfo _extensionRequirements = DependencyHelper.GetExtensionRequirements();
         private static string[] _builtinExtensionAssemblies = GetBuiltinExtensionAssemblies();
 
-        public ScriptStartupTypeLocator(string rootScriptPath, ILogger<ScriptStartupTypeLocator> logger, IExtensionBundleManager extensionBundleManager,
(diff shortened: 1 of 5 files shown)


src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
```diff
@@ -323,6 +332,50 @@ void CollectError(Type extensionType, Version minimumVersion, ExtensionStartupTy
...
+                bool applicationInsightsInstalled = startupTypes.Exists(t => t.Name.Equals("ApplicationInsightsWebJobsStartup"));
+                if (applicationInsightsInstalled && !IsApplicationInsightsSettingPresent())
+                {
+                    throw new HostInitializationException($"The Application Insights Extension is installed but is not properly configured. " +
+                                                          $"Please define the \"{EnvironmentSettingNames.AppInsightsConnectionString}\" app setting.");
+                }
+                else if (!applicationInsightsInstalled && IsApplicationInsightsSettingPresent())
+                {
+                    throw new HostInitializationException($"{EnvironmentSettingNames.AppInsightsConnectionString} or {EnvironmentSettingNames.AppInsightsInstrumentationKey} " +
+                                                          $"is defined but the Application Insights Extension is not installed. Please install the Application Insights Extension.");
```
gzuber: Would this break all extension bundle-configured apps if this change is released and used before extension bundles has been released with the App Insights extension?

gzuber:
I'm having a bit of trouble with the tests for this PR. I assumed I would add some `GetExtensionsStartupTypes` tests to test out `GetExtensionsStartupTypesAsync`, but the problem with that method is that it tries to _load_ the dll. I don't believe there's an example of any binding in those tests that isn't built in. Is there a better place to put these tests?

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
```diff
@@ -323,6 +332,50 @@ void CollectError(Type extensionType, Version minimumVersion, ExtensionStartupTy
...
+            }
+
+            return extension.Bindings.Intersect(bindingsSet, StringComparer.OrdinalIgnoreCase).Any();
+        }
+
+        private void ValidateApplicationInsightsConfig(List<Type> startupTypes)
+        {
+            if (_isSelfHost)
+            {
+                _logger.LogWarning("In order to use Application Insights in Azure Functions V4 and above, please install the Application Insights Extension.");
```
brettsam: let's make an aka.ms link for this so we can provide more information -- I'll ping you separately.

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
brettsam: Yes -- could we have another check to see if this is using bundles and only throw if it is not?

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
```diff
@@ -323,6 +332,54 @@ void CollectError(Type extensionType, Version minimumVersion, ExtensionStartupTy
...
+        }
+
+        internal static void ValidateApplicationInsightsConfig(List<Type> startupTypes, bool isSelfHost, bool isApplicationInsightsSettingPresent, bool isExtensionBundleConfigured, ILogger logger)
+        {
+            if (isSelfHost)
+            {
+                logger.LogWarning("In order to use Application Insights in Azure Functions V4 and above, please install the Application Insights Extension." +
+                                  "See https://aka.ms/func-applicationinsights-extension for more details.");
+            }
+            else if (!isExtensionBundleConfigured)
```
gzuber, line 358: I did this because in both cases, extensionbundles shouldn't be involved. If it's installed but there's no app setting, we shouldn't throw an exception because the app setting is being checked earlier in `IsValidBindingMatch` which is only called for bundles. If it's not installed and there's an app setting, I left a comment on why that could be problematic.

brettsam (approved):

PR #7736, opened 2021-10-01
refactoring dependencies for Abstractions

brettsam:
The Abstractions project was referencing a bunch of CodeAnalysis packages... but all it needed was System.Collections.Immutable. With this refactoring, we can remove a bunch of CodeAnalysis dependencies in v4.


pragnagopa (approved):
👍

fabiocav (approved):

PR #7743, opened 2021-10-05
[pack] Removing dependency on Microsoft.Build for generating project XML files

### LINKED ISSUE #7737: Use System.Xml to create msbuild xml file instead of Microsoft.Build
# [7737](https://github.com/Azure/azure-functions-host/issues/7737) Removing the Microsoft.Build dependency

Taking a dependency on the MSBuild library just to generate an XML file leads to ton of unused project dependencies that increase the size of our project. We can easily remove this library and use System.Xml to generate the exact same XML payload.

From the screenshot below, the only dependency we actually use is `System.Runtime.Loader` which we can import on our own.

Extra dependencies

The benefit of using the `Microsoft.Build` library is that we get some prebuilt functionality around creating msbuild.xml files i.e. we have a model to represent the schema for the XML file, and methods such as `AddPropertyGroup` which make it really easy to add new XML elements. However if all we're using this for is generating an XML file, we can easily create those elements ourselves, without having to take on all the additional dependencies that come with `Microsoft.Build` that we never use.

## Design

`Microsoft.Build` provides `ProjectRootElement` which enables us to create a new msbuild project file. It has functions such as `ProjectRootElement.Create`, `AddItemGroup()` and `AddPropertyGroup()` which are used to create something like the following XML output:

```xml
<Project Sdk="Microsoft.NET.Sdk">
    <PropertyGroup>
        <VersionFile>netstandard2.0</VersionFile>
    </PropertyGroup>
    <ItemGroup>
        <PackageReference Include="Microsoft.CodeAnalysis" Version="3.3.1" />
    <ItemGroup>
</Project>
```

Our implementation would need to replace those methods with our own to achieve the same purpose, but with less dependencies. All those methods do under the covers is create an XML file, so we should be able to just use the builtin `System.Xml` library to create the same XML payload.

### Identified functionality for parity

- Creating a `Project` root element with attributes
    - Create from file if it exists
    - Create from scratch
    - Should contain a searchable list of elements
- Creating a `PropertyGroup` with the following properties:
    - `TargetFramework`
    - `WarningsAsErrors` with the following properties:
- Creating an `ItemGroup`
    - `PackageReference`
- Creating an `Item`
- Create a `PackageReference`
- Remove a `PackageReference`
- Removing a child element

#### Option 1

We already have a class `ProjectExtensions` that extends the `ProjectRootElement`. We can reuse this class to extend `XmlDocument` and add all of the required functionality defined [here](#identified-functionality-for-parity)

#### Option 2

We create a new class to replace `ProjectRootElement` which extends `XmlDocument`. We essentially create our own type for the MSBuild Project and only implement the [core functionality we need](#identified-functionality-for-parity).

#### Microsoft.Build to System.Xml mappings

Either way, we will be surfacing System.Xml elements that MS Build wraps through `ProjectRootElement`, so we will need to define the mappings we replace, such as:

- [ProjectRootElement](https://docs.microsoft.com/en-us/dotnet/api/microsoft.build.construction.projectrootelement?view=msbuild-16-netcore) = XML document
- [ProjectItemElement](https://docs.microsoft.com/en-us/dotnet/api/microsoft.build.construction.projectitemelement?view=msbuild-16-netcore) = XML element `Item`
- [ProjectItemGroupElement](https://docs.microsoft.com/en-us/dotnet/api/microsoft.build.construction.projectitemgroupelement?view=msbuild-16-netcore) = XML element for `ItemGroup`

//TODO: class diagram

### Concerns

- We will essentially have to build our own wrapper for creating msbuild XML files. This could led to some maintenance down the road if the schema changes
    - On the plus side, we only need a select number of elements and properties, and how often will the schema be changing?

## Scope of Work

The scope of work is pretty small, it's mostly cleaning up unused references, but the biggest chunk of work would be writing our own `Project` XML file builder which comes from replacing use of `ProjectRootElement`.

- Remove the `Microsoft.Build` library
- Add the `System.Runtime.Loader` library
    - We're using this in our project but it's currently coming through `Microsoft.Build`
- `ExtensionsManager.cs`
    - Replace usage of `ProjectRootElement`
- `PackageManager.cs`
    - Replace usage of `ProjectRootElement`
- `ProjectExtensions.cs`
    - Update existing functions to work with `System.Xml`
    - Implement `Project` builder methods ([see design](#design))
- Clean up unused references to `Microsoft.Build`

liliankasem:
Removing dependency on Microsoft.Build for generating project XML files.

Resolves #7737 

### Pull request checklist


liliankasem:
@brettsam test failure:

```
The dependencies in WebHost have changed and MUST be reviewed before proceeding. Please follow up with brettsam, fabiocav or mathewc for approval

 Removed:
    - Microsoft.Build.dll: 15.1.0.0/15.8.166.59604
    - Microsoft.Build.Framework.dll: 15.1.0.0/15.8.166.59604
```

src/WebJobs.Script/BindingExtensions/ExtensionsManager.cs
```diff
@@ -67,14 +66,14 @@ public async Task AddExtensions(params ExtensionPackageReference[] references)
...
-        private async Task SaveAndProcessProjectAsync(ProjectRootElement project)
+        private async Task SaveAndProcessProjectAsync(XmlDocument project)
         {
             string baseFolder = Path.GetTempPath();
 
             var tempFolder = Path.Combine(baseFolder, "Functions", "Extensions", Guid.NewGuid().ToString());
             Directory.CreateDirectory(tempFolder);
 
-            File.WriteAllText(Path.Combine(tempFolder, ExtensionsProjectFileName), project.RawXml);
+            File.WriteAllText(Path.Combine(tempFolder, ExtensionsProjectFileName), project.InnerXml);
```
liliankasem: I think this actually need to be
```suggestion
            File.WriteAllText(Path.Combine(tempFolder, ExtensionsProjectFileName), project.OuterXml);
```

brettsam:
> @brettsam test failure:
> 
> ```
> The dependencies in WebHost have changed and MUST be reviewed before proceeding. Please follow up with brettsam, fabiocav or mathewc for approval
> 
>  Removed:
>     - Microsoft.Build.dll: 15.1.0.0/15.8.166.59604
>     - Microsoft.Build.Framework.dll: 15.1.0.0/15.8.166.59604
> ```

This makes sense -- I'll let you know how to clean this up once everything is all ready...

src/WebJobs.Script/BindingExtensions/ExtensionsManager.cs
```diff
@@ -104,13 +103,15 @@ public async Task<IEnumerable<ExtensionPackageReference>> GetExtensions()
             }
 
             var project = await GetOrCreateProjectAsync(extensionsProjectPath);
+            var projectElements = project.SelectNodes("//*").OfType<XmlElement>();
```
liliankasem: We do this a lot throughout this file and my concern is this being a potentially expensive operation (depending how how big this gets), so we should really try to optimise and run this less - but then the XmlDocument is constantly evolving so this list is changing. Might be a mute point but raising anyway in case others had thoughts here

src/WebJobs.Script/Extensions/ProjectExtensions.cs
```diff
@@ -1,52 +1,166 @@
...
             {
-                existingPackageReference.Parent.RemoveChild(existingPackageReference);
+                existingPackageReference.ParentNode?.RemoveChild(existingPackageReference);
             }
         }
 
-        public static ProjectItemGroupElement GetUniformItemGroupOrNew(this ProjectRootElement project, string itemName)
+        private static XmlElement GetUniformItemGroupOrNew(this XmlDocument project, string itemName)
         {
-            ProjectItemGroupElement group = project.ItemGroupsReversed.FirstOrDefault(g => g.Items.All(i => i.ItemType == itemName));
```
liliankasem, line 47: How important is it that we're using a reversed item group list here? I ask as we don't have an equivalent available through XmlDocument, so we would need to recreate our own reversed list of XmlElements if this is important. Probably through getting all elements and reversing it through `project.SelectNodes("//*").OfType<XmlElement>();`

src/WebJobs.Script/BindingExtensions/ExtensionsManager.cs
brettsam: Yes, I agree...

src/WebJobs.Script/BindingExtensions/ExtensionsManager.cs
brettsam: Did you look at using System.Xml.Linq for this? https://docs.microsoft.com/en-us/dotnet/standard/linq/linq-xml-overview

I think it'll make it more readable -- and maybe more efficient?

src/WebJobs.Script/BindingExtensions/ExtensionsManager.cs
```diff
@@ -105,12 +101,13 @@ public async Task<IEnumerable<ExtensionPackageReference>> GetExtensions()
 
             var project = await GetOrCreateProjectAsync(extensionsProjectPath);
 
-            return project.Items
-                .Where(i => PackageReferenceElementName.Equals(i.ItemType, StringComparison.Ordinal) && !MetadataGeneratorPackageId.Equals(i.Include, StringComparison.Ordinal))
+            return project.Descendants()?
+                .Where(i => i.Name == PackageReferenceElementName &&
+                            i.Attribute(PackageReferenceIncludeElementName)?.Value != MetadataGeneratorPackageId)
```
brettsam: use `.Equals` here -- `PackageReferenceElementName.Equals(i.Name, StringComparison.Ordinal)`... etc

src/WebJobs.Script/BindingExtensions/ExtensionsManager.cs
```diff
@@ -105,12 +101,13 @@ public async Task<IEnumerable<ExtensionPackageReference>> GetExtensions()
 
             var project = await GetOrCreateProjectAsync(extensionsProjectPath);
 
-            return project.Items
-                .Where(i => PackageReferenceElementName.Equals(i.ItemType, StringComparison.Ordinal) && !MetadataGeneratorPackageId.Equals(i.Include, StringComparison.Ordinal))
+            return project.Descendants()?
+                .Where(i => i.Name == PackageReferenceElementName &&
```
brettsam: Here, `<PackageReference>` needs to be a child of an `<ItemGroup>`, right? Can we ensure that here by checking the parents of each of these?

src/WebJobs.Script/Description/DotNet/PackageManager.cs
```diff
@@ -199,16 +199,16 @@ private static IList<LibraryRange> GetProjectDependencies(string projectFilePath
...
-                   .Select(i => new LibraryRange
-                   {
-                       Name = i.Include,
-                       VersionRange = VersionRange.Parse(i.Metadata.First(m => PackageReferenceVersionElementName.Equals(m.Name, StringComparison.Ordinal)).Value)
-                   })
-                   .ToList();
+                XDocument root = XDocument.Load(reader);
+
+                return root.Descendants()?
+                    .Where(i => i.Name == PackageReferenceElementName)
```
brettsam: use `PackageReferenceElementName.Equals`... here

src/WebJobs.Script/ScriptConstants.cs
```diff
@@ -132,10 +132,20 @@ public static class ScriptConstants
...
 
         public const string ExtensionsProjectFileName = "extensions.csproj";
+        public const string ExtensionsProjectSdkAttributeName = "Sdk";
+        public const string ExtensionsProjectSdkPackageId = "Microsoft.NET.Sdk";
         public const string MetadataGeneratorPackageId = "Microsoft.Azure.WebJobs.Script.ExtensionsMetadataGenerator";
         public const string MetadataGeneratorPackageVersion = "1.1.*";
+        public const string ProjectElementName = "Project";
         public const string PackageReferenceElementName = "PackageReference";
         public const string PackageReferenceVersionElementName = "Version";
+        public const string PackageReferenceIncludeElementName = "Include";
```
brettsam, line 139: Can we move these inside `ProjectExtensions` if that's the only place they're used?

src/WebJobs.Script/Extensions/ProjectExtensions.cs
```diff
@@ -2,51 +2,108 @@
...
+            {
+                existingPackageReference.Remove();
+            }
+        }
+
+        private static void CreateTargetFramework(this XDocument document, string innerText)
+        {
+            if (document.Root.Element(TargetFrameworkElementName) == null)
+            {
+                document.Root.Add(new XElement(TargetFrameworkElementName));
```
brettsam: Is this supposed to be adding a PropertyGroup instead of the TargetFramework?

brettsam: side note -- it'd be good to make each of these little methods `internal` and write some tests to validate they're doing exactly what you exepct.

src/WebJobs.Script/Extensions/ProjectExtensions.cs
brettsam: This should be easier with Linq-to-XML -- but what was this method doing in the first place? Looks like it was trying to get an itemgroup where all items were package refs?

liliankasem: Yup, I should revisit this since switching to XDoc. And yeah, I think that is what it's doing 🤔 I should probably rewrite this since it sounds like there could be ItemGroups where not all elements are of type PackageReferenc

src/WebJobs.Script/Extensions/ProjectExtensions.cs
liliankasem: Yup defo meant `PropertyGroupElementName` good catch :D 

I'll add some unit tests

src/WebJobs.Script/ScriptConstants.cs
liliankasem: can do!

src/WebJobs.Script/BindingExtensions/ExtensionsManager.cs
liliankasem: Honestly, I'm not sure? If they do, happy to add another query to check parent

liliankasem: Added the itemgroup check, feel free to reopen if still an issue

src/WebJobs.Script/Extensions/ProjectExtensions.cs
```diff
@@ -2,51 +2,61 @@
...
-                existingPackageReference.Parent.RemoveChild(existingPackageReference);
+                existingPackageReference.Remove();
             }
         }
 
-        public static ProjectItemGroupElement GetUniformItemGroupOrNew(this ProjectRootElement project, string itemName)
+        internal static XElement GetUniformItemGroupOrNew(this XDocument document, string itemName)
         {
-            ProjectItemGroupElement group = project.ItemGroupsReversed.FirstOrDefault(g => g.Items.All(i => i.ItemType == itemName));
+            XElement group = document.Descendants(ItemGroupElementName).LastOrDefault(g => g.Elements().All(i => i.Name.LocalName == itemName));
```
brettsam: Why do we use `LocalName` in some places and `Name` in others? I think `Name` should work for us (looks like Local removes any prefixes?)

test/WebJobs.Script.Tests/Extensions/ProjectExtensionsTests.cs
```diff
@@ -0,0 +1,159 @@
...
+        {
+            // Arrange
+            var version = "1.0.0";
+
+            // Act
+            testDocument.AddPackageReference("Microsoft.Azure.WebJobs.Extensions.Test", version);
+
+            // Assert
+            Assert.Equal(1, testDocument.Descendants("PackageReference").Count());
+            Assert.Equal(version, testDocument.Descendants("PackageReference").FirstOrDefault().Attribute("Version").Value);
```
brettsam, line 42: Can you also validate that it puts this inside an `ItemGroup`?

brettsam:
Just a couple of questions!

src/WebJobs.Script/Extensions/ProjectExtensions.cs
liliankasem: Good question! It's because of how we are doing the comparison.

If we use `.Equals()` for example: `PackageReferenceElementName.Equals(i.Name.LocalName, StringComparison.Ordinal)`,  we have to use `LocalName` because `Name` in this instance is of type `XName` and so we can't compare it to a string. (`cannot convert from 'System.Xml.Linq.XName' to 'string'`)

But when we're using `==` for comparison, it's able to compare and so `i.Name == itemName`  is possible.

I agree that we need to standardize so we are consistent in our code. Which way do you think we should go?

a) `x.Equals(i.Name.LocalName)` for all comparisons, or
b) ` x == i.Name` for all comparisons

I'm leaning option A

test/WebJobs.Script.Tests/Microsoft.Azure.WebJobs.Script.WebHost.deps.json
```diff
@@ -2655,12 +2598,12 @@
       "path": "microsoft.azure.functions.javaworker/2.0.0",
       "hashPath": "microsoft.azure.functions.javaworker.2.0.0.nupkg.sha512"
     },
-    "Microsoft.Azure.Functions.NodeJsWorker/4.0.0": {
+    "Microsoft.Azure.Functions.NodeJsWorker/3.0.0": {
```
madelinegordon: Does this node worker version correlate to v4?

brettsam: ... this may be out of date now -- we'll have to regen this file i think

brettsam (approved):

PR #7744, opened 2021-10-05
Ensure SyncTriggers clears secrets cache (#7698)

### LINKED ISSUE #7698: Functions Keys stale after rotating storage account until startup cache invalidated.

In an internal investigation of a customer reported incident, we found that after the customer had changed their storage account, there were several hours where the sync triggers operation was still using the stale keys from the old storage account.

This is because the SecretsManager was loading the secrets from the functions startup cache, so it didn't poll the new storage account to see that no keys existed and update with the new keys.

Only when the app went through a code path where the startup cache was not used were the new keys finally populated and used by the Functions Host.

#### Repro steps

1. Create a new storage account and query for the keys.
2. Change the storage connection string to a new storage account.

Note that this will only reproduce in the case where the Functions startup

#### Expected behavior

The keys should change, as the keys are stored in the storage account that the app is no longer referencing.

#### Actual behavior

If the startup cache was used, the keys will remain the same.

#### Known workarounds

Unsure if there is a manual way to bypass the startup cache.

#### Proposed Fix

In the sync triggers code path, bypass the startup cache values and go directly to storage so we catch the invalid cache scenario earlier.

mathewc:
Resolves https://github.com/Azure/azure-functions-host/issues/7698. See that issue my detailed repro steps.

### Pull request checklist

### Additional information

### DIFF
--- a/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
@@ -105,6 +105,8 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
             {
                 await _syncSemaphore.WaitAsync();
 
+                PrepareSyncTriggers();
+
                 var hashBlobClient = await GetHashBlobAsync();
                 if (isBackgroundSync && hashBlobClient == null && !_environment.IsKubernetesManagedHosting())
                 {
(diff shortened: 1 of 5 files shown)


src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -105,6 +105,8 @@ public async Task<SyncTriggersResult> TrySyncTriggersAsync(bool isBackgroundSync
             {
                 await _syncSemaphore.WaitAsync();
 
+                PrepareSyncTriggers();
```
balag0, line 108: We only need this when isBackgroundSync = false?

mathewc: We need it in all cases.

brettsam (approved):

PR #7752, opened 2021-10-07
Update PR template to check internal documentation update

liliankasem:


sidkri (approved):

.github/pull_request_template.md
```diff
@@ -2,16 +2,18 @@
 
 ### Issue describing the changes in this PR
 
-resolves #issue_for_this_pr
```
fabiocav, line 5: For GH keywords, most of our templates standardize on using them with no casing variation.

.github/pull_request_template.md
```diff
@@ -2,16 +2,18 @@
...
+Resolves #issue_for_this_pr
 
 ### Pull request checklist
 
-* [ ] My changes **do not** require documentation changes
-    * [ ] Otherwise: Documentation issue linked to PR
+* [ ] My changes **do not** require external documentation changes
```
fabiocav, line 9: Can you expand on the need to change this? The new template question only applies to internal contributors.

fabiocav (commented):
Not super comfortable having questions directed at internal contributors only. Is this addressing something we've had issues with?

.github/pull_request_template.md
liliankasem: I'll revert this

.github/pull_request_template.md
liliankasem: This is something @sidkri and @kashimiz and I discussed about trying to hold more accountability to updating documentation and design documents as things change, as we felt the previous documentation statement implied just external documentation. Also happy to revert this if it's a issue, I'll leave @sidkri  to comment.

fabiocav: Makes sense. It might be good to revert this for now and discuss how we can improve this flow for internal contributors while also making sure we have a consistent experience across the different repos we maintain, as a lot of the changes that would impact documentation would happen outside of the host.

PR #7755, opened 2021-10-08
working around config binding issue in .NET 6 RC2

brettsam:


src/WebJobs.Script/Config/ExtensionBundleConfigurationHelper.cs
```diff
@@ -28,9 +28,9 @@ public void Configure(ExtensionBundleOptions options)
 
             if (extensionBundleSection.Exists())
             {
+                ConfigureBundleVersion(extensionBundleSection, options);
```
brettsam, line 31: This actually works around the issue -- which manifests if we try to bind configuration to options and the `ExtensionBundleOptions.Version` property is null. By configuring this before the call to `Bind()`, things work. I've added the `AppContext` setting in `Program.cs` to protect customers from this.

fabiocav (approved):

PR #7760, opened 2021-10-11
New API to monitor function app invocation state

### LINKED ISSUE #7188: Add Inflight invocations completed status on the host

### LINKED ISSUE #7792: Service resolution exceptions for IDrainModeManager
We're seeing errors in production for the [drain API](https://github.com/Azure/azure-functions-host/blob/443869e7a41c820288d0d5cb87b31e7569ed118d/src/WebJobs.Script.WebHost/Controllers/HostController.cs#L110) where it can't resolve the service. This is because it's using [FromServices] to access a job host level service, which won't be available if a host isn't running. Related to https://github.com/Azure/azure-functions-host/issues/5161.

```
System.InvalidOperationException : No service for type 'Microsoft.Azure.WebJobs.Host.IDrainModeManager' has been registered.
   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService(IServiceProvider provider,Type serviceType)
   at Microsoft.AspNetCore.Mvc.ModelBinding.Binders.ServicesModelBinder.BindModelAsync(ModelBindingContext bindingContext)
   at async Microsoft.AspNetCore.Mvc.ModelBinding.ParameterBinder.BindModelAsync(ActionContext actionContext,IModelBinder modelBinder,IValueProvider valueProvider,ParameterDescriptor parameter,ModelMetadata metadata,Object value)
   at async Microsoft.AspNetCore.Mvc.Controllers.ControllerBinderDelegateProvider.<>c__DisplayClass0_0.<CreateBinderDelegate>g__Bind|0(??)
   at async Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.<InvokeInnerFilterAsync>g__Awaited|13_0(??)
   at async Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.<InvokeNextResourceFilter>g__Awaited|24_0(??)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.Rethrow(ResourceExecutedContextSealed context)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.Next(State& next,Scope& scope,Object& state,Boolean& isCompleted)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.InvokeFilterPipelineAsync()
   at async Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.<InvokeAsync>g__Logged|17_1(??)
   at async Microsoft.AspNetCore.Builder.RouterMiddleware.Invoke(HttpContext httpContext)
   at async Microsoft.Azure.WebJobs.Script.WebHost.Middleware.FunctionInvocationMiddleware.Invoke(HttpContext context) at D:\a\1\s\src\WebJobs.Script.WebHost\Middleware\FunctionInvocationMiddleware.cs : 43
   at async Microsoft.Azure.WebJobs.Script.WebHost.Middleware.ExceptionMiddleware.Invoke(HttpContext context) at D:\a\1\s\src\WebJobs.Script.WebHost\Middleware\ExceptionMiddleware.cs : 27
```

alrod:

resolves Azure/azure-functions-host#7188 
resolves Azure/azure-functions-host#7792 

### Pull request checklist

### Additional information

Additional PR information

Implement a new Functions Host drain status API that DataRole will poll to determine when it's safe to reclaim a draining instance.

Internal work [item](https://dev.azure.com/azfunctions/Planning/_workitems/edit/827/) and [design](https://microsoft.sharepoint.com/teams/Antares/_layouts/OneNote.aspx?id=%2fteams%2fAntares%2fShared%20Documents%2fAntares%20Feature%20Crew&wd=target%28Azure%20Functions.one%7cBD6BD264-4C3E-463B-8B54-35FCCBE73FA1%2fFunctions%20Retry%5c%2fDrain%20Issue%7cBA4C3DBF-F9CB-4FD2-B495-6232391B0020%2f%29)

Related webjobs sdk host PR Azure/azure-webjobs-sdk/pull/2775


liliankasem:
Hi @alrod, can you link to the issue tracking this change? Is there a design doc that talks about this somewhere? @soninaren and I are working on drain mode and weren't aware more work was coming in at the moment

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -116,6 +116,32 @@ public IActionResult Drain([FromServices] IDrainModeManager drainModeManager)
...
+        [Authorize(Policy = PolicyNames.AdminAuthLevelOrInternal)]
+        [RequiresRunningHost]
+        public IActionResult DrainStatus([FromServices] IScriptHostManager scriptHostManager)
+        {
+            if (Utility.TryGetHostService(scriptHostManager, out IFunctionActivityStatusProvider functionActivityStatusProvider))
+            {
+                var functionActivityStatus = functionActivityStatusProvider.GetStatus();
+                DrainModeStatus status = new DrainModeStatus()
+                {
+                    Status = (functionActivityStatus.OutstandingInvocations == 0 && functionActivityStatus.OutstandingRetries == 0) ?
```
soninaren: How are outstanding invocations calculated. What about events that may be cached and new invocations are scheduled

src/WebJobs.Script/Utility.cs
```diff
@@ -842,6 +841,15 @@ public static bool IsMediaTypeOctetOrMultipart(MediaTypeHeaderValue mediaType)
             {
                 throw new ArgumentNullException(nameof(retryOptions.MaxRetryCount));
             }
+            // Validate SKU limits
```
soninaren: isn't this a breaking change. What if somebody already has higher retry limits.

pragnagopa:
@alrod - Please update description if this PR will be resolving - https://github.com/Azure/azure-functions-host/issues/7188

pragnagopa:
cc @balag0 @divyagandhisethi @pgombar as FYI - as this API will help improve container dispose code paths in Linux Consumption

src/WebJobs.Script.WebHost/Controllers/HostController.cs
alrod: Related webjobs sdk host PR Azure/azure-webjobs-sdk/pull/2775. we idea of the change to monitor only active invocations and retries.

src/WebJobs.Script/Utility.cs
alrod: Retries is in preview also the customer logic would be broken anyway if the limits are exceeded.

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -116,6 +116,36 @@ public IActionResult Drain([FromServices] IDrainModeManager drainModeManager)
             return Ok();
         }
 
+        [HttpGet]
+        [HttpPost]
```
mathewc: Why would we support POST for this?

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -116,6 +116,36 @@ public IActionResult Drain([FromServices] IDrainModeManager drainModeManager)
...
+        [Route("admin/host/drain/status")]
+        [Authorize(Policy = PolicyNames.AdminAuthLevelOrInternal)]
+        [RequiresRunningHost]
```
mathewc: RRH attribute will result in a 503 if the host isn't up and running after a period of time. The other drain API isn't RRH so I'm wondering why these aren't consistent?

src/WebJobs.Script.WebHost/Models/DrainModeStatus.cs
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Models
+{
+    public class DrainModeStatus
+    {
+        [JsonProperty("status")]
+        public string Status { get; set; }
```
mathewc: For the status value, I think we should define an enum DrainModeState with values Disabled/InProgress/Completed or something like that.

Thus we'll have a DrainModeStatus.State property which gets rid of the oddity of a Status model type with a Status property, and is symmetric with what we've done for other APIs (e.g. host/status + HostStatus).

src/WebJobs.Script/Utility.cs
```diff
@@ -842,6 +841,15 @@ public static bool IsMediaTypeOctetOrMultipart(MediaTypeHeaderValue mediaType)
...
+            if (environment.IsWindowsElasticPremium() && retryOptions.MaximumInterval > TimeSpan.FromMinutes(60))
+            {
+                throw new InvalidOperationException("Maximum delay interval must not exceed 60 minutes for Elastic Premium plan");
```
mathewc: I think you should have a separate PR for this work which is unrelated to the new API. Having a separate PR/commit allows us to see/track the changes better in history, rather than sweeping it in with other unrelated changes.

mathewc: Also I don't understand the checks you're doing. E.g. you're only checking MaximumInterval for EP not Consumption, why? Similarly for DelayInterval. Also, the total retry duration for FixedDelay will be MaxRetryCount * DelayInterval, so likely most customer configurations that can run over would bypass your check. For exponential it seems the same issue exists - MaximumInterval is only for a single retry attempt but we'll actually allow MaxRetryCount of those.

alrod: I splitted the PR: https://github.com/Azure/azure-functions-host/pull/7788

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -116,6 +116,40 @@ public IActionResult Drain([FromServices] IDrainModeManager drainModeManager)
...
+                };
+
+                string message = $"Drain Status: {JsonConvert.SerializeObject(state, Formatting.Indented)}, Activity Status: {JsonConvert.SerializeObject(functionActivityStatus, Formatting.Indented)}";
+                _logger.LogInformation(message);
+
+                return Ok(status);
+            }
+            else
+            {
+                // This case should never happen. Because this action is marked RequiresRunningHost,
```
mathewc: These comments are out of date. I guess if we don't have access to an active host, we're not draining?

Once we add Resume capabilities, I do wonder how this API should behave if we've went through a series of Drain/Resume calls, such that we end up with an active host that isn't draining, but one or more orphaned hosts that are draining. Should this API return a result that reflects the drain status of any/all orphaned instances?

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -116,6 +116,40 @@ public IActionResult Drain([FromServices] IDrainModeManager drainModeManager)
             return Ok();
         }
 
+        [HttpGet]
+        [Route("admin/host/drain/status")]
+        [Authorize(Policy = PolicyNames.AdminAuthLevelOrInternal)]
+        public IActionResult DrainStatus([FromServices] IScriptHostManager scriptHostManager, [FromServices] IDrainModeManager drainModeManager)
```
mathewc: Isn't IDrainModeManager only available if there's an active host (since it's added by the WebJobs SDK when host starts)? Might be an existing issue for the other /drain API currently. You could test this by putting the host offline and calling these APIs - I'm thinking you'll see a 500 due to service resolution failure.

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -116,6 +116,40 @@ public IActionResult Drain([FromServices] IDrainModeManager drainModeManager)
...
+                       DrainModeState.Completed : DrainModeState.InProgress;
+                }
+
+                DrainModeStatus status = new DrainModeStatus()
+                {
+                    State = state
+                };
+
+                string message = $"Drain Status: {JsonConvert.SerializeObject(state, Formatting.Indented)}, Activity Status: {JsonConvert.SerializeObject(functionActivityStatus, Formatting.Indented)}";
+                _logger.LogInformation(message);
```
mathewc: Thinking this should be Debug?

test/WebJobs.Script.Tests/Controllers/Admin/HostControllerTests.cs
```diff
@@ -199,5 +202,57 @@ public async Task Ping_ReturnsOk()
             var result = (StatusCodeResult)(await _hostController.Ping(_mockScriptHostManager.Object));
             Assert.Equal((int)HttpStatusCode.OK, result.StatusCode);
         }
+
+        [Fact]
+        public void GetDrainStatus_ReturnsDisabled ()
```
mathewc: Seems like this test should be rolled into the one below - why separate?

test/WebJobs.Script.Tests/Controllers/Admin/HostControllerTests.cs
```diff
@@ -199,5 +202,57 @@ public async Task Ping_ReturnsOk()
...
+            var serviceProviderMock = scriptHostManagerMock.As<IServiceProvider>();
+            serviceProviderMock.Setup(x => x.GetService(typeof(IFunctionActivityStatusProvider))).Returns(functionActivityStatusProvider.Object);
+            drainModeManager.Setup(x => x.IsDrainModeEnabled).Returns(true);
+
+            var result = (OkObjectResult)_hostController.DrainStatus(scriptHostManagerMock.Object, drainModeManager.Object);
+            Assert.Equal((result.Value as DrainModeStatus).State, expectedState);
+        }
+
+        [Fact]
+        public void GetDrainStatus_ReturnsServiceUnavailable()
```
mathewc: For all these test names, please also include a descriptor of the condition which leads to the outcome. E.g. in this case "GetDrainStatus_HostNotRunning_ReturnsServiceUnavailable"

src/WebJobs.Script.WebHost/Controllers/HostController.cs
mathewc: I had previously made the comment about RRH attribute and how the drain methods should be consistent, and you've made them consistent. However, I'm wondering if they should both be marked RRH? Depends on what DataRole and other consumers are expecting.

pragnagopa:
cc @TsuyoshiUshio  as fyi

src/WebJobs.Script.WebHost/Controllers/HostController.cs
alrod: > Once we add Resume capabilities, I do wonder how this API should behave if we've went through a series of Drain/Resume calls, such that we end up with an active host that isn't draining, but one or more orphaned hosts that are draining. Should this API return a result that reflects the drain status of any/all orphaned instances?

In current code on drain call we only put the active instance in drain mode(there can be orphaned instances potentially), I assume new Resume API also will be restarting the listeners only in active instance. To be consistent I think we need to rely only on active instance for the DrainStatus API.

>I had previously made the comment about RRH attribute and how the drain methods should be consistent, and you've made them consistent. However, I'm wondering if they should both be marked RRH? Depends on what DataRole and other consumers are expecting.

IDrainModeManager is available in offline host - validated this by putting the host offline and then calling drain/drainstatus.

mathewc: I believe the plan for resume is for a new host to be started up. Thus, each time we call drain, the active host is orphaned. So you have an active host and one that is draining. If you call drain again, the process repeats, and now you have an active host and 2 draining instances, etc. Now if you call drain/status on the active instance it'll say it isn't draining, even though there are orphaned hosts that are, which from the standpoint of DataRole should prevent the instance from shutting down?

I'm bringing this all up because when resume is implemented we'll have to figure all that out. However, for the current codebase, it's not an issue - if we're in drain mode, there's only a single active (draining) host and no new hosts will be started.

src/WebJobs.Script.WebHost/Controllers/HostController.cs
alrod: IDrainModeManager is available in offline host - validated this by putting the host offline and then calling drain/drainstatus.

mathewc: I just checked in prod, and I see service resolution errors coming from admin/host/drain. I just logged https://github.com/Azure/azure-functions-host/issues/7792. If the ScriptHost isn't running, the service resolution will fail.

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SamplesEndToEndTests_Node_DrainStatus.cs
```diff
@@ -0,0 +1,83 @@
...
+using Xunit;
+
+namespace Microsoft.Azure.WebJobs.Script.Tests.EndToEnd
+{
+    [Trait(TestTraits.Category, TestTraits.EndToEnd)]
+    [Trait(TestTraits.Group, TestTraits.SamplesEndToEnd)]
+    public class SamplesEndToEndTests_Node_DrainStatus : DrainTestFixture
+    {
+        [Fact]
+        public async Task DrainStatus_RunningHost_ReturnsExpected()
```
liliankasem, line 22: Can we also add an E2E test for a non-happy case i.e. DrainStatus_HostOffline or similar

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SamplesEndToEndTests_Node_DrainStatus.cs
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    public class DrainTestFixture : EndToEndTestFixture
+    {
+        static DrainTestFixture()
+        {
+        }
+
+        public DrainTestFixture()
+            : base(Path.Combine(Environment.CurrentDirectory, @"..\..\..\..\..\sample\NodeDrain"), "samples", RpcWorkerConstants.NodeLanguageWorkerName)
```
liliankasem: Please use something like `Path.Join("..", "..", etc.`, this test wont run outside of Windows

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SamplesEndToEndTests_Node_DrainStatus.cs
alrod: `HostOffline` test is tricky as we use `TestFunctionHost` in the test and the offline behavior is different from WebHost.

I validated offline scenario manually and it works returns valid drain status as we have an active host in offline.

src/WebJobs.Script.WebHost/Controllers/HostController.cs
mathewc: You're using TryGetHostService below to defensively get the status provider, which will handle situations where this API is called when no active host exists. You should do the same for the drain manager and remove the [FromServices] param. That'll improve consistency, and will also ensure we don't get service resolution failures. That'll be the fix for #7792 as well. Wondering if you shouldn't just address that as well in this PR so these APIs are consistent? I.e. if the required services aren't available (no host running), we return the same status code in both cases (503).

mathewc (approved):

PR #7775, opened 2021-10-14
[pack] Porting patch changes from dev

yojagad:

resolves ##7314 for v4

Changes ported over:
https://github.com/Azure/azure-functions-host/pull/7346
https://github.com/Azure/azure-functions-host/pull/7468

Sample build with with patch version as 0
https://azfunc.visualstudio.com/Azure%20Functions/_build/results?buildId=39553&view=artifacts&pathAsName=false&type=publishedArtifacts

Sample build with non-zero patch version and base as 4.0.0
https://azfunc.visualstudio.com/Azure%20Functions/_build/results?buildId=39541&view=artifacts&pathAsName=false&type=publishedArtifacts

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/build/build-extensions.ps1
+++ b/build/build-extensions.ps1
@@ -1,8 +1,14 @@
 param (
-  [string]$buildNumber = "0",  
-  [string]$suffix = "",  
+  [string]$buildNumber = "0",
+  [string]$majorMinorVersion,
+  [string]$patchVersion,
+  [string]$suffix = "",
   [string]$hashesForHardlinksFile = "hashesForHardlinks.txt"
 )
+
+$extensionVersion = "$majorMinorVersion.$patchVersion"
+Write-Host "ExtensionVersion is $extensionVersion"
+Write-Host "BuildNumber is $buildNumber"
 
 $rootDir = Split-Path -Parent $PSScriptRoot
 $buildOutput = Join-Path $rootDir "buildoutput"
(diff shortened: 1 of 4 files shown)


yojagad:
>     <PreviewString Condition="'$(PreviewVersion)' != ''">-preview.$(PreviewVersion).$(BuildNumber)</PreviewString>

Since this should be pointing to the `next` version to be released, should the preview tags be removed ? This is assuming that the next version being released is just `4.0.0`

---
Refers to: build/common.props:11 in bf73f46. [](commit_id = bf73f466358106679ef86c3e6902622a263e5e92, deletion_comment = False)

brettsam:
> > ```
> > <PreviewString Condition="'$(PreviewVersion)' != ''">-preview.$(PreviewVersion).$(BuildNumber)</PreviewString>
> > ```
> 
> Since this should be pointing to the `next` version to be released, should the preview tags be removed ? This is assuming that the next version being released is just `4.0.0`
> 
> Refers to: build/common.props:11 in [bf73f46](https://github.com/Azure/azure-functions-host/commit/bf73f466358106679ef86c3e6902622a263e5e92). [](commit_id = [bf73f46](https://github.com/Azure/azure-functions-host/commit/bf73f466358106679ef86c3e6902622a263e5e92), deletion_comment = False)

I've got a PR out right now removing us from Preview -- https://github.com/Azure/azure-functions-host/pull/7770. Let's get that in first...

yojagad:
@brettsam / @pragnagopa  This is rebased and I updated the description with the updated test builds. Take a look when you have some time, please.

build/initialize-pipeline.ps1
```diff
@@ -9,4 +14,32 @@ if ($buildReason -eq "PullRequest") {
...
+  Write-Host "Setting 'PatchVersion' to $patchVersion"
+}
+
+#Update buildnumber with the same (Will be used by release pipelines)
+$customBuildNumber = "$majorVersion.$minorVersion.$patchVersion"
+if(($buildReason -eq "PullRequest") -or !($sourceBranch.ToLower().Contains("release")))
+{
+  $customBuildNumber = "$customBuildNumber-$buildNumber"
+}
+Write-Host "##vso[build.updatebuildnumber]$customBuildNumber"
```
brettsam, line 45: we shouldn't need to do any of this xml parsing -- there's a step in the build that calls the GenerateSiteExtensionVersion target in common.props, which writes out a version.txt that can then later be read to get the version.

build/common.props
```diff
@@ -5,14 +5,14 @@
     <LangVersion>latest</LangVersion>
     <MajorVersion>4</MajorVersion>
     <MinorVersion>0</MinorVersion>
-    <PatchVersion>1</PatchVersion>
+    <PatchVersion>0</PatchVersion>
```
brettsam, line 8: can you change this back to 1?

build/build-extensions.ps1
```diff
@@ -1,9 +1,15 @@
 param (
-  [string]$buildNumber = "0",  
-  [string]$suffix = "",  
+  [string]$buildNumber = "0",
+  [string]$majorMinorVersion,
+  [string]$patchVersion,
+  [string]$suffix = "",
```
brettsam, line 5: We shouldn't have to pass version details in now -- below it will call the GenerateSiteExtensionVersion target and read the output file. That way MSBuild always is in charge of the versions for us. There's some extra logic in there for previews and suffixes that is tough to reason through without MSBuild.

build/build-extensions.ps1
```diff
@@ -121,6 +127,105 @@ function CleanOutput([string] $rootPath) {
     Write-Host "  Current size: $(GetFolderSizeInMb $rootPath) Mb"
 }
 
+function CreatePatchedSiteExtension([string] $siteExtensionPath) {
+  try {
+    Write-Host "SiteExtensionPath is $siteExtensionPath"
+    $officialSiteExtensionPath = "$siteExtensionPath\$extensionVersion"
+    $baseVersion = "$majorMinorVersion.0"
```
brettsam, line 134: nuts... this logic won't work for v4 -- the versions are now "4.0.0.buildnumber" -- I did this so that every build would have a unique version number. I didn't realize it'd hit you here... 

i wonder if we need some special flag to say "this build should be patching build "4.0.0.xxxxx"?

brettsam (commented):
there's some initial comments here -- we may have to sync on behavior -- mostly around version numbers as i changed them slightly for v4.

build/build-extensions.ps1
yojagad: Is there a specific/strong reason for deviating from existing v3 versioning or am I missing something here ? I thought we decided to use `major.minor.patch` and have build number appended with a `hyphen` just like how we have in v3. The version should always be pointing to the successive version that is about to be released which gets updated on every release cut.

build/common.props
yojagad: Is `4.0.1` going to be the next version when a release is cut from here ?

build/build-extensions.ps1
pragnagopa: @brettsam  / @fabiocav  - Same question as above.  We updated Functions Runtime Versioning for releases to be able to support shipping patches on App Service without having to add full site extension. Did anything change?

build/initialize-pipeline.ps1
yojagad: This parsing is also done so that devops UI reflects the current build number as soon as initialize is run..

build/build-extensions.ps1
brettsam: Patches are still supported -- nothing changed there. The version out now is 4.0.1.xxxxx for example (we had to abandon 4.0.0.xxxx before UD0). And next release will be 4.1.0.xxxxx. I want it to be clear where the build came from when looking at the file version (i.e you don't get a bunch of 4.1.0.0 in kusto when the builds may be different).

Let me think through if we can drop that value from the site extension version (and add a dash for CI builds) -- but retain the value in the FileVersion.

build/common.props
brettsam: next release will be 4.1.0 -- but we should make that an explicit PR and not bundled in here.

build/initialize-pipeline.ps1
brettsam: Can you read version.txt? See build-siteextensions.ps1 for how it's generated. There's a decent amount of conditional logic in there for suffixes, previews, etc that really can't be checked without running MSBuild on it...

build/build-extensions.ps1
```diff
@@ -153,9 +258,21 @@ function CreateSiteExtensions() {
     Write-Host "======================================"
     Write-Host
     
+    # This needs to be determined if it's patch or not.
     $zipOutput = "$buildOutput\SiteExtension"
     New-Item -Itemtype directory -path $zipOutput -Force > $null
     ZipContent $siteExtensionPath "$zipOutput\Functions.$extensionVersion$runtimeSuffix.zip"
+
+    # Construct patch
+    if(([int]$patchVersion) -gt 0)
```
brettsam, line 267: Do we *always* do this during a patch? In the case I just had -- I never actually created a release as we abandoned it in UD0. I created a patch build and basically re-started...

brettsam: n/m -- discussed offline!

brettsam:
this was taken care of in #7814

PR #7780, opened 2021-10-15
Initiate DrainMode during application shutdown

pragnagopa:

resolves #issue_for_this_pr

Platform does not call DrainMode api - if worker assignment happens due to following:

- Code / File change
- Function app   restart
- Function app stop

Initiating DrainMode API during application shutdown will ensure call to stop listeners is made in all code paths that shutdown host - across Skus / Platforms

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
+++ b/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
@@ -763,16 +763,17 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
             _applicationLifetime.ApplicationStopping.Register(() =>
             {
                 Interlocked.Exchange(ref _applicationStopping, 1);
-                if (_environment.DrainOnApplicationStoppingEnabled())
+                var drainModeManager = _host?.Services.GetService<IDrainModeManager>();
+                if (drainModeManager != null)
                 {
-                    var drainModeManager = _host?.Services.GetService<IDrainModeManager>();
-                    if (drainModeManager != null)
-                    {
-                        _logger.LogDebug("Application Stopping: initiate drain mode");
-                        drainModeManager.EnableDrainModeAsync(CancellationToken.None);
-                        // Workaround until https://github.com/Azure/azure-functions-host/issues/7188 is addressed.
-                        Thread.Sleep(TimeSpan.FromMinutes(10));
-                    }
+                    _logger.LogDebug("Application Stopping: Initiate drain mode.");
+                    drainModeManager.EnableDrainModeAsync(CancellationToken.None);
+                }
+                if (_environment.DrainOnApplicationStoppingEnabled() && drainModeManager != null)
+                {
+                    _logger.LogDebug("Application Stopping. Waiting for 10mins to drain invocations.");
+                    // Workaround until https://github.com/Azure/azure-functions-host/issues/7188 is addressed.
+                    Thread.Sleep(TimeSpan.FromMinutes(10));
                 }
             });
 


pragnagopa:
cc @balag0 @divyagandhisethi @pgombar 

@mathewc / @soninaren  / @fabiocav  - Do you see any issues with "Initiating Drain Mode" as part of application shutdown? This will ensure listeners are stopped while functions host is shutting down across all skus consistently. How much grace period should be allowed to drain invocations is still owned by the platform.

mathewc:
Don't we already call stop on the host on application shutdown? If so that should have the same effect - stopping the listeners. Related to https://github.com/Azure/azure-webjobs-sdk/issues/2776.

pragnagopa:
cc @AnatoliB @alrod

pragnagopa:
> Don't we already call stop on the host on application shutdown

From my understanding - an explicit call is made to stop listeners from the platform to initiate DrainMode - this does not happen as part of the Stop.

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -763,16 +763,17 @@ private void RegisterApplicationLifetimeEvents()
...
-                        // Workaround until https://github.com/Azure/azure-functions-host/issues/7188 is addressed.
-                        Thread.Sleep(TimeSpan.FromMinutes(10));
-                    }
+                    _logger.LogDebug("Application Stopping: Initiate drain mode.");
+                    drainModeManager.EnableDrainModeAsync(CancellationToken.None);
+                }
+                if (_environment.DrainOnApplicationStoppingEnabled() && drainModeManager != null)
+                {
+                    _logger.LogDebug("Application Stopping. Waiting for 10mins to drain invocations.");
+                    // Workaround until https://github.com/Azure/azure-functions-host/issues/7188 is addressed.
```
liliankasem, line 775: Looks like #7188 is closed, do we still need the workaround here?

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -763,16 +763,17 @@ private void RegisterApplicationLifetimeEvents()
...
-                    var drainModeManager = _host?.Services.GetService<IDrainModeManager>();
-                    if (drainModeManager != null)
-                    {
-                        _logger.LogDebug("Application Stopping: initiate drain mode");
-                        drainModeManager.EnableDrainModeAsync(CancellationToken.None);
-                        // Workaround until https://github.com/Azure/azure-functions-host/issues/7188 is addressed.
-                        Thread.Sleep(TimeSpan.FromMinutes(10));
-                    }
+                    _logger.LogDebug("Application Stopping: Initiate drain mode.");
+                    drainModeManager.EnableDrainModeAsync(CancellationToken.None);
```
liliankasem, line 770: So it looks like we init drain even if `DrainOnApplicationStoppingEnabled` is not enabled, so do we need the check below? 

Just trying to understand why we're doing this 😄

pragnagopa: Trying to answer why shouldn't we stop listeners as part of application shutdown :) - this will avoid calling DrainMode api from the platform as part of worker unassignment flow

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
pragnagopa: Yea - we need to leave this in until Kubernetes - code paths can advantage of the changes made in #7188

fabiocav:
I have the same question as @mathewc above. A host disposal and shutdown will be initiated as part of the normal shutdown flow. Have you found evidence to the contrary?

pragnagopa:
@divyagandhisethi - can you help answer 

> A host disposal and shutdown will be initiated as part of the normal shutdown flow.

Are there any instances where shutting down the host does not stop listeners?

divyagandhisethi:
> @divyagandhisethi - can you help answer
> 
> > A host disposal and shutdown will be initiated as part of the normal shutdown flow.
> 
> Are there any instances where shutting down the host does not stop listeners?

If we are guaranteed that shutting down the host will stop the listeners, then the only solution is to call drain mode from platform side.  We had initially made this change to handle zombie containers, which never receive the shutdown call after deletion and hence don't stop the listeners as well.

fabiocav:
> If we are guaranteed that shutting down the host will stop the listeners, then the only solution is to call drain mode from platform side. We had initially made this change to handle zombie containers, which never receive the shutdown call after deletion and hence don't stop the listeners as well.

I'm not following this comment. Listeners will stop when the host is shutdown, so I don't believe there's anything to be solved by draining at that point.

If there's an issue with the platform failing to shutdown the host, we need to address that problem, instead of making this change.

pragnagopa:
> If there's an issue with the platform failing to shutdown the host, we need to address that problem, instead of making this change.

This right. There is an active issue on Atlas that does not delete containers as expected which in turn does not shutdown the host and invocations continue to happen. We had multiple CRIs for this. We are actively following up with Atlas to fix root cause. 

As long as listeners are shutdown when host shutdown, we do not need this PR. Closing.

PR #7789, opened 2021-10-19
updating AppContext switches

brettsam:


src/WebJobs.Script.WebHost/Program.cs
```diff
@@ -24,10 +24,6 @@ public class Program
     {
         public static void Main(string[] args)
         {
-            // Protect users from https://github.com/dotnet/runtime/pull/59716.
-            // Can be removed after running on .NET 6.0 GA
-            AppContext.SetSwitch("Microsoft.Extensions.Configuration.BindSingleElementsToArray", false);
```
brettsam, line 29: Makes more sense to put this in the runtimeconfig template. That ensures it is set as early as possible (even before any startup hooks)

mathewc: I get that you're moving this one, but what's the new one for?

brettsam: The issue linked there -- https://github.com/dotnet/runtime/issues/56052 and it's fix https://github.com/dotnet/runtime/pull/56873 -- was brought up by us. Originally it was found with some of the new Azure packages. Our host would use one version and add an EventSource, while the customer's extension would try to register that same EventSource and it'd throw. This is happening in separate AssemblyLoadContexts, but the EventSource infrastructure doesn't know that.

This was mitigated at the Azure.Core level but we bumped into it in some v4 testing in Core Tools with App Insights.

liliankasem (dismissed):

brettsam:
FYI -- I the SecretManager tests seem pretty flaky. I've added new logging so we can track the logs in devops when they fail... and of course they passed.

I want to keep that change in to help us catch them next time they fail, though... I can keep the commits separate when I merge.

liliankasem (approved):

PR #7793, opened 2021-10-21
Reverting removing built-in Application Insights support and references

gzuber:
### Issue describing the changes in this PR

resolves #issue_for_this_pr

### Pull request checklist

* [ ] My changes **do not** require documentation changes
    * [ ] Otherwise: Documentation issue linked to PR
* [ ] My changes **should not** be added to the release notes for the next release
    * [ ] Otherwise: I've added my notes to `release_notes.md`
* [ ] My changes **do not** need to be backported to a previous version
    * [ ] Otherwise: Backport tracked by issue/PR #issue_or_pr
* [ ] I have added all required tests (Unit tests, E2E tests)

### Additional information

Additional PR information

PR #7799, opened 2021-10-25
ignore and warn if AppInsights extension detected

brettsam:
As we are rolling back the preview support for the Application Insights extension (#7793) , we need to explicitly ignore it if we detect it. This prevents preview users from being broken with the next release.

### DIFF
--- a/src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
+++ b/src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
@@ -26,6 +26,8 @@ namespace Microsoft.Azure.WebJobs.Script.DependencyInjection
     /// </summary>
     public class ScriptStartupTypeLocator : IWebJobsStartupTypeLocator
     {
+        private const string ApplicationInsightsStartupType = "Microsoft.Azure.WebJobs.Extensions.ApplicationInsights.ApplicationInsightsWebJobsStartup, Microsoft.Azure.WebJobs.Extensions.ApplicationInsights, Version=1.0.0.0, Culture=neutral, PublicKeyToken=9475d07f10cb09df";
+
         private readonly string _rootScriptPath;
         private readonly ILogger _logger;
         private readonly IExtensionBundleManager _extensionBundleManager;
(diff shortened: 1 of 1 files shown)


src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
```diff
@@ -161,6 +163,13 @@ public async Task<IEnumerable<Type>> GetExtensionsStartupTypesAsync()
 
             foreach (var extensionItem in extensionItems)
             {
+                // We need to explicitly ignore ApplicationInsights extension
+                if (extensionItem.TypeName.Equals(ApplicationInsightsStartupType, StringComparison.Ordinal))
```
kshyju, line 167: Are we also supposed to update the extension bundle to remove app insights ?

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
```diff
@@ -161,6 +163,13 @@ public async Task<IEnumerable<Type>> GetExtensionsStartupTypesAsync()
 
             foreach (var extensionItem in extensionItems)
             {
+                // We need to explicitly ignore ApplicationInsights extension
```
kshyju, line 166: This is to handle those users who are migrating from preview5 to GA. rite? Should we also add a `// TO DO` to remove this after a reasonable time?

kshyju (approved):

gzuber (approved):

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
brettsam: We will, but it's not critical. It will be ignored by both v4.x and dev right now. We'll pull it out in parallel.

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
brettsam: We'll have to keep our eyes on it -- if we continue seeing these warnings in our logs, we may just leave this in forever.

Issue #4876, opened 2019-09-03
Evaluate breaking change -- AllowSynchronousIO now false in ASP.NET Core 3


brettsam:
https://github.com/aspnet/AspNetCore/issues/7644. 

Internally I had to change the placeholder warmup function to read asynchronously, but I suspect customers will get bit by this when updating to v3.

brettsam:
I actually reversed course on this a bit -- I left our ReadAsync calls b/c they're generally better. But I had to go through and set `AllowSynchronousIO = true` in a bunch of places to workaround json deserialization in the Http extension: https://github.com/Azure/azure-webjobs-sdk-extensions/blob/dev/src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs#L469.

Not sure if there's an efficient workaround for this without moving JSON.NET forward there. Will need investigation in the future.

brettsam:
https://github.com/Azure/azure-functions-host/pull/5064 will address this by allowing async IO when running in v2 compat mode.

PR #7801, opened 2021-10-25
Implement resume API for drain mode

### LINKED ISSUE #7774: [Resume API] Implement resume API endpoint in the functions host
Based on the design document for Drain Mode Resume, implement the resume API within the functions host.

Epic: #7772 
[Design document](https://dev.azure.com/msazure/One/_wiki/wikis/Antares%20Engineering%20Hub/189846/ResumeModeAPI)

liliankasem:
PR for the resume API endpoint as part of the wider drain mode feature.

resolves #7774

This PR is currently being tested with a private site extension in a private stamp

### Pull request checklist

    * [Resume API design doc](https://dev.azure.com/msazure/One/_git/AAPT-Antares-Docs?path=/TeamDocs/FunctionTeamDocs/Design/DrainMode/ResumeModeAPI.md&version=GBmaster&line=1&_a=preview)


test/WebJobs.Script.Tests.Integration/Controllers/ControllerScenarioTestFixture.cs
```diff
@@ -76,10 +78,8 @@ public virtual async Task InitializeAsync()
                 .ConfigureAppConfiguration(c => c.AddEnvironmentVariables());
 
             ConfigureWebHostBuilder(webHostBuilder);
-            
-            // TODO: https://github.com/Azure/azure-functions-host/issues/4876
```
liliankasem, line 80: Removing this comment as #4876 closed

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -155,6 +155,40 @@ public IActionResult DrainStatus([FromServices] IScriptHostManager scriptHostMan
...
+                }
+
+                if (!drainModeManager.IsDrainModeEnabled)
+                {
+                    _logger.LogWarning("Resume request received when not in drain mode");
+                    return Ok(new { hostStatus = scriptHostManager.State.ToString() });
+                }
+
+                _logger.LogDebug("Host is running and in drain mode, starting a new host");
+                await scriptHostManager.RestartHostAsync();
```
kshyju: Do we know how long this method call takes? I wonder we will run into a situation where the request from the caller(scale controller?) times out because this `RestartHostAsync` method takes a longer time to finish.

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -155,6 +155,40 @@ public IActionResult DrainStatus([FromServices] IScriptHostManager scriptHostMan
...
+                if (scriptHostManager.State != ScriptHostState.Running)
+                {
+                    _logger.LogDebug("Host is not running, cannot resume");
+                    return StatusCode(StatusCodes.Status409Conflict);
+                }
+
+                if (!drainModeManager.IsDrainModeEnabled)
+                {
+                    _logger.LogWarning("Resume request received when not in drain mode");
+                    return Ok(new { hostStatus = scriptHostManager.State.ToString() });
```
kshyju: Question: Is 200 OK the appropriate response code in this case? Should it be 405?

liliankasem: Yeah we discussed this as part of the design review. If the host is running and is not in drain mode, we want to return 200 as that signals to Antares that they can use this host and it's ready for them. Anything other than a 200 could potentially lead to this worker being killed on their side and that would be a waste when it's running as expected

src/WebJobs.Script.WebHost/Controllers/HostController.cs
liliankasem: This will vary based on SKU, language etc. but it should be faster than a full host start. Maybe @soninaren has an approximate number for this?

There is definitely potential for them to timeout and we talked about them retrying the resume call once that happens. If you have suggestions on how to better handle this I'd love to hear from you :D

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -155,6 +155,40 @@ public IActionResult DrainStatus([FromServices] IScriptHostManager scriptHostMan
...
+        [Authorize(Policy = PolicyNames.AdminAuthLevelOrInternal)]
+        public async Task<IActionResult> Resume([FromServices] IScriptHostManager scriptHostManager)
+        {
+            _logger.LogDebug($"Received request to resume a draining host.");
+
+            if (Utility.TryGetHostService(scriptHostManager, out IDrainModeManager drainModeManager))
+            {
+                _logger.LogDebug($"Host status: {scriptHostManager.State.ToString()}. Drain mode enabled: {drainModeManager.IsDrainModeEnabled}");
+
+                if (scriptHostManager.State != ScriptHostState.Running)
```
liliankasem: Move this outside the if statement on line 165

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -155,6 +155,40 @@ public IActionResult DrainStatus([FromServices] IScriptHostManager scriptHostMan
...
+                {
+                    _logger.LogDebug("Host is not running, cannot resume");
```
liliankasem: Add more context in the log. What do we mean by "host is not running"

409 = conflict = waiting for the host to be in the right state to resume. We can only except the request in a certain state:

- Drain mode is enabled
- Host is running

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -155,6 +155,40 @@ public IActionResult DrainStatus([FromServices] IScriptHostManager scriptHostMan
...
         }
 
+        [HttpPost]
+        [Route("admin/host/resume")]
+        [Authorize(Policy = PolicyNames.AdminAuthLevelOrInternal)]
+        public async Task<IActionResult> Resume([FromServices] IScriptHostManager scriptHostManager)
+        {
+            _logger.LogDebug($"Received request to resume a draining host - host status: {scriptHostManager?.State.ToString()}");
+
+            if (scriptHostManager?.State != ScriptHostState.Running)
```
brettsam: All of these state checks make me nervous about race conditions. For example, drain and resume called really close together so the states haven't been fully updated yet.

To help debug this, it may be a good idea to log when the State changes in the `WebJobsScriptHostService` -- look at what we do for `ActiveHost` today for an example.

sample/NodeResume/HttpTrigger/function.json
```diff
@@ -0,0 +1,23 @@
...
+        "post"
+      ]
+    },
+    {
+      "type": "http",
+      "direction": "out",
+      "name": "res"
+    }
+  ],
+  "retry": {
```
fabiocav: Is there a reason for retries configuration here?

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -113,7 +113,7 @@ public IActionResult Drain([FromServices] IScriptHostManager scriptHostManager)
 
             if (Utility.TryGetHostService(scriptHostManager, out IDrainModeManager drainModeManager))
             {
-                // Stop call to some listeners get stuck, Not waiting for the stop call to complete
+                // Stop call to some listeners get stuck, not waiting for the stop call to complete
                 drainModeManager.EnableDrainModeAsync(CancellationToken.None).ConfigureAwait(false);
```
fabiocav: Are failures being correctly handled and logged in the downstream API calls here? If unsure, it would be good to make sure we have a continuation handling any failures.

src/WebJobs.Script.WebHost/Controllers/HostController.cs
fabiocav: Indeed. Let's make sure we have enough logging here to trace through state changes. For some of the areas we don't expect to handle concurrent calls, we may also want to consider synchronizing access.

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -155,6 +155,40 @@ public IActionResult DrainStatus([FromServices] IScriptHostManager scriptHostMan
...
+        {
+            _logger.LogDebug($"Received request to resume a draining host - host status: {scriptHostManager?.State.ToString()}");
+
+            if (scriptHostManager?.State != ScriptHostState.Running)
+            {
+                _logger.LogWarning("The host is not in a state where we can resume. The host's state needs to be 'Running' and drain mode must be enabled for resume to work.");
+                return StatusCode(StatusCodes.Status409Conflict);
+            }
+
+            if (Utility.TryGetHostService(scriptHostManager, out IDrainModeManager drainModeManager))
```
fabiocav: Let's reverse this logic and log if unable to resolve the host service here?

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/DrainModeResumeEndToEndTests.cs
```diff
@@ -0,0 +1,113 @@
...
+using System.IO;
+using System.Net;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Microsoft.Azure.WebJobs.Script.Tests.EndToEnd
+{
+    [Trait(TestTraits.Category, TestTraits.EndToEnd)]
+    [Trait(TestTraits.Group, TestTraits.DrainModeEndToEnd)]
```
fabiocav, line 16: Thanks for scoping the tests!

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -155,6 +155,40 @@ public IActionResult DrainStatus([FromServices] IScriptHostManager scriptHostMan
...
+        [HttpPost]
+        [Route("admin/host/resume")]
+        [Authorize(Policy = PolicyNames.AdminAuthLevelOrInternal)]
+        public async Task<IActionResult> Resume([FromServices] IScriptHostManager scriptHostManager)
+        {
+            _logger.LogDebug($"Received request to resume a draining host - host status: {scriptHostManager?.State.ToString()}");
+
+            if (scriptHostManager?.State != ScriptHostState.Running)
+            {
+                _logger.LogWarning("The host is not in a state where we can resume. The host's state needs to be 'Running' and drain mode must be enabled for resume to work.");
```
mathewc: I think you should combine the check here with the one below - basically try to access IDrainMode manager and check drain status / state all in one go and return the error if needed in a single check? This warning seems out of sync with the logic below, where you return 200 if we're not in drain mode.

src/WebJobs.Script.WebHost/Controllers/HostController.cs
mathewc: If we need to return a body that the caller will be parsing, we should define and use a concrete Type, here and below

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/DrainModeResumeEndToEndTests.cs
```diff
@@ -0,0 +1,113 @@
...
+        {
+            // Validate the drain state is "Disabled" initially
+            var response = await SamplesTestHelpers.InvokeDrainStatus(this);
+            var responseString = response.Content.ReadAsStringAsync().Result;
+            var status = JsonConvert.DeserializeObject<DrainModeStatus>(responseString);
+
+            Assert.Equal(status.State, DrainModeState.Disabled);
+
+            ManualResetEvent resetEvent = new ManualResetEvent(false);
+            _ = Task.Run(async () =>
```
mathewc: Why do you need to use Task.Run here? I think it'd be cleaner if you removed it. The operations will run in parallel with your http trigger request below regardless.

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/DrainModeResumeEndToEndTests.cs
```diff
@@ -0,0 +1,113 @@
...
+                    var responseString = response.Content.ReadAsStringAsync().Result;
+                    var status = JsonConvert.DeserializeObject<DrainModeStatus>(responseString);
+
+                    return status.State == DrainModeState.Disabled;
+                }, 20000);
+
+                resetEvent.Set();
+            });
+
+            response = await SamplesTestHelpers.InvokeHttpTrigger(this, "HttpTrigger");
```
mathewc, line 61: Is the intent here to have this function run before drain mode is initiated? Might make the call before we initiate then, and also verify it runs after resume as well.

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/DrainModeResumeEndToEndTests.cs
```diff
@@ -0,0 +1,113 @@
...
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+            bool result = resetEvent.WaitOne(30000);
+            Assert.True(result);
```
mathewc: Might be good to call host/status before and after to compare InstanceId which will be different, verifying a new host is running?

liliankasem: Good idea!

sample/NodeResume/HttpTrigger/function.json
liliankasem: Ah nope, copy/paste. I can remove it :)

src/WebJobs.Script.WebHost/Controllers/HostController.cs
liliankasem: Will do :D

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/DrainModeResumeEndToEndTests.cs
liliankasem: Yup, I think you make a good point - I want to verify it works before and after. I'll update the test!

src/WebJobs.Script.WebHost/Controllers/HostController.cs
liliankasem: Good point, I'll do that!

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/DrainModeResumeEndToEndTests.cs
liliankasem: Hmm something funky was happening without this (it's been a minute so I don't remember the details), I'll look into doing it without Task.Run again and see if it works/ what the issue is

src/WebJobs.Script.WebHost/Controllers/HostController.cs
liliankasem: Great point about the warning message being out of sync, I can update that. 

RE having a single check: I think we want different messages for the host not running (409) vs the drainModeManager not being available (503) vs drain mode not enabled but host is running (200)?

We can combine trying to access IDrainMode and host state and just return 409 here? But I'd like to keep the drain mode state check separate as the return for that depends on host state:

drain mode enabled + running host = restart, return 200
drain mode disabled + running host = return 200

Updated the code to reflect this thought, let me know what you think?

src/WebJobs.Script.WebHost/Controllers/HostController.cs
liliankasem: Thoughts on just returning the existing HostStatus type but just filling in state?

```csharp
var status = new HostStatus { State = scriptHostManager.State.ToString() };
return Ok(status);
```

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/DrainModeResumeEndToEndTests.cs
liliankasem: Wasn't able to repro the issue I was having last time so I'm removing Task.Run as suggested, thank you!

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/DrainModeResumeEndToEndTests.cs
liliankasem: So the InstanceId == "" when I make the `/status` request in the E2E test and when I run the host locally in debug (both before and after calling drain/resume)

But I do see there is an instance Id in the logs:

`Host lock lease acquired by instance ID '000000000000000000000000CD51D200'`

And it is always `000000000000000000000000CD51D200`. This made me a little nervous so I checked the logs of a private site extension I deployed

```
Before /drain + /resume
Starting Host (HostId=blah, InstanceId=4ff35ca3-0270-48a2-a09a-3ad11887d12b,

After /resume
Starting Host (HostId=blah, InstanceId=3fd5d337-e970-4427-9655-c3a4028d7bef,
```

So I think it is working as expected, but I'm looking in the wrong place for the instance Id? Any suggestions?

src/WebJobs.Script.WebHost/Controllers/HostController.cs
liliankasem: I added trace logs for the state change - thoughts? https://github.com/Azure/azure-functions-host/pull/7801/commits/dab7a5862eea53434f9ddf51a06d3cc49ff6db05

RE: synchronizing access - I'd like to bring this topic to our design meeting Thursday, would love to brainstorm with the team on next steps

src/WebJobs.Script.WebHost/Controllers/HostController.cs
mathewc: Recommend defining a new ResumeStatus Type with a single HostState member for now

src/WebJobs.Script.WebHost/Controllers/HostController.cs
mathewc: Updates look good to me

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/DrainModeResumeEndToEndTests.cs
mathewc: It's the latter instance ID I was talking about, which I incorrectly assumed was what was returned via HostStatus.InstanceId. The goal is for the test to ensure that a new host instance is running. I think you can do that using the [HostInstanceId](https://github.com/Azure/azure-functions-host/blob/cbf9ad1b107c52531975f622d7f50f63ab31e592/test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/EndToEndTestFixture.cs#L75) property in the base test fixture.

liliankasem: I'll give that a shot, thanks for the pointer :)

src/WebJobs.Script.WebHost/Controllers/HostController.cs
liliankasem: I think that would change the data contract we have with the antares team from `{ "state": "running" }` to something like `{ "hostStatus": { "state": "running" } }` 

(just worth noting to sync with the other team + update docs)

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/DrainModeResumeEndToEndTests.cs
liliankasem: That worked :D

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -156,6 +156,32 @@ public IActionResult DrainStatus([FromServices] IScriptHostManager scriptHostMan
             }
         }
 
+        [HttpPost]
+        [Route("admin/host/resume")]
+        [Authorize(Policy = PolicyNames.AdminAuthLevelOrInternal)]
+        public async Task<IActionResult> Resume([FromServices] IScriptHostManager scriptHostManager)
+        {
+            _logger.LogDebug($"Received request to resume a draining host - host status: {scriptHostManager?.State.ToString()}");
+
```
mathewc: Because you're accessing State multiple times before the restart and it can change between these accesses, to avoid confusion I think you should capture it once in a local and reuse that state ` var currentState = scriptHostManager.State;`. You don't need the null checks on scriptHostManager either - it's an input parameter from DI so won't be null (see other examples in this file).

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -156,6 +156,32 @@ public IActionResult DrainStatus([FromServices] IScriptHostManager scriptHostMan
...
+            if (!Utility.TryGetHostService(scriptHostManager, out IDrainModeManager drainModeManager)
+                || scriptHostManager?.State != ScriptHostState.Running)
```
mathewc: Recommend checking state before accessing drain mode manager - can short circuit that check in many cases

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -156,6 +156,32 @@ public IActionResult DrainStatus([FromServices] IScriptHostManager scriptHostMan
...
+            {
+                _logger.LogWarning("The host is not in a state where we can resume.");
```
mathewc: Thinking this should be Debug rather than Warning, otherwise this may generate noise the user can't control and they may question?

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -156,6 +156,32 @@ public IActionResult DrainStatus([FromServices] IScriptHostManager scriptHostMan
...
+
+            _logger.LogDebug($"Drain mode enabled: {drainModeManager.IsDrainModeEnabled}");
+
+            if (drainModeManager.IsDrainModeEnabled)
+            {
+                _logger.LogDebug("Starting a new host");
+                await scriptHostManager.RestartHostAsync();
+            }
+
+            var status = new ResumeStatus { HostStatus = new HostStatus { State = scriptHostManager.State.ToString() } };
```
mathewc: I don't think you should include the mostly empty HostStatus in the response - just the ScriptHostState

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -130,7 +131,19 @@ private IHost ActiveHost
...
+        {
+            get
+            {
+                return _state;
+            }
+
+            private set
+            {
+                _state = value;
+                _logger?.HostStateChanged(_state);
```
mathewc: don't need the null check on _logger - it's set in ctor

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -236,15 +249,15 @@ private async Task UnsynchronizedStartHostAsync(ScriptHostStartupOperation activ
...
                 currentCancellationToken.ThrowIfCancellationRequested();
 
-                // if we were in an error state retain that,
+                // if we were in an error State retain that,
                 // otherwise move to default
-                if (State != ScriptHostState.Error)
+                if (_state != ScriptHostState.Error)
                 {
-                    State = ScriptHostState.Default;
+                    _state = ScriptHostState.Default;
```
mathewc: Why are you bypassing the property setter with these changes - you'll be bypassing the whole point of your new log statement? You should be using the property throughout.

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
mathewc: Also, recommend only writing the "state changed" log if it actually changed. Compare new value to existing first.

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SamplesTestHelpers.cs
```diff
@@ -50,7 +51,7 @@ public static async Task<HttpResponseMessage> InvokeDrain(EndToEndTestFixture fi
         {
             string masterKey = await fixture.Host.GetMasterKeyAsync();
             string uri = $"admin/host/drain?code={masterKey}";
-            HttpRequestMessage request = new HttpRequestMessage(HttpMethod.Get, uri);
```
mathewc, line 53: Strange that Drain supports both GET/POST, but Resume only supports POST. Guess this was an unnecessary addition when Drain was added? If there was no reason why GET is supported (looks like infra only uses POST), I guess it's fine that Resume does the right thing and only supports POST, though it is a discrepancy

liliankasem: Yeah ideally we want to remove GET support from drain, I spoke to the Atlas and Antares teams and they're both using POST so I think it is safe to get rid of it, but first I want to run a query to see if any GET requests are coming in at all from anywhere. Will address in a new PR.

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
liliankasem: Woops! Accident, will fix!!

src/WebJobs.Script.WebHost/Controllers/HostController.cs
liliankasem: Ah I must have misunderstood your comment about using  a `single HostState member`, ScriptHostState definitely makes more sense to me too. Will update

src/WebJobs.Script.WebHost/Controllers/HostController.cs
liliankasem: I think this is addressed, let me know what you think!

src/WebJobs.Script.WebHost/Controllers/HostController.cs
liliankasem: Update, we're going with `{ "state": "Running" }`

src/WebJobs.Script.WebHost/Controllers/HostController.cs
liliankasem: @fabiocav Added semaphore to address this, thoughts?

src/WebJobs.Script.WebHost/Controllers/HostController.cs
liliankasem: @fabiocav Added a continuation to handle exceptions, does this address your concerns?

sample/NodeResume/HttpTrigger/index.js
```diff
@@ -0,0 +1,14 @@
+﻿function delay(ms) {
```
nzthiago, line 1: Should we also test with an in-proc .NET Function and use/have an example of using the Cancellation Token? Not sure when the token would get raised to the function while the host is in drain mode, would be great to know so I can document. And then what happens when resume is called - is there a possibility where the Cancellation Token is raised AND resume is called?

liliankasem: I think if its drain related, I'd rather leave that to another PR. But for resume, maybe but I'm not sure? I don't think cancellation tokens play into resume - tagging @soninaren for his input

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SamplesTestHelpers.cs
liliankasem: New PR to deal with some drain mode cleanup, resolving this comment and we can discuss here if there are issues!

https://github.com/Azure/azure-functions-host/pull/7876

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -108,20 +109,30 @@ public async Task<IActionResult> GetHostStatus([FromServices] IScriptHostManager
...
+            await _drainModeSemaphore.WaitAsync();
+
+            // Stop call to some listeners gets stuck, not waiting for the stop call to complete
+            _ = drainModeManager.EnableDrainModeAsync(CancellationToken.None)
+                                .ContinueWith(
+                                    antecedent =>
+                                    {
+                                        if (antecedent.Status == TaskStatus.Faulted)
+                                        {
+                                            _logger.LogDebug($"Something went wrong enabling drain mode: {antecedent.Exception.GetBaseException().Message}");
```
brettsam: I think you should use `LogError` here (one of the overloads lets you pass in an exception directly, which will go to the Details column in Kusto).

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -156,6 +167,43 @@ public IActionResult DrainStatus([FromServices] IScriptHostManager scriptHostMan
...
+            if (currentState != ScriptHostState.Running
+                || !Utility.TryGetHostService(scriptHostManager, out IDrainModeManager drainModeManager))
+            {
+                _logger.LogDebug("The host is not in a state where we can resume.");
+                return StatusCode(StatusCodes.Status409Conflict);
+            }
+
+            try
+            {
+                await _drainModeSemaphore.WaitAsync();
```
brettsam, line 177: Should the state check be moved inside of this semaphore?

src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
```diff
@@ -176,6 +176,18 @@ public static class ScriptHostServiceLoggerExtension
                 new EventId(528, nameof(ExecutedHttpRequest)),
                 Properties.Resources.ExecutedHttpRequest);
 
+        private static readonly Action<ILogger, string, Exception> _hostStateChanged =
+            LoggerMessage.Define<string>(
+                LogLevel.Trace,
```
brettsam: Make this `Debug` -- logs with `Trace` don't show up in Kusto by default.

sample/NodeResume/HttpTrigger/index.js
nzthiago: I mean is there a possibility of a cancellation token being propagated to a running function because Drain Mode was called, and then Resume is called while the function code is dealing with the cancellation token? Likely an edge scenario, just wondering if there's any concerns or testing to do for it.

src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
```diff
@@ -176,6 +176,18 @@ public static class ScriptHostServiceLoggerExtension
...
+                new EventId(529, nameof(HostStateChanged)),
+                "Host state changed: {state}.");
```
brettsam: To make this super-explicit, let's pass in both the previous and new states and say `Host state changed from {previousState} to {newState}.`

src/WebJobs.Script.WebHost/Controllers/HostController.cs
liliankasem: Ah whoops, was supposed to do that! Thanks for pointing out :D

sample/NodeResume/HttpTrigger/index.js
mathewc: When drain is initiated for a host the cancellation tokens passed into any in progress invocations will be cancelled. When resume is called later, we do a restart - the existing (draining) host is "orphaned" and a new host instance is started. The orphaned (draining) instance continues to finish up in the background - once it's listeners are completely stopped and all outstanding invocations are complete, it's disposed. Thus these are separate host instances, so the answer to Thiago's initial question is no - once cancellation for a particular function invocation is triggered, that host is stopping/draining/orphaned and won't be used anymore. I.e. once a JobHost/ScriptHost transitions to draining/orphaned it stays in that state. I.e. the host for a cancelled/cancelling invocation doesn't state transition back to running.

liliankasem: @mathewc Thanks for confirming!

@nzthiago I can add a new test using cancellation in a new PR for drain mode. I tested something like this and it works as expected, this might help with your drain mode documentation?

```csharp
public static async Task<IActionResult> Run(HttpRequest req, ILogger log, CancellationToken token)
{
    log.LogInformation("C# HTTP trigger cancellation function processing a request.");

    try
    {
        await Task.Delay(5000, token);
        return new OkResult();

    }
    catch (OperationCanceledException)
    {
        log.LogInformation("Function invocation cancelled.");
        return new NotFoundResult();
    }
    catch (Exception e)
    {
        return new StatusCodeResult(500);
    }
}
```

mathewc (dismissed):

sample/NodeResume/HttpTrigger/index.js
nzthiago: @mathewc thank you, makes sense! I forgot about that recycle of JobHost 👍. @liliankasem yes that'd be great to have a recommended way to handle the token in a separate PR as part of a test for drain mode, and I can use it for the documentation update. Appreciate it

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -42,6 +42,7 @@ public class HostController : Controller
         private readonly IScriptHostManager _scriptHostManager;
         private readonly IFunctionsSyncManager _functionsSyncManager;
         private readonly HostPerformanceManager _performanceManager;
+        private readonly SemaphoreSlim _drainModeSemaphore = new SemaphoreSlim(1, 1);
```
fabiocav: This can't be class scoped, otherwise the locking will be specific to this instance and will not have the desired effect across API calls. In other words, this should be static.

fabiocav (changes requested):

src/WebJobs.Script.WebHost/Controllers/HostController.cs
liliankasem: Good catch, fixed!

mathewc (dismissed):

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -108,20 +109,30 @@ public async Task<IActionResult> GetHostStatus([FromServices] IScriptHostManager
...
+            await _drainModeSemaphore.WaitAsync();
+
+            // Stop call to some listeners gets stuck, not waiting for the stop call to complete
+            _ = drainModeManager.EnableDrainModeAsync(CancellationToken.None)
+                                .ContinueWith(
+                                    antecedent =>
+                                    {
+                                        if (antecedent.Status == TaskStatus.Faulted)
+                                        {
+                                            _logger.LogError("Something went wrong enabling drain mode", antecedent.Exception);
```
brettsam: This is subtle -- but I think the exception needs to go first here? This works b/c there's an overload that it'll match, but it won't recognize that second param as an exception:

```
// You want this one...
LogError(ILogger, Exception, String, Object[])

// But what you have will resolve to this one...
LogError(ILogger, String, Object[])
```

brettsam (commented):
Just one minor comment!

src/WebJobs.Script.WebHost/Controllers/HostController.cs
liliankasem: Thank you!!

fabiocav (approved):

PR #7814, opened 2021-10-29
[pack] updating devops versioning

brettsam:


azure-pipelines.yml
```diff
@@ -268,6 +262,16 @@ jobs:
       contents: '**\CodeSignSummary-*.md'
   - publish: $(Build.ArtifactStagingDirectory)\SiteExtension
     artifact: SiteExtension
+  - pwsh: |
+      if ((test-path $(Build.ArtifactStagingDirectory)\ZippedPatchSiteExtension))
+      {
+        Write-Host "Patched site extension detected."
+        Write-Host "##vso[task.setvariable variable=isPatchVersion;isOutput=true]true"
```
yojagad, line 269: >"##vso[task.setvariable variable=isPatchVersion;isOutput=true]true" [](http://example.com/codeflow?start=19&length=67)

Thanks for this additional condition. Can we set this in `build-extensions.ps1` in line `204`?

brettsam: `build-extensions.ps1` is meant to have no knowledge of where it's running -- (it's something you run locally as well). So I don't want to start mixing devops behavior in that file if we don't have to.

yojagad (approved):

PR #7842, opened 2021-11-08
Expose if function app is editable in /admin/host/status API

### LINKED ISSUE #7841: Expose IsFunctionAppEditable in the /admin/host/status API

Instead of relying on various app settings to determine if a function app can be edited in the Portal or not, instead leverage the existing flag `IsFileSystemReadOnly` from `ScriptApplicationHostOptions`. Do this by exposing it as another property in the /admin/host/status API.

pgombar:

resolves #7841

### Pull request checklist

### Additional information

Additional PR information


pgombar:
@VpOfEngineering testing these changes now. Will update here once that's done.

src/WebJobs.Script.WebHost/Models/HostStatus.cs
```diff
@@ -63,6 +63,12 @@ public class HostStatus
         [JsonProperty(PropertyName = "processUptime", DefaultValueHandling = DefaultValueHandling.Ignore)]
         public long ProcessUptime { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether the filesystem is read-only
+        /// </summary>
+        [JsonProperty(PropertyName = "isFunctionAppEditable", DefaultValueHandling = DefaultValueHandling.Ignore)]
+        public bool IsFunctionAppEditable { get; set; }
```
pragnagopa: Unit tests?

pgombar: Working on it. I created the PR to kick off Azure Pipelines and it confirmed my suspicion that this test never gets executed: `HostStatus_AdminLevel_Succeeds` in c:\git\azure-functions-host\test\WebJobs.Script.Tests.Integration\WebHostEndToEnd\SamplesEndToEndTests_CSharp.cs

Investigating why...

pgombar: Actually, it does get executed in the CI, but it passed. It should fail because I changed the number of properties in HostStatus

pgombar: Found a more suitable place for a new unit test.

src/WebJobs.Script.WebHost/Models/HostStatus.cs
```diff
@@ -63,6 +63,12 @@ public class HostStatus
         [JsonProperty(PropertyName = "processUptime", DefaultValueHandling = DefaultValueHandling.Ignore)]
         public long ProcessUptime { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether the filesystem is read-only
```
balag0: wondering if this should be enum (enabled/disabled/unknown)

basically we return enabled/disabled if we are sure and unknown otherwise. this will allow clients differentiate between editable = false since app is actually not editable / we failed to determine the status for some reason / scope this change to specific skus and let clients use some fallback logic if this api returns unknown.

src/WebJobs.Script.WebHost/Models/HostStatus.cs
```diff
@@ -63,6 +63,12 @@ public class HostStatus
...
+        /// </summary>
+        [JsonProperty(PropertyName = "isFunctionAppEditable", DefaultValueHandling = DefaultValueHandling.Ignore)]
```
balag0: isFunctionAppContentEditable?

src/WebJobs.Script.WebHost/Models/HostStatus.cs
pgombar: Not sure how that adds more value to the clients, since we want to be very restrictive with this feature, i.e. _only_ make the app contents editable if the host returns True here. Otherwise, both for Unknown/False, we want to turn it off.

pgombar: Discussed offline. Will be taking this approach since it accounts better for other SKUs (for example, we will return Unknown for non-consumption SKUs until we figure out the check for them).

pgombar:
This is now tested and ready for review, fyi @pragnagopa @balag0

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -86,6 +88,29 @@ public static bool ZipDeploymentAppSettingsExist(this IEnvironment environment)
                    !string.IsNullOrEmpty(environment.GetEnvironmentVariable(ScmRunFromPackage));
         }
 
+        public static bool AzureFilesAppSettingsExist(this IEnvironment environment)
```
balag0, line 91: Just a note for later, eventually it will be possible to have azure files enabled without explicitly having these settings when running in secretless mode. for now we will return false for such apps even though it might technically possible to edit (which is fine).

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -86,6 +88,29 @@ public static bool ZipDeploymentAppSettingsExist(this IEnvironment environment)
...
+        {
+            return !string.IsNullOrEmpty(environment.GetEnvironmentVariable(AzureFilesConnectionString)) &&
+                   !string.IsNullOrEmpty(environment.GetEnvironmentVariable(AzureFilesContentShare));
+        }
+
+        public static FunctionAppContentEditingState IsFunctionAppContentEditable(this IEnvironment environment, IOptions<ScriptApplicationHostOptions> applicationHostOptions)
```
balag0: Please add unit tests for these 2 new methods in EnvironmentExtensionsTests.

With that, IsFunctionAppContentEditable (and rename since this is no longer returning bool) could be a private method in HostController or a public method in some utility class.
currently it seems this method is not just looking up things from IEnvironment unlike the rest of the methods in this class, so feels out of place here.

balag0 (changes requested):

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
pgombar: Moved to Utility.

balag0 (approved):

brettsam (approved):

PR #7863, opened 2021-11-16
Include Extension and Dynamic Concurrency in Sync Trigger Payload

TsuyoshiUshio:
This PR is aiming for adding `host.json` information to the SyncTrigger Payload.  We add the following section from `host.json`
Make [this PR](https://github.com/Azure/azure-functions-host/pull/7527) effort to the production ready. 

# Overview
Adding the following section. The payload will be fetched from `IOptionsFormatter.Fromat` method for each extensions Options class. The Options class will be hydrated by `host.json` and `AppSettings`.  Also change the behavior of ArmCacheEnabled = false.

* `extensions` section
* `concurrency` section

**NOTE:**  AppSettings that is described in [here](https://stackoverflow.com/questions/59093584/how-to-override-logging-settings-when-running-azure-functions-locally). 

### Payload 
The order of the Payload will be changed by `IOptionsFormatter.Format` method implementation. The Payload will be camel case.

#### Default Value payload 

This is the sample payload of the default value. Each extension's version will be

##### Extension Versions

```xml
<PackageReference Include="Microsoft.Azure.WebJobs.Extensions.CosmosDB" Version="3.0.10" />
<PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask" Version="2.5.1" />
<PackageReference Include="Microsoft.Azure.WebJobs.Extensions.EventGrid" Version="2.1.0" />
<PackageReference Include="Microsoft.Azure.WebJobs.Extensions.EventHubs" Version="4.3.0" />
<PackageReference Include="Microsoft.Azure.WebJobs.Extensions.Kafka" Version="3.3.2" />
<PackageReference Include="Microsoft.Azure.WebJobs.Extensions.RabbitMQ" Version="1.1.0" />
<PackageReference Include="Microsoft.Azure.WebJobs.Extensions.SendGrid" Version="3.0.2" />
<PackageReference Include="Microsoft.Azure.WebJobs.Extensions.ServiceBus" Version="4.2.0" />
<PackageReference Include="Microsoft.Azure.WebJobs.Extensions.Storage" Version="5.0.0" />
<PackageReference Include="Microsoft.Azure.WebJobs.Extensions.Storage.Blobs" Version="5.0.0" />
<PackageReference Include="Microsoft.Azure.WebJobs.Extensions.Storage.Queues" Version="5.0.0" />
````
##### Payload 

```json
{
  "extensions": {
    "eventHubs": {
      "batchCheckpointFrequency": 1,
      "eventProcessorOptions": {
        "enableReceiverRuntimeMetric": false,
        "invokeProcessorAfterReceiveTimeout": false,
        "maxBatchSize": 10,
        "prefetchCount": 300,
        "receiveTimeout": "00:01:00"
      },
      "partitionManagerOptions": {
        "leaseDuration": "00:00:30",
        "renewInterval": "00:00:10"
      },
      "initialOffsetOptions": {
        "type": "",
        "enqueuedTimeUTC": ""
      }
    },
    "cosmosDB": {
      "connectionMode": null,
      "protocol": null,
      "leaseOptions": {
        "checkpointFrequency": {
          "explicitCheckpoint": false,
          "processedDocumentCount": null,
          "timeInterval": null
        },
        "feedPollDelay": "00:00:05",
        "isAutoCheckpointEnabled": true,
        "leaseAcquireInterval": "00:00:13",
        "leaseExpirationInterval": "00:01:00",
        "leasePrefix": null,
        "leaseRenewInterval": "00:00:17"
      }
    },
    "queue": {
      "batchSize": 16,
      "newBatchThreshold": 160,
      "maxPollingInterval": "00:01:00",
      "maxDequeueCount": 5,
      "visibilityTimeout": "00:00:00",
      "messageEncoding": "Base64"
    },
    "blob": {
      "maxDegreeOfParallelism": 160
    },
    "durableTask": {
      "httpSettings": {
        "defaultAsyncRequestSleepTimeMilliseconds": 30000
      },
      "hubName": "TestHubName",
      "storageProvider": {},
      "tracing": {
        "traceInputsAndOutputs": false,
        "allowVerboseLinuxTelemetry": false,
        "traceReplayEvents": false,
        "distributedTracingEnabled": false,
        "distributedTracingProtocol": "HttpCorrelationProtocol"
      },
      "notifications": {
        "eventGrid": null
      },
      "maxConcurrentActivityFunctions": null,
      "maxConcurrentOrchestratorFunctions": null,
      "localRpcEndpointEnabled": null,
      "extendedSessionsEnabled": false,
      "extendedSessionIdleTimeoutInSeconds": 30,
      "maxOrchestrationActions": 100000,
      "overridableExistingInstanceStates": 1,
      "entityMessageReorderWindowInMinutes": 30,
      "useGracefulShutdown": false,
      "rollbackEntityOperationsOnExceptions": true,
      "useAppLease": true,
      "appLeaseOptions": {
        "renewInterval": "00:00:25",
        "acquireInterval": "00:05:00",
        "leaseInterval": "00:01:00"
      }
    },
    "kafka": {
      "maxBatchSize": 64,
      "autoCommitIntervalMs": 200,
      "metadataMaxAgeMs": 180000,
      "socketKeepaliveEnable": true,
      "subscriberIntervalInSeconds": 1,
      "executorChannelCapacity": 1,
      "channelFullRetryIntervalInMs": 50
    },
    "RabbitMQ": {
      "hostName": null,
      "queueName": null,
      "port": 0,
      "prefetchCount": 30
    },
    "serviceBus": {
      "prefetchCount": 0,
      "messageHandlerOptions": {
        "autoComplete": true,
        "maxAutoRenewDuration": "00:05:00",
        "maxConcurrentCalls": 320
      },
      "sessionHandlerOptions": {
        "autoComplete": true,
        "maxAutoRenewDuration": "00:05:00",
        "maxConcurrentSessions": 2000,
        "messageWaitTimeout": "00:01:00"
      },
      "batchOptions": {
        "maxMessageCount": 1000,
        "operationTimeout": "00:01:00",
        "autoComplete": true
      }
    }
  },
  "concurrency": {
    "dynamicConcurrencyEnabled": false,
    "maximumFunctionConcurrency": 500,
    "cpuThreshold": 0.8,
    "snapshotPersistenceEnabled": true
  }
}
```

##### Service Bus with ver 5.0.0

```
    "serviceBus": {
      "clientRetryOptions": {
        "mode": "Exponential",
        "tryTimeout": "00:01:00",
        "delay": "00:00:00.8000000",
        "maxDelay": "00:01:00",
        "maxRetries": 3
      },
      "transportType": "AmqpTcp",
      "webProxy": "",
      "autoCompleteMessages": true,
      "prefetchCount": 0,
      "maxAutoLockRenewalDuration": "00:05:00",
      "maxConcurrentCalls": 320,
      "maxConcurrentSessions": 8,
      "maxMessageBatchSize": 1000,
      "sessionIdleTimeout": ""
    }
```

### Durable Functions
We don't support an Options class that doesn't implement `IOptionsFormatter`. However, `DurableFunctions` requires the payload for Arc scenario. Since it is already published on Arc. For avoiding the breaking change, we Serialize the DurableTaskOptions until they implement it. 
https://github.com/Azure/azure-functions-durable-extension/issues/2061

### ArmCacheEnabled behavior and Minimum Payload.

`ArmCacheEnabled = false` case and whey the SyncTriggerExceeded the MaxSize, the payload will include only `triggers`, `extensions`, and `concurrency` section. For more details refer to this comment. https://github.com/Azure/azure-functions-host/pull/7527#issuecomment-967292458

### Default Value
The default value is fetched with following order.
 
1. Get `host.json` value
2. Overwrite it by `AppSettings`
3. Bind Options object for each extensions

If an extension defined the default value on your extesions code, it won't reflected. If you define it on Options class, it can be fetched.

# Extension Owners
I tested the following extensions. The extensions payload will support the extensions that implements `IOptionsFormatter` for the Options class. We just ignore and not include the payload. 

## ServiceBus and some other extensions that has different payload for each versions
We found that ServiceBus 4.x and 5.x payload is different. For the ScaleController, we support both notations. However, be careful of the change of the `IOptionsFormatter.Format()` method.  

# Influence analysis of this PR. 

The SyncTrigger Payload is sent to the following places. There is no impact for the other components except for the size of the payload. We are carefully test it before the merge. 

## FrontEnd Role
The frontEnd receives the payload and update the database. Adding the payload doesn't matter. Also FrontEnd notify the change to the DataRole and GeoMaster.  Azure Functions Host hit the endpoint in here.  
[internal pointer A](https://msazure.visualstudio.com/One/_git/AAPT-Antares-Websites?path=/src/Hosting/frontend/krpbfe/Microsoft.Web.Hosting.Frontend/Authentication/OperationsUtils.cs&version=GBdev&_a=contents) 
[Internal pointer B](https://msazure.visualstudio.com/One/_git/AAPT-Antares-Websites?path=/src/Hosting/Microsoft.Web.Hosting/SiteManager.cs&version=GBdev&_a=contents)

## GeoMaster
Azure Portal sends request to the GeoMaster to get the functions information, in case of need sync trigger, 
it call the sync trigger api on Azure Functions Host.  Parsing is happening [FunctionAppInfo(Internal Pointer)](https://msazure.visualstudio.com/One/_git/AAPT-Antares-Websites?path=%2Fsrc%2FHosting%2FAdministrationService%2FMicrosoft.Web.Hosting.Administration.Api%2FWebEntities%2FFunctionAppInfo.cs&_a=contents&version=GBdev) includes all the parsing logic. 
[Internal Pointer C](https://msazure.visualstudio.com/One/_git/AAPT-Antares-Websites?path=/src/Hosting/AdministrationService/Microsoft.Web.Hosting.Administration.Csm.Common/Helpers/CsmFunctionsManager.cs&version=GBdev&_a=contents)
[Internal Pointer D](https://msazure.visualstudio.com/One/_git/AAPT-Antares-Websites?path=/src/Hosting/AdministrationService/Microsoft.Web.Hosting.Administration.Csm.Common/Helpers/CsmFunctionsManager.cs&version=GBdev&_a=contents)

## k4apps/KuduLite 
Currently, we are sending `extensions` section for Durable Functions on the k4apps SyncTrigger payload. No influence for this change. 
[KuduLite PR](https://github.com/Azure-App-Service/KuduLite/pull/209)
[Add host.json for sync trigger payload](https://github.com/Azure/azure-functions-host/pull/7386)
## Scale Controller 
Scale Controller fetch the Dynamic `concurrency` information from the Database. 

## ArmCacheEnabled configuration
We can configure `ArmCacheEnabled` by `WEBSITE_FUNCTIONS_ARMCACHE_ENABLED` environment variables on the Azure Functions Host side, the configuration is overwritten on the geo master by the GeoMaster configuration `FunctionsV2ARMCacheEnabled`.  If the `WEBSITE_FUNCTIONS_ARMCACHE_ENABLED`, the SyncTrigger payload will be the Array, that only includes trigger information. otherwise, it will contain the all the payload includes this change. 

# Issue describing the changes in this PR

[#issue_for_this_pr](https://github.com/Azure/azure-functions-host/issues/7394)

# Pull request checklist

* [x] My changes **do not** need to be backported to imageprevious version


src/WebJobs.Script/Config/HostJsonFileConfigurationSource.cs
```diff
@@ -71,6 +71,7 @@ public override void Load()
             {
                 JObject hostJson = LoadHostConfigurationFile();
                 ProcessObject(hostJson);
+                var a = Data;
```
TsuyoshiUshio, line 74: Ignore this. This is just for debugging and see the behavior.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -300,20 +312,26 @@ public async Task<SyncTriggersPayload> GetSyncTriggersPayload()
...
             var functionDetails = await WebFunctionsManager.GetFunctionMetadataResponse(listableFunctions, hostOptions, _hostNameProvider);
             result.Add("functions", new JArray(functionDetails.Select(p => JObject.FromObject(p))));
 
             // TEMP: refactor this code to properly add extensions in all scenario(#7394)
             // Add the host.json extensions to the payload
-            if (_environment.IsKubernetesManagedHosting())
+
+            // Glenna testing: Include the extensions payload for all apps
+            JToken hostJsonPayload = _configuration.Convert("AzureFunctionsJobHost");
+            JToken extensionsPayload = hostJsonPayload["extensions"];
```
chiangvincent: Would this result in a index invalid error? If the extensions section does not exist?

test/WebJobs.Script.Tests/Extensions/ConfigurationExtensionsTests.cs
```diff
@@ -0,0 +1,159 @@
...
+
+            // Actual value is string type, however, it is converted to int.
+            Assert.Equal(10, extensions.SelectToken("durableTask.maxConcurrentActivityFunctions"));
+            Assert.Equal(12, extensions.SelectToken("durableTask.maxConcurrentOrchestratorFunctions"));
+
+            Assert.Null(concurrency["dynamicConcurrencyEnabled"]);
+            Assert.Null(concurrency["snapshotPersistenceEnabled"]);
+
+            // An integer/boolean value will be transfered to a string value.
+            // HostJosnFileConfigurationSource is also convert these into a string value.
```
chiangvincent: nit: type in comment

TsuyoshiUshio: What do you mean "type in comment"? Let me help to understand. That is the good comment in this case? I updated the section.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
TsuyoshiUshio: Thanks. I double checked bymyself. It looks good. In this case.  hostJsonPayload will never be null. extensionPayload will be null. I have null check just after this section. 
image

TsuyoshiUshio:
I finish the task that becomes ready to testing. @chiangvincent  and @pragnagopa . 
Only one thing my concern is, ArmCacheEnabled = false case. 
I read through the code, and realized these things. 

* If it is enabled, It returns Payload A.  If not returns, Payload B. 
* The payload is sent to FrontEnd (internal pointer) or KuduLite (Arc usage) 
* FrontEnd will send the payload to the Database and notify it to GeoMaster and DataRole. 
* The Flag is enable/disabled by the environment variables.
* The ArmCacheEnabled set as EnvironmentVariables "WEBSITE_FUNCTIONS_ARMCACHE_ENABLED" However, it is overwritten by the Geo master config FunctionsV2ARMCacheEnabled. 
 
 What I don't know is, when the WEBSITE_FUNCTIONS_ARMCACHE_ENABLED is set as false on the Azure Functions Host side, In this case, the payload will be PayloadB that is very limited. Scale Controller will rely on the database payload that is posted through the FrontEnd that means, if the azure functions host has WEBSITE_FUNCTIONS_ARMCACHE_ENABLED = false can't get the enough payload to perform the scale controller to fetch. There might be the two options.  Do you have any good suggestion?

1. Keep the ArmCacheEnabled = false payload as is. If it enabled, it is not requires scaling.
2. Change the ArmCacheEnabled = false payload for including only triggers, extensions, and concurrency section. That might include the breaking change on Antres side. I see the code of FunctionAppInfo include the logic. Looks no problem. 

I'm leaning to option 2. What do you think? @mathewc In the case, we might need to deploy Private Stamp with Geo and check the logs. 

## Payload A
```
{"triggers":[{"authLevel":"anonymous","type":"httpTrigger","direction":"in","name":"req","functionName":"function1"},{"name":"myQueueItem","type":"orchestrationTrigger","direction":"in","queueName":"myqueue-items","connection":"DurableStorage","functionName":"function2","taskHubName":"TestHubValue"},{"name":"myQueueItem","type":"activityTrigger","direction":"in","queueName":"myqueue-items","connection":"DurableStorage","functionName":"function3","taskHubName":"TestHubValue"}],"functions":[{"name":"function1","script_root_path_href":"https://appName.azurewebsites.net/admin/vfs/function1/","script_href":"https://appName.azurewebsites.net/admin/vfs/function1/main.py","config_href":"https://appName.azurewebsites.net/admin/vfs/function1/function.json","test_data_href":"https://appName.azurewebsites.net/admin/vfs/ledata/function1.dat","href":"https://appName.azurewebsites.net/admin/functions/function1","invoke_url_template":"https://appname.azurewebsites.net/api/function1","language":null,"config":{"scriptFile":"main.py","disabled":false,"bindings":[{"authLevel":"anonymous","type":"httpTrigger","direction":"in","name":"req"},{"type":"http","direction":"out","name":"$return"}]},"files":null,"test_data":"{\r\n  \"scriptFile\": \"main.py\",\r\n  \"disabled\": false,\r\n  \"bindings\": [\r\n    {\r\n      \"authLevel\": \"anonymous\",\r\n      \"type\": \"httpTrigger\",\r\n      \"direction\": \"in\",\r\n      \"name\": \"req\"\r\n    },\r\n    {\r\n      \"type\": \"http\",\r\n      \"direction\": \"out\",\r\n      \"name\": \"$return\"\r\n    }\r\n  ]\r\n}","isDisabled":false,"isDirect":false,"isProxy":false},{"name":"function2","script_root_path_href":"https://appName.azurewebsites.net/admin/vfs/function2/","script_href":"https://appName.azurewebsites.net/admin/vfs/function2/main.js","config_href":"https://appName.azurewebsites.net/admin/vfs/function2/function.json","test_data_href":"https://appName.azurewebsites.net/admin/vfs/ledata/function2.dat","href":"https://appName.azurewebsites.net/admin/functions/function2","invoke_url_template":null,"language":"node","config":{"disabled":false,"scriptFile":"main.js","bindings":[{"name":"myQueueItem","type":"orchestrationTrigger","direction":"in","queueName":"myqueue-items","connection":""}]},"files":null,"test_data":"{\r\n  \"scriptFile\": \"main.py\",\r\n  \"disabled\": false,\r\n  \"bindings\": [\r\n    {\r\n      \"authLevel\": \"anonymous\",\r\n      \"type\": \"httpTrigger\",\r\n      \"direction\": \"in\",\r\n      \"name\": \"req\"\r\n    },\r\n    {\r\n      \"type\": \"http\",\r\n      \"direction\": \"out\",\r\n      \"name\": \"$return\"\r\n    }\r\n  ]\r\n}","isDisabled":false,"isDirect":false,"isProxy":false},{"name":"function3","script_root_path_href":"https://appName.azurewebsites.net/admin/vfs/function3/","script_href":"https://appName.azurewebsites.net/admin/vfs/function3/main.js","config_href":"https://appName.azurewebsites.net/admin/vfs/function3/function.json","test_data_href":"https://appName.azurewebsites.net/admin/vfs/ledata/function3.dat","href":"https://appName.azurewebsites.net/admin/functions/function3","invoke_url_template":null,"language":"node","config":{"disabled":false,"scriptFile":"main.js","bindings":[{"name":"myQueueItem","type":"activityTrigger","direction":"in","queueName":"myqueue-items","connection":""}]},"files":null,"test_data":"{\r\n  \"scriptFile\": \"main.py\",\r\n  \"disabled\": false,\r\n  \"bindings\": [\r\n    {\r\n      \"authLevel\": \"anonymous\",\r\n      \"type\": \"httpTrigger\",\r\n      \"direction\": \"in\",\r\n      \"name\": \"req\"\r\n    },\r\n    {\r\n      \"type\": \"http\",\r\n      \"direction\": \"out\",\r\n      \"name\": \"$return\"\r\n    }\r\n  ]\r\n}","isDisabled":false,"isDirect":false,"isProxy":false}],"extensions":{"durableTask":{"azureStorageConnectionStringName":"DurableStorage","HubName":"TestHubValue"}},"secrets":{"host":{"master":"aaa","function":{"TestHostFunctionKey1":"aaa","TestHostFunctionKey2":"bbb"},"system":{"TestSystemKey1":"aaa","TestSystemKey2":"bbb"}},"function":[{"name":"function1","secrets":{"TestFunctionKey1":"aaa","TestFunctionKey2":"bbb"}}]}}
```
## Payload B
```
[{"authLevel":"anonymous","type":"httpTrigger","direction":"in","name":"req","functionName":"function1"},{"name":"myQueueItem","type":"orchestrationTrigger","direction":"in","queueName":"myqueue-items","connection":"DurableStorage","functionName":"function2","taskHubName":"TestHubValue"},{"name":"myQueueItem","type":"activityTrigger","direction":"in","queueName":"myqueue-items","connection":"DurableStorage","functionName":"function3","taskHubName":"TestHubValue"}]
```

release_notes.md
```diff
@@ -3,6 +3,8 @@
 - My change description (#PR)
 -->
 
+* Add host.json payload (Concurrency and Extensions) to the SyncTriggerPayload (ArmCacheEnabled = true) (#7863)
```
mathewc: Let's reword this - we're not adding host.json. We're adding a few new select configuration sections. I'd update PR title as well to clarify

pragnagopa:
Copying comments from @mathewc - https://github.com/Azure/azure-functions-host/pull/7527#issuecomment-967292458

1. Add the "extensions" and "concurrency" sections to the sync triggers payload in both full and minimal formats. I.e. the minimal format returned when ArmCacheEnabled is true or when the full payload is oversize will now be an object format always including these two properties in addition to "triggers"
2. For minimal format, we'll no longer return an array format - we'll always return object. That ensures we can always include these properties. ScaleController and other consumers already handle both array and object formats of course.
3. Remove the Kubernetes specific changes added in https://github.com/Azure/azure-functions-host/pull/7386/files
4. Once ScaleController is updated to consume these changes, we can finally address https://github.com/Azure/azure-functions-host/issues/5390 by removing all the Durable specific code from the host. SC will have what it needs at that point.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -79,6 +82,15 @@ public FunctionsSyncManager(IConfiguration configuration, IHostIdProvider hostId
             _azureBlobStorageProvider = azureBlobStorageProvider;
         }
 
+        // Do not use in production. A method for testability.
+        internal IConfiguration Configuration
```
mathewc: I'd rather not add this. If you're trying to mock test, just pass the mock instance in on construction as needed

src/WebJobs.Script/Config/ScriptSettingsManager.cs
```diff
@@ -93,5 +96,18 @@ internal static IConfigurationBuilder CreateDefaultConfigurationBuilder()
                 .AddJsonFile("appsettings.json", optional: true)
                 .Add(new ScriptEnvironmentVariablesConfigurationSource());
         }
+
+        public static IConfiguration BuildDefaultConfigration(ScriptApplicationHostOptions applicationHostOptions, IEnvironment environment, ILoggerFactory loggerFactory, IMetricsLogger metricsLogger)
```
mathewc: Spelling errors. Please check throughout

src/WebJobs.Script/Extensions/ConfigurationExtensions.cs
```diff
@@ -0,0 +1,72 @@
...
+
+using System.Linq;
+using Microsoft.Extensions.Configuration;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs.Script.Extensions
+{
+    internal static class ConfigurationExtensions
+    {
+        internal static IConfigurationSection GetWebJobsConnectionStringSection(this IConfiguration configuration, string connectionStringName)
```
mathewc: Where are these new methods used in this PR - I don't see them being called (did a string search in PR)? This one and the other GetConnectionStringOrSettingSection below.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
TsuyoshiUshio: We can mock the IConfiguration, however, if we do it, We can not test the behavior with`HostJsonFileConfigurationSource ` and `ScriptEnvironmentVariablesConfigurationSource`  as an integration testing. If it's ok we don't need to test the behavior, I'd switch to the mock.

fabiocav:
@TsuyoshiUshio flagging this to set the expectation. I'll be reviewing this as well, but given the nature of the change and the fact that we're still iterating, it is possible we won't hear back from all reviewers until they're back from the holiday break.

pragnagopa:
> @TsuyoshiUshio flagging this to set the expectation. I'll be reviewing this as well, but given the nature of the change and the fact that we're still iterating, it is possible we won't hear back from all reviewers until they're back from the holiday break.


Thanks @fabiocav ! Yes, early feedback now is OK - which we already got. @TsuyoshiUshio  will be iterating and testing E2E and @chiangvincent will be looking into Scale Controller changes that will consume these changes. Targeting functions host release that will start in early January works !

TsuyoshiUshio:
Thank you, @fabiocav  and @mathewc ,  CC: @pragnagopa 
I understand the request that I missed. How we get the default values from extensions. I'll do the investigation until everyone is back from vacation and then we need to discuss more and that part might take time more. However, I'd like to unblock the release of Dynamic Concurrency with TargetBased Scaling. That is for very limited extensions with very limited parameter is referenced.  In this case, can we separate this PR to 2 parts to unblock the Dynamic Concurrency with TargetBased?

1. Release the first version for TargetBased Scaling.
    a. For Arc usage, we keep the same payload (send host.json info without default value) to keep the current behavior. 
    b.  Concurrency and Extensions section that only requires TargetBased scaling, we send payload with default value. 
2.  Release fully functional default value payload that includes whole extensions section

pragnagopa:
Adding notes from offline conversation with @TsuyoshiUshio  and @chiangvincent. Next steps
- Figure out how to get the default concurrency values for extensions - if this requires more changes in SDKs / host we will reach out to @fabiocav 
- Update Scale Controller with fall back logic in place - if syctrigger payload includes extension and dynamic concurrency section SC will use these otherwise will fall back to hardcoded default values

TsuyoshiUshio:
I wrote a draft pr for discuss the strategy of the default value hydration prototype. It requires `WebJobsSDK` change. The PR is one idea, that we can enable the default value hydration. However, there is several challenges. I'd like to discuss it after the vacation season has finished. I wrote the description and challenges to discuss. @pragnagopa @mathewc  @fabiocav @chiangvincent 
https://github.com/Azure/azure-webjobs-sdk/pull/2815

release_notes.md
TsuyoshiUshio: Re-written it. Could you have a look?

src/WebJobs.Script/Extensions/ConfigurationExtensions.cs
TsuyoshiUshio: Good point! Removed! Somewhat included by mistake!

src/WebJobs.Script/Config/ScriptSettingsManager.cs
TsuyoshiUshio: Thanks! Done!

TsuyoshiUshio:
Hi @chiangvincent 
if you want to merge this PR, only missing piece might be `ArmCacheEnabled` before I'm back. 
It is the old spec, and we can remove this section https://github.com/Azure/azure-functions-host/blob/ab666a2cba7d6b8b1c36b246e7ab76d10efe88b9/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs#L325-L334 and send the same payload when the ArmCacheEnable is on. We need to double check with Matthew. 
The change will introduce failure of this test.  Cache enabled should return the same payload. so that result will be the same as VerifyResultWithCacheOn() L259.   Feel free to update this PR. 

If you can wait, I'd happy to do it after I'm coming back. If it is not harmful, then you can release the current version.

pragnagopa:
Changes in this PR are tested E2E with Scale Controller on Private stamp. @fabiocav / @mathewc / @brettsam  - please review

cgillum:
Tagging @bachuv since we want to use this PR as a means of enabling Scale Controller integration with the new [Durable Functions storage backends](https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-storage-providers#comparing-storage-providers).

TsuyoshiUshio:
Hi @cgillum @bachuv 
I'm going to change this PR since we talked in offline and come to agreement with the architecture. I'm planning update this PR within this week with enabling default value. Keep on watching it.

TsuyoshiUshio:
Hi @cgillum @bachuv @chiangvincent @mathewc 

I create a sample payload for extensions section and concurrency section with the default value. Looks good? (the real one inlcudes triggers part)
I'm considering, I'll add only these extensions payload. eventHubs, cosmosDB, queue, blob,durableTask, kafka, http, serviceBus. However, it is ok to add others. I pick these for extensions that is supported by ScaleController. 

```
{
  "extensions": {
    "eventHubs": {
      "batchCheckpointFrequency": 1,
      "eventProcessorOptions": {
        "enableReceiverRuntimeMetric": false,
        "invokeProcessorAfterReceiveTimeout": false,
        "maxBatchSize": 10,
        "prefetchCount": 300,
        "receiveTimeout": "00:01:00"
      },
      "partitionManagerOptions": {
        "leaseDuration": "00:00:30",
        "renewInterval": "00:00:10"
      },
      "initialOffsetOptions": {
        "type": "",
        "enqueuedTimeUTC": ""
      }
    },
    "cosmosDB": {
      "connectionMode": null,
      "protocol": null,
      "leaseOptions": {
        "checkpointFrequency": {
          "explicitCheckpoint": false,
          "processedDocumentCount": null,
          "timeInterval": null
        },
        "feedPollDelay": "00:00:05",
        "isAutoCheckpointEnabled": true,
        "leaseAcquireInterval": "00:00:13",
        "leaseExpirationInterval": "00:01:00",
        "leasePrefix": null,
        "leaseRenewInterval": "00:00:17"
      }
    },
    "queue": {
      "batchSize": 16,
      "newBatchThreshold": 160,
      "maxPollingInterval": "00:01:00",
      "maxDequeueCount": 5,
      "visibilityTimeout": "00:00:00",
      "messageEncoding": "Base64"
    },
    "blob": {
      "maxDegreeOfParallelism": 160
    },
    "durableTask": {
      "httpSettings": {
        "defaultAsyncRequestSleepTimeMilliseconds": 30000
      },
      "hubName": "TestHubName",
      "storageProvider": {},
      "tracing": {
        "traceInputsAndOutputs": false,
        "allowVerboseLinuxTelemetry": false,
        "traceReplayEvents": false,
        "distributedTracingEnabled": false,
        "distributedTracingProtocol": "HttpCorrelationProtocol"
      },
      "notifications": {
        "eventGrid": null
      },
      "maxConcurrentActivityFunctions": null,
      "maxConcurrentOrchestratorFunctions": null,
      "localRpcEndpointEnabled": null,
      "extendedSessionsEnabled": false,
      "extendedSessionIdleTimeoutInSeconds": 30,
      "maxOrchestrationActions": 100000,
      "overridableExistingInstanceStates": 1,
      "entityMessageReorderWindowInMinutes": 30,
      "useGracefulShutdown": false,
      "rollbackEntityOperationsOnExceptions": true,
      "useAppLease": true,
      "appLeaseOptions": {
        "renewInterval": "00:00:25",
        "acquireInterval": "00:05:00",
        "leaseInterval": "00:01:00"
      }
    },
    "kafka": {
      "maxBatchSize": 64,
      "autoCommitIntervalMs": 200,
      "metadataMaxAgeMs": 180000,
      "socketKeepaliveEnable": true,
      "subscriberIntervalInSeconds": 1,
      "executorChannelCapacity": 1,
      "channelFullRetryIntervalInMs": 50
    },
    "serviceBus": {
      "prefetchCount": 0,
      "messageHandlerOptions": {
        "autoComplete": true,
        "maxAutoRenewDuration": "00:05:00",
        "maxConcurrentCalls": 320
      },
      "sessionHandlerOptions": {
        "autoComplete": true,
        "maxAutoRenewDuration": "00:05:00",
        "maxConcurrentSessions": 2000,
        "messageWaitTimeout": "00:01:00"
      },
      "batchOptions": {
        "maxMessageCount": 1000,
        "operationTimeout": "00:01:00",
        "autoComplete": true
      }
    }
  },
  "concurrency": {
    "dynamicConcurrencyEnabled": false,
    "maximumFunctionConcurrency": 500,
    "cpuThreshold": 0.8,
    "snapshotPersistenceEnabled": true
  }
}
```

src/WebJobs.Script.WebHost/ConcurrencyOptionProvider.cs
```diff
@@ -0,0 +1,42 @@
...
+using Microsoft.Azure.WebJobs.Script.Configuration;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using Newtonsoft.Json.Serialization;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost
+{
+    public class ConcurrencyOptionProvider : IConcurrencyOptionProvider
```
mathewc: Not sure why this is needed. If our goal is to simply include ConcurrencyOptions in the payload, FunctionSyncManager can simply take the options as a ctor param as normal

src/WebJobs.Script.WebHost/ExtensionsOptionProvider.cs
```diff
@@ -0,0 +1,76 @@
...
+                int idx = optionsType.Name.LastIndexOf("Options");
+                string name = optionsType.Name.Substring(0, 1).ToLower() + optionsType.Name.Substring(1, idx - 1);
+                if (_irregularNamingExtensions.ContainsKey(name))
+                {
+                    name = _irregularNamingExtensions[name];
+                }
+                var options = Activator.CreateInstance(optionsType);
+                var key = ConfigurationSectionNames.JobHost + ConfigurationPath.KeyDelimiter
+                     + "extensions" + ConfigurationPath.KeyDelimiter + name;
+                configuration.Bind(key, options);
```
mathewc: Why are you binding manually like this? In my prototype [here](https://github.com/Azure/azure-functions-host/blob/d7ac580aad39489592f3c64d88d52fb4517d5593/src/WebJobs.Script.WebHost/ExtensionOptionsProvider.cs#L43) I was able to allow the IOptions<> object to do that for me. My testing showed that worked - did you hit an issue? That way I didn't need to know configuration sections or anything. I.e. you shouldn't need _irregularNamingExtensions

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -62,8 +63,28 @@ public class FunctionsSyncManager : IFunctionsSyncManager, IDisposable
...
+        private IExtensionsOptionProvider _extensionsOptionProvider;
+
+        private IConcurrencyOptionProvider _concurrencyOptionProvider;
+
         private BlobClient _hashBlobClient;
 
+        private static HashSet<string> _supportedExtensions = new HashSet<string>
+        {
+            // Filter the extensions that supported for the SyncTrigger
+            "durableTask",
```
mathewc: The hope was that like in my prototype we could avoid having any static lists of extensions. Why is this necessary? I'd expect that ExtensionOptionsProvider only captures extension options types configured via [IWebJobsExtensionBuilder.ConfigureOptions](https://github.com/Azure/azure-webjobs-sdk/blob/3f4ec78be9f43bb041937425ced00b341883aa42/src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsExtensionBuilderExtensions.cs#L65). Was that not the case?

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -79,6 +100,22 @@ public FunctionsSyncManager(IConfiguration configuration, IHostIdProvider hostId
             _azureBlobStorageProvider = azureBlobStorageProvider;
         }
 
+        public IExtensionsOptionProvider ExtensionsOptionProvider
```
mathewc: Why the public properties for these? Instances should come via DI in ctor

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -327,20 +365,29 @@ public async Task<SyncTriggersPayload> GetSyncTriggersPayload()
 
             // Add all listable functions details to the payload
             JObject functions = new JObject();
-            string routePrefix = await WebFunctionsManager.GetRoutePrefix(hostOptions.RootScriptPath);
+            // TODO move before L387 and remove Console.WriteLine
```
mathewc: Clean up todos and get code into final shape

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -327,20 +365,29 @@ public async Task<SyncTriggersPayload> GetSyncTriggersPayload()
...
+            Console.WriteLine($"Concurrency Payload\n{concurrencyPayload.ToString()}");
+
             var listableFunctions = _functionMetadataManager.GetFunctionMetadata().Where(m => !m.IsCodeless());
             var functionDetails = await WebFunctionsManager.GetFunctionMetadataResponse(listableFunctions, hostOptions, _hostNameProvider);
             result.Add("functions", new JArray(functionDetails.Select(p => JObject.FromObject(p))));
 
             // TEMP: refactor this code to properly add extensions in all scenario(#7394)
             // Add the host.json extensions to the payload
-            if (_environment.IsKubernetesManagedHosting())
+
```
mathewc, line 335: The TEMP comment above can be removed as part of this PR right, since that's what we're addressing

src/WebJobs.Script.WebHost/Management/IFunctionsSyncManager.cs
```diff
@@ -2,11 +2,16 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host.Scale;
 
 namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
 {
     public interface IFunctionsSyncManager
     {
+        IExtensionsOptionProvider ExtensionsOptionProvider { set; }
```
mathewc: We don't want these properties in the public interface

src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
```diff
@@ -153,6 +154,14 @@ public static class WebScriptHostBuilderExtension
                     services.TryAddEnumerable(ServiceDescriptor.Singleton<IHostedService, IFileMonitoringService>(p => p.GetService<IFileMonitoringService>()));
 
                     ConfigureRegisteredBuilders(services, rootServiceProvider);
+
+                    // Hidrate ExtensionOptionProvider to FunctionSyncManager
```
mathewc, line 158: Not sure why all this is needed. Can't you simply add the new service as I did [here](https://github.com/Azure/azure-functions-host/blob/d7ac580aad39489592f3c64d88d52fb4517d5593/src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs#L157) and let DI take care of everything else?

src/WebJobs.Script/Config/HostJsonFileConfigurationSource.cs
```diff
@@ -198,7 +198,7 @@ internal JObject LoadHostConfig(string configFilePath)
                     JObject hostConfigObject;
                     try
                     {
-                        string json = File.ReadAllText(configFilePath);
+                        string json = FileUtility.ReadAllText(configFilePath);
```
mathewc, line 201: Why the change here?

src/WebJobs.Script.WebHost/ConcurrencyOptionProvider.cs
TsuyoshiUshio: We are having `IServiceCollection` and `IServiceProvider` inside of the `ExtensionsOptionProvider` we can't get it from `FunctionsSyncManager`. If I add it to `ExtensionsOptionProvider` it might be over responsibility because of the Naming. One idea is, combine ExtensionsOptionOpvider and ConcurrencyOptionProvider with `SyncTriggerOptionProvider` then it might be Ok. I'll explain why I can't use the DI hydration, I'll explain it the following comment. https://github.com/Azure/azure-functions-host/pull/7863#discussion_r794094949

src/WebJobs.Script/Config/ScriptSettingsManager.cs
```diff
@@ -93,5 +98,25 @@ internal static IConfigurationBuilder CreateDefaultConfigurationBuilder()
                 .AddJsonFile("appsettings.json", optional: true)
                 .Add(new ScriptEnvironmentVariablesConfigurationSource());
         }
+
+        public static Tuple<IServiceProvider, IServiceCollection> CreateDefaultConfigurationWithHostJsonFileAndEnvBuilder(ScriptApplicationHostOptions applicationHostOptions, IEnvironment environment, ILoggerFactory loggerFactory, IMetricsLogger metricsLogger)
```
mathewc: Looks like you're using this only in a single place in tests. Hoping we can not add it to this Type, and instead build the test host as needed in the test. If we need to do it in multiple places, then add a test helper for it, but didn't seem like it.

src/WebJobs.Script.WebHost/ExtensionsOptionProvider.cs
TsuyoshiUshio: I started testing with you code.  It creates an instance however, it didn't respect the `host.json` and AppSettings. Looks Default Value. If I change the code like it, at least it works.  However, if there is better way, I'd happy to change it.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
TsuyoshiUshio: This is a part that we need to discuss. I can remove the checking, however, what I thought was, we can use custom bindings. In this case, if the customer creates extremely big Options class (like EventHubs...) Then we'll got very big payload to the SyncTrigger, so that I thought it might be a good idea to limit supported extensions that need payload. However, if you prefer the dynamic approach rather than worrying about it, or better validation approach, it might be good idea as well.

test/WebJobs.Script.Tests/Managment/FunctionsSyncManagerTests.cs
```diff
@@ -1,110 +0,0 @@
...
-using System.Threading.Tasks;
-using Microsoft.Azure.WebJobs.Script.WebHost.Management;
-using Microsoft.Extensions.Logging;
-using Microsoft.Extensions.Options;
-using Moq;
-using Xunit;
-
-namespace Microsoft.Azure.WebJobs.Script.Tests.Managment
-{
-    public class FunctionsSyncManagerTests
```
mathewc, line 17: Where did these test go?

src/WebJobs.Script/Extensions/ConfigurationExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+                return jObject[key] = GetLowerCaseForBoolean(section.Value);
+            }
+            foreach (var child in children)
+            {
+                jObject.Add(child.Key, Parse(child));
+            }
+            return jObject;
+        }
+
+        // TODO This method is mitigation in case the caller of the SyncTrigger using DataContractJsonSerializer as  a serialization method.
```
mathewc: Why is this Type needed? Doesn't feel right - where are you using it?

test/WebJobs.Script.Tests.Integration/Management/FunctionsSyncManagerTests.cs
```diff
@@ -138,13 +147,26 @@ public FunctionsSyncManagerTests()
 
             _hostNameProvider = new HostNameProvider(_mockEnvironment.Object);
 
+            var (serviceProvider, services) = CreateDefaultConfigurationWithHostJsonFileAndEnvBuilder();
```
mathewc, line 153: Once you're taking the new provider types as ctor parameters on FunctionsSyncManager, you shouldn't need to build a separate host like this

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
TsuyoshiUshio: `FunctionSyncManager` reside in the different DI chain. The original branch, you inject `rootServiceProvider` that has `FunctionSyncManager` instance. However, `HostJsonFileConfigurationSource` and `ScriptEnvironmentVariablesConfigurationSource` is not reside in the `rootServiceProvider` instance. So that if I ran your branch, I see the FunctionSyncManager fails because no instance found about `ExtensionsOptionsProvider`.  My solution is adding Property and inject it to the `FunctionSyncManager`.  However, if it is better way exists, I'd happy to change it. 

## Doesn't work
```csharp
services.AddSingleton<ExtensionOptionsProvider>(new ExtensionOptionsProvider(rootServiceProvider, services));
```
## Works
```csharp
                   // Hidrate ExtensionOptionProvider to FunctionSyncManager
                    var functionSyncManager = rootServiceProvider.GetService<IFunctionsSyncManager>();
                    var serviceProvider = services.BuildServiceProvider();
                    functionSyncManager.ExtensionsOptionProvider = new ExtensionsOptionProvider(serviceProvider, services);
```

https://github.com/Azure/azure-functions-host/blob/d7ac580aad39489592f3c64d88d52fb4517d5593/src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs#L157

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
TsuyoshiUshio: Sure. Will do it!

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
TsuyoshiUshio: Sure! Thanks! Will do.

src/WebJobs.Script.WebHost/Management/IFunctionsSyncManager.cs
TsuyoshiUshio: Sure. I'll remove it from the interface.

src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
TsuyoshiUshio: I explained at the https://github.com/Azure/azure-functions-host/pull/7863#discussion_r794094949

test/WebJobs.Script.Tests.Integration/Management/FunctionsSyncManagerTests.cs
```diff
@@ -606,6 +692,116 @@ public async Task TrySyncTriggers_DurableV2ExtensionJson_V2ConfigPosted()
...
+        }
+
+        private string GetExpectedExtensionPayload()
+        {
+             return "{\"durableTask\":{\"httpSettings\":{\"defaultAsyncRequestSleepTimeMilliseconds\":30000},\"hubName\":\"DurableTask\",\"storageProvider\":{\"connectionStringName\":\"DurableConnection\"},\"tracing\":{\"traceInputsAndOutputs\":false,\"allowVerboseLinuxTelemetry\":false,\"traceReplayEvents\":false,\"distributedTracingEnabled\":false,\"distributedTracingProtocol\":\"HttpCorrelationProtocol\"},\"notifications\":{\"eventGrid\":null},\"maxConcurrentActivityFunctions\":null,\"maxConcurrentOrchestratorFunctions\":null,\"localRpcEndpointEnabled\":null,\"extendedSessionsEnabled\":false,\"extendedSessionIdleTimeoutInSeconds\":30,\"maxOrchestrationActions\":100000,\"overridableExistingInstanceStates\":1,\"entityMessageReorderWindowInMinutes\":30,\"useGracefulShutdown\":false,\"rollbackEntityOperationsOnExceptions\":true,\"useAppLease\":true,\"appLeaseOptions\":{\"renewInterval\":\"00:00:25\",\"acquireInterval\":\"00:05:00\",\"leaseInterval\":\"00:01:00\"}}}";
```
mathewc: Might make sense for us to have some static test file resources that we load for verification. That'll make it much easier to view/maintain. I.e. add baseline files to a Resources directory that is copied to test output (or embedded in test assembly).

src/WebJobs.Script/Config/HostJsonFileConfigurationSource.cs
TsuyoshiUshio: This is the requirement of the Integration testing.  
The current integration testing (not my update) , it mock the reading file operation when it use it in integration testing. `FileUtility` is simply delegate to File in production, for testing, it modify the request from test fixture.

src/WebJobs.Script/Config/ScriptSettingsManager.cs
TsuyoshiUshio: Sure. I'm removing from it. I just followed the old code's structure. :)

release_notes.md
```diff
@@ -3,6 +3,8 @@
 - My change description (#PR)
 -->
 
+* Add extensions and concurrency sections to the SyncTriggerPayload (ArmCacheEnabled = true) (#7863)
```
mathewc: Let's remove this ArmCacheEnabled comment. This feature should work regardless of ArmCacheEnabled state, right?

test/WebJobs.Script.Tests/Managment/FunctionsSyncManagerTests.cs
TsuyoshiUshio: The original author is me. I noticed that this test should be in the integration testing.... so that I remove this and do the integration testing.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -313,6 +350,7 @@ public async Task<SyncTriggersPayload> GetSyncTriggersPayload()
 
             if (!ArmCacheEnabled)
             {
+                // TODO: Discuss if we need other sections for Arm cache is not enabled.
```
mathewc: I described what I think we need to do for this in https://github.com/Azure/azure-functions-host/pull/7527#issuecomment-967292458

src/WebJobs.Script/Extensions/ConfigurationExtensions.cs
TsuyoshiUshio: Thanks. This doesn't need it anymore. I'll remove it.

test/WebJobs.Script.Tests.Integration/Management/FunctionsSyncManagerTests.cs
TsuyoshiUshio: I'll do some experiment because of your advice.

test/WebJobs.Script.Tests.Integration/Management/FunctionsSyncManagerTests.cs
TsuyoshiUshio: That's a good idea. Will do that.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
TsuyoshiUshio: will do!

release_notes.md
TsuyoshiUshio: Yes. will do it.

src/WebJobs.Script.WebHost/ExtensionsOptionContractResolver.cs
```diff
@@ -0,0 +1,77 @@
...
+using System.Collections.Generic;
+using System.Reflection;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost
+{
+    public class ExtensionsOptionContractResolver : DefaultContractResolver
+    {
+        private static readonly HashSet<string> _skipSerialization = new HashSet<string>
```
mathewc: I'm wondering if it would be a viable solution for us to just use the IOptionsFormatter.Format method for each options instance we construct to get the reduced set of safe values? The whole point of the formatter is to produce a curated result w/o secrets, including only the "important" values. That's exactly what we need here.

Now the contract for Format is that it returns a string, but by convention this is always a json object (can verify for all existing extensions). In that case we can call Format to get the JSON, then turn around and reparse to JObject for inclusion in the sync payload.

TsuyoshiUshio: Hi @mathewc 
I agree with you. It is clean and safe solution. 

For the current implementation, I checked several options class on each extension.  I did some research for the current implementation of each extension.
 
* EventHubsOptions -> OK
* CosmosDBOptions -> OK
* KafkaOptions -> OK
* QueuesOptions, BlobsOptions, ServiceBus -> Need to cast `IOptionsFormatter`. 
* Durable -> Not implement `IOptionsFormatter`

It looks only Durable doesn't support `IOptionFormatter`.   I just research popular extensions.

One of the reason why I didn't use this strategy, is the comment on the option. It says for logging. However, if all extensions follow the convention, I think it is clean / safer option. However, except for durable, call `Format()` might work. 

```
        /// <summary>
        /// Creates a string to be logged when the options are created from the IOptionsFactory.
        /// The returned value should be a JSON string with all secrets removed.
        /// </summary>
        /// <returns>A JSON string representing the options, with all secrets removed.</returns>
        string Format();
```

test/WebJobs.Script.Tests/Extensions/ConfigurationExtensionsTests.cs
```diff
@@ -0,0 +1,158 @@
...
+
+using System.IO;
+using System.Text;
+using Microsoft.Azure.WebJobs.Script.Extensions;
+using Microsoft.Extensions.Configuration;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using Xunit;
+
+namespace Microsoft.Azure.WebJobs.Script.Tests.Extensions
```
chiangvincent: Can you add a test showing the service bus and event hub payload that is being sent? I remember seeing this in a previous commit, perhaps I missed it? Mainly just to verify that `maxConcurrentCalls` and `maxBatchSize` are sent correctly for service bus and event hub, respectively.

chiangvincent: Following up on these two specifically because we need these values for target based scaling.
```
{
  "extensions": {
    "serviceBus": {
      "maxConcurrentCalls": 100
    },
    "eventHubs": {
      "eventProcessorOptions": {
        "maxBatchSize": 200
      }
    }
  }
}
```

src/WebJobs.Script.WebHost/ExtensionsOptionContractResolver.cs
TsuyoshiUshio: We talked offline we'll go this direction. FYI: @bachuv @cgillum  https://github.com/Azure/azure-functions-durable-extension/issues/2061

test/WebJobs.Script.Tests/Extensions/ConfigurationExtensionsTests.cs
TsuyoshiUshio: @chiangvincent  found that root cause ServiceBus doesn't work was, they have different structure for V5+ and old versions. He is working on it. However we need to be careful, the variance of the extensions. Thank you for sharing vincent!

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
TsuyoshiUshio: Done

src/WebJobs.Script/Extensions/ConfigurationExtensions.cs
TsuyoshiUshio: Done

release_notes.md
TsuyoshiUshio: Done

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
TsuyoshiUshio: Done

test/WebJobs.Script.Tests.Integration/Management/FunctionsSyncManagerTests.cs
TsuyoshiUshio: It won't get it from Constructor. https://github.com/Azure/azure-functions-host/pull/7863#discussion_r794094949

src/WebJobs.Script.WebHost/Management/IFunctionsSyncManager.cs
TsuyoshiUshio: Removed.

src/WebJobs.Script.WebHost/ConcurrencyOptionProvider.cs
TsuyoshiUshio: Removed

src/WebJobs.Script/Config/ScriptSettingsManager.cs
TsuyoshiUshio: Removed from the class.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
TsuyoshiUshio: For the new logic, extensions section will always exists.

test/WebJobs.Script.Tests/Extensions/ConfigurationExtensionsTests.cs
TsuyoshiUshio: Removed this file

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
TsuyoshiUshio: Removed

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
TsuyoshiUshio: Removed.

test/WebJobs.Script.Tests.Integration/Management/FunctionsSyncManagerTests.cs
TsuyoshiUshio: Done

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
TsuyoshiUshio: Done

TsuyoshiUshio:
Hi @mathewc  CC @pragnagopa @chiangvincent 
I updated the code and update the description https://github.com/Azure/azure-functions-host/pull/7863#issue-782193090 
Only thing we haven't discuss is the way of use of the DI and hydration of the object. The original solution didn't work, so that I make it work. I explained why it doesn't work on the previous comments.  Either way, I think the payload won't change anymore from now. So that I'll test with Arc cluster with Durable. 

Hi @cgillum @bachuv 
We talked with the extensions, concurrency payload on SyncTrigger. I updated the [description](https://github.com/Azure/azure-functions-host/pull/7863#issue-782193090) part.  I'm going to test from my end, If you guy want to test it feel free to do it.

TsuyoshiUshio:
New PR is introduced and merged

PR #7876, opened 2021-11-22
Drain mode cleanup: remove GET, return 202 Accepted

liliankasem:
1. We currently allow GET requests to the drain endpoint, however the design for the drain API is POST. Updating the controller to reflect this. I've spoken to the Antares and Atlas teams and they both already use POST so this will not be a breaking change

2. We're currently returning 200 OK on a fire and forget call where returning 202 Accepted is more fitting, the PR is updating the drain API to return 202. I have spoken to the Antares and Atlas teams about this and they both using [EnsureSuccessStatusCode](https://docs.microsoft.com/en-us/uwp/api/windows.web.http.httpresponsemessage.issuccessstatuscode?view=winrt-22000#property-value) which accepts any response in the 200 - 299 range, so this will not be a breaking change

3. Also added E2E tests for drain mode with a cancellation token sample

This PR has been tested with a private site extension in a private stamp

### Pull request checklist

        - [PR to update drain docs](https://dev.azure.com/msazure/One/_git/AAPT-Antares-Docs/pullrequest/5359725)

### DIFF
--- a/sample/CSharp/HttpTrigger-Cancellation/function.json
+++ b/sample/CSharp/HttpTrigger-Cancellation/function.json
@@ -1,0 +1,15 @@
+{
+    "bindings": [
+        {
+            "type": "httpTrigger",
+            "name": "req",
+            "direction": "in",
+            "methods": [ "get" ]
+        },
+        {
+            "type": "http",
+            "name": "$return",
+            "direction": "out"
+        }
+    ]
+}
--- a/sample/CSharp/HttpTrigger-Cancellation/run.csx
+++ b/sample/CSharp/HttpTrigger-Cancellation/run.csx
@@ -1,0 +1,26 @@
+﻿using System.Net;
+using System.Threading;
+using Microsoft.AspNetCore.Mvc;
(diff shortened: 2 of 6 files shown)


nzthiago (dismissed):
Thank you for adding the sample and extra tests. I found it strange that it's a csx function but I see all the others in the repo are as well - perhaps we should move to full C# in proc ones in the future.

brettsam (dismissed):

liliankasem:
@brettsam @nzthiago mind approving again, had to replace for pipeline run :)

nzthiago (dismissed):

brettsam (dismissed):

liliankasem:
> Thank you for adding the sample and extra tests. I found it strange that it's a csx function but I see all the others in the repo are as well - perhaps we should move to full C# in proc ones in the future.

Yeah, agree. Just following what we have for now, might be worth creating an issue to address this

nzthiago (dismissed):

nzthiago (approved):

PR #2794, opened 2018-05-04
Change Blobs.CentralizedPoisonQueue default (#1106)

mathewc:
Addresses https://github.com/Azure/azure-functions-host/issues/1106. With this change, poison messages for blobs will be co-located in the storage account of the source blob container.

When I fixed this in core SDK, we couldn't take the breaking change for v1, hence this switch default.

Issue #755, opened 2016-10-09
Using MBrace within an F# Azure Function stops the function from compiling


bruinbrown:
Currently using Functions to deploy work to an MBrace cluster on a recurring schedule. As soon as a call is made in the function body to connect to an Azure cluster, the script is picked up as being changed, the message appears showing that the script is about to be compiled but then no more log messages appear to say the compilation completes and trying to execute the function fails.
#### Repro steps
1. Create a new F# function and add the MBrace.Azure NuGet package
2. Edit the function body to connect to a cluster i.e.

```
let Run(input:string) =
    let cluster = AzureCluster.Connect("", "")
    ()
```
1. Click Save on the function
#### Expected behavior

The function should be saved and compiled ready to be executed
#### Actual behavior

The compiler hangs and the function does not get compiled
#### Known workarounds

Creating an external DLL which holds all of the MBrace related calls in a single function works i.e. having a function in a DLL which is simply called RunMBrace which handles the connecting and deploying of the cloud workflow, but I did need to ILMerge the MBrace assemblies into this DLL.

Alternatively using MBrace with a C# function works as expected and calling AzureCluster.Connect won't cause the function compilation to hang

lindydonna:
@fabiocav Thinks this is a NuGet reference issue

christopheranderson:
Next unless we have a repro for more packages to prove this is a general issue. Not all F# issues need to be done by rc, as well.

dsyme:
@bruinbrown  @lindydonna @fabiocav @christopheranderson - @sylvanc and I are actively looking into this bug this week. We think nearly all the existing F# issues boil down to the one same bug (not finding assemblies at runtime which have been resolved statically)

Suspected duplicate of #650, but we need to investigate further.

sylvanc:
This issue can be closed.

The compiler no longer hangs. Nuget's transitive package resolution and this particular set of packages don't seem to be interacting quite right. If you explicitly include MBrace.Runtime 1.3.1 it works:

```json
{
  "frameworks": {
    "net46": {
      "dependencies": {
        "MBrace.Azure": "1.4.3",
        "MBrace.Runtime": "1.3.1"
      }
    }
  }
}
```

sylvanc:
@fabiocav We're pretty sure there are some generic issues to do with non-isolation of project.json between Functions and between Function executions. We're pretty sure these aren't F# specific, as the same mechanism is used in C#.

[...]

PR #7908, opened 2021-11-30
Discussion: Functions Host Throughput Performance Improvements & Opportunities

NickCraver:
This is a draft PR **for discussion and not meant to be merged**, think "issue with some code". I figured this was the easiest way to talk about areas we can improve throughput performance in the least spammy way possible.

## Proposals Summary
These diffs are *not* proposed PRs. They are pain points or opportunities identified in profiling and experimentation where I think we can improve performance. It's likely some are viable and some are not due to constraints I'm unaware of (so let's talk!). The hope is that some of these findings help in making us faster and spur new discussions on the remaining. 

#### Goals
- Optimize what we have, with equivalent functionality - not remove any functionality.
- Don't break any use cases we have today
- Make any improvements as small single-thing PRs, not one big batch

A lot of these I file under "papercuts". Except 1 very recent regression in the `4.x` branch (which @fabiocav and @brettsam are looking at already), there is no large single item affecting performance that I can see. However, there are many small items we can potentially improve that add up. I'm numbering these in lists below so that they're easy to refer to and talk about in discussions in a quick way.

It's worth breaking these down and testing each we consider viable, though each may be a small gain in itself. In aggregate I'm seeing 24-38% gains (depending on scenario) with our typical 2 and 4 core VMs for a hello world scenario. We need to test this in App Services yet (private stamp issues...) and the gains will be proportionately smaller when the function itself is more expensive.

These proposals are split across 3 repos and need discussion PRs for each (I'll update this as I create PRs)
- (**This PR**) https://github.com/Azure/azure-functions-host/
- ([#2794](https://github.com/Azure/azure-webjobs-sdk/pull/2794)) https://github.com/Azure/azure-webjobs-sdk
- ([#755](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/755)) https://github.com/Azure/azure-webjobs-sdk-extensions

#### Optimizations for WebJobs Host

- [ ] 1. (Issue #8073 & WIP #7969) There is 1 medium/major regression in allocations from  which went live in 4.0 via [#7669](https://github.com/Azure/azure-functions-host/pull/7669) but have been behind a flag for ~14 months from [5176a5ea](https://github.com/Azure/azure-functions-host/commit/5176a5ea86861fde8ae15561d2413797eb10ed63). The last commit shows the use cases this was meant to resolve with transitive dependencies. Due to how DryIoc works, this is copying an immutable structure of service references on every call though, accounting for a large percentage of allocations.
   - Notes: @fabiocav and @brettsam are working on this already.
- [x] 2. (#7926) `IOptionsMonitor.CurrentValue` allocates on every call [due to the `_factory` capture in the lambda here](https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Options/src/OptionsMonitor.cs#L89). Instead of keeping a reference to the monitor and calling `.CurrentValue` often (which changes rarely, as I understand it - **TODO: question this assumption**), we can reference the actual value and subscribe for updates via `.OnChange` in the monitor, removing all of these allocations and overhead.
- [x] 3. (#7924) In `EtwEventGenerator` when it isn't enabled, we can avoid the `DateTime.ToString()` at the start of each activity. These strings account for ~4-5% of all allocations. If we zoom out, we can probably improve how we're passing values down through logging for `DateTime` and `Guid` specifically.
   - Open questions:
      -  Is this _always_ enabled in production? If so it's low priority but probably still worth doing. It's worth noting such a change dilutes inaccurate benchmarks though. If we could deal with less resolution (we're at 6 decimal places on `.ToString("O")`, we could lessen the allocations here, or serialize to stream potentially.
- [x] 4. (#7925) Misc captures especially in `.GetOrAdd()` for dictionaries - here are are specifically capturing context which means a lambda allocation per call, we can easily fix these items and with latest C# also add `static` to the lamda to prevent future occurences, e.g. `dict.GetOrAdd(binding, static (b) => something(b));`
- [x] 5. (#7942) We're allocating the JSON string for function metric data (via `.LogEvent()`) even though it's a rollup after the first event, which is also ~4-5% of all allocations. Given the limited cardinality of the data we can use a `ConcurrentDictionary` with a tuple or `readonly record struct` key to prevent re-allocating the string representation per call.
- [ ] 6. (#7936) `MetricsEventManager.QueuedEvents` uses a `ConcurrentDictinary<string, SystemMetricEvent>` but is allocating a string for the key on each lookup. Instead, we can use a `read-only record struct` (fastest hashing) for the lookup without an allocation.
- [ ] 7. (#7936) `MetricsEventManager.QueuedEvents` is also used in `.AddOrUpdate()` with 2 sets of opportunities to improve: We're capturing in the update lambda, but also locking under load in a `Monitor.Wait()`. Rather then doing this, we can change the structure slightly to only use `.GetOrAdd()` returning the object and _always_ updating the result. Using the `Lazy<T>` approach, we should not drop any metrics.
    - Notes: StyleCop breaks here due to an issue in the library, resolved in https://github.com/DotNetAnalyzers/StyleCopAnalyzers/pull/3401 - we'll need to upgrade StyleCop to not introduce sporadic build issues before applying this optimization.
- [ ] 8. (#7936) In the `FunctionActivityTracker` we have a `List<FunctionStartedEvent>` and locking around it to keep a running tally of all running functions (for periodic reporting). There's a lock on insert and periodically a lock while reporting which also prunes the list - this interaction creates contention under load. Since we already have a unique identifier per run (a `Guid`), we can use a `ConcurrentDictionary<Guid,FunctionStartedEvent>` and add/remove proactively rather than a large cull when logging to reduce lock contention.
    - Notes: per @mathewc, this was a problem area before and needs a history dig to assess further.
- [x] 9. (#8012) `ScriptLoggerFactory` calls `LoggerFactory.CreateLogger`, which has a `Monitor.Wait()` around the creation - so while we're not over-allocating we are stalling around loggers. We can create a `ConcurrentDictionary<string, ILogger>` here since they don't change. It's worth noting we're hitting this because we're calling the string name version of `CreateLogger`, where the generic version does not have this lock.
- [x] 10. (**Punted**) `SystemLogger` is the source of _most_ allocations, because although we're using structured logging we're also formatting it and allocating the string on the way out. This string is consumed by ETW, Linux, etc. The scope dictionary is allocated by `GetScopeDictionaryOrNull()` which we can change to an enumeration approach (or likely better: a broader think on how we might improve key/value), but it's non-trivial because `Utility.TryGetFunctionName` is in a project layer beneath the `IExternalScopeProvider` extensions.
    - Notes: this needs a larger think on improving, including version skew between libraries. If we want to change the `Dictionary` behavior, there are many questions, and this should be a thing we tackle by itself and not part of this papercut performance pass.
- [x] 11. (#7941) `Utility.TryGetFunctionName` is searching for 3 function names per call - this isn't terrible expensive in itself, but: is it for legacy or current? If we have `n` keys to look for, it would go into any key/value logging scope approach thinking.
- [x] 12. (#7938) `Dictionary` sizing: there are a hand full of spots we're allocating from resizes we can avoid.
- [ ] 13. There are a few places we're doing `.GetService<T>` per call when we can pass them in the constructor, if we can pass them once instead of live we can save lookup cost. Note that the measures for these were before number 1 above and need to be re-benchmarked solo since that's likely a compounding combination.
    - Open questions:
       -  Does this get from the correct scope? If global we're fine - if from the JobHsot we need to assess.
- [x] 14. (#7928) `HostAvailabilityCheckMiddleware` has a startup path and a runtime path. Since we're most often in the runtime path, we can elide the async/await state machine for the common case here with a local function and get a bit more performance out.
- [x] 15. (#7943) `SystemTraceMiddleware` is allocating a `Stopwatch` per call but we really only want the difference - we can shortcut without the reference allocation here via `.GetTimestamp()`.
- [ ] 16. (#7936) `FunctionMetrics` is a large allocation source under load but is a simple type - we can reduce overhead by using a `readonly struct` here.
- [ ] 17. (#7977) `loggerFactory.CreateLogger` is called many times for a function - but it's always the same logger so we could cache this somewhere like the `FunctionDescriptor`.
    - Open questions:
       - Are `FunctionDescriptors` re-created when a JobHost is? (It appears so via a fresh call through the `IFunctionIndexProvider`)
- [x] 18. (#7922) `FunctionInvokerBase` is allocating 3x `OfTypeIterator`s to grab a few specific binders - we can side-step this with a manual iteration.
- [x] 19. (#7937) `FunctionLoader` is currently using a `ReaderWriterLockSlim` for concurrency control of a single instance - instead we can `Interlocked.Exchange` the reference and save some critical section contention here.
- [x] 20. (**Punting**) `System.Environment` has an instance but doesn't have state, but more importantly we're using environmental variables a lot, and we can cache there.
     - Notes: this is worth benchmarking to re-evaluate, **but not doable in its current form dur to needing to observe sandbox changes**. One thought here is we could add code clarity and simplify configuration by having an Environmental variable -> `IOptionsMonitor<T>` that refreshed on some interval and triggered on `OnChange` when something shifted. This would eliminate a bunch of environmental variable calls spread across the codebase into a simpler structure. `AppServiceOptions` does similar things today on specialization as a reference - we could expand this class overall.

#### Other areas of interest

- `Guid`: We are parsing and `.ToString()`-ing GUIDs for logging and correlation in many places, we could probably keep a reference to both once and save a few per request.
- Logging scopes: This is a large source of overhead since we're allocating a `Dictionary<string, object>`, an execution context, and related bits at each `.BeginScope()`. Additionally, we're creating another `Dictionary<string, object>` and enumerator at the consumption site (via `.GetScopeDictionaryOrNull()`). I'm not sure how to crack this, but it feels like we can potentially find a better alternative for what's essentially key/value pairs in all cases. For example, if we had an `AsyncLocal<T>` with 1 dictionary perhaps. It's worth noting that most of the attributes are references to pre-allocated items like function descriptor components, with only a few (like start time) being per-request, so almost all of our overhead is not in the allocation of data but the structure we're conveying it over.
  - Notes: @mathewc already looked at this previously as well and if we can't change the structure there's indeed probably not much we can do, _can we change the structure?_ and keep compatibility is the primary question and that's a small investigation in itself.
- Strings: Though we're profiling on Windows predominately as that's where the tooling it the most mature (**TODO: Check assumption**), the Linux app service takes on far more allocations in the logging process due to additional formatting and normalization just before buffering (grep: `NormalizeString`).
  - TODO: Investigate if we could buffer structured log data and not allocate the string at all, instead formatting to the output stream (this may be bananapants).
- In building `StyleCop` has a race that fails around `record`...no clue there (that's why it's commented out in the branch), but something to be aware of in the `EventKey` optimization, something perhaps a newer version resolves.
- Sanitization: `SystemLogger` sanitizes a value (`formattedMessage = Sanitizer.Sanitize(formattedMessage);`) but doesn't use it during exception logging, this seems to be unintentional I just didn't want to forget it, so including here.
  - TODO: Look at history and propose.

#### Diff Notes
- Some of the changes are for benchmarking compatibility only (e.g. `samples/*` - please ignore)
- I'm not in love with passing the logger factory

I talked through most of the above with various people (thanks @mathewc @paulbatum and @fabiocav especially), and the next step is to start on the smaller and more straightforward PRs before moving to ones that require any discussion. If anyone has feedback of comments on any of the above proposals (presented here as code too), please add! I'm missing immense amounts of context and assume that I may have missed some edge cases which is why this discussion PR exists :)

### DIFF
--- a/sample/CSharp/HttpTrigger/function.json
+++ b/sample/CSharp/HttpTrigger/function.json
@@ -2,6 +2,7 @@
     "bindings": [
         {
             "type": "httpTrigger",
+            "authLevel": "anonymous",
             "name": "req",
             "direction": "in",
             "methods": [ "get" ]
--- a/sample/CSharp/host.json
+++ b/sample/CSharp/host.json
@@ -10,16 +10,17 @@
     },
     "functionTimeout": "00:05:00",
     "logging": {
-        "fileLoggingMode": "always"        
+        "fileLoggingMode": "never",
+        "console": {
+            "isEnabled": false
+        }
     },
     "extensions": {
         "sendGrid": {
             "from": "Azure Functions <samples@functions.com>"
         },
         "http": {
-            "routePrefix": "api",
-            "maxConcurrentRequests": 5,
-            "maxOutstandingRequests": 30
+            "routePrefix": "api"
         },
         "queues": {
             "visibilityTimeout": "00:00:10",
--- a/src/WebJobs.Script.WebHost/Diagnostics/AzureMonitorDiagnosticLogger.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/AzureMonitorDiagnosticLogger.cs
@@ -26,13 +26,13 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
         private readonly string _hostInstanceId;
         private readonly string _roleInstance;
         private readonly HostNameProvider _hostNameProvider;
-        private readonly IOptionsMonitor<AppServiceOptions> _appServiceOptions;
         private readonly IEventGenerator _eventGenerator;
         private readonly IEnvironment _environment;
         private readonly IExternalScopeProvider _scopeProvider;
+        private AppServiceOptions _appServiceOptions;
 
         public AzureMonitorDiagnosticLogger(string category, string hostInstanceId, IEventGenerator eventGenerator, IEnvironment environment, IExternalScopeProvider scopeProvider,
-            HostNameProvider hostNameProvider, IOptionsMonitor<AppServiceOptions> appServiceOptions)
+            HostNameProvider hostNameProvider, IOptionsMonitor<AppServiceOptions> appServiceOptionsMonitor)
         {
             _category = category ?? throw new ArgumentNullException(nameof(category));
             _hostInstanceId = hostInstanceId ?? throw new ArgumentNullException(nameof(hostInstanceId));
@@ -40,7 +40,10 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
             _environment = environment ?? throw new ArgumentNullException(nameof(environment));
             _scopeProvider = scopeProvider ?? throw new ArgumentNullException(nameof(scopeProvider));
             _hostNameProvider = hostNameProvider ?? throw new ArgumentNullException(nameof(hostNameProvider));
-            _appServiceOptions = appServiceOptions ?? throw new ArgumentNullException(nameof(appServiceOptions));
+            _ = appServiceOptionsMonitor ?? throw new ArgumentNullException(nameof(appServiceOptionsMonitor));
+
+            appServiceOptionsMonitor.OnChange(newOptions => _appServiceOptions = newOptions);
+            _appServiceOptions = appServiceOptionsMonitor.CurrentValue;
 
             _roleInstance = _environment.GetInstanceId();
 
@@ -104,7 +107,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
             using (JsonTextWriter writer = new JsonTextWriter(sw) { Formatting = Formatting.None })
             {
                 writer.WriteStartObject();
-                WritePropertyIfNotNull(writer, "appName", _appServiceOptions.CurrentValue.AppName);
+                WritePropertyIfNotNull(writer, "appName", _appServiceOptions.AppName);
                 WritePropertyIfNotNull(writer, "roleInstance", _roleInstance);
                 WritePropertyIfNotNull(writer, "message", formattedMessage);
                 WritePropertyIfNotNull(writer, "category", _category);
--- a/src/WebJobs.Script.WebHost/Diagnostics/EtwEventGenerator.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/EtwEventGenerator.cs
@@ -19,6 +19,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 
         public void LogFunctionTraceEvent(LogLevel level, string subscriptionId, string appName, string functionName, string eventName, string source, string details, string summary, string exceptionType, string exceptionMessage, string functionInvocationId, string hostInstanceId, string activityId, string runtimeSiteName, string slotName, DateTime eventTimeStamp)
         {
+            // TODO: String generated regardless of being enabled
             string formattedEventTimestamp = eventTimeStamp.ToString(EventTimestampFormat);
             using (FunctionsSystemLogsEventSource.SetActivityId(activityId))
             {
--- a/src/WebJobs.Script.WebHost/Diagnostics/EtwEventGenerator.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/EtwEventGenerator.cs
@@ -19,7 +19,11 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 
         public void LogFunctionTraceEvent(LogLevel level, string subscriptionId, string appName, string functionName, string eventName, string source, string details, string summary, string exceptionType, string exceptionMessage, string functionInvocationId, string hostInstanceId, string activityId, string runtimeSiteName, string slotName, DateTime eventTimeStamp)
         {
-            // TODO: String generated regardless of being enabled
+            if (!FunctionsSystemLogsEventSource.Instance.IsEnabled())
+            {
+                return;
+            }
+
             string formattedEventTimestamp = eventTimeStamp.ToString(EventTimestampFormat);
             using (FunctionsSystemLogsEventSource.SetActivityId(activityId))
             {
--- a/src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
@@ -178,12 +178,18 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics.Extensions
 
         public static void ExecutingHttpRequest(this ILogger logger, string mS_ActivityId, string httpMethod, string userAgent, string uri)
         {
-            _executingHttpRequest(logger, mS_ActivityId, httpMethod, userAgent, uri, null);
+            if (logger.IsEnabled(LogLevel.Information))
+            {
+                _executingHttpRequest(logger, mS_ActivityId, httpMethod, userAgent, uri, null);
+            }
         }
 
         public static void ExecutedHttpRequest(this ILogger logger, string mS_ActivityId, string identities, int statusCode, long duration)
         {
-            _executedHttpRequest(logger, mS_ActivityId, identities, statusCode, duration, null);
+            if (logger.IsEnabled(LogLevel.Information))
+            {
+                _executedHttpRequest(logger, mS_ActivityId, identities, statusCode, duration, null);
+            }
         }
 
         public static void ScriptHostServiceInitCanceledByRuntime(this ILogger logger)
--- a/src/WebJobs.Script.WebHost/Diagnostics/FunctionInstanceLogger.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/FunctionInstanceLogger.cs
@@ -20,6 +20,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 
         private readonly IMetricsLogger _metrics;
         private readonly IFunctionMetadataManager _metadataManager;
+        private ConcurrentDictionary<(string, string, bool, bool), string> _eventDataCache = new ConcurrentDictionary<(string, string, bool, bool), string>();
         private ConcurrentDictionary<BindingMetadata, string> _bindingMetricEventNames = new ConcurrentDictionary<BindingMetadata, string>();
 
         public FunctionInstanceLogger(
@@ -80,11 +81,11 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
             // log events for each of the binding types used
             foreach (var binding in metadata.Bindings)
             {
-                string eventName = _bindingMetricEventNames.GetOrAdd(binding, (existing) =>
+                string eventName = _bindingMetricEventNames.GetOrAdd(binding, static (b) =>
                 {
-                    return binding.IsTrigger ?
-                        string.Format(MetricEventNames.FunctionBindingTypeFormat, binding.Type) :
-                        string.Format(MetricEventNames.FunctionBindingTypeDirectionFormat, binding.Type, binding.Direction);
+                    return b.IsTrigger ?
+                        string.Format(MetricEventNames.FunctionBindingTypeFormat, b.Type) :
+                        string.Format(MetricEventNames.FunctionBindingTypeDirectionFormat, b.Type, b.Direction);
                 });
                 _metrics.LogEvent(eventName, metadata.Name);
             }
@@ -103,7 +104,12 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
             var function = startedEvent.FunctionMetadata;
             string eventName = success ? MetricEventNames.FunctionInvokeSucceeded : MetricEventNames.FunctionInvokeFailed;
             string functionName = function != null ? function.Name : string.Empty;
-            string data = string.Format(Microsoft.Azure.WebJobs.Script.Properties.Resources.FunctionInvocationMetricsData, startedEvent.FunctionMetadata.Language, functionName, success, Stopwatch.IsHighResolution);
+            var key = (startedEvent.FunctionMetadata.Language, functionName, success, Stopwatch.IsHighResolution);
+            if (!_eventDataCache.TryGetValue(key, out var data))
+            {
+                data = string.Format(Microsoft.Azure.WebJobs.Script.Properties.Resources.FunctionInvocationMetricsData, startedEvent.FunctionMetadata.Language, functionName, success, Stopwatch.IsHighResolution);
+                _eventDataCache[key] = data;
+            }
             _metrics.LogEvent(eventName, startedEvent.FunctionName, data);
 
             startedEvent.Data = data;
--- a/src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceFileLoggerFactory.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceFileLoggerFactory.cs
@@ -19,8 +19,8 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 
         public virtual LinuxAppServiceFileLogger GetOrCreate(string category)
         {
-            return Loggers.GetOrAdd(category,
-                c => new Lazy<LinuxAppServiceFileLogger>(() => new LinuxAppServiceFileLogger(c, _logRootPath, new FileSystem()))).Value;
+            return Loggers.GetOrAdd(category, 
+                static (c, path) => new Lazy<LinuxAppServiceFileLogger>(() => new LinuxAppServiceFileLogger(c, path, new FileSystem())), _logRootPath).Value;
         }
     }
 }
--- a/src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceFileLoggerFactory.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceFileLoggerFactory.cs
@@ -20,7 +20,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
         public virtual LinuxAppServiceFileLogger GetOrCreate(string category)
         {
             return Loggers.GetOrAdd(category,
-                c => new Lazy<LinuxAppServiceFileLogger>(() => new LinuxAppServiceFileLogger(category, _logRootPath, new FileSystem()))).Value;
+                c => new Lazy<LinuxAppServiceFileLogger>(() => new LinuxAppServiceFileLogger(c, _logRootPath, new FileSystem()))).Value;
         }
     }
 }
--- a/src/WebJobs.Script.WebHost/Diagnostics/MetricsEventManager.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/MetricsEventManager.cs
@@ -498,12 +498,15 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
                     executionTimespan = runningFunctionInfo.Duration.TotalMilliseconds;
                 }
 
+                // Don't allocate the GUID string twice, though we can probably optimize this further upstream.
+                var invocationId = runningFunctionInfo.InvocationId.ToString();
+
                 MetricsEventGenerator.LogFunctionExecutionEvent(
                     _executionId,
                     _appServiceOptions.AppName,
                     concurrency,
                     runningFunctionInfo.FunctionMetadata.Name,
-                    runningFunctionInfo.InvocationId.ToString(),
+                    invocationId,
                     executionStage.ToString(),
                     (long)executionTimespan,
                     runningFunctionInfo.Success);
@@ -512,7 +515,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
                 {
                     _metricsPublisher.AddFunctionExecutionActivity(
                         runningFunctionInfo.FunctionMetadata.Name,
-                        runningFunctionInfo.InvocationId.ToString(),
+                        invocationId,
                         concurrency,
                         executionStage.ToString(),
                         runningFunctionInfo.Success,
--- a/src/WebJobs.Script.WebHost/Diagnostics/MetricsEventManager.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/MetricsEventManager.cs
@@ -460,22 +460,27 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
                 // We only need to raise events here for functions that aren't completed.
                 // Events are raised immediately for completed functions elsewhere.
                 // Loop through and prune any completed runs
+                // Also collects the currently running functions in the same enumeration to minimize cost.
+                var running = new List<FunctionStartedEvent>();
                 foreach (var possiblyRunning in _runningFunctions)
                 {
                     if (possiblyRunning.Value.Completed)
                     {
                         _runningFunctions.TryRemove(possiblyRunning.Key, out _);
                     }
+                    else
+                    {
+                        running.Add(possiblyRunning.Value);
+                    }
                 }
 
-                // If not for the length, we could probably avoid this array allocation
-                var concurrency = _runningFunctions.Count;
+                var concurrency = running.Count;
 
                 // We calculate concurrency here based on count, since these events are raised
                 // on a background thread, so we want the actual count for this interval, not
                 // the current count.
                 var currentTime = DateTime.UtcNow;
-                foreach (var runningFunction in _runningFunctions.Values)
+                foreach (var runningFunction in running)
                 {
                     RaiseFunctionMetricEvent(runningFunction, concurrency, currentTime);
                 }
--- a/src/WebJobs.Script.WebHost/Diagnostics/MetricsEventManager.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/MetricsEventManager.cs
@@ -30,28 +30,32 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
         private readonly Timer _metricsFlushTimer;
         private readonly ILogger<MetricsEventManager> _logger;
         private bool _disposed;
-        private IOptionsMonitor<AppServiceOptions> _appServiceOptions;
+        private AppServiceOptions _appServiceOptions;
 
-        public MetricsEventManager(IOptionsMonitor<AppServiceOptions> appServiceOptions, IEventGenerator generator, int functionActivityFlushIntervalSeconds, IMetricsPublisher metricsPublisher, ILinuxContainerActivityPublisher linuxContainerActivityPublisher, ILogger<MetricsEventManager> logger, int metricsFlushIntervalMS = DefaultFlushIntervalMS)
+        public MetricsEventManager(IOptionsMonitor<AppServiceOptions> appServiceOptionsMonitor, IEventGenerator generator, int functionActivityFlushIntervalSeconds, IMetricsPublisher metricsPublisher, ILinuxContainerActivityPublisher linuxContainerActivityPublisher, ILogger<MetricsEventManager> logger, int metricsFlushIntervalMS = DefaultFlushIntervalMS)
         {
             _logger = logger ?? throw new ArgumentNullException(nameof(logger));
 
             // we read these in the ctor (not static ctor) since it can change on the fly
-            _appServiceOptions = appServiceOptions;
+            appServiceOptionsMonitor.OnChange(newOptions => _appServiceOptions = newOptions);
+            _appServiceOptions = appServiceOptionsMonitor.CurrentValue;
+
             _eventGenerator = generator;
             _functionActivityFlushIntervalSeconds = functionActivityFlushIntervalSeconds;
-            QueuedEvents = new ConcurrentDictionary<string, SystemMetricEvent>(StringComparer.OrdinalIgnoreCase);
+            QueuedEvents = new ConcurrentDictionary<EventKey, Lazy<SystemMetricEvent>>();
 
             // Initialize the periodic log flush timer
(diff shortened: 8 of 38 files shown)


NickCraver:
Leaving some perf data here on the proposed changes - note that these **do not** include a bypass for the first large regression above. If we did something like:
```diff
- scopedContext.UseInstance<IServiceProvider>(scope.ServiceProvider);
+ // scopedContext.UseInstance<IServiceProvider>(scope.ServiceProvider);
```
...we'd see even bigger gains.

All of these are running **100 connections for 60 seconds**. I've tested 15 second, 60 second, 5 minute, and 15 minute runs on each platform - while jitter exists (due to GC pause landing) at 15 seconds, the numbers are representative at 60 seconds so that's the number chosen to iterate reasonably. This is hitting the built-in sample for `/api/HttpTrigger?name=bob` doing minimal work. While `.csx` isn't representative for startup costs, it's should be equal for throughput (someone _please_ shout if this conclusion is incorrect).

This is "hello world" since we're looking for overhead in the pipeline. The percentage changes would of course decrease the more time the actual function took to run or allocated, but what we can optimize is the pipe so that's the methodology here. I'm happy to adjust if anyone sees flaws in the approach!

The test is using [crank](https://github.com/dotnet/crank) for a controlled environment with very little variance (happy to dive into this config - thanks to @sebastienros for helping me get going!). I've reproduced very similar results with a multitude of tools locally, so have confidence in the Crank runs being accurate from several angles. How these translate to running onto App Services and any bottlenecks there isn't a 1:1 because of other infrastructure pieces in play, but we'll improve one or many at a time :)

## Windows

#### 1 Core Worker

| application                             | v4x-baseline-1-win | v4x-craver-1-win |          |
| --------------------------------------- | ------------------ | ---------------- | -------- |
| CPU Usage (%)                           |                  7 |               13 |  +85.71% |
| Cores usage (%)                         |                186 |              352 |  +89.25% |
| Working Set (MB)                        |              1,101 |            2,337 | +112.26% |
| Private Memory (MB)                     |              1,621 |            3,116 |  +92.23% |
| Build Time (ms)                         |             33,751 |           12,658 |  -62.50% |
| Start Time (ms)                         |              9,987 |            8,694 |  -12.95% |
| Published Size (KB)                     |            669,886 |          669,889 |   +0.00% |
| .NET Core SDK Version                   |            6.0.100 |          6.0.100 |          |
| Max CPU Usage (%)                       |                 25 |                5 |  -80.00% |
| Max Working Set (MB)                    |              1,153 |            2,450 | +112.49% |
| Max GC Heap Size (MB)                   |                852 |            1,995 | +134.15% |
| Size of committed memory by the GC (MB) |                976 |            2,248 | +130.33% |
| Max Number of Gen 0 GCs / sec           |               3.00 |             2.00 |  -33.33% |
| Max Number of Gen 1 GCs / sec           |               2.00 |             1.00 |  -50.00% |
| Max Number of Gen 2 GCs / sec           |               1.00 |             1.00 |    0.00% |
| Max Time in GC (%)                      |               4.00 |             1.00 |  -75.00% |
| Max Gen 0 Size (B)                      |         35,144,192 |        6,321,216 |  -82.01% |
| Max Gen 1 Size (B)                      |         69,679,576 |      160,051,144 | +129.70% |
| Max Gen 2 Size (B)                      |         36,080,704 |      202,130,328 | +460.22% |
| Max LOH Size (B)                        |         11,438,296 |       55,808,352 | +387.91% |
| Max Allocation Rate (B/sec)             |      1,182,556,712 |    1,708,985,896 |  +44.52% |
| Max GC Heap Fragmentation               |                 83 |               19 |  -77.52% |
| # of Assemblies Loaded                  |                218 |              218 |    0.00% |
| Max Exceptions (#/s)                    |                282 |              225 |  -20.21% |
| Max Lock Contention (#/s)               |                196 |              173 |  -11.73% |
| Max ThreadPool Threads Count            |                295 |              356 |  +20.68% |
| Max ThreadPool Queue Length             |                 62 |               50 |  -19.35% |
| Max ThreadPool Items (#/s)              |            171,450 |          410,952 | +139.69% |
| Max Active Timers                       |                 87 |               20 |  -77.01% |
| IL Jitted (B)                           |          1,197,295 |        1,197,383 |   +0.01% |
| Methods Jitted                          |             12,355 |           12,368 |   +0.11% |

| load                   | v4x-baseline-1-win | v4x-craver-1-win |          |
| ---------------------- | ------------------ | ---------------- | -------- |
| CPU Usage (%)          |                  1 |                2 | +100.00% |
| Cores usage (%)        |                 38 |               45 |  +18.42% |
| Working Set (MB)       |                 38 |               38 |    0.00% |
| Private Memory (MB)    |                357 |              357 |    0.00% |
| Start Time (ms)        |                  0 |                0 |          |
| First Request (ms)     |                454 |              426 |   -6.17% |
| Requests/sec           |              3,350 |            4,214 |  +25.82% |
| Requests               |            201,234 |          253,278 |  +25.86% |
| Mean latency (ms)      |              37.19 |            12.22 |  -67.14% |
| Max latency (ms)       |           1,990.00 |         2,000.00 |   +0.50% |
| Bad responses          |                  0 |                0 |          |
| Socket errors          |                779 |              442 |  -43.26% |
| Read throughput (MB/s) |               0.51 |             0.65 |  +25.82% |
| Latency 50th (ms)      |               1.77 |             0.75 |  -57.40% |
| Latency 75th (ms)      |               6.77 |             0.90 |  -86.71% |
| Latency 90th (ms)      |              22.73 |             7.55 |  -66.78% |
| Latency 99th (ms)      |           1,020.00 |           327.61 |  -67.88% |

#### 2 Core Worker

| application                             | v4x-baseline-2-win | v4x-craver-2-win |          |
| --------------------------------------- | ------------------ | ---------------- | -------- |
| CPU Usage (%)                           |                  9 |                9 |    0.00% |
| Cores usage (%)                         |                242 |              241 |   -0.41% |
| Working Set (MB)                        |                950 |              581 |  -38.84% |
| Private Memory (MB)                     |              1,377 |            1,026 |  -25.49% |
| Build Time (ms)                         |             32,274 |           12,560 |  -61.08% |
| Start Time (ms)                         |             10,176 |            8,411 |  -17.34% |
| Published Size (KB)                     |            669,886 |          669,889 |   +0.00% |
| .NET Core SDK Version                   |            6.0.100 |          6.0.100 |          |
| Max CPU Usage (%)                       |                 46 |               99 | +115.22% |
| Max Working Set (MB)                    |                995 |              611 |  -38.59% |
| Max GC Heap Size (MB)                   |                816 |              261 |  -68.01% |
| Size of committed memory by the GC (MB) |                825 |              442 |  -46.42% |
| Max Number of Gen 0 GCs / sec           |               3.00 |             2.00 |  -33.33% |
| Max Number of Gen 1 GCs / sec           |               2.00 |             1.00 |  -50.00% |
| Max Number of Gen 2 GCs / sec           |               1.00 |             1.00 |    0.00% |
| Max Time in GC (%)                      |               1.00 |             1.00 |    0.00% |
| Max Gen 0 Size (B)                      |         18,206,800 |       37,828,648 | +107.77% |
| Max Gen 1 Size (B)                      |         42,753,280 |       44,379,560 |   +3.80% |
| Max Gen 2 Size (B)                      |         32,867,656 |       14,560,728 |  -55.70% |
| Max LOH Size (B)                        |          9,045,104 |       63,743,864 | +604.73% |
| Max Allocation Rate (B/sec)             |        555,466,808 |      365,475,816 |  -34.20% |
| Max GC Heap Fragmentation               |                 50 |               67 |  +33.00% |
| # of Assemblies Loaded                  |                218 |              218 |    0.00% |
| Max Exceptions (#/s)                    |                270 |              264 |   -2.22% |
| Max Lock Contention (#/s)               |                221 |              172 |  -22.17% |
| Max ThreadPool Threads Count            |                240 |              215 |  -10.42% |
| Max ThreadPool Queue Length             |                 74 |               34 |  -54.05% |
| Max ThreadPool Items (#/s)              |             91,739 |           84,523 |   -7.87% |
| Max Active Timers                       |                 53 |               16 |  -69.81% |
| IL Jitted (B)                           |          1,199,528 |        1,197,560 |   -0.16% |
| Methods Jitted                          |             12,384 |           12,367 |   -0.14% |

| load                   | v4x-baseline-2-win | v4x-craver-2-win |         |
| ---------------------- | ------------------ | ---------------- | ------- |
| CPU Usage (%)          |                  2 |                2 |   0.00% |
| Cores usage (%)        |                 46 |               62 | +34.78% |
| Working Set (MB)       |                 38 |               38 |   0.00% |
| Private Memory (MB)    |                357 |              357 |   0.00% |
| Start Time (ms)        |                  0 |                0 |         |
| First Request (ms)     |                430 |              411 |  -4.42% |
| Requests/sec           |              4,841 |            6,573 | +35.80% |
| Requests               |            290,898 |          394,870 | +35.74% |
| Mean latency (ms)      |              82.40 |            26.04 | -68.40% |
| Max latency (ms)       |           2,000.00 |         2,000.00 |   0.00% |
| Bad responses          |                  0 |                0 |         |
| Socket errors          |              1,098 |            1,237 | +12.66% |
| Read throughput (MB/s) |               0.74 |             1.01 | +35.89% |
| Latency 50th (ms)      |               1.33 |             0.91 | -31.58% |
| Latency 75th (ms)      |               5.88 |             1.23 | -79.08% |
| Latency 90th (ms)      |             241.80 |             4.36 | -98.20% |
| Latency 99th (ms)      |           1,330.00 |           995.14 | -25.18% |

#### 4 Core Worker

| application                             | v4x-baseline-4-win | v4x-craver-4-win |          |
| --------------------------------------- | ------------------ | ---------------- | -------- |
| CPU Usage (%)                           |                 16 |               16 |    0.00% |
| Cores usage (%)                         |                450 |              445 |   -1.11% |
| Working Set (MB)                        |                581 |              595 |   +2.41% |
| Private Memory (MB)                     |                964 |            1,041 |   +7.99% |
| Build Time (ms)                         |             13,298 |           12,789 |   -3.83% |
| Start Time (ms)                         |              8,222 |            8,502 |   +3.41% |
| Published Size (KB)                     |            669,886 |          669,889 |   +0.00% |
| .NET Core SDK Version                   |            6.0.100 |          6.0.100 |          |
| Max CPU Usage (%)                       |                 20 |               37 |  +85.00% |
| Max Working Set (MB)                    |                608 |              627 |   +3.12% |
| Max GC Heap Size (MB)                   |                313 |              276 |  -11.82% |
| Size of committed memory by the GC (MB) |                420 |              458 |   +9.05% |
| Max Number of Gen 0 GCs / sec           |               3.00 |             3.00 |    0.00% |
| Max Number of Gen 1 GCs / sec           |               2.00 |             2.00 |    0.00% |
| Max Number of Gen 2 GCs / sec           |               1.00 |             1.00 |    0.00% |
| Max Time in GC (%)                      |               2.00 |             2.00 |    0.00% |
| Max Gen 0 Size (B)                      |         55,087,368 |        1,319,096 |  -97.61% |
| Max Gen 1 Size (B)                      |         17,108,504 |       24,803,424 |  +44.98% |
| Max Gen 2 Size (B)                      |         49,881,520 |       15,008,672 |  -69.91% |
| Max LOH Size (B)                        |         11,535,976 |       78,325,736 | +578.97% |
| Max Allocation Rate (B/sec)             |        723,128,912 |      708,670,632 |   -2.00% |
| Max GC Heap Fragmentation               |                 72 |               43 |  -39.87% |
| # of Assemblies Loaded                  |                218 |              218 |    0.00% |
| Max Exceptions (#/s)                    |                252 |              253 |   +0.40% |
| Max Lock Contention (#/s)               |                304 |              399 |  +31.25% |
| Max ThreadPool Threads Count            |                282 |              284 |   +0.71% |
| Max ThreadPool Queue Length             |                 15 |               43 | +186.67% |
| Max ThreadPool Items (#/s)              |            117,139 |          151,139 |  +29.03% |
| Max Active Timers                       |                 24 |               16 |  -33.33% |
| IL Jitted (B)                           |          1,199,608 |        1,197,613 |   -0.17% |
| Methods Jitted                          |             12,394 |           12,369 |   -0.20% |

| load                   | v4x-baseline-4-win | v4x-craver-4-win |          |
| ---------------------- | ------------------ | ---------------- | -------- |
| CPU Usage (%)          |                  2 |                4 | +100.00% |
| Cores usage (%)        |                 63 |              101 |  +60.32% |
| Working Set (MB)       |                 38 |               38 |    0.00% |
| Private Memory (MB)    |                357 |              357 |    0.00% |
| Start Time (ms)        |                  0 |                0 |          |
| First Request (ms)     |                422 |              409 |   -3.08% |
| Requests/sec           |              9,771 |           13,126 |  +34.34% |
| Requests               |            587,219 |          788,849 |  +34.34% |
| Mean latency (ms)      |             287.51 |           301.69 |   +4.93% |
| Max latency (ms)       |           2,000.00 |         2,000.00 |    0.00% |
| Bad responses          |                  0 |                0 |          |
| Socket errors          |                383 |              623 |  +62.66% |
| Read throughput (MB/s) |               1.50 |             2.02 |  +34.67% |
| Latency 50th (ms)      |               1.47 |             0.98 |  -33.33% |
| Latency 75th (ms)      |             464.21 |           485.08 |   +4.50% |
| Latency 90th (ms)      |           1,090.00 |         1,200.00 |  +10.09% |
| Latency 99th (ms)      |           1,710.00 |         1,750.00 |   +2.34% |


## Linux

#### 1 Core Worker

| application                             | v4x-baseline-1-linux | v4x-craver-1-linux |          |
| --------------------------------------- | -------------------- | ------------------ | -------- |
| CPU Usage (%)                           |                    8 |                 13 |  +62.50% |
| Cores usage (%)                         |                  225 |                353 |  +56.89% |
| Working Set (MB)                        |                1,560 |              2,456 |  +57.44% |
| Private Memory (MB)                     |                2,251 |              3,197 |  +42.03% |
| Build Time (ms)                         |               15,736 |             12,853 |  -18.32% |
| Start Time (ms)                         |                8,831 |              8,862 |   +0.35% |
| Published Size (KB)                     |              669,886 |            669,889 |   +0.00% |
| .NET Core SDK Version                   |              6.0.100 |            6.0.100 |          |
| Max CPU Usage (%)                       |                  417 |                 12 |  -97.12% |
| Max Working Set (MB)                    |                1,632 |              2,571 |  +57.54% |
| Max GC Heap Size (MB)                   |                1,401 |              2,372 |  +69.31% |
| Size of committed memory by the GC (MB) |                1,446 |              2,343 |  +62.03% |
| Max Number of Gen 0 GCs / sec           |                 5.00 |               2.00 |  -60.00% |
| Max Number of Gen 1 GCs / sec           |                 3.00 |               1.00 |  -66.67% |
| Max Number of Gen 2 GCs / sec           |                 1.00 |               1.00 |    0.00% |
| Max Time in GC (%)                      |                10.00 |               2.00 |  -80.00% |
| Max Gen 0 Size (B)                      |          144,963,216 |         36,196,272 |  -75.03% |
| Max Gen 1 Size (B)                      |           86,562,448 |        275,733,184 | +218.54% |
| Max Gen 2 Size (B)                      |          101,344,376 |         95,365,216 |   -5.90% |
| Max LOH Size (B)                        |           14,452,784 |         36,669,488 | +153.72% |
| Max Allocation Rate (B/sec)             |        1,132,478,480 |      1,753,361,136 |  +54.83% |
| Max GC Heap Fragmentation               |                   89 |                 10 |  -88.67% |
| # of Assemblies Loaded                  |                  218 |                218 |    0.00% |
| Max Exceptions (#/s)                    |                  261 |                267 |   +2.30% |
| Max Lock Contention (#/s)               |                  200 |                120 |  -40.00% |
| Max ThreadPool Threads Count            |                  353 |                351 |   -0.57% |
| Max ThreadPool Queue Length             |                   90 |                 49 |  -45.56% |
| Max ThreadPool Items (#/s)              |              171,290 |            460,778 | +169.00% |
| Max Active Timers                       |                   55 |                 18 |  -67.27% |
| IL Jitted (B)                           |            1,197,336 |          1,197,242 |   -0.01% |
| Methods Jitted                          |               12,359 |             12,364 |   +0.04% |

| load                   | v4x-baseline-1-linux | v4x-craver-1-linux |          |
| ---------------------- | -------------------- | ------------------ | -------- |
| CPU Usage (%)          |                    1 |                  2 | +100.00% |
| Cores usage (%)        |                   37 |                 44 |  +18.92% |
| Working Set (MB)       |                   38 |                 37 |   -2.63% |
| Private Memory (MB)    |                  357 |                357 |    0.00% |
| Start Time (ms)        |                    0 |                  0 |          |
| First Request (ms)     |                  452 |                425 |   -5.97% |
| Requests/sec           |                3,605 |              4,328 |  +20.05% |
| Requests               |              216,637 |            259,966 |  +20.00% |
| Mean latency (ms)      |                38.65 |              10.86 |  -71.90% |
| Max latency (ms)       |             2,000.00 |           1,980.00 |   -1.00% |
| Bad responses          |                    0 |                  0 |          |
| Socket errors          |                  633 |                313 |  -50.55% |
| Read throughput (MB/s) |                 0.55 |               0.66 |  +20.05% |
| Latency 50th (ms)      |                 4.78 |               0.97 |  -79.71% |
| Latency 75th (ms)      |                17.75 |              11.46 |  -35.44% |
| Latency 90th (ms)      |                23.00 |              16.77 |  -27.09% |
| Latency 99th (ms)      |             1,070.00 |              83.94 |  -92.16% |

#### 2 Core Worker

| application                             | v4x-baseline-2-linux | v4x-craver-2-linux |            |
| --------------------------------------- | -------------------- | ------------------ | ---------- |
| CPU Usage (%)                           |                    9 |                  9 |      0.00% |
| Cores usage (%)                         |                  241 |                256 |     +6.22% |
| Working Set (MB)                        |                  768 |              1,043 |    +35.81% |
| Private Memory (MB)                     |                1,119 |              1,475 |    +31.81% |
| Build Time (ms)                         |               12,783 |             12,542 |     -1.89% |
| Start Time (ms)                         |                8,831 |              9,099 |     +3.03% |
| Published Size (KB)                     |              669,886 |            669,889 |     +0.00% |
| .NET Core SDK Version                   |              6.0.100 |            6.0.100 |            |
| Max CPU Usage (%)                       |                   69 |                 91 |    +31.88% |
| Max Working Set (MB)                    |                  792 |              1,092 |    +37.88% |
| Max GC Heap Size (MB)                   |                  575 |                367 |    -36.17% |
| Size of committed memory by the GC (MB) |                  631 |                927 |    +46.91% |
| Max Number of Gen 0 GCs / sec           |                 2.00 |               6.00 |   +200.00% |
| Max Number of Gen 1 GCs / sec           |                 1.00 |               4.00 |   +300.00% |
| Max Number of Gen 2 GCs / sec           |                 1.00 |               1.00 |      0.00% |
| Max Time in GC (%)                      |                 2.00 |               2.00 |      0.00% |
| Max Gen 0 Size (B)                      |           37,448,480 |         26,991,456 |    -27.92% |
| Max Gen 1 Size (B)                      |           22,032,840 |         47,746,544 |   +116.71% |
| Max Gen 2 Size (B)                      |           31,311,272 |         30,371,520 |     -3.00% |
| Max LOH Size (B)                        |            9,079,168 |        130,200,616 | +1,334.06% |
| Max Allocation Rate (B/sec)             |          545,570,904 |      1,230,643,288 |   +125.57% |
| Max GC Heap Fragmentation               |                   73 |                 75 |     +2.94% |
| # of Assemblies Loaded                  |                  218 |                218 |      0.00% |
| Max Exceptions (#/s)                    |                  258 |                273 |     +5.81% |
| Max Lock Contention (#/s)               |                  182 |                224 |    +23.08% |
| Max ThreadPool Threads Count            |                  284 |                356 |    +25.35% |
| Max ThreadPool Queue Length             |                   83 |                 31 |    -62.65% |
| Max ThreadPool Items (#/s)              |               82,432 |            210,714 |   +155.62% |
| Max Active Timers                       |                   58 |                 18 |    -68.97% |
| IL Jitted (B)                           |            1,199,552 |          1,197,167 |     -0.20% |
| Methods Jitted                          |               12,387 |             12,366 |     -0.17% |

| load                   | v4x-baseline-2-linux | v4x-craver-2-linux |         |
| ---------------------- | -------------------- | ------------------ | ------- |
| CPU Usage (%)          |                    2 |                  2 |   0.00% |
| Cores usage (%)        |                   44 |                 59 | +34.09% |
| Working Set (MB)       |                   37 |                 38 |  +2.70% |
| Private Memory (MB)    |                  357 |                357 |   0.00% |
| Start Time (ms)        |                    0 |                  0 |         |
| First Request (ms)     |                  431 |                405 |  -6.03% |
| Requests/sec           |                4,945 |              6,654 | +34.55% |
| Requests               |              297,165 |            399,681 | +34.50% |
| Mean latency (ms)      |                78.76 |              18.19 | -76.90% |
| Max latency (ms)       |             2,000.00 |           1,990.00 |  -0.50% |
| Bad responses          |                    0 |                  0 |         |
| Socket errors          |                1,088 |                947 | -12.96% |
| Read throughput (MB/s) |                 0.76 |               1.02 | +34.33% |
| Latency 50th (ms)      |                 1.38 |               0.90 | -34.78% |
| Latency 75th (ms)      |                14.35 |               1.34 | -90.66% |
| Latency 90th (ms)      |               204.56 |              10.85 | -94.70% |
| Latency 99th (ms)      |             1,370.00 |             652.57 | -52.37% |

#### 4 Core Worker

| application                             | v4x-baseline-4-linux | v4x-craver-4-linux |          |
| --------------------------------------- | -------------------- | ------------------ | -------- |
| CPU Usage (%)                           |                   17 |                 16 |   -5.88% |
| Cores usage (%)                         |                  462 |                437 |   -5.41% |
| Working Set (MB)                        |                  581 |                609 |   +4.82% |
| Private Memory (MB)                     |                  964 |              1,051 |   +9.02% |
| Build Time (ms)                         |               13,178 |             12,699 |   -3.63% |
| Start Time (ms)                         |                8,376 |              8,263 |   -1.35% |
| Published Size (KB)                     |              669,886 |            669,889 |   +0.00% |
| .NET Core SDK Version                   |              6.0.100 |            6.0.100 |          |
| Max CPU Usage (%)                       |                   20 |                 31 |  +55.00% |
| Max Working Set (MB)                    |                  608 |                637 |   +4.77% |
| Max GC Heap Size (MB)                   |                  256 |                326 |  +27.34% |
| Size of committed memory by the GC (MB) |                  422 |                466 |  +10.43% |
| Max Number of Gen 0 GCs / sec           |                 3.00 |               3.00 |    0.00% |
| Max Number of Gen 1 GCs / sec           |                 2.00 |               1.00 |  -50.00% |
| Max Number of Gen 2 GCs / sec           |                 1.00 |               1.00 |    0.00% |
| Max Time in GC (%)                      |                10.00 |               3.00 |  -70.00% |
| Max Gen 0 Size (B)                      |          180,660,400 |          4,014,040 |  -97.78% |
| Max Gen 1 Size (B)                      |           17,731,824 |         31,755,568 |  +79.09% |
| Max Gen 2 Size (B)                      |           49,771,384 |         22,430,624 |  -54.93% |
| Max LOH Size (B)                        |           10,225,032 |         79,406,688 | +676.59% |
| Max Allocation Rate (B/sec)             |          672,214,576 |        620,105,280 |   -7.75% |
| Max GC Heap Fragmentation               |                   86 |                 53 |  -38.23% |
| # of Assemblies Loaded                  |                  218 |                218 |    0.00% |
| Max Exceptions (#/s)                    |                  252 |                235 |   -6.75% |
| Max Lock Contention (#/s)               |                  396 |                186 |  -53.03% |
| Max ThreadPool Threads Count            |                  221 |                281 |  +27.15% |
| Max ThreadPool Queue Length             |                   44 |                 19 |  -56.82% |
| Max ThreadPool Items (#/s)              |              108,061 |            127,065 |  +17.59% |
| Max Active Timers                       |                   27 |                 17 |  -37.04% |
| IL Jitted (B)                           |            1,199,600 |          1,197,034 |   -0.21% |
| Methods Jitted                          |               12,393 |             12,364 |   -0.23% |

| load                   | v4x-baseline-4-linux | v4x-craver-4-linux |          |
| ---------------------- | -------------------- | ------------------ | -------- |
| CPU Usage (%)          |                    2 |                  4 | +100.00% |
| Cores usage (%)        |                   64 |                103 |  +60.94% |
| Working Set (MB)       |                   38 |                 38 |    0.00% |
| Private Memory (MB)    |                  357 |                357 |    0.00% |
| Start Time (ms)        |                    0 |                  0 |          |
| First Request (ms)     |                  421 |                400 |   -4.99% |
| Requests/sec           |                9,742 |             13,167 |  +35.17% |
| Requests               |              585,382 |            791,327 |  +35.18% |
| Mean latency (ms)      |               271.34 |             271.67 |   +0.12% |
| Max latency (ms)       |             2,000.00 |           2,000.00 |    0.00% |
| Bad responses          |                    0 |                  0 |          |
| Socket errors          |                  360 |                952 | +164.44% |
| Read throughput (MB/s) |                 1.50 |               2.02 |  +34.67% |
| Latency 50th (ms)      |                 1.48 |               0.94 |  -36.49% |
| Latency 75th (ms)      |               431.70 |             323.46 |  -25.07% |
| Latency 90th (ms)      |             1,020.00 |           1,170.00 |  +14.71% |
| Latency 99th (ms)      |             1,640.00 |           1,780.00 |   +8.54% |

Some notes:
- The increased LOH is from the metrics collector racking up results in the benchmark so IMO not something we should be concerned about.
- The increased 99th percentile latency in some cases is largely due to increased GC stalls duration. By allocating less doing so less often, we more evenly balance the heaps and cause a GC less often, but when it runs it'll can be larger and stall for slightly longer affecting the 99th percentile in a load test specifically. That's why it's also more likely to represents at higher core counts, because of the increased heap counts and amount of data involved.
  - On the balance, removal of some key critical sections and lock contention largely reduce latency for almost all cases and I think this is a good tradeoff, since we'd only see the 99th under high load and that's for a small percentage of requests. Latency for the vast majority of requests and overall net is lower.

NickCraver:
I'm pinging the folks I've talked to here, not meaning to leave owners out I'm just new :) If anyone should be tagged on an area, _please_ do so or let me know!

NickCraver:
On number 6 above, upgrading StyleCop to latest NuGet won't help - we'll see https://github.com/DotNetAnalyzers/StyleCopAnalyzers/pull/3401 which isn't related yet. I've pinged Sam to see when we might see an update there we can pull in.

liliankasem:
@NickCraver same comment as the other PR, we should use an issue to track progress on this and not a PR. Can always link to the branch diff if necessary

PR #7917, opened 2021-11-30
Performance: eliminate unintentional captures

NickCraver:
Part of #7908 for unintentional captures - this removes unintentional captures resulting in an allocation per call on these areas as well as adding a `static` guard to prevent doing so by accident in the future.

Resolves a component of #7908

### Pull request checklist

(Please double check the above - I'm guessing performance-only changes aren't something we note, but may in aggregate)

### DIFF
--- a/src/WebJobs.Script/Diagnostics/FunctionFileLoggerProvider.cs
+++ b/src/WebJobs.Script/Diagnostics/FunctionFileLoggerProvider.cs
@@ -50,7 +50,7 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics
             {
                 // Make sure that we return the same fileWriter if multiple loggers write to the same path. This happens
                 // with Function logs as Function.{FunctionName} and Function.{FunctionName}.User both go to the same file.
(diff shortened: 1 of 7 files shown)


src/WebJobs.Script/Diagnostics/FunctionFileLoggerProvider.cs
```diff
@@ -50,7 +50,7 @@ public ILogger CreateLogger(string categoryName)
             {
                 // Make sure that we return the same fileWriter if multiple loggers write to the same path. This happens
                 // with Function logs as Function.{FunctionName} and Function.{FunctionName}.User both go to the same file.
-                IFileWriter fileWriter = _fileWriterCache.GetOrAdd(filePath, (p) => _fileWriterFactory.Create(Path.Combine(_roogLogPath, filePath)));
+                IFileWriter fileWriter = _fileWriterCache.GetOrAdd(filePath, (path) => _fileWriterFactory.Create(Path.Combine(_roogLogPath, path)));
```
NickCraver, line 53: Note: this pulls in `_roogLogPath` and could be a local function to be more intuitive from the instance if we want - can fix the typo at the same time. Was aiming at minimal changes but happy to include here.

NickCraver:
Migrated to #7925 due to #7923

PR #7920, opened 2021-11-30
Performance: remove OfTypeIterator allocations

NickCraver:
Part of #7908 - in the hot path of getting function context, we allocate 3 `OfTypeIterator`s when for a little more code we can eliminate this and remove some GC overhead. This should be equivalent functionality without the overhead in both allocation and iteration (we iterate once rather than O (<3n) times.

Resolves a component of #7908

### Pull request checklist

### Additional information

Note that the exception thrown will be slightly different, but should be more informative if it occurs.

### DIFF
--- a/src/WebJobs.Script/Description/FunctionInvokerBase.cs
+++ b/src/WebJobs.Script/Description/FunctionInvokerBase.cs
@@ -84,14 +84,34 @@ namespace Microsoft.Azure.WebJobs.Script.Description
 
         private static FunctionInvocationContext GetContextFromParameters(object[] parameters, FunctionMetadata metadata)
         {
+            ExecutionContext functionExecutionContext = null;
+            Binder binder = null;
+            ILogger logger = null;
+
+            for (var i = 0; i < parameters.Length; i++)
+            {
+                switch (parameters[i])
+                {
+                    case ExecutionContext fc:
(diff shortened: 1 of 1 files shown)


src/WebJobs.Script/Description/FunctionInvokerBase.cs
```diff
@@ -84,15 +84,35 @@ public async Task<object> Invoke(object[] parameters)
...
+                    case ExecutionContext fc:
+                        functionExecutionContext ??= fc;
+                        break;
+                    case Binder b:
+                        binder ??= b;
+                        break;
+                    case ILogger l:
+                        logger ??= l;
+                        break;
+                }
```
NickCraver, line 104: Note that we can write this more compact, but I'm not sure everyone will find it intuitive:
```cs
for (var i = 0; i < parameters.Length; i++)
{
    _ = parameters[i] switch
    {
        ExecutionContext fc => functionExecutionContext ??= fc,
        Binder b => binder ??= b,
        ILogger l => logger ??= l,
        _ => (object)null,
    };
}
```

liliankasem (approved):

NickCraver:
Moved to #7922 to get on an internal branch (builds aren't cooperating on forks)

PR #7919, opened 2021-11-30
Performance: eliminate ETW date strings when disabled

NickCraver:
Part of #7908. I'm not sure for which cases this is disabled, but for any scenario where it's not activated in (which may be zero in production, to be fair), this accounts for roughly 5% of allocations from the `DateTime.ToString("O")` immediately below.

### Issue describing the changes in this PR

Resolves a component of #7908

### Pull request checklist

* [x] My changes **do not** require documentation changes
    * [ ] Otherwise: Documentation issue linked to PR
* [x] My changes **should not** be added to the release notes for the next release
    * [ ] Otherwise: I've added my notes to `release_notes.md`
* [x] My changes **do not** need to be backported to a previous version
    * [ ] Otherwise: Backport tracked by issue/PR #issue_or_pr
* [x] I have added all required tests (Unit tests, E2E tests)

PR #7924, opened 2021-12-01
Performance: eliminate ETW date strings when disabled

NickCraver:
Part of #7908. I'm not sure for which cases this is disabled, but for any scenario where it's not activated in (which may be zero in production, to be fair), this accounts for roughly 5% of allocations from the `DateTime.ToString("O")` immediately below.

Resolves a component of #7908

### Pull request checklist

(switching to internal branch - migration of #7919)

### DIFF
--- a/src/WebJobs.Script.WebHost/Diagnostics/EtwEventGenerator.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/EtwEventGenerator.cs
@@ -19,6 +19,11 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 
         public void LogFunctionTraceEvent(LogLevel level, string subscriptionId, string appName, string functionName, string eventName, string source, string details, string summary, string exceptionType, string exceptionMessage, string functionInvocationId, string hostInstanceId, string activityId, string runtimeSiteName, string slotName, DateTime eventTimeStamp)
         {
+            if (!FunctionsSystemLogsEventSource.Instance.IsEnabled())
+            {
+                return;
+            }
+
             string formattedEventTimestamp = eventTimeStamp.ToString(EventTimestampFormat);
             using (FunctionsSystemLogsEventSource.SetActivityId(activityId))
             {


liliankasem (approved):
lgtm but might be lacking context @brettsam?

src/WebJobs.Script.WebHost/Diagnostics/EtwEventGenerator.cs
```diff
@@ -19,6 +19,11 @@ public EtwEventGenerator()
 
         public void LogFunctionTraceEvent(LogLevel level, string subscriptionId, string appName, string functionName, string eventName, string source, string details, string summary, string exceptionType, string exceptionMessage, string functionInvocationId, string hostInstanceId, string activityId, string runtimeSiteName, string slotName, DateTime eventTimeStamp)
         {
+            if (!FunctionsSystemLogsEventSource.Instance.IsEnabled())
```
mathewc, line 22: There's also another method below where we're doing a date format inline. If our goal is to avoid all unnecessary formatting when disabled, I think we'd handle that upstream on creation of the generator instance [here](https://github.com/Azure/azure-functions-host/blob/ebe1aafa9bf01ed56202b21a1a47a1897d2cebf9/src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs#L117). If etw source is disabled, the final clause would just create and register a NullEventGenerator which noops.

As far as impact of these changes, the only time we'd be creating an ETW generator with the source disabled would be in local test scenarios. In production, the ETW generator will only be created when deployed to windows, and the source will be enabled there.

So while I don't think this will have production impact, I'm not opposed to a change like this.

NickCraver: I talked with @brettsam a bit this morning as well to catch up - if we're all onboard to `"{dateVariable:O}"` inline for better perf and fewer allocations, I can take a pass at that as a single PR and eliminate those items...but it won't help this specific case any since it's ETW.

Alternatively to both that and this change, I could move the date string down into the `FunctionsSystemLogsEventSource` calls within their existing IsEnabled checks (changing their signature from `string` to `DateTime`), though the code would be in a few spots - it'd be consolidated and like you'd likely expect in a disabled logger - thoughts on that as well?

Looking at this further, I see in many cases the recommended path is to pass a `long` for the `DateTime` (e.g. `.ToFileTime()`) into the `EventData` - that's a riskier change of course but any idea if that's been explore before and we could eliminate this allocation path completely? (only the `summary` would remain from `SystemLogger`)

danielmarbach: Sorry, just rando watching this PR because of educational interest. Here are a few thoughts. 

> Alternatively to both that and this change, I could move the date string down into the FunctionsSystemLogsEventSource calls within their existing IsEnabled checks

That seems to be the better alternative. Looking from the perspective of the event generator and the event source interaction, it seems the `IsEnabled` check that is supposed to be done inside the source is now leaking up one abstraction level higher. Changing the method signature on the source to datetime and make the formatting the responsibility of the source itself seems to be a reasonable choice, given that you then also free the callers from having to make the decision of how to format (unless that was the intent of the previous design)

fabiocav (approved):

PR #7925, opened 2021-12-01
Performance: eliminate unintentional captures

NickCraver:
Part of #7908 for unintentional captures - this removes unintentional captures resulting in an allocation per call on these areas as well as adding a `static` guard to prevent doing so by accident in the future.

Resolves a component of #7908

### Pull request checklist

(Please double check the above - I'm guessing performance-only changes aren't something we note, but may in aggregate)

(switching to internal branch - migration of #7917)

#### Crank summary

| application                             | dev         | captures    |         |
| Working Set (MB)                        |         378 |         372 |  -1.59% |
| Private Memory (MB)                     |       1,181 |       1,202 |  +1.78% |
| Build Time (ms)                         |      15,235 |      15,315 |  +0.53% |
| Start Time (ms)                         |       4,757 |       4,799 |  +0.88% |
| Published Size (KB)                     |     674,605 |     674,605 |   0.00% |
| .NET Core SDK Version                   |     6.0.100 |     6.0.100 |         |
| Max Working Set (MB)                    |         396 |         389 |  -1.77% |
| Max GC Heap Size (MB)                   |         145 |         123 | -15.17% |
| Size of committed memory by the GC (MB) |         207 |         202 |  -2.42% |
| Max Number of Gen 0 GCs / sec           |       13.00 |        9.00 | -30.77% |
| Max Time in GC (%)                      |        6.00 |        7.00 | +16.67% |
| Max Gen 0 Size (B)                      |  92,442,184 |  44,702,240 | -51.64% |
| Max Gen 1 Size (B)                      |  11,112,344 |  12,280,224 | +10.51% |
| Max Gen 2 Size (B)                      |  29,196,432 |  30,020,944 |  +2.82% |
| Max LOH Size (B)                        |   9,714,720 |   9,190,152 |  -5.40% |
| Max Allocation Rate (B/sec)             | 334,226,264 | 301,663,024 |  -9.74% |
| Max GC Heap Fragmentation               |          83 |          73 | -11.96% |
| Max Exceptions (#/s)                    |         258 |         264 |  +2.33% |
| Max Lock Contention (#/s)               |         109 |         144 | +32.11% |
| Max ThreadPool Threads Count            |          29 |          32 | +10.34% |
| Max ThreadPool Queue Length             |          92 |          89 |  -3.26% |
| Max ThreadPool Items (#/s)              |      36,134 |      36,670 |  +1.48% |
| Max Active Timers                       |          85 |          91 |  +7.06% |
| IL Jitted (B)                           |   1,054,396 |   1,054,433 |  +0.00% |
| Methods Jitted                          |      10,879 |      10,880 |  +0.01% |
| Avg Allocation Rate (B/sec)             | 303,164,832 | 272,884,539 |  -9.99% |
| 90th Allocation Rate (B/sec)            | 333,317,216 | 299,033,232 | -10.29% |

| load                   | dev     | captures |         |
| ---------------------- | ------- | -------- | ------- |
| CPU Usage (%)          |       1 |        1 |   0.00% |
| Cores usage (%)        |      37 |       37 |   0.00% |
| Working Set (MB)       |      38 |       38 |   0.00% |
| Private Memory (MB)    |     357 |      357 |   0.00% |
| Start Time (ms)        |       0 |        0 |         |
| First Request (ms)     |     338 |      340 |  +0.59% |
| Requests/sec           |   8,082 |    8,110 |  +0.35% |
| Requests               | 485,330 |  487,418 |  +0.43% |
| Mean latency (ms)      |   13.76 |    13.53 |  -1.67% |
| Max latency (ms)       |  205.02 |   164.49 | -19.77% |
| Bad responses          |       0 |        0 |         |
| Socket errors          |       0 |        0 |         |
| Read throughput (MB/s) |    1.24 |     1.25 |  +0.81% |
| Latency 50th (ms)      |   10.00 |    10.02 |  +0.20% |
| Latency 75th (ms)      |   15.87 |    15.64 |  -1.45% |
| Latency 90th (ms)      |   27.73 |    26.78 |  -3.43% |
| Latency 99th (ms)      |   62.69 |    59.40 |  -5.25% |


src/WebJobs.Script/Diagnostics/FunctionFileLoggerProvider.cs
```diff
@@ -50,7 +50,7 @@ public ILogger CreateLogger(string categoryName)
             {
                 // Make sure that we return the same fileWriter if multiple loggers write to the same path. This happens
                 // with Function logs as Function.{FunctionName} and Function.{FunctionName}.User both go to the same file.
-                IFileWriter fileWriter = _fileWriterCache.GetOrAdd(filePath, (p) => _fileWriterFactory.Create(Path.Combine(_roogLogPath, filePath)));
+                IFileWriter fileWriter = _fileWriterCache.GetOrAdd(filePath, (path) => _fileWriterFactory.Create(Path.Combine(_roogLogPath, path)));
```
NickCraver: Note: this pulls in `_roogLogPath` and could be a local function to be more intuitive from the instance if we want - can fix the typo at the same time. Was aiming at minimal changes but happy to include here.

src/WebJobs.Script.WebHost/Helpers/MediaTypeMap.cs
```diff
@@ -28,11 +28,11 @@ public MediaTypeHeaderValue GetMediaType(string fileExtension)
             }
```
sebastienros: dotnet 6 has `ArgumentNullException.ThrowIfNull` which prevents you from invoking throw. Having `throw` in a method prevents it from being inlined. So at least for `ArgumentNullException` I would suggest to use it when possible (like always). And for the other exceptions it might make sense to use a custom helper methods.

src/WebJobs.Script.WebHost/Helpers/MediaTypeMap.cs
```diff
@@ -28,11 +28,11 @@ public MediaTypeHeaderValue GetMediaType(string fileExtension)
...
 
             return _mediatypeMap.GetOrAdd(fileExtension,
-                (extension) =>
+                static (extension) =>
                 {
                     try
                     {
-                        if (_mimeMapping.TryGetContentType(fileExtension, out string mediaTypeValue))
+                        if (_mimeMapping.TryGetContentType(extension, out string mediaTypeValue))
```
sebastienros, line 32: If you know that extensions is `".<something>"` then you can access `FileExtensionContentTypeProvider.Mapping` directly and it will prevent the internal substring or lookups on the argument since it expects a path.

sebastienros: But apparently this is just to lazily create new `MediaTypeHeaderValue` instances based on the values from `FileExtensionContentTypeProvider`. In that case why not just eagerly create all the entries from the `Mapping` dictionary, and not use a `ConcurrentDictionary` at all, but a static `Dictionary` that already contains all the possible values.

NickCraver: Another good question - I just did some benchmarking to compare here. Overall: the spin-up would be much slower (more so on a 2-core machine we're typically on). If we did the full spin-up we would indeed gain a bit of throughput, but at the expensive of startup time which is a major effort as well. Comparison of spin-up only:

|               Method |        Mean |       Error |    StdDev |  Gen 0 |  Gen 1 | Allocated |
|--------------------- |------------:|------------:|----------:|-------:|-------:|----------:|
|     ConcurrentSpinup |    407.1 ns |    296.8 ns |  16.27 ns | 0.1068 | 0.0010 |      2 KB |
| FullDictionarySpinup | 36,976.1 ns | 15,953.0 ns | 874.44 ns | 2.7466 | 0.4272 |     46 KB |

I think more importantly though: there may be a breaking change around Markdown - the `Mappings` version has `text/markdown` where we have `text/plain` here...not sure if that may subtle break someone but definitely possible. Here's the benchmark setup if anyone's curious: https://gist.github.com/NickCraver/e4515051e60d393ba4c124cb1526e660

Now that's not a blocker IMO, since we can simply overwrite it last thing to keep compat - just noting a practical difference from a full move.

I'll add this to the look-at list - my next 2 big things to look into are: 1) startup time, and 2) out-of-process workers that light up all the gRPC paths. 

Curious on thoughts here ^

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceFileLoggerFactory.cs
```diff
@@ -20,7 +20,7 @@ public LinuxAppServiceFileLoggerFactory()
         public virtual LinuxAppServiceFileLogger GetOrCreate(string category)
         {
             return Loggers.GetOrAdd(category,
-                c => new Lazy<LinuxAppServiceFileLogger>(() => new LinuxAppServiceFileLogger(category, _logRootPath, new FileSystem()))).Value;
+                static (c, path) => new Lazy<LinuxAppServiceFileLogger>(() => new LinuxAppServiceFileLogger(c, path, new FileSystem())), _logRootPath).Value;
```
danielmarbach, line 23: The lambda of the lazy could also be marked as static

src/WebJobs.Script/Diagnostics/FunctionFileLoggerProvider.cs
danielmarbach: You could pass in the fileWriterFactory as an argument by using

https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.getoradd?view=net-5.0#System_Collections_Concurrent_ConcurrentDictionary_2_GetOrAdd__1__0_System_Func__0___0__1____0_

and then make this static too if I'm not mistaken

NickCraver: Oh damn I have exactly this but it didn't make it into the diff branch (maybe be on the other laptop) - thanks for pointing out! I'll get this one amended.

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceFileLoggerFactory.cs
NickCraver: The lazy bits _are_ using context, so no static there - it'd fail :)

src/WebJobs.Script/Diagnostics/FunctionFileLoggerProvider.cs
NickCraver: Ah I remember where this was now - we need 2 bits of context. I was going to benchmark the ValueTuple needed here vs. the lambda creation per call and see what the overall overhead is...and then forgot. I'll do that Monday and report back :)

danielmarbach: Ah yeah it needs two contexts. These things are sometimes hard to spot in the diff view only. I confess I didn't checkout the code 🙄

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceFileLoggerFactory.cs
danielmarbach: Damn I missed that in the diff view. Sorry for that

src/WebJobs.Script.WebHost/Helpers/MediaTypeMap.cs
NickCraver: Ooooh I indeed forgot this was added - thanks! I'll do a pass on the solution separately to this now that we're all .NET 6 here, SDK is trickier but there may be some places it benefits measurably.

fabiocav: This is interesting! @NickCraver we can engage more folks here if that would be helpful.

src/WebJobs.Script/Diagnostics/FunctionFileLoggerProvider.cs
fabiocav: Here, you should just go ahead with the change.

sebastienros: Reminder that CD is not locking on `GetOrAdd` so multiple threads can concurrently invoke the lambda, even though a single value will be added to the dictionary. It might be recommended to use the `Lazy` pattern.

sebastienros: And I checked `FileWriter` which should be disposed, so it's even more important to not lose these unnecessary instances. I don't know the cardinality of these writers, but they each create a `Timer`, which is expected to be disposed. There is no finalizer, so it might be necessary to add one, even more that the Dispose is calling GC.SuppressFinlaize so it apparently expects one to be here.

NickCraver: ACK - done!

src/WebJobs.Script.WebHost/Helpers/MediaTypeMap.cs
NickCraver: I was curious to see how much impact it'd have - minor as expected but this doesn't include bindings or the SDK which would increase effect - here's a PR moving to `ThrowIfNull` :) #7958

fabiocav (approved):

PR #7926, opened 2021-12-01
Performance: IOptionsMonitor.CurrentValue optimizations

NickCraver:
Part of #7908. The structure of this for reloading is great - and I think we can extend it further to simplify how we use environmental variables. In profiling though, turns out this allocates every call [due to a `_factory` capture in the lambda](https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Options/src/OptionsMonitor.cs#L89). This change instead references the current instance and subscribes to updates - increasing the speed of every call and removing allocations from it.

Resolves a component of #7908

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.WebHost/Diagnostics/AzureMonitorDiagnosticLogger.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/AzureMonitorDiagnosticLogger.cs
@@ -26,13 +26,13 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
         private readonly string _hostInstanceId;
         private readonly string _roleInstance;
         private readonly HostNameProvider _hostNameProvider;
-        private readonly IOptionsMonitor<AppServiceOptions> _appServiceOptions;
         private readonly IEventGenerator _eventGenerator;
         private readonly IEnvironment _environment;
         private readonly IExternalScopeProvider _scopeProvider;
+        private AppServiceOptions _appServiceOptions;
 
         public AzureMonitorDiagnosticLogger(string category, string hostInstanceId, IEventGenerator eventGenerator, IEnvironment environment, IExternalScopeProvider scopeProvider,
-            HostNameProvider hostNameProvider, IOptionsMonitor<AppServiceOptions> appServiceOptions)
+            HostNameProvider hostNameProvider, IOptionsMonitor<AppServiceOptions> appServiceOptionsMonitor)
         {
             _category = category ?? throw new ArgumentNullException(nameof(category));
             _hostInstanceId = hostInstanceId ?? throw new ArgumentNullException(nameof(hostInstanceId));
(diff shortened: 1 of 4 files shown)


brettsam (approved):

src/WebJobs.Script.WebHost/Diagnostics/AzureMonitorDiagnosticLogger.cs
```diff
@@ -26,21 +26,24 @@ public class AzureMonitorDiagnosticLogger : ILogger
...
             _category = category ?? throw new ArgumentNullException(nameof(category));
             _hostInstanceId = hostInstanceId ?? throw new ArgumentNullException(nameof(hostInstanceId));
             _eventGenerator = eventGenerator ?? throw new ArgumentNullException(nameof(eventGenerator));
             _environment = environment ?? throw new ArgumentNullException(nameof(environment));
             _scopeProvider = scopeProvider ?? throw new ArgumentNullException(nameof(scopeProvider));
             _hostNameProvider = hostNameProvider ?? throw new ArgumentNullException(nameof(hostNameProvider));
-            _appServiceOptions = appServiceOptions ?? throw new ArgumentNullException(nameof(appServiceOptions));
+            _ = appServiceOptionsMonitor ?? throw new ArgumentNullException(nameof(appServiceOptionsMonitor));
+
+            appServiceOptionsMonitor.OnChange(newOptions => _appServiceOptions = newOptions);
```
danielmarbach, line 45: How does the additional closure here compare to the allocations of the factory during the CurrentValue call?

src/WebJobs.Script.WebHost/Diagnostics/AzureMonitorDiagnosticLogger.cs
```diff
@@ -104,7 +107,7 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
             using (JsonTextWriter writer = new JsonTextWriter(sw) { Formatting = Formatting.None })
             {
                 writer.WriteStartObject();
-                WritePropertyIfNotNull(writer, "appName", _appServiceOptions.CurrentValue.AppName);
+                WritePropertyIfNotNull(writer, "appName", _appServiceOptions.AppName);
```
danielmarbach, line 110: Given that this introduces the risk of stale reads as well as updates to the options happening concurrent to the reads can that be problematic?

danielmarbach (commented):
Hope I don't derail things. I'm here on this PR to learn a thing or two. Please don't hesitate to entirely ignore my questions.

src/WebJobs.Script.WebHost/Diagnostics/AzureMonitorDiagnosticLogger.cs
NickCraver: In this case we're talking about 1 closure allocation for the event -per logger_ (so, pretty much 1), whereas today we're talking about _per function execution_ which is an unbounded `n`. So indeed this isn't free, but it's way cheaper and only paid once :)

src/WebJobs.Script.WebHost/Diagnostics/AzureMonitorDiagnosticLogger.cs
NickCraver: In practice, there's little to no race difference here. For example, if you called `.CurrentValue.X` 5 times in a row (see `SystemLogger` for this actually happening), you'll already tear work across the reference boundary and get skew during a single function call. You can work around this with `IOptionsSnapshot` (which allocates a copy per scope) or getting a ref and re-using once in the method, but the reality is somewhere in a request pipe between methods, things will change and that should be expected if options are _ever_ changing.

I'm not sure the phrasing is awesome there, but the baseline is: yep, it's expected values may change as you go. That's why you should try to avoid:

```cs
if (setting.Thing != null)
{
    var a = setting.Thing.Something();
}
```
...because after that `if` check, it may _become_ null if volatile/reloadable. Instead, strive for:
```cs
var a = setting.Thing?.Something();
```
...or:
```cs
var snapshot = setting.Thing;
if (snapshot != null)
{
    var a = snapshot.Something();
}
```
In general, be aware that these can change _literally at any time_ and plan accordingly, especially if nullable.

PR #6547, opened 2020-08-21
Removing unnecessary per invocation allocations

mathewc:
Removing per invocation allocations by removing FunctionInstanceMonitor/RunningFunctionInfo types.

As part of these changes, also improving the locking strategy MetricsEventManager uses when tracking in-progress functions to reduce lock contention. Performance profiles after these changes show that our current "hot-lock" issue is addressed.

PR #7927, opened 2021-12-01
Performance: MetricEventManager _runningFunctions to ConcurrentDictionary

NickCraver:
Part of #7908 - `_runningFunctions` currently causes locking under load especially when a metrics flush happens to cull the list. I caught up on history with @mathewc and @brettsam here. It seems a combination of `ConcurrentDictionary` semantics and previous circular refs were the reasoning for going this way in the past. The last change to this was in #6547, but since then `ConcurrentDictionary` has improved and we can further improve the mechanism by not using a string allocation on the key front. Since an `InvocationId` is already unique we can key off it and opportunistically remove it.

Overall: this moves back to the original storage structure with a tuned key approach now that .NET has changed the locking bits internally (since 5.x - see https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-5/#collections) and it's no longer a hot path in profiling.

Note that I kept the culling for safety as the `List<T>` approach did - which ultimately may have been one cause of the previous memory leaks - an incomplete function could manifest as a hung reference. This may be impossible now allowing us to remove it, but it's not causing harm and is of minimal cost to keep the security here rather than relying on perfection in the pipe.

Resolves a component of #7908

### Pull request checklist


brettsam:
I went back and looked -- I was wrong about why we moved away from `ConcurentDictionary` here. What I was remembering was some issues we saw way back with `ConcurrentBag` (see here -- https://github.com/Azure/azure-webjobs-sdk/pull/1971#issue-371703839).

src/WebJobs.Script.WebHost/Diagnostics/MetricsEventManager.cs
```diff
@@ -462,30 +460,32 @@ internal void Flush()
...
+                foreach (var possiblyRunning in _runningFunctions)
                 {
-                    // effectively we're pruning all the completed invocations here
-                    _runningFunctions = _runningFunctions.Where(p => !p.Completed).ToList();
-
-                    // create a snapshot within the lock so we can enumerate below
-                    runningFunctionsSnapshot = _runningFunctions.ToArray();
+                    if (possiblyRunning.Value.Completed)
+                    {
+                        _runningFunctions.TryRemove(possiblyRunning.Key, out _);
```
brettsam, line 468: Does this not throw a `Collection was modified...` exception?

NickCraver: Nope! We're safe since it's a concurrent collection now. This safety mechanism is even _likely_ to race as the metrics flush will often overlap a function's own end statement but that race should be safe and not cause any issues. Please sanity check my thinking there!

brettsam: Yup -- you're right! https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.getenumerator?view=net-6.0#remarks

src/WebJobs.Script.WebHost/Diagnostics/MetricsEventManager.cs
```diff
@@ -462,30 +460,32 @@ internal void Flush()
...
+                    }
                 }
 
-                // we calculate concurrency here based on count, since these events are raised
+                // If not for the length, we could probably avoid this array allocation
+                var concurrency = _runningFunctions.Count;
```
danielmarbach: Doesn't count acquire all the locks? Since you have already iterated through the collection above why not simply count there and than use that as the concurrency value or does the value need to be "precise" at the moment when you take it possibly including things that have been added in the meantime? I guess since there is no atomicity between those two iterations anyway and you take a snapshot of count counting above would be sufficient

src/WebJobs.Script.WebHost/Diagnostics/MetricsEventManager.cs
```diff
@@ -462,30 +460,32 @@ internal void Flush()
...
+
+                // We calculate concurrency here based on count, since these events are raised
                 // on a background thread, so we want the actual count for this interval, not
                 // the current count.
                 var currentTime = DateTime.UtcNow;
-                foreach (var runningFunction in runningFunctionsSnapshot)
+                foreach (var runningFunction in _runningFunctions.Values)
```
danielmarbach: Values also acquires the locks and creates a new collection/snapshot. Is that necessary, or could another enumeration be good enough?

NickCraver: We could get more complicated here, but we'd also quickly get into diminishing returns vs. maintenance of something more complex, IMO. While `GetValues()` under the hood does a lock acquisition and snapshot, we're generally talking about very few items here. The cost amortization is also important - this runs once every `n` seconds to dump based on config but with the new cleanup only contains the running values (+/- instantaneous delta inside any method), then we're only locking _once per cycle_ rather than a lock on every op (where we see a great deal of contention in profiling).

Talking through this though - we _could_ change it up. We don't need _another_ enumeration though, we've already got one above. If we're creating the same list of items _anyway_ we could just pretend we're microseconds earlier and go with an `else` in the loop above, creating our "running" collection in the existing enumeration. It'd eat the same number of resizes in the list (still relatively rarely in the overall scheme) as the current `.Values` call.

NickCraver: @danielmarbach updated :)

src/WebJobs.Script.WebHost/Diagnostics/MetricsEventManager.cs
```diff
@@ -462,30 +460,37 @@ internal void Flush()
...
                     return;
                 }
 
                 // We only need to raise events here for functions that aren't completed.
                 // Events are raised immediately for completed functions elsewhere.
-                FunctionStartedEvent[] runningFunctionsSnapshot = null;
-                lock (_runningFunctionsSyncLock)
+                // Loop through and prune any completed runs
+                // Also collects the currently running functions in the same enumeration to minimize cost.
+                var running = new List<FunctionStartedEvent>();
```
danielmarbach, line 463: Depending how hot this path is the list could be Pooled. Hard to say though without more context of that is worth it

fabiocav (approved):

src/WebJobs.Script.WebHost/Diagnostics/MetricsEventManager.cs
```diff
@@ -333,15 +333,14 @@ private class FunctionActivityTracker : IDisposable
...
-            private readonly object _runningFunctionsSyncLock = new object();
             private readonly Timer _activityTimer;
             private readonly ILogger<MetricsEventManager> _logger;
 
             private ulong _totalExecutionCount = 0;
             private int _activeFunctionCount = 0;
             private int _functionActivityFlushInterval;
             private ConcurrentQueue<FunctionMetrics> _functionMetricsQueue = new ConcurrentQueue<FunctionMetrics>();
-            private List<FunctionStartedEvent> _runningFunctions = new List<FunctionStartedEvent>();
+            private ConcurrentDictionary<Guid, FunctionStartedEvent> _runningFunctions = new ConcurrentDictionary<Guid, FunctionStartedEvent>();
```
mathewc, line 333: We had intentionally moved away from a ConcurrentDictionary due to hot lock issues in performance investigations. The changes we made resolved that. My concern is that we may regress here. The relevant PR was https://github.com/Azure/azure-functions-host/pull/6547 where we had moved from ConcurrentDictionary to a List

NickCraver: Yep agreed - I dug into this more as well and tried to add detail in the PR description. The locking internals have changed since when those changes have made - given we're on .NET 5+ this is a much less contentious path and results in more throughput.

I'd love to get the anon/benchmark bits in so we can add Crank data to each of the PRs here - thoughts on https://github.com/Azure/azure-functions-host/pull/7941#issuecomment-986947273? It'd make it much easier to provide numbers for each, though they're all small and add-up I totally agree it's worthwhile.

NickCraver:
Superseded by #7936 (requires both to really test and benchmark the changes since we have 2 overlapping locks here).

PR #7928, opened 2021-12-01
Performance: HostAvailabilityCheckMiddleware: elide async when possible

NickCraver:
Part of #7908. The common case path through `HostAvailabiltyCheckMiddleware` _that we care about_ is when the host is ready and serving traffic - until then its impact is of minimal concern (and we indeed need to await availability anyway).

This change elides the async/await and state machine overhead in the common case once the host is ready using the local function pattern. If preferred we could make this another method - no preference at all, whatever the team is most comfortable with maintaining here.

Resolves a component of #7908

### Pull request checklist

### Additional PR information

Reminder to review this with whitespace off for a much easier time!

### DIFF
--- a/src/WebJobs.Script.WebHost/Middleware/HostAvailabilityCheckMiddleware.cs
+++ b/src/WebJobs.Script.WebHost/Middleware/HostAvailabilityCheckMiddleware.cs
@@ -25,37 +25,44 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Middleware
             _scriptHostManager = scriptHostManager;
         }
 
-        public async Task Invoke(HttpContext httpContext)
+        public Task Invoke(HttpContext httpContext)
         {
-            if (_scriptHostManager.State != ScriptHostState.Offline)
+            // Slow path with async/await for spin-up only
+            static async Task InvokeAwaited(HttpContext context, RequestDelegate next, ILogger<HostAvailabilityCheckMiddleware> logger, IScriptHostManager scriptHostManager)
             {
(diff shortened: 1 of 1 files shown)


src/WebJobs.Script.WebHost/Middleware/HostAvailabilityCheckMiddleware.cs
```diff
@@ -25,37 +25,44 @@ public HostAvailabilityCheckMiddleware(RequestDelegate next, ILoggerFactory logg
             _scriptHostManager = scriptHostManager;
         }
 
-        public async Task Invoke(HttpContext httpContext)
+        public Task Invoke(HttpContext httpContext)
         {
-            if (_scriptHostManager.State != ScriptHostState.Offline)
+            // Slow path with async/await for spin-up only
+            static async Task InvokeAwaited(HttpContext context, RequestDelegate next, ILogger<HostAvailabilityCheckMiddleware> logger, IScriptHostManager scriptHostManager)
```
brettsam: I'd personally like to see this as a separate method... it's typically what we do throughout the codebase although there's no real concrete rules we follow.

brettsam (dismissed):

src/WebJobs.Script.WebHost/Middleware/HostAvailabilityCheckMiddleware.cs
NickCraver: Absolutely - changed to a private method and tried to make clear with comments, would appreciate another pass!

brettsam (approved):

PR #7931, opened 2021-12-01
Reset cached secrets on SyncTriggers (#7698)

### LINKED ISSUE #7698: Functions Keys stale after rotating storage account until startup cache invalidated.

mathewc:
Resolves https://github.com/Azure/azure-functions-host/issues/7698. In addition, the changes here pave the way for addressing https://github.com/Azure/azure-functions-host/issues/5922 because a guard pattern is introduced for secret manager creation. I've improved the current state of things here for that issue a bit in this PR already by checking before accessing SecretManager in FunctionsSyncManager.

I'll be backporting this to 3.x as well.

### Pull request checklist

    * [x] Otherwise: Backport PR https://github.com/Azure/azure-functions-host/pull/7947

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
@@ -106,6 +106,8 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
             {
                 await _syncSemaphore.WaitAsync();
 
+                PrepareSyncTriggers();
+
                 var hashBlobClient = await GetHashBlobAsync();
                 if (isBackgroundSync && hashBlobClient == null && !_environment.IsKubernetesManagedHosting())
                 {
@@ -158,6 +160,31 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
             }
 
             return result;
+        }
+
+        /// <summary>
+        /// SyncTriggers is performed whenever deployments or other changes are made to the application.
+        /// There are some operations we want to perform whenever this happens.
+        /// </summary>
+        private void PrepareSyncTriggers()
+        {
+            // We clear cache to ensure that secrets are reloaded. This is important because secrets are part
+            // of the StartupContext payload (see StartupContextProvider) and that payload comes from the
+            // SyncTriggers operation. So there's a chicken and egg situation here. Consider the following scenario:
+            //   - app is using blob storage for keys
+            //   - a SyncTriggers operation has happened previously and the StartupContext has key info
+            //   - app instances initialize keys from StartupContext (keys aren't loaded from storage)
+            //   - user updates the app to use a new storage account
+            //   - a SyncTriggers operation is performed
+            //   - the app initializes from StartupContext, and **previous old key info is loaded**
+            //   - the SyncTriggers operation uses this old key info, so trigger cache is never updated with new key info
+            //   - Portal/ARM APIs will continue to show old key info.
+            // By clearing cache, we ensure that this host instance reloads keys when they're requested, and the SyncTriggers
+            // operation will contain current keys.
+            if (_secretManagerProvider.SecretsEnabled)
+            {
+                _secretManagerProvider.Current.ClearCache();
+            }
         }
 
         internal static bool IsSyncTriggersEnvironment(IScriptWebHostEnvironment webHostEnvironment, IEnvironment environment)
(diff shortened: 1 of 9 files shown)


src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -106,6 +106,8 @@ public async Task<SyncTriggersResult> TrySyncTriggersAsync(bool isBackgroundSync
             {
                 await _syncSemaphore.WaitAsync();
 
+                PrepareSyncTriggers();
```
mathewc, line 109: I'm basically reverting the revert done in https://github.com/Azure/azure-functions-host/pull/7915/files here.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -160,6 +162,31 @@ public async Task<SyncTriggersResult> TrySyncTriggersAsync(bool isBackgroundSync
...
+            //   - a SyncTriggers operation has happened previously and the StartupContext has key info
+            //   - app instances initialize keys from StartupContext (keys aren't loaded from storage)
+            //   - user updates the app to use a new storage account
+            //   - a SyncTriggers operation is performed
+            //   - the app initializes from StartupContext, and **previous old key info is loaded**
+            //   - the SyncTriggers operation uses this old key info, so trigger cache is never updated with new key info
+            //   - Portal/ARM APIs will continue to show old key info.
+            // By clearing cache, we ensure that this host instance reloads keys when they're requested, and the SyncTriggers
+            // operation will contain current keys.
+            if (_secretManagerProvider.SecretsEnabled)
```
mathewc, line 184: This SecretsEnabled check addresses the issue we had requiring the revert. This will return false for static web apps and other scenarios where secret storage isn't configured.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -316,45 +343,48 @@ public async Task<SyncTriggersPayload> GetSyncTriggersPayload()
...
             }
 
-            // Add functions secrets to the payload
-            // Only secret types we own/control can we cache directly
-            // Encryption is handled by Antares before storage
-            var secretsStorageType = _environment.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebJobsSecretStorageType);
-            if (string.IsNullOrEmpty(secretsStorageType) ||
-                string.Compare(secretsStorageType, "files", StringComparison.OrdinalIgnoreCase) == 0 ||
-                string.Compare(secretsStorageType, "blob", StringComparison.OrdinalIgnoreCase) == 0)
+            if (_secretManagerProvider.SecretsEnabled)
```
mathewc, line 346: Bonus check which should improve things for static web apps if an explicit sync triggers is performed.

src/WebJobs.Script.WebHost/Security/KeyManagement/DefaultSecretManagerProvider.cs
```diff
@@ -45,75 +46,147 @@ public sealed class DefaultSecretManagerProvider : ISecretManagerProvider
...
+                    repository = new BlobStorageSecretsRepository(Path.Combine(_options.CurrentValue.SecretsPath, "Sentinels"),
+                                                                    ConnectionStringNames.Storage,
+                                                                    siteSlotName,
+                                                                    _loggerFactory.CreateLogger<BlobStorageSecretsRepository>(),
+                                                                    _environment,
+                                                                    _azureBlobStorageProvider);
+                }
+            }
+
+            if (repository == null)
```
balag0, line 128: I am curious what it would look like if instead of throwing we log and return null here.
with this current change, it is up to callers to first check SecretsEnabled before invoking and it might easy to miss when writing new code  #Resolved

mathewc: Returning null here would still result in errors upstream, since when a caller accesses/uses ISecretManagerProvider.Current to manage secrets, they expect to get a working instance. Even if we were to move to a model where we created a NullSecretManager that nooped, I don't think that would be the right solution, since keys APIs and other code would return success but not actually do anything. If we're going to support "keyless", then we'll need to instrument places in the code that use keys and handle the lack appropriately. E.g. keys APIs might return an appropriate error code.

So the changes here keep the existing design where any usage of ISecretManagerProvider.Current expects that a working manager can be created/returned.

balag0 (approved):

src/WebJobs.Script.WebHost/Security/KeyManagement/DefaultSecretManagerProvider.cs
fabiocav: Thank you, @mathewc . Yes, changing this behavior here would introduce the problems mentioned and a change in behavior not expected in existing code paths, in addition to an API that is harder to work with and more prone to misuse.

PR #7937, opened 2021-12-02
Performance: Move FunctionLoader to Interlocked

NickCraver:
Part of #7908. This simplifies the locking pattern in `FunctionLoader` since ultimately the set of operations we're doing is exchanging and disposing. There is a subtle assumption here in that a specific `FunctionValueLoader<T>` being disposed _while we create another_ is okay to overlap - I believe that assumption to be valid but more eyes should validate.

Overall, this is a 2-3% RPS gain under load (and much more so in the latency department as it is contention-based).

Resolves a component of #7908

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script/Description/FunctionLoader.cs
+++ b/src/WebJobs.Script/Description/FunctionLoader.cs
@@ -11,7 +11,6 @@ namespace Microsoft.Azure.WebJobs.Script.Description
 {
     internal sealed class FunctionLoader<T> : IDisposable
     {
-        private readonly ReaderWriterLockSlim _functionValueLoaderLock = new ReaderWriterLockSlim();
         private readonly Func<CancellationToken, Task<T>> _valueFactory;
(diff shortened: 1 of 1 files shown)


fabiocav (approved):
Your assumption is correct, it should be safe to dispose while a new loader is being created.

Curious about your test scenarios for this; are you mainly validating CSX scenarios?

NickCraver:
@fabiocav Indeed! But that's only because it's what we have in the repo and am focusing on throughput - I'd love to get benchmarks for `.csx` and a ReadyToRun style .NET function in the repo so we can test those (out of proc should already be good). I figure my next 2 areas to poke at are:
1. Out of proc scenario (probably Python or JavaScript or both)
2. Cold start, just to see if I can spot anything new on top of current work.

fabiocav:
Sounds good. Just wanted to make sure the in-proc, R2R enabled, pre-compiled scenarios are on the radar :)

PR #7939, opened 2021-12-02
Performance: Memorize FunctionInstanceLogger data format strings

NickCraver:
Part of #7908. Today we format a string per call but the number of permutations of these strings is pretty low - we can cache/lookup and do better than the GC cost here with a lookup (in the 1-3% range depending on load).

This can be a bit faster by moving to a `readonly record struct` key (due to faster hashing by ~33%), but since StyleCop is broken with that - it'll need to wait on a 1.2.x upgrade. Let's get the win now and improve it more later.

Resolves a component of #7908

### Pull request checklist

### Additional information

I'm not concerned with the supserset of memory usage here since the cardinality should be pretty finite, but is this a valid assumption? It warrants checking by those with more context to make sure that's correct.


NickCraver:
Branch error on secrets - fixed via #7942 branch.

PR #7940, opened 2021-12-02
Performance: Utility.TryGetFunctionName: search for current function key first

NickCraver:
Part of #7908. In talking with @mathewc and @paulbatum, we're not entirely sure the other keys are even needed - presumably they were at some point of compat along the way. Regardless, we don't have to take a risk here - we can optimize at least a little by moving the most common hit first and saving 2 lookups but still fallback if needed.

Note: the cleanup is still much better in the KeyValuePair case to execute because it's run in a loop to match (or miss) and we're doing an extra 2 case-insensitive searches per loop.

Are these fallbacks still needed for some case I can't find in the current solutions? (some extensibility point?)

Resolves a component of #7908

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script/Utility.cs
+++ b/src/WebJobs.Script/Utility.cs
@@ -506,9 +506,9 @@ namespace Microsoft.Azure.WebJobs.Script
 
             object scopeValue = null;
             if (scopeProps != null && scopeProps.Count > 0 &&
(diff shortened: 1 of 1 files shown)


NickCraver:
Doh, GitHub switched back to creating off the fork - fixed in #7941

PR #7941, opened 2021-12-02
Performance: Utility.TryGetFunctionName: search for current function key first

NickCraver:
Part of #7908. In talking with @mathewc and @paulbatum, we're not entirely sure the other keys are even needed - presumably they were at some point of compat along the way. Regardless, we don't have to take a risk here - we can optimize at least a little by moving the most common hit first and saving 2 lookups but still fallback if needed.

Note: the cleanup is still much better in the KeyValuePair case to execute because it's run in a loop to match (or miss) and we're doing an extra 2 case-insensitive searches per loop.

Are these fallbacks still needed for some case I can't find in the current solutions? (some extensibility point?)

Resolves a component of #7908

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script/Utility.cs
+++ b/src/WebJobs.Script/Utility.cs
@@ -506,9 +506,9 @@ namespace Microsoft.Azure.WebJobs.Script
 
             object scopeValue = null;
             if (scopeProps != null && scopeProps.Count > 0 &&
-                (scopeProps.TryGetValue("functionName", out scopeValue) ||
-                 scopeProps.TryGetValue(LogConstants.NameKey, out scopeValue) ||
-                 scopeProps.TryGetValue(ScopeKeys.FunctionName, out scopeValue)) && scopeValue != null)
+                (scopeProps.TryGetValue(ScopeKeys.FunctionName, out scopeValue) ||
+                 scopeProps.TryGetValue("functionName", out scopeValue) ||
+                 scopeProps.TryGetValue(LogConstants.NameKey, out scopeValue)) && scopeValue != null)
             {
                 functionName = scopeValue.ToString();
             }
@@ -518,9 +518,9 @@ namespace Microsoft.Azure.WebJobs.Script
 
         public static bool IsFunctionName(KeyValuePair<string, object> p)
         {
-            return string.Equals(p.Key, "functionName", StringComparison.OrdinalIgnoreCase) ||
-                string.Equals(p.Key, LogConstants.NameKey, StringComparison.OrdinalIgnoreCase) ||
-                string.Equals(p.Key, ScopeKeys.FunctionName, StringComparison.OrdinalIgnoreCase);
+            return string.Equals(p.Key, ScopeKeys.FunctionName, StringComparison.OrdinalIgnoreCase) ||
+                string.Equals(p.Key, "functionName", StringComparison.OrdinalIgnoreCase) ||
+                string.Equals(p.Key, LogConstants.NameKey, StringComparison.OrdinalIgnoreCase);
         }
 
         public static string GetValueFromScope(IDictionary<string, object> scopeProperties, string key)


liliankasem (approved):
lgtm, can't say I have enough context to know if the fallbacks are needed. Maybe @mathewc or @brettsam have more insight here!

mathewc:
Yes, the other fallbacks are needed. If you search the code, you should find that there are places where those values are added. I'm good with the reordering of ScopeKeys.FunctionName first, since [BeginFunctionScope](https://github.com/Azure/azure-webjobs-sdk/blob/8c90eb9bf2f6562cdd82463371fa38f427206315/src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/LoggerExtensions.cs#L234) adds this for all invocations.

Regarding these perf PRs - if possible it'd be helpful to also include Crank and/or Benchmark test results for each of the changes so we can get a sense of the improvements, e.g. like in PRs like [this](https://github.com/Azure/azure-functions-host/pull/6840) or [this](https://github.com/Azure/azure-functions-host/pull/6764).

mathewc (approved):

NickCraver:
@mathewc I'd like to get them on every PR as well - right now it's a lot of overhead though (something I'd like to tackle as a set of PRs specifically).

The main barrier at the moment is _I can't Crank test this branch_, because of 2 things: throttling in `host.json` and the authentication in the primary `HttpTrigger`. To baseline anything, I need to make basically this commit: https://github.com/Azure/azure-functions-host/commit/9146f760fcc7775c4ea5381cc563ef3159a0f7b8

What do you think of adding another sample that's "Benchmarking" or some such which means we could test against any branch? I believe that or some similar alternative where we don't need to make a baseline change to benchmark is necessary to drastically lower the barrier here. I was creating 2 branches (one for benchmarking and 1 for PR) side-by-side but splitting these out that's currently maintaining ~40 branches and adds up to a lot :( If we're open to adding some sort of baseline hello world benchmark to the dev branch, it'd be _very_ easy to add Crank to all of the PRs. Thoughts? Happy to open an issue on this if easier - wasn't sure the best spot to discuss.

cc @brettsam @fabiocav @liliankasem for thoughts too ^

liliankasem:
> @mathewc I'd like to get them on every PR as well - right now it's a lot of overhead though (something I'd like to tackle as a set of PRs specifically).
> 
> The main barrier at the moment is _I can't Crank test this branch_, because of 2 things: throttling in `host.json` and the authentication in the primary `HttpTrigger`. To baseline anything, I need to make basically this commit: [9146f76](https://github.com/Azure/azure-functions-host/commit/9146f760fcc7775c4ea5381cc563ef3159a0f7b8)
> 
> What do you think of adding another sample that's "Benchmarking" or some such which means we could test against any branch? I believe that or some similar alternative where we don't need to make a baseline change to benchmark is necessary to drastically lower the barrier here. I was creating 2 branches (one for benchmarking and 1 for PR) side-by-side but splitting these out that's currently maintaining ~40 branches and adds up to a lot :( If we're open to adding some sort of baseline hello world benchmark to the dev branch, it'd be _very_ easy to add Crank to all of the PRs. Thoughts? Happy to open an issue on this if easier - wasn't sure the best spot to discuss.
> 
> cc @brettsam @fabiocav @liliankasem for thoughts too ^

I think we might have another sample that's already using anonymous auth, maybe the python one? If that doesn't work for this use case then I think a benchmarking sample is a good idea. Unless @fabiocav knows a good way for us to grab the function key to use for testing :)

NickCraver:
@liliankasem I think Python won't work _for these numbers_ because we're in/out of proc. Out-of-proc is the next scenario I want to poke at though, so that's great if we have a path there already.

I created a branch for such a proposal here - effectively the CSharp HttpTrigger only but in anon/benchmarkable form: https://github.com/Azure/azure-functions-host/compare/craver/benchmark-sample?expand=1 If we're okay with such a thing, I'll happily PR and update all other PR branches and get some numbers in them :)

NickCraver:
Opened a PR for the above - will merge this on in to get the list down but do want to add benchmarks for all PRs we can.

PR #7942, opened 2021-12-02
Performance: Memorize FunctionInstanceLogger data format strings

NickCraver:
Part of #7908. Today we format a string per call but the number of permutations of these strings is pretty low - we can cache/lookup and do better than the GC cost here with a lookup (in the 1-3% range depending on load).

This can be a bit faster by moving to a `readonly record struct` key (due to faster hashing by ~33%), but since StyleCop is broken with that - it'll need to wait on a 1.2.x upgrade. Let's get the win now and improve it more later.

Resolves a component of #7908

### Pull request checklist

### Additional information

I'm not concerned with the supserset of memory usage here since the cardinality should be pretty finite, but is this a valid assumption? It warrants checking by those with more context to make sure that's correct.

### DIFF
--- a/src/WebJobs.Script.WebHost/Diagnostics/FunctionInstanceLogger.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/FunctionInstanceLogger.cs
@@ -105,7 +105,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
             string eventName = success ? MetricEventNames.FunctionInvokeSucceeded : MetricEventNames.FunctionInvokeFailed;
             string functionName = function != null ? function.Name : string.Empty;
 
-            // This has low cadinality but we allocate the string every call (even though ironically it's often not used due to rollups)
+            // This has low cardinality but we allocate the string every call (even though ironically it's often not used due to rollups)
             // It's cheaper to cache and lookup rather than generate and pay for GC here.
             // Note: this would be faster with a readonly record struct key, but StyleCop is angry with it until an upcoming 1.2.x beta release
             var key = (startedEvent.FunctionMetadata.Language, functionName, success, Stopwatch.IsHighResolution);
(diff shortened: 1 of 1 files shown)


src/WebJobs.Script.WebHost/Diagnostics/FunctionInstanceLogger.cs
```diff
@@ -103,7 +104,16 @@ private void EndFunction(FunctionInstanceLogEntry item)
             var function = startedEvent.FunctionMetadata;
             string eventName = success ? MetricEventNames.FunctionInvokeSucceeded : MetricEventNames.FunctionInvokeFailed;
             string functionName = function != null ? function.Name : string.Empty;
-            string data = string.Format(Microsoft.Azure.WebJobs.Script.Properties.Resources.FunctionInvocationMetricsData, startedEvent.FunctionMetadata.Language, functionName, success, Stopwatch.IsHighResolution);
+
+            // This has low cadinality but we allocate the string every call (even though ironically it's often not used due to rollups)
```
danielmarbach:
```suggestion
            // This has low cardinality but we allocate the string every call (even though ironically it's often not used due to rollups)
```

src/WebJobs.Script.WebHost/Diagnostics/FunctionInstanceLogger.cs
```diff
@@ -103,7 +104,16 @@ private void EndFunction(FunctionInstanceLogEntry item)
...
+            // It's cheaper to cache and lookup rather than generate and pay for GC here.
+            // Note: this would be faster with a readonly record struct key, but StyleCop is angry with it until an upcoming 1.2.x beta release
+            var key = (startedEvent.FunctionMetadata.Language, functionName, success, Stopwatch.IsHighResolution);
```
danielmarbach, line 111: Out of curiosity. From the grand scheme of things, does this really pay off in comparison to for example trying to use a faster way of allocating the string with for example `string.Create`? Now this requires a not so small ValueTuple and a dictionary lookup which also comes with runtime costs.

NickCraver: Good question! Yes it pays off, because that string we're allocating below is _far bigger_, and it needs to be garbage collected later - to boot, it's often not even used.

This is one of those things were "if you do X a lot, even in gen0, it isn't free" - that's still cost to cleanup the mess later. I'd like to get load tests as part of PRs in general, so that amortized cost of GC is measured and recognized.

brettsam (approved):

fabiocav (approved):

PR #7943, opened 2021-12-02
Performance: Eliminate Stopwatch allocations

NickCraver:
Part of #7908. This approach eliminates the Stopwatch allocations every function call (we allocate several per request). For consistency, this uses `ValueStopwatch` as seen in other parts (internally) in the framework, e.g. https://github.com/dotnet/runtime/blob/main/src/libraries/Common/src/Extensions/ValueStopwatch/ValueStopwatch.cs. This approach allows us to add an analyzer down the road to prevent usage of `Stopwatch` and keep us lean on `ValueStopwatch`.

For context: ValueStopwatch keeps track of exactly 1 thing: when it we started. Anything after that is "how long" - since it's only a value of "started ticks" there's no restart (which wouldn't do what you may think on a value type anyway), all work after that is "how long since?" - so on restarts if ever needed just throw it away and replace (as see in the PR here - see `WorkerConcurrencyManager` for an example delta).

Overall, this isn't a huge gain, but: papercuts! I'm poking at what we're allocating every function call and getting that down overall reducing our time both doing the initial work and in GC pauses later.

Leaving this tagged as needs-discussion since this is a code/approach change and not a one-off optimization. Hopefully agreeable but can change course if not.

Resolves a component of #7908

### Pull request checklist

#### Crank comparison
| application                             | dev-sw-1    | stopwatch-6 |         |
| Working Set (MB)                        |         364 |         377 |  +3.57% |
| Private Memory (MB)                     |       1,102 |       1,150 |  +4.36% |
| Build Time (ms)                         |      17,373 |      18,348 |  +5.61% |
| Start Time (ms)                         |       4,768 |       4,770 |  +0.04% |
| Published Size (KB)                     |     674,686 |     674,687 |  +0.00% |
| .NET Core SDK Version                   |     6.0.101 |     6.0.101 |         |
| Max Working Set (MB)                    |         379 |         394 |  +3.96% |
| Max GC Heap Size (MB)                   |         185 |         138 | -25.41% |
| Size of committed memory by the GC (MB) |         191 |         211 | +10.47% |
| Max Number of Gen 0 GCs / sec           |       14.00 |       12.00 | -14.29% |
| Max Time in GC (%)                      |        7.00 |        6.00 | -14.29% |
| Max Gen 0 Size (B)                      |  38,895,824 |  33,088,472 | -14.93% |
| Max Gen 1 Size (B)                      |  10,841,312 |  10,549,832 |  -2.69% |
| Max Gen 2 Size (B)                      |  28,461,904 |  26,697,336 |  -6.20% |
| Max LOH Size (B)                        |   8,830,720 |  10,108,048 | +14.46% |
| Max Allocation Rate (B/sec)             | 335,214,208 | 334,253,624 |  -0.29% |
| Max GC Heap Fragmentation               |          70 |          69 |  -2.61% |
| Max Exceptions (#/s)                    |         249 |         275 | +10.44% |
| Max Lock Contention (#/s)               |         127 |         122 |  -3.94% |
| Max ThreadPool Threads Count            |          30 |          32 |  +6.67% |
| Max ThreadPool Queue Length             |          94 |          93 |  -1.06% |
| Max ThreadPool Items (#/s)              |      36,332 |      36,133 |  -0.55% |
| Max Active Timers                       |          86 |          94 |  +9.30% |
| IL Jitted (B)                           |   1,054,476 |   1,054,978 |  +0.05% |
| Methods Jitted                          |      10,881 |      10,884 |  +0.03% |
| Avg Allocation Rate (B/sec)             | 297,712,948 | 302,074,347 |  +1.46% |
| 90th Allocation Rate (B/sec)            | 329,426,128 | 330,418,592 |  +0.30% |

| load                   | dev-sw-1 | stopwatch-6 |         |
| ---------------------- | -------- | ----------- | ------- |
| CPU Usage (%)          |        1 |           1 |   0.00% |
| Cores usage (%)        |       37 |          38 |  +2.70% |
| Working Set (MB)       |       37 |          37 |   0.00% |
| Private Memory (MB)    |      357 |         357 |   0.00% |
| Start Time (ms)        |        0 |           0 |         |
| First Request (ms)     |      340 |         339 |  -0.29% |
| Requests/sec           |    8,055 |       8,094 |  +0.49% |
| Requests               |  484,085 |     485,996 |  +0.39% |
| Mean latency (ms)      |    13.92 |       13.75 |  -1.22% |
| Max latency (ms)       |   198.99 |      167.93 | -15.61% |
| Bad responses          |        0 |           0 |         |
| Socket errors          |        0 |           0 |         |
| Read throughput (MB/s) |     1.24 |        1.24 |   0.00% |
| Latency 50th (ms)      |    10.04 |       10.02 |  -0.20% |
| Latency 75th (ms)      |    15.90 |       15.70 |  -1.26% |
| Latency 90th (ms)      |    28.15 |       27.70 |  -1.60% |
| Latency 99th (ms)      |    64.70 |       62.60 |  -3.25% |


src/WebJobs.Script/Utility.cs
```diff
@@ -974,6 +977,15 @@ public static FunctionAppContentEditingState GetFunctionAppContentEditingState(I
             }
         }
 
+        /// <summary>
+        /// Gets a timestamp given start ticks (from <see cref="Stopwatch.GetTimestamp"/>) without allocating a Stopwatch.
+        /// </summary>
+        public static TimeSpan GetDuration(long startTicks)
```
sebastienros: Why not have it return milliseconds directly instead of creating a `TimeSpan` just to read the `TotalMilliseconds` ?

NickCraver: Good question! This was for consistency if we went with one thing across projects, because of the direct use of `TimeSpan` here: https://github.com/Azure/azure-webjobs-sdk/pull/2794/files#diff-673ae62ecb874d5346bc217f774d055a5da4728f418a2e169028d205428dd020R149 - I'd appreciate any thoughts!

sebastienros: I think the main value is to encapsulate the call to `GetTimeStamp()` and the conversion between ticks and ms. So maybe create a `TimeSpan` from the caller when you need one, with `new TimeSpan(GetDuration(ticks))`.

But it's not much, didn't realize it was used somewhere else.

danielmarbach:
FYI Azure SDK also uses the ValueStopWatch approach. I contributed that a while ago to the core

https://github.com/Azure/azure-sdk-for-net/blob/4162f6fa2445b2127468b9cfd080f01c9da88eba/sdk/core/Azure.Core/src/Shared/ValueStopwatch.cs

src/WebJobs.Script/Utility.cs
NickCraver: Getting back to this - pulling in `ValueStopwatch` overall and globally replacing!

src/WebJobs.Script/Diagnostics/ValueStopwatch.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Diagnostics;
+
+namespace Microsoft.Azure.WebJobs.Script.Diagnostics
+{
+    // Based on: https://github.com/dotnet/aspnetcore/blob/master/src/Shared/ValueStopwatch/ValueStopwatch.cs
+    public struct ValueStopwatch
+    {
+        private static readonly double TimestampToTicks = TimeSpan.TicksPerSecond / (double)Stopwatch.Frequency;
+
+        private long _startTimestamp;
```
kshyju, line 15: Should it be `private readonly` ?

src/WebJobs.Script/Diagnostics/ValueStopwatch.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        private ValueStopwatch(long startTimestamp)
+        {
+            _startTimestamp = startTimestamp;
+        }
+
+        public bool IsActive => _startTimestamp != 0;
```
kshyju, line 22: Question out of curiosity: Would there be a case where this prop returns `false`? (based on the comment on line 27)

src/WebJobs.Script/Utility.cs
```diff
@@ -5,6 +5,7 @@
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Collections.Specialized;
+using System.Diagnostics;
```
kshyju: Unused?

kshyju (dismissed):

src/WebJobs.Script/Diagnostics/ValueStopwatch.cs
```diff
@@ -0,0 +1,40 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Diagnostics;
+
+namespace Microsoft.Azure.WebJobs.Script.Diagnostics
+{
+    // Based on: https://github.com/dotnet/aspnetcore/blob/master/src/Shared/ValueStopwatch/ValueStopwatch.cs
+    public struct ValueStopwatch
```
mathewc, line 10: Let's add some basic test coverage for this new type - would be good for it to have it's own test suite

mathewc:
There are many usages of Stopwatch in the WebJobs SDK as well https://github.com/Azure/azure-webjobs-sdk/search?q=Stopwatch. Once we finalize the approach here (looks good to me), we'll want to apply there as well, since those are also accruing to Azure Functions

src/WebJobs.Script/Diagnostics/ValueStopwatch.cs
NickCraver: `readonly` creates a defensive copy in the runtime, so that's why it's left off here - can add a comment to make clear!

src/WebJobs.Script/Diagnostics/ValueStopwatch.cs
NickCraver: 👍 - this may exist already at the source and reusable, either way will get this added

liliankasem (dismissed):

src/WebJobs.Script/Diagnostics/ValueStopwatch.cs
NickCraver: Brought in the runtime tests and added one relevant to some of our usages.

src/WebJobs.Script/Diagnostics/ValueStopwatch.cs
NickCraver: Comment added!

src/WebJobs.Script/Diagnostics/ValueStopwatch.cs
NickCraver: Yep - this is a `struct` to avoid the allocations, so a `default(ValueStopwatch)` would be false and appropriately go boom if ever used. Added in the test suite to illustrate :)

src/WebJobs.Script/Utility.cs
NickCraver: Indeed! From an earlier iteration, thought I got this but must not have, fixed!

brettsam (approved):

NickCraver:
Build failure is another thing - fix pending in #8056

kshyju (approved):

mathewc (approved):

PR #7593, opened 2021-08-09
Adds Metric Logging support for GRPC workers

brandonh-msft:
Host-level fix to support gRPC logging of custom metric types

### Issue describing the changes in this PR

For https://github.com/Azure/azure-functions-dotnet-worker/issues/543

This adds `RpcMetric` as a gRPC message type + support to log it to ILogger in such a way that it shows in Application Insights logs as a custom metric. This provides functionality similar to what we had in in-proc languages by simply using the `.LogMetric()` ILogger extension method.

This PR encompasses the necessary changes required to complete https://github.com/Azure/azure-functions-dotnet-worker/issues/543

### Pull request checklist

* [x] My changes **do not** require documentation changes
    * [ ] Otherwise: Documentation issue linked to PR
* [ ] My changes **should not** be added to the release notes for the next release
    * [x] Otherwise: I've added my notes to `release_notes.md`
* [x] My changes **do not** need to be backported to a previous version
    * [ ] Otherwise: Backport tracked by issue/PR #issue_or_pr
* [x] I have added all required tests (Unit tests, E2E tests)

### Additional information

After completion, https://github.com/Azure/azure-functions-dotnet-worker/pull/548 should be completed to fully implement this functionality.

PR #7956, opened 2021-12-06
Adds Metric Logging support for GRPC workers

liliankasem:
- [ Branch transferred from here: #7593 ]

Host-level fix to support gRPC logging of custom metric types

For https://github.com/Azure/azure-functions-dotnet-worker/issues/543

This adds `RpcMetric` as a gRPC message type + support to log it to ILogger in such a way that it shows in Application Insights logs as a custom metric. This provides functionality similar to what we had in in-proc languages by simply using the `.LogMetric()` ILogger extension method.

This PR encompasses the necessary changes required to complete https://github.com/Azure/azure-functions-dotnet-worker/issues/543

### Pull request checklist

### Additional information

After completion, https://github.com/Azure/azure-functions-dotnet-worker/pull/548 should be completed to fully implement this functionality.

### DIFF
--- a/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
@@ -13,6 +13,7 @@ using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Threading.Tasks.Dataflow;
+using Microsoft.Azure.WebJobs.Logging;
 using Microsoft.Azure.WebJobs.Script.Description;
 using Microsoft.Azure.WebJobs.Script.Diagnostics;
 using Microsoft.Azure.WebJobs.Script.Eventing;
@@ -75,18 +76,18 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
         private System.Timers.Timer _timer;
 
         internal GrpcWorkerChannel(
-           string workerId,
-           IScriptEventManager eventManager,
-           RpcWorkerConfig workerConfig,
-           IWorkerProcess rpcWorkerProcess,
-           ILogger logger,
-           IMetricsLogger metricsLogger,
-           int attemptCount,
-           IEnvironment environment,
-           IOptionsMonitor<ScriptApplicationHostOptions> applicationHostOptions,
-           ISharedMemoryManager sharedMemoryManager,
-           IFunctionDataCache functionDataCache,
-           IOptions<WorkerConcurrencyOptions> workerConcurrencyOptions)
+            string workerId,
+            IScriptEventManager eventManager,
+            RpcWorkerConfig workerConfig,
+            IWorkerProcess rpcWorkerProcess,
+            ILogger logger,
+            IMetricsLogger metricsLogger,
+            int attemptCount,
+            IEnvironment environment,
+            IOptionsMonitor<ScriptApplicationHostOptions> applicationHostOptions,
+            ISharedMemoryManager sharedMemoryManager,
+            IFunctionDataCache functionDataCache,
+            IOptions<WorkerConcurrencyOptions> workerConcurrencyOptions)
         {
             _workerId = workerId;
             _eventManager = eventManager;
@@ -122,7 +123,7 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
                 .Subscribe(async (msg) => await InvokeResponse(msg.Message.InvocationResponse)));
 
             _inboundWorkerEvents.Where(msg => msg.MessageType == MsgType.WorkerStatusResponse)
-               .Subscribe((msg) => ReceiveWorkerStatusResponse(msg.Message.RequestId, msg.Message.WorkerStatusResponse));
+                .Subscribe((msg) => ReceiveWorkerStatusResponse(msg.Message.RequestId, msg.Message.WorkerStatusResponse));
 
             _startLatencyMetric = metricsLogger?.LatencyEvent(string.Format(MetricEventNames.WorkerInitializeLatency, workerConfig.Description.Language, attemptCount));
 
@@ -657,14 +658,29 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
                 // Restore the execution context from the original invocation. This allows AsyncLocal state to flow to loggers.
                 System.Threading.ExecutionContext.Run(context.AsyncExecutionContext, (s) =>
                 {
-                    if (rpcLog.Exception != null)
+                    if (rpcLog.LogCategory == RpcLogCategory.CustomMetric)
                     {
-                        var exception = new Workers.Rpc.RpcException(rpcLog.Message, rpcLog.Exception.Message, rpcLog.Exception.StackTrace);
-                        context.Logger.Log(logLevel, new EventId(0, rpcLog.EventId), rpcLog.Message, exception, (state, exc) => state);
+                        if (rpcLog.PropertiesMap.TryGetValue(LogConstants.NameKey, out var metricName)
+                            && rpcLog.PropertiesMap.TryGetValue(LogConstants.MetricValueKey, out var metricValue))
+                        {
+                            // Strip off the name/value entries in the dictionary passed to Log Message and include the rest as the property bag passed to the backing ILogger
+                            var rpcLogProperties = rpcLog.PropertiesMap
+                                                    .Where(i => i.Key != LogConstants.NameKey && i.Key != LogConstants.MetricValueKey)
+                                                    .ToDictionary(i => i.Key, i => i.Value.ToObject());
+                            context.Logger.LogMetric(metricName.String, metricValue.Double, rpcLogProperties);
+                        }
                     }
                     else
                     {
-                        context.Logger.Log(logLevel, new EventId(0, rpcLog.EventId), rpcLog.Message, null, (state, exc) => state);
+                        if (rpcLog.Exception != null)
+                        {
+                            var exception = new Workers.Rpc.RpcException(rpcLog.Message, rpcLog.Exception.Message, rpcLog.Exception.StackTrace);
+                            context.Logger.Log(logLevel, new EventId(0, rpcLog.EventId), rpcLog.Message, exception, (state, exc) => state);
+                        }
+                        else
+                        {
+                            context.Logger.Log(logLevel, new EventId(0, rpcLog.EventId), rpcLog.Message, null, (state, exc) => state);
+                        }
                     }
                 }, null);
             }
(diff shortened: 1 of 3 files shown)


src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs
```diff
@@ -88,25 +85,23 @@ public static async Task<TypedData> ToRpc(this object value, ILogger logger, Grp
...
+                    break;
+                case double[] arrDouble:
+                    typedData = arrDouble.ToRpcDoubleArray();
+                    break;
+                case long[] arrLong:
+                    typedData = arrLong.ToRpcLongArray();
+                    break;
+                default:
+                    typedData = value.ToRpcDefault();
+                    break;
```
NickCraver: I'm not sure if there's a reason we're pinned at C# 7.2, but if the team is interested things like this and the above get a bit more terse in C# 8 and above, for example this could be:
```cs
internal static TypedData ToRpcCollection(this object value)
{
    return value switch
    {
        byte[][] arrBytes => arrBytes.ToRpcByteArray(),
        string[] arrStr => arrStr.ToRpcStringArray(),
        double[] arrDouble => arrDouble.ToRpcDoubleArray(),
        long[] arrLong => arrLong.ToRpcLongArray(),
        _ => value.ToRpcDefault(),
    };
}
```

This could just be preference as well that I'm just learning - wanted to point it out as I see some things new language features can make easier though!

src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs
```diff
@@ -57,29 +57,26 @@ public static async Task<TypedData> ToRpc(this object value, ILogger logger, Grp
...
+                case double dbl:
+                    typedData.Double = dbl;
+                    break;
+                case HttpRequest request:
+                    typedData = await request.ToRpcHttp(logger, capabilities);
+                    break;
+                default:
+                    typedData = value.GetType().IsArray && IsTypedDataCollectionSupported(capabilities) ? value.ToRpcCollection() : value.ToRpcDefault();
+                    break;
```
NickCraver: Adding to the below, since this is an optimization. I noticed we're building and then discarding the initially created the `typedData` up top in the `HttpRequest` and default cases (existing issue) - we could avoid this allocation by moving the creations inline:
```cs
public static async Task<TypedData> ToRpc(this object value, ILogger logger, GrpcCapabilities capabilities)
{
    return value switch
    {
        null => new TypedData(),
        byte[] arr => new TypedData() { Bytes = ByteString.CopyFrom(arr) },
        JObject jobj => new TypedData() { Json = jobj.ToString(Formatting.None) },
        string str => new TypedData() { String = str },
        double dbl => new TypedData() { Double = dbl },
        HttpRequest request => await request.ToRpcHttp(logger, capabilities),
        _ => value.GetType().IsArray && IsTypedDataCollectionSupported(capabilities)
            ? value.ToRpcCollection()
            : value.ToRpcDefault(),
    };
}
```

src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs
liliankasem: I like how simple that new approach is! @brettsam & @fabiocav might be able to answer regarding being pinned to C# 7.2?

src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs
liliankasem: Switched to using inline creations for TypedData!

NickCraver: Nice! I think we can simplify and improve even more here if you're game - just looked at the new version and noticed this is really a branched switch overall. We can collapse down and simplify if there's no reason not to - for example:

```cs
            switch (value)
            {
                case null:
                    return new TypedData();
                case byte[] arr:
                    return new TypedData() { Bytes = ByteString.CopyFrom(arr) };
                case JObject jobj:
                    return new TypedData() { Json = jobj.ToString(Formatting.None) };
                case string str:
                    return new TypedData() { String = str };
                case double dbl:
                    return new TypedData() { Double = dbl };
                case HttpRequest request:
                    return await request.ToRpcHttp(logger, capabilities);
                case byte[][] arrBytes when IsTypedDataCollectionSupported(capabilities):
                    return arrBytes.ToRpcByteArray();
                case string[] arrStr when IsTypedDataCollectionSupported(capabilities):
                    return arrStr.ToRpcStringArray();
                case double[] arrDouble when IsTypedDataCollectionSupported(capabilities):
                    return arrDouble.ToRpcDoubleArray();
                case long[] arrLong when IsTypedDataCollectionSupported(capabilities):
                    return arrLong.ToRpcLongArray();
                default:
                    return value.ToRpcDefault();
            }
```
...and eliminate the `ToRpcCollection` method doing resolution of type again. Thoughts?

In later versions of C#, that goes all the way down to:
```cs
        public static async Task<TypedData> ToRpc(this object value, ILogger logger, GrpcCapabilities capabilities) => value switch
        {
            null => new TypedData(),
            byte[] arr => new TypedData() { Bytes = ByteString.CopyFrom(arr) },
            JObject jobj => new TypedData() { Json = jobj.ToString(Formatting.None) },
            string str => new TypedData() { String = str },
            double dbl => new TypedData() { Double = dbl },
            HttpRequest request => await request.ToRpcHttp(logger, capabilities),
            byte[][] arrBytes when IsTypedDataCollectionSupported(capabilities) => arrBytes.ToRpcByteArray(),
            string[] arrStr when IsTypedDataCollectionSupported(capabilities) => arrStr.ToRpcStringArray(),
            double[] arrDouble when IsTypedDataCollectionSupported(capabilities) => arrDouble.ToRpcDoubleArray(),
            long[] arrLong when IsTypedDataCollectionSupported(capabilities) => arrLong.ToRpcLongArray(),
            _ => value.ToRpcDefault(),
        };
```
...if people find it readable, up to the team there on these kinds of things, wouldn't want to push them if not interested.

Overall, I think we may want to factor out `HttpRequest` specifically into another method if this is a hot path, because in those cases we're adding `async/await` overhead to a method where only one variant does so - we can potentially save additive overhead there (not proposing such in this PR, just noting something to look at).

liliankasem: Makes sense to me, will update :D I did not even notice those two were related (not my PR originally, just helping someone else's code get merged 🙈🤦🏻‍♀️)

RE: HttpRequest, not sure if it is a hot path or not, but could be worth creating an optimization issue to look into it :)

NickCraver: 👍 To clarify: if `ToRpc()` at all is a hot path (without looking I'd guess "probably") - we're paying that overhead everywhere when we _could_ likely reduce it to only in the `HttpRequest` case - happy to poke when I start looking at out-of-proc if no one else has time first :)

src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs
liliankasem: @NickCraver turns out we're not actually pinned and use latest: https://github.com/Azure/azure-functions-host/commit/a68438b21a65ca7d5c209d4e23bad9f52276ffda

Something is off with common.props in my PR - will fix that then start using this new switch approach :D

NickCraver: Ahhhh, that's what was throwing me - this is baseline behind dev, awesome - good catch!

liliankasem: Thanks to @brettsam ofc :)

NickCraver (approved):
I'm unsure about the logging, could use more experienced eyes but :+1: on the `ToRpc` bits, thanks!

src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs
```diff
@@ -24,93 +24,35 @@ internal static class GrpcMessageConversionExtensions
...
-            else
-            {
-                typedData = value.ToRpcDefault();
-            }
+                    _ => throw new InvalidOperationException($"Unknown RpcDataType: {typedData.DataCase}")
+            };
 
-            return typedData;
-        }
+        public static async Task<TypedData> ToRpc(this object value, ILogger logger, GrpcCapabilities capabilities) =>
```
brettsam, line 41: this isn't part of your change, but this seems like a prime candidate to get rid of `async` here and return `Tasks` directly for perf.

brettsam (approved):

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -657,14 +658,29 @@ internal void Log(GrpcEvent msg)
...
+                                                    .ToDictionary(i => i.Key, i => i.Value.ToObject());
+                            context.Logger.LogMetric(metricName.String, metricValue.Double, rpcLogProperties);
+                        }
                     }
                     else
                     {
-                        context.Logger.Log(logLevel, new EventId(0, rpcLog.EventId), rpcLog.Message, null, (state, exc) => state);
+                        if (rpcLog.Exception != null)
+                        {
+                            var exception = new Workers.Rpc.RpcException(rpcLog.Message, rpcLog.Exception.Message, rpcLog.Exception.StackTrace);
```
fabiocav, line 677: I know this is the current pattern, but `RpcException` is becoming a source of confusion for users, particularly when reviewing .NET exceptions (i.e. .NET isolated). Here's an issue we recently reviewed: https://github.com/Azure/azure-functions-dotnet-worker/issues/370

fabiocav (approved):
We don't need to block on the comment, but I'm flagging as this is something we need to discuss since it impacts the user experience and is causing confusion.

src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs
liliankasem: Taking note to address in another PR

liliankasem:
@brettsam can you run `/backport` on this for me? Will do manually if it fails

PR #7961, opened 2021-12-07
Copied patch logic from dev

yojagad:
This PR fixes build step that uploads artifacts


VpOfEngineering:
Confirmed with Yogesh, this makes sense.

pragnagopa (approved):

PR #7964, opened 2021-12-08
Samples: add a benchmarkable sample

NickCraver:
This would allow us to run Crank benchmarks against the branch without modifications. Currently the `Samples/CSharp` path has throttling limiting throughput as well as authentication, which means effectively we can't pull the `dev` branch and benchmark it. Today to benchmark we need to mutate the baseline branch (removing or changing auth) and do the same to the proposed branch (so that it's not in the proposed PR), meaning we have 3 new branches to baseline 1 against `dev` every time...let's fix that!

It'd be better if we could test authentication with a known key so that we're invoking the authentication path as well, but I'm unsure how to do this (couldn't find it in docs). @fabiocav has ideas - but maybe this can go in as-is and we add the authenticated one beside it so we have benchmarks for both available as neighboring endpoints?

This would unblock me much more easily adding baseline vs. new benchmarks in performance PRs.

### Pull request checklist

### DIFF
--- a/sample/CSharpBenchmark/HttpTriggerAnon/function.json
+++ b/sample/CSharpBenchmark/HttpTriggerAnon/function.json
@@ -1,0 +1,16 @@
+{
+    "bindings": [
+        {
+            "type": "httpTrigger",
+            "authLevel": "anonymous",
+            "name": "req",
+            "direction": "in",
+            "methods": [ "get" ]
+        },
+        {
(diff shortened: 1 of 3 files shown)


liliankasem (approved):

NickCraver:
Adding this in, and a follow-up should be a secrets path for benchmarking right beside it so we'll have both options :)

PR #7969, opened 2021-12-09
Experiment: DryIoc Removal (Orchard Approach)

NickCraver:
This is a WIP PR for test runs and such as we attempt to remove DryIoc from the solution and use the child container approach that Orchard is (based on built-in .NET DI). Ignore the few sample project changes - branch noise that would not go in.

This isn't something we could do without a _lot_ of validation and possibly a major version, but given the performance we're seeing some DryIoc bits and the percentage of time it accounts for in cold start - it's an experiment to explore what gains are on the table.


fabiocav:
@brettsam I'm currently validating service resolution scenarios with this, but it would be good to have you take a look as well.

fabiocav:
@brettsam one of the things we'll need to review are the test scenarios.

NickCraver:
Adding here: the main item of discussion that @fabiocav and I found: how missing services are handled.

The issue is how DryIoc handles this vs. the built-in .NET bits. It's kind of expected (I think) that a missing dependency will throw at constructor call time when you're missing a service. Let's say you have:

```cs
public class MyServiceA 
{
   public MyserviceA(SomeServiceB foo) { }
}
```

If `SomeServiceB` isn't registered in the container, then the service provider builder isn't able to resolve what to feed to the constructor of `MyServiceA`, and it goes boom. This is what the built-in DI does. Evidently, it's _not_ what DryIoc does. It would happily pass a `null` in instead. This subtle but important difference means users could have silently been building up invalid combinations that would be "enforced" by this change.

This ultimate place this happens in .NET DI is in the constructor call builder here: https://github.com/dotnet/aspnetcore/blob/5adbdcb18ce1167321e919b930d575cbbfb7512c/src/Shared/ActivatorUtilities/ActivatorUtilities.cs#L376

This means even the simpler case of a constructor of:
```cs
public class MyServiceA 
{
   public MyserviceA(string foo) { }
}
```
...will fail the same way, unless you registered `string` in your service list. However, default parameters specifically work as expected, e.g.
```cs
public class MyServiceA 
{
   public MyserviceA(string foo = "someDefault") { }
}
```

This probably would have been fine on 4.0 as a change but now...we've gotta figure that out.

Note: this appears even in our tests - we silently haven't had a service registered that's needed, but hasn't bitten us:
```
Message: 
    System.InvalidOperationException : Unable to resolve service for type 'Microsoft.Azure.WebJobs.Script.Workers.SharedMemoryDataTransfer.ISharedMemoryManager' while attempting to activate 'Microsoft.Azure.WebJobs.Script.Grpc.GrpcWorkerChannelFactory'.

  Stack Trace: 
    CallSiteFactory.CreateArgumentCallSites(Type implementationType, CallSiteChain callSiteChain, ParameterInfo[] parameters, Boolean throwIfCallSiteNotFound)
    CallSiteFactory.CreateConstructorCallSite(ResultCache lifetime, Type serviceType, Type implementationType, CallSiteChain callSiteChain)
    CallSiteFactory.TryCreateExact(ServiceDescriptor descriptor, Type serviceType, CallSiteChain callSiteChain, Int32 slot)
    CallSiteFactory.TryCreateExact(Type serviceType, CallSiteChain callSiteChain)
    CallSiteFactory.CreateCallSite(Type serviceType, CallSiteChain callSiteChain)
    CallSiteFactory.CreateArgumentCallSites(Type implementationType, CallSiteChain callSiteChain, ParameterInfo[] parameters, Boolean throwIfCallSiteNotFound)
    CallSiteFactory.CreateConstructorCallSite(ResultCache lifetime, Type serviceType, Type implementationType, CallSiteChain callSiteChain)
    CallSiteFactory.TryCreateExact(ServiceDescriptor descriptor, Type serviceType, CallSiteChain callSiteChain, Int32 slot)
    CallSiteFactory.TryCreateExact(Type serviceType, CallSiteChain callSiteChain)
    CallSiteFactory.CreateCallSite(Type serviceType, CallSiteChain callSiteChain)
    <32 more frames...>
    CallSiteRuntimeResolver.VisitRootCache(ServiceCallSite callSite, RuntimeResolverContext context)
    CallSiteVisitor`2.VisitCallSite(ServiceCallSite callSite, TArgument argument)
    CallSiteRuntimeResolver.Resolve(ServiceCallSite callSite, ServiceProviderEngineScope scope)
    ServiceProvider.CreateServiceAccessor(Type serviceType)
    ConcurrentDictionary`2.GetOrAdd(TKey key, Func`2 valueFactory)
    ServiceProvider.GetService(Type serviceType, ServiceProviderEngineScope serviceProviderEngineScope)
    ServiceProviderServiceExtensions.GetService[T](IServiceProvider provider)
    TestHostExtensions.GetScriptHost(IHost host) line 19
    TestFixture.get_ScriptHost() line 1456
    TestFixture.InitializeAsync() line 1479
```

NickCraver:
For performance numbers - here's a baseline of this change:

| application                             | exp-lin-2-ioc-pre | exp-lin-2-ioc-post |          |
| --------------------------------------- | ----------------- | ------------------ | -------- |
| CPU Usage (%)                           |                 7 |                  7 |    0.00% |
| Cores usage (%)                         |               201 |                201 |    0.00% |
| Working Set (MB)                        |               354 |                456 |  +28.81% |
| Private Memory (MB)                     |             1,175 |              1,277 |   +8.68% |
| Build Time (ms)                         |            18,593 |             69,902 | +275.96% |
| Start Time (ms)                         |             6,818 |             12,005 |  +76.08% |
| Published Size (KB)                     |           674,604 |            674,338 |   -0.04% |
| .NET Core SDK Version                   |           6.0.100 |            6.0.100 |          |
| Max CPU Usage (%)                       |                99 |                 99 |    0.00% |
| Max Working Set (MB)                    |               370 |                478 |  +29.19% |
| Max GC Heap Size (MB)                   |               154 |                277 |  +79.87% |
| Size of committed memory by the GC (MB) |               167 |                281 |  +68.26% |
| Max Number of Gen 0 GCs / sec           |             12.00 |              13.00 |   +8.33% |
| Max Number of Gen 1 GCs / sec           |              5.00 |               5.00 |    0.00% |
| Max Number of Gen 2 GCs / sec           |              1.00 |               1.00 |    0.00% |
| Max Time in GC (%)                      |             13.00 |               9.00 |  -30.77% |
| Max Gen 0 Size (B)                      |        91,262,848 |         58,245,424 |  -36.18% |
| Max Gen 1 Size (B)                      |        15,316,832 |         18,024,208 |  +17.68% |
| Max Gen 2 Size (B)                      |        36,507,472 |         29,568,776 |  -19.01% |
| Max LOH Size (B)                        |        11,025,944 |         12,501,688 |  +13.38% |
| Max Allocation Rate (B/sec)             |       334,968,992 |        322,204,376 |   -3.81% |
| Max GC Heap Fragmentation               |                80 |                 69 |  -13.81% |
| # of Assemblies Loaded                  |               212 |                212 |    0.00% |
| Max Exceptions (#/s)                    |               255 |                258 |   +1.18% |
| Max Lock Contention (#/s)               |               167 |                169 |   +1.20% |
| Max ThreadPool Threads Count            |                32 |                 33 |   +3.12% |
| Max ThreadPool Queue Length             |                90 |                 94 |   +4.44% |
| Max ThreadPool Items (#/s)              |            36,324 |             23,182 |  -36.18% |
| Max Active Timers                       |                98 |                100 |   +2.04% |
| IL Jitted (B)                           |         1,246,255 |          1,109,622 |  -10.96% |
| Methods Jitted                          |            12,926 |             12,358 |   -4.39% |
| Avg Allocation Rate (B/sec)             |       302,589,018 |        290,538,748 |   -3.98% |
| 90th Allocation Rate (B/sec)            |       331,157,288 |        319,837,040 |   -3.42% |


| load                   | exp-lin-2-ioc-pre | exp-lin-2-ioc-post |          |
| ---------------------- | ----------------- | ------------------ | -------- |
| CPU Usage (%)          |                 1 |                  2 | +100.00% |
| Cores usage (%)        |                39 |                 42 |   +7.69% |
| Working Set (MB)       |                38 |                 38 |    0.00% |
| Private Memory (MB)    |               357 |                357 |    0.00% |
| Start Time (ms)        |                 0 |                  0 |          |
| First Request (ms)     |               328 |                302 |   -7.93% |
| Requests/sec           |             8,118 |              9,352 |  +15.20% |
| Requests               |           487,855 |            562,041 |  +15.21% |
| Mean latency (ms)      |             14.09 |              12.67 |  -10.08% |
| Max latency (ms)       |            222.82 |             237.03 |   +6.38% |
| Bad responses          |                 0 |                  0 |          |
| Socket errors          |                 0 |                  0 |          |
| Read throughput (MB/s) |              1.25 |               1.44 |  +15.20% |
| Latency 50th (ms)      |              9.93 |               8.59 |  -13.49% |
| Latency 75th (ms)      |             15.93 |              13.84 |  -13.12% |
| Latency 90th (ms)      |             28.80 |              26.00 |   -9.72% |
| Latency 99th (ms)      |             69.74 |              68.90 |   -1.20% |

(gains are also 8-10% on Windows but we don't trust those benchmarks just enough yet - using Linux for stable results here)

src/WebJobs.Script.WebHost/DependencyInjection/ServiceProviderExtensions.cs
```diff
@@ -0,0 +1,202 @@
...
+                // If all services of the same type are singletons.
+                else if (services.All(s => s.Lifetime == ServiceLifetime.Singleton))
+                {
+                    // We can resolve them from the main container.
+                    var instances = serviceProvider.GetServices(services.Key);
+
+                    for (var i = 0; i < services.Count(); i++)
+                    {
+                        clonedCollection.CloneSingleton(services.ElementAt(i), instances.ElementAt(i));
+                    }
```
davidfowl, line 104: ~This is N^2~ I just realized services isn't the full collection.

src/WebJobs.Script.Grpc/Server/AspNetCoreGrpcServer.cs
```diff
@@ -60,10 +60,11 @@ protected async ValueTask DisposeAsync(bool disposing)
...
             if (!_disposed)
             {
-                if (disposing)
+                var temp = _grpcHost;
+                if (disposing && temp != null)
                 {
-                    await _grpcHost.StopAsync();
-                    _grpcHost.Dispose();
+                    await temp.StopAsync();
+                    temp.Dispose();
```
NickCraver: cc @brettsam - I _think_ this is an existing issue, if we register (but never start) the server then dispose it, we'll null ref. I noticed in this PR (I'm not sure how it worked before this, some other delta in the testing landscape), but the fix I think is probably one to lift to dev independently - thoughts?

src/WebJobs.Script.WebHost/DependencyInjection/ServiceProviderExtensions.cs
```diff
@@ -0,0 +1,202 @@
+﻿// TODO: Licensing header, and split these? - these extensions are based on Orchard (https://github.com/OrchardCMS/OrchardCore)
```
sebastienros: I think recommendation for other MS projects has been to flow the license header, which should be made out of the following block, and also mention the project in the NOTICE file.

```
BSD 3-Clause License
https://opensource.org/licenses/BSD-3-Clause

Copyright (c) .NET Foundation. All rights reserved.
```

src/WebJobs.Script.WebHost/DependencyInjection/ServiceProviderExtensions.cs
```diff
@@ -0,0 +1,202 @@
...
+                if (ChildContainerIgnoredTypes.Contains(services.Key))
+                {
+                    // Do nothing - we don't want to copy these to a child scope for reasons above
+                }
+
+                // A generic type definition is rather used to create other constructed generic types.
+                else if (services.Key.IsGenericTypeDefinition)
+                {
+                    // So, we just need to pass the descriptor.
+                    foreach (var service in services)
```
halter73, line 49: Not cloning singleton open generic service descriptors is probably not ideal.

`IOptions<T>` (`UnamedOptionsManager<T>`) comes to mind as a common open generic service where this might have undesirable side effects because `IConfigureOptions<T>` callbacks will be rerun on a new `T` even if the options have already been resolved by the parent container.

Unfortunately, I don't come with solutions because you cannot provide a single implementation instance for the `IDisposable` case, and you cannot get the concrete type in the implantation factory in the non-`IDisposable` case. I do think the comment should be updated to indicate that just passing through singleton open generic service descriptors is problematic.

halter73: Or maybe it's a feature because you want to be able to reconfigure options in each child container? In that case, it's kind of interesting that the one scenario the child container doesn't use the parent container's singletons, it ends up helping.

NickCraver: @sebastienros is this an implication you've already thought through in Orchard?

src/WebJobs.Script.WebHost/DependencyInjection/ServiceProviderExtensions.cs
NickCraver: hmmm we don't have a NOTICE today - @fabiocav recommendations? I was trying to find DryIoc mentions but don't see past the original files, perhaps we should add/adopt the `NOTICE` pattern if going this route? Happy to do that, but maybe independently of this PR and add to it if we go this route (to make sure we capture anything else in there).

NickCraver:
Another interesting oddity here - @safihamid was running some cold start test on this and in the throttled case (we're not in throughput samples generally), `HttpThrottleMiddleware` is lit up in the use path. This breaks because it's trying to access the `HttpRequestQueue` on the `.Invoke()` (via injection) which would be coming not from the application host but the child container for the job host at that point. I'm slightly puzzled as to how this _was_ working with DryIoc, because the failure makes sense.

I think as Orchard does, we'd need to swap the service scope for the request in the Http case - perhaps as a middleware or something else. Curious on thoughts here - IMO the current setup is a bit odd as we transition from a set of middlewares that are global into ones that are pseudo-JobHost specific (like the throttle) - it seems like we could pick a point and set a boundary, and change the services we resolve from that that point, when we enter the tenant.

Stack of error:
```
System.InvalidOperationException : Unable to resolve service for type 'Microsoft.Azure.WebJobs.Script.WebHost.HttpRequestQueue' while attempting to Invoke middleware 'Microsoft.Azure.WebJobs.Script.WebHost.Middleware.HttpThrottleMiddleware'.
at Microsoft.AspNetCore.Builder.UseMiddlewareExtensions.GetService(IServiceProvider sp,Type type,Type middleware)
at lambda_method2(Closure ,Object ,HttpContext ,IServiceProvider )
at Microsoft.AspNetCore.Builder.UseMiddlewareExtensions.<>c__DisplayClass5_1.<UseMiddleware>b__2(HttpContext context)
at Microsoft.AspNetCore.Builder.UseWhenExtensions.<>c__DisplayClass0_1.<UseWhen>b__1(HttpContext context)
at Microsoft.AspNetCore.Builder.UseWhenExtensions.<>c__DisplayClass0_1.<UseWhen>b__1(HttpContext context)
at async Microsoft.Azure.WebJobs.Script.WebHost.Middleware.ExceptionMiddleware.Invoke(HttpContext context) at D:\git\azure-functions-host\src\WebJobs.Script.WebHost\Middleware\ExceptionMiddleware.cs : 27
```

src/WebJobs.Script.Grpc/Server/AspNetCoreGrpcServer.cs
brettsam: yeah, good call -- can we move this to a separate PR and get it in?

NickCraver: Done: #7981

PR #7973, opened 2021-12-09
Solution: simplify Samples inclusion.

NickCraver:
Instead of having every sample addition or removal be a solution file change, we can use the https://github.com/Microsoft/MSBuildSdks/tree/main/src/NoTargets SDK type in a minimal `.csproj` file to include the whole directory. This simplifies the solution file greatly with the minimal change of moving Samples to a direct project at the bottom (rather than a folder).

I noticed this when I forgot the solution addition in #7964, and we're planning to add some more...so how about we don't modify the `.sln` file for this? I hope that's appealing :)

### Pull request checklist

### Additional information

Here's a comparison (note the second is taller because we were missing some samples...)
| Before | After |
|--------|-------|
|before|after|


NickCraver:
I...didn't break those tests, I promise! Seems to be something up with the Node runner at the moment.

liliankasem (approved):

brettsam (approved):

PR #7974, opened 2021-12-09
Ensure child worker processes are monitored after specialization

mathewc:
Currently, child worker processes are only registered for monitoring when the process is started. That's a problem if the host is started in placeholder mode and later specialized. In that case, the worker channels are NOT restarted, and the child processes are not registered for monitoring by the new specialized host instance.

For testing, I've configured my local functions host to start in placeholder mode. I can see that placeholder worker channels are used. After triggering specialization, I can see the worker channels are specialized. Previously before my fix, the newly restarted/specialized host wouldn't be monitoring the child processes, because the were only registered on process start before. Now the worker process instance handles the HostStartEvent and registers.

Caught this issue in production testing. For a Python Consumption app, after specialization only the main host process is being monitored, not the worker process. That impacts DynamicConcurrency - it's unable to make correct concurrency decisions because it doesn't have a full view of CPU usage. In previous rounds of testing, I was testing with private site extensions, with placeholders disabled.

Will send a PR backporting to v3 as well, once this one is signed off on.

### Pull request checklist

    * [x] Backport PR https://github.com/Azure/azure-functions-host/pull/7978

### DIFF
--- a/src/WebJobs.Script.WebHost/Configuration/ScriptApplicationHostOptionsSetup.cs
+++ b/src/WebJobs.Script.WebHost/Configuration/ScriptApplicationHostOptionsSetup.cs
@@ -69,7 +69,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Configuration
                 options.IsStandbyConfiguration = true;
             }
 
-            options.IsFileSystemReadOnly = IsZipDeployment(out bool isScmRunFromPackage);
+            options.IsFileSystemReadOnly |= IsZipDeployment(out bool isScmRunFromPackage);
             options.IsScmRunFromPackage = isScmRunFromPackage;
         }
 
--- a/test/WebJobs.Script.Tests/WebJobsScriptHostServiceTests.cs
+++ b/test/WebJobs.Script.Tests/WebJobsScriptHostServiceTests.cs
@@ -9,6 +9,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Script.Config;
 using Microsoft.Azure.WebJobs.Script.Diagnostics;
+using Microsoft.Azure.WebJobs.Script.Eventing;
 using Microsoft.Azure.WebJobs.Script.Scale;
 using Microsoft.Azure.WebJobs.Script.WebHost;
 using Microsoft.Azure.WebJobs.Script.WebHost.DependencyInjection;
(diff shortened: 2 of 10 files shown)


src/WebJobs.Script.WebHost/Configuration/ScriptApplicationHostOptionsSetup.cs
```diff
@@ -69,7 +69,7 @@ public void Configure(string name, ScriptApplicationHostOptions options)
                 options.IsStandbyConfiguration = true;
             }
 
-            options.IsFileSystemReadOnly = IsZipDeployment(out bool isScmRunFromPackage);
+            options.IsFileSystemReadOnly |= IsZipDeployment(out bool isScmRunFromPackage);
```
mathewc, line 72: To test this fix, I had to simulate specialization locally. That requires configuration that ensures placeholder channels are used. For this specific change, the check [here](https://github.com/Azure/azure-functions-host/blob/b1c5107d2759c1a7a916fffda8d244ae346127b5/src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs#L152) requires a RO file system. So I'm making this change to facilitate testability, however it's a logical change to make regardless.

test/WebJobs.Script.Tests/WebJobsScriptHostServiceTests.cs
```diff
@@ -87,6 +88,33 @@ private Mock<IHost> CreateMockHost(SemaphoreSlim disposedSemaphore = null)
...
+                _monitor, hostBuilder.Object, _loggerFactory,
+                _mockScriptWebHostEnvironment.Object, _mockEnvironment.Object,
+                _hostPerformanceManager, _healthMonitorOptions,
+                metricsLogger, new Mock<IApplicationLifetime>().Object,
+                _mockConfig, mockEventManager.Object);
+
+            await _hostService.StartAsync(CancellationToken.None);
+
+            // add general post startup validations here
+            mockEventManager.Verify(_ => _.Publish(It.IsAny<HostStartEvent>()), Times.Once());
```
mathewc, line 115: Added this StartAsync test so I had a place to add verification for the new event

test/WebJobs.Script.Tests/Workers/Rpc/RpcWorkerProcessTests.cs
```diff
@@ -42,6 +51,78 @@ public RpcWorkerProcessTests()
...
+            _hostProcessMonitorMock.Verify(p => p.RegisterChildProcess(process), Times.Once);
+
+            // registration is skipped if attempted again for the same monitor
+            _rpcWorkerProcess.RegisterWithProcessMonitor();
+            _hostProcessMonitorMock.Verify(p => p.RegisterChildProcess(process), Times.Once);
+
+            // if the monitor changes (e.g. due to a host restart)
+            // registration is performed
+            _hostProcessMonitorMock = new Mock<IHostProcessMonitor>(MockBehavior.Strict);
+            _hostProcessMonitorMock.Setup(p => p.RegisterChildProcess(process));
```
pragnagopa, line 99: this is good test coverage! I recommend updating this test to publish `HostStartEvent` and verifying `RegisterWithProcessMonitor` .

pragnagopa (approved):
Looks good! Added minor comment!

test/WebJobs.Script.Tests/Workers/Rpc/RpcWorkerProcessTests.cs
mathewc: The event manager used by this test suite is a mock, so can't be used to publish. However I have test validating that the constructor registers the event handler and that when that handler (OnHostStart) is called it does the right thing. I'm intentionally not testing that IScriptEventManager delivers an event when it's published - that's outside of the scope of these unit tests (i.e. I'm not testing IScriptEventManager).

PR #7976, opened 2021-12-10
[v3.x] Drain mode cleanup: remove GET, return 202 Accepted

liliankasem:
Backport #7876 to v3.x


brettsam (dismissed):

brettsam:
@liliankasem -- when you merge, can you remove the old PR number from the commit message?

Linking to the original -- #7876

brettsam (approved):

PR #7977, opened 2021-12-10
Performance: Cache ILogger on FunctionDesciptor

NickCraver:
Part of #7908. Cache the logger on the `FunctionDescriptor` to avoid the `.CreateLogger()` call every invocation.

Internally, `.CreateLogger()` (specifically the string variable) has a lock which we pile up on under load and we can avoid both the lock and the service fetch here. Since `FunctionDescriptor`s are re-created with each JobHost, memoizing the logger like this should be okay and change on any host change as we want.

Resolves part of #7908

### Pull request checklist

#### Crank comparison

| application                             | dev         | fd-logger   |         |
| Working Set (MB)                        |         378 |         382 |  +1.06% |
| Private Memory (MB)                     |       1,181 |       1,199 |  +1.52% |
| Build Time (ms)                         |      15,235 |      17,833 | +17.05% |
| Start Time (ms)                         |       4,757 |       4,759 |  +0.04% |
| Published Size (KB)                     |     674,605 |     674,605 |   0.00% |
| .NET Core SDK Version                   |     6.0.100 |     6.0.100 |         |
| Max Working Set (MB)                    |         396 |         400 |  +1.01% |
| Max GC Heap Size (MB)                   |         145 |         152 |  +4.83% |
| Size of committed memory by the GC (MB) |         207 |         225 |  +8.70% |
| Max Number of Gen 0 GCs / sec           |       13.00 |       13.00 |   0.00% |
| Max Time in GC (%)                      |        6.00 |        7.00 | +16.67% |
| Max Gen 0 Size (B)                      |  92,442,184 |  32,369,448 | -64.98% |
| Max Gen 1 Size (B)                      |  11,112,344 |   9,449,160 | -14.97% |
| Max Gen 2 Size (B)                      |  29,196,432 |  27,484,952 |  -5.86% |
| Max LOH Size (B)                        |   9,714,720 |  11,418,936 | +17.54% |
| Max Allocation Rate (B/sec)             | 334,226,264 | 340,185,376 |  +1.78% |
| Max GC Heap Fragmentation               |          83 |          68 | -17.46% |
| Max Exceptions (#/s)                    |         258 |         252 |  -2.33% |
| Max Lock Contention (#/s)               |         109 |         101 |  -7.34% |
| Max ThreadPool Threads Count            |          29 |          32 | +10.34% |
| Max ThreadPool Queue Length             |          92 |          92 |   0.00% |
| Max ThreadPool Items (#/s)              |      36,134 |      39,213 |  +8.52% |
| Max Active Timers                       |          85 |          95 | +11.76% |
| IL Jitted (B)                           |   1,054,396 |   1,055,449 |  +0.10% |
| Methods Jitted                          |      10,879 |      10,887 |  +0.07% |
| Avg Allocation Rate (B/sec)             | 303,164,832 | 295,241,931 |  -2.61% |
| 90th Allocation Rate (B/sec)            | 333,317,216 | 328,697,416 |  -1.39% |

| load                   | dev     | fd-logger |         |
| ---------------------- | ------- | --------- | ------- |
| CPU Usage (%)          |       1 |         1 |   0.00% |
| Cores usage (%)        |      37 |        39 |  +5.41% |
| Working Set (MB)       |      38 |        37 |  -2.63% |
| Private Memory (MB)    |     357 |       357 |   0.00% |
| Start Time (ms)        |       0 |         0 |         |
| First Request (ms)     |     338 |       344 |  +1.78% |
| Requests/sec           |   8,082 |     8,467 |  +4.76% |
| Requests               | 485,330 |   508,824 |  +4.84% |
| Mean latency (ms)      |   13.76 |     13.10 |  -4.80% |
| Max latency (ms)       |  205.02 |    169.65 | -17.25% |
| Bad responses          |       0 |         0 |         |
| Socket errors          |       0 |         0 |         |
| Read throughput (MB/s) |    1.24 |      1.30 |  +4.84% |
| Latency 50th (ms)      |   10.00 |      9.57 |  -4.30% |
| Latency 75th (ms)      |   15.87 |     15.12 |  -4.73% |
| Latency 90th (ms)      |   27.73 |     26.23 |  -5.41% |
| Latency 99th (ms)      |   62.69 |     59.60 |  -4.93% |

Note I'm not checking the unit test box because I'm not sure if we have a reliable way to test after-reloading-the-host to make sure the `ILogger` reference on the descriptor changes. IMO, this would be a good test to add but I could use some guidance on the best place for that.


src/WebJobs.Script/Host/ScriptHost.cs
```diff
@@ -846,7 +848,7 @@ private void HandleHostError(Exception exception)
         {
             if (exception == null)
             {
-                throw new ArgumentNullException(nameof(exception));
+                throw new ArgumentNullException("exception");
```
fabiocav: Curious about the reason for this change.

src/WebJobs.Script/Host/ScriptHost.cs
```diff
@@ -846,7 +848,7 @@ private void HandleHostError(Exception exception)
         {
             if (exception == null)
```
fabiocav: Move this to use `ArgumentNullException.ThrowIfNull`?

src/WebJobs.Script/Description/FunctionDescriptor.cs
```diff
@@ -39,6 +41,7 @@ internal FunctionDescriptor()
             TriggerBinding = InputBindings?.SingleOrDefault(p => p.Metadata.IsTrigger);
             HttpTriggerAttribute = GetTriggerAttributeOrNull<HttpTriggerAttribute>();
             LogCategory = LogCategories.CreateFunctionCategory(Name);
+            Logger = loggerFactory.CreateLogger(LogCategory);
```
mathewc, line 44: Previously similar to the improvement you're making here, I added LogCategory to cache it. Now that we're caching the logger, LogCategory isn't needed anymore in the public surface it seems. It was only used in that one place in the middleware, which you've removed.

src/WebJobs.Script.WebHost/Middleware/FunctionInvocationMiddleware.cs
```diff
@@ -86,8 +86,7 @@ private async Task<IActionResult> GetResultAsync(HttpContext context, IFunctionE
             {
                 // Add the request to the logging scope. This allows the App Insights logger to
                 // record details about the request.
-                ILoggerFactory loggerFactory = context.RequestServices.GetService<ILoggerFactory>();
```
mathewc, line 86: Are we using the same LoggerFactory in the new code? Want to make sure there isn't a subtle change here.

src/WebJobs.Script/Host/ScriptHost.cs
NickCraver: This was an inadvertent revert when carving up branches - will definitely fix!

src/WebJobs.Script/Description/FunctionDescriptor.cs
NickCraver: ACK - will simplify

src/WebJobs.Script/Host/ScriptHost.cs
NickCraver: The reason I didn't here is this project is `netstandard2.1` from the common build props - didn't want to change it _just_ for this to `net6.0` but if we're okay aligning the projects happy to adjust - thoughts?

brettsam (commented):
This looks good to me but I see a couple of outstanding comments/questions.

liliankasem:
Tagging @fabiocav & @mathewc for a second pass, I think your comments were addressed?

src/WebJobs.Script.WebHost/Middleware/FunctionInvocationMiddleware.cs
fabiocav: Yes, with the change, the logger factory used is that gets injected into the `ScriptHost`, which will have the same scope of what is being resolved here, so we should be good.

fabiocav (approved):
Changes look good to me.

src/WebJobs.Script/Description/FunctionDescriptor.cs
mathewc: Also, ILoggerFactory is IDisposable - do we need to make FunctionDescriptor disposable now as well and ensure it's disposed at the right time?

liliankasem (approved):

src/WebJobs.Script/Description/FunctionDescriptor.cs
jviau: No - it is an external dependency, this should not handle disposing it. Also `ILoggerFactory` is a singleton. Host shutdown handles disposing it.

PR #7980, opened 2021-12-13
Host to fall back on its own indexing if worker denies

### LINKED ISSUE #7916: Worker Indexing - Host to fall back on its indexing if worker denies by checking legacy payload
To add ability for worker to check the payload and send message to host the it cannot index the legacy payload. Host to then fall back to its indexing.

[Documentation](https://eng.ms/docs/cloud-ai-platform/devdiv/serverless-paas-balam/serverless-paas-balam/app-service-web-apps/antares-team-documents/functionteamdocs/design/WorkerIndexing/WorkerIndexing#35-host-fallback-scenario)

surgupta-msft:

resolves #7916 

[Documentation](https://eng.ms/docs/cloud-ai-platform/devdiv/serverless-paas-balam/serverless-paas-balam/app-service-web-apps/antares-team-documents/functionteamdocs/design/WorkerIndexing/WorkerIndexing#35-host-fallback-scenario)

### Pull request checklist
- Taken changes from Protobuf repo needed for this change. [PR merged](https://github.com/Azure/azure-functions-language-worker-protobuf/pull/61/files).
- Made WorkerFunctionMetadataProvider as AggregateFunctionMetadataProvider which first uses worker indexing and if worker denies then it falls back to host indexing by calling HostFunctionMetadataProvider
- Worker has the ability to deny indexing by setting flag functionMetadataResponse.UseDefaultMetadataIndexing to True.


src/WebJobs.Script.Grpc/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
```diff
@@ -78,7 +78,7 @@ message StreamingMessage {
```
liliankasem, line 78: Can you raise a separate PR for bumping the proto version?

https://github.com/Azure/azure-functions-language-worker-protobuf#pulling-updates

src/WebJobs.Script/Host/AggregateFunctionMetadataProvider.cs
```diff
@@ -17,44 +17,69 @@
...
             _logger.FunctionMetadataProviderParsingFunctions();
+
+            bool workerIndexing = Utility.CanWorkerIndex(workerConfigs, environment);
+
             if (_functions.IsDefaultOrEmpty || forceRefresh)
             {
-                if (_dispatcher != null)
+                if (_dispatcher == null && workerIndexing == true)
+                {
+                    throw new ArgumentNullException(nameof(_dispatcher));
```
liliankasem: What happens if dispatcher is null and worker indexing is false? Do we also throw? Just wondering if  throwing `ArgumentNullException` is better were it was originally in the constructor, it would also simplify this function?

src/WebJobs.Script/Host/AggregateFunctionMetadataProvider.cs
```diff
@@ -17,44 +17,69 @@
...
+                    }
+                    else
+                    {
+                        // If worker denies indexing then falling back to the host for Indexing
+                        _functions = _hostFunctionMetadataProvider.GetFunctionMetadataAsync(workerConfigs, SystemEnvironment.Instance, forceRefresh).GetAwaiter().GetResult();
+                    }
                 }
+                else if (workerIndexing == false)
+                {
+                    _functions = _hostFunctionMetadataProvider.GetFunctionMetadataAsync(workerConfigs, SystemEnvironment.Instance, forceRefresh).GetAwaiter().GetResult();
```
liliankasem: nit: might be cleaner to separate this out into its own if statement  instead of if/else, and add a check for rawFunctions == null, so that way we only need to write this line of code once

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -486,45 +486,48 @@ internal Task<List<RawFunctionMetadata>> SendFunctionMetadataRequest()
...
 
-            foreach (var metadataResponse in functionMetadataResponses.FunctionLoadRequestsResults)
+            if (functionMetadataResponse.UseDefaultMetadataIndexing == false)
             {
-                var metadata = metadataResponse.Metadata;
-                if (metadata != null)
+                foreach (var metadata in functionMetadataResponse.FunctionMetadataResults)
                 {
-                    if (metadata.Status != null && metadata.Status.IsFailure(out Exception metadataRequestEx))
+                    if (metadata != null)
```
liliankasem: nit: wonder if we can reverse some of these if statements to simplify, i.e. `if (metadata == null) break;` ?

src/WebJobs.Script/Host/AggregateFunctionMetadataProvider.cs
```diff
@@ -17,44 +17,69 @@
...
+                }
+
+                _functions = functions.ToImmutableArray();
```
liliankasem: just wondering why this needed to be moved up?

src/WebJobs.Script/Host/AggregateFunctionMetadataProvider.cs
```diff
@@ -17,44 +17,69 @@
...
-            _dispatcher = invocationDispatcher ?? throw new ArgumentNullException(nameof(invocationDispatcher));
+            _dispatcher = invocationDispatcher;
+            _hostFunctionMetadataProvider = hostFunctionMetadataProvider;
         }
 
         public ImmutableDictionary<string, ImmutableArray<string>> FunctionErrors
            => _functionErrors.ToImmutableDictionary(kvp => kvp.Key, kvp => kvp.Value.ToImmutableArray());
 
-        public async Task<ImmutableArray<FunctionMetadata>> GetFunctionMetadataAsync(IEnumerable<RpcWorkerConfig> workerConfigs, bool forceRefresh)
+        public async Task<ImmutableArray<FunctionMetadata>> GetFunctionMetadataAsync(IEnumerable<RpcWorkerConfig> workerConfigs, IEnvironment environment, bool forceRefresh)
```
liliankasem, line 42: Do we need to add new unit tests to cover the new paths in this function?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
NickCraver: +1 - existing issue, but I think `continue` here if intent hasn't changed? I'm not 100% how this _should_ work - just commenting on the 1:1

liliankasem: my bad, totally meant `continue` :D

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -486,45 +486,48 @@ internal Task<List<RawFunctionMetadata>> SendFunctionMetadataRequest()
...
+                            EntryPoint = metadata.EntryPoint,
+                            Name = metadata.Name
+                        };
 
-                    var bindings = new List<string>();
-                    foreach (string binding in metadata.RawBindings)
-                    {
-                        bindings.Add(binding);
-                    }
+                        functionMetadata.SetFunctionId(metadata.Name);
```
satvu: Not 100% sure on the FunctionId convention but in the previous design with `FunctionLoadRequests` both python and dotnet returned a new guid as the function id (instead of the function name).

[This](https://github.com/Azure/azure-functions-host/blob/431f1f87db87c2ce70d7009fcc99a1b0b7a6921d/src/WebJobs.Script/Extensions/FunctionMetadataExtensions.cs#L36) `GetFunctionId()` method in `FunctionMetadataExtensions` also returns a guid if the function id does not already exist.

surgupta-msft: @satvu Thanks for mentioning this. This helped me in fixing some tests failure.

@fabiocav  As part of this [PR merge](https://github.com/Azure/azure-functions-language-worker-protobuf/pull/61/files#:~:text=repeated-,RpcFunctionMetadata,-function_metadata_results%20%3D%201%3B) in Protobuf repo, we replaced **FunctionLoadRequest** by **RpcFunctionMetadata**. Due to this, the *FunctionId* present in **FunctionLoadRequest** got removed, causing tests to fail and also the scenario @satvu is indicating. So, I am adding FunctionId in this [PR](https://github.com/Azure/azure-functions-host/pull/7980/files#:~:text=//%20unique%20function%20identifier,%7D) in host repo. If this looks good to you, I will also update the Protobuf repo.

src/WebJobs.Script/Host/AggregateFunctionMetadataProvider.cs
surgupta-msft: Thanks for catching. Updated it.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
surgupta-msft: Done. Thanks.

src/WebJobs.Script/Host/AggregateFunctionMetadataProvider.cs
surgupta-msft: I have cleaned up the code a bit here. The reason I cannot throw ArgumentNullException in constructor is - if workerIndexing = false, dispatcher can be null but if workerIndexing = true then it can't be and both of these scenarios are being handled in this Aggregate class.

src/WebJobs.Script/Host/AggregateFunctionMetadataProvider.cs
```diff
@@ -17,44 +18,65 @@
...
+                        await _dispatcher.FinishInitialization(functions);
+                    }
+                }
+                if (IsNullOrEmpty(rawFunctions) || !workerIndexing)
+                {
+                    // If worker denies indexing then falling back to the host for Indexing
+                    _functions = _hostFunctionMetadataProvider.GetFunctionMetadataAsync(workerConfigs, SystemEnvironment.Instance, forceRefresh).GetAwaiter().GetResult();
                 }
             }
+            _functions = _functions.ToImmutableArray();
```
liliankasem: nit: do we need to change the order here? if not might be best to leave it after the logger for minimal changes

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -486,22 +486,26 @@ internal Task<List<RawFunctionMetadata>> SendFunctionMetadataRequest()
...
         // parse metadata response into RawFunctionMetadata objects for WorkerFunctionMetadataProvider to further parse and validate
-        internal void ProcessFunctionMetadataResponses(FunctionMetadataResponses functionMetadataResponses)
+        internal void ProcessFunctionMetadataResponses(FunctionMetadataResponse functionMetadataResponse)
         {
             _workerChannelLogger.LogDebug("Received the worker function metadata response from worker {worker_id}", _workerId);
 
             var functions = new List<RawFunctionMetadata>();
 
-            foreach (var metadataResponse in functionMetadataResponses.FunctionLoadRequestsResults)
+            if (functionMetadataResponse.UseDefaultMetadataIndexing == false)
```
soninaren, line 495: nit: UseDefaultFunctionIndexing

src/WebJobs.Script/Host/AggregateFunctionMetadataProvider.cs
```diff
@@ -17,44 +18,65 @@
...
+        private IFunctionMetadataProvider _hostFunctionMetadataProvider;
 
-        public WorkerFunctionMetadataProvider(ILogger<WorkerFunctionMetadataProvider> logger, IFunctionInvocationDispatcher invocationDispatcher)
+        public AggregateFunctionMetadataProvider(
+            ILogger<AggregateFunctionMetadataProvider> logger,
+            IFunctionInvocationDispatcher invocationDispatcher,
+            IFunctionMetadataProvider hostFunctionMetadataProvider)
         {
             _logger = logger;
-            _dispatcher = invocationDispatcher ?? throw new ArgumentNullException(nameof(invocationDispatcher));
```
soninaren, line 30: why move this inside the function?

src/WebJobs.Script/Host/AggregateFunctionMetadataProvider.cs
```diff
@@ -17,44 +18,65 @@
...
+            bool workerIndexing = Utility.CanWorkerIndex(workerConfigs, environment);
+
             if (_functions.IsDefaultOrEmpty || forceRefresh)
             {
-                if (_dispatcher != null)
+                if (workerIndexing)
                 {
+                    if (_dispatcher == null)
+                    {
+                        throw new ArgumentNullException(nameof(_dispatcher));
```
soninaren: This is not the correct exception. dispatcher is not an argument to this function.

src/WebJobs.Script/Host/AggregateFunctionMetadataProvider.cs
```diff
@@ -17,44 +18,65 @@
...
+                    }
+
                     // start up GRPC channels
                     await _dispatcher.InitializeAsync(new List<FunctionMetadata>());
 
                     // get function metadata from worker, then validate it
                     rawFunctions = await _dispatcher.GetWorkerMetadata();
```
soninaren, line 63: this should not be called if `WorkerIndexing` is false already.  If worker config itself does not indicate that the worker can index, there is no point sending indexing request.

src/WebJobs.Script/Host/AggregateFunctionMetadataProvider.cs
```diff
@@ -17,44 +18,65 @@
...
-                    // set up invocation buffers and send load requests
-                    await _dispatcher.FinishInitialization(functions);
+                    if (!IsNullOrEmpty(rawFunctions))
+                    {
+                        functions = ValidateMetadata(rawFunctions);
+                        // set up invocation buffers and send load requests
+                        await _dispatcher.FinishInitialization(functions);
+                    }
+                }
+                if (IsNullOrEmpty(rawFunctions) || !workerIndexing)
```
soninaren: empty rawFunctions does not mean that the worker rejected indexing. We need a definite indicator / evaluation of a field that explicitly indicates that the worker does not identify this as a valid payload and cannot index.

soninaren (changes requested):

src/WebJobs.Script/Host/AggregateFunctionMetadataProvider.cs
NickCraver: We should also `await` here, rather than doing `.GetAwaiter().GetResult()`

src/WebJobs.Script/Host/FunctionMetadataManager.cs
```diff
@@ -132,11 +132,9 @@ internal ImmutableArray<FunctionMetadata> LoadFunctionMetadata(bool forceRefresh
             ImmutableArray<FunctionMetadata> immutableFunctionMetadata;
             var workerConfigs = _languageWorkerOptions.Value.WorkerConfigs;
 
-            IFunctionMetadataProvider metadataProvider = Utility.CanWorkerIndex(workerConfigs, _environment)
-                ? new WorkerFunctionMetadataProvider(_loggerFactory.CreateLogger<WorkerFunctionMetadataProvider>(), dispatcher)
-                : _functionMetadataProvider;
+            IFunctionMetadataProvider metadataProvider = new AggregateFunctionMetadataProvider(_loggerFactory.CreateLogger<AggregateFunctionMetadataProvider>(), dispatcher, _functionMetadataProvider);
 
-            immutableFunctionMetadata = metadataProvider.GetFunctionMetadataAsync(workerConfigs, forceRefresh).GetAwaiter().GetResult();
+            immutableFunctionMetadata = metadataProvider.GetFunctionMetadataAsync(workerConfigs, SystemEnvironment.Instance, forceRefresh).GetAwaiter().GetResult();
```
NickCraver, line 137: FWIW this is another path that should be async (and another in here not present in the PR diff) - can we change the interface to remedy these? If we're loading _all_ of these asynchronously as an end game (100% out of proc), we should be async on the paths.

I think we should also be passing `_environment` here, rather than tying to a specific implementation.

src/WebJobs.Script/Host/AggregateFunctionMetadataProvider.cs
surgupta-msft: updated code here.

liliankasem (dismissed):

src/WebJobs.Script.Grpc/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
```diff
@@ -78,7 +78,7 @@ message StreamingMessage {
 
     // Worker indexing message types
     FunctionsMetadataRequest functions_metadata_request = 29;
-    FunctionMetadataResponses function_metadata_responses = 30;
+    FunctionMetadataResponse function_metadata_response = 30;
```
kshyju, line 81: Question: So, this would not be a breaking change because the prop names are not really used in GRPC payload, and we should be good as long as we are not changing the order?

src/WebJobs.Script.Grpc/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
```diff
@@ -288,12 +294,15 @@ message FunctionsMetadataRequest {
 }
 
 // Worker sends function metadata back to host
-message FunctionMetadataResponses {
+message FunctionMetadataResponse {
   // list of function indexing responses
-  repeated FunctionLoadRequest function_load_requests_results = 1;
```
kshyju, line 293: Question: This property is currently only being used for worker indexing/stein feature? So, nothing else would break I assume ?

src/WebJobs.Script.Grpc/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
```diff
@@ -288,12 +294,15 @@ message FunctionsMetadataRequest {
...
+  repeated RpcFunctionMetadata function_metadata_results = 1;
 
   // status of overall metadata request
   StatusResult result = 2;
+
+  // if set to true then host will perform indexing
+  bool useDefaultMetadataIndexing = 3;
```
kshyju, line 305: nit: Should we consider being more specific name? Instead of "Default" use some terminology which explicitly says it is done on host.

src/WebJobs.Script/Host/AggregateFunctionMetadataProvider.cs
```diff
@@ -17,44 +18,65 @@
...
            => _functionErrors.ToImmutableDictionary(kvp => kvp.Key, kvp => kvp.Value.ToImmutableArray());
 
-        public async Task<ImmutableArray<FunctionMetadata>> GetFunctionMetadataAsync(IEnumerable<RpcWorkerConfig> workerConfigs, bool forceRefresh)
+        public async Task<ImmutableArray<FunctionMetadata>> GetFunctionMetadataAsync(IEnumerable<RpcWorkerConfig> workerConfigs, IEnvironment environment, bool forceRefresh)
         {
             IEnumerable<RawFunctionMetadata> rawFunctions = new List<RawFunctionMetadata>();
             IEnumerable<FunctionMetadata> functions = new List<FunctionMetadata>();
             _logger.FunctionMetadataProviderParsingFunctions();
+
+            bool workerIndexing = Utility.CanWorkerIndex(workerConfigs, environment);
```
kshyju: Move this inside the if condition (just before line 52) so that it is executed only if the `if` expression in line 50 yields `true`

src/WebJobs.Script/Host/AggregateFunctionMetadataProvider.cs
kshyju: Should the null check and throw be done inside the constructor where this is injected in?

src/WebJobs.Script/Host/AggregateFunctionMetadataProvider.cs
```diff
@@ -17,44 +18,65 @@
...
+            _hostFunctionMetadataProvider = hostFunctionMetadataProvider;
         }
 
         public ImmutableDictionary<string, ImmutableArray<string>> FunctionErrors
            => _functionErrors.ToImmutableDictionary(kvp => kvp.Key, kvp => kvp.Value.ToImmutableArray());
 
-        public async Task<ImmutableArray<FunctionMetadata>> GetFunctionMetadataAsync(IEnumerable<RpcWorkerConfig> workerConfigs, bool forceRefresh)
+        public async Task<ImmutableArray<FunctionMetadata>> GetFunctionMetadataAsync(IEnumerable<RpcWorkerConfig> workerConfigs, IEnvironment environment, bool forceRefresh)
         {
             IEnumerable<RawFunctionMetadata> rawFunctions = new List<RawFunctionMetadata>();
```
kshyju: Declare closer to usage (line 63) so you will not allocate a list if the if conditions are not passing below.

src/WebJobs.Script/Host/AggregateFunctionMetadataProvider.cs
```diff
@@ -155,5 +177,14 @@ internal static FunctionMetadata ValidateBindings(IEnumerable<string> rawBinding
 
             return function;
         }
+
+        private bool IsNullOrEmpty(IEnumerable<RawFunctionMetadata> functions)
```
kshyju, line 194: See we already have an extension method in the repo which does this on `IEnumerable<T>` type.

src/WebJobs.Script/Host/AggregateFunctionMetadataProvider.cs
```diff
@@ -17,44 +18,65 @@
...
 
                     // get function metadata from worker, then validate it
                     rawFunctions = await _dispatcher.GetWorkerMetadata();
-                    functions = ValidateMetadata(rawFunctions);
 
-                    // set up invocation buffers and send load requests
-                    await _dispatcher.FinishInitialization(functions);
+                    if (!IsNullOrEmpty(rawFunctions))
+                    {
+                        functions = ValidateMetadata(rawFunctions);
```
kshyju, line 67: This comment is for line 86 😁 on which I am not able to leave the comment.
Use `IEnumerable.Any()` instead of `Count()`

src/WebJobs.Script/Host/AggregateFunctionMetadataProvider.cs
kshyju: Same for the next list (functions) as well.

src/WebJobs.Script/Host/AggregateFunctionMetadataProvider.cs
```diff
@@ -17,44 +18,65 @@
...
+                        // set up invocation buffers and send load requests
+                        await _dispatcher.FinishInitialization(functions);
+                    }
+                }
+                if (IsNullOrEmpty(rawFunctions) || !workerIndexing)
+                {
+                    // If worker denies indexing then falling back to the host for Indexing
```
kshyju, line 74: Should we also log this info?

src/WebJobs.Script/Host/AggregateFunctionMetadataProvider.cs
kshyju: You could also consider using the IsNullOrEmpty method after changing to support IEnumerable<T>.

src/WebJobs.Script/Host/AggregateFunctionMetadataProvider.cs
```diff
@@ -155,5 +177,14 @@ internal static FunctionMetadata ValidateBindings(IEnumerable<string> rawBinding
 
             return function;
         }
+
+        private bool IsNullOrEmpty(IEnumerable<RawFunctionMetadata> functions)
+        {
+            if (functions == null || (functions != null && !functions.Any()))
```
kshyju: I think you can remove the `functions != null` expression.

src/WebJobs.Script.Grpc/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
surgupta-msft: Yes, that's right.

src/WebJobs.Script.Grpc/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
surgupta-msft: Yes, correct.

src/WebJobs.Script.Grpc/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
surgupta-msft: This naming came out from a discussion with few folks. Will take your feedback and check again.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
surgupta-msft: This name had been discussed in broader sync and PR was merged in protobuf repo for it - https://github.com/Azure/azure-functions-language-worker-protobuf/pull/61. If there are any major concerns, we can again bring it up with broader group.

src/WebJobs.Script/Host/AggregateFunctionMetadataProvider.cs
surgupta-msft: Discussed offline. _dispatcher can be null if using hostIndexing and it cannot be null if using workerIndexing.

src/WebJobs.Script/Host/AggregateFunctionMetadataProvider.cs
surgupta-msft: updated the code here.

src/WebJobs.Script.Grpc/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
surgupta-msft: I tried this several times. Somehow, the pulling updates is not working for me (though I have done this few months back successfully). Let me know if you have done this recently. We can chat offline and get this done.

src/WebJobs.Script/Host/FunctionMetadataManager.cs
surgupta-msft: I agree with the issue on async pattern and it not being highlighted in PR diff. I will discuss with folks in the team about things to do to improve the pattern here.

src/WebJobs.Script/Host/AggregateFunctionMetadataProvider.cs
```diff
@@ -17,51 +18,73 @@
...
+                        functions = ValidateMetadata(rawFunctions);
+                        // set up invocation buffers and send load requests
+                        await _dispatcher.FinishInitialization(functions);
+                    }
+                }
+                if (!workerIndexing || IsDefaultIndexingRequired(rawFunctions))
+                {
+                    // If worker denies indexing then falling back to the host for Indexing
+                    _logger.LogDebug("Fallback to host indexing as worker denied indexing");
+                    functions = await _hostFunctionMetadataProvider.GetFunctionMetadataAsync(workerConfigs, SystemEnvironment.Instance, forceRefresh);
```
NickCraver: Should this be using `_environment` over `SystemEnvironment.Instance`? There's a mismatch here in methods that take an instance but then pass the global static anyway. It seems like we should either be using the global static everywhere and not pass it anywhere, or the other way around.

surgupta-msft: Agreed. Corrected the code here.

soninaren (approved):

liliankasem (approved):

PR #7985, opened 2021-12-13
fixing concurrency bug in CacheMetadataResolver

### LINKED ISSUE #7574: DotNet Applications without pre-compilation can sometimes cause a concurrency bug leading to permanent unhealthy state.
#### Problem Description
During servicing loop, a bug was discovered that can occasionally occur in Dotnet Function Applications that do not use pre-compilation. During Function Invocation, sometimes a concurrent operation is executed on a non-concurrent collection.  This causes the function to become unhealthy but does not affect the application as a whole. Therefore, the function app is not migrated to a different host and executions of the unhealthy function will fail until the app is reset. 

When executing the [TryGetValue on the referenceCache linked here](https://github.com/Azure/azure-functions-host/blob/852eed9ceef6ef56b431428a8eb31f1bd9c97f3b/src/WebJobs.Script/Description/CacheMetadataResolver.cs#L34) FunctionsHost occasionally throws an error related to concurrency shown below.  This causes the function to become permanently unhealthy until app restart.  The full stack of this error can be viewed below. 
Error :
The 'FUNCTIONAME' function is in error: Operations that change non-concurrent collections must have exclusive access. A concurrent update was performed on this collection and corrupted its state. The collection's state is no longer correct.

Full Stack Trace : 
image

Unable to provide PII for the function. 

- Timestamp: (2021-07-08 12:00:00)-(2021-07-09 14:00:00)
- Function App version: 3.0.15733
- Function App name: N/A
- Function name(s) (as appropriate): N/A 
- Invocation ID: 126cfc74-b2a2-4aa3-bb68-ee34cc7661c7
- Region: WEU

#### Repro steps

1. Generate a dotnet function application without pre-compilation
2. Invoke the function
3. Function loads (if necessary) and executes

#### Expected behavior

When the customer invokes the function, it should start a healthy function and return a response. 

#### Actual behavior

The function sometimes experiences a concurrency issue and does not start. Returns a 404

#### Known workarounds

The suggested mitigation at this time was to reset the function application when the customer experiences the issue.

#### Related information 

* Programming language used : Dotnet
* Bindings used : HttpTrigger

brettsam:
Fixes #7574

### DIFF
--- a/src/WebJobs.Script/Description/CacheMetadataResolver.cs
+++ b/src/WebJobs.Script/Description/CacheMetadataResolver.cs
@@ -2,7 +2,7 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Generic;
+using System.Collections.Concurrent;
 using System.Collections.Immutable;
 using Microsoft.CodeAnalysis;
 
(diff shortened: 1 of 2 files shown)


brettsam:
Was able to confirm the test fails in CI (and it fails 100% of the time for me locally). Now I'll switch to `ConcurrentDictionary`.

image

kshyju (approved):

liliankasem (approved):

fabiocav (approved):

PR #8004, opened 2021-12-15
fixing build numbering for releases

brettsam:
For releases (those in the `release/...` branch), we are not supposed to append the build version to the build number. This affects tags and the automated release. The behavior was already built into the powershell module, but I wasn't correctly calling it in this situation.

### DIFF
--- a/build/common.props
+++ b/build/common.props
@@ -5,7 +5,7 @@
     <LangVersion>latest</LangVersion>
     <MajorVersion>4</MajorVersion>
     <MinorVersion>1</MinorVersion>
-    <PatchVersion>0</PatchVersion>
+    <PatchVersion>1</PatchVersion>
     <BuildNumber Condition="'$(BuildNumber)' == '' ">0</BuildNumber>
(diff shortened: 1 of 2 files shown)


build/common.props
```diff
@@ -5,7 +5,7 @@
     <LangVersion>latest</LangVersion>
     <MajorVersion>4</MajorVersion>
     <MinorVersion>1</MinorVersion>
-    <PatchVersion>0</PatchVersion>
+    <PatchVersion>1</PatchVersion>
```
yojagad, line 8: I'm thinking after cherry picking this into hotfix branch, this will read
4
2
0

?

yojagad (approved):

build/common.props
brettsam: release/4.x has 4.1.0 in it so i think this should work as expected when cherry-picked

https://github.com/Azure/azure-functions-host/blob/release/4.x/build/common.props#L6-L8

fabiocav (approved):

PR #8031, opened 2022-01-06
SharedMemoryMapTests: optimize GetStream_VerifyContentMatches

NickCraver:
The comparison here was an implementation we can optimize lowering the test runtime by reducing allocations, copies, and a suboptimal comparison (.NET 6 further optimizes sequence comparisons). I don't see a quick way to optimize it further that doesn't compromise what we want to test, so stopping here. This is out of an 8 min runtime locally for all test (I don't have everything configured) so overall decent savings for test runs.

If you're looking just above these changes at:
```cs
[InlineData(1024 * 1024 * 1024, Skip = "Results in OOM when run on test infra")] // 1GB
```
...then why doesn't the 1GB test skip locally? I'm not sure, but it runs which is why I looked at this. I'll poke at that, but it still helps shave off the 256MB even with it fixed later.

### Pull request checklist

Before:
image
After:
image


NickCraver:
Test failures aren't related - random flakiness both runs.

gohar94 (dismissed):
Thanks, Nick.

test/WebJobs.Script.Tests/Workers/SharedMemoryMapTests.cs
```diff
@@ -209,14 +211,17 @@ public async Task GetStream_VerifyContentMatches(int contentSize)
 
             Stream readContent = await sharedMemoryMap.GetStreamAsync();
 
-            // Read content into MemoryStream to easily put them into a byte[].
-            using (MemoryStream tempStream = new MemoryStream())
+            static IEnumerable<byte> StreamToEmulerable(Stream stream)
```
surgupta-msft: nit: Enumerable typo in StreamToEmulerable

surgupta-msft (dismissed):

test/WebJobs.Script.Tests/Workers/SharedMemoryMapTests.cs
NickCraver: I...don't know what happened there :( Fixed!

gohar94 (approved):

surgupta-msft (approved):

PR #8045, opened 2022-01-09
Preserve IsPackable Property if Set

### LINKED ISSUE #8044: Preserve IsPackable Property if Set
While not common, there are some scenarios where a developer may wish to publish their Azure Functions as NuGet packages. However, there are a couple targets files in the toolchain that set the `<IsPackable>` property to `false`. This has already been talked about in other repositories, like here https://github.com/Azure/azure-functions-vs-build-sdk/issues/110, where the targets file [`Microsoft.NET.Sdk.Functions.targets`](https://github.com/Azure/azure-functions-vs-build-sdk/commit/0992e21cc695c18b08bcf9de22ca0dc2dcab9713#diff-60eb2d9d1ecbc014c8c0a62346773539L20-L35) now respects an existing value for `<IsPackable>` using a condition. However, the targets file [`Microsoft.Azure.WebJobs.Script.ExtensionsMetadataGenerator.targets`](https://github.com/Azure/azure-functions-host/blob/dev/tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator/Targets/Microsoft.Azure.WebJobs.Script.ExtensionsMetadataGenerator.targets) requires a similar change.

Microsoft.Azure.WebJobs.Script.ExtensionsMetadataGenerator.targets should be updated to respect any existing value for `IsPackable`. Eg. `<IsPackable Condition="'$(IsPackable)' == ''">false</IsPackable>`

You can of course set `<IsPackable>true</IsPackable>` in a `Directory.Build.targets` which should resolve after all of the imported targets file.

wsugarman:

Resolves https://github.com/Azure/azure-functions-host/issues/8044

### Pull request checklist


fabiocav:
Thanks!

PR #8048, opened 2022-01-10
improve scoping of FunctionRpcService; remove call to CreateDefaultBu…

brettsam:
1. Using the root-registered `FunctionRpcService` when running the `AspNetCoreGrpcService`. We were missing logs because the new inner Grpc host did not have all of the root loggers registered. By using this service from the root scope, the logger is automatically wired up as we expect.
2. Removing call to `CreateDefaultBuilder()` for the host running the `AspNetCoreGrpcService`. This was having cold start impact in single-core Linux instances, which we've attributed to the FileWatchers set up by the default builder.


liliankasem (approved):

fabiocav (approved):

fabiocav:
@pragnagopa / @balag0 FYI: merging this. We can pull this into a hotfix if needed.

PR #8054, opened 2022-01-11
Sending single load request in case of multiple functions

### LINKED ISSUE #7880: Sending single FunctionLoadRequest in case of multiple functions
Host should be sending single FunctionLoadRequest even if multiple functions are there in Function App. Need to investigate if current code already has this behavior and fix if it does not already exist.

surgupta-msft:

resolves #7880 
PR in Protobuf repo - https://github.com/Azure/azure-functions-language-worker-protobuf/pull/63

Host should have an ability to send single FunctionLoadRequest in case of multiple functions in Function App. Adding new message that can send a list of FunctionLoadRequest and new methods to process that in GRPCWorkerChannel.cs


src/WebJobs.Script.Grpc/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
```diff
@@ -79,6 +79,9 @@ message StreamingMessage {
     // Worker indexing message types
     FunctionsMetadataRequest functions_metadata_request = 29;
     FunctionMetadataResponse function_metadata_response = 30;
+
+    // Host sends required metadata to worker to load functions
+    FunctionLoadRequests function_load_requests = 31;
```
surgupta-msft: PR in protobuf repo - https://github.com/Azure/azure-functions-language-worker-protobuf/pull/63

satvu (dismissed):
My only hesitation is that some of the new gRPC message types and methods added have the same names as older messages/methods but with an extra 's' at the end. I feel like there's potential for confusion there. 

Is the plan to keep the current model of sending function requests to workers one at a time? Will this new model ever move from a capability to the standard behavior, replacing the current one?

src/WebJobs.Script.Grpc/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
```diff
@@ -220,7 +223,12 @@ message CloseSharedMemoryResourcesResponse {
   map<string, bool> close_map_results = 1;
 }
 
-// Host tells the worker to load a Function
+// Host tells the worker to load a list of Functions
+message FunctionLoadRequests {
```
kshyju: Should this be singular? `FunctionLoadRequest` ? The child property can be plural as it is a collection type.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -299,13 +299,47 @@ public void SendFunctionLoadRequests(ManagedDependencyOptions managedDependencyO
                     _eventSubscriptions.Add(_inboundWorkerEvents.Where(msg => msg.MessageType == MsgType.FunctionLoadResponse)
                         .Subscribe((msg) => LoadResponse(msg.Message.FunctionLoadResponse), HandleWorkerFunctionLoadError));
                 }
-                foreach (FunctionMetadata metadata in _functions.OrderBy(metadata => metadata.IsDisabled()))
+
+                // Check if the worker supports this feature
+                bool capabilityEnabled = !string.IsNullOrEmpty(_workerCapabilities.GetCapabilityState(RpcWorkerConstants.AcceptsListOfFunctionLoadRequests));
+                if (capabilityEnabled)
+                {
+                    SendFunctionLoadRequestsList(_functions, managedDependencyOptions);
```
kshyju: naming nit: function name sounds like it sends multiple requests. but in reality, it sends only one grpc request. Should we use `SendFunctionLoadRequest`?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -299,13 +299,47 @@ public void SendFunctionLoadRequests(ManagedDependencyOptions managedDependencyO
                     _eventSubscriptions.Add(_inboundWorkerEvents.Where(msg => msg.MessageType == MsgType.FunctionLoadResponse)
                         .Subscribe((msg) => LoadResponse(msg.Message.FunctionLoadResponse), HandleWorkerFunctionLoadError));
                 }
-                foreach (FunctionMetadata metadata in _functions.OrderBy(metadata => metadata.IsDisabled()))
+
+                // Check if the worker supports this feature
+                bool capabilityEnabled = !string.IsNullOrEmpty(_workerCapabilities.GetCapabilityState(RpcWorkerConstants.AcceptsListOfFunctionLoadRequests));
```
kshyju, line 304: With this logic, the worker should not send "false"  for "AcceptsListOfFunctionLoadRequests" and expect it to be disabled. correct?

kshyju: You may also consider using [IsNullOrWhiteSpace](https://docs.microsoft.com/en-us/dotnet/api/system.string.isnullorwhitespace?redirectedfrom=MSDN&view=net-6.0#remarks)

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -299,13 +299,47 @@ public void SendFunctionLoadRequests(ManagedDependencyOptions managedDependencyO
...
+                    }
                 }
             }
         }
 
+        internal void SendFunctionLoadRequestsList(IEnumerable<FunctionMetadata> functions, ManagedDependencyOptions managedDependencyOptions)
+        {
+            _functionLoadRequestResponseEvent = _metricsLogger.LatencyEvent(MetricEventNames.FunctionLoadRequestResponse);
+
+            // send load requests for the registered functions
```
kshyju, line 327: Should we also do `_workerChannelLogger.LogDebug` similar to what happens in line 386? If so, your UT can be fixed to verify that as the expected log message.

kshyju (dismissed):

src/WebJobs.Script.Grpc/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
satvu: `FunctionLoadRequest` singular already exists and is used in the current implementation for sending each function load request one by one from host to worker. The definition is a few lines down in this file. 

In order to not break the current implementation but provide a path for improving performance, `FunctionLoadRequests` wraps the singular `FunctionLoadRequest` messages that are currently being sent one by one into one payload instead. Workers will have to opt into this using the capability @surgupta-msft added. 

Please let me know if I understand correctly @surgupta-msft ... this is also what I meant about the naming confusion.

liliankasem: Agree, please make sure this is singular

src/WebJobs.Script.Grpc/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
```diff
@@ -220,7 +223,12 @@ message CloseSharedMemoryResourcesResponse {
   map<string, bool> close_map_results = 1;
 }
 
-// Host tells the worker to load a Function
+// Host tells the worker to load a list of Functions
```
liliankasem: I want to make sure we follow the right proto update practice here, ideally the [proto PR goes in first ](https://github.com/Azure/azure-functions-language-worker-protobuf/pull/63), then we have a PR in the host to just bump the version. I'd like to keep the proto changes separate from the code changes in case we need to revert

src/WebJobs.Script.Grpc/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
liliankasem: That makes sense, I misread the code :)

src/WebJobs.Script.Grpc/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
surgupta-msft: Agreed.

Discussed offline about errors I am getting while following documentation due to unrelated histories. We will try again to bump the version after files are same.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
surgupta-msft: yes

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
liliankasem: I think `SendFunctionLoadRequestCollection / List` is good (removing s from requests

src/WebJobs.Script.Grpc/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
surgupta-msft: @satvu yes your understanding is correct. I have updated the code as well to avoid naming confusion.

surgupta-msft:
>

surgupta-msft:
> My only hesitation is that some of the new gRPC message types and methods added have the same names as older messages/methods but with an extra 's' at the end. I feel like there's potential for confusion there.
> 
> Is the plan to keep the current model of sending function requests to workers one at a time? Will this new model ever move from a capability to the standard behavior, replacing the current one?

@satvu updated the code to avoid naming confusion.

liliankasem:
If you rebase after this PR goes in, you can remove protobuf changes from this PR

https://github.com/Azure/azure-functions-host/pull/8072

test/WebJobs.Script.Tests/Workers/Rpc/GrpcWorkerChannelTests.cs
```diff
@@ -356,6 +356,31 @@ public void SendLoadRequests_PublishesOutboundEvents()
...
+        {
+            StartStream startStream = new StartStream()
+            {
+                WorkerId = _workerId
+            };
+            StreamingMessage startStreamMessage = new StreamingMessage()
+            {
+                StartStream = startStream
+            };
+            GrpcEvent rpcEvent = new GrpcEvent(_workerId, startStreamMessage);
```
liliankasem, line 370: nit: can we add some new lines to clean this up for easier readability? not a big deal but it's nicer to read :)

``` 
// Arrange

// Act

// Assert

liliankasem (approved):

satvu (approved):

PR #8056, opened 2022-01-12
RpcFunctionInvocationDispatcher: capture async void errors

NickCraver:
In test runs we're seeing failures like this:
```
The active test run was aborted. Reason: Test host process crashed : Unhandled exception. System.Threading.Tasks.TaskCanceledException: A task was canceled.
   at Microsoft.Azure.WebJobs.Script.Grpc.GrpcWorkerChannel.StartWorkerProcessAsync(CancellationToken cancellationToken) in /_/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs:line 155
   at Microsoft.Azure.WebJobs.Script.Workers.Rpc.RpcFunctionInvocationDispatcher.InitializeJobhostLanguageWorkerChannelAsync(Int32 attemptCount) in /_/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs:line 119
   at Microsoft.Azure.WebJobs.Script.Workers.Rpc.RpcFunctionInvocationDispatcher.StartWorkerChannel(String runtime) in /_/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs:line 535
   at Microsoft.Azure.WebJobs.Script.Workers.Rpc.RpcFunctionInvocationDispatcher.StartWorkerChannel(String runtime) in /_/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs:line 548
   at Microsoft.Azure.WebJobs.Script.Workers.Rpc.RpcFunctionInvocationDispatcher.DisposeAndRestartWorkerChannel(String runtime, String workerId, Exception workerException) in /_/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs:line 497
   at Microsoft.Azure.WebJobs.Script.Workers.Rpc.RpcFunctionInvocationDispatcher.WorkerError(WorkerErrorEvent workerError) in /_/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs:line 442
   at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__128_1(Object state)
   at System.Threading.QueueUserWorkItemCallbackDefaultContext.Execute()
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
   at System.Threading.PortableThreadPool.WorkerThread.WorkerThreadStart()
   at System.Threading.Thread.StartCallback()
```

This is overall because of the async void usage in the event handlers for the `ScriptEvent` pipe. By kicking off a task that could error (what `async void` does under the covers), its only option is to crash the runtime when an error happens. We have the options of:
1. Let the runtime crash (what happens today)
2. Capture only the `TaskCancelledException` case (what this PR does)
3. Capture and log all errors

The problem with #3 is it introduces a state I'm tooo naïve to reason about: what happens when an error happens _and we fail to restart the worker_? This could be a downstream net win from a state standpoint of "well, we restart everything and recover". It's not fast though, vs. restarting just the worker. If that's _not_ a new state and it's handled correctly, 3 is the better/global option.

The interaction here with the test suite is:
1. We throw an worker error (in testing)
2. An event is triggered kicking off an `async void` to restart the worker
3. We dispose of the Rpc bits (when the test finishes)
4. That background test is cancelled, throwing an error and crashing our runtime

I think it's worth noting this is fixing this class only - similar usages (grep `async void`) lie in other areas and is something we should address globally. Writing this up to make sure the direction is correct/agreeable as a first step.

### Pull request checklist

cc @brettsam not sure about backport here - if the case exists and this is causing any prod troubles, happy to help.

### DIFF
--- a/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
+++ b/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
@@ -435,16 +435,26 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
                 return;
             }
 
-            if (string.Equals(_workerRuntime, workerError.Language))
+            try
             {
-                _logger.LogDebug("Handling WorkerErrorEvent for runtime:{runtime}, workerId:{workerId}. Failed with: {exception}", workerError.Language, _workerRuntime, workerError.Exception);
-                AddOrUpdateErrorBucket(workerError);
-                await DisposeAndRestartWorkerChannel(workerError.Language, workerError.WorkerId, workerError.Exception);
+                if (string.Equals(_workerRuntime, workerError.Language))
+                {
+                    _logger.LogDebug("Handling WorkerErrorEvent for runtime:{runtime}, workerId:{workerId}. Failed with: {exception}", workerError.Language, _workerRuntime, workerError.Exception);
+                    AddOrUpdateErrorBucket(workerError);
+                    await DisposeAndRestartWorkerChannel(workerError.Language, workerError.WorkerId, workerError.Exception);
+                }
+                else
+                {
+                    _logger.LogDebug("Received WorkerErrorEvent for runtime:{runtime}, workerId:{workerId}", workerError.Language, workerError.WorkerId);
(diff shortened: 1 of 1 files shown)


pragnagopa (approved):
Changes LGTM. Scoping the changes to `TaskCanceledException` for now makes sense. Please wait for approval from @brettsam as well.

fabiocav (approved):

PR #8058, opened 2022-01-13
Prototyping Default Value Fetching for SyncTriggerPayload

TsuyoshiUshio:
This is the Draft PR for sharing the idea of how to fetch Default Value for host.json parameters. 

For more information refer to 
https://github.com/Azure/azure-webjobs-sdk/pull/2815


TsuyoshiUshio:
We use the different architecture with this one. close.

PR #8061, opened 2022-01-13
Emitting metric to identify .NET applications not using R2R

### LINKED ISSUE #7901: Risk Alert if deployed customer function app is not deployed with Ready2Run
Need to log a statement if .NET Function app is not deployed with Ready2Run. This log entry will be read and alert to customers in cold start Risk Alert section of the portal.

Code to check for the condition:

using System.Reflection.Metadata;
using System.Reflection.PortableExecutable;

static bool IsReadyToRun(Stream stream)
{
    try
    {
        using (PEReader peFile = new PEReader(stream))
        {
            return peFile.PEHeaders.CorHeader?.ManagedNativeHeaderDirectory.Size != 0;
        }
    }
    catch
    {
        return false;
    }
}

fabiocav:
This change adds logic to perform analysis of function assemblies to identify whether they have been R2R optimized. This enables us to emit telemetry that will power Risk Alerts and other notifications to provide guidance to .NET Function Apps not using R2R

No tests checked in as part of this change, but I ran through validation with a number of scenarios and wanted to get this in to give @safihamid a chance to look and make sure we can get this in the next release. I'll follow up with additional tests.

resolves #7901 

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.WebHost/AssemblyAnalysis/AssemblyAnalysisService.cs
+++ b/src/WebJobs.Script.WebHost/AssemblyAnalysis/AssemblyAnalysisService.cs
@@ -1,0 +1,159 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Reflection.PortableExecutable;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.AssemblyAnalyzer
+{
+    internal class AssemblyAnalysisService : IHostedService, IDisposable
+    {
+        private readonly IEnvironment _environment;
+        private readonly ILoggerFactory _loggerFactory;
+        private readonly IOptionsMonitor<StandbyOptions> _standbyOptionsMonitor;
+        private readonly WebJobsScriptHostService _scriptHost;
+        private CancellationTokenSource _cancellationTokenSource;
+        private Task _analysisTask;
+        private bool _disposed;
+        private bool _analysisScheduled;
+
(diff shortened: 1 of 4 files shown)


src/WebJobs.Script.WebHost/AssemblyAnalysis/AssemblyAnalysisService.cs
```diff
@@ -0,0 +1,158 @@
...
+            {
+                if (_cancellationTokenSource.IsCancellationRequested)
+                {
+                    return;
+                }
+
+                if (item.Metadata.Properties.TryGetValue(ScriptConstants.FunctionMetadataDirectTypeKey, out Type type)
+                    && !assemblies.Contains(type.Assembly))
+                {
+                    if (!IsReadyToRun(type.Assembly))
```
safihamid: ####                     if (!IsReadyToRun(type.Assembly))

---

There is no need to loop thru all Functions, if one is not optimized the rest will be not optimized as well.

I suggest we exit the loop if/when we detected one and logged it. that's enough for Risk Alerts. no? #Resolved

fabiocav: For most applications, that will be the case, but not all functions have to come from the same assembly, and although it’s probably safe to assume that in the normal publishing flow, either all assemblies defining functions would be either be optimized or all of them wouldn’t, there are scenarios where that could deviate from that, though, but pretty unlikely ones.

I agree with you that for the risk alerts, simply flagging that we have unoptimized assemblies and exiting would be sufficient to meet our requirements, and debated doing that, but ended up doing this as I thought it would be helpful to also identify all of the assemblies and provide that information. In the vast majority of cases, we’ll be dealing with a single assembly for all functions, so the actual analysis would only be performed once anyway, and given the average number of functions per app and how this runs, I don’t believe we would see significant impact if we exit early, but either way is really fine for our purposes. 

I don’t think the current approach is a problem, and it might be good to list all assemblies, but I’m happy to change to simply flag the app, as I agree this is not super important. Thoughts?

fabiocav: Made the change to flag the app only. We can come back to this and add assembly names in the future if needed.

safihamid: ok sounds good.

safihamid (dismissed):
:shipit:

safihamid (approved):

brettsam (approved):

PR #8067, opened 2022-01-15
Updating end to end test to consume cancellation token with drain mode

soninaren:

resolves https://github.com/Azure/azure-webjobs-sdk/pull/2821

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/sample/CSharp/HttpTrigger-Cancellation/run.csx
+++ b/sample/CSharp/HttpTrigger-Cancellation/run.csx
@@ -6,21 +6,7 @@ using Microsoft.Extensions.Primitives;
 public static async Task<IActionResult> Run(HttpRequest req, ILogger log, CancellationToken token)
 {
     log.LogInformation("C# HTTP trigger cancellation function processing a request.");
-
-    try
-    {
-        await Task.Delay(10000, token);
-        log.LogInformation("Function invocation successful.");
-        return new OkResult();
-
-    }
-    catch (OperationCanceledException)
-    {
-        log.LogInformation("Function invocation cancelled.");
-        return new NotFoundResult();
-    }
-    catch (Exception e)
-    {
-        return new StatusCodeResult(500);
-    }
+    await Task.Delay(10000, token);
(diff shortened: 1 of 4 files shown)


sample/CSharp/HttpTrigger-Cancellation/run.csx
```diff
@@ -5,22 +5,8 @@ using Microsoft.Extensions.Primitives;
 
 public static async Task<IActionResult> Run(HttpRequest req, ILogger log, CancellationToken token)
```
liliankasem, line 6: Do we even need a cancellation sample if we've gotten rid of the cancellation token?

soninaren: In a way this test is not needed as it is testing for something that should not happen. However, since we are explicitly removing the cancellation token, this helps validate that change.

liliankasem: Then it would make sense to keep the sample code the same, but change the assert to check that it returns OK, right?

soninaren: We can certainly do that. The only reason i changed the sample code is because the original function code seems to indicate that it expects function OperationCanceledException. Somone reading the test / sample code without context would be confused as why it is there in the first place.

soninaren: Have added a tracking item for test changes.  https://github.com/Azure/azure-functions-host/issues/8084.

fabiocav: @liliankasem flagging this as resolved as @soninaren will continue to iterate and follow up on your concerns using the issue he created.

fabiocav (approved):

brettsam (approved):

nzthiago:
Related WebJobs SDK PR https://github.com/Azure/azure-webjobs-sdk/pull/2821

PR #8070, opened 2022-01-18
Sync host with Protobuf - removing retry_options and config_source

surgupta-msft:
Host protobuf file has extra fields as compared to Protobuf repo. Removing those fields to bring them in sync.
Discussion - https://github.com/Azure/azure-functions-host/issues/7849

### DIFF
--- a/src/WebJobs.Script.Grpc/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
+++ b/src/WebJobs.Script.Grpc/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
@@ -274,12 +274,6 @@ message RpcFunctionMetadata {
   // Raw binding info
   repeated string raw_bindings = 10;
 
-  // Retry Options: string representation of JObject retry options (maxRetryCount, intervals, etc.)
-  string retry_options = 11;
-
-  // Configuration Source: string representation of JToken configuration source property in function metadata
-  string config_source = 12;
-
   // unique function identifier (avoid name collisions, facilitate reload case)
   string function_id = 13;
 
--- a/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
@@ -525,8 +525,6 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
                     {
                         Metadata = functionMetadata,
                         Bindings = bindings,
-                        RetryOptions = metadata.RetryOptions,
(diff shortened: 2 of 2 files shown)


liliankasem (approved):
In future let's make sure we're all following [these steps](https://github.com/Azure/azure-functions-language-worker-protobuf#pulling-updates) to get the proto in sync. Ideally the proto change goes into the protobuf repo, and then we have a single PR in the host to bump the version

src/WebJobs.Script.Grpc/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
```diff
@@ -274,12 +274,6 @@ message RpcFunctionMetadata {
...
   repeated string raw_bindings = 10;
 
-  // Retry Options: string representation of JObject retry options (maxRetryCount, intervals, etc.)
-  string retry_options = 11;
-
-  // Configuration Source: string representation of JToken configuration source property in function metadata
-  string config_source = 12;
-
   // unique function identifier (avoid name collisions, facilitate reload case)
   string function_id = 13;
```
kshyju, line 278: GRPC newbie question: Does the order change of properties has any impact? (11 and 12 are gone now)

surgupta-msft: even 5 is not there in this message.

liliankasem: Field numbers are an important part of Protobuf. They're used to identify fields in the binary encoded data, which means they can't change from version to version of your service. The advantage is that backward compatibility and forward compatibility are possible. Clients and services will ignore field numbers that they don't know about, as long as the possibility of missing values is handled.

https://docs.microsoft.com/en-us/dotnet/architecture/grpc-for-wcf-developers/protobuf-messages

surgupta-msft: @soninaren Any thoughts on this?

liliankasem: If I understood correctly, we shouldn't change the numbers to keep the protobuf backwards compatible

kshyju: Thank you!

kshyju (approved):

PR #8071, opened 2022-01-19
Update key generation to provide fixed signatures and encoded checksums (for improved security tooling)

### LINKED ISSUE #7662: Consider updating key generation to increase identifiability (and glean security value from improved tooling)

michaelcfanning:
Resolves #7662

### Pull request checklist

@liliankasem @fabiocav @brettsam @mathewc 

Will send some additional context offline.

### DIFF
--- a/src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
+++ b/src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
@@ -630,12 +630,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
 
         internal static string GenerateIdentifiableSecret(ulong seed)
         {
-            string secret = IdentifiableSecrets.GenerateIdentifiableKey(seed, 40, AzureFunctionsSignature);
-
-            secret = secret.Replace('+', '-');
-            secret = secret.Replace('/', '_');
-
-            return secret;
+            return IdentifiableSecrets.GenerateBase64Key(seed, 40, AzureFunctionsSignature, encodeForUrl: true);
         }
 
         private void OnSecretsChanged(object sender, SecretsChangedEventArgs e)
--- a/src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
+++ b/src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
@@ -16,6 +16,8 @@ using Microsoft.Azure.WebJobs.Extensions.Http;
 using Microsoft.Azure.WebJobs.Script.Diagnostics;
 using Microsoft.Azure.WebJobs.Script.WebHost.Properties;
 using Microsoft.Extensions.Logging;
+using Microsoft.Security.Utilities;
+
 using DataProtectionCostants = Microsoft.Azure.Web.DataProtection.Constants;
 
 namespace Microsoft.Azure.WebJobs.Script.WebHost
@@ -35,6 +37,24 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
         private IMetricsLogger _metricsLogger;
         private string _repositoryClassName;
         private DateTime _lastCacheResetTime;
+
+        internal const string AzureFunctionsSignature = "AzFu";
+
+        // Checksum seed for master keys. The string transformation here results in a
+        // a seed that is versioned in the least significant bits (as a ulong):
+        // BitConverter.ToUInt64(Encoding.ASCII.GetBytes("Master00").Reverse().ToArray());
+        // If there is a need to version, use "Master01" as the next string literal.
+        internal const ulong MasterKeySeed = 0x4d61737465723030;
+
+        // Checksum seed for system keys.
+        // BitConverter.ToUInt64(Encoding.ASCII.GetBytes("System00").Reverse().ToArray());
+        // If there is a need to version, use "System01" as the next string literal.
+        internal const ulong SystemKeySeed = 0x53797374656d3030;
+
+        // Checksum seed for function keys.
+        // BitConverter.ToUInt64(Encoding.ASCII.GetBytes("Functi00").Reverse());
+        // If there is a need to version, use "Functi01" next.
+        internal const ulong FunctionKeySeed = 0x46756e6374693030;
 
         // for testing
         public SecretManager()
@@ -254,7 +274,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
                 if (value == null)
                 {
                     // Generate a new secret (clear)
-                    masterKey = GenerateSecret();
+                    masterKey = GenerateIdentifiableSecret(MasterKeySeed);
                     result = OperationResult.Created;
                 }
                 else
@@ -303,7 +323,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
         {
             OperationResult result = OperationResult.NotFound;
 
-            secret = secret ?? GenerateSecret();
+            secret = secret ?? GenerateIdentifiableSecret(FunctionKeySeed);
 
             await ModifyFunctionSecretsAsync(secretsType, keyScope, secrets =>
             {
@@ -493,10 +513,10 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
         {
             return new HostSecrets
             {
-                MasterKey = GenerateKey(ScriptConstants.DefaultMasterKeyName),
+                MasterKey = GenerateKey(ScriptConstants.DefaultMasterKeyName, MasterKeySeed),
                 FunctionKeys = new List<Key>
             {
-                GenerateKey(ScriptConstants.DefaultFunctionKeyName)
+                GenerateKey(ScriptConstants.DefaultFunctionKeyName, FunctionKeySeed)
             },
                 SystemKeys = new List<Key>()
             };
@@ -506,10 +526,10 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
         {
             if (secrets.MasterKey.IsEncrypted)
             {
-                secrets.MasterKey.Value = GenerateSecret();
+                secrets.MasterKey.Value = GenerateIdentifiableSecret(MasterKeySeed);
             }
-            secrets.SystemKeys = RegenerateKeys(secrets.SystemKeys);
-            secrets.FunctionKeys = RegenerateKeys(secrets.FunctionKeys);
+            secrets.SystemKeys = RegenerateKeys(secrets.SystemKeys, SystemKeySeed);
+            secrets.FunctionKeys = RegenerateKeys(secrets.FunctionKeys, FunctionKeySeed);
             return secrets;
         }
 
@@ -519,24 +539,24 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
             {
                 Keys = new List<Key>
                 {
-                    GenerateKey(ScriptConstants.DefaultFunctionKeyName)
+                    GenerateKey(ScriptConstants.DefaultFunctionKeyName, FunctionKeySeed)
                 }
             };
         }
 
         private FunctionSecrets GenerateFunctionSecrets(FunctionSecrets secrets)
         {
-            secrets.Keys = RegenerateKeys(secrets.Keys);
+            secrets.Keys = RegenerateKeys(secrets.Keys, FunctionKeySeed);
             return secrets;
         }
 
-        private IList<Key> RegenerateKeys(IList<Key> list)
+        private IList<Key> RegenerateKeys(IList<Key> list, ulong seed)
         {
             return list.Select(k =>
             {
                 if (k.IsEncrypted)
                 {
-                    k.Value = GenerateSecret();
+                    k.Value = GenerateIdentifiableSecret(seed);
                 }
                 return k;
             }).ToList();
@@ -594,9 +614,9 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
             };
         }
 
-        private Key GenerateKey(string name = null)
+        private Key GenerateKey(string name, ulong seed)
         {
-            string secret = GenerateSecret();
+            string secret = GenerateIdentifiableSecret(seed);
 
             return CreateKey(name, secret);
         }
@@ -608,17 +628,14 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
             return _keyValueConverterFactory.WriteKey(key);
         }
 
-        internal static string GenerateSecret()
+        internal static string GenerateIdentifiableSecret(ulong seed)
         {
-            using (var rng = RandomNumberGenerator.Create())
-            {
-                byte[] data = new byte[40];
-                rng.GetBytes(data);
-                string secret = Convert.ToBase64String(data);
+            string secret = IdentifiableSecrets.GenerateIdentifiableKey(seed, 40, AzureFunctionsSignature);
 
-                // Replace pluses as they are problematic as URL values
-                return secret.Replace('+', 'a');
-            }
+            secret = secret.Replace('+', '-');
+            secret = secret.Replace('/', '_');
+
+            return secret;
         }
 
         private void OnSecretsChanged(object sender, SecretsChangedEventArgs e)
--- a/src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
+++ b/src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
@@ -18,7 +18,7 @@ using Microsoft.Azure.WebJobs.Script.WebHost.Properties;
 using Microsoft.Extensions.Logging;
 using Microsoft.Security.Utilities;
 
-using DataProtectionCostants = Microsoft.Azure.Web.DataProtection.Constants;
+using DataProtectionConstants = Microsoft.Azure.Web.DataProtection.Constants;
 
 namespace Microsoft.Azure.WebJobs.Script.WebHost
 {
@@ -323,7 +323,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
         {
             OperationResult result = OperationResult.NotFound;
 
-            secret = secret ?? GenerateIdentifiableSecret(FunctionKeySeed);
+            secret ??= GenerateIdentifiableSecret(FunctionKeySeed);
 
             await ModifyFunctionSecretsAsync(secretsType, keyScope, secrets =>
             {
@@ -513,11 +513,11 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
         {
             return new HostSecrets
             {
-                MasterKey = GenerateKey(ScriptConstants.DefaultMasterKeyName, MasterKeySeed),
+                MasterKey = GenerateMasterKey(),
                 FunctionKeys = new List<Key>
-            {
-                GenerateKey(ScriptConstants.DefaultFunctionKeyName, FunctionKeySeed)
-            },
+                {
+                    GenerateFunctionKey(),
+                },
                 SystemKeys = new List<Key>()
             };
         }
@@ -539,7 +539,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
             {
                 Keys = new List<Key>
                 {
-                    GenerateKey(ScriptConstants.DefaultFunctionKeyName, FunctionKeySeed)
+                    GenerateFunctionKey()
                 }
             };
         }
@@ -614,7 +614,21 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
             };
         }
 
-        private Key GenerateKey(string name, ulong seed)
+        private Key GenerateMasterKey()
+        {
+            string secret = GenerateIdentifiableSecret(MasterKeySeed);
+
+            return CreateKey(ScriptConstants.DefaultMasterKeyName, secret);
+        }
+
+        private Key GenerateFunctionKey()
+        {
+            string secret = GenerateIdentifiableSecret(FunctionKeySeed);
+
+            return CreateKey(ScriptConstants.DefaultFunctionKeyName, secret);
+        }
+
+        private Key CreateKey(string name, ulong seed)
         {
             string secret = GenerateIdentifiableSecret(seed);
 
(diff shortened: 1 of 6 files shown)


src/WebJobs.Script.WebHost/WebJobs.Script.WebHost.csproj
```diff
@@ -70,6 +70,7 @@
     <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.32" />
     <PackageReference Include="Microsoft.Azure.WebJobs.Host.Storage" Version="5.0.0-beta.1" />
     <PackageReference Include="Microsoft.Azure.WebSites.DataProtection" Version="2.1.91-alpha" />
+    <PackageReference Include="Microsoft.Security.Utilities" Version="1.1.0" />
```
michaelcfanning: >Utilities [](http://example.com/codeflow?start=50&length=9)

This package contains key generation logic (and tests) that have been integrated in several other Azure resource providers. Note there is a pending change in this package possibly relevant to your RP (to produce URL friendly base64-encoded keys) that we could consider to update this PR.

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
```diff
@@ -36,6 +38,24 @@ public class SecretManager : IDisposable, ISecretManager
         private string _repositoryClassName;
         private DateTime _lastCacheResetTime;
 
+        internal const string AzureFunctionsSignature = "AzFu";
+
+        // Checksum seed for master keys. The string transformation here results in a
+        // a seed that is versioned in the least significant bits (as a ulong):
+        // BitConverter.ToUInt64(Encoding.ASCII.GetBytes("Master00").Reverse().ToArray());
+        // If there is a need to version, use "Master01" as the next string literal.
+        internal const ulong MasterKeySeed = 0x4d61737465723030;
```
michaelcfanning, line 46: >MasterKeySeed [](http://example.com/codeflow?start=29&length=13)

An important question for the team is how to 'bucket' keys by generating each set with a unique seed. We'd typically do this in situations where response/remediation may take a different path. In general, a good strategy is to group all user-facing (customer handled) secrets with one seed, and two organize other generated keys (which are intended to be internal-facing only) with another.

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
```diff
@@ -254,7 +274,7 @@ public async Task<KeyOperationResult> SetMasterKeyAsync(string value = null)
                 if (value == null)
                 {
                     // Generate a new secret (clear)
-                    masterKey = GenerateSecret();
+                    masterKey = GenerateIdentifiableSecret(MasterKeySeed);
```
michaelcfanning, line 267: >GenerateIdentifiableSecret [](http://example.com/codeflow?start=32&length=26)

Assuming we retain a distinct seed for master keys, we should consider a distinct helper GenerateMasterKey or similar, which would ensure that the master key seed is always utilized. I did not make this change, though, wanting to get feedback from you on whether the grouping I've proposed here, master, system, and function keys, makes sense.

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
```diff
@@ -608,17 +628,14 @@ private Key CreateKey(string name, string secret)
...
 
-        internal static string GenerateSecret()
+        internal static string GenerateIdentifiableSecret(ulong seed)
         {
-            using (var rng = RandomNumberGenerator.Create())
-            {
-                byte[] data = new byte[40];
-                rng.GetBytes(data);
-                string secret = Convert.ToBase64String(data);
+            string secret = IdentifiableSecrets.GenerateIdentifiableKey(seed, 40, AzureFunctionsSignature);
```
michaelcfanning: >GenerateIdentifiableKey [](http://example.com/codeflow?start=48&length=23)

This is the essential change. This helper will produce a base64-encoded key, seven bytes of which are dedicated to a fixed signature (when encoded) and a checksum. The remaining entropy of the key is sufficient to meet security standards. Alternately, we could increase the size of the generated key in order to preserve full 320 bits of entropy (as exists today).

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
```diff
@@ -608,17 +628,14 @@ private Key CreateKey(string name, string secret)
...
 
-                // Replace pluses as they are problematic as URL values
-                return secret.Replace('+', 'a');
-            }
+            secret = secret.Replace('+', '-');
```
michaelcfanning: >secret [](http://example.com/codeflow?start=12&length=6)

Previously, this code eliminated all '+' chars from the generated key, I believe because in at least one code path, for extensions, a key was returned as part of a URL. Generally, the '/' character is also replaced or eliminated if URL-friendly base64-encoded values are sought (and the '=' sign padding is also replaced.

What I've done here is simply to replace '+' with '-' and '/' with '_', a common convention for maintaining a base64-encoding with a complete character set.

This change, however, will change the alphabet of your keys moving forward. It will break any consumers who may have depended on the previous behavior to use functions like Convert.FromBase64String to convert a secret to a byte array.

The Azure Identity package provides an alternate which could now be used, the Base64UrlEncoder class. It would be important for this change to determine whether there are any code locations which s/be updated to use this API. Alternately, those code locations could simply find/replace back to the 'special' character set before calling Convert.

test/WebJobs.Script.Tests/Security/SecretManagerTests.cs
```diff
@@ -54,6 +55,88 @@ public SecretManagerTests(ITestOutputHelper outputHelper)
             _startupContextProvider = new StartupContextProvider(_testEnvironment, loggerFactory.CreateLogger<StartupContextProvider>());
         }
 
+        [Fact]
+        public void SecretManager_GenerateIdentifiableSecret_HonorsSeed()
```
michaelcfanning, line 61: >SecretManager_GenerateIdentifiableSecret_HonorsSeed [](http://example.com/codeflow?start=20&length=51)

I've been conservative about updating tests until we could have a preliminary conversation on the key grouping (i.e., the # of seeds used to bucket allocations). Once that's decided, several of these tests could be updated to exercise the key generation logic and to verify that populated values have the expected signature/checksum.

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
mathewc: In our case, all of these keys are user-facing - the keys just have different authorization levels. So the severity of an issue would vary based on key type. However, in all cases I think the response/remediation would be the same. So I'm not sure if we need the three buckets. @brettsam / @fabiocav what do you think?

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
mathewc: As discussed, our keys need to continue to be URL friendly. They are passed as query params, but not in path segments which is why '/' isn't an issue. Also, we don't have any customers decoding these back to byte array - that's not a concern.

src/WebJobs.Script.WebHost/WebJobs.Script.WebHost.csproj
mathewc: What nuget feed is this on? Looks like on nuget.org the name is just reserved https://www.nuget.org/packages/Microsoft.Security.Utilities/

test/WebJobs.Script.Tests/Microsoft.Azure.WebJobs.Script.WebHost.deps.json
```diff
@@ -6,7 +6,7 @@
   "compilationOptions": {},
   "targets": {
     ".NETCoreApp,Version=v6.0": {
-      "Microsoft.Azure.WebJobs.Script.WebHost/4.1.0": {
+      "Microsoft.Azure.WebJobs.Script.WebHost/4.1.1": {
```
mathewc: Not sure if all these deps file changes are needed - we'll need to carefully review this.

michaelcfanning: Yes, this jumped out to me as worrisome as well. Ideally, this file should only inject references specific to the sole dependency I've introduced, Microsoft.Security.Utilities. However, a specific unit test failed on my machine making that focused change, and so I went ahead and baselined. 

Several other tests fail as well, so my best bet is probably to consult with any machine/testing setup guidance you have to get the broader functionals working. One test appears to indicate Java be installed. Others require a missing connection string.

I'll be giving this PR attention tomorrow (to fix the API factoring improvements we discussed offline) and I'll reset this deps.json file. If you have a suggestion on ensuring all tests pass, let me know, otherwise will go digging around for info in the repo.

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
michaelcfanning: Makes sense. Only the '+' character is a problem if tokens are only passed as query params. The trailing '=' padding chars shouldn't be a concern (as far as parameter parsing) if you're using the '&' as a separator.

Still, I think converting to a full 64-character (URL friendly) alphabet and eliding the trailing equal signs characters is a good change, since you've confirmed there shouldn't be anyone decoding to byte arrays (using standard base64) today.

One minor question I have is whether the absence of trailing padding might result in some incremental change in the # of false positive regex matches (that will get eliminated by the checksum validation). I plan to test this on a fairly large data set (a body of public OSS code) this week, will follow up here on results.

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
michaelcfanning: The bucketing is most helpful for routing problems to the appropriate audience for response. If all keys are user-facing, you could definitely get by with one.

There is arguably a benefit in response, however, in having a more granular categorization, simple because more information in reporting maximizes actionability. i.e., if the distinction between master, system, and function keys in any ways maximizes motivation to fix, lowers costs to get oriented and respond, could be worth it. Certainly the scanning costs aren't high for bucketing.

My personal opinion is that it's a close call but I'd favor having more information/retaining the three buckets. But I wouldn't argue if you'd prefer to keep it simple.

src/WebJobs.Script.WebHost/WebJobs.Script.WebHost.csproj
michaelcfanning: Can you see this version? https://www.nuget.org/packages/Microsoft.Security.Utilities/1.2.0. This is the one I plan to use.

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
michaelcfanning: @mathewc , removed all deps.json changes (resulted from me baselining a debug build, I believe). will work on getting all functional tests running today.

src/WebJobs.Script.WebHost/WebJobs.Script.WebHost.csproj
michaelcfanning: We just published v1.3 of this package, in order to provide a helper that generates a URL safe token without eliding the equal sign padding. We completed some offline analysis that suggests this will improve the accuracy of the preliminary regex match.

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
```diff
@@ -608,17 +642,10 @@ private Key CreateKey(string name, string secret)
...
-            {
-                byte[] data = new byte[40];
-                rng.GetBytes(data);
-                string secret = Convert.ToBase64String(data);
-
-                // Replace pluses as they are problematic as URL values
-                return secret.Replace('+', 'a');
-            }
+            string secret = IdentifiableSecrets.GenerateBase64Key(seed, 40, AzureFunctionsSignature, encodeForUrl: true);
+            return secret.EndsWith("==") ? secret : secret + "==";
```
mathewc: Why the suffix logic "=="?

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
```diff
@@ -16,7 +16,9 @@
 using Microsoft.Azure.WebJobs.Script.Diagnostics;
 using Microsoft.Azure.WebJobs.Script.WebHost.Properties;
 using Microsoft.Extensions.Logging;
-using DataProtectionCostants = Microsoft.Azure.Web.DataProtection.Constants;
+using Microsoft.Security.Utilities;
+
+using DataProtectionConstants = Microsoft.Azure.Web.DataProtection.Constants;
```
mathewc, line 21: thanks for fixing the spelling :)

test/WebJobs.Script.Tests/DefaultScriptWebHookProviderTests.cs
```diff
@@ -18,38 +19,68 @@ namespace Microsoft.Azure.WebJobs.Script.Tests
...
+                    "testextension_extension",
+                    It.IsAny<string>(),
+                    HostKeyScopes.SystemKeys,
+                    ScriptSecretsType.Host))
+                        .Callback<string, string, string, ScriptSecretsType>((key, secret, scope, type) => secretValue = secret)
+                        .Returns(() => Task.FromResult(new KeyOperationResult(secretValue, OperationResult.Created)));
+
+            // When an extension has no existing secret, one should be generated using
+            // the Azure Functions system key seed and standard fixed signature.
+            hostSecrets.SystemKeys = new Dictionary<string, string>();
```
mathewc: What does this line exist for? Doesn't look like these key values are accessed anywhere below?

test/WebJobs.Script.Tests/Security/SecretManagerTests.cs
```diff
@@ -1105,7 +1188,7 @@ private Mock<IKeyValueConverterFactory> GetConverterFactoryMock(bool simulateWri
             return mockValueConverterFactory;
         }
 
-        private SecretManager CreateSecretManager(string secretsPath, ILogger logger = null, IMetricsLogger metricsLogger = null, IKeyValueConverterFactory keyConverterFactory = null, bool createHostSecretsIfMissing = false, bool simulateWriteConversion = true, bool setStaleValue = true)
+        internal SecretManager CreateSecretManager(string secretsPath, ILogger logger = null, IMetricsLogger metricsLogger = null, IKeyValueConverterFactory keyConverterFactory = null, bool createHostSecretsIfMissing = false, bool simulateWriteConversion = true, bool setStaleValue = true)
```
mathewc: Is this change necessary? Didn't see it being accessed from other classes?

test/WebJobs.Script.Tests/Security/SecretManagerTests.cs
```diff
@@ -54,6 +55,88 @@ public SecretManagerTests(ITestOutputHelper outputHelper)
...
+                // Because we don't initialize any disk context, we can configure
+                // the secret manager to allocate fresh keys on start-up.
+                using (var secretManager = CreateSecretManager(directory.Path, createHostSecretsIfMissing: true))
+                {
+                    var hostSecrets = await secretManager.GetHostSecretsAsync();
+
+                    // Here and elsewhere we elide the '!' character which is prepended to every
+                    // generated secret value, as a clue that actual encryption is simulated in
+                    // the test case.
+                    ValidateSecret(hostSecrets.MasterKey.Substring(1), SecretManager.MasterKeySeed);
```
mathewc, line 84: I don't understand this - who adds the "!"?

test/WebJobs.Script.Tests/Security/SecretManagerTests.cs
```diff
@@ -54,6 +55,88 @@ public SecretManagerTests(ITestOutputHelper outputHelper)
...
+                    }
+                }
+            }
+        }
+
+        private static void ValidateSecret(string secret, ulong seed)
+        {
+            // All Azure function keys are 40 bytes in length, 56 chars
+            // when base64-encoded. A URL friendly encoding elides the
+            // '==' padding, however.
```
mathewc: I don't understand this comment - SecretManager.GenerateIdentifiableSecret is appending == when not present.

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
mathewc: I'm fine with retaining the buckets, thanks.

mathewc, line 46: How about for these three seeds we add SecretManager unit tests that include the code you give and serve as the location where we can keep these comments and sample code. The tests run and are debuggable, modifiable for the future if we need to version. We could have just a simple comment referencing the tests for details?

test/WebJobs.Script.Tests/Security/SecretManagerTests.cs
mathewc: Agree that it would be good to update a few of the existing tests that add/update keys by adding a verification step at the end ensuring the right seed was used, the secrets are identifiable, etc. I.e. something that would catch a mistake if someone used the wrong seed value.

michaelcfanning: Yes, makes sense. I will first merge this branch from main and then improve some tests. Are we all agreed that we're keeping the 3-seed breakdown, then? Master, system, and function?

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
michaelcfanning: Right, so we ran some tests and the absence of this trailing padding results in a significant number of matches against your core pattern (at GitHub public OSS scale). Adding the trailing padding will go a long way to reduce false positives for this preliminary match. We took a change to Security.Utilities to bring this configurability (padding or no padding) for URL friendly tokens into the API. I will update this PR package version and consume that new API.

test/WebJobs.Script.Tests/Security/SecretManagerTests.cs
mathewc: Yes, I'm fine with that. If @fabiocav / @brettsam don't have any opinions on this, I say go with what you have.

michaelcfanning: ACK on this and your other notes, which I'm cleaning acting on today. Thanks for the advice.

test/WebJobs.Script.Tests/Microsoft.Azure.WebJobs.Script.WebHost.deps.json
michaelcfanning: I've reverted everything not relevant to MS.Security.Utilities.

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
michaelcfanning: Great suggestion, done.

test/WebJobs.Script.Tests/Security/SecretManagerTests.cs
michaelcfanning: Whoops, neglected to reset after deciding to add tests in-place.

test/WebJobs.Script.Tests/Security/SecretManagerTests.cs
michaelcfanning: OK, so I have acted on your note, to update some tests to provide the validation you describe. I also made an effort to provide helpers that route people to the correct seeds.

mathewc (dismissed):

test/WebJobs.Script.Tests/Microsoft.Azure.WebJobs.Script.WebHost.deps.json
```diff
@@ -1322,6 +1323,14 @@
           }
         }
       },
+      "Microsoft.Security.Utilities/1.1.0": {
```
mathewc: @fabiocav / @brettsam to help review these deps changes

michaelcfanning: I just updated this deps file to properly reflect the current Microsoft.Security.Utilities dependency and the corresponding test passes. So just need to validate with everyone we are comfortable with picking up this Microsoft NuGet dependency.

test/WebJobs.Script.Tests/DefaultScriptWebHookProviderTests.cs
michaelcfanning: This code was added to account for the fact that the working system requires a non-null dictionary for this property. I went ahead and just baked this into the default initialization of the mock object (in the ctor).

test/WebJobs.Script.Tests/Microsoft.Azure.WebJobs.Script.WebHost.deps.json
michaelcfanning: @mathewc, FYI, I advertently cleared your previous approval, oops.

mathewc (dismissed):

fabiocav (dismissed):
Thank you for the changes and the patience, @michaelcfanning !

The changes look good. The only thing we need to do is make sure we're updating the runtime [assemblies manifest](https://github.com/Azure/azure-functions-host/blob/9cf4d03e54bcc98cc5b5e165467b56d04d0b6f99/src/WebJobs.Script/runtimeassemblies.json) to flag the new dependency ('Microsoft.Security.Utilities') as private, but I'm not going to block on this as I'd be happy to make that change if you'd like.

fabiocav (approved):

mathewc (approved):

michaelcfanning:
Thanks, @fabiocav. I would, in fact, appreciate your help marking the new dependency as private. :) Thanks for that.

mathewc:
Merged in mirror PR https://github.com/Azure/azure-functions-host/pull/8171

PR #8075, opened 2022-01-19
adding logs to shutdown path

brettsam:
While investigating https://github.com/Azure/azure-webjobs-sdk/issues/2823, it became clear that we have some blind spots during shutdown. For example, some scenarios seem to completely skip the call to `StopAsync()` and immediately dispose of the listeners. I'm adding some additional logging to see if we can catch this and possibly address it.

### DIFF
--- a/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
+++ b/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
@@ -694,15 +694,23 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
         /// <param name="instance">The <see cref="IHost"/> instance to remove</param>
         private async Task Orphan(IHost instance, CancellationToken cancellationToken = default)
         {
+            if (instance == null)
+            {
+                _logger.LogDebug("Cannot stop or dispose a null host instance.");
+                return;
+            }
+
             var isStandbyHost = false;
+            string hostInstanceId = "unknown";
             try
             {
-                var scriptHost = (ScriptHost)instance?.Services.GetService<ScriptHost>();
+                var scriptHost = instance.Services.GetService<ScriptHost>();
                 if (scriptHost != null)
                 {
                     scriptHost.HostInitializing -= OnHostInitializing;
                     scriptHost.HostInitialized -= OnHostInitialized;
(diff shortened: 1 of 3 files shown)


liliankasem (dismissed):

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -712,45 +720,43 @@ private async Task Orphan(IHost instance, CancellationToken cancellationToken =
...
+                        }
+                    }, cancellationToken);
+
+                    _logger.LogDebug("Standby ScriptHost marked for disposal");
+                }
+                else
+                {
+                    DisposeDependencyTrackingModule(instance);
+                    instance.Dispose();
+                    _logger.LogDebug("ScriptHost disposed: '{hostInstanceId}'", hostInstanceId);
```
fabiocav: This is one of those cases where I do feel we have a number of queries and potentially detectors relying on this log message, and this change could cause them to break. Thoughts?

brettsam: I was trying to keep it so that those could run with "startswith" -- but I do agree. Let me see if there's another way I can add this information.

NickCraver:
Just to add an avenue here - in hunting some of these things it's akin to connection leaks, one of the simpler but more successful things we've found in hunting these down is adding a call in the deconstructor. If we called Dispose() manually: great! If it was the finalizer cleaning us up via GC: uh oh, that's an unintentional leak. Luckily I have an OSS version of this handy if it's of use: https://github.com/StackExchange/StackExchange.Redis/blob/main/src/StackExchange.Redis/ConnectionMultiplexer.cs#L75-L85 - if that's an appealing path happy to chat or explain more, just noticed this and thought the idea may be useful :)

In other places where you don't know where it's created or disposed (where is the leak entry point?), using caller member attributes in the constructors, and having that string when the finalizer hits so record where your original source was is also handy. I don't have an example at the moment to point to, but if of interest happy to cook something up!

fabiocav (dismissed):

liliankasem (dismissed):

liliankasem (approved):

fabiocav (approved):

Issue #5922, opened 2020-04-28
Add support for AzureWebJobsSecretStorageType=none


balag0:
For customers not using AzureWebjobsStorage / AzureWebJobsSecretStorageSas there is no where to store function/host keys. This causes failures anytime an attempt is made to access secrets, e.g., during synctriggers, secrets management admin APIs, etc.

Proposal is to support AzureWebJobsSecretStorageType=none, For apps using this configuration we will not generate or persist any keys.

Instead all authentication will be based on jwt tokens based on AzureWebEncryptionKey.

@mathewc / @balag0 for additional details.

fabiocav:
Work currently in progress. Moving to sprint 96

mathewc:
@balag0  - can you remind me what the source of this work item was? Are customers running into this? Definitely if we have codepaths that are accessing secrets (e.g. admin keys APIs, synctriggers) and not failing gracefully in absence of a secrets repository we should address that.

fabiocav:
@mathewc I believe this is for static web apps.

balag0:
Correct. It is for SWA. + @anthonychu as well who is tracking other related requirements around this scenario.

anthonychu:
Yes this is the issue: https://github.com/Azure/static-web-apps/issues/165

It’s because BlueRidge apps don’t have storage accounts. @annaji-msft can provide more details.

This is quite common because folks are using a function as the callback for external auth providers and they pass a token in the query string with the name `code`.

anthonychu:
The desired outcome is if no AzureWebJobsStorage and the http function has authLevel anonymous, requests with `code=` in the query string should work, and the parameter value is accessible in the function.

fabiocav:
Moving to next sprint as this work has not started yet.

FYI @anthonychu

fabiocav:
@anthonychu an update on this issue; @kashimiz has the initial set of changes and is iterating with additional people prior to opening a PR. We expect this to be done at the beginning of 104.

fabiocav:
@kashimiz should this be moved to 106?

kashimiz:
@fabiocav Yes, we should move this to 106.

fabiocav:
@kashimiz leaving this in 106 pending an update. Let me know if this needs to move.

fabiocav:
Moving this to sprint 116 as we still have open discussions on this.

/cc @mathewc @kashimiz

fabiocav:
[...]

PR #8083, opened 2022-01-20
Auth handler gated on SecretsEnabled check.

kashimiz:

Addresses #5922.

Ported forward to #8239.

### Pull request checklist

### DIFF
--- a/test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SWAEndToEndTests.cs
+++ b/test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SWAEndToEndTests.cs
@@ -1,0 +1,157 @@
+﻿using Microsoft.Azure.WebJobs.Script.WebHost;
+using Microsoft.Azure.WebJobs.Script.WebHost.Management;
+using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+using Newtonsoft.Json;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Microsoft.Azure.WebJobs.Script.Tests.Integration.WebHostEndToEnd
+{
+    /// <summary>
+    /// Tests verifying the Static Web Apps deployment configuration.
+    /// </summary>
+    public class SWAEndToEndTests : IClassFixture<SWAEndToEndTests.TestFixture>
+    {
+        private TestFixture _fixture;
+
+        public SWAEndToEndTests(TestFixture fixture)
+        {
+            _fixture = fixture;
+        }
+
+        [Fact]
+        public void NoStorageConfigured_SecretsDisabled()
+        {
+            Assert.Null(Environment.GetEnvironmentVariable("AzureWebJobsStorage"));
+            Assert.False(_fixture.Host.SecretManagerProvider.SecretsEnabled);
+        }
+
+        [Fact]
+        public async Task InvokeFunction_AnonymousLevel_Succeeds()
+        {
+            var code = "test";
+
+            // send along a code query param, and ensure that the invocation succeeds
+            // the token isn't interpreted as a function key in this case since keys are disabled
+            var content = new StringContent(JsonConvert.SerializeObject(new { scenario = "swa" }));
+            var response = await _fixture.Host.HttpClient.PostAsync($"api/HttpTrigger-Scenarios?code={code}", content);
+            response.EnsureSuccessStatusCode();
+
+            var responseBody = await response.Content.ReadAsStringAsync();
+            Assert.Equal(code, responseBody);
+        }
+
+        [Fact]
+        public async Task InvokeFunction_FunctionLevel_NoAuthToken_Fails()
+        {
+            // the function declares an auth level of Function, so invocations should fail
+            var response = await _fixture.Host.HttpClient.GetAsync("api/HttpTrigger-FunctionAuth?code=test");
+            Assert.Equal(HttpStatusCode.Unauthorized, response.StatusCode);
+
+            response = await _fixture.Host.HttpClient.GetAsync("api/HttpTrigger-FunctionAuth");
+            Assert.Equal(HttpStatusCode.Unauthorized, response.StatusCode);
+        }
+
+        [Fact]
+        public async Task InvokeFunction_FunctionLevel_ValidToken_Succeeds()
+        {
+            // if an admin token is passed, the function invocation succeeds
+            HttpRequestMessage request = new HttpRequestMessage(HttpMethod.Get, "api/HttpTrigger-FunctionAuth?code=test");
+            string token = _fixture.Host.GenerateAdminJwtToken();
+            request.Headers.Authorization = new AuthenticationHeaderValue("Bearer", token);
+
+            var response = await _fixture.Host.HttpClient.SendAsync(request);
+            response.EnsureSuccessStatusCode();
+        }
+
+        [Fact]
+        public async Task SyncTriggers_Succeeds()
+        {
+            _fixture.Host.ClearLogMessages();
+
+            HttpRequestMessage request = new HttpRequestMessage(HttpMethod.Post, "admin/host/synctriggers");
+            string token = _fixture.Host.GenerateAdminJwtToken();
+            request.Headers.Authorization = new System.Net.Http.Headers.AuthenticationHeaderValue("Bearer", token);
+
+            var response = await _fixture.Host.HttpClient.SendAsync(request);
+
+            // the sync request will fail because we're not running in Antares and can't
+            // communicate with the FE, but it suffices to verify that we attempted to send the request
+            var logs = _fixture.Host.GetScriptHostLogMessages().Where(p => p.Category == typeof(FunctionsSyncManager).FullName).ToArray();
+            var log = logs[0].FormattedMessage;
+            Assert.True(log.StartsWith("Making SyncTriggers request"));
+            Assert.True(log.Contains("HttpTrigger-FunctionAuth"));
+            Assert.True(log.Contains("HttpTrigger-Scenarios"));
+        }
+
+        public class TestFixture : EndToEndTestFixture
+        {
+            private TestScopedEnvironmentVariable _scopedEnvironment;
+
+            public TestFixture() : base(@"TestScripts\CSharp", "csharp", RpcWorkerConstants.DotNetLanguageWorkerName, addTestSettings: false)
+            {
+                var testKeyBytes = TestHelpers.GenerateKeyBytes();
+                var testKey = TestHelpers.GenerateKeyHexString(testKeyBytes);
+
+                var settings = new Dictionary<string, string>()
+                {
+                    { "AzureWebEncryptionKey", testKey },
+                    { EnvironmentSettingNames.WebSiteAuthEncryptionKey, testKey },
+                    { "AzureWebJobsStorage", null },
+                    { EnvironmentSettingNames.AzureWebsiteName, "testsite" }
+                };
+                _scopedEnvironment = new TestScopedEnvironmentVariable(settings);
+            }
+
+            public override void ConfigureScriptHost(IWebJobsBuilder webJobsBuilder)
+            {
+                base.ConfigureScriptHost(webJobsBuilder);
+
+                webJobsBuilder.Services.Configure<ScriptJobHostOptions>(o =>
+                {
+                    o.Functions = new[]
+                    {
+                        "HttpTrigger-Scenarios",
+                        "HttpTrigger-FunctionAuth"
+                    };
+                });
+            }
+
+            public override void ConfigureScriptHost(IServiceCollection services)
+            {
+                base.ConfigureScriptHost(services);
+
+                // replace the base mock FunctionsSyncManager
+                var service = services.FirstOrDefault(d => d.ServiceType == typeof(IFunctionsSyncManager));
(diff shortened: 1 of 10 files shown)


src/WebJobs.Script.WebHost/Security/KeyManagement/DefaultSecretManagerProvider.cs
```diff
@@ -74,64 +74,56 @@ private void ResetSecretManager()
             Interlocked.Exchange(ref _secretManagerLazy, new Lazy<ISecretManager>(Create));
         }
 
-        private ISecretManager Create() => new SecretManager(CreateSecretsRepository(), _loggerFactory.CreateLogger<SecretManager>(), _metricsLogger, _hostNameProvider, _startupContextProvider);
+        private ISecretManager Create() => SecretsEnabled ? new SecretManager(CreateSecretsRepository(), _loggerFactory.CreateLogger<SecretManager>(), _metricsLogger, _hostNameProvider, _startupContextProvider) : null;
```
mathewc: Changes to this file shouldn't be needed

test/WebJobs.Script.Tests.Integration/Security/AuthenticationLevelHandlerTests.cs
```diff
@@ -0,0 +1,114 @@
...
+            }
+        }
+    }
+
+    public class AuthenticationLevelHandlerTestFixture : IDisposable
+    {
+        public TestFunctionHost Host;
+        private bool _disposed;
+
+        public AuthenticationLevelHandlerTestFixture(string scriptRoot, string testRuntime, IDictionary<string, string> webHostConfig)
```
mathewc: the webHostConfig you're passing in doesn't appear to be used? So how are you configuring the test host with SWA config?

test/WebJobs.Script.Tests.Integration/Security/AuthenticationLevelHandlerTests.cs
```diff
@@ -0,0 +1,114 @@
...
+using Moq;
+using Newtonsoft.Json;
+using Xunit;
+
+namespace Microsoft.Azure.WebJobs.Script.Tests.Integration.Security
+{
+    [Trait(TestTraits.Category, TestTraits.EndToEnd)]
+    [Trait(TestTraits.Group, nameof(AuthenticationLevelHandlerTests))]
+    public class AuthenticationLevelHandlerTests
+    // TODO: This should be a more general E2E test suite - maybe for SWA? Or combine with storage/secret E3E tests?
```
mathewc: Agree that it would be good to have a dedicated test suite for the SWA scenario. I was experimenting myself, and sketched one out in branch https://github.com/Azure/azure-functions-host/tree/swa-fix. That might be a good starting point - it also addresses some of the issues I point out below.

test/WebJobs.Script.Tests.Integration/Security/AuthenticationLevelHandlerTests.cs
```diff
@@ -0,0 +1,114 @@
...
+            // test when AzureWebJobsSecretStorageType=None
+            // test when AzureWebJobsSecretStorageType= something else
+            // test when AzureWebJobsSecretStorageType is not set
+
+            // test for Node scenarios too
+        }
+
+        [Theory]
+        [MemberData(nameof(GetData))]
+        public async Task NoStorage_ReturnsExpected(string scriptRoot, string runtime, IDictionary<string, string> config)
```
mathewc: We need tests for other SWA scenarios. I sketched some out in the following suite, and also provided solution to the TestFunctionHost issues you pointed out https://github.com/Azure/azure-functions-host/blob/eb5e2e2f003634d42adf01ade43a1c5e7477850d/test/WebJobs.Script.Tests.Integration/ScriptHostEndToEnd/SWAEndToEndTests.cs#L24

test/WebJobs.Script.Tests.Integration/Security/AuthenticationLevelHandlerTests.cs
```diff
@@ -0,0 +1,114 @@
...
+        public static IEnumerable<object[]> GetData()
+        {
+            yield return new object[] { @"TestScripts\CSharp", RpcWorkerConstants.DotNetLanguageWorkerName,
+                new Dictionary<string, string>() { ["AzureWebJobsStorage"] = string.Empty }};
+
+            // test when AzureWebJobsSecretStorageType=None
+            // test when AzureWebJobsSecretStorageType= something else
+            // test when AzureWebJobsSecretStorageType is not set
+
+            // test for Node scenarios too
```
mathewc: These changes are language agnostic, so we don't need per language tests. Using some .NET functions will suffice

kashimiz: Great. Removing this todo.

test/WebJobs.Script.Tests.Integration/Security/AuthenticationLevelHandlerTests.cs
kashimiz: Thanks for the catch; I ended up not needing it. My understanding based on discussions with Fabio and Anthony is that we're this behavior change is meant to affect apps that meet the following two conditions, of which SWA apps are a subset:

1. No `AzureWebJobsStorage` setting / `AzureWebJobsStorage` is null/empty
2. `AzureWebEncryptionKey` exists (for JWT auth path)

Initially I set up webHostConfig to overwrite the test `AzureWebJobsStorage` setting with an empty value. Turns out if you do that, `IConfigurationBuilder` will go back through its sources for the last non-null/empty value for a setting, which is why `TestConfigurationBuilderExtensions.RemoveTestSettings()` now exists.

And 2 is set as an environment variable, so `webHostConfig` became an artifact. I've removed it.

src/WebJobs.Script.WebHost/Security/KeyManagement/DefaultSecretManagerProvider.cs
kashimiz: Reverted.

test/WebJobs.Script.Tests.Integration/Security/AuthenticationLevelHandlerTests.cs
kashimiz: Transferred your test suite over. Did add the step of wiping existing test settings; the build machine has an `appsettings.test.json` that provides an `AzureWebJobsStorage` value otherwise. (See note in TestConfigurationBuilderExtensions.)

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SWAEndToEndTests.cs
```diff
@@ -0,0 +1,151 @@
...
+namespace Microsoft.Azure.WebJobs.Script.Tests.Integration.WebHostEndToEnd
+{
+    /// <summary>
+    /// Tests verifying the Static Web Apps deployment configuration.
+    /// </summary>
+    public class SWAEndToEndTests : IClassFixture<SWAEndToEndTests.TestFixture>
+    {
+        private TestFixture _fixture;
+
+        public SWAEndToEndTests(TestFixture fixture)
```
mathewc, line 25: I had a NoStorageConfigured_SecretsDisabled test that in my branch [here](https://github.com/Azure/azure-functions-host/commit/1c35d0bda882324df17e2d16e3d7c348b7d54957) that you should retain - it ensures we're running the SWA scenario and secrets aren't enabled.

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SWAEndToEndTests.cs
```diff
@@ -0,0 +1,151 @@
...
+            // the token isn't interpreted as a function key in this case since keys are disabled
+            var content = new StringContent(JsonConvert.SerializeObject(new { scenario = "swa" }));
+            var response = await _fixture.Host.HttpClient.PostAsync("api/HttpTrigger-Scenarios?code=test", content);
+            response.EnsureSuccessStatusCode();
+
+            // TODO: should this verify that the code was interpreted as a parameter?
+        }
+
+        [Fact]
+        public async Task InvokeFunction_FunctionLevel_NoAuthToken_Fails()
```
mathewc, line 53: Also need the InvokeFunction_FunctionLevel_ValidToken_Succeeds test I had - why didn't you retain that?

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SWAEndToEndTests.cs
```diff
@@ -0,0 +1,151 @@
...
+        [Fact]
+        public async Task InvokeFunction_AnonymousLevel_Succeeds()
+        {
+            // send along a code query param, and ensure that the invocation succeeds
+            // the token isn't interpreted as a function key in this case since keys are disabled
+            var content = new StringContent(JsonConvert.SerializeObject(new { scenario = "swa" }));
+            var response = await _fixture.Host.HttpClient.PostAsync("api/HttpTrigger-Scenarios?code=test", content);
+            response.EnsureSuccessStatusCode();
+
+            // TODO: should this verify that the code was interpreted as a parameter?
```
mathewc: I don't think so - if it were interpreted as an auth code, we'd attempt to access secrets and we'd have a 500 right? You could verify that in the debugger by forcing the auth handler down that path, but that's what I saw.

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SWAEndToEndTests.cs
```diff
@@ -0,0 +1,151 @@
...
+                services.Remove(service);
+                services.AddSingleton<IFunctionsSyncManager, FunctionsSyncManager>();
+            }
+
+            public override void ConfigureScriptHost(IConfigurationBuilder configBuilder)
+            {
+                base.ConfigureScriptHost(configBuilder);
+
+                // wipe existing test settings
+                configBuilder.RemoveTestSettings();
```
mathewc: I think an easier way for you to do this would be to just add a parameter to the EndToEndTestFixture ctor "addTestSettings" that defaults to true. You can override this in your suite ctor above. This value would be passed to the TestFunctionHost that the fixture creates, and the calls to AddTestSettings would be skipped when false. That way your two ConfigureScriptHost/ConfigureWebHost overrides aren't needed, nor is the new RemoveTestSettings method.

test/WebJobs.Script.Tests.Integration/WebJobs.Script.Tests.Integration.csproj
```diff
@@ -19,6 +19,12 @@
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <WarningsAsErrors />
   </PropertyGroup>
+  <ItemGroup>
```
mathewc: What is this change for?

test/WebJobs.Script.Tests.Shared/TestConfigurationBuilderExtensions.cs
```diff
@@ -3,16 +3,31 @@
...
+using System.Linq;
 using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Configuration.Json;
 
 namespace Microsoft.Azure.WebJobs.Script.Tests
 {
     public static class TestConfigurationBuilderExtensions
     {
-        public static IConfigurationBuilder AddTestSettings(this IConfigurationBuilder builder)
+        private const string ConfigFile = "appsettings.tests.json";
```
mathewc, line 13: With my suggestion above, changes to this file are likely not needed

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SWAEndToEndTests.cs
kashimiz: Right, this is why 10 AM brain should check what 1 AM brain did. That was an oversight; I'll add those back in.

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SWAEndToEndTests.cs
kashimiz: That is what happens in that scenario, yes, and it can be assumed that if not interpreted as an auth code, it'll be read as a parameter. I looked back at Anthony's requirements in https://github.com/Azure/azure-functions-host/issues/5922#issuecomment-781760639 and 'parameter value accessible in function' was explicitly called out, so I thought to verify.

test/WebJobs.Script.Tests.Integration/WebJobs.Script.Tests.Integration.csproj
kashimiz: A very good question. Removed.

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SWAEndToEndTests.cs
kashimiz: Oh dear. Selective blindness again. Added this one in as well.

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SWAEndToEndTests.cs
kashimiz: Agreed; that's a cleaner way of doing things.

test/WebJobs.Script.Tests.Shared/TestConfigurationBuilderExtensions.cs
kashimiz: Indeed they're not. Removed.

mathewc (dismissed):

mathewc:
Need to port to v4 as well.

mathewc (approved):

PR #8101, opened 2022-01-27
Use appropriate log level for worker console logs for v3

ejizba:
Currently, worker console logs are always categorized as "Debug" even though they can contain useful warnings or errors. For example, the Node.js worker displays warnings/errors when validating the Node.js version prior to initializing the rpc channel.

This may be a regression caused by https://github.com/Azure/azure-functions-host/pull/6576. That PR doesn't mention anything about changing log level, so I'm pretty sure it was unintentional, but cc @pragnagopa @AnatoliB to confirm

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs
+++ b/src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs
@@ -190,7 +190,7 @@ namespace Microsoft.Azure.WebJobs.Script.Workers
             };
             if (WorkerProcessUtilities.IsConsoleLog(msg))
             {
-                _workerProcessLogger?.LogDebug(WorkerProcessUtilities.RemoveLogPrefix(msg));
+                _workerProcessLogger?.Log(level, WorkerProcessUtilities.RemoveLogPrefix(msg));
             }
             else
             {
--- a/test/WebJobs.Script.Tests/Workers/WorkerConsoleLogServiceTests.cs
+++ b/test/WebJobs.Script.Tests/Workers/WorkerConsoleLogServiceTests.cs
@@ -17,7 +17,8 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.Workers
     {
         private IScriptEventManager _eventManager;
         private IProcessRegistry _processRegistry;
-        private TestLogger _testLogger = new TestLogger("test");
+        private TestLogger _testUserLogger = new TestLogger("Host.Function.Console");
+        private TestLogger _testSystemLogger = new TestLogger("Worker.rpcWorkerProcess");
         private WorkerConsoleLogService _workerConsoleLogService;
         private WorkerConsoleLogSource _workerConsoleLogSource;
         private Mock<IServiceProvider> _serviceProviderMock;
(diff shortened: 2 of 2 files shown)


src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs
```diff
@@ -190,7 +190,7 @@ internal void BuildAndLogConsoleLog(string msg, LogLevel level)
             };
             if (WorkerProcessUtilities.IsConsoleLog(msg))
             {
-                _workerProcessLogger?.LogDebug(WorkerProcessUtilities.RemoveLogPrefix(msg));
+                _workerProcessLogger?.Log(level, WorkerProcessUtilities.RemoveLogPrefix(msg));
```
pragnagopa, line 193: This was an intentional change - as we should consider all Worker logs as System Logs - system logs are noisy for Cx but Debug is the default logLevel for Kusto - product team has access to all debug level system logs.

pragnagopa (changes requested):
If you do need few system logs to be logged at Info level - we need to come up a distinction for System Log - Info level. I recommend leaving the default loglevel here to be Debug.

src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs
ejizba: Aren't those two separate things? I agree they are system logs, but normally I see us choose a level _and_ a category. The level can be "Error", "Warning", "Information", "Debug", etc. and the category can be either "System" or "User". In this case I thought `_workerProcessLogger` was used for system logs and `_consoleLogSource` (in the "else" case below) was used for user logs.

Without proper levels, it makes supportability more difficult for the workers because any warning/error the worker logs is ignored - users _can't_ see it and the support team _won't_ see it because they often filter by log level.

Finally - at least when I'm running locally the default log level for the "Worker" category is "warning", so it's not noisy at all in my experience.

fabiocav: Interesting. @pragnagopa it does look like the change has an impact on the logs surfaced to the end user, which is something we're trying to address. Is it possible the original change had a broader impact than anticipated and left us with a gap that makes troubleshooting more difficult for customers?

pragnagopa: Looking at the previous PR and the test added, the intention was to force language worker logs to be treated as  system logs similar to what we do in the host.  Also, condition here ` if (WorkerProcessUtilities.IsConsoleLog(msg))` - is specifically to ensure the log has known prefix.

If you would like to go ahead with this change - I recommend scoping the change to language workers that the team owns - as of now they are shipped and packaged with the host - to avoid logging info level logs from language workers not owned by the team.

liliankasem: That sounds good to me, I would like to go ahead with the change if possible - @ejizba is it possible to scope this as Pragna suggested?

AnatoliB:
cc @gzuber @RohitRanjanMS - FYI

src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs
pragnagopa: @ejizba - I do not think host has a list of language workers that are well known - open a separate for the recommendation above. Approving this PR!

pragnagopa (approved):
Please do wait for final approval from @fabiocav / @liliankasem

liliankasem (approved):

src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs
ejizba: Filed this issue to follow up: https://github.com/Azure/azure-functions-host/issues/8109

PR #8107, opened 2022-02-01
Adding ability to allow worker to send Function load responses in batch

### LINKED ISSUE #8082: Single load response - design discussion and implementation
With reference to the issue - https://github.com/Azure/azure-functions-host/issues/7880 mentioning that host should have the ability to send single FunctionLoadRequest when multiple functions are there in a Function App, we also need to discuss about having Single FunctionLoadResponse. So, when worker advertise through capabilities that it can accept an array of FunctionLoadRequest then it should also return the response in the form of array of FunctionLoadResponse.

As they are performance improvement tasks, we should ensure that we are not adding more delays in case some functions take more time to load (or due to any other scenarios). As per [this code](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs#:~:text=internal%20void-,LoadResponse,-(FunctionLoadResponse%20loadResponse)), it seems that as soon as Host gets LoadResponse, it sends its InvocationRequest. So, should we wait for getting all LoadResponses before sending any function's InvocationRequest?

Need to discuss, design and implement as needed.

surgupta-msft:
resolves #8082

Adding ability to allow worker to send Function load responses in batch. Worker can choose to send load responses one by one using _FunctionLoadResponse_ field or in batch using _FunctionLoadResponseCollection_ field.

### DIFF
--- a/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
@@ -295,7 +295,7 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
                         .Subscribe((msg) => LoadResponse(msg.Message.FunctionLoadResponse), HandleWorkerFunctionLoadError));
 
                     _functionLoadTimeout = functionTimeout.Value > _functionLoadTimeout ? functionTimeout.Value : _functionLoadTimeout;
-                    _eventSubscriptions.Add(_inboundWorkerEvents.Where(msg => msg.MessageType == MsgType.FunctionLoadResponse)
+                    _eventSubscriptions.Add(_inboundWorkerEvents.Where(msg => msg.MessageType == MsgType.FunctionLoadResponseCollection)
                         .Timeout(_functionLoadTimeout)
                         .Take(_functions.Count())
                         .Subscribe((msg) => LoadResponse(msg.Message.FunctionLoadResponseCollection), HandleWorkerFunctionLoadError));
@@ -305,7 +305,7 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
                     _eventSubscriptions.Add(_inboundWorkerEvents.Where(msg => msg.MessageType == MsgType.FunctionLoadResponse)
                         .Subscribe((msg) => LoadResponse(msg.Message.FunctionLoadResponse), HandleWorkerFunctionLoadError));
 
-                    _eventSubscriptions.Add(_inboundWorkerEvents.Where(msg => msg.MessageType == MsgType.FunctionLoadResponse)
+                    _eventSubscriptions.Add(_inboundWorkerEvents.Where(msg => msg.MessageType == MsgType.FunctionLoadResponseCollection)
                         .Subscribe((msg) => LoadResponse(msg.Message.FunctionLoadResponseCollection), HandleWorkerFunctionLoadError));
                 }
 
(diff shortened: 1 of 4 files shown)


satvu (dismissed):

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -293,11 +293,20 @@ public void SendFunctionLoadRequests(ManagedDependencyOptions managedDependencyO
                         .Timeout(_functionLoadTimeout)
                         .Take(_functions.Count())
                         .Subscribe((msg) => LoadResponse(msg.Message.FunctionLoadResponse), HandleWorkerFunctionLoadError));
+
+                    _functionLoadTimeout = functionTimeout.Value > _functionLoadTimeout ? functionTimeout.Value : _functionLoadTimeout;
```
liliankasem: I don't see a difference between this `_functionLoadTimeout` and the one defined on line 291, we can probably delete this line as `_functionLoadTimeout` is already defined. Or is there a reason we need to define it twice?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -461,6 +470,16 @@ internal void LoadResponse(FunctionLoadResponse loadResponse)
             _inputLinks.Add(disposableLink);
         }
 
+        internal void LoadResponse(FunctionLoadResponseCollection loadResponseCollection)
```
liliankasem, line 472: Just curious, how often does FunctionLoadResponse / FunctionLoadResponseCollection get called? Is it just on startup/restart? Or more frequently than that?

test/WebJobs.Script.Tests/Workers/Rpc/GrpcWorkerChannelTests.cs
```diff
@@ -487,6 +487,21 @@ public void ReceivesInboundEvent_FunctionLoadResponse()
             Assert.True(traces.Any(m => string.Equals(m.FormattedMessage, "Received FunctionLoadResponse for function: 'js1' with functionId: 'TestFunctionId1'.")));
         }
 
+        [Fact]
+        public void ReceivesInboundEvent_FunctionLoadResponses()
```
liliankasem, line 507: Can you think of any failure scenarios that we haven't tested at all? I'd like us to test more of what happens when things go wrong. Also is it possible to add an integration test for this? Maybe something under: test/WebJobs.Script.Tests.Integration/Rpc

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
surgupta-msft: Removed

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
surgupta-msft: It will be called frequently than that. Host sends request to worker to load functions and gets load responses. So, 1 call for each function in the app. They can reduce number of calls by sending/receiving collection instead of one by one.

soninaren (dismissed):
Looks good to me. Would be good to test / verify failure scenarios as Lilian mentioned.

liliankasem (approved):

soninaren (approved):

PR #8128, opened 2022-02-09
Removing binding direction from event name when emitting binding metric to logs.

kshyju:
Currently, for in-proc function apps, only trigger bindings were included in the function meta data. So, no other bindings (out/return) are being logged to the functions metrics logs. This is being fixed with this PR: [Include all bindings when generating function metadata](https://github.com/Azure/azure-functions-vs-build-sdk/pull/542)

As part of fixing this issue, we have decided to remove the binding direction value ("in" or "out") from the event name when emitting this to the functions metrics logs since we cannot reliably get this value.

### DIFF
--- a/src/WebJobs.Script.WebHost/Diagnostics/FunctionInstanceLogger.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/FunctionInstanceLogger.cs
@@ -83,9 +83,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
             {
                 string eventName = _bindingMetricEventNames.GetOrAdd(binding, static (b) =>
                 {
-                    return b.IsTrigger ?
-                        string.Format(MetricEventNames.FunctionBindingTypeFormat, b.Type) :
-                        string.Format(MetricEventNames.FunctionBindingTypeDirectionFormat, b.Type, b.Direction);
+                    return string.Format(MetricEventNames.FunctionBindingTypeFormat, b.Type);
                 });
                 _metrics.LogEvent(eventName, metadata.Name);
             }
--- a/test/WebJobs.Script.Tests/Diagnostics/FunctionInstanceLoggerTests.cs
+++ b/test/WebJobs.Script.Tests/Diagnostics/FunctionInstanceLoggerTests.cs
@@ -30,24 +30,20 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.Diagnostics
                 Name = "TestFunction"
             };
             metadata.Bindings.Add(new BindingMetadata { Type = "httpTrigger" });
-            metadata.Bindings.Add(new BindingMetadata { Type = "blob", Direction = BindingDirection.In });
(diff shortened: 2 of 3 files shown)


liliankasem (dismissed):

test/WebJobs.Script.Tests/Diagnostics/FunctionInstanceLoggerTests.cs
```diff
@@ -30,24 +30,20 @@ public void LogInvocationMetrics_EmitsExpectedEvents()
                 Name = "TestFunction"
             };
             metadata.Bindings.Add(new BindingMetadata { Type = "httpTrigger" });
-            metadata.Bindings.Add(new BindingMetadata { Type = "blob", Direction = BindingDirection.In });
```
mathewc: I think you should leave the existing test inputs as is, since this represents a valid user scenario (in and out bindings of the same type). Instead just update the log verification.

src/WebJobs.Script.WebHost/Diagnostics/FunctionInstanceLogger.cs
```diff
@@ -83,9 +83,7 @@ internal object LogInvocationMetrics(FunctionMetadata metadata)
             {
                 string eventName = _bindingMetricEventNames.GetOrAdd(binding, static (b) =>
                 {
-                    return b.IsTrigger ?
```
mathewc, line 86: I wonder if we should retain the information we have w.r.t triggers. So triggers would be logged with a metric name that indicates it's a trigger, but all other bindings are logged w/o a direction.

mathewc: Never mind, I realize now the type will indicate that it is a trigger, since the binding type will say "Trigger"

test/WebJobs.Script.Tests/Diagnostics/FunctionInstanceLoggerTests.cs
kshyju: Test updated in the latest iteration.

mathewc (dismissed):

src/WebJobs.Script.WebHost/Diagnostics/FunctionInstanceLogger.cs
```diff
@@ -83,9 +83,7 @@ internal object LogInvocationMetrics(FunctionMetadata metadata)
...
-                        string.Format(MetricEventNames.FunctionBindingTypeFormat, b.Type) :
-                        string.Format(MetricEventNames.FunctionBindingTypeDirectionFormat, b.Type, b.Direction);
```
mathewc, line 88: Let's remove the now unused format string

kshyju: Removed unused constant.

mathewc (approved):

liliankasem (approved):

PR #8152, opened 2022-02-17
Update Jit Trace Integration Test

VpOfEngineering:
Currently, the `JitPreparesTest` checks for failure percentages for the linux and windows JIT trace files. We are moving this test to instead check if the file exists and has not dramatically reduced the number of lines in it.

### Pull request checklist

### Additional information

Additional PR information


test/WebJobs.Script.Tests.Integration/PreJIT/JitPreparesTest.cs
```diff
@@ -16,22 +16,17 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.PreJIT
...
             Assert.True(file.Exists, $"Expected PGO file '{file.FullName}' does not exist. The file was either renamed or deleted.");
-
-            JitTraceRuntime.Prepare(file, out int successfulPrepares, out int failedPrepares);
-
-            var failurePercentage = (double)failedPrepares / successfulPrepares * 100;
-
-            // using 1% as approximate number of allowed failures before we need to regenrate a new PGO file.
-            Assert.True(failurePercentage < threshold, $"Number of failed PGOs are more than {threshold} percent! Current number of failures are {failedPrepares}. This will definitely impact cold start! Time to regenrate PGOs and update the {fileName} file!");
+            var lineCount = File.ReadAllLines(path).Length;
+            Assert.True(lineCount > 6500, $"Jit Trace file line count of {lineCount} for {fileName} is less than 6500 lines! There is likely a bug removing lines from the linux trace.");
```
pragnagopa, line 46: this would mean if release process skips generating file we would not catch it?

azure-pipelines.yml
```diff
@@ -589,7 +589,6 @@ jobs:
         **\WebJobs.Script.Tests.Integration.csproj
   - task: DotNetCoreCLI@2
     displayName: "Release verification tests"
-    condition: eq(variables['isReleaseBranch'], 'True')
```
pragnagopa: my recommendation even though test is not perfect - we should keep this test in some format that verifies checked in file with the code.
Also, leave this test in release branch only to ensure PR builds are not blocked. But add the file read test in dev branch

cc @safihamid

VpOfEngineering:
I have made the updates @pragnagopa - I was unsure if you meant to keep the original test in, so I did and re-enabled it for release branches only.

The unit test is still failing due to the short Windows JIT trace - should I adjust the file size check for Windows?

test/WebJobs.Script.Tests/PreJit/JitReadTest.cs
```diff
@@ -0,0 +1,32 @@
...
+using Microsoft.Azure.WebJobs.Script.WebHost.Middleware;
+using Microsoft.Diagnostics.JitTrace;
+using Microsoft.WebJobs.Script.Tests;
+using Xunit;
+
+namespace Microsoft.Azure.WebJobs.Script.Tests.PreJIT
+{
+    [Trait(TestTraits.Category, TestTraits.EndToEnd)]
+    [Trait(TestTraits.Group, TestTraits.ReleaseTests)]
+    public class JitReadTest
```
pragnagopa: please add this test in the same test file - [JitPreparesTest.cs]

pragnagopa:
> The unit test is still failing due to the short Windows JIT trace - should I adjust the file size check for Windows?

Yes ! @safihamid  ?

pragnagopa (changes requested):
Updated the target branch to v3.x - commits should only go into dev branches and we should cherry-pick commits into release branch.

test/WebJobs.Script.Tests/PreJit/JitReadTest.cs
VpOfEngineering: Got it. So, both tests should run only during releases?

pragnagopa: possible to apply ReleaseTests trait at test level?

VpOfEngineering (commented):
Again, the current test will still fail once we merge into the release branch.

pragnagopa:
> Again, the current test will still fail once we merge into the release branch.

Probably missing some context here. Do we understand the issue?

VpOfEngineering:
> > Again, the current test will still fail once we merge into the release branch.
> 
> Probably missing some context here. Do we understand the issue?

Yes, this was as @balag0 discussed in the release channel. 
`public void ColdStart_JitFailuresTest(string fileName, double threshold)` needs to be removed and we need to simply check if the file exists and is larger than 6500 lines.

pragnagopa:
> Yes, this was as @balag0 discussed in the release channel.
public void ColdStart_JitFailuresTest(string fileName, double threshold) needs to be removed and we need to simply check if the file exists and is larger than 6500 lines.

@balag0  / @safihamid  - how would this ensure JIT trace file is updated as part of the release? Static file size check would always be true?

test/WebJobs.Script.Tests.Integration/PreJIT/JitPreparesTest.cs
```diff
@@ -33,5 +33,19 @@ public void ColdStart_JitFailuresTest(string fileName, double threshold)
             // using 1% as approximate number of allowed failures before we need to regenrate a new PGO file.
             Assert.True(failurePercentage < threshold, $"Number of failed PGOs are more than {threshold} percent! Current number of failures are {failedPrepares}. This will definitely impact cold start! Time to regenrate PGOs and update the {fileName} file!");
         }
+
+        [Theory]
+        [InlineData(WarmUpConstants.JitTraceFileName)]
+        [InlineData(WarmUpConstants.LinuxJitTraceFileName)]
```
safihamid, line 37: I think we should have 2 different tests here, one for Windows and one for Linux and each with different line count.

VpOfEngineering: Got it. What should the line count be for Windows?

balag0: I think, for windows, we should just use the existing assert (% of failures)
just for linux, instead of % of failures use line count (until we can figure out something better)
Btw these tests will be failing locally as well and if required we should feel free to adjust the threshold of 6500 to something else as required.

pragnagopa: Yes, leaving % of failures for windows and moving to checking file size for linux is OK!

safihamid: sounds good to me :)

pragnagopa (approved):
Please go ahead and resolve comments with notes before merging

PR #8166, opened 2022-02-24
Conditional based profiles for workers

### LINKED ISSUE #8184: Implement worker config profiles
Profiles enable workers to change how the worker is configured based on certain conditions, for example a worker may want to configure itself differently for a consumption application vs a dedicated application. The profiles will be defined in the `worker.config.json` file and will essentially override the default worker description based on a set of conditions.

Work to do:

- design the profile schema
- implement the profile parser logic

[Design doc](https://dev.azure.com/msazure/One/_git/AAPT-Antares-Docs/pullrequest/5715621?_a=files)

shreyas-gopalakrishna:
resolves #8184

Profiles enable workers to change how the worker is configured based on certain conditions, for example a worker may want to configure
itself differently for a consumption application vs a dedicated application. The profiles will be defined in the `worker.config.json` file
and will essentially override the default worker description based on a set of conditions.

The Java worker team is working on a prototype where an Application Insights (AI) agent is deployed along side the Java worker
to handle telemetry and logging to AI. Depending on the SKU and environment variables, the AI agent should be used. 
Profiles will be used to add arguments required to enable the agent. 

Sample `worker.config.json` with profiles 
``` json 
{
  "description": { ... }, // this will always be the default worker configuration
  "profiles": [
    {
      "name": "profile1",
      "conditions": [
        { "type": "environment", "name": "APP_SETTING_NAME", "expression": "app_setting_value" }
      ],
      "description": {
        "arguments": ["-DmyNewArg"]
      }
    },
    {
      "name": "profile2",
      "conditions": [
        { "type": "hostProperties", "name": "sku", "expression": "consumption" }
      ],
      "description": {
        "defaultWorkerPath": "",
        "arguments": []
      }
    }
  ]
}
```

### Pull request checklist


src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
```diff
@@ -118,6 +119,24 @@ internal void AddProvider(string workerDir)
...
+                    {
+                        if (!_environment.IsAppInsightsAgentOptOut())
+                        {
+                            workerDescription = GetWorkerDescriptionFromProfiles(RpcWorkerConstants.WorkerDescriptionAppInsightsAgentEnvironmentProfileName, descriptionProfiles, workerDescription);
+                        }
+                    }
+                    else
+                    {
+                        if (_environment.IsAppInsightsAgentEnabled())
+                        {
```
pragnagopa: @shreyas-gopalakrishna - We should not be relying on the App Setting  here. @liliankasem - can you please share notes from the meeting on approach discussed?

pragnagopa:
@shreyas-gopalakrishna - looking at the description - profile is added per app setting - profile based on Sku is more generic . If AppSetting based profile is more useful - lets discuss Schema with @liliankasem

src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
```diff
@@ -23,6 +23,8 @@ public static class EnvironmentSettingNames
         public const string AzureWebJobsSecretStorageSas = "AzureWebJobsSecretStorageSas";
         public const string AppInsightsInstrumentationKey = "APPINSIGHTS_INSTRUMENTATIONKEY";
         public const string AppInsightsConnectionString = "APPLICATIONINSIGHTS_CONNECTION_STRING";
+        public const string AppInsightsEnableAgent = "APPLICATIONINSIGHTS_ENABLE_AGENT";
+        public const string OptOutAppInsightsAgent = "OPT_OUT_APPLICATIONINSIGHTS_AGENT";
```
liliankasem: I thought we were not going to have a an environment variable for opt out?

liliankasem:
> @shreyas-gopalakrishna - looking at the description - profile is added per app setting - profile based on Sku is more generic . If AppSetting based profile is more useful - lets discuss Schema with @liliankasem

Yeah definitely worth discussing because it's a 2 part problem:

- The agent process is started with the worker based on the SKU 
- The agent process sends application insights data based on an environment variable

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerDescriptionProfile.cs
```diff
@@ -0,0 +1,67 @@
...
+        [JsonProperty(PropertyName = "description")]
+        public RpcWorkerDescription ProfileDescription { get; set; }
+
+        public bool Validate()
+        {
+            if (string.IsNullOrEmpty(Name))
+            {
+                return false;
+                // throw new ValidationException($"WorkerDescriptionProfile {nameof(Name)} cannot be empty");
+            }
```
shreyas-gopalakrishna: Validate conditions

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
```diff
@@ -159,6 +166,91 @@ internal void AddProvider(string workerDir)
...
+        {
+            List<RpcWorkerDescriptionProfile> descriptionProfiles = new List<RpcWorkerDescriptionProfile>();
+            JToken profiles = workerConfig.GetValue(RpcWorkerConstants.WorkerDescriptionProfiles);
+            if (profiles != null)
+            {
+                try
+                {
+                    descriptionProfiles = profiles.ToObject<List<RpcWorkerDescriptionProfile>>();
+                }
+                catch (Exception ex)
```
shreyas-gopalakrishna: Check if we should continue after exception or stop host

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
```diff
@@ -159,6 +166,91 @@ internal void AddProvider(string workerDir)
...
+        }
+
+        private RpcWorkerDescription LoadWorkerDescriptionFromProfiles(List<RpcWorkerDescriptionProfile> descriptionProfiles, RpcWorkerDescription defaultWorkerDescription)
+        {
+            foreach (var profile in descriptionProfiles)
+            {
+                if ( profile.Validate() && CheckProfileConditions(profile))
+                {
+                    RpcWorkerDescription profileDescription = profile.ProfileDescription;
+                    profileDescription.Arguments = profileDescription.Arguments?.Count > 0 ? profileDescription.Arguments : defaultWorkerDescription.Arguments;
```
liliankasem, line 194: Since `profileDescription.Arguments` should already be the default, I wonder if one of these might be cleaner to use:

a) single line if

src/WebJobs.Script/Workers/Rpc/Configuration/ProfileCondition.cs
```diff
@@ -0,0 +1,47 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Runtime.Serialization;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
+{
+    public enum ConditionType
+    {
+        [EnumMember(Value = "environment")]
```
fabiocav: I understand the approach, but ideally, this wouldn't be an enumeration, but instead, we'd resolve the conditions from a factory and keep this extensible, so adding a condition would require no modification to existing code, just a new implementation of a condition. This would also enable conditions to be added by simply registering new "condition providers" the factory could use, which could even be done/implemented outside of the host.

src/WebJobs.Script/Workers/Rpc/Configuration/ProfileCondition.cs
```diff
@@ -0,0 +1,47 @@
...
+        Environment,
+        [EnumMember(Value = "hostProperty")]
+        HostProperty
+    }
+
+    public enum ConditionHostPropertyName
+    {
+        [EnumMember(Value = "sku")]
```
fabiocav: These would ideally all be encapsulated by the "HostEnvironmentCondition" implementation.

src/WebJobs.Script/Workers/Rpc/Configuration/ProfileCondition.cs
```diff
@@ -0,0 +1,47 @@
...
+        HostVersion
+    }
+
+    public class ProfileCondition
+    {
+        /// <summary>
+        /// Gets or sets the type of the condition
+        /// </summary>
+        [JsonProperty(PropertyName = "type")]
+        public ConditionType Type { get; set; }
```
fabiocav: This would ideally be the only property we care about for a condition implementation.

If you think about the interface for this (e.g. something like `IWorkerProfileCondition`), it would have a "Type" property (`string`), and a method for evaluation.

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerDescriptionProfile.cs
```diff
@@ -0,0 +1,67 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
+{
+    public class RpcWorkerDescriptionProfile
```
fabiocav: nit: move this to `Microsoft.Azure.WebJobs.Script.Workers` and rename to `WorkerDescriptionProfile`

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
```diff
@@ -159,6 +166,91 @@ internal void AddProvider(string workerDir)
...
+                {
+                    return false;
+                }
+            }
+            return true;
+        }
+
+        private bool ValidateProfileCondition(ProfileCondition condition)
+        {
+            string value = string.Empty;
```
fabiocav: With the condition implementation approach described above, this logic is encapsulated by each condition, removing the responsibility from this type, and the profile evaluation becomes a matter of ensuring all conditions return true from their validation method.

fabiocav (commented):
The logic here is looking great! 

Providing an initial round of feedback on the implementation.

src/WebJobs.Script/Workers/Rpc/Configuration/ProfileCondition.cs
liliankasem: Ahh I like the sound of that!

src/WebJobs.Script/Workers/Rpc/Configuration/ConditionType.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) .NET Foundation. All rights reserved.
```
shreyas-gopalakrishna: Check: Move files not related to Rpc to src/WebJobs.Script/Workers/

src/WebJobs.Script/Workers/Rpc/Configuration/EnvironmentCondition.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        public EnvironmentCondition(ILogger logger, ISystemRuntimeInformation systemRuntimeInformation, IEnvironment environment, string name, string expression)
+        {
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            _environment = environment ?? throw new ArgumentNullException(nameof(environment));
+            _systemRuntimeInformation = systemRuntimeInformation ?? throw new ArgumentNullException(nameof(systemRuntimeInformation));
+            Type = ConditionType.Environment;
+            Name = name;
+            Expression = expression;
+            Validate();
```
shreyas-gopalakrishna: Is it right to call `Validate();` condition here?

src/WebJobs.Script/Workers/Rpc/Configuration/IWorkerProfileCondition.cs
```diff
@@ -0,0 +1,18 @@
...
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.WebJobs.Script.Workers
+{
+    public interface IWorkerProfileCondition
+    {
+        /// <summary>
+        /// Gets the type of the condition
+        /// </summary>
+        [JsonProperty(PropertyName = "type")]
```
shreyas-gopalakrishna: Check if `[JsonProperty(PropertyName = "type")]` is necessary

src/WebJobs.Script/Workers/Rpc/Configuration/IWorkerProfileConditionFactory.cs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Script.Workers
+{
+    internal interface IWorkerProfileConditionFactory
+    {
+        IWorkerProfileCondition CreateWorkerProfileCondition(string type, string name, string expression);
```
shreyas-gopalakrishna: `CreateWorkerProfileCondition` is still dependent on name and expression. How to remove this dependency? Do we pass the JToken/JObject around as a context to the factory?

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
```diff
@@ -159,6 +168,59 @@ internal void AddProvider(string workerDir)
...
+            List<WorkerDescriptionProfile> descriptionProfiles = new List<WorkerDescriptionProfile>();
+            JArray profiles = workerConfig.GetValue(RpcWorkerConstants.WorkerDescriptionProfiles) as JArray;
+            if (profiles != null)
+            {
+                try
+                {
+                    //descriptionProfiles = profiles.ToObject<List<WorkerDescriptionProfile>>();
+                    foreach (JObject item in profiles)
+                    {
+                        descriptionProfiles.Add(JsonConvert.DeserializeObject<WorkerDescriptionProfile>(item.ToString(Formatting.None), new WorkerDescriptionProfileConverter(_logger, _systemRuntimeInformation, _environment)));
```
shreyas-gopalakrishna: Do we use a custom JsonConverter `WorkerDescriptionProfileConverter` or reply on the default converter. This is the only place I could think of using the `IWorkerProfileConditionFactory`. Is there a better way to do this?

src/WebJobs.Script/Workers/Rpc/Configuration/WorkerDescriptionProfileConverter.cs
```diff
@@ -0,0 +1,61 @@
...
+            workerDescriptionProfile.ProfileDescription = jObject.Property(WorkerConstants.WorkerDescription).Value.ToObject<RpcWorkerDescription>();
+
+            WorkerProfileConditionFactory workerProfileConditionFactory = new WorkerProfileConditionFactory(_logger, _systemRuntimeInformation, _environment);
+            workerDescriptionProfile.Conditions = new List<IWorkerProfileCondition>();
+
+            JArray conditions = jObject.GetValue("conditions") as JArray;
+
+            foreach (JObject conditionJObject in conditions)
+            {
+                IWorkerProfileCondition workerProfileCondition = workerProfileConditionFactory.CreateWorkerProfileCondition((string)conditionJObject["type"], (string)conditionJObject["name"], (string)conditionJObject["expression"]);
```
shreyas-gopalakrishna: `(string)conditionJObject["type"]` check if this is hardcoding is okay or if it must be defined as constants?

src/WebJobs.Script/Workers/Rpc/Configuration/WorkerProfileConditionFactory.cs
```diff
@@ -0,0 +1,36 @@
...
+        private readonly IEnvironment _environment;
+
+        public WorkerProfileConditionFactory(ILogger logger, ISystemRuntimeInformation systemRuntimeInfo, IEnvironment environment)
+        {
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            _environment = environment ?? throw new ArgumentNullException(nameof(environment));
+            _systemRuntimeInformation = systemRuntimeInfo ?? throw new ArgumentNullException(nameof(systemRuntimeInfo));
+        }
+
+        public IWorkerProfileCondition CreateWorkerProfileCondition(string type, string name, string expression)
```
shreyas-gopalakrishna: check if factory must depend on only type and a context instead of name and expression here

src/WebJobs.Script/Workers/WorkerDescriptionProfile.cs
```diff
@@ -0,0 +1,50 @@
...
+
+namespace Microsoft.Azure.WebJobs.Script.Workers
+{
+    //[JsonConverter(typeof(WorkerDescriptionProfileConverter))]
+    public class WorkerDescriptionProfile
+    {
+        /// <summary>
+        /// Gets or sets the name of the profile.
+        /// </summary>
+        [JsonProperty(PropertyName = "name")]
```
shreyas-gopalakrishna: Check if `[JsonProperty(PropertyName = "name")]` must be used or the `[JsonConverter(typeof(WorkerDescriptionProfileConverter))]` must be used.

src/WebJobs.Script/Workers/Rpc/Configuration/WorkerDescriptionProfileConverter.cs
```diff
@@ -0,0 +1,61 @@
...
+using System.Collections.Generic;
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs.Script.Workers
+{
+    public class WorkerDescriptionProfileConverter : JsonConverter
+    {
+        private ILogger _logger;
```
shreyas-gopalakrishna: Check if these can be dependency injected?

src/WebJobs.Script/Workers/Rpc/Configuration/WorkerProfileConditionFactory.cs
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Script.Workers
+{
+    internal class WorkerProfileConditionFactory : IWorkerProfileConditionFactory
```
shreyas-gopalakrishna: Check if data to the factory can be dependency injected

src/WebJobs.Script/Workers/Rpc/Configuration/IWorkerProfileCondition.cs
fabiocav: Shouldn't be required. Ultimately, the less we're tied to JSON.NET here, the better.

src/WebJobs.Script/Workers/Rpc/Configuration/IWorkerProfileConditionFactory.cs
fabiocav: Don't take a dependency on those types, instead, just pass the properties in as a dictionary or something along those lines.

src/WebJobs.Script/Workers/Rpc/Configuration/ConditionType.cs
```diff
@@ -0,0 +1,15 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Runtime.Serialization;
+
+namespace Microsoft.Azure.WebJobs.Script.Workers
+{
+    public enum ConditionType
+    {
+        [EnumMember(Value = "environment")]
```
fabiocav: We want the ability to expand the condition types without having to modify host types, also making them injectable. This enum shouldn't be required.

src/WebJobs.Script/Workers/Rpc/Configuration/IWorkerProfileCondition.cs
```diff
@@ -0,0 +1,18 @@
...
+        public ConditionType Type { get; }
```
fabiocav: `Type` here should be a string.

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
```diff
@@ -159,6 +168,59 @@ internal void AddProvider(string workerDir)
             }
         }
 
+        private List<WorkerDescriptionProfile> ReadWorkerDescriptionFromProfiles(JObject workerConfig)
```
fabiocav: nit: rename method to `ReadWorkerDescriptionProfiles`

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
```diff
@@ -159,6 +168,59 @@ internal void AddProvider(string workerDir)
...
+        {
+            List<WorkerDescriptionProfile> descriptionProfiles = new List<WorkerDescriptionProfile>();
```
fabiocav: Move the actual instantiation info your `if`, and create the list with the correct count. Return empty if no profiles are defined. This will avoid instantiating a list here if not required.

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
```diff
@@ -159,6 +168,59 @@ internal void AddProvider(string workerDir)
...
+                    profileDescription.Extensions = profileDescription.Extensions?.Count > 0 ? profileDescription.Extensions : defaultWorkerDescription.Extensions;
+                    profileDescription.Language = string.IsNullOrEmpty(profileDescription.Language) ? defaultWorkerDescription.Language : profileDescription.Language;
+                    profileDescription.WorkerDirectory = string.IsNullOrEmpty(profileDescription.WorkerDirectory) ? defaultWorkerDescription.WorkerDirectory : profileDescription.WorkerDirectory;
+                    return profileDescription;
+                }
+            }
+            return defaultWorkerDescription;
+        }
+
+        private bool CheckProfileConditions(WorkerDescriptionProfile profile)
```
fabiocav: Move this logic into the `WorkerDescriptionProfile` and expose a method called something like `EvaluateConditions`

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
```diff
@@ -159,6 +168,59 @@ internal void AddProvider(string workerDir)
...
+                }
+            }
+            return descriptionProfiles;
+        }
+
+        private RpcWorkerDescription LoadWorkerDescriptionFromProfiles(List<WorkerDescriptionProfile> workerDescriptionProfiles, RpcWorkerDescription defaultWorkerDescription)
+        {
+            foreach (var profile in workerDescriptionProfiles)
+            {
+                if (CheckProfileConditions(profile))
```
fabiocav: Call the `EvaluateConditions` method mentioned below on the profile.

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
```diff
@@ -159,6 +168,59 @@ internal void AddProvider(string workerDir)
...
+                    profileDescription.Arguments = profileDescription.Arguments?.Count > 0 ? profileDescription.Arguments : defaultWorkerDescription.Arguments;
+                    profileDescription.DefaultExecutablePath = string.IsNullOrEmpty(profileDescription.DefaultExecutablePath) ? defaultWorkerDescription.DefaultExecutablePath : profileDescription.DefaultExecutablePath;
+                    profileDescription.DefaultWorkerPath = string.IsNullOrEmpty(profileDescription.DefaultWorkerPath) ? defaultWorkerDescription.DefaultWorkerPath : profileDescription.DefaultWorkerPath;
+                    profileDescription.Extensions = profileDescription.Extensions?.Count > 0 ? profileDescription.Extensions : defaultWorkerDescription.Extensions;
+                    profileDescription.Language = string.IsNullOrEmpty(profileDescription.Language) ? defaultWorkerDescription.Language : profileDescription.Language;
+                    profileDescription.WorkerDirectory = string.IsNullOrEmpty(profileDescription.WorkerDirectory) ? defaultWorkerDescription.WorkerDirectory : profileDescription.WorkerDirectory;
+                    return profileDescription;
```
fabiocav: Move this logic to the profile, exposing a method called `ApplyProfile` that takes the description. 

There, you can also introduce some helpers to reduce repetition with the null/empty and other checks.

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
fabiocav: Here we'd use the factory directly, after deserializing the profile and getting the type and dictionary of properties.

src/WebJobs.Script/Workers/Rpc/Configuration/WorkerProfileConditionFactory.cs
```diff
@@ -0,0 +1,36 @@
...
+            _environment = environment ?? throw new ArgumentNullException(nameof(environment));
+            _systemRuntimeInformation = systemRuntimeInfo ?? throw new ArgumentNullException(nameof(systemRuntimeInfo));
+        }
+
+        public IWorkerProfileCondition CreateWorkerProfileCondition(string type, string name, string expression)
+        {
+            Enum.TryParse(type, out ConditionType conditionType);
+            switch (conditionType)
+            {
+                case ConditionType.HostProperty:
```
fabiocav: This logic should be directly on the string type. The exception below should only be thrown if no condition provider matches the type

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
```diff
@@ -40,6 +41,7 @@ public RpcWorkerConfigFactory(IConfiguration config, ILogger logger, ISystemRunt
             {
                 WorkersDirPath = workersDirectorySection.Value;
             }
+            _environment.SetEnvironmentVariable(EnvironmentSettingNames.WorkerDirectoryPath, WorkersDirPath);
```
fabiocav: Why is this being set? Seems outside of the scope of this change.

fabiocav (commented):
Quick round of feedback. We can sync on the comments, if needed.

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
liliankasem: We can create a new PR just for this change, but it's semi profiles related; essentially this is for the workers to be able to get the worker path through something like  `%WORKER_PATH%` in the arguments property in the worker.config file etc.

src/WebJobs.Script/Workers/Rpc/Configuration/ConditionType.cs
liliankasem: yeah if it's not rpc you can move it, if it makes sense to create a folder for worker config / profile we can do that but /workers should also be ok

src/WebJobs.Script/Workers/Rpc/Configuration/EnvironmentCondition.cs
liliankasem: Calling this here mean that we every time we instantiate a new condition, it would be validated. Ideally this should only happen once so I think it's fine to call here. The only other place I can think of would be at the factory level maybe. But I think this is fine - @fabiocav ?

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
liliankasem: nit: this can just be var since it's very clear what the type is on the right side of the expression
```suggestion
            var descriptionProfiles = new List<WorkerDescriptionProfile>();
```

src/WebJobs.Script/Workers/Rpc/Configuration/WorkerDescriptionProfileConverter.cs
liliankasem: These 3 should already be dependency injection in the project so they should be available for you to use

src/WebJobs.Script/Workers/Rpc/Configuration/WorkerDescriptionProfileConverter.cs
liliankasem: I personally prefer defining constants than hardcoding strings but here it is probably ok. If I understood @fabiocav correct, you would not need this custom converter (but the logic for the condition parsing would stay)

src/WebJobs.Script/Workers/HostPropertyCondition.cs
```diff
@@ -0,0 +1,93 @@
...
+        public void Validate()
+        {
+            if (string.IsNullOrEmpty(Name))
+            {
+               throw new ValidationException($"HostPropertyCondition {nameof(Name)} cannot be empty.");
+            }
+
+            if (!Enum.GetNames(typeof(ConditionHostPropertyName)).Any(x => x.ToLower().Contains(Name)))
+            {
+               throw new ValidationException($"HostPropertyCondition {nameof(Name)} cannot be a valid host property name.");
```
liliankasem:
```suggestion
               throw new ValidationException($"HostPropertyCondition {nameof(Name)} is not a valid host property name.");
```

src/WebJobs.Script/Workers/EnvironmentCondition.cs
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) .NET Foundation. All rights reserved.
```
liliankasem: the number of conditions might grow in future so it might be worth create a new directory such as `/Profiles/Conditions`

src/WebJobs.Script/Workers/WorkerProfileConditionFactory.cs
```diff
@@ -0,0 +1,36 @@
...
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            _environment = environment ?? throw new ArgumentNullException(nameof(environment));
+            _systemRuntimeInformation = systemRuntimeInfo ?? throw new ArgumentNullException(nameof(systemRuntimeInfo));
+        }
+
+        public IWorkerProfileCondition CreateWorkerProfileCondition(IDictionary<string, object> properties)
+        {
+            switch (properties[WorkerConstants.WorkerDescriptionProfileConditionType])
+            {
+                case "hostProperty":
```
liliankasem: i think we should make all of these profile strings these constants

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
```diff
@@ -159,6 +167,51 @@ internal void AddProvider(string workerDir)
             }
         }
 
+        private List<WorkerDescriptionProfile> ReadWorkerDescriptionProfiles(JObject workerConfig)
+        {
+            JArray profiles = workerConfig.GetValue(RpcWorkerConstants.WorkerDescriptionProfiles) as JArray;
+            if (profiles != null)
```
liliankasem, line 174: wondering if it might be cleaner to reserve this if statement i.e `if null .. do xyz`

src/WebJobs.Script/Workers/WorkerDescriptionProfile.cs
```diff
@@ -0,0 +1,79 @@
...
+            }
+            return true;
+        }
+
+        /// <summary>
+        /// Overrides the worker description parameters to that available in the profile
+        /// </summary>
+        public RpcWorkerDescription ApplyProfile(RpcWorkerDescription defaultWorkerDescription)
+        {
+            defaultWorkerDescription.Arguments = ProfileDescription.Arguments?.Count > 0 ? ProfileDescription.Arguments : defaultWorkerDescription.Arguments;
```
liliankasem: I think Fabio mentioned maybe using a helper function here to reduce the repetition of `ProfileDescription.Arguments?.Count > 0 ? ProfileDescription.Arguments : defaultWorkerDescription.Arguments;`

PR #8200, opened 2022-03-04
Host to send load request correctly after fallback is requested by the worker

### LINKED ISSUE #8199: Project stein - Host not sending load request after fallback requested from worker
For worker indexing/ project stein, host is falling back to its own indexing after worker denies. The functions are being indexed correctly by host but load requests are not being sent correctly. 

Fix - A minor change is needed to call finish initialization method from RPCFunctionInvocationDispatcher which will send the load requests after host fallback.

surgupta-msft:
resolves #8199 

For worker indexing/ project stein, host is falling back to its own indexing after worker denies. The functions are being indexed correctly by the host, but load requests are not being sent correctly. This PR has implementation to fix the issue as well as added new tests around issue.


satvu (approved):

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -590,6 +590,14 @@ internal void ProcessFunctionMetadataResponses(FunctionMetadataResponse function
                     });
                 }
             }
+            else
+            {
+                functions.Add(new RawFunctionMetadata()
+                {
+                    UseDefaultMetadataIndexing = functionMetadataResponse.UseDefaultMetadataIndexing
```
soninaren, line 597: renaming to `UseHostIndexing` would be little less ambiguous, Since default indexing can move between host and worker

src/WebJobs.Script/Host/AggregateFunctionMetadataProvider.cs
```diff
@@ -62,17 +62,21 @@ public async Task<ImmutableArray<FunctionMetadata>> GetFunctionMetadataAsync(IEn
                     // get function metadata from worker, then validate it
                     rawFunctions = await _dispatcher.GetWorkerMetadata();
 
-                    if (!IsNullOrEmpty(rawFunctions))
+                    if (IsDefaultIndexingRequired(rawFunctions))
```
soninaren, line 65: same here, can we rename to `HostIndexingRequired`

src/WebJobs.Script/Host/AggregateFunctionMetadataProvider.cs
```diff
@@ -62,17 +62,21 @@ public async Task<ImmutableArray<FunctionMetadata>> GetFunctionMetadataAsync(IEn
...
+
+                    // set up invocation buffers and send load requests
+                    await _dispatcher.FinishInitialization(functions);
                 }
-                if (!workerIndexing || IsDefaultIndexingRequired(rawFunctions))
+                else
                 {
-                    // If worker denies indexing then falling back to the host for Indexing
-                    _logger.LogDebug("Fallback to host indexing as worker denied indexing");
                     functions = await _hostFunctionMetadataProvider.GetFunctionMetadataAsync(workerConfigs, environment, forceRefresh);
```
soninaren, line 80: What scenario would reach here? both if and else call `_hostFunctionMetadataProvider.GetFunctionMetadataAsync`

src/WebJobs.Script/Host/AggregateFunctionMetadataProvider.cs
```diff
@@ -27,7 +27,7 @@ public class AggregateFunctionMetadataProvider : IFunctionMetadataProvider
         private IFunctionMetadataProvider _hostFunctionMetadataProvider;
 
         public AggregateFunctionMetadataProvider(
-            ILogger<AggregateFunctionMetadataProvider> logger,
```
soninaren, line 30: why is this type changed?

soninaren (changes requested):

src/WebJobs.Script/Host/AggregateFunctionMetadataProvider.cs
surgupta-msft: I referred to other classes like FunctionMetadataManager and RPCFunctionInvocationDispatcher, they do not have typed Logger. As I was adding more tests for AggregateFunctionMetadataProvider, I noticed that the typed Logger is not allowing me to pass **TestLogger** which is also of type _ILogger_. So, I thought of just following the pattern used in other classes and their Test files and made it just _ILogger._

src/WebJobs.Script/Host/AggregateFunctionMetadataProvider.cs
surgupta-msft: Host fallback scenario will reach to this _if block_. The final _else_ is for a case where workerIndexing is not enabled at all. 

For host fallback, I still need to do `FinishInitialization(functions);` which is not needed for legacy flow.

src/WebJobs.Script/Host/AggregateFunctionMetadataProvider.cs
surgupta-msft: Done

soninaren (approved):

Issue #7591, opened 2021-08-07
BlobStorageSecretsRepository Write failure should fail key creation request


mathewc:
We're seeing cases where a PUT to admin/host/functionkeys/testkey0 succeed and return 201 created, even though there was an exception thrown from the blob persistence logic [here](https://github.com/Azure/azure-functions-host/blob/91b7d1c780f2e040d32b02b38f6af5bab4799c96/src/WebJobs.Script.WebHost/Security/KeyManagement/BlobStorageSecretsRepository.cs#L191). So somewhere upstream we must be swallowing this exception. If the persistence fails, the PUT request should fail. 

Additionally, the error message is rather poor in these cases. We only log "There was an error performing a write operation on the Blob Storage Secret Repository. Please ensure the 'AzureWebJobsStorage' connection string is valid.". We should provide more details on the actual underlying storage exception, rather than offering the unhelpful bit on checking the storage connection string.

See recent CRI #251344724

fabiocav:
@soninaren it would be good to validate the read paths as @liliankasem ran into some issues that could be similar in the read path.

PR #8240, opened 2022-03-17
Add exception details to error logging for blob repository

liliankasem:
Logging the exception for blob repository errors so we have more details when a failure occurs. 

Related to #7591

### Pull request checklist

**before**
Screen Shot 2022-03-29 at 6 56 09 PM

**after**
Screen Shot 2022-03-30 at 4 56 11 PM

### DIFF
--- a/src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceSecurityExtension.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceSecurityExtension.cs
@@ -9,18 +9,17 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics.Extensions
     public static class ScriptHostServiceSecurityExtension
     {
         // EventId range is 600-699
-
         private static readonly Action<ILogger, string, string, Exception> _blobStorageSecretRepoError =
             LoggerMessage.Define<string, string>(
                 LogLevel.Error,
                 new EventId(600, nameof(BlobStorageSecretRepoError)),
-                "There was an error performing a {operation} operation on the Blob Storage Secret Repository. Please ensure the '{appSettingName}' connection string is valid. {exception}.");
+                "There was an error performing a {operation} operation on the Blob Storage Secret Repository.\nException: {exception}.");
 
         private static readonly Action<ILogger, string, string, Exception> _blobStorageSecretSasRepoError =
             LoggerMessage.Define<string, string>(
                 LogLevel.Error,
                 new EventId(601, nameof(BlobStorageSecretSasRepoError)),
-                "There was an error performing a {operation} operation on the Blob Storage Secret Repository. Please ensure the '{appSettingName}' SAS URL has Read, Write, and List permissions. {exception}");
+                "There was an error performing a {operation} operation on the Blob Storage Secret Repository.\nException: {exception}");
 
         public static void BlobStorageSecretRepoError(this ILogger logger, string operation, string appSettingName, Exception exception)
         {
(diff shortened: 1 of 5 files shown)


src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceSecurityExtension.cs
```diff
@@ -14,22 +14,22 @@ public static class ScriptHostServiceSecurityExtension
             LoggerMessage.Define<string, string>(
                 LogLevel.Error,
                 new EventId(600, nameof(BlobStorageSecretRepoError)),
-                "There was an error performing a {operation} operation on the Blob Storage Secret Repository. Please ensure the '{appSettingName}' connection string is valid.");
+                "There was an error performing a {operation} operation on the Blob Storage Secret Repository. Please ensure the '{appSettingName}' connection string is valid. {exception}.");
```
kshyju: Do we need to explicitly include the exception in the format string? I think [it gets passed to the internal Log method](https://github.com/dotnet/runtime/blob/37c5a3e6bf685bb08550baba54a742d21c91ffe7/src/libraries/Microsoft.Extensions.Logging.Abstractions/src/LoggerMessage.cs#L184) automagically.

kshyju: Also, the number of named parameters in message string (3) not matching with the definition (2).

liliankasem: From the kusto logs I was looking at, we don't see any exception details

liliankasem: @kshyju I doubled checked this and you're right it is included. So I will remove it from the message string :)

src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceSecurityExtension.cs
```diff
@@ -13,13 +13,13 @@ public static class ScriptHostServiceSecurityExtension
             LoggerMessage.Define<string, string>(
                 LogLevel.Error,
                 new EventId(600, nameof(BlobStorageSecretRepoError)),
-                "There was an error performing a {operation} operation on the Blob Storage Secret Repository.");
+                "There was an error performing a {operation} operation on the Blob Storage Secret Repository. Please ensure the '{appSettingName}' connection string is valid.");
```
kshyju: I wonder the "_Please ensure the '{appSettingName}' connection string is valid_." part of message may cause confusion. Is it possible that this error logging code is executed for cases where the connection string is valid, but the READ/WRITE/LIST operations threw an exception for other reasons? For example, the `BlobContainerClient.GetBlobsAsync` could [throw a RequestFailedException](https://docs.microsoft.com/en-us/dotnet/api/azure.storage.blobs.blobcontainerclient.getblobsasync?view=azure-dotnet#remarks)

kshyju (approved):
LGTM. One question about the message we are logging.

src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceSecurityExtension.cs
liliankasem: Yeah I am debating this one myself, it's almost a red herring and is might be best removed. But yes, this error shows up when it is not an auth issue which is where this PR came from in the first place :D 

On one hand, it's useful information when the issue is auth related, on the other hand if the issue is something else, this is useless information. I think I'm leaning towards removing that part of the string and just making sure our documentation is updated. 

Something else worth noting, I am not seeing this error surface in AI or the log stream - should it be? Or should we leave this for our internal logs?

satvu (approved):

PR #8260, opened 2022-03-24
Add hex exit code to message when worker fails

ejizba:
We sometimes gets issues and CRIs like [this](https://github.com/Azure/azure-functions-nodejs-worker/issues/556) where the node process exits, but the integer exit code (`-1073741819`) doesn't come up with many results on google. By contrast, the hex version (`0xC0000005`) usually has a lot more results, including docs like [this](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/596a1078-e883-4972-9bbc-49e60bebca55).

## Previous message

> node exited with code -1073741819

## New message

> node exited with code -1073741819 (0xC0000005)

### Pull request checklist

    * [x] Otherwise: Not entirely sure this needs to be backported, although I supposed it could help supportability so might be worth it. Let's see what people think in the first place

### DIFF
--- a/src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs
+++ b/src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs
@@ -158,7 +158,7 @@ namespace Microsoft.Azure.WebJobs.Script.Workers
                 }
                 else
                 {
-                    var processExitEx = new WorkerProcessExitException($"{Process.StartInfo.FileName} exited with code {Process.ExitCode}", new Exception(exceptionMessage));
+                    var processExitEx = new WorkerProcessExitException($"{Process.StartInfo.FileName} exited with code {Process.ExitCode} (0x{Process.ExitCode.ToString("X")})", new Exception(exceptionMessage));
                     processExitEx.ExitCode = Process.ExitCode;
                     processExitEx.Pid = Process.Id;
                     HandleWorkerProcessExitError(processExitEx);


liliankasem (approved):

src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs
```diff
@@ -156,7 +156,7 @@ private void OnProcessExited(object sender, EventArgs e)
                 }
                 else
                 {
-                    var processExitEx = new WorkerProcessExitException($"{Process.StartInfo.FileName} exited with code {Process.ExitCode}\n {exceptionMessage}");
+                    var processExitEx = new WorkerProcessExitException($"{Process.StartInfo.FileName} exited with code {Process.ExitCode} (0x{Process.ExitCode.ToString("X")})\n {exceptionMessage}");
```
kshyju: Do we know anything like a detector is relying on the exact message (previous version), which will break with this change?

kshyju (approved):

src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs
ejizba: I haven't contributed to detectors yet, is there a source repo for those that I could search?

kshyju: I am not sure since I also haven't contributed to detectors :) @kashimiz might know.

fabiocav:
@ejizba going through some of the old PRs and it would be good to have this merged. Can you please resolve the conflict so we can get his change in?
Thanks!

src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs
ejizba: I discussed with Anatoli a bit and it does not seem like there is a searchable repo for this. The only place he said we could contribute detectors was in the "Develop" section of app lense, which only has "Your detectors" listed, not everyone's. Otherwise you can go to each individual detector and check the "Develop" section for that detector.

I will do my best to check some of the default detectors that show up on a function app in App Lense, but overall I personally think the change is worth the potential risk of temporarily breaking a detector or two

kshyju: Makes sense. Thanks!

PR #8263, opened 2022-03-24
Updating identity dependencies to address an issue with their internal type usage

### LINKED ISSUE #8249: Rollback of function runtime 4.2.1 in Azure?
Hi All,

It appears that on Friday all our Function App Runtimes were upgraded to v4.2.1.17812 which caused Library compatibility issues on every one of our functions in every environment, 
Specifically with the upgrade of Microsoft.IdentityModel.Tokens to v6.15.1 found in commit https://github.com/Azure/azure-functions-host/commit/d02fe0862445d0662ec68695c062555bbc1e0864

We mitigated this by pinning all our runtimes to a previously working version v  4.1.3.17473.

This allowed our services to come back while we upgrade our code to move to the updated IdentityModel library.

We are running net6.0 'In-process' functions under v4, so host library dependency changes like this is a massive problem for us. 

When trying to verify the library upgrade, and force the broken behaviour in Azure by manually setting the runtime to v4.2.1 I get an error in the "Function runtime settings"

image

This seems to indicate that the release was rolled back, can someone confirm this?

Does this mean that there is no need to migrate to the IdentityModel 6.15.1? Or will it come out at a later stage?

Is there a place that publishes when breaking changes like this are happening?  or that shows the library versions of each released runtime version?

Regards,

Basil

fabiocav:

resolves #8249

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.WebHost/WebJobs.Script.WebHost.csproj
+++ b/src/WebJobs.Script.WebHost/WebJobs.Script.WebHost.csproj
@@ -10,6 +10,7 @@
     <AspNetCoreHostingModel>InProcess</AspNetCoreHostingModel>
     <TieredCompilation>false</TieredCompilation>
     <NoWarn>NU5104</NoWarn>
+    <IdentityDependencyVersion>6.15.1</IdentityDependencyVersion>
   </PropertyGroup>
   <PropertyGroup Condition="'$(RuntimeIdentifier)' != ''">
     <PublishReadyToRun>true</PublishReadyToRun>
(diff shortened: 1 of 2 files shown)


fabiocav:
Tests that will ensure this continues to be validated will land with a PR @kashimiz is working on.

I'll have some follow up changes to ensure more information is available when the token validation process fails, but this will come in a separate PR.

mathewc (approved):

kashimiz (approved):

brettsam (approved):

Issue #8258, opened 2022-03-23
Analyzer: Illegal Function Name


kashimiz:
Part of https://github.com/Azure/azure-functions-host/issues/7531

See https://github.com/Azure/azure-webjobs-sdk/blob/4e287eb443e2445a298a8a959f7cd01efc46e2c9/src/Analyzer/WebJobs.VSAnalyzer.Core/ErrorList.cs#L104

PR #8264, opened 2022-03-24
WebJobs Analyzer: Function name validation

kashimiz:
Addresses #8258 

Related: https://github.com/Azure/azure-webjobs-sdk/pull/2842

Requires release of updated WebJobs SDK to build.


src/WebJobs.Script.Analyzers/AssemblyCache.cs
```diff
@@ -0,0 +1,190 @@
...
+        bool _registered;
+
+        // Assembly Display Name --> Path
+        Dictionary<string, string> _map = new Dictionary<string, string>();
+
+        // Assembly Display Name --> loaded Assembly object
+        Dictionary<string, Assembly> _mapRef = new Dictionary<string, Assembly>();
+
+        const string WebJobsAssemblyName = "Microsoft.Azure.WebJobs";
+        const string WebJobsHostAssemblyName = "Microsoft.Azure.WebJobs.Host";
```
brettsam: can you explicitly mark all these fields as `private`?

src/WebJobs.Script.Analyzers/AssemblyCache.cs
```diff
@@ -0,0 +1,190 @@
...
+
+                    _map[dispName] = path;
+                }
+
+                // Builtins
+                _mapRef["mscorlib"] = typeof(object).Assembly;
+                _mapRef[WebJobsAssemblyName] = typeof(Microsoft.Azure.WebJobs.FunctionNameAttribute).Assembly;
+                _mapRef[WebJobsHostAssemblyName] = typeof(Microsoft.Azure.WebJobs.JobHost).Assembly;
+
+                // JSON.Net?
```
brettsam: remove this comment?

src/WebJobs.Script.Analyzers/AssemblyCache.cs
```diff
@@ -0,0 +1,190 @@
...
+
+        bool _registered;
+
+        // Assembly Display Name --> Path
+        Dictionary<string, string> _map = new Dictionary<string, string>();
+
+        // Assembly Display Name --> loaded Assembly object
+        Dictionary<string, Assembly> _mapRef = new Dictionary<string, Assembly>();
+
+        const string WebJobsAssemblyName = "Microsoft.Azure.WebJobs";
```
liliankasem: Do we want to put these in a constants file?

src/WebJobs.Script.Analyzers/AssemblyCache.cs
```diff
@@ -0,0 +1,190 @@
...
+                try
+                {
+                    // See GetNuGetPackagesPath for details
+                    // Script runtime is already setup with assembly resolution hooks, so use LoadFrom
+                    assembly = Assembly.LoadFrom(path);
+
+                    string asmName = new AssemblyName(assembly.FullName).Name;
+                    _mapRef[asmName] = assembly;
+
+                    var test = assembly.GetCustomAttributes<WebJobsStartupAttribute>().Select(a => a.WebJobsStartupType);
```
liliankasem: Not sure that I understand why the variable is named test? It looks like you're selecting a startup type? I might be missing some context

src/WebJobs.Script.Analyzers/AssemblyCache.cs
```diff
@@ -0,0 +1,190 @@
...
+            {
+                return;
+            }
+            AppDomain.CurrentDomain.AssemblyResolve += CurrentDomain_AssemblyResolve;
+            _registered = true;
+        }
+
+        private Assembly CurrentDomain_AssemblyResolve(object sender, ResolveEventArgs args)
+        {
+            var an = new AssemblyName(args.Name);
```
liliankasem: nit: I think it might be easier to follow the code with the full variable name
```suggestion
            var assemblyName = new AssemblyName(args.Name);
```

src/WebJobs.Script.Analyzers/AssemblyCache.cs
```diff
@@ -0,0 +1,190 @@
...
+            var context = args.RequestingAssembly;
+
+            Assembly asm2;
```
liliankasem: what does asm2 stand for?

src/WebJobs.Script.Analyzers/AssemblyCache.cs
```diff
@@ -0,0 +1,190 @@
...
+
+                // Simplifying assumption: assume dll name matches assembly name.
+                // Use this as a filter to limit the number of file-touches.
+                if (string.Equals(filename, an.Name, StringComparison.OrdinalIgnoreCase))
+                {
+                    var an2 = AssemblyName.GetAssemblyName(path);
+
+                    if (string.Equals(an2.FullName, an.FullName, StringComparison.OrdinalIgnoreCase))
+                    {
+                        var a = Assembly.LoadFrom(path);
```
liliankasem: similar nit here around using meaningful variable names i.e. `loadedAssembly`

src/WebJobs.Script.Analyzers/CompilationWebJobsStartuptypeLocator.cs
```diff
@@ -0,0 +1,37 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Linq;
+using System.Reflection;
+using Microsoft.Azure.WebJobs.Hosting;
+
+namespace Microsoft.Azure.Functions.Analyzers
+{
+    public class CompilationWebJobsStartupTypeLocator : IWebJobsStartupTypeLocator
```
liliankasem: Mind adding a class comment here?

src/WebJobs.Script.Analyzers/Constants.cs
```diff
@@ -14,6 +14,7 @@ internal static class DiagnosticsCategories
         {
             public const string Reliability = "Reliability";
             public const string Usage = "Usage";
+            public const string WebJobs = "WebJobs";    // TODO: Too vague?
```
liliankasem: What's the usage of this and what are you trying to represent? Context might help me answer that "todo" question :D

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
```diff
@@ -0,0 +1,141 @@
...
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class WebJobsAttributeAnalyzer : DiagnosticAnalyzer
+    {
+        // TODO: Scope this to per-project
+        JobHostMetadataProvider _tooling;
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get { return ImmutableArray.Create(DiagnosticDescriptors.IllegalFunctionName); } }
+
+    public static void VerifyWebJobsLoaded()
```
liliankasem:
```suggestion
        public static void VerifyWebJobsLoaded() 
```

nit: misaligned

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
```diff
@@ -0,0 +1,141 @@
...
+    public class WebJobsAttributeAnalyzer : DiagnosticAnalyzer
+    {
+        // TODO: Scope this to per-project
+        JobHostMetadataProvider _tooling;
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get { return ImmutableArray.Create(DiagnosticDescriptors.IllegalFunctionName); } }
+
+    public static void VerifyWebJobsLoaded() 
+        {
+            var x = new JobHost(new OptionsWrapper<JobHostOptions>(new JobHostOptions()), null);
```
liliankasem: nit?
```suggestion
            var jobHost = new JobHost(new OptionsWrapper<JobHostOptions>(new JobHostOptions()), null);
```

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
```diff
@@ -0,0 +1,141 @@
...
+
+        // First argument to the FunctionName ctor.
+        private string GetFunctionNameFromAttribute(SemanticModel semantics, AttributeSyntax attributeSyntax)
+        {
+            foreach (var arg in attributeSyntax.ArgumentList.Arguments)
+            {
+                var val = semantics.GetConstantValue(arg.Expression);
+                if (!val.HasValue)
+                {
+                return null;
```
liliankasem, line 42: I think you can eliminate this return if you inverse the if to check for a value then return `return val.Value as string;` within the if condition, unless you're trying to not loop through everything and want to return early?

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
```diff
@@ -0,0 +1,141 @@
...
+            return null;
+        }
+
+        // Does the method have a [FunctionName] attribute?
+        // This provides a quick check before we get into the more intensive analysis work.
+        private bool HasFunctionNameAttribute(SyntaxNodeAnalysisContext context, MethodDeclarationSyntax methodDeclarationSyntax)
+        {
+            foreach (var attrListSyntax in methodDeclarationSyntax.AttributeLists)
+            {
+                foreach (AttributeSyntax attributeSyntax in attrListSyntax.Attributes)
```
liliankasem, line 57: The double foreach scares be a bit perf wise, but I'm not sure if there is a better way without a data structure change? @brettsam ?

liliankasem (commented):
I don't have a lot of context around assemblies so probably not the right person to sign off on this, but this looks great! Nice work, and I think you did a great job commenting your code which I appreciate a lot. Overall, just a few nits and questions :)

src/WebJobs.Script.Analyzers/AssemblyCache.cs
kashimiz: Thanks for all the comments! This is a pretty straight port of code @MikeStall wrote in [azure/azure-webjobs-sdk ](https://github.com/azure/azure-webjobs-sdk/), so compliments go to him. :) (I added the tests.)

Good catch here; I'll move this into Constants.

src/WebJobs.Script.Analyzers/AssemblyCache.cs
kashimiz: I think it means "I'm testing to see if this assembly has a `WebJobsStartupAttribute`" but I agree the name is a little indirect. I'll change it to `startupAttr`.

src/WebJobs.Script.Analyzers/AssemblyCache.cs
kashimiz: Agreed here; changing.

src/WebJobs.Script.Analyzers/AssemblyCache.cs
kashimiz: I think it stands for "assembly2". There's not really an assembly1, so just changing this to `assembly` and changing other `asm[#]` instances.

src/WebJobs.Script.Analyzers/AssemblyCache.cs
kashimiz: Changed.

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
```diff
@@ -0,0 +1,141 @@
...
+
+        private void AnalyzeCompilation(CompilationStartAnalysisContext context)
+        {
+            var compilation = context.Compilation;
+
+            AssemblyCache.Instance.Build(compilation);
+            _tooling = AssemblyCache.Instance.Tooling;
+
+            // cast to PortableExecutableReference which has a file path
+            var x1 = compilation.References.OfType<PortableExecutableReference>().ToArray();
```
madelinegordon: nit: Could "x1" be made more descriptive?

src/WebJobs.Script.Analyzers/CompilationWebJobsStartuptypeLocator.cs
kashimiz: Added!

test/WebJobs.Script.Tests.Analyzers/WebJobsAttributeAnalyzerTests.cs
```diff
@@ -0,0 +1,88 @@
...
+using System.Collections.Immutable;
+
+namespace WebJobs.Script.Tests.Analyzers
+{
+    public class WebJobsAttributeAnalyzerTests
+    {
+        [Fact]
+        public async Task ValidFunctionName_NoDiagnostic()
+        {
+            string testCode = @"
```
madelinegordon, line 18: nit: Could this be extracted to a separate method like GetTestCode(string functionName), to make the test more readable?

madelinegordon (approved):

src/WebJobs.Script.Analyzers/CompilationWebJobsStartuptypeLocator.cs
kashimiz: Added!

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
kshyju: I think we are trying to check only the first argument. IMHO, explicitly reading the first argument might be a bit more readable. Something like this.
````
if (attributeSyntax.ArgumentList.Arguments.Count == 0)
{
    return null; 
}

var firstArgument = attributeSyntax.ArgumentList.Arguments[0];
var val = semantics.GetConstantValue(firstArgument.Expression);

return val.Value as string;
````

File: src/WebJobs.Script.Analyzers/Constants.cs (hunk 9 above)
Comment (kashimiz) [reply src/WebJobs.Script.Analyzers/Constants.cs]: This is a fun discussion opportunity. From what I've gleaned, we can define categories for our analyzers. Here's an example from Roslyn: https://github.com/dotnet/roslyn-analyzers/blob/main/src/Utilities/Compiler/DiagnosticCategory.cs

You can use categories to configure settings, ex. severity, suppression, etc: https://docs.microsoft.com/en-us/visualstudio/code-quality/use-roslyn-analyzers?view=vs-2022#set-rule-severity-of-multiple-analyzer-rules-at-once-in-an-editorconfig-file

Since we only have a few analyzers right now, I'm not sure how we want to divide them + future ones up. Since these analyzers operate on the WebJobs binding model, so not fully "Functions", I went with `WebJobs`, but also considered `Binding` as a possible name. Love to hear your thoughts here! + @fabiocav

File: src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs (hunk 10 above)
Comment (kashimiz) [reply src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs]: Whoops! Thanks for the catch.

File: src/WebJobs.Script.Analyzers/AssemblyCache.cs [hunk 16]
```diff
@@ -0,0 +1,190 @@
...
+
+            // Produce tooling object
+            var webjobsStartups = new List<Type>();
+            foreach (var path in _map.Values)
+            {
+                // We don't want to load and reflect over every dll.
+                // By convention, restrict based on filenames.
+                var filename = Path.GetFileName(path);
+                // TODO: I assume this is so general to avoid missing custom extensions, but can it be tightened up for performance?
+                if (!filename.ToLowerInvariant().Contains("extension"))
```
soninaren: `DurableTask.SqlServer.AzureFunctions.dll` does not have extension in the name. We may have to update this to include durable assemblies

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
kashimiz: Better name. Thanks!

src/WebJobs.Script.Analyzers/AssemblyCache.cs
soninaren: Not sure if it is possible to read assemblyNames from extensions.json

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
```diff
@@ -0,0 +1,141 @@
...
+        {
+            if (_tooling == null) // Not yet initialized
+            {
+                return;
+            }
+
+            var methodDecl = (MethodDeclarationSyntax)context.Node;
+            var methodName = methodDecl.Identifier.ValueText;
+
+            if (!HasFunctionNameAttribute(context, methodDecl))
```
kshyju: Do we need the if condition and return block? Calling `HasFunctionNameAttribute` is good enough correct? I also think we should rename that method (HasFunctionNameAttribute) to be more explicit that it checks function name and report.

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
```diff
@@ -0,0 +1,141 @@
...
+        // Analyze the method signature to validate binding attributes + types on the parameters
+        private void AnalyzeMethodDeclarationNode(SyntaxNodeAnalysisContext context)
+        {
+            if (_tooling == null) // Not yet initialized
+            {
+                return;
+            }
+
+            var methodDecl = (MethodDeclarationSyntax)context.Node;
+            var methodName = methodDecl.Identifier.ValueText;
```
kshyju: nit: remove unused.

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
kashimiz: Very much so. Changed to `references`

test/WebJobs.Script.Tests.Analyzers/WebJobsAttributeAnalyzerTests.cs
kashimiz: Yes; this approach worked when there were only a few tests, but now it's getting unwieldy. There are some other possible approaches, such as loading the code from .cs files in the test project. I'd like to flag this as a near-future enhancement (before things get too out of hand) and make those changes in a separate PR.

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
kashimiz: Removed.

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
kshyju: The inner foreach won't be executed for any method declarations which does not have attribute decorations.

src/WebJobs.Script.Analyzers/AssemblyCache.cs
```diff
@@ -0,0 +1,190 @@
...
+
+        // Assembly Display Name --> loaded Assembly object
+        Dictionary<string, Assembly> _mapRef = new Dictionary<string, Assembly>();
+
+        const string WebJobsAssemblyName = "Microsoft.Azure.WebJobs";
+        const string WebJobsHostAssemblyName = "Microsoft.Azure.WebJobs.Host";
+
+        JobHostMetadataProvider _tooling;
+
+        internal JobHostMetadataProvider Tooling => _tooling;
```
soninaren: Not sure why this is named tooling, Should we call it `metadataProvider or jobHostMetadataProvider `

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
```diff
@@ -0,0 +1,141 @@
...
+
+        // Does the method have a [FunctionName] attribute?
+        // This provides a quick check before we get into the more intensive analysis work.
+        private bool HasFunctionNameAttribute(SyntaxNodeAnalysisContext context, MethodDeclarationSyntax methodDeclarationSyntax)
+        {
+            foreach (var attrListSyntax in methodDeclarationSyntax.AttributeLists)
+            {
+                foreach (AttributeSyntax attributeSyntax in attrListSyntax.Attributes)
+                {
+                    // Perf - Can we get the name without doing a symbol resolution?
```
kshyju, line 59: Would `GetText` method give us this info(name) ?

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
```diff
@@ -0,0 +1,141 @@
...
+                return null;
+                }
+                return val.Value as string;
+            }
+            return null;
+        }
+
+        // Does the method have a [FunctionName] attribute?
+        // This provides a quick check before we get into the more intensive analysis work.
+        private bool HasFunctionNameAttribute(SyntaxNodeAnalysisContext context, MethodDeclarationSyntax methodDeclarationSyntax)
```
kshyju: IMHO, `Validate/CheckMethodsAndReport` or something along that line seems more readable (or something where we say we are reporting as well).

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
```diff
@@ -0,0 +1,141 @@
...
+            {
+                foreach (AttributeSyntax attributeSyntax in attrListSyntax.Attributes)
+                {
+                    // Perf - Can we get the name without doing a symbol resolution?
+                    var symAttributeCtor = context.SemanticModel.GetSymbolInfo(attributeSyntax).Symbol;
+                    if (symAttributeCtor != null)
+                    {
+                        var attrType = symAttributeCtor.ContainingType;
+                        if (attrType.Name == nameof(FunctionNameAttribute))
+                        {
```
kshyju, line 65: Also, consider inverting `if` blocks to reduce nesting. Helps with readability.

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
```diff
@@ -0,0 +1,141 @@
...
+                                var error = Diagnostic.Create(DiagnosticDescriptors.IllegalFunctionName, attributeSyntax.GetLocation(), functionName);
+                                context.ReportDiagnostic(error);
+                            }
+
+                            return true;
+                        }
+                    }
+                }
+            }
+            return false;
```
kshyju: Remove return value as those seems to be unused on the caller site.

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
```diff
@@ -0,0 +1,141 @@
...
+                {
+                return null;
+                }
+                return val.Value as string;
+            }
+            return null;
+        }
+
+        // Does the method have a [FunctionName] attribute?
+        // This provides a quick check before we get into the more intensive analysis work.
```
kshyju: nit: Are these comments reflecting the latest version of the method?

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
```diff
@@ -0,0 +1,141 @@
...
+
+            var methodDecl = (MethodDeclarationSyntax)context.Node;
+            var methodName = methodDecl.Identifier.ValueText;
+
+            if (!HasFunctionNameAttribute(context, methodDecl))
+            {
+                return;
+            }
+
+            // TODO: Attribute validation
```
kshyju: nit: Cleanup if these are not needed anymore.

src/WebJobs.Script.Analyzers/AssemblyCache.cs
kashimiz: `jobHostMetadataProvider` it is.

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
```diff
@@ -0,0 +1,141 @@
...
+            // Hook compilation to get the assemblies' references and build the WebJob tooling interfaces.
+            context.RegisterCompilationStartAction(AnalyzeCompilation);
+        }
+
+        private void AnalyzeCompilation(CompilationStartAnalysisContext context)
+        {
+            var compilation = context.Compilation;
+
+            AssemblyCache.Instance.Build(compilation);
+            _tooling = AssemblyCache.Instance.Tooling;
```
kshyju: Could we document what are the intentions of this method/why we are doing what we are doing here, to help future readers ? Also, for a present reader😁 I am lost here 😕

src/WebJobs.Script.Analyzers/AssemblyCache.cs
```diff
@@ -0,0 +1,190 @@
...
+using Microsoft.Azure.WebJobs;
+using Microsoft.Azure.WebJobs.Host;
+using Microsoft.Azure.WebJobs.Hosting;
+using Microsoft.CodeAnalysis;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+namespace Microsoft.Azure.Functions.Analyzers
+{
+    // Only support extensions and WebJobs core.
```
kshyju: Could we add some summary comment to this type explaining what this class does?

src/WebJobs.Script.Analyzers/AssemblyCache.cs
kashimiz: This logic is looking for `IWebJobsStartup` implementations to pull the relevant binding rules and validation attributes. The Durable Functions bindings are defined in `Microsoft.Azure.WebJobs.Extensions.DurableTask`, so I think skipping `DurableTask.SqlServer.AzureFunctions` would be fine.

If there are binding attributes or `IWebJobsStartup` implementations in an assembly that doesn't follow the `Microsoft.Azure.WebJobs.Extensions.<name>` convention, we'd need to adjust this.

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
kashimiz: `SyntaxNode.GetText` would give a string like `FunctionName("MyFunction")` that the function name could be extracted from. This approach would lose the symbol information, but we're only using the `NamedType.Name` string for comparison. It might be a bit fragile, though.

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
kashimiz: Given that this isn't a simple boolean check and has some side effects, I think that's a good idea.

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
kashimiz: That was addressed in this PR, so I'll take that comment out: https://github.com/Azure/azure-functions-host/pull/8400/

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
kshyju: Got it. Thanks!

src/WebJobs.Script.Analyzers/AssemblyCache.cs
soninaren: You are right, I saw the startup class implementation `DurableTask.SqlServer.AzureFunctions.dll` i assumed it to be implementing IWebJobsStartup , it is implementing 
 FunctionsStartup instead.

src/WebJobs.Script.Analyzers/Constants.cs
liliankasem: Ahh I see, would `WebJobsBindings` be a good category name?

src/WebJobs.Script.Analyzers/CompilationWebJobsStartuptypeLocator.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            _assemblies = assemblies;
+        }
+
+        public Type[] GetStartupTypes()
+        {
+            return _startupTypes.Value;
+        }
+
+        public Type[] GetTypes()
```
kshyju: Does this need to be public?

src/WebJobs.Script.Analyzers/DiagnosticDescriptors.cs
```diff
@@ -27,5 +27,11 @@ private static DiagnosticDescriptor Create(string id, string title, string messa
                 messageFormat: "Reuse HttpClient instances to avoid holding more connections than necessary. See helplink for more information.",
                 category: Constants.DiagnosticsCategories.Reliability,
                 severity: DiagnosticSeverity.Warning);
+
+        public static DiagnosticDescriptor IllegalFunctionName { get; }
+            = Create(id: "AZF0003", title: "Illegal Function name",
```
kshyju, line 29: Which repo needs to be updated to make the https://docs.microsoft.com/azure/azure-functions/errors-diagnostics/sdk-rules/AZF0003 link working?

src/WebJobs.Script.Analyzers/WebJobs.Script.Analyzers.csproj
```diff
@@ -41,4 +45,7 @@
     </ItemGroup>
   </Target>
 
+  <Target Name="_AddNuGetDlls">
```
kshyju: Did you mean to have this empty target?

src/WebJobs.Script.Analyzers/Constants.cs
kashimiz: That'll do!

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
kashimiz: Shyju is correct; the loop will hit a return statement on its first iteration. I'm not sure why it was written the way it was, but his suggested implementation is more straightforward. I'll make the substitution.

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
kashimiz: Good catch. I'm guessing this was implemented with a return value for flexibility, but ended up only being called in one place.

src/WebJobs.Script.Analyzers/CompilationWebJobsStartuptypeLocator.cs
kashimiz: No, no it does not.

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
kashimiz: Removed `AssemblyCache` for now after conversation offline; will reintroduce in the PR for the analyzer that actually makes use of it.

src/WebJobs.Script.Analyzers/AssemblyCache.cs
kashimiz: Removed for now, but when it comes back it'll be with comments. :)

src/WebJobs.Script.Analyzers/DiagnosticDescriptors.cs
kashimiz: There's an additional PR in [MicrosoftDocs/azure-docs](https://github.com/MicrosoftDocs/azure-docs) that I'll be linking.

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
kashimiz: To the best of my knowledge, it did. I'm updating it to reflect the changes I'm pushing.

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
kashimiz: Addressed in later comment.

src/WebJobs.Script.Analyzers/WebJobs.Script.Analyzers.csproj
kashimiz: Ah, thanks. This is a remnant; found a cleaner way to include the WebJobs DLLs in the analyzer nuget. Removing.

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
```diff
@@ -0,0 +1,126 @@
...
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class WebJobsAttributeAnalyzer : DiagnosticAnalyzer
+    {
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get { return ImmutableArray.Create(DiagnosticDescriptors.IllegalFunctionName); } }
+
+        public static void VerifyWebJobsLoaded()
+        {
+            // Check if WebJobs types can be loaded
+            var jobHost = new JobHost(new OptionsWrapper<JobHostOptions>(new JobHostOptions()), null);
```
liliankasem: If we're not going to use the variable
```suggestion
            _ = new JobHost(new OptionsWrapper<JobHostOptions>(new JobHostOptions()), null);
```

src/WebJobs.Script.Analyzers/DiagnosticDescriptors.cs
```diff
@@ -27,5 +27,11 @@ private static DiagnosticDescriptor Create(string id, string title, string messa
...
+                messageFormat: "Function name can't be '{0}'",
```
brettsam: Is there any reasonable way for us to explain the regex to provide clues? Like "cannot be longer than 32 characters and must be alphanumeric?"

liliankasem (approved):
lgtm! just one last comment from me!

test/WebJobs.Script.Tests.Analyzers/AssemblyCacheTests.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
```
brettsam: delete this file for now since it's empty?

test/WebJobs.Script.Tests.Analyzers/WebJobs.Script.Tests.Analyzers.csproj
```diff
@@ -7,6 +7,8 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.33" />
+    <PackageReference Include="Microsoft.Azure.WebJobs.Host.Storage" Version="4.0.4" />
```
brettsam: do we need this storage ref here?

brettsam (commented):
Couple minor comments...

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
kshyju: I do not think this `VerifyWebJobsLoaded` method is needed in this analyzer PR anymore. Based on our last sync, we decided to keep only the code needed for the analyzer for function name validation. Feel free to correct me @kashimiz

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
```diff
@@ -0,0 +1,126 @@
...
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+            context.EnableConcurrentExecution();
+
+            VerifyWebJobsLoaded();
+
+            // Analyze method signatures.
+            context.RegisterSyntaxNodeAction(AnalyzeMethodDeclarationNode, SyntaxKind.MethodDeclaration);
+
+            // Hook compilation to get the assemblies' references and build the WebJob tooling interfaces.
+            context.RegisterCompilationStartAction(AnalyzeCompilation);
```
kshyju: Do not think we need this one either in this analyzer PR as this is not really needed for function name validation analyzer.

src/WebJobs.Script.Analyzers/CompilationWebJobsStartuptypeLocator.cs
```diff
@@ -0,0 +1,44 @@
...
+namespace Microsoft.Azure.Functions.Analyzers
+{
+    /// <summary>
+    /// An implementation of an <see cref="IWebJobsStartupTypeLocator"/> that "locates" startup types
+    /// from a Microsoft.CodeAnalysis.Compilation's dependencies.
+    /// 
+    /// Note: Assemblies are derived from the Compilation by AssemblyCache. 
+    /// This class pulls the WebJobsStartupAttributes from assemblies in the list.
+    /// </summary>
+    public class CompilationWebJobsStartupTypeLocator : IWebJobsStartupTypeLocator
```
kshyju: Can delete this file as we do not need this file for the function name analyzer (current PR we want to merge). Add in the next PR where this will be actually used, as needed.

src/WebJobs.Script.Analyzers/Constants.cs
```diff
@@ -5,15 +5,22 @@ namespace Microsoft.Azure.Functions.Analyzers
 {
     internal static class Constants
     {
+        internal static class Assemblies
```
kshyju: nit: Can name the class as `AssemblyNames` and const as `WebJobs`. So it will be used as `AssemblyNames.WebJobs`

src/WebJobs.Script.Analyzers/Constants.cs
```diff
@@ -5,15 +5,22 @@ namespace Microsoft.Azure.Functions.Analyzers
...
         internal static class Types
         {
             public const string FunctionNameAttribute = "Microsoft.Azure.WebJobs.FunctionNameAttribute";
         }
 
         internal static class DiagnosticsCategories
         {
-            public const string Reliability = "Reliability";
-            public const string Usage = "Usage";
+            public const string Reliability = nameof(Reliability);
```
kshyju: nit question: it's cool to know we can use nameof here. But If someone decides to do a rename on the caller site (by doing Ctrl + R +R ) on the constant name, will it also update the value?

kashimiz: It would. "Which shouldn't be a problem because the tests would break," I thought, then looked at the tests, and it turns out they don't check expected vs actual `Category`. (There's not a way to do so without adding a custom `DiagnosticVerifier` either.)

tl;dr It's cool but you make a good point. Reverting to strings.

src/WebJobs.Script.Analyzers/Constants.cs
kashimiz: This is less redundant; I like it.

src/WebJobs.Script.Analyzers/CompilationWebJobsStartuptypeLocator.cs
kashimiz: Missed this; removing.

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
kashimiz: True; this analyzer doesn't use the tooling object. Thanks for the catch.

test/WebJobs.Script.Tests.Analyzers/WebJobs.Script.Tests.Analyzers.csproj
kashimiz: Not for this analyzer, no. Thanks for the catch.

test/WebJobs.Script.Tests.Analyzers/AssemblyCacheTests.cs
kashimiz: Deleting.

src/WebJobs.Script.Analyzers/DiagnosticDescriptors.cs
kashimiz: How about: `Function name can't be '{0}'. It must start with a letter, be no longer than 128 characters, and contain only alphanumeric characters, underscores, or hyphens.`

128-character limit: https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs/FunctionNameAttribute.cs#L35

What logic imposes a 32-character limit at runtime? AFAIK we do this for host IDs, but I wasn't aware we imposed the same limit for function names.

src/WebJobs.Script.Analyzers/AssemblyCache.cs
kashimiz: I have removed the entire file. Will make these updates when I put it back.

src/WebJobs.Script.Analyzers/AssemblyCache.cs
kashimiz: I have removed the entire file. Will make these updates when I put it back.

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
kashimiz: True; removed.

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
```diff
@@ -0,0 +1,86 @@
...
+                        var attrType = symAttributeCtor.ContainingType;
+                        if (attrType.Name != nameof(FunctionNameAttribute))
+                        {
+                            return;
+                        }
+
+                        // Validate the FunctionName
+                        var functionName = GetFunctionNameFromAttribute(context.SemanticModel, attributeSyntax);
+
+                        bool match = FunctionNameAttribute.FunctionNameValidationRegex.IsMatch(functionName);
```
kshyju: The `IsMatch` method can throw when the input is `null`. `GetFunctionNameFromAttribute` have some code path to return `null`. So we should do a null check before calling the Match method.

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
```diff
@@ -0,0 +1,86 @@
...
+        // Quick check for [FunctionName] attribute on a method.
+        // Reports a diagnostic if the name doesn't meet requirements.
+        private void CheckForFunctionNameAttributeAndReport(SyntaxNodeAnalysisContext context, MethodDeclarationSyntax methodDeclarationSyntax)
+        {
+            foreach (var attrListSyntax in methodDeclarationSyntax.AttributeLists)
+            {
+                foreach (AttributeSyntax attributeSyntax in attrListSyntax.Attributes)
+                {
+                    // Perf - Can we get the name without doing a symbol resolution?
+                    var test = attributeSyntax.GetText();
```
kshyju: Remove this line. The variable `test` seems to be unused.

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
```diff
@@ -0,0 +1,86 @@
...
+namespace Microsoft.Azure.Functions.Analyzers
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class WebJobsAttributeAnalyzer : DiagnosticAnalyzer
+    {
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get { return ImmutableArray.Create(DiagnosticDescriptors.IllegalFunctionName); } }
+
+        public override void Initialize(AnalysisContext context)
+        {
+            // https://stackoverflow.com/questions/62638455/analyzer-with-code-fix-project-template-is-broken
```
kshyju, line 20: nit: Do we need to keep this comment?

src/WebJobs.Script.Analyzers/WebJobs.Script.Analyzers.csproj
```diff
@@ -25,6 +25,9 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="2.9.8" />
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="3.3.1" />
+	  
+	<!-- Packed in Microsoft.Azure.Functions.Analyzers NuGet -->
```
kshyju, line 29: Do we really need this package reference? I wonder we should keep a copy of the Regex in the analyzer so that the analyzer won't have a dependency on WebJobs package. (I don't have a strong opinion though).

test/WebJobs.Script.Tests.Analyzers/WebJobsAttributeAnalyzerTests.cs
```diff
@@ -0,0 +1,88 @@
...
+        {
+            httpClient.GetAsync(""https://www.microsoft.com"");
+        }
+    }
+}
+";
+
+            var test = new AnalyzerTest();
+            test.ReferenceAssemblies = ReferenceAssemblies.Net.Net50.WithPackages(ImmutableArray.Create(
+                new PackageIdentity("Microsoft.NET.Sdk.Functions", "3.0.11"),
```
kshyju: 3.1.1 and 4.1.1 are the latest versions of `Microsoft.NET.Sdk.Functions` package.

kshyju (approved):
LGTM.

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
kashimiz: I would prefer to; it might save someone time/frustration in the future. (When I picked up analyzer work again after a break, I forgot to add the line below and lost a chunk of time figuring out why nothing was working. I have every reason to believe this will happen again. Probably to me.)

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
kashimiz: Good catch; thank you. Adding a null check.

src/WebJobs.Script.Analyzers/WebJobs.Script.Analyzers.csproj
kashimiz: The regex is unlikely to change, but I'd rather not give it the opportunity to get out of sync with what's in the WebJobs SDK. We'll need this reference later for the analyzer that runs through the binding rules (https://github.com/Azure/azure-functions-host/issues/8255) in any case.

src/WebJobs.Script.Analyzers/WebJobsAttributeAnalyzer.cs
kashimiz: Removed.

test/WebJobs.Script.Tests.Analyzers/WebJobsAttributeAnalyzerTests.cs
kashimiz: The other analyzers' tests need this change as well. I'll update this reference, but keep it as a separate commit when merging this PR.

src/WebJobs.Script.Analyzers/DiagnosticDescriptors.cs
brettsam: That sounds good to me!

The 32-character limit is for host ids b/c we use that to create queues for some internal stuff, which have stricter limits on length. That logic is here -- https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Host/ScriptHostIdProvider.cs#L87

brettsam (approved):

PR #8276, opened 2022-03-31
Ensure _maxProcessCount is always initialized

liliankasem:
### Pull request checklist

### DIFF
--- a/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
+++ b/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
@@ -32,6 +32,7 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
         private readonly TimeSpan _thresholdBetweenRestarts = TimeSpan.FromMinutes(WorkerConstants.WorkerRestartErrorIntervalThresholdInMinutes);
         private readonly IOptions<WorkerConcurrencyOptions> _workerConcurrencyOptions;
         private readonly IEnumerable<RpcWorkerConfig> _workerConfigs;
+        private readonly Lazy<Task<int>> _maxProcessCount;
 
         private IScriptEventManager _eventManager;
         private IWebHostRpcWorkerChannelManager _webHostLanguageWorkerChannelManager;
@@ -39,7 +40,6 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
         private IDisposable _workerErrorSubscription;
         private IDisposable _workerRestartSubscription;
         private ScriptJobHostOptions _scriptOptions;
-        private int _maxProcessCount;
         private IRpcFunctionInvocationDispatcherLoadBalancer _functionDispatcherLoadBalancer;
         private bool _disposed = false;
         private bool _disposing = false;
@@ -91,6 +91,8 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
 
             _shutdownStandbyWorkerChannels = ShutdownWebhostLanguageWorkerChannels;
             _shutdownStandbyWorkerChannels = _shutdownStandbyWorkerChannels.Debounce(milliseconds: 5000);
+
+            _maxProcessCount = new Lazy<Task<int>>(GetMaxProcessCount);
         }
 
         public FunctionInvocationDispatcherState State { get; private set; }
@@ -101,9 +103,25 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
 
         internal ConcurrentStack<WorkerErrorEvent> LanguageWorkerErrors => _languageWorkerErrors;
 
-        internal int MaxProcessCount => _maxProcessCount;
-
         internal IWebHostRpcWorkerChannelManager WebHostLanguageWorkerChannelManager => _webHostLanguageWorkerChannelManager;
+
+        internal Task<int> MaxProcessCount => _maxProcessCount.Value;
+
+        private async Task<int> GetMaxProcessCount()
+        {
+            if (_maxProcessCount.IsValueCreated)
+            {
+                return await _maxProcessCount.Value;
+            }
+
+            if (_environment.IsPlaceholderModeEnabled())
+            {
+                return (await GetAllWorkerChannelsAsync()).ToArray().Count();
+            }
+
+            var workerConfig = _workerConfigs.Where(c => c.Description.Language.Equals(_workerRuntime, StringComparison.InvariantCultureIgnoreCase)).FirstOrDefault();
+            return _environment.IsWorkerDynamicConcurrencyEnabled() ? _workerConcurrencyOptions.Value.MaxWorkerCount : workerConfig.CountOptions.ProcessCount;
+        }
 
         internal async Task InitializeJobhostLanguageWorkerChannelAsync()
         {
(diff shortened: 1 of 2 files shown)


src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -214,6 +214,7 @@ public async Task InitializeAsync(IEnumerable<FunctionMetadata> functions, Cance
 
             if (placeholderModeEnabled)
             {
+                _maxProcessCount = _webHostLanguageWorkerChannelManager.GetChannels(_workerRuntime).Count();
```
fabiocav: I realize this avoids the uninitialized flow, but do we want to do something like we've discussed and put this behind a property to ensure initialization in all cases? 

Also, do we have a race between starting the channels at the web host level and the point where this count is performed?

liliankasem: Yup, I'll make this a property.

Good point about a potential race here, any suggestions?

pragnagopa: Yes, race conditions are possible here - we have seen some variation of races here earlier as well - see

https://github.com/Azure/azure-functions-host/blob/2597f221d7643b24e206712ab9d8c3a01972b50d/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs#L405-L421

Instead of relying on the count - refactor the code above -get the count from webhostchannels - this awaits on the Tasks that start in webhostchannelmanager

test/WebJobs.Script.Tests/Workers/Rpc/RpcFunctionInvocationDispatcherTests.cs
```diff
@@ -279,6 +279,29 @@ public async Task ShutdownTests_WithInfinitelyRunningTasks_Timesout()
...
+            int expectedJobHostProcessCount = 0;
+
+            var functionDispatcher = GetTestFunctionDispatcher(runtime: RpcWorkerConstants.JavaLanguageWorkerName, placeholder: true, addWebhostChannel: true);
+            await WaitForWebhostWorkerChannelsToStartup(functionDispatcher.WebHostLanguageWorkerChannelManager, expectedWebHostProcessCount, RpcWorkerConstants.JavaLanguageWorkerName);
+            var channels = functionDispatcher.WebHostLanguageWorkerChannelManager.GetChannels(RpcWorkerConstants.JavaLanguageWorkerName).Values;
+            channels.FirstOrDefault().Task.Result.SetupFunctionInvocationBuffers(null);
+
+            // Act
+            await functionDispatcher.InitializeAsync(GetTestFunctionsList(RpcWorkerConstants.JavaLanguageWorkerName));
+            await functionDispatcher.ShutdownAsync();
```
pragnagopa, line 296: add check for MaxProcessCount

pragnagopa (commented):
added minor comment for the test. Rest looks good.!

test/WebJobs.Script.Tests/Workers/Rpc/RpcFunctionInvocationDispatcherTests.cs
liliankasem: done!

liliankasem: I ended up refactoring this so that we use lazy initialization, and since maxProcessCount is never used outside of this class I made that field private to avoid something else triggering the initialization. So I think we are okay to not check the maxProcessCount field directly here - thoughts?

liliankasem:
@alrod adding you for review as there was some overlap in the code changes we made and want to make sure I'm not breaking your stuff :)

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -103,10 +103,25 @@ internal class RpcFunctionInvocationDispatcher : IFunctionInvocationDispatcher
...
 
-        internal int MaxProcessCount => _maxProcessCount;
-
         internal IWebHostRpcWorkerChannelManager WebHostLanguageWorkerChannelManager => _webHostLanguageWorkerChannelManager;
 
+        private async Task<int> GetMaxProcessCount()
+        {
+            if (_workerConfigs != null
+                && _environment != null
+                && _workerConcurrencyOptions != null
```
pragnagopa: move null ref checks on _evironment and _workerconfigs to constructor - if these are null - DI setup is has gone wrong!

test/WebJobs.Script.Tests/Workers/Rpc/RpcFunctionInvocationDispatcherTests.cs
pragnagopa: exposing internal property to verify in unit tests is OK!

pragnagopa (approved):
approving with minor comments. Please wait for approval from @fabiocav  as well. Thanks!

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
liliankasem: great point, will move!

test/WebJobs.Script.Tests/Workers/Rpc/RpcFunctionInvocationDispatcherTests.cs
liliankasem: Done, but just FYI I think we were cautious of making it internal so that it didn't get invoked outside of this class i.e giving something else the ability to call `(await _maxProcessCount.Value)` to init the value before we actually need it. technically shouldn't happen as this is not used outside of this class but more of a safety thing

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -103,10 +103,22 @@ internal class RpcFunctionInvocationDispatcher : IFunctionInvocationDispatcher
...
+            if (_workerConcurrencyOptions != null && !string.IsNullOrEmpty(_workerRuntime))
+            {
+                var workerConfig = _workerConfigs.Where(c => c.Description.Language.Equals(_workerRuntime, StringComparison.InvariantCultureIgnoreCase)).FirstOrDefault();
+                if (workerConfig != null)
+                {
+                    return _environment.IsWorkerDynamicConcurrencyEnabled() ? _workerConcurrencyOptions.Value.MaxWorkerCount : workerConfig.CountOptions.ProcessCount;
+                }
+            }
+
+            return (await GetAllWorkerChannelsAsync()).ToArray().Count();
```
fabiocav: We should be able to skip `ToArray` here

fabiocav (approved):
Small comment. Once that is addressed, :shipit:

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
liliankasem: Yup, you're right - I don't know why I had to do that in the first place xD

pragnagopa:
Can you port this to v3.x as well? Thanks!

liliankasem:
> Can you port this to v3.x as well? Thanks!

Can do!

liliankasem:
Alrighty, manual it is :)

PR #8317, opened 2022-04-22
Exception handling for IHostedService implementations in start/stop methods

### LINKED ISSUE #4891: Revisit: IHostedServices can now throw exceptions on Start/Stop in 3.0
This behavior changed in .NET Core 3.0: https://github.com/aspnet/AspNetCore/pull/7657.

We had one test (`StandbyModeE2E_LinuxContainer`) that was failing at startup because the python language worker wasn't present. Our test in 2.0 passed happily because the host would swallow this exception. Now it failed. I've wrapped the startup in a try/catch, but we need to re-evaluate how we want to handle this if StartAsync() fails in our hosted services.

surgupta-msft:
resolves #4891 

The PR adds try-catch block in start/stop methods of all the implementations of IHostedService and IManagedHostedService.

### Pull request checklist

### Additional information

Additional PR information


src/WebJobs.Script.WebHost/ContainerManagement/LinuxContainerActivityPublisher.cs
```diff
@@ -75,27 +75,41 @@ private void OnStandbyOptionsChange()
...
+            try
+            {
+                _logger.LogInformation($"Stopping {nameof(LinuxContainerActivityPublisher)}");
 
-            // stop the timer if it has been started
-            _timer?.Change(Timeout.Infinite, Timeout.Infinite);
+                // stop the timer if it has been started
+                _timer?.Change(Timeout.Infinite, Timeout.Infinite);
+            }
+            catch (Exception ex)
```
kshyju, line 109: When the cancellation token is cancelled (_as a result of the infra code is signaling that the app is shutting down_), Would it throw `TaskCanceledException` or `OperationCanceledException` ?. Should we exclude that case before logging it as an error? 

Something like
````
catch (Exception ex)
{
    if (!cancellationToken.IsCancellationRequested)
    {
      //log error
    }
}
````

kshyju, line 109: When the cancellation token is cancelled (_as a result of the infra code is signaling that the app is shutting down_), it is going to throw `TaskCanceledException` or `OperationCanceledException`. Should we exclude that case before logging it as an error? 

Something like
````
catch (Exception ex)
{
    if (!cancellationToken.IsCancellationRequested)
    {
      //log error
    }
}
````

brettsam (commented):
Looking at this -- I think that we only want to try/catch the startup of services that are optional. In a lot of cases, the host cannot really run without some of these running and it should fail to startup if an exception is thrown.

In fact, the only one that I think should have any try/catch around its startup is the `AssemblyAnalysisService`.  I think we should backout the changes from all the others.

liliankasem:
> Looking at this -- I think that we only want to try/catch the startup of services that are optional. In a lot of cases, the host cannot really run without some of these running and it should fail to startup if an exception is thrown.
> 
> In fact, the only one that I think should have any try/catch around its startup is the `AssemblyAnalysisService`. I think we should backout the changes from all the others.

In the other cases (outside of AssemblyAnalysisService), is it worth logging the error in a catch statement and then throwing? Or are the logs not valuable here?

brettsam:
> In the other cases (outside of AssemblyAnalysisService), is it worth logging the error in a catch statement and then throwing? Or are the logs not valuable here?

The logs will bubble up and prevent the Host from completing its startup. All of that will be logged automatically.

PR #8323, opened 2022-04-25
Implement unique semaphore for release and drain API calls

### LINKED ISSUE #8333: /resume call timesout as it's waiting for the /drain semaphore to release
Calls to `/resume` a draining host can timeout if host is draining invocations as the `/drain` controller is still holding onto a semaphore

##### Expected

1. Implement function app that reads from an event hub and sleeps for 15 minutes
1. Call /drain
1. Call /resume after 5 mins
1. Function host should resume

##### Actual

The /resume called timeouts because it's waiting on the semaphore used by /drain
to be released. This is because we release the semaphore in the drain controller
in a background thread that is waiting for the listeners to stop.

liliankasem:

resolves #8333

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/Controllers/HostController.cs
+++ b/src/WebJobs.Script.WebHost/Controllers/HostController.cs
@@ -111,28 +111,33 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Controllers
         [Authorize(Policy = PolicyNames.AdminAuthLevelOrInternal)]
         public async Task<IActionResult> Drain([FromServices] IScriptHostManager scriptHostManager)
         {
-            _logger.LogDebug("Received request for draining host");
-
-            if (!Utility.TryGetHostService(scriptHostManager, out IDrainModeManager drainModeManager))
+            try
             {
-                return StatusCode(StatusCodes.Status503ServiceUnavailable);
-            }
+                _logger.LogDebug("Received request to drain the host");
 
-            await _drainModeSemaphore.WaitAsync();
+                if (!Utility.TryGetHostService(scriptHostManager, out IDrainModeManager drainModeManager))
+                {
+                    return StatusCode(StatusCodes.Status503ServiceUnavailable);
+                }
 
-            // Stop call to some listeners gets stuck, not waiting for the stop call to complete
-            _ = drainModeManager.EnableDrainModeAsync(CancellationToken.None)
-                                .ContinueWith(
-                                    antecedent =>
-                                    {
-                                        if (antecedent.Status == TaskStatus.Faulted)
+                await _drainModeSemaphore.WaitAsync();
+
+                // Stop call to some listeners gets stuck, not waiting for the stop call to complete
+                _ = drainModeManager.EnableDrainModeAsync(CancellationToken.None)
+                                    .ContinueWith(
+                                        antecedent =>
                                         {
-                                            _logger.LogError(antecedent.Exception, "Something went wrong enabling drain mode");
-                                        }
-
-                                        _drainModeSemaphore.Release();
-                                    });
-            return Accepted();
+                                            if (antecedent.Status == TaskStatus.Faulted)
+                                            {
+                                                _logger.LogError(antecedent.Exception, "Something went wrong invoking drain mode");
+                                            }
+                                        });
+                return Accepted();
+            }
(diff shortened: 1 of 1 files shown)


src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -111,28 +111,33 @@ public async Task<IActionResult> GetHostStatus([FromServices] IScriptHostManager
...
-                                .ContinueWith(
-                                    antecedent =>
-                                    {
-                                        if (antecedent.Status == TaskStatus.Faulted)
-                                        {
-                                            _logger.LogError(antecedent.Exception, "Something went wrong enabling drain mode");
-                                        }
+                await _drainModeSemaphore.WaitAsync();
 
-                                        _drainModeSemaphore.Release();
```
sidkri, line 133: The original logic here is to prevent two or more concurrent executions of `EnableDrainModeAsync()` as some `IListeners` should not be stopped concurrently.  The listeners themselves have logic to serialize calls to `StopAsync()` (e.g. [ServiceBus Ext.](https://github.com/Azure/azure-functions-servicebus-extension/blob/4c64c9263a849b0ba0e5f502889150dffd94060d/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs#L127)) so having a semaphore here might be redundant.

The semaphore does not seem to serve any purpose with the new change as `EnableDrainModeAsync()` will return a task immediately and I suggest reverting this change and discussing the need for this semaphore.

The issue is that the same semaphore (`_drainModeSemaphore`) is used in both `Drain()` and [`Resume()`](https://github.com/Azure/azure-functions-host/blob/1db1782f1bda1718215ce7337e89ebde32515325/src/WebJobs.Script.WebHost/Controllers/HostController.cs#L177).  This will cause `Resume()` to block until draining completes.  We should understand why there was a need to share the same semaphore.  Thinking through the logic around drain and resume, I don't see the need for concurrency control between them.  In fact, we want to be able to resume while we are draining (i.e. [`ScriptHostState.Running`](https://github.com/Azure/azure-functions-host/blob/1db1782f1bda1718215ce7337e89ebde32515325/src/WebJobs.Script.WebHost/Controllers/HostController.cs#L183) and [`drainModeManager.IsDrainModeEnabled = true`](https://github.com/Azure/azure-functions-host/blob/1db1782f1bda1718215ce7337e89ebde32515325/src/WebJobs.Script.WebHost/Controllers/HostController.cs#L192) which I believe will only be the case while drain is in progress.  `Resume()` itself might not need a semaphore as `scriptHostManager.RestartHostAsync()` looks thread safe and will [cancel any in-progress startup](https://github.com/Azure/azure-functions-host/blob/86efe24eaf1d7d016975fe72c9d193670bebf66b/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs#L470) of the active host. but it ofcourse doesn't hurt to have a semaphore preventing concurrent execution.  The semaphore in `Resume()` should be distinct from that used by `Drain()` - I feel this is the fix we need.

sidkri (changes requested):

src/WebJobs.Script.WebHost/Controllers/HostController.cs
liliankasem: Thanks for the feedback, this makes a lot of sense. I remember Mathew & Fabio asking me to include a semaphore in the /resume case because we're checking state so I believe that needs to stay. I don't know why we need a semaphore in /drain - maybe @soninaren has context.

Sounds like the fix is to make the semaphores distinct as you suggested, or remove the drain one?

sidkri: Suggest making them distinct.  It doesn't hurt to have a semaphore in drain so we can leave it in there and ask Naren for guidance about removing it.

sidkri (approved):

sidkri:
Latest changes look good!  Could you link the issue and backport PR/issue here?

liliankasem:
> Latest changes look good! Could you link the issue and backport PR/issue here?

Added an issue for this bug! I should be able to automate a backport using the bot when this merges

soninaren (approved):

PR #8329, opened 2022-04-26
correctly wiring up OnChange for ScriptApplicationHostOptions

brettsam:


test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SpecializationE2ETests.cs
```diff
@@ -246,6 +243,51 @@ public async Task Specialization_ResetsSharedLoadContext()
             }
         }
 
+        [Fact]
+        public async Task Specialization_ResetsSecretManagerRepository()
```
liliankasem, line 247: Are there any edge cases we can test for as well?

liliankasem (approved):

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SpecializationE2ETests.cs
brettsam: Trying to think through some but can't come up with any that seem worthwhile. We're relying on the `Options` infrastructure to handle this (now that it's properly wired up) and the usage of `Lazy` so I think most of the corner cases are handled by those things.

Any ideas you had in mind?

liliankasem: nope :D

kashimiz (approved):
Looks good. Confirmed Function-level auth works when https://github.com/Azure/azure-functions-host/commit/2ba05ad5df0d63b82865c95a23314a415b64fbaa is layered on top.

PR #8335, opened 2022-04-27
[PlaceholderMode]Read RuntimeAssemblyFiles

### LINKED ISSUE #8285: Read content of all loaded modules in Functions process during placeholder mode for cold start improvement on higher percentiles
Based on Cold Start Profiles, This will improve cold start on higher percentiles.
We want to read all those dlls in 4K bytes and access one byte every 4K so they are actually read from disk and loaded in OS cache.
We do not actually want to load actual assemblies into the process just to read the content of files.

Best if this is done as part of /warmup call that DWAS does into the Functions Host process.

pragnagopa:

resolves #8285

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/Middleware/HostWarmupMiddleware.cs
+++ b/src/WebJobs.Script.WebHost/Middleware/HostWarmupMiddleware.cs
@@ -56,37 +56,40 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Middleware
             try
             {
                 string[] allFiles = Directory.GetFiles(_assemblyLocalPath, "*.dll", SearchOption.TopDirectoryOnly);
+                // Read File content in 4K chunks
+                int maxBuffer = 4 * 1024;
+                byte[] chunk = new byte[maxBuffer];
+                Random random = new Random();
                 foreach (string file in allFiles)
                 {
                     // Read file content to avoid disk reads during specialization. This is only to page-in bytes.
-                    ReadFileInChunks(file);
+                    ReadFileInChunks(file, chunk, maxBuffer, random);
                 }
-                _logger.LogDebug(new EventId(100, nameof(ReadRuntimeAssemblyFiles)), $"Number of files read: {allFiles.Count()}.");
+                _logger.LogDebug(new EventId(100, nameof(ReadRuntimeAssemblyFiles)), "Number of files read: '{allFilesCount}'. AssemblyLocalPath: '{assemblyLocalPath}' ", allFiles.Count(), _assemblyLocalPath);
             }
             catch (Exception ex)
             {
-                _logger.LogDebug(new EventId(100, nameof(ReadRuntimeAssemblyFiles)), ex, "Reading ReadRuntimeAssemblyFiles failed");
+                _logger.LogError(new EventId(100, nameof(ReadRuntimeAssemblyFiles)), ex, "Reading ReadRuntimeAssemblyFiles failed. AssemblyLocalPath: '{assemblyLocalPath}'", _assemblyLocalPath);
             }
         }
 
-        private void ReadFileInChunks(string file)
+        private void ReadFileInChunks(string file, byte[] chunk, int maxBuffer, Random random)
         {
             try
             {
-                // Read File content in 4K chunks
-                const int MAX_BUFFER = 4 * 1024;
-                byte[] chunk = new byte[MAX_BUFFER];
-                int bytesRead;
                 using (FileStream fileStream = new FileStream(file, FileMode.Open, FileAccess.Read))
                 {
-                    while ((bytesRead = fileStream.Read(chunk, 0, MAX_BUFFER)) != 0)
+                    int bytesRead;
+                    while ((bytesRead = fileStream.Read(chunk, 0, maxBuffer)) != 0)
                     {
+                        // Read one random byte for every 4K bytes - 4K is default OS page size. This will help avoid disk read during specialization
+                        var randomByte = Convert.ToInt32(chunk[random.Next(0, bytesRead - 1)]);
                     }
                 }
             }
             catch (Exception ex)
             {
-                _logger.LogDebug(new EventId(100, nameof(ReadFileInChunks)), ex, $"Reading file '{file}' failed");
+                _logger.LogError(new EventId(100, nameof(ReadFileInChunks)), ex, "Reading file '{file}' failed. AssemblyLocalPath: '{assemblyLocalPath}'", file, _assemblyLocalPath);
             }
         }
 
--- a/src/WebJobs.Script.WebHost/Middleware/HostWarmupMiddleware.cs
+++ b/src/WebJobs.Script.WebHost/Middleware/HostWarmupMiddleware.cs
@@ -83,6 +83,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Middleware
                     while ((bytesRead = fileStream.Read(chunk, 0, maxBuffer)) != 0)
                     {
                         // Read one random byte for every 4K bytes - 4K is default OS page size. This will help avoid disk read during specialization
+                        // see for details on OS page buffering in Windows - https://docs.microsoft.com/en-us/windows/win32/fileio/file-buffering
                         var randomByte = Convert.ToInt32(chunk[random.Next(0, bytesRead - 1)]);
                     }
                 }
--- a/src/WebJobs.Script.WebHost/Middleware/HostWarmupMiddleware.cs
+++ b/src/WebJobs.Script.WebHost/Middleware/HostWarmupMiddleware.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.IO;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
@@ -20,6 +21,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Middleware
         private readonly IEnvironment _environment;
         private readonly IScriptHostManager _hostManager;
         private readonly ILogger _logger;
+        private string _assemblyLocalPath;
 
         public HostWarmupMiddleware(RequestDelegate next, IScriptWebHostEnvironment webHostEnvironment, IEnvironment environment, IScriptHostManager hostManager, ILogger<HostWarmupMiddleware> logger)
         {
@@ -28,6 +30,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Middleware
             _environment = environment;
             _hostManager = hostManager;
             _logger = logger;
+            _assemblyLocalPath = Path.GetDirectoryName(new Uri(typeof(HostWarmupMiddleware).Assembly.Location).LocalPath);
         }
 
         public async Task Invoke(HttpContext httpContext)
@@ -40,17 +43,58 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Middleware
                     PreJitPrepare(WarmUpConstants.LinuxJitTraceFileName);
                 }
 
+                ReadRuntimeAssemblyFiles();
+
                 await WarmUp(httpContext.Request);
             }
 
             await _next.Invoke(httpContext);
         }
 
+        internal void ReadRuntimeAssemblyFiles()
+        {
+            try
+            {
+                string[] allFiles = Directory.GetFiles(_assemblyLocalPath, "*.dll", SearchOption.TopDirectoryOnly);
+                foreach (string file in allFiles)
+                {
+                    // Read file content to avoid disk reads during specialization. This is only to page-in bytes.
+                    ReadFileInChunks(file);
+                }
+                _logger.LogDebug(new EventId(100, nameof(ReadRuntimeAssemblyFiles)), $"Number of files read: {allFiles.Count()}.");
+            }
+            catch (Exception ex)
+            {
+                _logger.LogDebug(new EventId(100, nameof(ReadRuntimeAssemblyFiles)), ex, "Reading ReadRuntimeAssemblyFiles failed");
+            }
+        }
+
+        private void ReadFileInChunks(string file)
+        {
+            try
+            {
+                // Read File content in 4K chunks
+                const int MAX_BUFFER = 4 * 1024;
+                byte[] chunk = new byte[MAX_BUFFER];
+                int bytesRead;
+                using (FileStream fileStream = new FileStream(file, FileMode.Open, FileAccess.Read))
+                {
+                    while ((bytesRead = fileStream.Read(chunk, 0, MAX_BUFFER)) != 0)
+                    {
+                    }
+                }
+            }
+            catch (Exception ex)
+            {
+                _logger.LogDebug(new EventId(100, nameof(ReadFileInChunks)), ex, $"Reading file '{file}' failed");
+            }
+        }
+
         private void PreJitPrepare(string jitTraceFileName)
         {
             // This is to PreJIT all methods captured in coldstart.jittrace file to improve cold start time
             var path = Path.Combine(
-                Path.GetDirectoryName(new Uri(typeof(HostWarmupMiddleware).Assembly.Location).LocalPath),
+                _assemblyLocalPath,
                 WarmUpConstants.PreJitFolderName, jitTraceFileName);
 
             var file = new FileInfo(path);
--- a/test/WebJobs.Script.Tests/Middleware/HostWarmupMiddlewareTests.cs
+++ b/test/WebJobs.Script.Tests/Middleware/HostWarmupMiddlewareTests.cs
@@ -2,8 +2,12 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using System.Linq;
+using Microsoft.AspNetCore.Hosting;
 using Microsoft.Azure.WebJobs.Script.WebHost;
+using Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics.Extensions;
 using Microsoft.Azure.WebJobs.Script.WebHost.Middleware;
+using Microsoft.Extensions.Logging;
 using Microsoft.WebJobs.Script.Tests;
 using Moq;
 using Xunit;
@@ -52,6 +56,22 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.Middleware
             environment.SetEnvironmentVariable(EnvironmentSettingNames.ContainerName, "TestContainer");
             Assert.True(environment.IsLinuxConsumption());
             Assert.True(HostWarmupMiddleware.IsWarmUpRequest(request, hostEnvironment.InStandbyMode, environment));
+        }
+
+        [Fact]
+        public void ReadRuntimeAssemblyFiles_VerifyLogs()
+        {
+            var environment = new TestEnvironment();
+            var hostEnvironment = new ScriptWebHostEnvironment(environment);
+            var testLoggerFactory = new LoggerFactory();
+            TestLoggerProvider testLoggerProvider = new TestLoggerProvider();
+            testLoggerFactory.AddProvider(testLoggerProvider);
+            ILogger<HostWarmupMiddleware> testLogger = testLoggerFactory.CreateLogger<HostWarmupMiddleware>();
+            HostWarmupMiddleware hostWarmupMiddleware = new HostWarmupMiddleware(null, new Mock<IScriptWebHostEnvironment>().Object, environment, new Mock<IScriptHostManager>().Object, testLogger);
+            hostWarmupMiddleware.ReadRuntimeAssemblyFiles();
+            // Assert
+            var traces = testLoggerProvider.GetAllLogMessages();
+            Assert.True(traces.Any(m => m.FormattedMessage.Contains("Number of files read:")));
         }
     }
 }


src/WebJobs.Script.WebHost/Standby/StandbyManager.cs
```diff
@@ -209,6 +209,19 @@ private async Task CreateStandbyWarmupFunctions()
...
 
             _logger.LogInformation("StandbyMode placeholder function directory created");
+
+            ReadRuntimeAssemblyFiles();
+        }
+
+        private void ReadRuntimeAssemblyFiles()
+        {
+            string assemblyLocalPath = Path.GetDirectoryName(new Uri(typeof(StandbyManager).Assembly.Location).LocalPath);
+            string[] allFiles = Directory.GetFiles(assemblyLocalPath, "*.dll", SearchOption.TopDirectoryOnly);
```
pragnagopa: @safihamid  - reading files from TopDirectory enough?

src/WebJobs.Script.WebHost/Standby/StandbyManager.cs
```diff
@@ -209,6 +209,19 @@ private async Task CreateStandbyWarmupFunctions()
...
+            foreach (string file in allFiles)
+            {
+                // Read file content to avoid disk reads during specialization. This is only to page-in bytes.
+                File.ReadAllBytes(file);
```
safihamid: Can you please read files in 4K byte chunks so OS doesn't optimize over and make sure they are actually read?

src/WebJobs.Script.WebHost/Standby/StandbyManager.cs
```diff
@@ -209,6 +209,19 @@ private async Task CreateStandbyWarmupFunctions()
             File.WriteAllText(Path.Combine(functionPath, "run.csx"), content);
 
             _logger.LogInformation("StandbyMode placeholder function directory created");
+
+            ReadRuntimeAssemblyFiles();
+        }
+
+        private void ReadRuntimeAssemblyFiles()
+        {
+            string assemblyLocalPath = Path.GetDirectoryName(new Uri(typeof(StandbyManager).Assembly.Location).LocalPath);
```
safihamid, line 218: Add a log for the assemblyLocalPath that we are reading from?

src/WebJobs.Script.WebHost/Standby/StandbyManager.cs
safihamid: yes top directory is enough

src/WebJobs.Script.WebHost/Standby/StandbyManager.cs
safihamid: To be super safe, let's put the entire method as well as reading individual file bytes in try/catch and log exceptions if they throw but continue for the rest.

src/WebJobs.Script.WebHost/Standby/StandbyManager.cs
```diff
@@ -209,6 +209,19 @@ private async Task CreateStandbyWarmupFunctions()
             File.WriteAllText(Path.Combine(functionPath, "run.csx"), content);
 
             _logger.LogInformation("StandbyMode placeholder function directory created");
+
+            ReadRuntimeAssemblyFiles();
+        }
+
+        private void ReadRuntimeAssemblyFiles()
```
safihamid: I think it's better to put this method in HostWarmupMiddleware and run it next to PGO, ... if it is a warmup request. that way if we add more logic for warmup it can all be in same location

src/WebJobs.Script.WebHost/Middleware/HostWarmupMiddleware.cs
```diff
@@ -40,12 +41,54 @@ public async Task Invoke(HttpContext httpContext)
...
             }
 
             await _next.Invoke(httpContext);
         }
 
+        internal void ReadRuntimeAssemblyFiles()
+        {
+            try
+            {
+                string assemblyLocalPath = Path.GetDirectoryName(new Uri(typeof(StandbyManager).Assembly.Location).LocalPath);
```
safihamid: nit: let's use typeof(HostWarmupMiddleware) here similar to line 96 below

src/WebJobs.Script.WebHost/Middleware/HostWarmupMiddleware.cs
```diff
@@ -40,12 +41,54 @@ public async Task Invoke(HttpContext httpContext)
...
+                string[] allFiles = Directory.GetFiles(assemblyLocalPath, "*.dll", SearchOption.TopDirectoryOnly);
+                foreach (string file in allFiles)
+                {
+                    // Read file content to avoid disk reads during specialization. This is only to page-in bytes.
+                    ReadFileInChunks(file);
+                }
+                _logger.LogDebug(new EventId(101, nameof(ReadRuntimeAssemblyFiles)), $"Number of files read: {allFiles.Count()}.");
```
safihamid: should this be EventId 101 or 100?

src/WebJobs.Script.WebHost/Standby/StandbyManager.cs
pragnagopa: done!

src/WebJobs.Script.WebHost/Standby/StandbyManager.cs
pragnagopa: done!

src/WebJobs.Script.WebHost/Middleware/HostWarmupMiddleware.cs
pragnagopa: fixed! we can use the same event id

safihamid (dismissed):
:shipit:

safihamid (dismissed):
:shipit:

pragnagopa:
@brettsam  / @fabiocav  - can you please review as well?

src/WebJobs.Script.WebHost/Middleware/HostWarmupMiddleware.cs
```diff
@@ -28,6 +30,7 @@ public HostWarmupMiddleware(RequestDelegate next, IScriptWebHostEnvironment webH
             _environment = environment;
             _hostManager = hostManager;
             _logger = logger;
+            _assemblyLocalPath = Path.GetDirectoryName(new Uri(typeof(HostWarmupMiddleware).Assembly.Location).LocalPath);
```
fabiocav, line 33: Are you just trying to get the working directory? I see this is just reusing (moving) logic we already had, but we should be able to just get the directory without going through the assembly's location, right?

src/WebJobs.Script.WebHost/Middleware/HostWarmupMiddleware.cs
```diff
@@ -40,17 +43,58 @@ public async Task Invoke(HttpContext httpContext)
...
+                foreach (string file in allFiles)
+                {
+                    // Read file content to avoid disk reads during specialization. This is only to page-in bytes.
+                    ReadFileInChunks(file);
+                }
+                _logger.LogDebug(new EventId(100, nameof(ReadRuntimeAssemblyFiles)), $"Number of files read: {allFiles.Count()}.");
+            }
+            catch (Exception ex)
+            {
+                _logger.LogDebug(new EventId(100, nameof(ReadRuntimeAssemblyFiles)), ex, "Reading ReadRuntimeAssemblyFiles failed");
```
fabiocav: Did you intend to reuse the event ID here?

src/WebJobs.Script.WebHost/Middleware/HostWarmupMiddleware.cs
```diff
@@ -40,17 +43,58 @@ public async Task Invoke(HttpContext httpContext)
...
+        {
+            try
+            {
+                string[] allFiles = Directory.GetFiles(_assemblyLocalPath, "*.dll", SearchOption.TopDirectoryOnly);
+                foreach (string file in allFiles)
+                {
+                    // Read file content to avoid disk reads during specialization. This is only to page-in bytes.
+                    ReadFileInChunks(file);
+                }
+                _logger.LogDebug(new EventId(100, nameof(ReadRuntimeAssemblyFiles)), $"Number of files read: {allFiles.Count()}.");
```
fabiocav: Can you pass the count as arguments?

src/WebJobs.Script.WebHost/Middleware/HostWarmupMiddleware.cs
```diff
@@ -40,17 +43,58 @@ public async Task Invoke(HttpContext httpContext)
...
+                using (FileStream fileStream = new FileStream(file, FileMode.Open, FileAccess.Read))
+                {
+                    while ((bytesRead = fileStream.Read(chunk, 0, MAX_BUFFER)) != 0)
+                    {
+                    }
+                }
+            }
+            catch (Exception ex)
+            {
+                _logger.LogDebug(new EventId(100, nameof(ReadFileInChunks)), ex, $"Reading file '{file}' failed");
```
fabiocav: Pass the file as arguments. Is the event id intentional?

src/WebJobs.Script.WebHost/Middleware/HostWarmupMiddleware.cs
```diff
@@ -40,17 +43,58 @@ public async Task Invoke(HttpContext httpContext)
...
+            }
+        }
+
+        private void ReadFileInChunks(string file)
+        {
+            try
+            {
+                // Read File content in 4K chunks
+                const int MAX_BUFFER = 4 * 1024;
+                byte[] chunk = new byte[MAX_BUFFER];
```
fabiocav: Can we please avoid allocating this buffer with every iteration? Since we don't need to keep this, we should be able to reuse this buffer and make this more efficient (for a standard build of the host, we'd be able to avoid over 350KB of memory allocation here)

src/WebJobs.Script.WebHost/Middleware/HostWarmupMiddleware.cs
```diff
@@ -40,17 +43,58 @@ public async Task Invoke(HttpContext httpContext)
...
+                int bytesRead;
```
fabiocav: this local shouldn't be needed.

brettsam (commented):
To be honest, I don't really understand the magic here with the 4k buffer size, etc. Maybe this is all well-known OS behavior, but can we document it or link to it here somewhere? I don't understand why going larger than 4k causes issues, for example. And this works for both Windows and Linux?

This would definitely confuse someone coming into this code for the first time.

I also see this comment in the issue:

>We want to read all those dlls in 4K bytes and access one byte every 4K so they are actually read from disk and loaded in OS cache.

Is this PR accessing one byte every 4k somewhere?

src/WebJobs.Script.WebHost/Middleware/HostWarmupMiddleware.cs
pragnagopa: Yes, reusing the same event id  used for prejit - do you recommend defining a new id?

src/WebJobs.Script.WebHost/Middleware/HostWarmupMiddleware.cs
pragnagopa: yes, using the same event id number as prejit - in line 108

src/WebJobs.Script.WebHost/Middleware/HostWarmupMiddleware.cs
pragnagopa: good point. Addressed.

src/WebJobs.Script.WebHost/Middleware/HostWarmupMiddleware.cs
pragnagopa: done!

src/WebJobs.Script.WebHost/Middleware/HostWarmupMiddleware.cs
pragnagopa: Yes - we need the directory where host is running from - when running in Azure it would be SiteExtensions folder - @safihamid  can we change this `GetCurrentDirectory`?

pragnagopa:
@safihamid  - can you expand more on reading in 4k bytes instead of the whole file?

pragnagopa:
> Is this PR accessing one byte every 4k somewhere?

@brettsam  - good call out . @safihamid  - do we need to read the files periodically?

pragnagopa:
> Is this PR accessing one byte every 4k somewhere?

@brettsam  - good call out . @safihamid  - do we need to read the files periodically?

safihamid:
Uh good call I missed it. Yes please access one byte every 4K. This is all so OS doesn't page bytes to page file unless it has to to make sure we don't do disk reads during specialization.

src/WebJobs.Script.WebHost/Middleware/HostWarmupMiddleware.cs
safihamid: I am not sure if it works, plus this whole logic is all happening during placeholder mode so I am not too concerned about it perfwise.

src/WebJobs.Script.WebHost/Middleware/HostWarmupMiddleware.cs
```diff
@@ -40,17 +43,59 @@ public async Task Invoke(HttpContext httpContext)
...
+        private void ReadFileInChunks(string file, byte[] chunk, int maxBuffer)
+        {
+            try
+            {
+                using (FileStream fileStream = new FileStream(file, FileMode.Open, FileAccess.Read))
+                {
+                    while (fileStream.Read(chunk, 0, maxBuffer) != 0)
+                    {
+                        // Read one byte for every 4K bytes - 4K is default OS page size. This will help avoid disk read during specialization
+                        var randomByte = Convert.ToInt32(chunk[0]);
```
pragnagopa: @safihamid / @brettsam - updated PR to read 1 byte for every 4K

safihamid: Can you please read a random byte in that 4K array instead of element 0? This is to make sure OS paging doesn't outsmart byte 0.

safihamid:
FYI this is the query based on prod profiles which shows the MemoryHardFaults that causes disk reads. I am not expecting all of this to go away with this PR but it should get much better from the current stage for both Windows and Linux cold starts on higher percentiles.

FunctionsColdStartAnalyzer
| where TIMESTAMP > ago(15d)
| where HttpStatus == 200
| where EventPrimaryStampName == "waws-prod-euapbn1-003"
| extend AssemblyNames = extractall("(.*)\\(count: .*", FunctionsDetailedMemoryHardFaults)
| mv-expand AssemblyNames
| where AssemblyNames has "SiteExtensions"
| summarize count() by tostring(AssemblyNames)
| order by count_ desc

src/WebJobs.Script.WebHost/Middleware/HostWarmupMiddleware.cs
pragnagopa: done!

src/WebJobs.Script.WebHost/Middleware/HostWarmupMiddleware.cs
```diff
@@ -40,17 +43,61 @@ public async Task Invoke(HttpContext httpContext)
...
+                string[] allFiles = Directory.GetFiles(_assemblyLocalPath, "*.dll", SearchOption.TopDirectoryOnly);
+                // Read File content in 4K chunks
+                int maxBuffer = 4 * 1024;
+                byte[] chunk = new byte[maxBuffer];
+                foreach (string file in allFiles)
+                {
+                    // Read file content to avoid disk reads during specialization. This is only to page-in bytes.
+                    ReadFileInChunks(file, chunk, maxBuffer);
+                }
+                _logger.LogDebug(new EventId(100, nameof(ReadRuntimeAssemblyFiles)), "Number of files read: {allFilesCount}.", allFiles.Count());
```
safihamid, line 67: Would you please log _assemblyLocalPath as well in both success and failure logs here?

src/WebJobs.Script.WebHost/Middleware/HostWarmupMiddleware.cs
```diff
@@ -40,17 +43,61 @@ public async Task Invoke(HttpContext httpContext)
...
+        {
+            try
+            {
+                using (FileStream fileStream = new FileStream(file, FileMode.Open, FileAccess.Read))
+                {
+                    int bytesRead;
+                    while ((bytesRead = fileStream.Read(chunk, 0, maxBuffer)) != 0)
+                    {
+                        // Read one random byte for every 4K bytes - 4K is default OS page size. This will help avoid disk read during specialization
+                        Random r = new Random();
```
safihamid: You don't need to allocate this object on every iteration, it can be allocated only once similar to chunk byte[]

src/WebJobs.Script.WebHost/Middleware/HostWarmupMiddleware.cs
```diff
@@ -40,17 +43,61 @@ public async Task Invoke(HttpContext httpContext)
...
+                        var randomByte = Convert.ToInt32(chunk[r.Next(0, bytesRead - 1)]);
+                    }
+                }
+            }
+            catch (Exception ex)
+            {
+                _logger.LogDebug(new EventId(100, nameof(ReadFileInChunks)), ex, "Reading file '{file}' failed", file);
```
safihamid: Can you change the exception logs to LogError instead please?

src/WebJobs.Script.WebHost/Middleware/HostWarmupMiddleware.cs
```diff
@@ -40,17 +43,61 @@ public async Task Invoke(HttpContext httpContext)
...
+                foreach (string file in allFiles)
+                {
+                    // Read file content to avoid disk reads during specialization. This is only to page-in bytes.
+                    ReadFileInChunks(file, chunk, maxBuffer);
+                }
+                _logger.LogDebug(new EventId(100, nameof(ReadRuntimeAssemblyFiles)), "Number of files read: {allFilesCount}.", allFiles.Count());
+            }
+            catch (Exception ex)
+            {
+                _logger.LogDebug(new EventId(100, nameof(ReadRuntimeAssemblyFiles)), ex, "Reading ReadRuntimeAssemblyFiles failed");
```
safihamid: Same here, _logger.LogError inside the catch exception

src/WebJobs.Script.WebHost/Middleware/HostWarmupMiddleware.cs
pragnagopa: Error level logs - will show up for Cx as most apps have default loglevel set to Info  - can you confirm if HostMiddleware is only used in PlaceholderMode?

safihamid: Yes, this is all being called if only in placeholder mode for both Windows and Linux. take a look at IsWarmUpRequest method

safihamid (approved):
:shipit:

pragnagopa:
@fabiocav / @brettsam  - gentle ping for review !

src/WebJobs.Script.WebHost/Middleware/HostWarmupMiddleware.cs
```diff
@@ -40,17 +43,61 @@ public async Task Invoke(HttpContext httpContext)
...
 
+                ReadRuntimeAssemblyFiles();
+
                 await WarmUp(httpContext.Request);
             }
 
             await _next.Invoke(httpContext);
         }
 
+        internal void ReadRuntimeAssemblyFiles()
```
brettsam, line 54: I'd still like to see a nice comment here explaining why we're doing this. Something that explains that it forces the OS to cache the file bytes, etc -- someone new coming to this method could be pretty confused without all the context.

brettsam:
Have we been able to run this in a private stamp to ensure it improves what we're looking for?

safihamid:
@brettsam this is all observed based on profiles and helps with higher percentiles. and yes we know when there is no DISK read kernel ETW events the cold start improves. on P50s we see no disk reads.

src/WebJobs.Script.WebHost/Middleware/HostWarmupMiddleware.cs
pragnagopa: is adding a link to  - https://stackoverflow.com/questions/16976544/why-the-page-size-of-os-is-4k-and-why-it-still-stick-to-be-4k-as-the-memory-can OK?

pragnagopa:
> Have we been able to run this in a private stamp to ensure it improves what we're looking for?

This is similar change to https://github.com/Azure/azure-functions-host/pull/7822 --> @safihamid verified the production logs it did help with cold start for java function apps

src/WebJobs.Script.WebHost/Middleware/HostWarmupMiddleware.cs
safihamid: You can add link to more official doc: https://docs.microsoft.com/en-us/windows/win32/fileio/file-buffering

also FYI output from running this on nested VM:

Fsutil fsinfo ntfsinfo c:

Number Sectors :                   0x0000000007704fff
Total Clusters :                   0x0000000000ee09ff
Free Clusters  :                   0x00000000004412ed
Total Reserved :                   0x000000000000bc27
Bytes Per Sector  :                512
Bytes Per Physical Sector :        512
**Bytes Per Cluster :                4096**
Bytes Per FileRecord Segment    :  1024
Clusters Per FileRecord Segment :  0
Mft Valid Data Length :            0x000000001fa80000
Mft Start Lcn  :                   0x00000000000c0000
Mft2 Start Lcn :                   0x0000000000000002
Mft Zone Start :                   0x0000000000aaa780
Mft Zone End   :                   0x0000000000aab580
Max Device Trim Extent Count :     4096
Max Device Trim Byte Count :       0xffffffff
Max Volume Trim Extent Count :     62
Max Volume Trim Byte Count :       0x40000000

pragnagopa: done! added link to https://docs.microsoft.com/en-us/windows/win32/fileio/file-buffering

pragnagopa:
@brettsam / @fabiocav - gentle ping for review. Shared email with cold start numbers with prev change we did to read language worker file in memory.

brettsam (approved):

src/WebJobs.Script.WebHost/Middleware/HostWarmupMiddleware.cs
fabiocav: Looks like the request to pass the file as arguments was missed here.

PR #8358, opened 2022-05-04
Logging framework versions for arc cluster

chandrod:
We use to receive customer tickets for azure functions, where to reproduce customer issues we need to fetch some basic data from customer’s cluster like :
a)host version  b)Framework   c)Framework version  d)Trigger used etc. (these are examples)
Some of this data is already available in the environment variables and we wanted to fetch it.
From what information I have so far, when we push logs (platform/app) for a functionapp, those logs are also written to a file in logProcessor pods and finally fluentd helps in flowing
that data to logs analytics workspace.

With that intention I wish to introduce the log statements which would be dumped every1 hour (change verified in logs analytics workspace)
Although we as developers can find this information from logs by doing some grep/regex, but for customers this process becomes cumbersome.
With this change customer can go to their logs Analytics workspace and send us the query result of last 1 hour or so.

### Pull request checklist

Only logging, none of these apply.

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
@@ -1,0 +1,131 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host.Scale;
+using Microsoft.Azure.WebJobs.Hosting;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.WebJobs.Script.Diagnostics
+{
+    /// <summary>
+    /// Service responsible for logging versions
+    /// </summary>
+    public class FunctionsVersionLogger : IHostedService, IDisposable
+    {
+        private readonly ILogger _logger;
+        private readonly Timer _timer;
+        private readonly TimeSpan _interval;
+        private readonly IEnvironment _environment;
+        private bool _disposed;
+
+        public FunctionsVersionLogger(ILoggerFactory loggerFactory, IEnvironment environment)
+        {
+            _logger = loggerFactory.CreateLogger<FunctionsVersionLogger>();
+            _interval = TimeSpan.FromMinutes(10);
+            _timer = new Timer(OnTimer, null, Timeout.Infinite, Timeout.Infinite);
+            _environment = environment;
+        }
+
+        public Task StartAsync(CancellationToken cancellationToken)
+        {
+            // start the timer by setting the due time
+            SetTimerInterval((int)_interval.TotalMilliseconds);
+
+            return Task.CompletedTask;
+        }
+
+        public Task StopAsync(CancellationToken cancellationToken)
+        {
+            // stop the timer if it has been started
+            _timer.Change(Timeout.Infinite, Timeout.Infinite);
+
+            return Task.CompletedTask;
+        }
+
+        private async void OnTimer(object state)
+        {
+            if (true)
+            {
+                await PublishLogsSamplesAsync();
+            }
+
+            SetTimerInterval((int)_interval.TotalMilliseconds);
+        }
+
+        private Task PublishLogs()
+        {
+            try
+            {
+                if (_environment.IsKubernetesManagedHosting())
+                {
+                    _logger.LogInformation($"FUNCTIONS_EXTENSION_VERSION  : '{Environment.GetEnvironmentVariable("FUNCTIONS_EXTENSION_VERSION")}'");
+                    _logger.LogInformation($"Framework : '{Environment.GetEnvironmentVariable("FRAMEWORK")}'");
+                    _logger.LogInformation($"Framework version : '{Environment.GetEnvironmentVariable("FRAMEWORK_VERSION")}'");
+                    _logger.LogInformation($"Slot : '{Environment.GetEnvironmentVariable("WEBSITE_SLOT_NAME")}'");
+                }
+                return Task.CompletedTask;
+            }
+            catch (Exception exc) when (!exc.IsFatal())
+            {
+                return null;
+            }
+        }
+
+        private async Task PublishLogsSamplesAsync()
+        {
+            try
+            {
+                await PublishLogs();
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex, "Failed to publis logs");
+            }
+        }
+
+        private void SetTimerInterval(int dueTime)
+        {
+            if (!_disposed)
+            {
+                var timer = _timer;
+                if (timer != null)
+                {
+                    try
+                    {
+                        _timer.Change(dueTime, Timeout.Infinite);
+                    }
+                    catch (ObjectDisposedException)
+                    {
+                        // might race with dispose
+                    }
+                }
+            }
+        }
+
+        private void Dispose(bool disposing)
+        {
+            if (!_disposed)
+            {
+                if (disposing)
+                {
+                    _timer.Dispose();
+                }
+
+                _disposed = true;
+            }
+        }
+
+        public void Dispose()
+        {
+            Dispose(true);
+        }
+    }
+}
--- a/src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLoggerService.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLoggerService.cs
@@ -59,12 +59,12 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics
             return Task.CompletedTask;
         }
 
-        private async void OnTimer(object state)
+        private void OnTimer(object state)
         {
-            await PublishLogsSamplesAsync();
+            PublishLogsSamples();
         }
 
-        private Task PublishLogs()
+        private void PublishLogs()
         {
             try
             {
@@ -72,19 +72,18 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics
                 _logger.LogInformation("Framework : {0}", _environment.GetEnvironmentVariable(EnvironmentSettingNames.Framework));
                 _logger.LogInformation("FrameworkVersion : {0}", _environment.GetEnvironmentVariable(EnvironmentSettingNames.FrameworkVersion));
                 _logger.LogInformation("SlotName : {0}", _environment.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebsiteSlotName));
-                return Task.CompletedTask;
             }
             catch (Exception exc) when (!exc.IsFatal())
             {
-                return null;
+                throw;
             }
         }
 
-        private async Task PublishLogsSamplesAsync()
+        private void PublishLogsSamples()
         {
             try
             {
-                await PublishLogs();
+                PublishLogs();
             }
             catch (Exception ex)
             {
@@ -103,7 +102,7 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics
             {
                 try
                 {
-                    _timer.Change(dueTime, Timeout.Infinite);
+                    _timer.Change(0, dueTime);
                 }
                 catch (Exception)
                 {
--- a/src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLoggerService.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLoggerService.cs
@@ -18,20 +18,20 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics
     /// <summary>
     /// Service responsible for logging versions
     /// </summary>
-    public class FunctionsVersionLogger : IHostedService, IDisposable
+    public class FunctionsVersionLoggerService : IHostedService, IDisposable
     {
-        private readonly ILogger _logger;
+        private readonly ILogger<FunctionsVersionLoggerService> _logger;
         private readonly Timer _timer;
         private readonly TimeSpan _interval;
         private readonly IEnvironment _environment;
         private bool _disposed;
 
-        public FunctionsVersionLogger(ILoggerFactory loggerFactory, IEnvironment environment)
+        public FunctionsVersionLoggerService(ILogger<FunctionsVersionLoggerService> logger, IEnvironment environment)
         {
-            _logger = loggerFactory.CreateLogger<FunctionsVersionLogger>();
-            _interval = TimeSpan.FromMinutes(10);
             _timer = new Timer(OnTimer, null, Timeout.Infinite, Timeout.Infinite);
-            _environment = environment;
+            _interval = TimeSpan.FromMinutes(60);
+            _environment = environment ?? throw new ArgumentNullException(nameof(environment));
+            _logger = logger;
         }
 
         public Task StartAsync(CancellationToken cancellationToken)
@@ -45,19 +45,14 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics
         public Task StopAsync(CancellationToken cancellationToken)
         {
             // stop the timer if it has been started
-            _timer.Change(Timeout.Infinite, Timeout.Infinite);
+            _timer?.Change(Timeout.Infinite, Timeout.Infinite);
 
             return Task.CompletedTask;
         }
 
         private async void OnTimer(object state)
         {
-            if (true)
-            {
-                await PublishLogsSamplesAsync();
-            }
-
-            SetTimerInterval((int)_interval.TotalMilliseconds);
+            await PublishLogsSamplesAsync();
         }
 
         private Task PublishLogs()
@@ -66,10 +61,10 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics
             {
                 if (_environment.IsKubernetesManagedHosting())
                 {
-                    _logger.LogInformation($"FUNCTIONS_EXTENSION_VERSION  : '{Environment.GetEnvironmentVariable("FUNCTIONS_EXTENSION_VERSION")}'");
-                    _logger.LogInformation($"Framework : '{Environment.GetEnvironmentVariable("FRAMEWORK")}'");
-                    _logger.LogInformation($"Framework version : '{Environment.GetEnvironmentVariable("FRAMEWORK_VERSION")}'");
-                    _logger.LogInformation($"Slot : '{Environment.GetEnvironmentVariable("WEBSITE_SLOT_NAME")}'");
+                    _logger.LogInformation($"{EnvironmentSettingNames.FunctionsExtensionVersion}  : {_environment.GetEnvironmentVariable(EnvironmentSettingNames.FunctionsExtensionVersion)}");
+                    _logger.LogInformation($"{EnvironmentSettingNames.Framework} : {_environment.GetEnvironmentVariable(EnvironmentSettingNames.Framework)}");
+                    _logger.LogInformation($"{EnvironmentSettingNames.FrameworkVersion}: {_environment.GetEnvironmentVariable(EnvironmentSettingNames.FrameworkVersion)}");
+                    _logger.LogInformation($"{EnvironmentSettingNames.AzureWebsiteSlotName} : {_environment.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebsiteSlotName)}");
                 }
                 return Task.CompletedTask;
             }
@@ -87,45 +82,39 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics
             }
             catch (Exception ex)
             {
-                _logger.LogError(ex, "Failed to publis logs");
+                _logger.LogWarning(ex, "Failed to publish function version logs");
             }
         }
 
         private void SetTimerInterval(int dueTime)
         {
-            if (!_disposed)
+            if (_disposed)
             {
-                var timer = _timer;
-                if (timer != null)
-                {
-                    try
-                    {
-                        _timer.Change(dueTime, Timeout.Infinite);
-                    }
-                    catch (ObjectDisposedException)
-                    {
-                        // might race with dispose
-                    }
-                }
+                return;
             }
-        }
-
-        private void Dispose(bool disposing)
-        {
-            if (!_disposed)
+            var timer = _timer;
+            if (timer != null)
             {
-                if (disposing)
+                try
                 {
-                    _timer.Dispose();
+                    _timer.Change(dueTime, Timeout.Infinite);
                 }
-
-                _disposed = true;
+                catch (ObjectDisposedException)
+                {
+                    // might race with dispose
+                    _logger.LogWarning("Exception in SetTimerInterval");
+                }
             }
         }
 
         public void Dispose()
         {
-            Dispose(true);
+            if (_disposed)
+            {
+                return;
+            }
+            _timer?.Dispose();
+            _disposed = true;
         }
     }
 }
--- a/src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLoggerService.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLoggerService.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Host.Scale;
@@ -20,18 +21,26 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics
     /// </summary>
     public class FunctionsVersionLoggerService : IHostedService, IDisposable
     {
+        private const double IntervalInSecondsDev = 1;    // interval for dev unit testing (1 s) in which the logs would be dumped
         private readonly ILogger<FunctionsVersionLoggerService> _logger;
         private readonly Timer _timer;
         private readonly TimeSpan _interval;
         private readonly IEnvironment _environment;
         private bool _disposed;
+        private double intervalInSecondsProd = 3600;      // default interval (1 hour) in which the logs would be dumped
 
         public FunctionsVersionLoggerService(ILogger<FunctionsVersionLoggerService> logger, IEnvironment environment)
         {
             _timer = new Timer(OnTimer, null, Timeout.Infinite, Timeout.Infinite);
-            _interval = TimeSpan.FromMinutes(60);
             _environment = environment ?? throw new ArgumentNullException(nameof(environment));
+            _interval = TimeSpan.FromSeconds(this.Interval);
             _logger = logger;
+        }
+
+        public virtual double Interval
+        {
+            get { return (_environment.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebsiteSlotName) == "development") ? IntervalInSecondsDev : intervalInSecondsProd; }
+            set { intervalInSecondsProd = value; }
         }
 
         public Task StartAsync(CancellationToken cancellationToken)
@@ -59,13 +68,10 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics
         {
             try
             {
-                if (_environment.IsKubernetesManagedHosting())
-                {
-                    _logger.LogInformation($"{EnvironmentSettingNames.FunctionsExtensionVersion}  : {_environment.GetEnvironmentVariable(EnvironmentSettingNames.FunctionsExtensionVersion)}");
-                    _logger.LogInformation($"{EnvironmentSettingNames.Framework} : {_environment.GetEnvironmentVariable(EnvironmentSettingNames.Framework)}");
-                    _logger.LogInformation($"{EnvironmentSettingNames.FrameworkVersion}: {_environment.GetEnvironmentVariable(EnvironmentSettingNames.FrameworkVersion)}");
-                    _logger.LogInformation($"{EnvironmentSettingNames.AzureWebsiteSlotName} : {_environment.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebsiteSlotName)}");
-                }
+                _logger.LogInformation("FunctionsExtensionVersion : {0}", _environment.GetEnvironmentVariable(EnvironmentSettingNames.FunctionsExtensionVersion));
+                _logger.LogInformation("Framework : {0}", _environment.GetEnvironmentVariable(EnvironmentSettingNames.Framework));
+                _logger.LogInformation("FrameworkVersion : {0}", _environment.GetEnvironmentVariable(EnvironmentSettingNames.FrameworkVersion));
+                _logger.LogInformation("SlotName : {0}", _environment.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebsiteSlotName));
                 return Task.CompletedTask;
             }
             catch (Exception exc) when (!exc.IsFatal())
@@ -92,17 +98,17 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics
             {
                 return;
             }
-            var timer = _timer;
-            if (timer != null)
+
+            if (_timer != null)
             {
                 try
                 {
                     _timer.Change(dueTime, Timeout.Infinite);
                 }
-                catch (ObjectDisposedException)
+                catch (Exception)
                 {
                     // might race with dispose
-                    _logger.LogWarning("Exception in SetTimerInterval");
+                    _logger.LogWarning("Exception : timer disposed, so unable to set timer interval");
                 }
             }
         }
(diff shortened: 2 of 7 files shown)


src/WebJobs.Script.WebHost/Program.cs
```diff
@@ -81,12 +81,26 @@ public static IWebHostBuilder CreateWebHostBuilder(string[] args = null)
                 .UseStartup<Startup>();
         }
 
+        internal static void LogVersions(object state)
+        {
+            Console.WriteLine($"HOST_VERSION : {Environment.GetEnvironmentVariable("HOST_VERSION")}");
+            Console.WriteLine($"FUNCTIONS_EXTENSION_VERSION : {Environment.GetEnvironmentVariable("FUNCTIONS_EXTENSION_VERSION")}");
+            Console.WriteLine($"FRAMEWORK : {Environment.GetEnvironmentVariable("FRAMEWORK")}");
+            Console.WriteLine($"FRAMEWORK_VERSION : {Environment.GetEnvironmentVariable("FRAMEWORK_VERSION")}");
+        }
```
divyagandhisethi: Do these values change with time? Why do we need to have a timer for this method?

src/WebJobs.Script.WebHost/Program.cs
```diff
@@ -81,12 +81,26 @@ public static IWebHostBuilder CreateWebHostBuilder(string[] args = null)
...
+
         /// <summary>
         /// Perform any process level initialization that needs to happen BEFORE
         /// the WebHost is initialized.
         /// </summary>
         private static void InitializeProcess()
         {
+            Timer tTimer = null;
+            tTimer = new Timer(
+                        new TimerCallback(LogVersions),
```
divyagandhisethi: Please add a check to scope this only to arc environments

src/WebJobs.Script.WebHost/Program.cs
```diff
@@ -81,12 +81,26 @@ public static IWebHostBuilder CreateWebHostBuilder(string[] args = null)
                 .UseStartup<Startup>();
         }
 
+        internal static void LogVersions(object state)
+        {
+            Console.WriteLine($"HOST_VERSION : {Environment.GetEnvironmentVariable("HOST_VERSION")}");
+            Console.WriteLine($"FUNCTIONS_EXTENSION_VERSION : {Environment.GetEnvironmentVariable("FUNCTIONS_EXTENSION_VERSION")}");
+            Console.WriteLine($"FRAMEWORK : {Environment.GetEnvironmentVariable("FRAMEWORK")}");
+            Console.WriteLine($"FRAMEWORK_VERSION : {Environment.GetEnvironmentVariable("FRAMEWORK_VERSION")}");
```
divyagandhisethi: I think you can still get all this data from the last 30 mins result from their log analytics workspace. We don't need to explicitly log environment variables.

divyagandhisethi (changes requested):

src/WebJobs.Script.WebHost/Program.cs
chandrod: For sampling, I exported data for last 12 hours for a cluster which had 4 functionapps already running : 
https://microsoft-my.sharepoint.com/:x:/p/shubhamc/EQ1SJNdAXvhCvaKHXMbMSJIBu_4tM_ZEJtWwDBRuUxwkAQ?e=T4DDxU
it does not have the host version. Although out of the 4 apps in the cluster, one of them has. So, I wanted to just publish explicitly/uniformly. Maybe it is fine to just publish maybe once every 30 mins or so. Would be interested to know your opinion.

src/WebJobs.Script.WebHost/Program.cs
chandrod: These values don't change usually unless the customer explicitly change them in funcapp configuration. Also, if I don't put timer it might get published when the pod/app starts and never again. Although I am totally in favour of increasing the time duration so increasing it to 15 mins. Please let me know what do you think

src/WebJobs.Script.WebHost/Program.cs
chandrod: Done

src/WebJobs.Script.WebHost/Program.cs
divyagandhisethi: Yea these values are generally never changed by customers - I would suggest either only publish at startup or publishing at a much lower frequency - ex: once per hour.

src/WebJobs.Script.WebHost/Program.cs
divyagandhisethi: I don't see any logs in this link above..it's only logging the lock call to the pod. Do we not send the stdout log stream of the pod to customer's Azure Monitor?

chandrod: It is because the lock acquire runs periodically every few seconds/mins and fills up most of the logs. So the log analytics workspace come under Azure monitor only and we can actually see the functionapp pod logs here if lock call logs don't fill it up repeatedly. Please feel free to reopen, we can discuss offline as well if this does not suffice.

src/WebJobs.Script.WebHost/Program.cs
chandrod: Done

liliankasem:
Hi @chandrod, I think there's potentially a different approach you can take here; we'd like to avoid setting up a timer in the start up flow. The approach we think you should take here is to register a custom hosted service instead. here is an example you can use as a reference:

- [FunctionsScaleMonitorService]( https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Scale/FunctionsScaleMonitorService.cs)
 
For Arc, are you running your own build of the host?

- If yes, you can register the hosted services outside of our repo, injecting in your own service
- If not, you can still use this service approach and we can conditionally register your service based on environment

chandrod:
@liliankasem your comment makes total sense. Made those changes. Please take a look.

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
```diff
@@ -0,0 +1,131 @@
...
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.WebJobs.Script.Diagnostics
+{
+    /// <summary>
+    /// Service responsible for logging versions
+    /// </summary>
+    public class FunctionsVersionLogger : IHostedService, IDisposable
+    {
+        private readonly ILogger _logger;
```
liliankasem: Shouldn't have to use the logger factory in the constructor

```suggestion
        private readonly ILogger<FunctionsVersionLogger> _logger;
```

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
```diff
@@ -0,0 +1,131 @@
...
+        private readonly Timer _timer;
+        private readonly TimeSpan _interval;
+        private readonly IEnvironment _environment;
+        private bool _disposed;
+
+        public FunctionsVersionLogger(ILoggerFactory loggerFactory, IEnvironment environment)
```
liliankasem:
```suggestion
        public FunctionsVersionLogger(ILogger<FunctionsVersionLogger> logger, IEnvironment environment)
```

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
```diff
@@ -0,0 +1,131 @@
...
+        {
+            _logger = loggerFactory.CreateLogger<FunctionsVersionLogger>();
+            _interval = TimeSpan.FromMinutes(10);
+            _timer = new Timer(OnTimer, null, Timeout.Infinite, Timeout.Infinite);
+            _environment = environment;
```
liliankasem: nit: move this to above `_interval = TimeSpan.FromMinutes(10);` to keep parameter items together

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
```diff
@@ -0,0 +1,131 @@
...
+        private readonly ILogger _logger;
+        private readonly Timer _timer;
+        private readonly TimeSpan _interval;
+        private readonly IEnvironment _environment;
+        private bool _disposed;
+
+        public FunctionsVersionLogger(ILoggerFactory loggerFactory, IEnvironment environment)
+        {
+            _logger = loggerFactory.CreateLogger<FunctionsVersionLogger>();
+            _interval = TimeSpan.FromMinutes(10);
```
liliankasem: I thought this was supposed to only log every hour?

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
```diff
@@ -0,0 +1,131 @@
...
+        {
+            // stop the timer if it has been started
+            _timer.Change(Timeout.Infinite, Timeout.Infinite);
+
+            return Task.CompletedTask;
+        }
+
+        private async void OnTimer(object state)
+        {
+            if (true)
```
liliankasem: Why do we need this if-statement if it's always true?

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
```diff
@@ -0,0 +1,131 @@
...
+            {
+                await PublishLogsSamplesAsync();
+            }
+
+            SetTimerInterval((int)_interval.TotalMilliseconds);
```
liliankasem: Wondering why we need to set the timer interval again here, we already do it on StartAsync?

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
```diff
@@ -0,0 +1,131 @@
...
+
+        private async Task PublishLogsSamplesAsync()
+        {
+            try
+            {
+                await PublishLogs();
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex, "Failed to publis logs");
```
liliankasem:
```suggestion
                _logger.LogError(ex, "Failed to publish function version logs");
```

Also wondering if this should be an error, it doesn't break anything and might be better logged as warning

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
```diff
@@ -0,0 +1,131 @@
...
+                    {
+                        // might race with dispose
+                    }
+                }
+            }
+        }
+
+        private void Dispose(bool disposing)
+        {
+            if (!_disposed)
```
liliankasem: Inverse this statement:

```
if (_disposed)
{
    return;
}
```

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
```diff
@@ -0,0 +1,131 @@
...
+                    }
+                    catch (ObjectDisposedException)
+                    {
+                        // might race with dispose
+                    }
+                }
+            }
+        }
+
+        private void Dispose(bool disposing)
```
liliankasem: Why do we need a private dispose method here? I suggest scrapping this and moving the logic down to `public void Dispose()`, getting rid of `disposing`

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
```diff
@@ -0,0 +1,131 @@
...
+        {
+            if (!_disposed)
+            {
+                if (disposing)
+                {
+                    _timer.Dispose();
```
liliankasem:
```suggestion
                    _timer?.Dispose();
```

src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
```diff
@@ -152,6 +152,8 @@ public static class WebScriptHostBuilderExtension
                     services.AddSingleton<IFileMonitoringService, FileMonitoringService>();
                     services.TryAddEnumerable(ServiceDescriptor.Singleton<IHostedService, IFileMonitoringService>(p => p.GetService<IFileMonitoringService>()));
 
+                    services.TryAddEnumerable(ServiceDescriptor.Singleton<IHostedService, FunctionsVersionLogger>());
```
liliankasem: Why `TryAddEnumerable` over `services.AddSingleton<IHostedService, FunctionsVersionLogger >();` ?

Also @fabiocav is this the correct place for this? Wondering if it should be in `WebHostServiceCollectionExtensions`?

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
```diff
@@ -0,0 +1,131 @@
...
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.WebJobs.Script.Diagnostics
+{
+    /// <summary>
+    /// Service responsible for logging versions
+    /// </summary>
+    public class FunctionsVersionLogger : IHostedService, IDisposable
```
liliankasem:
```suggestion
    public class FunctionsVersionLoggerService : IHostedService, IDisposable
```

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
```diff
@@ -0,0 +1,131 @@
...
+        private Task PublishLogs()
+        {
+            try
+            {
+                if (_environment.IsKubernetesManagedHosting())
+                {
+                    _logger.LogInformation($"FUNCTIONS_EXTENSION_VERSION  : '{Environment.GetEnvironmentVariable("FUNCTIONS_EXTENSION_VERSION")}'");
+                    _logger.LogInformation($"Framework : '{Environment.GetEnvironmentVariable("FRAMEWORK")}'");
+                    _logger.LogInformation($"Framework version : '{Environment.GetEnvironmentVariable("FRAMEWORK_VERSION")}'");
+                    _logger.LogInformation($"Slot : '{Environment.GetEnvironmentVariable("WEBSITE_SLOT_NAME")}'");
```
liliankasem: These environment variables should already exist in a constants file somewhere e.g. https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Environment/EnvironmentSettingNames.cs

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
```diff
@@ -0,0 +1,131 @@
...
+            }
+
+            SetTimerInterval((int)_interval.TotalMilliseconds);
+        }
+
+        private Task PublishLogs()
+        {
+            try
+            {
+                if (_environment.IsKubernetesManagedHosting())
```
liliankasem: I think we want to have this environment check where we do the DI instead, so that the service is not added for all

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
```diff
@@ -0,0 +1,131 @@
...
+                {
+                    _logger.LogInformation($"FUNCTIONS_EXTENSION_VERSION  : '{Environment.GetEnvironmentVariable("FUNCTIONS_EXTENSION_VERSION")}'");
```
liliankasem: Should this be:

```suggestion
                    _logger.LogInformation($"FUNCTIONS_EXTENSION_VERSION  : '{_environment.GetEnvironmentVariable("FUNCTIONS_EXTENSION_VERSION")}'");
```

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
```diff
@@ -0,0 +1,131 @@
...
+            // start the timer by setting the due time
+            SetTimerInterval((int)_interval.TotalMilliseconds);
+
+            return Task.CompletedTask;
+        }
+
+        public Task StopAsync(CancellationToken cancellationToken)
+        {
+            // stop the timer if it has been started
+            _timer.Change(Timeout.Infinite, Timeout.Infinite);
```
liliankasem:
```suggestion
            _timer?.Change(Timeout.Infinite, Timeout.Infinite);
```

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
liliankasem:
```suggestion
            _environment = environment ?? throw new ArgumentNullException(nameof(environment));
```

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
liliankasem: Should check if disposed here

liliankasem (changes requested):

liliankasem:
Please also add some tests

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
```diff
@@ -0,0 +1,131 @@
...
+
+        private Task PublishLogs()
+        {
+            try
+            {
+                if (_environment.IsKubernetesManagedHosting())
+                {
+                    _logger.LogInformation($"FUNCTIONS_EXTENSION_VERSION  : '{Environment.GetEnvironmentVariable("FUNCTIONS_EXTENSION_VERSION")}'");
+                    _logger.LogInformation($"Framework : '{Environment.GetEnvironmentVariable("FRAMEWORK")}'");
+                    _logger.LogInformation($"Framework version : '{Environment.GetEnvironmentVariable("FRAMEWORK_VERSION")}'");
```
liliankasem: For all logging here, please pass the values as arguments into the log methods instead of using interpolation/concatenating the strings.

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
chandrod: Done

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
chandrod: Done

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
chandrod: Done

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
chandrod: Updated

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
chandrod: Done

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
chandrod: StartAsync uses this function to set the interval, sends 60 mins as a parameter

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
chandrod: Done

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
chandrod: Done

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
chandrod: Done

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
chandrod: Done

src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
chandrod: Integration test was failing with AddSingleton, please let me know if this is an absolute priority

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
chandrod: Done

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
chandrod: Done

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
chandrod: Done

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
chandrod: Done

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
chandrod: Done

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
chandrod: Done

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLogger.cs
chandrod: > Please also add some tests

Done

chandrod: > For all logging here, please pass the values as arguments into the log methods instead of using interpolation/concatenating the strings.
Done

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLoggerService.cs
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        private void SetTimerInterval(int dueTime)
+        {
+            if (_disposed)
+            {
+                return;
+            }
+            var timer = _timer;
+            if (timer != null)
```
liliankasem: this seems unnecessary? Any reason to not use _timer directly?
```suggestion
            if (_timer != null)
```

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLoggerService.cs
```diff
@@ -0,0 +1,128 @@
...
+            }
+        }
+
+        private void SetTimerInterval(int dueTime)
+        {
+            if (_disposed)
+            {
+                return;
+            }
+            var timer = _timer;
```
liliankasem: nit: new line after closing }

```suggestion

            var timer = _timer;
```

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLoggerService.cs
```diff
@@ -0,0 +1,128 @@
...
+        private async void OnTimer(object state)
+        {
+            await PublishLogsSamplesAsync();
+        }
+
+        private Task PublishLogs()
+        {
+            try
+            {
+                if (_environment.IsKubernetesManagedHosting())
```
liliankasem: We already have this check in `WebScriptHostBuilderExtension.cs` so we can remove it here, I don't see a reason to check twice

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLoggerService.cs
```diff
@@ -0,0 +1,128 @@
...
+        {
+            _timer = new Timer(OnTimer, null, Timeout.Infinite, Timeout.Infinite);
+            _environment = environment ?? throw new ArgumentNullException(nameof(environment));
+            _interval = TimeSpan.FromSeconds(this.Interval);
+            _logger = logger;
+        }
+
+        public virtual double Interval
+        {
+            get { return (_environment.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebsiteSlotName) == "development") ? 3 : intervalInSecondsProd; }
```
liliankasem: Can we assign the value `3`  to a named variable for readability

liliankasem: Also why is it 3 seconds in development? That seems really excessive

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLoggerService.cs
```diff
@@ -0,0 +1,128 @@
...
+            if (timer != null)
+            {
+                try
+                {
+                    _timer.Change(dueTime, Timeout.Infinite);
+                }
+                catch (ObjectDisposedException)
+                {
+                    // might race with dispose
+                    _logger.LogWarning("Exception in SetTimerInterval");
```
liliankasem: Could other exceptions occur here or just `ObjectDisposedException`? Wondering if we need to account for a more generic Exception or not?

Also please have a clearly exception message here, i.e. something about the timer being disposed so we are unable to set the timer interval

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLoggerService.cs
```diff
@@ -0,0 +1,128 @@
...
+            }
+        }
+
+        public void Dispose()
+        {
+            if (_disposed)
+            {
+                return;
+            }
+            _timer?.Dispose();
```
liliankasem, line 122: nit: new line after closing brace

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLoggerService.cs
chandrod: Done

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLoggerService.cs
chandrod: Done

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLoggerService.cs
chandrod: Changed to 1s and used a named variable. Done

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLoggerService.cs
chandrod: Done

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLoggerService.cs
chandrod: Done

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLoggerService.cs
```diff
@@ -0,0 +1,128 @@
...
+            _timer = new Timer(OnTimer, null, Timeout.Infinite, Timeout.Infinite);
+            _environment = environment ?? throw new ArgumentNullException(nameof(environment));
+            _interval = TimeSpan.FromSeconds(this.Interval);
+            _logger = logger;
+        }
+
+        public virtual double Interval
+        {
+            get { return (_environment.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebsiteSlotName) == "development") ? intervalInSecondsDev : intervalInSecondsProd; }
+            set { intervalInSecondsProd = value; }
```
madelinegordon, line 43: Should there be a condition (dev vs. prod) in the set here too? And if not, should intervalInSecondsDev be declared as a constant?

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLoggerService.cs
chandrod: Done

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLoggerService.cs
chandrod: Done (declared as constant)

chandrod:
> Hi @chandrod, I think there's potentially a different approach you can take here; we'd like to avoid setting up a timer in the start up flow. The approach we think you should take here is to register a custom hosted service instead. here is an example you can use as a reference:
> 
> * [FunctionsScaleMonitorService](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Scale/FunctionsScaleMonitorService.cs)
> 
> For Arc, are you running your own build of the host?
> 
> * If yes, you can register the hosted services outside of our repo, injecting in your own service
> * If not, you can still use this service approach and we can conditionally register your service based on environment

Done

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLoggerService.cs
```diff
@@ -0,0 +1,127 @@
...
+
+            return Task.CompletedTask;
+        }
+
+        private async void OnTimer(object state)
+        {
+            await PublishLogsSamplesAsync();
+        }
+
+        private Task PublishLogs()
```
brettsam: Does this need to be a `Task`? You're not doing anything asynchronously here. It also removes the need for the `async void` for `OnTimer`.

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLoggerService.cs
```diff
@@ -0,0 +1,127 @@
...
+            if (_disposed)
+            {
+                return;
+            }
+
+            if (_timer != null)
+            {
+                try
+                {
+                    _timer.Change(dueTime, Timeout.Infinite);
```
brettsam: This will only run the timer once, I think. Is that what you wanted?

What you may have intended was `_timer.Change(0, dueTime)`, which would invoke the timer immediately, then every `dueTime` after that.

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLoggerService.cs
chandrod: Done

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLoggerService.cs
chandrod: Done

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLoggerService.cs
```diff
@@ -0,0 +1,126 @@
...
+                _logger.LogInformation("FrameworkVersion : {0}", _environment.GetEnvironmentVariable(EnvironmentSettingNames.FrameworkVersion));
+                _logger.LogInformation("SlotName : {0}", _environment.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebsiteSlotName));
+            }
+            catch (Exception exc) when (!exc.IsFatal())
+            {
+                throw;
+            }
+        }
+
+        private void PublishLogsSamplesAsync()
```
brettsam: nit: remove `Async` from this method name....

brettsam (approved):
Just one tiny comment.

src/WebJobs.Script.WebHost/Diagnostics/FunctionsVersionLoggerService.cs
chandrod: Done

chandrod:
@CooperLink. Had a short discussion with functions-host team. We need your approval as well if we wish to merge this PR. Could you please take a look ? Thank you

pragnagopa:
Curious to understand why this change is needed when most of the information is already logged in script startup message -  - https://github.com/Azure/azure-functions-host/blob/15e1c2bc4b610f42a1e40ad5de81dba478db6dd3/src/WebJobs.Script/Diagnostics/Extensions/ScriptHostLoggerExtension.cs#L126-L132 as part of ScriptHost startup

Are these logs missing in log analytics?

liliankasem:
Closing as migrated to #8711

PR #8361, opened 2022-05-04
returning error if the Storage operation for secret management fails

### LINKED ISSUE #7591: BlobStorageSecretsRepository Write failure should fail key creation request

soninaren:

resolves #7591

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
+++ b/src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
@@ -330,9 +330,16 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
                 return secrets;
             }, secretsFactory).ContinueWith(t =>
             {
-                if (t.IsFaulted && t.Exception.InnerException is KeyVaultErrorException)
+                if (t.IsFaulted)
                 {
-                    result = OperationResult.Forbidden;
+                    _logger.LogError("Error adding or updating secrets", t.Exception.InnerException);
+
+                    result = OperationResult.Error;
+
+                    if (t.Exception.InnerException is KeyVaultErrorException)
+                    {
+                        result = OperationResult.Forbidden;
+                    }
                 }
             });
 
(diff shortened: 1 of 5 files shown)


src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
```diff
@@ -330,9 +330,16 @@ public async Task<bool> DeleteSecretAsync(string secretName, string keyScope, Sc
...
             {
-                if (t.IsFaulted && t.Exception.InnerException is KeyVaultErrorException)
+                if (t.IsFaulted)
                 {
-                    result = OperationResult.Forbidden;
+                    _logger.LogError("Error adding or updating secrets", t.Exception.InnerException);
+
+                    result = OperationResult.Error;
+
+                    if (t.Exception.InnerException is KeyVaultErrorException)
```
mathewc: @alrod - It looks like you added this KeyVaultErrorException=>403 conversion code when you initially added the keyvault provider [here](https://github.com/Azure/azure-functions-host/commit/2b8c2b851e2a415d70b40e7d47bc415a0a82475a#diff-5edb54e856f748486383109589f7c8980f5d5a68bb4e0a809ff20d6374c53452R318). Is there a reason why we need to retain this? I'm thinking that across the board, if SecretManager receives a Fault updating secrets, it should just return a 500.

soninaren: agreed, @alrod any concerns changing error code to 500?

alrod: it should be 500, please change this

liliankasem (dismissed):

mathewc (approved):

liliankasem (approved):

PR #8362, opened 2022-05-05
Return 500 instead of 403 when writing secrets to storage fails

### LINKED ISSUE #7591: BlobStorageSecretsRepository Write failure should fail key creation request

soninaren:

resolves #7591

### Pull request checklist

    * [x] Otherwise: Backport tracked by issue/PR #8361

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
+++ b/src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
@@ -329,10 +329,19 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
                 return secrets;
             }, secretsFactory).ContinueWith(t =>
             {
-                if (t.IsFaulted && t.Exception.InnerException is RequestFailedException)
+                if (t.IsFaulted)
                 {
-                    result = OperationResult.Forbidden;
+                    _logger.LogError("Error adding or updating secrets", t.Exception.InnerException);
+
+                    result = OperationResult.Error;
+
+                    if (t.Exception.InnerException is RequestFailedException)
+                    {
+                        result = OperationResult.Forbidden;
+                    }
                 }
+
+                return result;
             });
 
             return new KeyOperationResult(secret, result);
--- a/src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
+++ b/src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
@@ -328,24 +328,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
                 }
 
                 return secrets;
-            }, secretsFactory).ContinueWith(t =>
-            {
-                if (t.IsFaulted)
-                {
-                    _logger.LogError("Error adding or updating secrets", t.Exception.InnerException);
-
-                    result = OperationResult.Error;
-
-                    var exception = t.Exception.InnerException as RequestFailedException;
-
-                    if (exception?.Status == (int)HttpStatusCode.Forbidden)
-                    {
-                        result = OperationResult.Forbidden;
-                    }
-                }
-
-                return result;
-            });
+            }, secretsFactory);
 
             return new KeyOperationResult(secret, result);
         }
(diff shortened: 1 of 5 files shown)


src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
```diff
@@ -329,10 +329,19 @@ public async Task<bool> DeleteSecretAsync(string secretName, string keyScope, Sc
...
             {
-                if (t.IsFaulted && t.Exception.InnerException is RequestFailedException)
+                if (t.IsFaulted)
                 {
-                    result = OperationResult.Forbidden;
+                    _logger.LogError("Error adding or updating secrets", t.Exception.InnerException);
+
+                    result = OperationResult.Error;
+
+                    if (t.Exception.InnerException is RequestFailedException)
```
mathewc: I don't think we need to retain this special 403 conversion code. @liliankasem updated this logic in [this commit](https://github.com/Azure/azure-functions-host/commit/a642d6d3aadeb05eff1e2031fc8c7824b32b97fb) - it used to be scoped specifically to a KeyVaultErrorException, but was changed to the more general RequestFailedException. We shouldn't be converting general failures to 403s.

soninaren: agreed, @liliankasem any concerns changing error response to 500?

liliankasem: Ohhh I remember this. With the change to `Azure.Identity`, they return a `RequestFailedException` instead of a `KeyVaultErrorException` so that had to change. I think it's fair that we should not return 403 for all `RequestFailedException`, but I still want us to return 403 if the exception was due to being forbidden

liliankasem: I wonder if we need to check the status code somewhere from the initial request to make that happen

mathewc: Based on discussion of the original code in PR https://github.com/Azure/azure-functions-host/pull/8361, I think we're agreeing that we shouldn't have been doing the conversion in the first place, so the special handling will go away in both cases.

liliankasem: Sounds good, then let's remove it!

liliankasem (approved):

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
soninaren: @liliankasem I had some further discussion with mathew, I realized that returning 403 would be incorrect because to the user that would mean the call to generate the key does not have the correct auth, which is not true. If we fail to generate / write a key we should be returning 500. Have updated my code to reflect the same.

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
```diff
@@ -327,13 +328,7 @@ public async Task<bool> DeleteSecretAsync(string secretName, string keyScope, Sc
                 }
 
                 return secrets;
-            }, secretsFactory).ContinueWith(t =>
-            {
-                if (t.IsFaulted && t.Exception.InnerException is RequestFailedException)
-                {
-                    result = OperationResult.Forbidden;
```
mathewc, line 334: Remove OperationResult.Forbidden enum since it's not used anymore

soninaren: done

soninaren: done

liliankasem (approved):

mathewc (approved):

PR #8363, opened 2022-05-05
Adding ability to allow worker to send Function load responses in batch

surgupta-msft:

Issue - https://github.com/Azure/azure-functions-host/issues/8082

Adding ability to allow worker to send Function load responses in batch. Worker can choose to send load responses one by one using FunctionLoadResponse field or in batch using FunctionLoadResponseCollection field.

### Pull request checklist

### Additional information

Additional PR information


src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -287,20 +287,6 @@ public void SendFunctionLoadRequests(ManagedDependencyOptions managedDependencyO
...
-                }
-                else
-                {
-                    _eventSubscriptions.Add(_inboundWorkerEvents.Where(msg => msg.MessageType == MsgType.FunctionLoadResponse)
-                        .Subscribe((msg) => LoadResponse(msg.Message.FunctionLoadResponse), HandleWorkerFunctionLoadError));
-                }
-
                 // Load Request is also sent for disabled function as it is invocable using the portal and admin endpoints
                 // Loading disabled functions at the end avoids unnecessary performance issues. Refer PR #5072 and commit #38b57883be28524fa6ee67a457fa47e96663094c
                 _functions = _functions.OrderBy(metadata => metadata.IsDisabled());
```
soninaren, line 292: I could be wrong, but this would put the disabled functions first in the list.

soninaren (commented):
Just a nit comment

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
surgupta-msft: This SO link says that when you order based on a boolean value then false values will be kept first and then true values - https://stackoverflow.com/a/5408347

soninaren (approved):

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -309,10 +295,36 @@ public void SendFunctionLoadRequests(ManagedDependencyOptions managedDependencyO
                 bool capabilityEnabled = !string.IsNullOrEmpty(_workerCapabilities.GetCapabilityState(RpcWorkerConstants.AcceptsListOfFunctionLoadRequests));
                 if (capabilityEnabled)
```
fabiocav, line 296: Outside of the scope of you changes, but can we have a better name for this variable? Perhaps something like `supportsLoadResponseCollection`?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -309,10 +295,36 @@ public void SendFunctionLoadRequests(ManagedDependencyOptions managedDependencyO
...
                 {
+                    if (functionTimeout.HasValue)
```
fabiocav, line 299: Is this the right timeout? Isn't this for the function execution timeout? We need to use the appropriate response timeout here (which would be in seconds)

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -309,10 +295,36 @@ public void SendFunctionLoadRequests(ManagedDependencyOptions managedDependencyO
...
+                    {
+                        _functionLoadTimeout = functionTimeout.Value > _functionLoadTimeout ? functionTimeout.Value : _functionLoadTimeout;
+                        _eventSubscriptions.Add(_inboundWorkerEvents.Where(msg => msg.MessageType == MsgType.FunctionLoadResponse)
+                            .Timeout(_functionLoadTimeout)
+                            .Take(_functions.Count())
+                            .Subscribe((msg) => LoadResponse(msg.Message.FunctionLoadResponse), HandleWorkerFunctionLoadError));
+                    }
+                    else
+                    {
+                        _eventSubscriptions.Add(_inboundWorkerEvents.Where(msg => msg.MessageType == MsgType.FunctionLoadResponse)
```
fabiocav: Do we not need the `Take` here?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
surgupta-msft: Done

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
surgupta-msft: This code was already there for sending load requests - both the Timeout and the Take. They are using FunctionLoadTimeout which is set to 10 minutes. Let me know if it needs to be updated.

Older commit from oct 2021 -
https://github.com/Azure/azure-functions-host/blob/c744199f4cb5dee95f4941a79b0d0e591abefa56/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs#L289-L301

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
surgupta-msft: Updated code as discussed offline.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
surgupta-msft: Updated code as discussed offline.

liliankasem (approved):

PR #8370, opened 2022-05-10
Enabling DiagnosticEvents, minor bug fixes

soninaren:

- No issue required, just enabling a feature

### Pull request checklist

    * [ ] Otherwise: Backport tracked by issue/PR #8371

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEvent.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEvent.cs
@@ -21,7 +21,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 
         public int HitCount { get; set; }
 
-        public DateTime LastTimeStamp { get; set; }
+        public DateTimeOffset TimeStamp { get; set; }
 
         public string Message { get; set; }
 
--- a/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
@@ -184,7 +184,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
             {
                 ErrorCode = errorCode,
                 HelpLink = helpLink,
-                LastTimeStamp = timestamp,
+                TimeStamp = timestamp,
                 Message = message,
                 LogLevel = level,
                 Details = exception?.ToFormattedString(),
--- a/src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
+++ b/src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
@@ -90,7 +90,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
                         loggingBuilder.Services.AddSingleton<ILoggerProvider, AzureMonitorDiagnosticLoggerProvider>();
                     }
 
-                    if (FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagEnableDiagnosticEventLogging))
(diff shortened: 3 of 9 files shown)


liliankasem (approved):

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEvent.cs
```diff
@@ -21,7 +21,7 @@ public DiagnosticEvent(string hostId, DateTime timestamp)
 
         public int HitCount { get; set; }
 
-        public DateTime LastTimeStamp { get; set; }
+        public DateTimeOffset TimeStamp { get; set; }
```
kshyju, line 24: Should it be `Timestamp` ?

kshyju: Another place where it is called `Timestamp` instead of `TimeStamp` https://github.com/Azure/azure-functions-host/blob/dev/test/WebJobs.Script.Tests.Shared/FunctionExecutionEvent.cs#L27

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -184,7 +184,7 @@ public void WriteDiagnosticEvent(DateTime timestamp, string errorCode, LogLevel
             {
                 ErrorCode = errorCode,
                 HelpLink = helpLink,
-                LastTimeStamp = timestamp,
+                TimeStamp = timestamp,
```
kshyju, line 187: nit question: Should the constructor of `DiagnosticEvent` also set the ~Timestamp` property? I see we are already passing that piece of data?

src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
```diff
@@ -90,7 +90,7 @@ public static class WebScriptHostBuilderExtension
                         loggingBuilder.Services.AddSingleton<ILoggerProvider, AzureMonitorDiagnosticLoggerProvider>();
                     }
 
-                    if (FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagEnableDiagnosticEventLogging))
```
kshyju, line 93: IMHO, the `Enable` version (instead of the new Disable version) of the feature flag name seems less confusing.

soninaren: The disable is just a safeguard in case a customer is facing an issue with the first release. We will remove disable eventually.

kshyju (approved):
What minor bug are we fixing? Might be better to add the details to the PR description to help future readers.

soninaren:
> What minor bug are we fixing? Might be better to add the details to the PR description to help future readers.

LogError, was adding a log an info level

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEvent.cs
soninaren: actually this field needs to be removed as it is part of the TableEntity already. I think i missed adding one more commit

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
soninaren: Good point, let me double check if need setters to be called explicitly.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEvent.cs
soninaren: the TimeStamp field needs to be removed as table entity already has it.

PR #8373, opened 2022-05-10
Updating httpClient to not follow redirects with EnableForwardingHttpRequest is true

### LINKED ISSUE #7860: [Custom Handlers] Add option for host to not follow redirects
For custom handlers, when `enableForwardingHttpRequest` is enabled, when the custom handler returns a 301, 302, or 307, this is not propagated back in the HTTP response from the function host. Instead, the host [follows](https://github.com/Azure/azure-functions-host/blob/1010912485d019bedb30bcb32086e9a54ab2cbca/src/WebJobs.Script/Workers/Http/DefaultHttpWorkerService.cs#L120) the redirect and returns the response from that. 

There should be an option to disable following redirects and simply propagate the 30x responses back to the caller.

No known workarounds.

#### Additional context
The current behavior is probably considered a bug but since it's a breaking change, it needs to be behind a flag.

soninaren:

resolves #7860 

### Pull request checklist

### Additional information

Additional PR information


liliankasem (commented):
do we need to add some sort of unit test for this?

soninaren:
> do we need to add some sort of unit test for this?

Can't really add a unit test because there is no way to inspect httpClient AllowAutoRedirect of an httpClient. That is what unit test would supposedly cover. It would check httpClient if AllowAutoRedirect in relation to EnableForwardingHttpRequest

liliankasem (approved):

Issue #3296, opened 2018-08-20
Allow workers to terminate gracefully by sending `WorkerTerminate` message


mhoeger:
We should send a [WorkerTerminate](https://github.com/Azure/azure-functions-language-worker-protobuf/blob/58b3dc0dbde065b15bd979bd2e27f36b7ee6273e/src/proto/FunctionRpc.proto#L39) message to from host to worker to allow the worker to terminate itself gracefully instead of immediately killing it.

Step 1 to allowing https://github.com/Azure/azure-functions-host/issues/2153.

cc: @pragnagopa

peterhuene:
Is there any update to this issue?

I notice that my worker process is rudely terminated with `SIGKILL` when gracefully shutting down the host with Ctrl-C.

Unfortunately, the host notices that the process errored and attempts to restart it.  The restarted worker attempts to connect but fails because RPC was shut down and exits with an error.  This causes the host to notice again that the process errored and restarts it a second time.  This could repeat a number of times depending on how fast the rest of the host shuts down.

Example output with host version 2.0.12285.0:

```
[2/12/19 3:31:14 AM] Shuttingdown Rpc Channels Manager
[2/12/19 3:31:14 AM] Shutting down language worker channel for runtime:Rust
[2/12/19 3:31:14 AM] Language Worker Process exited.
[2/12/19 3:31:14 AM] workers/rust/rust_worker exited with code 137
[2/12/19 3:31:14 AM]  .
[2/12/19 3:31:14 AM] Starting language worker process:workers/rust/rust_worker  run --host 127.0.0.1 --port 58568 --workerId 8830d120-0f09-4304-8310-88286e4fc6b8 --requestId 85ef34b2-7879-4049-ba30-ca6863bc05a1 --grpcMaxMessageLength 134217728
[2/12/19 3:31:14 AM] workers/rust/rust_worker process with Id=46248 started
[2/12/19 3:31:14 AM] Connecting to Azure Functions host at 127.0.0.1:58568.
[2/12/19 3:31:14 AM] thread '<unnamed>thread 'main' panicked at '' panicked at 'failed to receive message: RpcFailure(RpcStatus { status: Unavailable, details: Some("Connect Failed") }).failed to send message to host: RpcFinished(Some(RpcStatus { status: Unavailable, details: Some("Connect Failed") }))', ', /Users/peterhuene/src/azure-functions-rs/azure-functions/src/rpc/client.rs:src/libcore/result.rs:1361009::335
[...]

Issue #8366, opened 2022-05-05
Testing for timeout exceptions


surgupta-msft:
We need a mechanism to test timeout exceptions during development time. This is important when we are doing changes related to timeouts, we don't realize if that piece of code will throw any error after timeout is reached (10 minutes or so). In some cases, these timeouts are around 10 minutes which means we need to wait for 10 minutes to see such errors happening. An approach could be to allow error showing up with reduced timeout during integration testing.

**Example scenario -** 
We were getting workerTimeout exceptions due to event being incorrectly configured.
Commit reverted to solve the issue - https://github.com/Azure/azure-functions-host/pull/8346
The issue was that there were 2 subscriptions for 1 GRPC request (incorrectly configured), so the other subscription was getting timed out. 

Also, another time related improvement for unit tests is to have a way to run UTs for a shorter wait time (instead of actual wait time).

For example - Below UT waits for actual wait time (25 seconds), instead we can have a simulated shorter time to quickly run the test
https://github.com/Azure/azure-functions-host/blob/b2ed903d20416157cc8d3730246e642d3e6ed9e5/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs#L147

PR #8385, opened 2022-05-13
Added support to gracefully shutdown language worker

### LINKED ISSUE #2308: Add support to gracefully shutdown language worker
FunctionRpc proto file defines: [WorkerTerminate ](https://github.com/Azure/azure-functions-host/blob/a32a63925c9e67a4ce1af4a49ae1dd3b16d7f7e2/src/WebJobs.Script.Grpc/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto#L160-L162). Need to update host to send WorkerTerminate message with a grace period.

surgupta-msft:

resolves #2308 
[Design doc](https://eng.ms/docs/cloud-ai-platform/devdiv/serverless-paas-balam/serverless-paas-benbyrd/app-service-web-apps/antares-team-documents/functionteamdocs/design/host/gracefulshutdownofworker)

Sending a [WorkerTerminate](https://github.com/Azure/azure-functions-language-worker-protobuf/blob/58b3dc0dbde065b15bd979bd2e27f36b7ee6273e/src/proto/FunctionRpc.proto#L39) message from host to worker to allow the worker to terminate itself gracefully instead of immediately killing it. Refer #3296 for more details.

### Pull request checklist

### Additional information

Additional PR information


src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -873,10 +874,32 @@ protected virtual void Dispose(bool disposing)
 
         public void Dispose()
         {
+            StopWorkerProcessAsync();
```
liliankasem: can you move this under setting `_disposing = true;`

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -873,10 +874,32 @@ protected virtual void Dispose(bool disposing)
...
             _disposing = true;
             Dispose(true);
         }
 
+        public void StopWorkerProcessAsync()
+        {
+            bool capabilityEnabled = !string.IsNullOrEmpty(_workerCapabilities.GetCapabilityState(RpcWorkerConstants.HandlesWorkerTerminateMessage));
+            if (capabilityEnabled)
```
liliankasem: nit: would it be cleaner to inverse this if condition

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -873,10 +874,32 @@ protected virtual void Dispose(bool disposing)
...
+            {
+                var gracePeriod = TimeSpan.FromMilliseconds((double)WorkerConstants.ProcessExitTimeoutInMilliseconds);
```
liliankasem: just to confirm, this value is set by each worker? or set by the host? Just wondering if we should be getting this grace period from each worker config

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SamplesEndToEndTests_Java.cs
```diff
@@ -46,6 +46,7 @@ public async Task JavaProcess_Different_AfterHostRestart()
             // Trigger a restart
             await _fixture.Host.RestartAsync(CancellationToken.None);
             await HttpTrigger_Java_Get_Succeeds();
+            Thread.Sleep(TimeSpan.FromSeconds(2));
```
liliankasem: why do we need to sleep for 2 seconds?

test/WebJobs.Script.Tests/Workers/Rpc/GrpcWorkerChannelTests.cs
```diff
@@ -145,6 +145,48 @@ public async Task StartWorkerProcessAsync_ThrowsTaskCanceledException_IfDisposed
...
+
+            // Send worker init request and enable the capabilities
+            GrpcEvent rpcEvent = new GrpcEvent(_workerId, startStreamMessage);
+            _workerChannel.SendWorkerInitRequest(rpcEvent);
+            _testFunctionRpcService.PublishWorkerInitResponseEvent(capabilities);
+
+            _workerChannel.Dispose();
+            var traces = _logger.GetLogMessages();
+            var expectedLogMsg = $"Sending WorkerTerminate message with grace period {WorkerConstants.ProcessExitTimeoutInMilliseconds}.";
+            Assert.True(traces.Any(m => string.Equals(m.FormattedMessage, expectedLogMsg)));
```
liliankasem, line 176: Is it possible for us to add an E2E test where we check if we actually sent a WorkerTerminate message, and assert the value of `GracePeriod`?

liliankasem (commented):
looking good so far, just a few comments and questions

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -873,10 +874,32 @@ protected virtual void Dispose(bool disposing)
...
+
+                var workerTerminate = new WorkerTerminate()
+                {
+                    GracePeriod = Duration.FromTimeSpan(gracePeriod)
```
kshyju: Can you remind me how will the language workers use this grace period value when they get this GRPC message?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
surgupta-msft: Little unsure of that as `StopWorkerProcessAsync( )` is trying to inform the worker about termination while actual `dispose()` is starting after that. I think we can follow same pattern as `StartWorkerProcessAsync()` where it tells the worker first and then set the state.

https://github.com/Azure/azure-functions-host/blob/9a81e05c957e492c8887a8c2cd180d927f594b35/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs#L155-L156

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
surgupta-msft: Grace period is set by Host. It represents the time Host will wait for worker to exit the process before killing it. I am assigning it the same value which we were already using in the code for this purpose.

https://github.com/Azure/azure-functions-host/blob/9a81e05c957e492c8887a8c2cd180d927f594b35/src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs#L219-L220

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
surgupta-msft: With grace period, we are just telling the worker that Host will wait for some time (grace period) before killing the process. It gives an ability to worker to perform needed cleanup tasks and gracefully shutdown the process within this time limit otherwise Host will kill it after timeout. Previously, Host was directly killing the process without giving any ability to worker to perform graceful shutdown.

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SamplesEndToEndTests_Java.cs
surgupta-msft: Not sure if the concern is about 2 seconds or sleep. I will share more context here - we were discussing the grace period we want to keep and finalized it to be 10 seconds. Updated the code with it. The test is restarting the service so internally host will wait for 10 seconds for process to get exited and then kill the process. Thus, added the delay.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
surgupta-msft: Updated the code here.

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SamplesEndToEndTests_Java.cs
liliankasem: Is there a way to check if the process has exited? Happy for the delay to be the fallback if that is not possible, but generally adding sleeps leads to flakey tests. So if the host has a way of checking if a process has exited, that would be the ideal way to do that

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -873,10 +874,32 @@ protected virtual void Dispose(bool disposing)
 
         public void Dispose()
         {
+            StopWorkerProcessAsync();
             _disposing = true;
             Dispose(true);
         }
 
+        public void StopWorkerProcessAsync()
```
liliankasem: Should this be called `StopWorkerProcess`? I don't think we're awaiting anything or returning a task for it to be async

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -873,10 +874,34 @@ protected virtual void Dispose(bool disposing)
 
         public void Dispose()
         {
+            StopWorkerProcessAsync();
             _disposing = true;
             Dispose(true);
         }
 
+        public async void StopWorkerProcessAsync()
```
kshyju: Should this need to be `public`?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -873,10 +874,34 @@ protected virtual void Dispose(bool disposing)
...
+            };
+
+            _workerChannelLogger.LogDebug($"Sending WorkerTerminate message with grace period {WorkerConstants.WorkerTerminateGracePeriodInSeconds} seconds.");
+
+            SendStreamingMessage(new StreamingMessage
+            {
+                WorkerTerminate = workerTerminate
+            });
+
+            await Task.Delay(TimeSpan.FromSeconds(WorkerConstants.WorkerTerminateGracePeriodInSeconds));
```
kshyju: So we want to delay the `Dispose(true)` method call 10 seconds?, not doing the wait in the `Process.WaitForExit` code path anymore?

test/WebJobs.Script.Tests/Workers/Rpc/GrpcWorkerChannelTests.cs
kshyju: Wonder there is a way for us to use a different wait time for the test runs so that UTs are done faster.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
kshyju: Do we really need to wait for 10 seconds here? The fact that this code path is called form the Dispose method means that some of the components are already shut down/not live any more. In that case,  do we expect the host to process any messages coming from the worker during that 10 seconds timeframe?

kshyju (commented):
Some questions out of curiosity.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
surgupta-msft: Yeah, I thought the same. I just kept it consistent with `StartWorkerProcessAsync()`. Let me know if you feel we should still update it to a private method.

https://github.com/Azure/azure-functions-host/blob/b2ed903d20416157cc8d3730246e642d3e6ed9e5/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs#L147

kshyju: I think we should change the visibility of this method to private since it is being called only from within the same class.

liliankasem (approved):

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
surgupta-msft: WorkerTerminate message will be just to inform the worker about termination. Host will not expect any response coming from worker here. 10 seconds grace period is just to allow the worker to perform any cleanup tasks for graceful termination before Host killing it.

test/WebJobs.Script.Tests/Workers/Rpc/GrpcWorkerChannelTests.cs
surgupta-msft: Yes, that's a good point. On checking code, it seems we will need some changes in the test setup so that we can have a `TestConstants` file which will be passed in the classes. On quick check, found another test which waits for actual time (25 sec) instead of shorter one. https://github.com/Azure/azure-functions-host/blob/5194284aaeecf1734a14e45871930d9cc2c7f813/test/WebJobs.Script.Tests/HttpWorker/DefaultHttpWorkerServiceTests.cs#L481

There is an old issue #8366 , I created for testing timeouts. May be, we can also add this scenario there to cover improvements in tests for time related scenarios.

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SamplesEndToEndTests_Java.cs
surgupta-msft: Changes in this file not needed now.

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -434,7 +434,7 @@ public async Task StopAsync(CancellationToken cancellationToken)
             var currentHost = ActiveHost;
             ActiveHost = null;
             Task stopTask = Orphan(currentHost, cancellationToken);
-            Task result = await Task.WhenAny(stopTask, Task.Delay(TimeSpan.FromSeconds(10), cancellationToken));
```
surgupta-msft, line 437: This is removed in order to handle exception - "Host did not shutdown within allotted time". Because of Cancellation token being signaled by ctrl+c in shutdown path, code is not waiting here for 10 seconds and throwing above warning. Discussed with some folks in team that cancellation token in shutdown path is not needed.

src/WebJobs.Script/Workers/ProcessManagement/JobObjectRegistry.cs
```diff
@@ -8,7 +8,7 @@
 namespace Microsoft.Azure.WebJobs.Script.Workers
 {
     // Registers processes on windows with a job object to ensure disposal after parent exit
-    internal class JobObjectRegistry : IProcessRegistry, IDisposable
```
surgupta-msft: This is removed because JobObjectRegistry is disposing the worker process early (due to race condition). Code is required to wait for few seconds for graceful termination of worker before killing the process.

src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs
```diff
@@ -216,10 +216,10 @@ public void Dispose()
                 {
                     if (!Process.HasExited)
                     {
-                        Process.Kill();
                         if (!Process.WaitForExit(processExitTimeoutInMilliseconds))
                         {
-                            _workerProcessLogger.LogWarning($"Worker process has not exited despite waiting for {processExitTimeoutInMilliseconds} ms");
+                            Process.Kill();
+                            _workerProcessLogger.LogInformation($"Worker process has not exited despite waiting for {processExitTimeoutInMilliseconds} ms");
```
surgupta-msft: Changing log level from _warning_ to _info_ otherwise for all the workers not handling WorkerTerminate, we will see warnings in logs. Not a hard preference though from my end, I can switch it back to warning if it is needed for something else.
@fabiocav, @brettsam let me know your thoughts.

src/WebJobs.Script/Workers/ProcessManagement/JobObjectRegistry.cs
liliankasem: Do we still get to dispose the JobObjectRegistry at some point or we no longer need to?

Edit: I see that you are now disposing this in the WorkerProcess :)

src/WebJobs.Script/Workers/ProcessManagement/EmptyProcessRegistry.cs
```diff
@@ -8,5 +8,9 @@ namespace Microsoft.Azure.WebJobs.Script.Workers
     internal class EmptyProcessRegistry : IProcessRegistry
     {
         public bool Register(Process process) => true;
+
+        public void Close()
```
liliankasem, line 12: Just curious, why do we need an empty process registry?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -883,10 +884,34 @@ protected virtual void Dispose(bool disposing)
...
+            };
+
+            _workerChannelLogger.LogDebug($"Sending WorkerTerminate message with grace period {WorkerConstants.WorkerTerminateGracePeriodInSeconds} seconds.");
+
+            SendStreamingMessage(new StreamingMessage
+            {
+                WorkerTerminate = workerTerminate
+            });
+
+            WorkerProcess.WaitForProcessTermination();
```
liliankasem: Just to confirm: we send a worker terminate message to the worker (if they have the worker terminate capability), then we tell the process to wait 5 seconds (or whatever `WorkerTerminateGracePeriodInSeconds` is) before exiting. And WaitForExit blocks the current thread of execution until the time has elapsed or the process has exited. So by the time we call Dispose here on line 889, we have already waited the grace period or we could still be waiting?

src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs
liliankasem: Looks like this comment is no longer relevant as the code has changed? But I vote for warning still :)

surgupta-msft: Yup this not valid now. Thanks

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
surgupta-msft: By line 889, we would have already completed waiting for grace period.

src/WebJobs.Script/Workers/ProcessManagement/EmptyProcessRegistry.cs
surgupta-msft: saw usage of EmptyProcessRegistry in Tests projects.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -901,10 +902,34 @@ protected virtual void Dispose(bool disposing)
...
         public void Dispose()
         {
+            StopWorkerProcess();
             _disposing = true;
             Dispose(true);
         }
 
+        private void StopWorkerProcess()
+        {
+            bool capabilityEnabled = !string.IsNullOrEmpty(_workerCapabilities.GetCapabilityState(RpcWorkerConstants.HandlesWorkerTerminateMessage));
```
madelinegordon, line 912: Is there a possibility that this could evaluate to not empty but have the value "false"?

liliankasem: I don't think so, my understanding of capabilities is that if it is in the list then we treat it as "true" / something the worker can handle - the value provided doesn't matter.

E.g. if a worker added the capability `HandlesWorkerTerminateMessage = false` to their init response message, we don't actual care about the value provided, we only check that `HandlesWorkerTerminateMessage` [is in the capability list provided by the worker](https://github.com/Azure/azure-functions-host/blob/5a79ab338e0e6ebd961a30b9950fc19ac9645fa0/src/WebJobs.Script.Grpc/Channel/GrpcCapabilities.cs#L23)

src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs
```diff
@@ -53,8 +54,9 @@ internal WorkerProcess(IScriptEventManager eventManager, IProcessRegistry proces
 
         internal Queue<string> ProcessStdErrDataQueue => _processStdErrDataQueue;
 
-        // for testing
-        internal Process Process { get; set; }
+        public Process Process { get; set; }
```
liliankasem, line 57: Why did we end up making this public?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -901,10 +902,34 @@ protected virtual void Dispose(bool disposing)
...
+            };
+
+            _workerChannelLogger.LogDebug($"Sending WorkerTerminate message with grace period {WorkerConstants.WorkerTerminateGracePeriodInSeconds} seconds.");
+
+            SendStreamingMessage(new StreamingMessage
+            {
+                WorkerTerminate = workerTerminate
+            });
+
+            WorkerProcess.WaitForProcessExit();
```
liliankasem: I feel like we should pass in the wait time here so it's clear when reading the code that we are waiting for the same grace period that we are sending in the WorkerTerminate message. It feels like this method should be in control of the time period it chooses rather than whatever is inside WaitForProcessExit().

```suggestion
            WorkerProcess.WaitForProcessExit(WorkerConstants.WorkerTerminateGracePeriodInSeconds);
```

liliankasem (approved):

src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs
surgupta-msft: We are using it in another class JobObjectRegistry to check the status of process termination.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
surgupta-msft: I agree. Updated the code here. Thanks.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -915,19 +915,21 @@ private void StopWorkerProcess()
...
-            _workerChannelLogger.LogDebug($"Sending WorkerTerminate message with grace period {WorkerConstants.WorkerTerminateGracePeriodInSeconds} seconds.");
+            _workerChannelLogger.LogDebug($"Sending WorkerTerminate message with grace period {gracePeriod} seconds.");
 
             SendStreamingMessage(new StreamingMessage
             {
                 WorkerTerminate = workerTerminate
             });
 
-            WorkerProcess.WaitForProcessExit();
+            WorkerProcess.WaitForProcessExitInMilliSeconds(gracePeriod * 1000);
```
liliankasem, line 932: super nit here, but I don't feel we need to change the name, the method definition/summary should be enough to indicate if the parameter it takes is milliseconds or second

satvu (approved):
LGTM!

PR #8435, opened 2022-06-03
Adding a new logger for tooling console log entries from language workers so that they are always logged as Informational

kshyju:
Adding a new logger for tooling console log entries from language workers so that they are always logged as Informational irrespective of Host category level present in HOST.JSON

The dotnet language worker has a startup hook which emits some meta information about the language worker process (ex: processId). This information is emitted via [standard output using Console.WriteLine API](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/src/DotNetWorker.Core/StartupHook.cs#L52). The Host reads this std output stream and log that information using ILogger. We have a console logger which will print this in the console when in development mode, which IDE/Debuggers may read (the processId) so that they can attach to the process.

When the `host.JSON` of dotnet isolated function app specify "Error" as the minimum log level for "Host" category, this information is not surfaced to console as ILogger will discard any log message with level below "Error". The tooling console log messages[ are sent as "Information" level](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs#L127) one and thus being dropped.

The fix here in this PR is having a separate ILogger for the tooling related log messages coming from worker. We identify the tooling related messages by checking the existence of a special prefix(`azfuncjsonlog:`), and if detected, we will use the new logger for logging that message. This log level of this special logger is set to `Informational` in code (Customers/Users are not supposed to change this), so that the log level value of "Host" category is not impacting this log entries.

Relevant issue: https://github.com/Azure/azure-functions-dotnet-worker/issues/900

### DIFF
--- a/src/WebJobs.Script.WebHost/Diagnostics/SystemLoggerProvider.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/SystemLoggerProvider.cs
@@ -49,7 +49,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 
         private bool IsUserLogCategory(string categoryName)
         {
-            return LogCategories.IsFunctionUserCategory(categoryName) || categoryName.Equals(WorkerConstants.FunctionConsoleLogCategoryName, StringComparison.OrdinalIgnoreCase);
+            return LogCategories.IsFunctionUserCategory(categoryName) || categoryName.Equals(WorkerConstants.ConsoleLogCategoryName, StringComparison.OrdinalIgnoreCase);
         }
 
         public void Dispose()
--- a/src/WebJobs.Script/Workers/WorkerConsoleLogService.cs
+++ b/src/WebJobs.Script/Workers/WorkerConsoleLogService.cs
@@ -13,6 +13,7 @@ namespace Microsoft.Azure.WebJobs.Script.Workers
     internal class WorkerConsoleLogService : IHostedService, IDisposable
     {
         private readonly ILogger _logger;
+        private readonly Lazy<ILogger> _toolingConsoleJsonLoggerLazy;
         private readonly IWorkerConsoleLogSource _source;
         private readonly CancellationTokenSource _cts = new CancellationTokenSource();
         private Task _processingTask;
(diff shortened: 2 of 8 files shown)


satvu (approved):

src/WebJobs.Script/Workers/WorkerConsoleLogService.cs
```diff
@@ -27,12 +28,14 @@ public WorkerConsoleLogService(ILoggerFactory loggerFactory, IWorkerConsoleLogSo
 
             _source = consoleLogSource ?? throw new ArgumentNullException(nameof(consoleLogSource));
             _logger = loggerFactory.CreateLogger(WorkerConstants.FunctionConsoleLogCategoryName);
+            _toolingConsoleJsonLoggerLazy = new Lazy<ILogger>(() => loggerFactory.CreateLogger(WorkerConstants.FunctionToolingConsoleJsonCategoryName), isThreadSafe: true);
         }
 
-        internal WorkerConsoleLogService(ILogger logger, IWorkerConsoleLogSource consoleLogSource)
+        internal WorkerConsoleLogService(ILogger logger, Lazy<ILogger> toolingConsoleJsonLoggerLazy, IWorkerConsoleLogSource consoleLogSource)
```
kshyju, line 34: This constructor is being used by UTs to inject test version of dependencies.

src/WebJobs.Script/Workers/WorkerConstants.cs
```diff
@@ -35,6 +35,17 @@ public static class WorkerConstants
         public const string LanguageWorkerConsoleLogPrefix = "LanguageWorkerConsoleLog";
         public const string FunctionConsoleLogCategoryName = "Host.Function.Console";
 
+        /// <summary>
+        /// The log category to be used for logging Tooling log entries emitted from language workers.
+        /// </summary>
+        public const string FunctionToolingConsoleJsonCategoryName = "Host.Function.ToolingConsoleJson";
```
madelinegordon: nit - Thinking it would be more consistent to add "Log" to this var name but that might make the name too long - I am wondering if "Function" in "FunctionConsoleLogCategoryName" is necessary (/what it indicates?) and if that can be removed from both var names so that you'd have "ToolingJsonConsoleLogCategoryName" and "ConsoleLogCategoryName" ?

madelinegordon (approved):

src/WebJobs.Script/Workers/WorkerConstants.cs
liliankasem: I wonder if json has to be in the name, I feel like `Host.Function.ToolingConsoleLog` is more appropriate. Will there be any other kind of tooling log what we would have to differentiate between json and not json?

Agree with Maddie to add Log to the name for consistency

liliankasem (approved):
lgtm, only comment I have is around naming (see above comment)

src/WebJobs.Script/Workers/WorkerConstants.cs
kshyju: Good suggestions. Made the below changes in the latest iteration pushed.

- Renamed the logger category name to "Host.Function.ToolingConsoleLog"
- Renamed the constants to remove the "Functions" prefix.

src/WebJobs.Script.WebHost/Diagnostics/SystemLoggerProvider.cs
```diff
@@ -49,7 +49,7 @@ public ILogger CreateLogger(string categoryName)
 
         private bool IsUserLogCategory(string categoryName)
         {
-            return LogCategories.IsFunctionUserCategory(categoryName) || categoryName.Equals(WorkerConstants.FunctionConsoleLogCategoryName, StringComparison.OrdinalIgnoreCase);
+            return LogCategories.IsFunctionUserCategory(categoryName) || categoryName.Equals(WorkerConstants.ConsoleLogCategoryName, StringComparison.OrdinalIgnoreCase);
```
kshyju, line 52: Reflecting the change of renaming the constant to remove the "Function" prefix.

PR #8449, opened 2022-06-09
[pack] Adding Multi language Support in Azure Function v3.x

ghost:

Azure functions is currently designed to execute functions written in a single language within a single function app. The goal of this feature is to enable Logic App v3 to start more than one language worker within the same function app.

### Pull request checklist

    * [ ] Otherwise: I have added/updated all related diagnostic events and their documentation (Documentation issue linked to PR)


src/WebJobs.Script/Description/Workers/Rpc/MultiLanguageFunctionDescriptorProvider.cs
```diff
@@ -0,0 +1,41 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Azure.WebJobs.Script.Extensibility;
+using Microsoft.Azure.WebJobs.Script.Workers;
+using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Script.Description
+{
+    internal class MultiLanguageFunctionDescriptorProvider : WorkerFunctionDescriptorProvider
```
daviburg, line 15: XML comments on all classes and methods?

src/WebJobs.Script/Description/Workers/Rpc/MultiLanguageFunctionDescriptorProvider.cs
```diff
@@ -0,0 +1,41 @@
...
+            {
+                throw new ArgumentNullException(nameof(functionMetadata));
+            }
+
+            if (!Utility.IsFunctionMetadataLanguageSupportedByWorkerRuntime(functionMetadata, _workerConfig))
+            {
+                return (false, null);
+            }
+
+            return await base.TryCreate(functionMetadata);
```
daviburg, line 50: needs ConfigureAwaitable
Yes I see rest of the repo has not done it :-(

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -143,24 +154,47 @@ internal async Task InitializeJobhostLanguageWorkerChannelAsync(int attemptCount
...
         }
 
+        internal async Task InitializeJobhostLanguageWorkerChannelAsync(int attemptCount, IEnumerable<string> languages)
+        {
+            foreach (string language in languages)
+            {
+                var rpcWorkerChannel = _rpcWorkerChannelFactory.Create(_scriptOptions.RootScriptPath, language, _metricsLogger, attemptCount, _workerConfigs);
+                _jobHostLanguageWorkerChannelManager.AddChannel(rpcWorkerChannel, language);
+                await rpcWorkerChannel.StartWorkerProcessAsync();
+                _logger.LogDebug("Adding jobhost language worker channel for runtime: {language}. workerId:{id}", language, rpcWorkerChannel.Id);
```
daviburg, line 153: How does this work? a format string usually has numerical placeholders. An interpolated string would have the $ start marker.

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -260,12 +296,22 @@ public async Task InitializeAsync(IEnumerable<FunctionMetadata> functions, Cance
...
+                _restartWait = _workerConfigs.Max(wc => wc.CountOptions.ProcessRestartInterval);
+                _shutdownTimeout = _workerConfigs.Max(wc => wc.CountOptions.ProcessShutdownTimeout);
+            }
+            else
+            {
+                _processStartupInterval = workerConfig.CountOptions.ProcessStartupInterval;
+                _restartWait = workerConfig.CountOptions.ProcessRestartInterval;
+                _shutdownTimeout = workerConfig.CountOptions.ProcessShutdownTimeout;
+            }
+            ErrorEventsThreshold = 3 * await _maxProcessCount.Value;
```
daviburg, line 307: Avoid hardcoded magical values in code. Consider defining a named constant.

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -306,6 +352,10 @@ public async Task InitializeAsync(IEnumerable<FunctionMetadata> functions, Cance
                         await InitializeJobhostLanguageWorkerChannelAsync();
                         StartWorkerProcesses(1, InitializeJobhostLanguageWorkerChannelAsync);
                     }
+                    else if (_environment.IsMultiLanguageRuntimeEnvironment())
+                    {
+                        StartWorkerProcesses(startIndex: 0, startAction: InitializeJobhostLanguageWorkerChannelAsync, functionLanguages: functions.Select(f => f.Language).Distinct());
```
daviburg: avoid uncommon abbreviation. Use fully formed English words for variable names.

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -426,17 +477,19 @@ internal async Task<IEnumerable<IRpcWorkerChannel>> GetAllWorkerChannelsAsync()
...
         {
-            IEnumerable<IRpcWorkerChannel> workerChannels = await GetAllWorkerChannelsAsync();
+            language ??= _workerRuntime;
+            IEnumerable<IRpcWorkerChannel> workerChannels = await GetAllWorkerChannelsAsync(language);
             IEnumerable<IRpcWorkerChannel> initializedWorkers = workerChannels.Where(ch => ch.IsChannelReadyForInvocations());
-            int workerCount = await _maxProcessCount.Value;
 
-            if (initializedWorkers.Count() > workerCount)
+            int workerCount = await _maxProcessCount.Value;
+            // In case of multi language environment, we are not aware of workers we need until we get function metadata. Thus skipping this check
```
daviburg: blank line before comment line

src/WebJobs.Script/Workers/Rpc/JobHostRpcWorkerChannelManager.cs
```diff
@@ -12,45 +13,86 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
     internal class JobHostRpcWorkerChannelManager : IJobHostRpcWorkerChannelManager
     {
         private readonly ILogger _logger;
-        private ConcurrentDictionary<string, IRpcWorkerChannel> _channels = new ConcurrentDictionary<string, IRpcWorkerChannel>();
+        private ConcurrentDictionary<string, ConcurrentDictionary<string, IRpcWorkerChannel>> _channels = new ConcurrentDictionary<string, ConcurrentDictionary<string, IRpcWorkerChannel>>();
```
daviburg: That's a complex nested generic class. Should it be a declared explicit class instead?

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/MultiLanguageEndToEndTests.cs
```diff
@@ -0,0 +1,220 @@
...
+
+                    var javascriptFunctionKey = await host.GetFunctionSecretAsync("JavascriptFunction");
+                    using var javascriptHttpTriggerResponse = await host.HttpClient.GetAsync($"http://localhost/api/JavascriptFunction?name=Azure&code={javascriptFunctionKey}");
+                    Assert.Equal(HttpStatusCode.OK, javascriptHttpTriggerResponse.StatusCode);
+                    Assert.Equal("Hello Azure", await javascriptHttpTriggerResponse.Content.ReadAsStringAsync());
+                }
+            }
+        }
+
+
```
daviburg: duplicate empty line

test/WebJobs.Script.Tests/Workers/Rpc/JobHostRpcWorkerChannelManagerTests.cs
```diff
@@ -0,0 +1,121 @@
...
+
+        [Fact]
+        public void TestAddChannel()
+        {
+            _jobHostRpcWorkerChannelManager.ShutdownChannels();
+            _jobHostRpcWorkerChannelManager.AddChannel(_workerChannelJava1.Object, "java");
+            _jobHostRpcWorkerChannelManager.AddChannel(_workerChannelPs1.Object, "powershell");
+            _jobHostRpcWorkerChannelManager.AddChannel(_workerChannelJs1.Object, "node");
+
+            Assert.Equal(_jobHostRpcWorkerChannelManager.GetChannels().Count(), 3);
```
daviburg: use named parameters due to the ambiguity in same type arguments order.

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/MultiLanguageEndToEndTests.cs
```diff
@@ -0,0 +1,220 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Azure.WebJobs.Script.Description;
```
daviburg: what's the using ordering rule for this repo? I see a mix of different ordering logics in different files for this PR.

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
ghost: It's logged this way in whole function's code. They are using [`MSLogger`](https://docs.microsoft.com/en-us/dotnet/core/extensions/logging?tabs=command-line#create-logs)

src/WebJobs.Script/Description/Workers/Rpc/MultiLanguageFunctionDescriptorProvider.cs
ghost: Adding XML Doc to all the new classes that i create. To keep the XML Doc consistent in a class.

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
siddharth-ms: Which abbreviation?

test/WebJobs.Script.Tests/Workers/Rpc/JobHostRpcWorkerChannelManagerTests.cs
ghost: Not followed anywhere in function code, will wait for owners response on this

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/MultiLanguageEndToEndTests.cs
ghost: Not sure, I observe different `stylecop` for unit & integration tests. Waiting for owners comments on this!

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
ghost: It was used before too, will wait for owners to comment on why 3?

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -109,6 +109,10 @@ internal class RpcFunctionInvocationDispatcher : IFunctionInvocationDispatcher
 
         private async Task<int> GetMaxProcessCount()
         {
+            if (_environment.IsMultiLanguageRuntimeEnvironment())
+            {
+                return 1;
```
liliankasem: If this isn't something we can get from the worker config, please define a constant for this value

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -143,24 +154,47 @@ internal async Task InitializeJobhostLanguageWorkerChannelAsync(int attemptCount
...
+                _logger.LogDebug("Adding jobhost language worker channel for runtime: {language}. workerId:{id}", language, rpcWorkerChannel.Id);
+
+                // if the worker is indexing, we will not have function metadata yet so we cannot perform these next three lines
+                if (!_workerIndexing)
+                {
+                    rpcWorkerChannel.SetupFunctionInvocationBuffers(_functions);
+                    rpcWorkerChannel.SendFunctionLoadRequests(_managedDependencyOptions.Value, _scriptOptions.FunctionTimeout);
+                }
+            }
+            SetFunctionDispatcherStateToInitializedAndLog();
```
liliankasem: @surgupta-msft In the other `InitializeJobhostLanguageWorkerChannelAsync` method, is there a reason`SetFunctionDispatcherStateToInitializedAndLog` is happening inside the `if (!_workerIndexing)` branch and not on the outside like here?

Just wondering why the two methods would handle this differently

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -122,17 +126,24 @@ private async Task<int> GetMaxProcessCount()
             return (await GetAllWorkerChannelsAsync()).Count();
         }
 
-        internal async Task InitializeJobhostLanguageWorkerChannelAsync()
+        internal async Task InitializeJobhostLanguageWorkerChannelAsync(IEnumerable<string> languages = null)
         {
-            await InitializeJobhostLanguageWorkerChannelAsync(0);
+            if (languages == null)
```
praveensri, line 133: Do you need to check if languages is not empty?

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -122,17 +126,24 @@ private async Task<int> GetMaxProcessCount()
...
+                await InitializeJobhostLanguageWorkerChannelAsync(0, _workerRuntime);
+            }
+            else
+            {
+                await InitializeJobhostLanguageWorkerChannelAsync(0, languages);
+            }
         }
 
-        internal async Task InitializeJobhostLanguageWorkerChannelAsync(int attemptCount)
+        internal async Task InitializeJobhostLanguageWorkerChannelAsync(int attemptCount, string language)
```
praveensri: can you add description summary of this method?

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
praveensri: do we need check here if Lang metadata is used before making the max process count for Lang worker to 1. Just a thought?

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
ghost: Good catch! I think i should move it inside the condition. Will fix it in next iteration.

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
ghost: Skipping this, only adding doc for new files i am creating.

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
ghost: That will never happen since its not user input, but i think its still good idea to check for empty also!

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
ghost: This is like a configuration value which is set at constructor level, Thus, it's not really possible to make a lot of decisions.

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -143,24 +154,47 @@ internal async Task InitializeJobhostLanguageWorkerChannelAsync(int attemptCount
             }
         }
 
+        internal async Task InitializeJobhostLanguageWorkerChannelAsync(int attemptCount, IEnumerable<string> languages)
```
surgupta-msft, line 146: Should the method name have _channels_ instead of _channel_ as it is dealing with multiple languages and channels?

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -143,24 +154,47 @@ internal async Task InitializeJobhostLanguageWorkerChannelAsync(int attemptCount
...
-        internal async Task InitializeWebhostLanguageWorkerChannel()
+        internal async Task InitializeWebhostLanguageWorkerChannel(IEnumerable<string> languages = null)
         {
-            _logger.LogDebug("Creating new webhost language worker channel for runtime:{workerRuntime}.", _workerRuntime);
-            IRpcWorkerChannel workerChannel = await _webHostLanguageWorkerChannelManager.InitializeChannelAsync(_workerRuntime);
-
-            // if the worker is indexing, we will not have function metadata yet so we cannot perform the next two lines
-            if (!_workerIndexing)
+            languages ??= new[] { _workerRuntime };
+            foreach (string language in languages)
```
surgupta-msft, line 178: The variable `languages` will always have only 1 item - `_workerRuntime`. Curious why we need to add `_workerRuntime` to `languages` and do a loop here?

src/WebJobs.Script/Workers/Rpc/IRpcWorkerChannelHolder.cs
```diff
@@ -0,0 +1,14 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Collections.Concurrent;
+
+namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
+{
+    /// <summary>
+    /// Class to hold list of <see cref="IRpcWorkerChannel"/>
+    /// </summary>
+    internal class IRpcWorkerChannelHolder : ConcurrentDictionary<string, IRpcWorkerChannel>
```
surgupta-msft: nit: should we name this class IRpcWorkerChannelDictionary?

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
brettsam: Specifically, it's using structured logging: https://docs.microsoft.com/en-us/dotnet/core/extensions/logging?tabs=command-line#log-message-template

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -143,24 +154,47 @@ internal async Task InitializeJobhostLanguageWorkerChannelAsync(int attemptCount
             }
         }
 
+        internal async Task InitializeJobhostLanguageWorkerChannelAsync(int attemptCount, IEnumerable<string> languages)
+        {
+            foreach (string language in languages)
+            {
```
brettsam, line 149: Is everything in this `for` loop identical to `InitializeJobhostLanguageWorkerChannelAsync(int attemptCount, string language)`? Can we just call that inside this loop to ensure the logic is the same for both cases? 

something like...
```
foreach (string language in languages)
{
    await InitializeJobhostLanguageWorkerChannelAsync(0, language)
}
```

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
brettsam: Yeah, this has been around for a long time -- don't worry about updating it here (just to keep the changes minimal).

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -260,12 +296,22 @@ public async Task InitializeAsync(IEnumerable<FunctionMetadata> functions, Cance
             }
 
             _functions = functions ?? new List<FunctionMetadata>();
-            _processStartupInterval = workerConfig.CountOptions.ProcessStartupInterval;
-            _restartWait = workerConfig.CountOptions.ProcessRestartInterval;
-            _shutdownTimeout = workerConfig.CountOptions.ProcessShutdownTimeout;
-            ErrorEventsThreshold = 3 * (await _maxProcessCount.Value);
 
-            if (Utility.IsSupportedRuntime(_workerRuntime, _workerConfigs))
+            if (_environment.IsMultiLanguageRuntimeEnvironment())
```
brettsam, line 295: Can we cache this in something like a private `_isMultiLanguageRuntimeEnvironment` field in the constructor? While not super-expensive, it does call `ToLower()` and `Contains()` every time it's called... would be better to cache that if it's being re-checked a lot of places in here.

src/WebJobs.Script/Workers/Rpc/IJobHostRpcWorkerChannelManager.cs
```diff
@@ -9,12 +9,14 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
...
-        void AddChannel(IRpcWorkerChannel channel);
+        void AddChannel(IRpcWorkerChannel channel, string language);
 
         Task<bool> ShutdownChannelIfExistsAsync(string channelId, Exception workerException);
 
         void ShutdownChannels();
 
+        IEnumerable<IRpcWorkerChannel> GetChannels(string language);
+
         IEnumerable<IRpcWorkerChannel> GetChannels();
```
brettsam, line 20: Can we get rid of `GetChannels()`?

src/WebJobs.Script/Workers/Rpc/JobHostRpcWorkerChannelManager.cs
```diff
@@ -12,45 +13,86 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
...
+            if (_channels.TryGetValue(language, out IRpcWorkerChannelHolder channels))
+            {
+                channels.TryAdd(channel.Id, channel);
+            }
+            else
+            {
+                _channels.TryAdd(language, new IRpcWorkerChannelHolder
+                {
+                    [channel.Id] = channel,
+                });
```
brettsam: I don't think this would be thread-safe -- for example if a bunch of threads all `AddChannel("node", ...)` at once, you could end up missing some of those as they all raced to call `_channels.TryAdd("node"...)`. I think we need a lock around this if it's being called multi-threaded anywhere.

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -234,7 +268,8 @@ public async Task InitializeAsync(IEnumerable<FunctionMetadata> functions, Cance
 
             _workerRuntime = _workerRuntime ?? Utility.GetWorkerRuntime(functions, _environment);
 
-            if (string.IsNullOrEmpty(_workerRuntime) || _workerRuntime.Equals(RpcWorkerConstants.DotNetLanguageWorkerName, StringComparison.InvariantCultureIgnoreCase))
+            // In case of multi language runtime, _workerRuntime has no significance, thus skipping this check for multi language runtime environment
+            if ((string.IsNullOrEmpty(_workerRuntime) || _workerRuntime.Equals(RpcWorkerConstants.DotNetLanguageWorkerName, StringComparison.InvariantCultureIgnoreCase)) && !_environment.IsMultiLanguageRuntimeEnvironment())
```
surgupta-msft, line 267: Curious, how are we handling the case where one of the languages in multi-lang is Dotnet? As I am seeing different handling for Dotnet here.

src/WebJobs.Script/Workers/Rpc/JobHostRpcWorkerChannelManager.cs
```diff
@@ -12,45 +13,86 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
...
             return Task.FromResult(false);
         }
 
         public void ShutdownChannels()
         {
-            foreach (string channelId in _channels.Keys)
+            foreach (string language in _channels.Keys)
             {
-                if (_channels.TryRemove(channelId, out IRpcWorkerChannel removedChannel))
+                if (_channels.TryRemove(language, out IRpcWorkerChannelHolder channels))
```
surgupta-msft: very nit: renaming language to languageChannelId and Channels to RemovedChannels

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
liliankasem: I think this is because its an overload of `InitializeJobhostLanguageWorkerChannelAsync`

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
liliankasem: If that's the case, we can probably move that foreach up to line 137 to avoid an additional method overload

src/WebJobs.Script/Workers/Rpc/IJobHostRpcWorkerChannelManager.cs
surgupta-msft: I think `GetChannels()` is returning _all channels_ of all languages, so if we can't get rid of it then probably renaming it to `GetAllChannels()` and the other one to `GetLanguageChannel()`

src/WebJobs.Script/Utility.cs
```diff
@@ -673,9 +673,15 @@ internal static bool IsFunctionMetadataLanguageSupportedByWorkerRuntime(Function
             {
                 return true;
             }
+
             return !string.IsNullOrEmpty(functionMetadata.Language) && functionMetadata.Language.Equals(workerRuntime, StringComparison.OrdinalIgnoreCase);
         }
 
+        internal static bool IsFunctionMetadataLanguageSupportedByWorkerRuntime(FunctionMetadata functionMetadata, IList<RpcWorkerConfig> workerConfig)
```
madelinegordon: nit -  for consistency with `CanWorkerIndex(IEnumerable<RpcWorkerConfig> workerConfigs,...` method params, could change param name to `workerConfigs`

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -143,24 +154,47 @@ internal async Task InitializeJobhostLanguageWorkerChannelAsync(int attemptCount
...
+        internal async Task InitializeJobhostLanguageWorkerChannelAsync(int attemptCount, IEnumerable<string> languages)
+        {
+            foreach (string language in languages)
+            {
+                var rpcWorkerChannel = _rpcWorkerChannelFactory.Create(_scriptOptions.RootScriptPath, language, _metricsLogger, attemptCount, _workerConfigs);
+                _jobHostLanguageWorkerChannelManager.AddChannel(rpcWorkerChannel, language);
+                await rpcWorkerChannel.StartWorkerProcessAsync();
+                _logger.LogDebug("Adding jobhost language worker channel for runtime: {language}. workerId:{id}", language, rpcWorkerChannel.Id);
+
+                // if the worker is indexing, we will not have function metadata yet so we cannot perform these next three lines
```
madelinegordon: nit - in case the content of the conditional changes in the future, could change comment from "cannot perform these next three lines" to something like "cannot set up invocation buffers or send load requests".

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -143,24 +154,47 @@ internal async Task InitializeJobhostLanguageWorkerChannelAsync(int attemptCount
...
+            languages ??= new[] { _workerRuntime };
+            foreach (string language in languages)
             {
-                workerChannel.SetupFunctionInvocationBuffers(_functions);
-                workerChannel.SendFunctionLoadRequests(_managedDependencyOptions.Value, _scriptOptions.FunctionTimeout);
-                SetFunctionDispatcherStateToInitializedAndLog();
+                _logger.LogDebug("Creating new webhost language worker channel for runtime:{workerRuntime}.", language);
+                IRpcWorkerChannel workerChannel = await _webHostLanguageWorkerChannelManager.InitializeChannelAsync(language);
+
+                // if the worker is indexing, we will not have function metadata yet so we cannot perform the next two lines
```
madelinegordon: nit - same as above regarding "next two lines" in the comment

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/MultiLanguageEndToEndTests.cs
```diff
@@ -0,0 +1,260 @@
...
+                    var javaFunctionKey = await host.GetFunctionSecretAsync("JavaFunction");
+                    using var javaHttpTriggerResponse = await host.HttpClient.GetAsync($"http://localhost/api/JavaFunction?code={javaFunctionKey}");
+                    Assert.Equal(HttpStatusCode.BadRequest, javaHttpTriggerResponse.StatusCode);
+                    Assert.Equal("Please pass a name on the query string or in the request body", await javaHttpTriggerResponse.Content.ReadAsStringAsync());
+                }
+            }
+        }
+
+        /// <summary>
+        /// Runs tests with Node language provider funtion.
```
madelinegordon: nit - "function"

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/MultiLanguageEndToEndTests.cs
```diff
@@ -0,0 +1,260 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Azure.WebJobs.Script.Description;
+using Microsoft.Azure.WebJobs.Script.WebHost.Management;
+using Microsoft.Extensions.DependencyInjection;
+using Moq;
+using System;
```
madelinegordon, line 4: nit  - ordering of usings

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/MultiLanguageEndToEndTests.cs
madelinegordon: `System` using's on the top!

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
ghost: True

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
ghost: Not necessarily, we since we have added multiple language support, they can be multiple, but since languages has a default and is not provided we fallback to the default scenario of starting `_workerRuntime`.

src/WebJobs.Script/Workers/Rpc/IJobHostRpcWorkerChannelManager.cs
ghost: `GetChannels` previously used to return all channels, i added a overloaded function with language to avoid changes in places where `GetChannels` is used. It's still used in some cases.

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
ghost: Dotnet is running as in-proc function and does not create and workers. I have added test to call C#, Java, Node function. Let me know if i am missing something?

src/WebJobs.Script/Workers/Rpc/JobHostRpcWorkerChannelManager.cs
ghost: `languageChannelId` doesn't makes sense here, it's the language only and can have values like "java", "node", etc.  Renaming `removedChannels`.

src/WebJobs.Script/Workers/Rpc/JobHostRpcWorkerChannelManager.cs
ghost: We were using `ConcurrentDictionary` before too. Currently we are creating single worker per language in sequence. What do you suggest here, should we have locks/waits here?

ghost: Looks like the issue would be introduced in case of language worker handling and might not have existed before, added a synchronized block for this. Also added a test case to test the concurrency.

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
ghost: Actually there can be cases where languages can be empty, Lets say in case of no language worker is needed. Thus not making this change.

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -23,6 +23,8 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
 {
     internal class RpcFunctionInvocationDispatcher : IFunctionInvocationDispatcher
     {
+        private static readonly int MultiLanguageDefaultPrcocessCount = 1;
```
soninaren: Nit: Spelling error

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -299,14 +301,26 @@ public static string GetAntaresComputerName(this IEnvironment environment)
...
             string appKind = environment.GetEnvironmentVariable(AppKind)?.ToLower();
             return !string.IsNullOrEmpty(appKind) && appKind.Contains(ScriptConstants.WorkFlowAppKind);
         }
 
+        /// <summary>
+        /// Gets if runtime environment needs multi language
+        /// </summary>
+        public static bool IsMultiLanguageRuntimeEnvironment(this IEnvironment environment, bool refreshCache = false)
+        {
+            if (refreshCache)
```
soninaren: Relying on the ScriptHost to set its initial value seems a little fragile, may want to use an additional flag to set its initial value in case something else calls it first.

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
soninaren: Just an observation here, while there should not be any standbyChannels, running the shutdown code is not particularly harmful.

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -306,6 +347,10 @@ public async Task InitializeAsync(IEnumerable<FunctionMetadata> functions, Cance
                         await InitializeJobhostLanguageWorkerChannelAsync();
                         StartWorkerProcesses(1, InitializeJobhostLanguageWorkerChannelAsync);
                     }
+                    else if (_environment.IsMultiLanguageRuntimeEnvironment())
+                    {
+                        StartWorkerProcesses(startIndex: 0, startAction: InitializeJobhostLanguageWorkerChannelAsync, functionLanguages: functions.Select(function => function.Language).Distinct());
```
soninaren: nit: saving the function list in a separate variable might be easier to read.

src/WebJobs.Script/Workers/Rpc/JobHostRpcWorkerChannelManager.cs
```diff
@@ -12,45 +13,89 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
...
         public JobHostRpcWorkerChannelManager(ILoggerFactory loggerFactory)
         {
             _logger = loggerFactory.CreateLogger<JobHostRpcWorkerChannelManager>();
         }
 
-        public void AddChannel(IRpcWorkerChannel channel)
+        public void AddChannel(IRpcWorkerChannel channel, string language)
         {
-            _channels.TryAdd(channel.Id, channel);
+            lock (_channels)
```
soninaren: Might want to use a separate object for the lock

soninaren (changes requested):
**[Resolved]** Given the number of changes to the critical components. I would recommend putting this feature behind a feature flag. This would reduce the bar for PR merge and also would allow us to introduce breaking changes if needed.

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
ghost: Tried to use `bool?` and initialize it if its `null` and use it if `not null`. But the problem is the test cases fail on cloudtest (doesn't fail locally though). I am assuming that the `bool?` flag is keeping its value from last run even though new test is running.
Any suggestion on this?

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
ghost: I suppressed it because in case of Multi Language, `_workerRuntime` is alwys `""`

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
ghost: I can move the code to constructor of `SystemEnvironment` but not sure how good that is?
**Note: Removed Cache for now as it's not a huge performance hit.**

soninaren (dismissed):

src/WebJobs.Script/Description/Workers/Rpc/MultiLanguageFunctionDescriptorProvider.cs
```diff
@@ -0,0 +1,53 @@
...
+        /// <param name="bindingProviders">List of <see cref="IScriptBindingProvider"/> instances.</param>
+        /// <param name="dispatcher"><see cref="IFunctionInvocationDispatcher"/> instance.</param>
+        /// <param name="loggerFactory"><see cref="ILoggerFactory"/> instance.</param>
+        /// <param name="applicationLifetime"><see cref="IApplicationLifetime"/> instance.</param>
+        /// <param name="workerInitializationTimeout">Worker initialization timeout.</param>
+        public MultiLanguageFunctionDescriptorProvider(ScriptHost host, IList<RpcWorkerConfig> workerConfig, ScriptJobHostOptions config, ICollection<IScriptBindingProvider> bindingProviders,
+            IFunctionInvocationDispatcher dispatcher, ILoggerFactory loggerFactory, IApplicationLifetime applicationLifetime, TimeSpan workerInitializationTimeout)
+            : base(host, config, bindingProviders, dispatcher, loggerFactory, applicationLifetime, workerInitializationTimeout)
+        {
+            _workerConfig = workerConfig;
```
liliankasem:
```suggestion
            _workerConfig = workerConfig ?? throw new ArgumentNullException(nameof(workerConfig));
```

src/WebJobs.Script/Utility.cs
```diff
@@ -906,9 +912,9 @@ public static bool IsMediaTypeOctetOrMultipart(MediaTypeHeaderValue mediaType)
 
         public static bool CanWorkerIndex(IEnumerable<RpcWorkerConfig> workerConfigs, IEnvironment environment)
         {
-            var workerRuntime = environment.GetEnvironmentVariable(EnvironmentSettingNames.FunctionWorkerRuntime);
-            if (workerConfigs != null)
+            if (workerConfigs != null && !environment.IsMultiLanguageRuntimeEnvironment())
```
liliankasem, line 915: Wondering if we need to cache `IsMultiLanguageRuntimeEnvironment()` at the environment extensions level since it's getting used to much

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -245,7 +275,8 @@ public async Task InitializeAsync(IEnumerable<FunctionMetadata> functions, Cance
 
             var workerConfig = _workerConfigs.Where(c => c.Description.Language.Equals(_workerRuntime, StringComparison.InvariantCultureIgnoreCase)).FirstOrDefault();
 
-            if (workerConfig == null && (functions == null || functions.Count() == 0))
+            // We are skipping this check for multi-language environments because they use multiple workers and thus doesn't honor 'FUNCTIONS_WORKER_RUNTIME'
+            if (workerConfig == null && (functions == null || functions.Count() == 0) && !_environment.IsMultiLanguageRuntimeEnvironment())
```
liliankasem, line 279: It doesn't look like `IsMultiLanguageRuntimeEnvironment` is being cached yet as per Brett's comment which was resolved

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -426,17 +473,19 @@ internal async Task<IEnumerable<IRpcWorkerChannel>> GetAllWorkerChannelsAsync()
...
+        internal async Task<IEnumerable<IRpcWorkerChannel>> GetInitializedWorkerChannelsAsync(string language = null)
         {
-            IEnumerable<IRpcWorkerChannel> workerChannels = await GetAllWorkerChannelsAsync();
+            language ??= _workerRuntime;
+            IEnumerable<IRpcWorkerChannel> workerChannels = await GetAllWorkerChannelsAsync(language);
             IEnumerable<IRpcWorkerChannel> initializedWorkers = workerChannels.Where(ch => ch.IsChannelReadyForInvocations());
             int workerCount = await _maxProcessCount.Value;
 
-            if (initializedWorkers.Count() > workerCount)
+            // In case of multi language environment, we are not aware of workers we need until we get function metadata. Thus skipping this check
```
liliankasem: Will there be a similar check once you are aware of how many workers you need?

src/WebJobs.Script/Workers/Rpc/JobHostRpcWorkerChannelManager.cs
```diff
@@ -4,53 +4,99 @@
...
+            {
+                return channels.Values;
+            }
+            return Enumerable.Empty<IRpcWorkerChannel>();
+        }
+
         public IEnumerable<IRpcWorkerChannel> GetChannels()
         {
-            return _channels.Values;
+            List<IRpcWorkerChannel> rpcWorkerChannels = new List<IRpcWorkerChannel>();
```
liliankasem:
```suggestion
            var rpcWorkerChannels = new List<IRpcWorkerChannel>();
```

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
ghost: Having a cache in static variable was causing issues while running tests in `CloudTest`. Removed the caching because most of these checks are at host start and string operations might not be very expensive given the size of `APP_KIND/FEATURE_FLAG` is not very huge. Here is a revert of that change for your reference - [cc66cc](https://github.com/Azure/azure-functions-host/pull/8449/commits/cc66cc10e644fd57d867dd0c59004faf908abd95)
What do you think?

Screenshot 2022-07-13 054830

src/WebJobs.Script/Utility.cs
ghost: Closing this and tracking this in comment below!

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
ghost: Currently there is no such check, is that needed? because in case of multi language we hard code the worker process count to be 1 thus no more than X(number of languages) should be instantiated.

ghost: My Bad looks like this code change is not needed. We can still have it in case of multi language. Removing diff.

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
ghost: One thing which can be done maybe is that, cache it twice at [ScriptHost.cs](https://github.com/Azure/azure-functions-host/pull/8449/files#diff-6d3165a3c204458bede6be5209ce39874c572862494969cb364f31e0d1c7d1b3)[ScriptHost.cs](https://github.com/Azure/azure-functions-host/pull/8449/files#diff-6d3165a3c204458bede6be5209ce39874c572862494969cb364f31e0d1c7d1b3)  and [RpcFunctionInvocationDispatcher.cs](https://github.com/Azure/azure-functions-host/pull/8449/files/cc66cc10e644fd57d867dd0c59004faf908abd95#diff-5a241b8ff64ccf9edb3fc0cc63256d99705ba2ffa011019fa8371d16d86b6cd6) level. Should we cache it twice?

ghost: Issue is resolved now!

soninaren (approved):

liliankasem (approved):

daviburg (approved):

PR #8481, opened 2022-06-16
Parse and log WorkerMetadata information

### LINKED ISSUE #8452: Log WorkerMetadata details from the WorkerInitResponse message
The protobuf has been updated to include a `WorkerMetadata` property within the `WorkerInitResponse` message. 
The host should log the `WorkerMetadata` information when we handle a `WorkerInitResponse`.

[PR](https://github.com/Azure/azure-functions-language-worker-protobuf/pull/76) for reference on how the fields look.

You will need to pull in the latest protobuf version first.

```proto
message WorkerMetadata {
  // The runtime/stack name
  string runtime_name = 1;

  // The version of the runtime/stack
  string runtime_version = 2;

  // The version of the worker or worker SDK
  string worker_version = 3;

  // The worker bitness/architecture
  string worker_bitness = 4;

  // Optional additional custom properties
  map<string, string> custom_properties = 5;
}
```

liliankasem:
Logging WorkerMetadata information as a debug log and a metric event. Defaulting to information from worker config for language and runtime version

resolves #8452

### Pull request checklist

### Testing

Metric shows up in FunctionMetrics table:

Screen Shot 2022-06-17 at 3 00 20 PM

Log shows up in FunctionLogs table:

Screen Shot 2022-06-17 at 2 59 27 PM

### DIFF
--- a/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
@@ -258,9 +258,10 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
 
             if (_initMessage.WorkerMetadata != null)
             {
-                _initMessage.WorkerMetadata = ParseWorkerMetadata(_initMessage.WorkerMetadata);
-                _metricsLogger.LogEvent(MetricEventNames.WorkerMetadata, functionName: null, _initMessage.WorkerMetadata.ToString());
-                _workerChannelLogger.LogDebug($"Worker metadata: {_initMessage.WorkerMetadata}");
+                _initMessage.ParseWorkerMetadata(_workerConfig);
+                var workerMetadata = _initMessage.WorkerMetadata.ToString();
+                _metricsLogger.LogEvent(MetricEventNames.WorkerMetadata, functionName: null, workerMetadata);
+                _workerChannelLogger.LogDebug($"Worker metadata: {workerMetadata}");
             }
 
             if (_initMessage.Result.IsFailure(out Exception exc))
@@ -281,15 +282,6 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
             }
 
             _workerInitTask.SetResult(true);
-        }
-
-        private WorkerMetadata ParseWorkerMetadata(WorkerMetadata workerMetadata)
-        {
-            workerMetadata.RuntimeName = string.IsNullOrEmpty(workerMetadata.RuntimeName)
-                                            ? _workerConfig.Description.Language : workerMetadata.RuntimeName;
-            workerMetadata.RuntimeVersion = string.IsNullOrEmpty(workerMetadata.RuntimeVersion)
-                                            ? _workerConfig.Description.DefaultRuntimeVersion : workerMetadata.RuntimeVersion;
-            return workerMetadata;
         }
 
         public void SetupFunctionInvocationBuffers(IEnumerable<FunctionMetadata> functions)
--- a/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
@@ -254,13 +254,9 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
 
             _workerChannelLogger.LogDebug("Received WorkerInitResponse. Worker process initialized");
             _initMessage = initEvent.Message.WorkerInitResponse;
+            _workerChannelLogger.LogDebug($"Worker capabilities: {_initMessage.Capabilities}");
 
             ParseAndLogWorkerMetadata();
-
-            using (_workerChannelLogger.BeginScope(_initMessage.WorkerMetadata))
-            {
-                _workerChannelLogger.LogDebug($"Worker capabilities: {_initMessage.Capabilities}");
-            }
 
             if (_initMessage.Result.IsFailure(out Exception exc))
             {
@@ -282,7 +278,7 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
             _workerInitTask.SetResult(true);
         }
 
-        internal void ParseAndLogWorkerMetadata()
+        private void ParseAndLogWorkerMetadata()
         {
             if (_initMessage?.WorkerMetadata == null)
             {
@@ -294,7 +290,8 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
             _initMessage.WorkerMetadata.RuntimeVersion = string.IsNullOrEmpty(_initMessage.WorkerMetadata.RuntimeVersion)
                                                         ? _workerConfig.Description.DefaultRuntimeVersion : _initMessage.WorkerMetadata.RuntimeVersion;
 
-            _metricsLogger.LogEvent(RpcWorkerConstants.WorkerMetadataMetricEvent, null, _initMessage.WorkerMetadata.ToString());
+            _metricsLogger.LogEvent(MetricEventNames.WorkerMetadata, null, _initMessage.WorkerMetadata.ToString());
+            _workerChannelLogger.LogDebug($"Worker metadata: {_initMessage.WorkerMetadata}");
         }
 
         public void SetupFunctionInvocationBuffers(IEnumerable<FunctionMetadata> functions)
--- a/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
@@ -256,7 +256,12 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
             _initMessage = initEvent.Message.WorkerInitResponse;
             _workerChannelLogger.LogDebug($"Worker capabilities: {_initMessage.Capabilities}");
 
-            ParseAndLogWorkerMetadata();
+            if (_initMessage.WorkerMetadata != null)
+            {
+                _initMessage.WorkerMetadata = ParseWorkerMetadata(_initMessage.WorkerMetadata);
+                _metricsLogger.LogEvent(MetricEventNames.WorkerMetadata, functionName: null, _initMessage.WorkerMetadata.ToString());
+                _workerChannelLogger.LogDebug($"Worker metadata: {_initMessage.WorkerMetadata}");
+            }
 
             if (_initMessage.Result.IsFailure(out Exception exc))
             {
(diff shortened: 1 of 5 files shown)


src/WebJobs.Script/Workers/Rpc/RpcWorkerConstants.cs
```diff
@@ -63,5 +63,8 @@ public static class RpcWorkerConstants
         public const string PythonThreadpoolThreadCount = "PYTHON_THREADPOOL_THREAD_COUNT";
         public const string PSWorkerInProcConcurrencyUpperBound = "PSWorkerInProcConcurrencyUpperBound";
         public const string DefaultConcurrencyLimit = "1000";
+
+        // Metric events
+        public const string WorkerMetadataMetricEvent = "WorkerMetadata";
```
liliankasem, line 68: Is there a style guide we follow for event names? would something like `host.workerinitresponse.workermetadata` be a more fitting event name?

liliankasem: For reference, here are what some of the events look like in the table:

Screen Shot 2022-06-17 at 11 34 39 AM

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -274,6 +278,22 @@ internal void WorkerInitResponse(GrpcEvent initEvent)
             _workerInitTask.SetResult(true);
         }
 
+        private void ParseAndLogWorkerMetadata()
+        {
+            if (_initMessage?.WorkerMetadata == null)
```
kshyju: I see we use a global variable here. I understand init event happens only once Ideally. But I think it might be good idea to pass this instance as a parameter to this method, to prevent any future bugs.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -274,6 +278,22 @@ internal void WorkerInitResponse(GrpcEvent initEvent)
...
+            {
+                return;
+            }
+
+            _initMessage.WorkerMetadata.RuntimeName = string.IsNullOrEmpty(_initMessage.WorkerMetadata.RuntimeName)
+                                                        ? _workerConfig.Description.Language : _initMessage.WorkerMetadata.RuntimeName;
```
kshyju: nit: Could be simplified to 

````
_initMessage.WorkerMetadata.RuntimeName ??= _workerConfig.Description.Language;
````

File: src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs [hunk 4]
```diff
@@ -274,6 +278,22 @@ internal void WorkerInitResponse(GrpcEvent initEvent)
...
+            _initMessage.WorkerMetadata.RuntimeVersion = string.IsNullOrEmpty(_initMessage.WorkerMetadata.RuntimeVersion)
+                                                        ? _workerConfig.Description.DefaultRuntimeVersion : _initMessage.WorkerMetadata.RuntimeVersion;
+
+            _metricsLogger.LogEvent(MetricEventNames.WorkerMetadata, null, _initMessage.WorkerMetadata.ToString());
+            _workerChannelLogger.LogDebug($"Worker metadata: {_initMessage.WorkerMetadata}");
```
kshyju: Oh, The `ToString` call here on `WorkerMetadata` produces a JSON string. 👍

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -274,6 +278,22 @@ internal void WorkerInitResponse(GrpcEvent initEvent)
...
+            {
+                return;
+            }
+
+            _initMessage.WorkerMetadata.RuntimeName = string.IsNullOrEmpty(_initMessage.WorkerMetadata.RuntimeName)
+                                                        ? _workerConfig.Description.Language : _initMessage.WorkerMetadata.RuntimeName;
+            _initMessage.WorkerMetadata.RuntimeVersion = string.IsNullOrEmpty(_initMessage.WorkerMetadata.RuntimeVersion)
+                                                        ? _workerConfig.Description.DefaultRuntimeVersion : _initMessage.WorkerMetadata.RuntimeVersion;
+
+            _metricsLogger.LogEvent(MetricEventNames.WorkerMetadata, null, _initMessage.WorkerMetadata.ToString());
```
kshyju: nit: Consider using the argument name to help with readability for the `null` param.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -274,6 +278,22 @@ internal void WorkerInitResponse(GrpcEvent initEvent)
...
+        private void ParseAndLogWorkerMetadata()
+        {
+            if (_initMessage?.WorkerMetadata == null)
+            {
+                return;
+            }
+
+            _initMessage.WorkerMetadata.RuntimeName = string.IsNullOrEmpty(_initMessage.WorkerMetadata.RuntimeName)
+                                                        ? _workerConfig.Description.Language : _initMessage.WorkerMetadata.RuntimeName;
+            _initMessage.WorkerMetadata.RuntimeVersion = string.IsNullOrEmpty(_initMessage.WorkerMetadata.RuntimeVersion)
```
kshyju: So, we are falling back to workerConfig values if worker GRPC message does not have certain properties. It may be possible for us to log a mix of what was sent from worker and what is present in workerconfig., Would that cause any confusion? Should we be explicit that we are using the fallback version value?

kshyju (approved):

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
liliankasem: I actually started with that originally but it doesn't work in this case as `??=` only checks null and we need to check for null or empty. I debugged this with the dotnet worker and when `RuntimeName` was empty, it did not use the worker config language field

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
liliankasem: Yeah I was debating if I should pass it in or not, happy to do that just in case of future issues

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
liliankasem: Yup

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
liliankasem: I don't think it would cause confusion given these two values are intentionally optional, and this is just for telemetry so there is not code impact. I can add a log here to indicate we are using the worker config but I don't feel it is necessary

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
liliankasem: Done! Though it does feel uglier this way xD

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
kshyju: You are right. My bad 😔

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
kshyju: Makes sense. Agree that it may be unnecessary.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -255,12 +255,21 @@ internal void WorkerInitResponse(GrpcEvent initEvent)
             _workerChannelLogger.LogDebug("Received WorkerInitResponse. Worker process initialized");
             _initMessage = initEvent.Message.WorkerInitResponse;
             _workerChannelLogger.LogDebug($"Worker capabilities: {_initMessage.Capabilities}");
+
+            if (_initMessage.WorkerMetadata != null)
+            {
+                _initMessage.WorkerMetadata = ParseWorkerMetadata(_initMessage.WorkerMetadata);
+                _metricsLogger.LogEvent(MetricEventNames.WorkerMetadata, functionName: null, _initMessage.WorkerMetadata.ToString());
+                _workerChannelLogger.LogDebug($"Worker metadata: {_initMessage.WorkerMetadata}");
```
fabiocav: This is serializing the metadata to JSON twice here.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -255,12 +255,21 @@ internal void WorkerInitResponse(GrpcEvent initEvent)
             _workerChannelLogger.LogDebug("Received WorkerInitResponse. Worker process initialized");
             _initMessage = initEvent.Message.WorkerInitResponse;
             _workerChannelLogger.LogDebug($"Worker capabilities: {_initMessage.Capabilities}");
+
+            if (_initMessage.WorkerMetadata != null)
+            {
+                _initMessage.WorkerMetadata = ParseWorkerMetadata(_initMessage.WorkerMetadata);
+                _metricsLogger.LogEvent(MetricEventNames.WorkerMetadata, functionName: null, _initMessage.WorkerMetadata.ToString());
```
fabiocav: Do you have an example of what the serialized payload will look like with custom properties? It would be good to understand whether the way we'll be persisting this will make it easy/efficient for querying.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -255,12 +255,21 @@ internal void WorkerInitResponse(GrpcEvent initEvent)
             _workerChannelLogger.LogDebug("Received WorkerInitResponse. Worker process initialized");
             _initMessage = initEvent.Message.WorkerInitResponse;
             _workerChannelLogger.LogDebug($"Worker capabilities: {_initMessage.Capabilities}");
+
+            if (_initMessage.WorkerMetadata != null)
+            {
+                _initMessage.WorkerMetadata = ParseWorkerMetadata(_initMessage.WorkerMetadata);
```
fabiocav: The object passed in gets updated, so it makes it a bit odd to reassign to this property. Perhaps updating the name (or adding the method to the message) would make this clearer

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
liliankasem: @fabiocav yup, there's a screenshot in the PR description - it's a json object so should be easy to query

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
liliankasem: Yeah I didn't really like this approach either - but adding the method to the message as an extension sounds like the nicest way to do this!

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
liliankasem: Just to confirm, this is because I'm passing `_initMessage.WorkerMetadata` twice?

fabiocav (approved):

PR #8517, opened 2022-06-29
Active profile tracking bug fix and adding logs

### LINKED ISSUE #8516: Worker Profiles - Active profile not being tracked
During specialization, we want to restart job host to use the right profile in case it has changed.
Currently active profiles and existing profiles are not visible at WebHostRpcWorkerChannelManager

Currently the profile manager is initialized at the [RpcWorkerConfigFactory](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs#L54)  and also at [WebHostRpcWorkerChannelManager.cs](https://github.com/Azure/azure-functions-host/blob/71c6243600ec70af83ceb57fc520c850cbb95fdd/src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs#L145-L175) . Since both are instantiated in different contexts, they currently do not share the existing profiles and active profile information. We want to mainly have one copy/instance of the profile manager so we have consistent profiles data.

Since Job host is restarted multiple times, the profile manager is initialized multiple times. To avoid this it is decided to make the service itself a global application singleton by moving it to the Webhost.

Profiles PR - https://github.com/Azure/azure-functions-host/pull/8365

shreyas-gopalakrishna:
Active profile is not being tracked. Updating the WorkerProfileManager to fix this
resolves #8516

### Pull request checklist

    * [x] Otherwise: Backport tracked by issue/PR #8530

### DIFF
--- a/src/WebJobs.Script/Workers/Profiles/WorkerProfileManager.cs
+++ b/src/WebJobs.Script/Workers/Profiles/WorkerProfileManager.cs
@@ -16,15 +16,13 @@ namespace Microsoft.Azure.WebJobs.Script.Workers
     {
         private readonly ILogger _logger;
         private readonly IEnumerable<IWorkerProfileConditionProvider> _conditionProviders;
-        private Dictionary<string, List<WorkerDescriptionProfile>> _profiles;
-        private string _activeProfile;
+        private static Dictionary<string, List<WorkerDescriptionProfile>> _profiles = new Dictionary<string, List<WorkerDescriptionProfile>>();
+        private static string _activeProfile = string.Empty;
 
         public WorkerProfileManager(ILogger logger, IEnumerable<IWorkerProfileConditionProvider> conditionProviders)
         {
             _logger = logger ?? throw new ArgumentNullException(nameof(logger));
             _conditionProviders = conditionProviders ?? throw new ArgumentNullException(nameof(conditionProviders));
-            _profiles = new Dictionary<string, List<WorkerDescriptionProfile>>();
-            _activeProfile = string.Empty;
         }
 
         /// <inheritdoc />
@@ -42,6 +40,7 @@ namespace Microsoft.Azure.WebJobs.Script.Workers
                 {
                     if (profile.EvaluateConditions())
                     {
+                        _logger?.LogInformation($"Conditions evaluated succesfully for profile with name: {profile.Name} and ID: {profile.ProfileId}");
                         evaluatedProfile = profile;
                         return true;
                     }
@@ -89,6 +88,7 @@ namespace Microsoft.Azure.WebJobs.Script.Workers
             {
                profileId = profile.ProfileId;
             }
+            _logger?.LogInformation($"Active profile: {_activeProfile} Evaluated profile: {profileId}");
             return _activeProfile.Equals(profileId);
         }
     }
--- a/src/WebJobs.Script/Workers/Profiles/WorkerProfileManager.cs
+++ b/src/WebJobs.Script/Workers/Profiles/WorkerProfileManager.cs
@@ -40,7 +40,7 @@ namespace Microsoft.Azure.WebJobs.Script.Workers
                 {
                     if (profile.EvaluateConditions())
                     {
-                        _logger?.LogInformation($"Conditions evaluated succesfully for profile with name: {profile.Name} and ID: {profile.ProfileId}");
+                        _logger.LogInformation($"Conditions evaluated succesfully for profile with name: {profile.Name} and ID: {profile.ProfileId}");
                         evaluatedProfile = profile;
                         return true;
                     }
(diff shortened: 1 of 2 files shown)


src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
```diff
@@ -169,6 +163,13 @@ private bool UsePlaceholderChannel(string workerRuntime)
                     // Use if readonly and not v2 compatible on ~3 extension
                     return _applicationHostOptions.CurrentValue.IsFileSystemReadOnly && !_environment.IsV2CompatibileOnV3Extension();
                 }
+
+                // If a profile evaluates to true and was not previously loaded, restart worker process
+                if (!_profileManager.IsCorrectProfileLoaded(workerRuntime))
+                {
```
madelinegordon, line 169: Is the ordering of the checks in this method important? If not, wondering if it would make sense for this condition to be added to the check in line 151 to avoid duplication of "return false" ?

liliankasem: I'm not 100% sure, but I think ordering might be important here. @fabiocav & @pragnagopa probably know more about this code path

shreyas-gopalakrishna: Not sure about the ordering. I felt it was better to respect previous conditions to not use placeholder channel over the newly introduced profiles change.

src/WebJobs.Script/Workers/Profiles/WorkerProfileManager.cs
```diff
@@ -42,6 +40,7 @@ private bool GetEvaluatedProfile(string language, out WorkerDescriptionProfile e
                 {
                     if (profile.EvaluateConditions())
                     {
+                        _logger?.LogInformation($"Conditions evaluated succesfully for profile with name: {profile.Name} and ID: {profile.ProfileId}");
```
liliankasem: Can probably remove the `?` as it's been null check in the constructor

src/WebJobs.Script/Workers/Profiles/WorkerProfileManager.cs
```diff
@@ -89,6 +88,7 @@ public bool IsCorrectProfileLoaded(string workerRuntime)
             {
                profileId = profile.ProfileId;
             }
+            _logger?.LogInformation($"Active profile: {_activeProfile} Evaluated profile: {profileId}");
```
liliankasem: Can probably remove the `?` as it's been null check in the constructor

src/WebJobs.Script/Workers/Profiles/WorkerProfileManager.cs
```diff
@@ -16,15 +16,13 @@ internal class WorkerProfileManager : IWorkerProfileManager
     {
         private readonly ILogger _logger;
         private readonly IEnumerable<IWorkerProfileConditionProvider> _conditionProviders;
-        private Dictionary<string, List<WorkerDescriptionProfile>> _profiles;
-        private string _activeProfile;
+        private static Dictionary<string, List<WorkerDescriptionProfile>> _profiles = new Dictionary<string, List<WorkerDescriptionProfile>>();
```
liliankasem, line 19: @fabiocav any concerns with this being made static here? Or do you have a better recommendation for addressing this issue?

liliankasem (approved):
This will also need to be backported to v3.x

src/WebJobs.Script/Workers/Profiles/WorkerProfileManager.cs
fabiocav: This change is indeed concerning as this service is instantiated in multiple scopes and sharing this may introduce fragility and impact other people working on this code.

@shreyas-gopalakrishna we may want to consider reusing this service across different scopes instead (having a singleton instance of this service) if the multiple instances are causing problems, but this will bring additional complexities with you having to handle specialization events within the service implementation. That's not a problem, but something you need to be aware of.

src/WebJobs.Script/Workers/Profiles/WorkerProfileManager.cs
```diff
@@ -42,6 +40,7 @@ private bool GetEvaluatedProfile(string language, out WorkerDescriptionProfile e
                 {
                     if (profile.EvaluateConditions())
                     {
+                        _logger.LogInformation($"Conditions evaluated succesfully for profile with name: {profile.Name} and ID: {profile.ProfileId}");
```
fabiocav, line 43: For all logging here, please pass the values as arguments into the log methods instead of using interpolation/concatenating the strings.

fabiocav (changes requested):

src/WebJobs.Script/Workers/Profiles/WorkerProfileManager.cs
shreyas-gopalakrishna: @fabiocav 
The goal here is to have 1 instance/copy of `_profiles` and `_activeProfile`. 
Is singleton the best approach we can take here? Once we have a singleton, any info on how to handle the specialization events? I recollect you pointing to AssemblyAnalysisService. Curious as to what changes to check here, is it the `IOptionsMonitor<StandbyOptions> standbyOptionsMonitor`? 

https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.WebHost/AssemblyAnalysis/AssemblyAnalysisService.cs#L31

PR #8524, opened 2022-07-05
Implementing diagnostic listener service

fabiocav:

This change introduces a new hosted service that subscribes to event sources defined in the host and persists events from those sources into our logging pipeline.

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticListenerService.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticListenerService.cs
@@ -28,16 +28,18 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
             _logger = logger;
             _listenerSubscriptions = new Dictionary<string, IDisposable>();
 
+            SetDebugState();
+
             if (standbyOptions.CurrentValue.InStandbyMode)
             {
-                _standbyChangeHandler = standbyOptions.OnChange(o => SubscribeListeners());
+                _standbyChangeHandler = standbyOptions.OnChange(o => SetDebugState());
             }
         }
 
+        private static void SetDebugState() => _debugTraceEnabled = FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagEnableDebugTracing);
+
         private void SubscribeListeners()
         {
-            _debugTraceEnabled = FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagEnableDebugTracing);
-
             if (_allListenersSubscription != null)
             {
                 _allListenersSubscription.Dispose();
@@ -126,7 +128,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
                 }
                 else
                 {
-                    _logger.LogDebug("Listener '{listener}' emitted event '{eventName}': {payload}", _listenerName, kvp.Key, kvp.Value);
+                    _logger.LogDebug("Diagnostic source '{source}' emitted event '{eventName}': {payload}", _listenerName, kvp.Key, kvp.Value);
                 }
             }
         }
--- a/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticListenerService.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticListenerService.cs
@@ -1,0 +1,134 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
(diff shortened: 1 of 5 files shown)


src/WebJobs.Script.WebHost/Diagnostics/DiagnosticListenerService.cs
```diff
@@ -0,0 +1,134 @@
...
+        private static bool _debugTraceEnabled;
+        private IDisposable _allListenersSubscription;
+        private bool _disposed;
+
+        public DiagnosticListenerService(ILogger<DiagnosticListenerService> logger, IOptionsMonitor<StandbyOptions> standbyOptions)
+        {
+            _logger = logger;
+            _listenerSubscriptions = new Dictionary<string, IDisposable>();
+
+            if (standbyOptions.CurrentValue.InStandbyMode)
```
liliankasem, line 33: Do we need to do something like this for null checking?

`standbyOptions?.CurrentValue?.InStandbyMode`

liliankasem (approved):

VpOfEngineering:
Please hold on merging this until tomorrow - to preserve dev for the JIT Trace. Let me know if that is not possible :)

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticListenerService.cs
```diff
@@ -0,0 +1,134 @@
...
+            {
+                _standbyChangeHandler = standbyOptions.OnChange(o => SubscribeListeners());
```
brettsam: Do you have to do this? If this is at the WebHost level, I thought that your `Subscribe` callback would get called for all future diagnostic events as well.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticListenerService.cs
```diff
@@ -0,0 +1,134 @@
...
+
+            public void OnNext(KeyValuePair<string, object> kvp)
+            {
+                if (kvp.Value is IHostDiagnosticEvent diagnosticEvent)
+                {
+                    diagnosticEvent.LogEvent(_logger);
+                }
+                else
+                {
+                    _logger.LogDebug("Listener '{listener}' emitted event '{eventName}': {payload}", _listenerName, kvp.Key, kvp.Value);
```
brettsam: Using the `Listener` terminology could be confusing... can we say something else to make it clearer that this isn't a WebJobs Listener?

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticListenerService.cs
fabiocav: Are there scenarios you would you expect the monitor or its value to be `null` here?

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticListenerService.cs
fabiocav: For new listeners it would, but if they're setup on something that won't change (e.g., the assembly load context), it wouldn't, but I see why you're asking, as when I initially implemented this, the logic enabled full sources (source names) to be for debug only, which means we'd not subscribe to them in placeholder mode, but that was moved to the event name, so this wouldn't be necessary.

I'll still need to handle the change, but only to flip the debug flag. Good catch!

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticListenerService.cs
fabiocav: Yeah, I agree. I actually think it would make sense to just treat this as a source. Thinking about changing `Listener` to `Diagnostic source` here. Thoughts?

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticListenerService.cs
liliankasem: I'm not sure! It depends on where we are getting it from and how it's populated so let's have a look:

We DI `StandbyOptionsSetup ` and looking at the [code](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.WebHost/Standby/StandbyOptionsSetup.cs) for that, it looks like we'll throw if environment is null. So I guess things would fail before we get here and so `standbyOptions` should not be null at this stage. 

In general I'm always caution of items being injected not initializing properly, but if things fail before this then we should be fine not to have an extra check?

brettsam (approved):

VpOfEngineering:
You are good to merge from my end, release has been cut

PR #8538, opened 2022-07-11
[pack] Adding Multiple Languages Support in Functions

ghost:
Port of #8449

### DIFF
--- a/src/WebJobs.Script/Environment/EnvironmentExtensions.cs
+++ b/src/WebJobs.Script/Environment/EnvironmentExtensions.cs
@@ -19,6 +19,8 @@ namespace Microsoft.Azure.WebJobs.Script
     {
         private static bool? isApplicationInsightsAgentEnabled;
 
+        private static bool? isMultiLanguageEnabled;
+
         internal static string BaseDirectory { get; set; }
 
         public static string GetEnvironmentVariableOrDefault(this IEnvironment environment, string name, string defaultValue)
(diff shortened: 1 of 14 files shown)


src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -311,14 +314,26 @@ public static string GetAntaresComputerName(this IEnvironment environment)
         }
 
         /// <summary>
-        /// Gets the computer name.
+        /// Gets if runtime enviromnent is logic apps.
```
madelinegordon: nit - "environment"

src/WebJobs.Script/Host/ScriptHost.cs
```diff
@@ -297,7 +297,8 @@ public async Task InitializeAsync(CancellationToken cancellationToken = default)
                         // Windows Consumption as well.
                         string runtimeVersion = _environment.GetEnvironmentVariable(RpcWorkerConstants.FunctionWorkerRuntimeVersionSettingName);
 
-                        if (!string.IsNullOrEmpty(runtimeVersion))
+                        // If the environemt is multi language, we are running multiple language worker, and thus not honoring 'FUNCTIONS_WORKER_RUNTIME_VERSION'
```
madelinegordon: nit - "environment" and "workers"

src/WebJobs.Script/Host/ScriptHost.cs
```diff
@@ -560,7 +568,7 @@ private void AddFunctionDescriptors(IEnumerable<FunctionMetadata> functionMetada
                 var workerConfig = _languageWorkerOptions.Value.WorkerConfigs?.FirstOrDefault(c => c.Description.Language.Equals(_workerRuntime, StringComparison.OrdinalIgnoreCase));
 
                 // If there's no worker config, use the default (for legacy behavior; mostly for tests).
-                TimeSpan initializationTimeout = workerConfig?.CountOptions?.InitializationTimeout ?? WorkerProcessCountOptions.DefaultInitializationTimeout;
+                var initializationTimeout = workerConfig?.CountOptions?.InitializationTimeout ?? WorkerProcessCountOptions.DefaultInitializationTimeout;
```
madelinegordon: Just curious - why was this type changed?

ghost: Kept it consistent with line above, where `workerConfig` var is used instead of `ClassName`.

liliankasem: I think using Timespan makes more sense here. With `workerConfig`, we know the type of the variable, but with `initializationTimeout` it's not clear through assignment what the type is and so it makes sense to be clear with that.

ghost: Reverted back, will also backport it to v3.x

madelinegordon (approved):

soninaren (approved):

PR #8556, opened 2022-07-19
Send InvocationCancel request to out-of-proc workers

### LINKED ISSUE #2152: Provide out-of-proc support for cancellation tokens
In C# functions you can take a `CancellationToken` argument to allow your function to detect that a host shutdown is occurring and abort gracefully. This work item tracks provide an equivalent mechanism for languages that execute out of proc. Exactly what type/data structure would be used would be language dependent.

liliankasem:

resolves #2152

Implementing the `InvocationCancel` message so provide cancellation token support for out-of-proc language workers.

First, we setup the CancellationToken from WebJobs to be passed into the invocation context parameters list. This enables us to flow the CancellationToken to the InvokeCore method and eventually to SendInvocationRequest. 

Within SendInvocationRequest, we register a callback for the CancellationToken so that when it is cancelled, we send the InvocationCancel message to the worker. This feature is built behind a worker capability flag, if a worker does not have a cancellation capability, we will not send the InvocationCancel message

Design document available in the [wiki](https://github.com/Azure/azure-functions-host/wiki/Out-of-proc-Support-for-Cancellation-Tokens).

### Pull request checklist


src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -500,36 +500,61 @@ internal async Task SendInvocationRequest(ScriptInvocationContext context)
...
+                    _workerChannelLogger.LogDebug("Cancellation has been requested, cancelling invocation request");
+                    context.ResultSource.SetCanceled();
+                    return;
                 }
+
+                var invocationRequest = await context.ToRpcInvocationRequest(_workerChannelLogger, _workerCapabilities, _isSharedMemoryDataTransferEnabled, _sharedMemoryManager);
+                AddAdditionalTraceContext(invocationRequest.TraceContext.Attributes, context);
+                _executingInvocations.TryAdd(invocationRequest.InvocationId, context);
+
+                context.CancellationToken.Register(() => SendInvocationCancel(invocationRequest.InvocationId));
```
kshyju: Would this line be executed if the `context.CancellationToken.IsCancellationRequested` expression in line 513 yields `true` (because we have a return statement in that block)?

src/WebJobs.Script/Description/Workers/WorkerFunctionDescriptorProvider.cs
```diff
@@ -54,6 +55,9 @@ protected override IFunctionInvoker CreateFunctionInvoker(string scriptFilePath,
         {
             var parameters = await base.GetFunctionParametersAsync(functionInvoker, functionMetadata, triggerMetadata, methodAttributes, inputBindings, outputBindings);
 
+            // Add cancellation token
+            parameters.Add(new ParameterDescriptor(ScriptConstants.SystemCancellationTokenParameterName, typeof(CancellationToken)));
```
kshyju, line 59: Pardon my silly question :) Do we need to add it like this? If OOP function has CancellationToken param, will it be not added like other bindings/parameters automatically with existing code? Also where is this parameter (`_cancellationToken`) being used?

src/WebJobs.Script/Description/Workers/WorkerFunctionInvoker.cs
```diff
@@ -187,6 +187,16 @@ private object TransformInput(object input, Dictionary<string, object> bindingDa
...
         }
 
+        private CancellationToken HandleCancellationTokenParameter(object input)
+        {
+            if (input == null)
+            {
+                return CancellationToken.None;
+            }
+
+            return (CancellationToken)input;
```
kshyju, line 197: Another silly question: Will the OOP worker actually use this CancellationToken instance from this payload in the invocation request? I thought (for example, in the dotnet worker), it is going to create a token source for each invocation and when the `InvocationCancel` GRPC message arrives, it call Cancel on the token.

kshyju (commented):
Few questions to understand the implementation better. Thanks!

release_notes.md
```diff
@@ -2,9 +2,10 @@
...
 - My change description (#PR)
 -->
-- Update Python Worker Version to [4.3.0](https://github.com/Azure/azure-functions-python-worker/releases/tag/4.3.0)
 
-- Surface exception type and details from out-of-proc user code to App Insights (#8314)
+- Update Python Worker Version to [4.3.0](https://github.com/Azure/azure-functions-python-worker/releases/tag/4.3.0)
+- Surface exception type and details from out-of-proc user code to App Insights ([#8314](https://github.com/Azure/azure-functions-host/pull/8314))
+- Host support for out-of-proc cancellation tokens ([#2153](https://github.com/Azure/azure-functions-host/issues/2152))
 
 **Release sprint:** Sprint 124
```
madelinegordon: nit - sprint 126?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -500,36 +500,61 @@ internal async Task SendInvocationRequest(ScriptInvocationContext context)
...
+                    InvocationRequest = invocationRequest
+                });
             }
             catch (Exception invokeEx)
             {
                 context.ResultSource.TrySetException(invokeEx);
             }
         }
 
+        internal void SendInvocationCancel(string invocationId)
```
madelinegordon, line 543: Should there be a null check somewhere in here for invocationId?

madelinegordon (approved):

src/WebJobs.Script/Workers/Rpc/RpcInitializationService.cs
```diff
@@ -59,7 +59,7 @@ public async Task StartAsync(CancellationToken cancellationToken)
 
         public async Task StopAsync(CancellationToken cancellationToken)
         {
-            _logger.LogDebug("Shuttingdown Rpc Channels Manager");
+            _logger.LogDebug("Shutting down Rpc Channels Manager");
```
surgupta-msft: Curious, do we have any process around keeping our detectors updated when changing log messages (if any detector is dependent on this log line)?
cc @kashimiz

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
liliankasem: Nope, this line will not execute and we don't want it to. At this point a cancellation has already been requested so we can to cancel the invocation. We only want to register the callback when an invocation is going to be sent to the worker / will be in-flight

src/WebJobs.Script/Description/Workers/WorkerFunctionDescriptorProvider.cs
liliankasem: I think @fabiocav might know the answer better, but from my understanding WebJobs has a cancellation token and this line is allowing us to grab the `_cancellationToken` param from WebJobs (this name is defined by WebJobs I think?). Once we have the cancellation token in the parameters object, it flows down to [InvokeCore](https://github.com/Azure/azure-functions-host/blob/422c5b4059edc005ecf94b7d4804155b35da5e61/src/WebJobs.Script/Description/Workers/WorkerFunctionInvoker.cs#L58) where we add the CT to the invocation context.

src/WebJobs.Script/Description/Workers/WorkerFunctionInvoker.cs
liliankasem: Nope, it will not use this instance. The OOP worker will create its own CancellationTokenSource and cancel it when it receives the InvocationCancel message (in the dotnet scenario, this will ofc look different for each langauge).

This cancellation token comes from WebJobs and when cancelled (for timeout or other reasons), will lead to the host sending a InvocationCancel message. The workers will not have any ties to this CT

src/WebJobs.Script/Workers/Rpc/RpcInitializationService.cs
liliankasem: Good question, I am also worried about that. Happy to revert for now and have a cleanup PR in future

src/WebJobs.Script/Description/Workers/WorkerFunctionInvoker.cs
liliankasem: @kshyju a quick draft I have been using to test this E2E: https://github.com/Azure/azure-functions-dotnet-worker/pull/972

Feel free to suggest a different approach here, I'm not familiar with the dotnet worker codebase and best practices!

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
liliankasem: I don't believe this can ever be null when we get here, this value comes from [WebJobs](https://github.com/Azure/azure-webjobs-sdk/blob/40153369ff35184314dc0c827a32d0be63ccbcc0/src/Microsoft.Azure.WebJobs.Host/Bindings/ExecutionContext/ExecutionContext.cs#L17) and things would have already failed way before we get to invocation if this item is null. You could trace the invocationId and see if it can be null by the time we get to [Invoke](https://github.com/Azure/azure-functions-host/blob/6c80f8b0649f964899c30228241969d73275acb7/src/WebJobs.Script/Description/FunctionInvokerBase.cs#L79-L83)

satvu (approved):
lgtm!

src/WebJobs.Script.Grpc/MessageExtensions/StatusResultExtensions.cs
```diff
@@ -43,8 +43,8 @@ public static bool IsInvocationSuccess<T>(this StatusResult status, TaskCompleti
                     return false;
 
                 case StatusResult.Types.Status.Cancelled:
-                    tcs.SetCanceled();
-                    return false;
+                    var cancellationException = GetRpcException(status);
+                    throw new OperationCanceledException("Invocation was cancelled", cancellationException);
```
liliankasem, line 47: Just want to confirm, we definitely want to throw here?

todo item for me: validate what happens in-proc

src/WebJobs.Script.Grpc/MessageExtensions/StatusResultExtensions.cs
```diff
@@ -43,7 +43,8 @@ public static bool IsInvocationSuccess<T>(this StatusResult status, TaskCompleti
                     return false;
 
                 case StatusResult.Types.Status.Cancelled:
-                    tcs.SetCanceled();
+                    var exception = new TaskCanceledException("Invocation cancelled");
+                    tcs.SetException(exception);
```
liliankasem: Walked through what happens in-proc and it's the same as what we're doing here. The task status is getting set to cancelled, and we throw a `TaskCanceledException` exception

Only concern here is that we're missing the stack trace - @fabiocav thoughts?

liliankasem: Updated to use rpc exception

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -501,36 +501,61 @@ internal async Task SendInvocationRequest(ScriptInvocationContext context)
...
+                    _workerChannelLogger.LogDebug("Cancellation has been requested, cancelling invocation request");
+                    context.ResultSource.SetCanceled();
+                    return;
                 }
+
+                var invocationRequest = await context.ToRpcInvocationRequest(_workerChannelLogger, _workerCapabilities, _isSharedMemoryDataTransferEnabled, _sharedMemoryManager);
+                AddAdditionalTraceContext(invocationRequest.TraceContext.Attributes, context);
+                _executingInvocations.TryAdd(invocationRequest.InvocationId, context);
+
+                context.CancellationToken.Register(() => SendInvocationCancel(invocationRequest.InvocationId));
```
kshyju: Should we move the "capability enabled" check(which we are doing at 540) here so that we register this callback only when the capability is enabled?

liliankasem: Good call, will move the check up!

PR #8518, opened 2022-06-29
Add compile errors to the functionerror collection

chiangvincent:
This PR surfaces compilation errors for customers using C# script (.csx files). This is part of an ongoing effort to make scale decisions in scale controller more intelligent -- if we see that a function is failing to compile, it should not be triggering scale out votes to begin with. The scale controller will poll the `admin/functions/<name>/status` API upon site metadata refreshes to check the health of each function in a given function app. If a function returns any errors from this endpoint, its scale out votes will be overridden with scale in votes, given it meets these other two conditions: 

1. It is not the only function in the app
2. The current worker count is greater than 1

The scale controller PR is [here](https://msazure.visualstudio.com/One/_git/AAPT-Antares-ScaleController/pullrequest/6526663?path=/src/Microsoft.Azure.AppService.ScaleController.Extension/Shared/ScaleControllerConfiguration.cs).

A sample app on private stamp that surfaces compilation errors (simple HTTP trigger using C# script, with undefined global variable) 

[Kusto](https://dataexplorer.azure.com/clusters/wawstest/databases/wawskustotest?query=H4sIAAAAAAAAA2XMsQqDMBRG4b1PcXFSqGJSoiA4lY4tBfsCafqjKSaG5KJLH75OXboeDt9g9Izz4jku84x4WeE5HT60TYigwTJu2oH6nrLVTFb7cTMhIYhVZL/tHmFswsM6DKxdoCd4A3z+0gzeay5rKcu6LYUkWXdKdidVqaaRbaMKqioSzvpi90Jc3jD8Jx7pipT0iC/AHLhxsAAAAA==)

ScaleControllerEvents
| where SiteName == "vchiangwcpsep1v1"
| where PreciseTimeStamp between(datetime(2022-07-12 20:52:35.5662765) .. 1min)
| project PreciseTimeStamp, Message

Sample API response returned from browser, using private site extension on production function app with broken HTTP trigger (undefined variable test in global scope):
image

Portal experience is not affected by compilation error surfacing:
image

image

### Issue describing the changes in this PR

resolves #issue_for_this_pr

### Pull request checklist

* [ ] My changes **do not** require documentation changes
    * [ ] Otherwise: Documentation issue linked to PR
* [ ] My changes **should not** be added to the release notes for the next release
    * [ ] Otherwise: I've added my notes to `release_notes.md`
* [ ] My changes **do not** need to be backported to a previous version
    * [ ] Otherwise: Backport tracked by issue/PR #issue_or_pr
[...]

PR #8609, opened 2022-08-05
[pack] build compileerrors PR

pragnagopa:

resolves #issue_for_this_pr

### Pull request checklist

### Additional information

Additional PR information


liliankasem:
@pragnagopa is there an issue you can link to & update description to provide more info on what this change is about

liliankasem (approved):

src/WebJobs.Script/Utility.cs
```diff
@@ -541,10 +541,10 @@ internal static void AddFunctionError(IDictionary<string, ICollection<string>> f
         {
             functionName = isFunctionShortName ? functionName : Utility.GetFunctionShortName(functionName);
 
-            ICollection<string> functionErrorCollection = new Collection<string>();
+            ICollection<string> functionErrorCollection = new HashSet<string>();
```
jviau, line 544: I think there is a chance for some further optimization here:
1. We can avoid the extra hash set allocation and move it to be declared as part of the `out` param.
    - `!functionErrors.TryGetValue(functionName, out ICollection<string> functionErrorCollection)`
2. The check for `!string.IsNullorEmpty(functionName)` is interesting. If it is null or empty we won't even add `functionErrorCollection` to the dictionary, in which case we can just exit after line 542. But this begs the question - is null or empty even valid here, or is this indicative of a misuse of this function? And maybe we should throw an `ArgumentException` (fail fast principal)?

jviau:
I think this change has already been checked in: #8518

@pragnagopa considering this is already present in the repo, any reason we should proceed with the PR? Although I think the other PR may have gone in _after_ you made this PR - so not your fault there.

PR #8624, opened 2022-08-11
[pack] make channel lookups case-insensitive

brettsam:
Worker channels were stored in a dictionary with case-sensitive keys, meaning that mismatches between `FUNCTIONS_WORKER_RUNTIME` and the worker config's `language` value would cause the exception `Did not find any initialized language workers`.

### DIFF
--- a/test/WebJobs.Script.Tests/Workers/Rpc/WebHostRpcWorkerChannelManagerTests.cs
+++ b/test/WebJobs.Script.Tests/Workers/Rpc/WebHostRpcWorkerChannelManagerTests.cs
@@ -86,6 +86,9 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.Workers.Rpc
             Assert.NotNull(initializedChannel);
             Assert.Equal(javaWorkerChannel.Id, initializedChannel.Id);
             Assert.Equal(_rpcWorkerChannelManager.GetChannels(RpcWorkerConstants.JavaLanguageWorkerName).Count(), 2);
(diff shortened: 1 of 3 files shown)


test/WebJobs.Script.Tests/Workers/Rpc/WebHostRpcWorkerChannelManagerTests.cs
```diff
@@ -86,6 +86,9 @@ public async Task CreateChannels_Succeeds()
             Assert.NotNull(initializedChannel);
             Assert.Equal(javaWorkerChannel.Id, initializedChannel.Id);
             Assert.Equal(_rpcWorkerChannelManager.GetChannels(RpcWorkerConstants.JavaLanguageWorkerName).Count(), 2);
+
+            // test case insensitivity
+            Assert.Equal(_rpcWorkerChannelManager.GetChannels("Java").Count(), 2);
```
brettsam, line 91: The WebHost manager was already case-insensitive, but added an explicit test just to prevent changes in the future.

https://github.com/Azure/azure-functions-host/blob/c52bb6f0f2a8eb31eebe0367fca3d38873ac10f9/src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs#L35

surgupta-msft (approved):

madelinegordon (approved):

kshyju (approved):
💰

PR #8634, opened 2022-08-16
InstanceManager for Legion Runtime

### LINKED ISSUE #8616: Support FunctionsRuntime on Legion Host

Tonewall:
Created a Legion specific IInstanceManager (LegionInstanceManager) for Legion runtime. In Legion, the runtime no longer mounts/unzips the app content since that is handled by Legion. When assigning, we can assume that the app content is already mounted/unzipped in the correct file location. So, we don't need a lot of methods that are available in the current InstanceManager.cs.

resolves #8616 

### Pull request checklist

    * [x] Otherwise: Backport tracked by issue/PR #8616 

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/ContainerManagement/LinuxContainerInitializationHostService.cs
+++ b/src/WebJobs.Script.WebHost/ContainerManagement/LinuxContainerInitializationHostService.cs
@@ -18,12 +18,12 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.ContainerManagement
     public class LinuxContainerInitializationHostService : IHostedService
     {
         private readonly IEnvironment _environment;
-        private readonly IInstanceManager _instanceManager;
+        private readonly LinuxInstanceManager _instanceManager;
         private readonly ILogger _logger;
         private readonly StartupContextProvider _startupContextProvider;
         private CancellationToken _cancellationToken;
 
-        public LinuxContainerInitializationHostService(IEnvironment environment, IInstanceManager instanceManager, ILogger<LinuxContainerInitializationHostService> logger, StartupContextProvider startupContextProvider)
+        public LinuxContainerInitializationHostService(IEnvironment environment, LinuxInstanceManager instanceManager, ILogger<LinuxContainerInitializationHostService> logger, StartupContextProvider startupContextProvider)
         {
             _environment = environment;
             _instanceManager = instanceManager;
--- a/src/WebJobs.Script.WebHost/ContainerManagement/LinuxContainerInitializationHostService.cs
+++ b/src/WebJobs.Script.WebHost/ContainerManagement/LinuxContainerInitializationHostService.cs
@@ -18,12 +18,12 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.ContainerManagement
     public class LinuxContainerInitializationHostService : IHostedService
     {
         private readonly IEnvironment _environment;
-        private readonly LinuxInstanceManager _instanceManager;
+        private readonly IInstanceManager _instanceManager;
         private readonly ILogger _logger;
         private readonly StartupContextProvider _startupContextProvider;
         private CancellationToken _cancellationToken;
 
-        public LinuxContainerInitializationHostService(IEnvironment environment, LinuxInstanceManager instanceManager, ILogger<LinuxContainerInitializationHostService> logger, StartupContextProvider startupContextProvider)
+        public LinuxContainerInitializationHostService(IEnvironment environment, IInstanceManager instanceManager, ILogger<LinuxContainerInitializationHostService> logger, StartupContextProvider startupContextProvider)
         {
             _environment = environment;
             _instanceManager = instanceManager;
--- a/src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
+++ b/src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
@@ -20,11 +20,11 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Controllers
     public class InstanceController : Controller
     {
         private readonly IEnvironment _environment;
-        private readonly IInstanceManager _instanceManager;
+        private readonly LinuxInstanceManager _instanceManager;
         private readonly ILogger _logger;
         private readonly StartupContextProvider _startupContextProvider;
 
-        public InstanceController(IEnvironment environment, IInstanceManager instanceManager, ILoggerFactory loggerFactory, StartupContextProvider startupContextProvider)
+        public InstanceController(IEnvironment environment, LinuxInstanceManager instanceManager, ILoggerFactory loggerFactory, StartupContextProvider startupContextProvider)
         {
             _environment = environment;
             _instanceManager = instanceManager;
--- a/src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
+++ b/src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
@@ -56,6 +56,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Controllers
             {
                 return StatusCode(StatusCodes.Status500InternalServerError, error);
             }
+
             var succeeded = _instanceManager.StartAssignment(assignmentContext);
 
             return succeeded
--- a/src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
+++ b/src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
@@ -41,22 +41,27 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Controllers
 
             var assignmentContext = _startupContextProvider.SetContext(encryptedAssignmentContext);
 
-            // before starting the assignment we want to perform as much
-            // up front validation on the context as possible
-            string error = await _instanceManager.ValidateContext(assignmentContext);
-            if (error != null)
-            {
-                return StatusCode(StatusCodes.Status400BadRequest, error);
-            }
+            string error;
+            
+            // Skip MSI Specialization and validation for Legion
+            if (!_environment.IsLinuxConsumptionOnLegion()) {
+                // before starting the assignment we want to perform as much
+                // up front validation on the context as possible
+                error = await _instanceManager.ValidateContext(assignmentContext);
+                if (error != null)
+                {
+                    return StatusCode(StatusCodes.Status400BadRequest, error);
+                }
 
-            // Wait for Sidecar specialization to complete before returning ok.
-            // This shouldn't take too long so ok to do this sequentially.
-            error = await _instanceManager.SpecializeMSISidecar(assignmentContext);
-            if (error != null)
-            {
-                return StatusCode(StatusCodes.Status500InternalServerError, error);
+                // Wait for Sidecar specialization to complete before returning ok.
+                // This shouldn't take too long so ok to do this sequentially.
+                error = await _instanceManager.SpecializeMSISidecar(assignmentContext);
+                if (error != null)
+                {
+                    return StatusCode(StatusCodes.Status500InternalServerError, error);
+                }
             }
-
+            
             var succeeded = _instanceManager.StartAssignment(assignmentContext);
 
             return succeeded
--- a/src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
+++ b/src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
@@ -20,11 +20,11 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Controllers
     public class InstanceController : Controller
     {
         private readonly IEnvironment _environment;
-        private readonly LinuxInstanceManager _instanceManager;
+        private readonly IInstanceManager _instanceManager;
         private readonly ILogger _logger;
         private readonly StartupContextProvider _startupContextProvider;
 
-        public InstanceController(IEnvironment environment, LinuxInstanceManager instanceManager, ILoggerFactory loggerFactory, StartupContextProvider startupContextProvider)
+        public InstanceController(IEnvironment environment, IInstanceManager instanceManager, ILoggerFactory loggerFactory, StartupContextProvider startupContextProvider)
         {
             _environment = environment;
             _instanceManager = instanceManager;
--- a/src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
+++ b/src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
@@ -41,27 +41,21 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Controllers
 
             var assignmentContext = _startupContextProvider.SetContext(encryptedAssignmentContext);
 
-            string error;
-            
-            // Skip MSI Specialization and validation for Legion
-            if (!_environment.IsLinuxConsumptionOnLegion()) {
-                // before starting the assignment we want to perform as much
-                // up front validation on the context as possible
-                error = await _instanceManager.ValidateContext(assignmentContext);
-                if (error != null)
-                {
-                    return StatusCode(StatusCodes.Status400BadRequest, error);
-                }
-
-                // Wait for Sidecar specialization to complete before returning ok.
-                // This shouldn't take too long so ok to do this sequentially.
-                error = await _instanceManager.SpecializeMSISidecar(assignmentContext);
-                if (error != null)
-                {
-                    return StatusCode(StatusCodes.Status500InternalServerError, error);
-                }
+            // before starting the assignment we want to perform as much
+            // up front validation on the context as possible
+            string error = await _instanceManager.ValidateContext(assignmentContext);
+            if (error != null)
+            {
+                return StatusCode(StatusCodes.Status400BadRequest, error);
             }
-            
+
+            // Wait for Sidecar specialization to complete before returning ok.
+            // This shouldn't take too long so ok to do this sequentially.
+            error = await _instanceManager.SpecializeMSISidecar(assignmentContext);
+            if (error != null)
+            {
+                return StatusCode(StatusCodes.Status500InternalServerError, error);
+            }
             var succeeded = _instanceManager.StartAssignment(assignmentContext);
 
             return succeeded
--- a/src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
@@ -40,7 +40,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
         public Task<string> SpecializeMSISidecar(HostAssignmentContext context)
         {
             // Legion will take care of MSI Specialization
-            return null;
+            return Task.FromResult((string)null);
         }
 
         public bool StartAssignment(HostAssignmentContext context)
@@ -64,41 +64,41 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
                 return true;
             }
 
-            if (_assignmentContext == null)
+            if (_assignmentContext != null)
             {
-                lock (_assignmentLock)
-                {
-                    if (_assignmentContext != null)
-                    {
-                        return _assignmentContext.Equals(context);
-                    }
-                    _assignmentContext = context;
-                }
-
-                _logger.LogInformation($"Starting Assignment. Cloud Name: {_environment.GetCloudName()}");
-
-                // set a flag which will cause any incoming http requests to buffer
-                // until specialization is complete
-                // the host is guaranteed not to receive any requests until AFTER assign
-                // has been initiated, so setting this flag here is sufficient to ensure
-                // that any subsequent incoming requests while the assign is in progress
-                // will be delayed until complete
-                _webHostEnvironment.DelayRequests();
-
-                // start the specialization process in the background
-                Task.Run(async () => await Assign(context));
-
-                return true;
+                // No lock needed here since _assignmentContext is not null when we are here
+                return _assignmentContext.Equals(context);
             }
 
-            // No lock needed here since _assignmentContext is not null when we are here
-            return _assignmentContext.Equals(context);
+            lock (_assignmentLock)
+            {
+                if (_assignmentContext != null)
+                {
+                    return _assignmentContext.Equals(context);
+                }
+                _assignmentContext = context;
+            }
+
+            _logger.LogInformation("Starting Assignment. Cloud Name: {cloudName}", _environment.GetCloudName());
+
+            // set a flag which will cause any incoming http requests to buffer
+            // until specialization is complete
+            // the host is guaranteed not to receive any requests until AFTER assign
+            // has been initiated, so setting this flag here is sufficient to ensure
+            // that any subsequent incoming requests while the assign is in progress
+            // will be delayed until complete
+            _webHostEnvironment.DelayRequests();
+
+            // start the specialization process in the background
+            Task.Run(async () => await Assign(context));
+
+            return true;
         }
 
         public Task<string> ValidateContext(HostAssignmentContext assignmentContext)
         {
             // Don't need to validate RunFromPackageContext in Legion
-            return null;
+            return Task.FromResult((string)null);
         }
 
         private async Task Assign(HostAssignmentContext assignmentContext)
@@ -107,7 +107,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
             {
                 // first make all environment and file system changes required for
                 // the host to be specialized
-                _logger.LogInformation("Applying {count} app setting(s)", assignmentContext.Environment.Count);
+                _logger.LogInformation("Applying {environmentCount} app setting(s)", assignmentContext.Environment.Count);
                 assignmentContext.ApplyAppSettings(_environment, _logger);
             }
             catch (Exception ex)
@@ -132,8 +132,8 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
         {
             return new Dictionary<string, string>
             {
-                { "FUNCTIONS_EXTENSION_VERSION", ScriptHost.Version },
-                { "WEBSITE_NODE_DEFAULT_VERSION", "8.5.0" }
+                { EnvironmentSettingNames.FunctionsExtensionVersion, ScriptHost.Version },
+                { EnvironmentSettingNames.WebsiteNodeDefaultVersion, "8.5.0" }
             };
         }
 
--- a/src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
@@ -90,7 +90,8 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
             _webHostEnvironment.DelayRequests();
 
             // start the specialization process in the background
-            Task.Run(async () => await Assign(context));
+            //Task.Run(async () => await Assign(context));
+            Task.Run(() => Assign(context));
 
             return true;
         }
--- a/src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
@@ -40,7 +40,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
         public Task<string> SpecializeMSISidecar(HostAssignmentContext context)
         {
             // Legion will take care of MSI Specialization
-            throw new NotImplementedException();
+            return null;
         }
 
         public bool StartAssignment(HostAssignmentContext context)
@@ -102,7 +102,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
         public Task<string> ValidateContext(HostAssignmentContext assignmentContext)
         {
             // Don't need to validate RunFromPackageContext in Legion
-            throw new NotImplementedException();
+            return null;
         }
 
         private async Task Assign(HostAssignmentContext assignmentContext)
@@ -135,7 +135,11 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
 
         public IDictionary<string, string> GetInstanceInfo()
         {
-            throw new NotImplementedException();
+            return new Dictionary<string, string>
+            {
+                { "FUNCTIONS_EXTENSION_VERSION", ScriptHost.Version },
+                { "WEBSITE_NODE_DEFAULT_VERSION", "8.5.0" }
+            };
         }
 
         // for testing
--- a/src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
@@ -1,0 +1,199 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.Http;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Script.Diagnostics;
+using Microsoft.Azure.WebJobs.Script.WebHost.Models;
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
+{
+    public class LegionInstanceManager : IInstanceManager
+    {
+        private static readonly object _assignmentLock = new object();
+        private static HostAssignmentContext _assignmentContext;
+
+        private readonly ILogger _logger;
+        private readonly IMetricsLogger _metricsLogger;
+        private readonly IMeshServiceClient _meshServiceClient;
+        private readonly IEnvironment _environment;
+        private readonly HttpClient _client;
+        private readonly IScriptWebHostEnvironment _webHostEnvironment;
+
+        public LegionInstanceManager(IHttpClientFactory httpClientFactory, IScriptWebHostEnvironment webHostEnvironment,
+            IEnvironment environment, ILogger<LegionInstanceManager> logger, IMetricsLogger metricsLogger, IMeshServiceClient meshServiceClient)
+        {
+            _client = httpClientFactory?.CreateClient() ?? throw new ArgumentNullException(nameof(httpClientFactory));
+            _webHostEnvironment = webHostEnvironment ?? throw new ArgumentNullException(nameof(webHostEnvironment));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            _metricsLogger = metricsLogger;
+            _meshServiceClient = meshServiceClient;
+            _environment = environment ?? throw new ArgumentNullException(nameof(environment));
+        }
+
+        public async Task<string> SpecializeMSISidecar(HostAssignmentContext context)
+        {
+            // No cold start optimization needed for side car scenarios
+            if (context.IsWarmupRequest)
+            {
+                return null;
+            }
+
+            var msiEnabled = context.IsMSIEnabled(out var endpoint);
+
+            _logger.LogInformation($"MSI enabled status: {msiEnabled}");
+
+            if (msiEnabled)
+            {
+                if (context.MSIContext == null)
+                {
+                    _logger.LogWarning("Skipping specialization of MSI sidecar since MSIContext was absent");
+                    await _meshServiceClient.NotifyHealthEvent(ContainerHealthEventType.Fatal, this.GetType(),
+                        "Could not specialize MSI sidecar since MSIContext was empty");
+                }
+                else
+                {
+                    using (_metricsLogger.LatencyEvent(MetricEventNames.LinuxContainerSpecializationMSIInit))
+                    {
+                        var uri = new Uri(endpoint);
+                        var address = $"http://{uri.Host}:{uri.Port}{ScriptConstants.LinuxMSISpecializationStem}";
+
+                        _logger.LogDebug($"Specializing sidecar at {address}");
+
+                        var requestMessage = new HttpRequestMessage(HttpMethod.Post, address)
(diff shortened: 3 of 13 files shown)


liliankasem:
Please update the PR name so it's more clear on what the change is, and update description to describe the changes that are being made - thanks!

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
```diff
@@ -0,0 +1,147 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Script.Diagnostics;
+using Microsoft.Azure.WebJobs.Script.WebHost.Models;
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
+{
+    public class LegionInstanceManager : IInstanceManager
```
yojagad: Where is this being injected ?

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
```diff
@@ -0,0 +1,147 @@
...
+            {
+                // first make all environment and file system changes required for
+                // the host to be specialized
+                _logger.LogInformation($"Applying {assignmentContext.Environment.Count} app setting(s)");
+                assignmentContext.ApplyAppSettings(_environment, _logger);
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex, "Assign failed");
+                await _meshServiceClient.NotifyHealthEvent(ContainerHealthEventType.Fatal, GetType(), "Assign failed");
```
yojagad: Do we still need this for now? :)

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
```diff
@@ -41,22 +41,27 @@ public async Task<IActionResult> Assign([FromBody] EncryptedHostAssignmentContex
...
-            // up front validation on the context as possible
-            string error = await _instanceManager.ValidateContext(assignmentContext);
-            if (error != null)
-            {
-                return StatusCode(StatusCodes.Status400BadRequest, error);
-            }
+            string error;
+            
+            // Skip MSI Specialization and validation for Legion
+            if (!_environment.IsLinuxConsumptionOnLegion()) {
```
balag0: we should get the correct instancemanager type injected by DI. (no explicit checks here)

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
```diff
@@ -0,0 +1,147 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
```
balag0, line 1: + unit tests

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
Tonewall: So currently, the LegionInstanceManager will be injected already, and the ValidateContext() just throws a NotImplementedException. Should I just get rid of this explicit check and just have the ValidateContext() as no-op?

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
```diff
@@ -0,0 +1,147 @@
...
+            // Legion will take care of MSI Specialization
+            throw new NotImplementedException();
+        }
+
+        public bool StartAssignment(HostAssignmentContext context)
+        {
+            if (!_webHostEnvironment.InStandbyMode)
+            {
+                // This is only true when specializing pinned containers.
+                if (!context.Environment.TryGetValue(EnvironmentSettingNames.ContainerStartContext, out string startContext))
```
liliankasem: Can this be simplified into one if-statement with the `!_webHostEnvironment.InStandbyMode` check?

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
```diff
@@ -0,0 +1,147 @@
...
+                _logger.LogError("Assign called while container is marked as specialized.");
+                return false;
+            }
+
+            if (context.IsWarmupRequest)
+            {
+                // No need to download since file already exists
+                return true;
+            }
+            else if (_assignmentContext == null)
```
liliankasem: This can be an independent if statement since the previous one returns 
```suggestion
           if (_assignmentContext == null)
```

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
```diff
@@ -0,0 +1,147 @@
...
+                // start the specialization process in the background
+                Task.Run(async () => await Assign(context));
+
+                return true;
+            }
+            else
+            {
+                // No lock needed here since _assignmentContext is not null when we are here
+                return _assignmentContext.Equals(context);
+            }
```
liliankasem:
```suggestion
            
            // No lock needed here since _assignmentContext is not null when we are here
            return _assignmentContext.Equals(context);
```

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
```diff
@@ -0,0 +1,147 @@
...
+            throw new NotImplementedException();
+        }
+
+        private async Task Assign(HostAssignmentContext assignmentContext)
+        {
+            try
+            {
+                // first make all environment and file system changes required for
+                // the host to be specialized
+                _logger.LogInformation($"Applying {assignmentContext.Environment.Count} app setting(s)");
```
liliankasem: please use structured logging 
```suggestion
                _logger.LogInformation("Applying {count} app setting(s)", assignmentContext.Environment.Count);
```

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
```diff
@@ -0,0 +1,147 @@
...
+            }
+            finally
+            {
+                // all assignment settings/files have been applied so we can flip
+                // the switch now on specialization
+                // even if there are failures applying context above, we want to
+                // leave placeholder mode
+                _logger.LogInformation("Triggering specialization");
+                _webHostEnvironment.FlagAsSpecializedAndReady();
+
```
liliankasem: nit
```suggestion
```

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
```diff
@@ -0,0 +1,147 @@
...
+        private readonly IScriptWebHostEnvironment _webHostEnvironment;
+
+        public LegionInstanceManager(IHttpClientFactory httpClientFactory, IScriptWebHostEnvironment webHostEnvironment,
+            IEnvironment environment, ILogger<LegionInstanceManager> logger, IMetricsLogger metricsLogger, IMeshServiceClient meshServiceClient)
+        {
+            _client = httpClientFactory?.CreateClient() ?? throw new ArgumentNullException(nameof(httpClientFactory));
+            _webHostEnvironment = webHostEnvironment ?? throw new ArgumentNullException(nameof(webHostEnvironment));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            _metricsLogger = metricsLogger;
+            _meshServiceClient = meshServiceClient;
```
liliankasem: Any reason we don't throw a `ArgumentNullException` for this? And the metrics logger? Usage of these items will throw if they're null

liliankasem (dismissed):

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
Tonewall: Resolved, thank you for the feedback!

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
Tonewall: Fixed!

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
Tonewall: Good point, I was using the existing instancemanager for this class, but I do think they should be required. Just made the fix!

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
Tonewall: Fixed!

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
```diff
@@ -0,0 +1,146 @@
...
+                _webHostEnvironment.FlagAsSpecializedAndReady();
+                _webHostEnvironment.ResumeRequests();
+            }
+        }
+
+        public IDictionary<string, string> GetInstanceInfo()
+        {
+            return new Dictionary<string, string>
+            {
+                { "FUNCTIONS_EXTENSION_VERSION", ScriptHost.Version },
```
liliankasem: These should exist or be added to the env constants file https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Environment/EnvironmentSettingNames.cs

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
```diff
@@ -294,6 +302,8 @@ private static void AddLinuxContainerServices(this IServiceCollection services)
                 return NullLinuxContainerActivityPublisher.Instance;
             });
 
+            
+
```
liliankasem: nit: remove new lines

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
```diff
@@ -0,0 +1,146 @@
...
+                lock (_assignmentLock)
+                {
+                    if (_assignmentContext != null)
+                    {
+                        return _assignmentContext.Equals(context);
+                    }
+                    _assignmentContext = context;
+                }
+
+                _logger.LogInformation($"Starting Assignment. Cloud Name: {_environment.GetCloudName()}");
```
liliankasem: structured logged here too + any logging statements
```suggestion
                _logger.LogInformation("Starting Assignment. Cloud Name: {name}", _environment.GetCloudName());
```

jviau: +1 - but I would recommend `{cloudName}` over `{name}`, as you need to consider what other log scopes might be active. "name" is generic and may conflict.

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
```diff
@@ -0,0 +1,146 @@
...
+                // set a flag which will cause any incoming http requests to buffer
+                // until specialization is complete
+                // the host is guaranteed not to receive any requests until AFTER assign
+                // has been initiated, so setting this flag here is sufficient to ensure
+                // that any subsequent incoming requests while the assign is in progress
+                // will be delayed until complete
+                _webHostEnvironment.DelayRequests();
+
+                // start the specialization process in the background
+                Task.Run(async () => await Assign(context));
```
jviau: Small nit: don't think you need `async`/`await` here.
```suggestion
                Task.Run(() => Assign(context));
```

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
```diff
@@ -0,0 +1,146 @@
...
+
+                return true;
+            }
+
+            // No lock needed here since _assignmentContext is not null when we are here
+            return _assignmentContext.Equals(context);
```
jviau: Small nit: if you invert the `if` statement above, you can avoid an extra level of nesting. To me it reads slightly cleaner as the structure is more procedural: all short circuits are upfront, then core logic is after all the `if` statements.

``` CSharp
if (_assignmentContext is not null)
{
    return _assignmentContext.Equals(context);
}

// rest of the logic
```

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        public Task<string> ValidateContext(HostAssignmentContext assignmentContext)
+        {
+            // Don't need to validate RunFromPackageContext in Legion
+            return null;
```
jviau: Don't return null tasks - anyone that `await`'s this will throw.

```suggestion
            return Task.FromResult((string)null); // also better to put this in a readonly static to avoid allocating a task object each call.
```

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        private async Task Assign(HostAssignmentContext assignmentContext)
+        {
+            try
+            {
+                // first make all environment and file system changes required for
+                // the host to be specialized
+                _logger.LogInformation("Applying {count} app setting(s)", assignmentContext.Environment.Count);
```
jviau: nit: `{environmentCount}` - `{count}` is not very clear _what_ is the subject of the count.

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
```diff
@@ -0,0 +1,146 @@
...
+            }
+            finally
+            {
+                // all assignment settings/files have been applied so we can flip
+                // the switch now on specialization
+                // even if there are failures applying context above, we want to
+                // leave placeholder mode
+                _logger.LogInformation("Triggering specialization");
+                _webHostEnvironment.FlagAsSpecializedAndReady();
+                _webHostEnvironment.ResumeRequests();
```
jviau: If assign fails, the `_webHostEnvironment.DelayRequests()` will never be undone - is that intentional?

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
```diff
@@ -0,0 +1,146 @@
...
+using Microsoft.Azure.WebJobs.Script.Diagnostics;
+using Microsoft.Azure.WebJobs.Script.WebHost.Models;
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
+{
+    public class LegionInstanceManager : IInstanceManager
+    {
+        private static readonly object _assignmentLock = new object();
```
jviau: This class is registered as a singleton - can we use instance fields instead and rely on singleton behavior?

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
```diff
@@ -0,0 +1,146 @@
...
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            _metricsLogger = metricsLogger ?? throw new ArgumentNullException(nameof(metricsLogger));
+            _meshServiceClient = meshServiceClient ?? throw new ArgumentNullException(nameof(meshServiceClient));
+            _environment = environment ?? throw new ArgumentNullException(nameof(environment));
+        }
+
+        public Task<string> SpecializeMSISidecar(HostAssignmentContext context)
+        {
+            // Legion will take care of MSI Specialization
+            return null;
```
jviau: Don't return null tasks - anyone that await's this will throw. See the other comment regarding this.

jviau (changes requested):
The null `Task` returning methods may cause issues.

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
Tonewall: But the Assign() method invokes the async _meshServiceClient.NotifyHealthEvent() task. Don't we need to await this task?

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
Tonewall: If assign fails, we mark the container unhealthy and it will marked for recycling and be disposed. So, I don't think it will matter.

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
Tonewall: Changed! Thank you for the feedback!

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
Tonewall: Fixed!

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
Tonewall: Fixed!

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
Tonewall: Fixed!

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
Tonewall: Fixed!

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
Tonewall: Fixed!

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
Tonewall: Fixed!

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
jviau: It will still be awaited. `Task.Run` just wants a Task-returning method passed in. `() => Assign(context)` and `async () => await Assign(context)` have the same `Task` returning signature. Async-await is a compiler directive to generate a state machine to make task continuations easier. But in this case you have no continuation (specifically, your lambda does not do any work after `Assign` is finished). So you are safe to just directly pass the returned `Task` from `Assign` up to `Task.Run` and let it await it. Now if you did something with the return, then you would have to async/await here. Something like:

``` CSharp
Task.Run(async () =>
{
    await Assign(context);
    // extra post-assign work.
});
```

yojagad (approved):

liliankasem:
@Tonewall is #8803 replacing this PR? Is so, please close it.

Tonewall:
> @Tonewall is #8803 replacing this PR? Is so, please close it.

I created the other PR to compare the different approaches, but I think Fabio's suggested approach of using the IInstanceManager interface with a base abstract class is the better approach, so I will edit this PR with that implementation

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
Tonewall: Good catch, forgot to edit the WebHostServiceCollectionExtensions.cs

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
Tonewall: I think we'd need it eventually. Should we get rid of it for right now?

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
```diff
@@ -128,12 +128,33 @@ public static void AddWebJobsScriptHost(this IServiceCollection services, IConfi
...
+            services.AddSingleton<IInstanceManager>(s =>
+            {
+                var httpClientFactory = s.GetService<IHttpClientFactory>();
+                var scriptWebHostEnvironment = s.GetService<IScriptWebHostEnvironment>();
+                var metricsLogger = s.GetService<IMetricsLogger>();
+                var meshServiceClient = s.GetService<IMeshServiceClient>();
+
+                var environment = s.GetService<IEnvironment>();
+
+                if (environment.IsLinuxConsumptionOnLegion())
```
jviau: Don't suppose there is a way for this check to be done outside of this delegate?

But if not, what we can do to simplify this logic is use `ActivatorUtilities`:

``` CSharp
services.AddSingleton<IInstanceManager>(s =>
{
    return environment.IsLinuxConsumptionOnLegion()
        ? ActivatorUtilities.CreateInstance<LegionInstanceManager>(s)
        : ActivatorUtilities.CreateInstance<InstanceManager>(s);
});

```

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
```diff
@@ -128,12 +128,33 @@ public static void AddWebJobsScriptHost(this IServiceCollection services, IConfi
...
+                    var legionLogger = s.GetService<ILogger<LegionInstanceManager>>();
+                    return new LegionInstanceManager(httpClientFactory, scriptWebHostEnvironment, environment, legionLogger, metricsLogger, meshServiceClient);
+                }
+
+                var optionsFactory = s.GetService<IOptionsFactory<ScriptApplicationHostOptions>>();
+                var logger = s.GetService<ILogger<InstanceManager>>();
+                var runFromPackageHandler = s.GetService<IRunFromPackageHandler>();
+                var packageDownloadHandler = s.GetService<IPackageDownloadHandler>();
+
+                return new InstanceManager(optionsFactory, httpClientFactory, scriptWebHostEnvironment, environment, logger, metricsLogger, meshServiceClient, runFromPackageHandler, packageDownloadHandler);
```
jviau: Can we rename `InstanceManager` to `AtlasInstanceManager`? (I am assuming that is what it represents now).

src/WebJobs.Script.WebHost/ContainerManagement/LinuxContainerInitializationHostService.cs
```diff
@@ -18,12 +18,12 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.ContainerManagement
...
     {
         private readonly IEnvironment _environment;
-        private readonly IInstanceManager _instanceManager;
+        private readonly LinuxInstanceManager _instanceManager;
         private readonly ILogger _logger;
         private readonly StartupContextProvider _startupContextProvider;
         private CancellationToken _cancellationToken;
 
-        public LinuxContainerInitializationHostService(IEnvironment environment, IInstanceManager instanceManager, ILogger<LinuxContainerInitializationHostService> logger, StartupContextProvider startupContextProvider)
+        public LinuxContainerInitializationHostService(IEnvironment environment, LinuxInstanceManager instanceManager, ILogger<LinuxContainerInitializationHostService> logger, StartupContextProvider startupContextProvider)
```
jviau: Is this change necessary? If not, can we remove it?

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
```diff
@@ -20,11 +20,11 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Controllers
     public class InstanceController : Controller
     {
         private readonly IEnvironment _environment;
-        private readonly IInstanceManager _instanceManager;
+        private readonly LinuxInstanceManager _instanceManager;
         private readonly ILogger _logger;
         private readonly StartupContextProvider _startupContextProvider;
 
-        public InstanceController(IEnvironment environment, IInstanceManager instanceManager, ILoggerFactory loggerFactory, StartupContextProvider startupContextProvider)
+        public InstanceController(IEnvironment environment, LinuxInstanceManager instanceManager, ILoggerFactory loggerFactory, StartupContextProvider startupContextProvider)
```
jviau: Same here, if this change is not necessary lets not include it.

src/WebJobs.Script.WebHost/Management/LinuxInstanceManager.cs
```diff
@@ -0,0 +1,153 @@
...
+                _logger.LogInformation("Triggering specialization");
+                _webHostEnvironment.FlagAsSpecializedAndReady();
+
+                _webHostEnvironment.ResumeRequests();
+            }
+        }
+
+        protected abstract Task ApplyContext(HostAssignmentContext assignmentContext);
+
+        protected abstract Task<string> DownloadWarmupAsync(RunFromPackageContext context);
```
jviau, line 138: small nit: "downloading" is an implementation detail of the atlas manager. Can we keep the name more generic?
```suggestion
        protected abstract Task<string> WarmupAsync(RunFromPackageContext context);
```

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
```diff
@@ -128,13 +128,27 @@ public static void AddWebJobsScriptHost(this IServiceCollection services, IConfi
...
-            services.AddSingleton<IInstanceManager, InstanceManager>();
             services.AddHttpClient();
             services.AddSingleton<StartupContextProvider>();
             services.AddSingleton<IFileSystem>(_ => FileUtility.Instance);
             services.AddTransient<VirtualFileSystem>();
             services.AddTransient<VirtualFileSystemMiddleware>();
 
+            if (SystemEnvironment.Instance.IsLinuxConsumptionOnAtlas())
+            {
+                System.Console.WriteLine("[TEST] Atlas Linux Consumption");
```
jviau: Make sure to remove these before checking in.

src/WebJobs.Script.WebHost/Management/LinuxInstanceManager.cs
```diff
@@ -0,0 +1,153 @@
...
+                // even if there are failures applying context above, we want to
+                // leave placeholder mode
+                _logger.LogInformation("Triggering specialization");
+                _webHostEnvironment.FlagAsSpecializedAndReady();
+
+                _webHostEnvironment.ResumeRequests();
+            }
+        }
+
+        protected abstract Task ApplyContext(HostAssignmentContext assignmentContext);
```
jviau: nit:
```suggestion
        protected abstract Task ApplyContextAsync(HostAssignmentContext assignmentContext);
```

src/WebJobs.Script.WebHost/Management/LinuxInstanceManager.cs
```diff
@@ -0,0 +1,153 @@
...
+            else
+            {
+                // No lock needed here since _assignmentContext is not null when we are here
+                return _assignmentContext.Equals(context);
+            }
+        }
+
+        public abstract Task<string> ValidateContext(HostAssignmentContext assignmentContext);
+
+        private async Task Assign(HostAssignmentContext assignmentContext)
```
jviau: nit:
```suggestion
        private async Task AssignAsync(HostAssignmentContext assignmentContext)
```

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
```diff
@@ -128,13 +128,27 @@ public static void AddWebJobsScriptHost(this IServiceCollection services, IConfi
...
+                services.AddSingleton<LinuxInstanceManager, InstanceManager>();
+            }
+            else if (SystemEnvironment.Instance.IsLinuxConsumptionOnLegion())
+            {
+                System.Console.WriteLine("[TEST] Legion Linux Consumption");
+                services.AddSingleton<LinuxInstanceManager, LegionInstanceManager>();
+            }
+            else
+            {
+                System.Console.WriteLine("[TEST] Else");
```
jviau: Should we throw here? Or can we have a default, like assume Atlas if unable to tell from environment? This will retain previous behavior.

Drawback of that is it will hide any issues where we expect the environment to tell us legion or atlas, but it isn't, and we just continue on assuming it is Atlas (when that may be incorrect).

This boils down to: do we have confidence all environments will correctly report legion vs atlas (and so we follow fail fast principal and throw when it is not clear), or will we require a fallback behavior here (to Atlas) when environment details are not available?

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
```diff
@@ -128,13 +128,27 @@ public static void AddWebJobsScriptHost(this IServiceCollection services, IConfi
...
             services.AddHttpClient();
             services.AddSingleton<StartupContextProvider>();
             services.AddSingleton<IFileSystem>(_ => FileUtility.Instance);
             services.AddTransient<VirtualFileSystem>();
             services.AddTransient<VirtualFileSystemMiddleware>();
 
+            if (SystemEnvironment.Instance.IsLinuxConsumptionOnAtlas())
+            {
+                System.Console.WriteLine("[TEST] Atlas Linux Consumption");
+                services.AddSingleton<LinuxInstanceManager, InstanceManager>();
```
jviau: We should retain registering this as `IInstanceManager`. Same with the legion registration below.
```suggestion
                services.AddSingleton<IInstanceManager, InstanceManager>();
```

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
```diff
@@ -222,7 +236,7 @@ private static void AddLinuxContainerServices(this IServiceCollection services)
                 //todo: Replace with legion specific service
                 if (environment.IsAnyLinuxConsumption())
                 {
-                    var instanceManager = s.GetService<IInstanceManager>();
+                    var instanceManager = s.GetService<LinuxInstanceManager>();
```
jviau: After the change above back to `IInstanceManager`, make sure to undo this change.

jviau (changes requested):
Looking much cleaner. Some small feedback to address.

src/WebJobs.Script.WebHost/ContainerManagement/LinuxContainerInitializationHostService.cs
Tonewall: Wouldn't the type for LegionInstanceManager and InstanceManager (will change to AtlasInstanceManager) be LinuxInstanceManager since both are subclasses of LinuxInstanceManager and not IInstanceManager?

jviau: `LinuxInstanceManager` implements `IInstanceManager`, so through inheritance both `LegionInstanceManager` and `AtlasInstanceManager` are `IInstanceManager`.

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
Tonewall: Made it default to AtlasInstanceManager since that was the previous InstanceManager injected. The IsLinuxConsumptionOnAtlas and IsLinuxConsumptionOnLegion checks should be accurate, but if they cause an issue, we should change the checks and leave the dependency injection logic

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
Tonewall: I was checking the environment condition before registering the service outside of the lambda as suggested in the other PR. How does the updated current dependency injection look?

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
Tonewall: Changed the design of the interface to include base class

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
```diff
@@ -128,13 +128,26 @@ public static void AddWebJobsScriptHost(this IServiceCollection services, IConfi
...
             services.AddSingleton<IFunctionMetadataProvider, HostFunctionMetadataProvider>();
             services.AddSingleton<IWebFunctionsManager, WebFunctionsManager>();
-            services.AddSingleton<IInstanceManager, InstanceManager>();
             services.AddHttpClient();
             services.AddSingleton<StartupContextProvider>();
             services.AddSingleton<IFileSystem>(_ => FileUtility.Instance);
             services.AddTransient<VirtualFileSystem>();
             services.AddTransient<VirtualFileSystemMiddleware>();
 
+            if (SystemEnvironment.Instance.IsLinuxConsumptionOnAtlas())
```
jviau: This statement is not needed if the `else` statement will perform the same logic. We can boil it down to:

``` CSharp
if (SystemEnvironment.Instance.IsLinuxConsumptionOnLegion())
{
    services.AddSingleton<IInstanceManager, LegionInstanceManager>();
 }
else
{
    // Default IInstanceManager
    services.AddSingleton<IInstanceManager, AtlasInstanceManager>();
}
```

src/WebJobs.Script.WebHost/Management/LinuxInstanceManager.cs
```diff
@@ -0,0 +1,153 @@
...
+using Microsoft.Azure.WebJobs.Script.WebHost.Models;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
+{
+    public abstract class LinuxInstanceManager : IInstanceManager
+    {
+        private static readonly object _assignmentLock = new object();
```
jviau: For a later PR, it would be nice to remove these `static`. These services are registered as a singleton, I don't see why they should be using static fields over instance fields.

jviau (approved):

jviau:
Lets also get a review from @fabiocav if he has time.

src/WebJobs.Script.WebHost/Management/LinuxInstanceManager.cs
fabiocav: Marking as unresolved to ask: is there a reason this shouldn't be done as part of this PR?

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
```diff
@@ -0,0 +1,47 @@
...
+    public class LegionInstanceManager : LinuxInstanceManager
+    {
+        public LegionInstanceManager(IHttpClientFactory httpClientFactory, IScriptWebHostEnvironment webHostEnvironment,
+            IEnvironment environment, ILogger<LegionInstanceManager> logger, IMetricsLogger metricsLogger, IMeshServiceClient meshServiceClient) : base(httpClientFactory, webHostEnvironment,
+            environment, logger, metricsLogger, meshServiceClient) { }
+
+        public override Task<string> SpecializeMSISidecar(HostAssignmentContext context)
+        {
+            // Legion will take care of MSI Specialization
+            return Task.FromResult((string)null);
```
fabiocav: nit: `Task.FromResult((string)null);` -> `Task.FromResult<string>(null);` here and in the other methods.

fabiocav (approved):
Left a couple of comments/questions, but nothing blocking.

balag0 (approved):

src/WebJobs.Script.WebHost/Management/LegionInstanceManager.cs
Tonewall: Fixed, thank you for your review!

src/WebJobs.Script.WebHost/Management/LinuxInstanceManager.cs
Tonewall: Fixed, thank you for your review!

src/WebJobs.Script.WebHost/Management/LinuxInstanceManager.cs
```diff
@@ -0,0 +1,153 @@
...
+
+        private HostAssignmentContext _assignmentContext;
+
+        public LinuxInstanceManager(IHttpClientFactory httpClientFactory, IScriptWebHostEnvironment webHostEnvironment,
+            IEnvironment environment, ILogger<LinuxInstanceManager> logger, IMetricsLogger metricsLogger, IMeshServiceClient meshServiceClient)
+        {
+            _client = httpClientFactory?.CreateClient() ?? throw new ArgumentNullException(nameof(httpClientFactory));
+            _webHostEnvironment = webHostEnvironment ?? throw new ArgumentNullException(nameof(webHostEnvironment));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            _meshServiceClient = meshServiceClient;
```
liliankasem, line 41: Do we not want to handle if this is null too? Or we don't care?

src/WebJobs.Script.WebHost/Management/LinuxInstanceManager.cs
```diff
@@ -0,0 +1,153 @@
...
+        protected abstract Task ApplyContextAsync(HostAssignmentContext assignmentContext);
+
+        protected abstract Task<string> DownloadWarmupAsync(RunFromPackageContext context);
+
+        public IDictionary<string, string> GetInstanceInfo()
+        {
+            return new Dictionary<string, string>
+            {
+                { EnvironmentSettingNames.FunctionsExtensionVersion, ScriptHost.Version },
+                { EnvironmentSettingNames.WebsiteNodeDefaultVersion, "8.5.0" }
```
liliankasem: Can we make this a const variable somewhere?

liliankasem (approved):
2 quick comments otherwise lgtm

src/WebJobs.Script.WebHost/Management/LinuxInstanceManager.cs
Tonewall: Fixed! Thank you for your review!

src/WebJobs.Script.WebHost/Management/LinuxInstanceManager.cs
Tonewall: Yes, there are some instances when we inject a NullMeshServiceClient ([here](https://github.com/Azure/azure-functions-host/blob/ba112b76a93ffc145690fb9015eef41bcab71d88/src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs#L271)). So, we wouldn't want to throw for those cases

PR #8642, opened 2022-08-18
Include empty entries when building InvocationRequest payload for OOP workers

### LINKED ISSUE #8499: Introduce new app setting to control skipping null entries from event payload when sending to OOP layer
Background:
For out-of-process functions, the host code gets the trigger data from the listener/processor module & then we convert the event data object [to RPC invocation request](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.Grpc/Extensions/ScriptInvocationContextExtensions.cs#L25) instance which we will send to the out-of-process worker via GRPC. In the current implementation, we are skipping the empty entries in the [ToRpcStringArray method](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs#L276) which gets called internally inside [ToRpcInvocationRequest ](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.Grpc/Extensions/ScriptInvocationContextExtensions.cs#L25) method.

Empty-Entry-Skipped-ToRpc 

This behavior causes confusion when dealing with batched event processing where the event body of one of the entries in the batch is `null`. The above screenshot is taken from an example where we are processing event hub messages(batched) where one of the message has null body, but the message still has valid meta properties. The oop worker will not be able to associate the meta properties of the message payload entries (message body) with the meta properties array.  See [this issue](https://github.com/Azure/azure-functions-dotnet-worker/issues/876) for a complete example.

Ask:

We need to introduce a new app setting, which when set to True, will skip the code which do the[ _null check and discard_](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs#L282) codepath. Default value of this setting will be "False", so no existing apps are impacted.

kshyju:
Fixes #8499 

The current behavior is to skip entries which are null or empty. This causes issues like  [this](https://github.com/Azure/azure-functions-dotnet-worker/issues/876). So we decided to introduce a new capability called `IncludeEmptyEntriesInMessagePayload` which language workers can advertise and if it is present, we will include the empty entries (empty, not null) in the array we create for the message payload.

### Pull request checklist

    * [x] Otherwise: Backport tracked by same issue. I will add once this change is in v4.

### Additional information

Additional PR information
Please refer #8499 which has details about the current behavior and problems caused by that.

### DIFF
--- a/src/WebJobs.Script/Workers/Rpc/RpcWorkerConstants.cs
+++ b/src/WebJobs.Script/Workers/Rpc/RpcWorkerConstants.cs
@@ -55,6 +55,11 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
         public const string HandlesWorkerTerminateMessage = "HandlesWorkerTerminateMessage";
         public const string HandlesInvocationCancelMessage = "HandlesInvocationCancelMessage";
 
+        /// <summary>
+        /// Indicates whether empty entries in the trigger message should be included when sending RpcInvocation data to OOP workers.
+        /// </summary>
+        public const string IncludeEmptyEntriesInMessagePayload = "IncludeEmptyEntriesInMessagePayload";
+
         // Host Capabilities
         public const string V2Compatable = "V2Compatable";
 
--- a/src/WebJobs.Script/Workers/Rpc/RpcWorkerConstants.cs
+++ b/src/WebJobs.Script/Workers/Rpc/RpcWorkerConstants.cs
@@ -55,11 +55,6 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
         public const string HandlesWorkerTerminateMessage = "HandlesWorkerTerminateMessage";
         public const string HandlesInvocationCancelMessage = "HandlesInvocationCancelMessage";
 
-        /// <summary>
-        /// Indicates whether empty entries in the trigger message should be included when sending RpcInvocation data to OOP workers.
-        /// </summary>
-        public const string IncludeEmptyEntriesInMessagePayload = "IncludeEmptyEntriesInMessagePayload";
-
         // Host Capabilities
         public const string V2Compatable = "V2Compatable";
 
--- a/test/WebJobs.Script.Tests/Workers/Rpc/GrpcMessageConversionExtensionsTests.cs
+++ b/test/WebJobs.Script.Tests/Workers/Rpc/GrpcMessageConversionExtensionsTests.cs
@@ -486,6 +486,41 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.Workers.Rpc
         }
 
         [Fact]
+        public async Task ToRpc_Collection_String_IgnoreEmptyEntries_When_Capability_Is_Not_Present()
+        {
+            var logger = MockNullLoggerFactory.CreateLogger();
+            var capabilities = new GrpcCapabilities(logger);
+            capabilities.UpdateCapabilities(new MapField<string, string>
+                {
+                    { RpcWorkerConstants.TypedDataCollection, bool.TrueString },
+                });
+
+            string[] arrString = { "element1", string.Empty, "element_2" };
+            TypedData actual = await arrString.ToRpc(logger, capabilities);
+
+            var expected = new RepeatedField<string> { "element1", "element_2" };
+            Assert.Equal(expected, actual.CollectionString.String);
+        }
+
+        [Fact]
+        public async Task ToRpc_Collection_String_IncludeEmptyEntries_When_Capability_Is_Present()
+        {
+            var logger = MockNullLoggerFactory.CreateLogger();
+            var capabilities = new GrpcCapabilities(logger);
+            capabilities.UpdateCapabilities(new MapField<string, string>
+                {
+                    { RpcWorkerConstants.TypedDataCollection, bool.TrueString },
+                    { RpcWorkerConstants.IncludeEmptyEntriesInMessagePayload, bool.TrueString }
+                });
+
+            string[] arrString = { "element1", string.Empty, "element_2", null };
+            TypedData actual = await arrString.ToRpc(logger, capabilities);
+
+            var expected = new RepeatedField<string> { "element1", string.Empty, "element_2" }; // null entry should be still skipped
+            Assert.Equal(expected, actual.CollectionString.String);
+        }
+
+        [Fact]
         public async Task ToRpc_Collection_Long_With_Capabilities_Value()
         {
             var logger = MockNullLoggerFactory.CreateLogger();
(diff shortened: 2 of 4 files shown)


src/WebJobs.Script/Workers/Rpc/RpcWorkerConstants.cs
```diff
@@ -55,6 +55,11 @@ public static class RpcWorkerConstants
         public const string HandlesWorkerTerminateMessage = "HandlesWorkerTerminateMessage";
         public const string HandlesInvocationCancelMessage = "HandlesInvocationCancelMessage";
 
+        /// <summary>
+        /// Indicates whether empty entries in the trigger message should be included when sending RpcInvocation data to OOP workers.
+        /// </summary>
+        public const string IncludeEmptyEntriesInMessagePayload = "IncludeEmptyEntriesInMessagePayload";
```
kshyju, line 61: This is a new capability and none of the worker are sending it now. We will start with dotnet worker and then share the info with other language workers so that they can also opt in for this behavior.

mathewc: Are you saying that if we simply updated the ToRpcStringArray method to correctly add nulls with full fidelity, existing workers won't be able to handle that? Can you link to the corresponding worker code showing where the null-ref occurs?

kshyju: > Are you saying that if we simply updated the ToRpcStringArray method to correctly add nulls with full fidelity, existing workers won't be able to handle that? Can you link to the corresponding worker code showing where the null-ref occurs?

No. I did not mean that.  The fix is correcting the current incorrect behavior/bug. So probably do not need to be considered as a breaking change or so. I will remove the capability check.

kshyju: Latest iteration has this capability check removed.

liliankasem (approved):

src/WebJobs.Script/Workers/Rpc/RpcWorkerConstants.cs
mathewc: So we'll now send along empty entries, but still skip nulls. For empty entries, are we sure workers handle this w/o breaking changes? Probably need to verify with some testing for the triggers supporting cardinality many (EventHubs, ServiceBus Batches).

This will address the linked issue (empty strings) but what about nulls? Is it possible for the triggers to send nulls here? If so, in those cases, we'll have the same issue.

kshyju: null cannot be added as `RepeatedField<string>.Add` will throw when adding a null entry.

The original idea of having the capability check was to prevent this being a breaking change for workers when host version update (Even though the change is to fix a bug/incorrect behavior). Do you think we should go back to that route to be safe/reduce potential impact? I am assuming none of the worker implementations having specific code to handle the current incorrect data (without empty entries) coming in. If they do, we bite the bullet with this change(host version change) and fix it now? 

I tested with EventHubs batching use case in dotnet isolated worker and it is behaving correctly.(the original incorrect behavior issue was reported in dotnet worker as mentioned

mathewc: If we can verify that for all of the language workers as they exist now they correctly handle the new empty array elements we may send them without blowing up, then we don't need a capability flag I think.

wesselkranenborg: Is it really that this cannot be null? We are using `Microsoft.Azure.Devices.Client.Message` without a body (there is only constructor where you don't need to specify a body). We only set properties and a messageId on this Message. That IoTHub message is then routed to this eventhub. Is the default body of an IoT Hub message `string.Empty`?

kshyju: Language worker testing is currently blocked due to [this issue](https://github.com/Azure/azure-functions-python-worker/issues/1089) on python.  Tested JS, TS and Dotnetworker and they look good.

test/WebJobs.Script.Tests/Workers/Rpc/GrpcMessageConversionExtensionsTests.cs
```diff
@@ -485,6 +485,23 @@ public async Task ToRpc_Collection_String_Without_Capabilities_Value()
             Assert.Equal(typedData.Json, returned_typedata.Json);
         }
 
+        [Fact]
+        public async Task ToRpc_Collection_String_IncludeEmptyEntries_for_TypedDataCollection()
```
madelinegordon: nit - name seems a little disjointed? Is "Rpc collection" a single concept? (e.g. would "ToRpcCollection_String_IncludeEmptyEntriesForTypedDataCollection" make sense?)

src/WebJobs.Script/Workers/Rpc/RpcWorkerConstants.cs
fabiocav: Historically, any behavior change like this is based on capability advertisement. That’s the safest way if you want to expedite the fix. If you don’t want to introduce a capability, thorough testing of all language workers will be required, and that wouldn’t cover external ones other teams may be maintaining.

is there a concern about the new capability?

kshyju: Brought back the capability check based on offline sync. Thank you!

satvu (approved):

test/WebJobs.Script.Tests/Workers/Rpc/GrpcMessageConversionExtensionsTests.cs
kshyju: ToRpc is the method name and other tests are using this format. So I will keep it as it is for consistency.

src/WebJobs.Script/Workers/Rpc/RpcWorkerConstants.cs
wesselkranenborg: We are now testing this bugfix in our azure function but it looks like we are still hitting this issue. Nobody also replied to my comment above...

> Is it really that this cannot be null? We are using `Microsoft.Azure.Devices.Client.Message` without a body (there is only constructor where you don't need to specify a body). We only set properties and a messageId on this Message. That IoTHub message is then routed to this eventhub. Is the default body of an IoT Hub message `string.Empty`?

It looks like our payload was just null and this bugfix is only helping when you are sending string.Empty. Which `Microsoft.Azure.Devices.Client.Message` doesn't do by default.

CC @kshyju, @fabiocav

PR #8680, opened 2022-08-30
Set environment variable to indicate an application started from placeholder mode

### LINKED ISSUE #8638: Worker Profiles - Investigate if the placeholder env var is sufficient for detecting placeholder mode
`WEBSITE_PLACEHOLDER_MODE` is expected to change from 1 to 0 when host instance starts in placeholder mode and then goes through specialization. This is still a good reliable setting to rely on. Can you change the condition to check if `WEBSITE_PLACEHOLDER_MODE` exists instead of checking for its value

Investigate to answer the following question:

- [x] Is it possible to use a worker profile condition to check if `WEBSITE_PLACEHOLDER_MODE` exists instead of checking for its value?
- [ ] Validate if using `WEBSITE_PLACEHOLDER_MODE` (as mentioned ^) loads the correct profile and expected behaviour (the java process starts with the agent for placeholders)

If we cannot check if an environment variable exists or not, we may need to create a new profile condition that enables us to check for that.

If we cannot use `WEBSITE_PLACEHOLDER_MODE` to validate a site is in placeholder mode, we will need to design another approach for this.

liliankasem:

resolves #8638 

New environment variable `INITIALIZED_FROM_PLACEHOLDER` to indicate that a function app was started from placeholder mode.

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/Standby/StandbyManager.cs
+++ b/src/WebJobs.Script.WebHost/Standby/StandbyManager.cs
@@ -166,6 +166,9 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
                     {
                         await CreateStandbyWarmupFunctions();
 
+                        // Flag to indicate a function was initialized from placeholder mode
+                        _environment.SetEnvironmentVariable(EnvironmentSettingNames.InitializedFromPlaceholder, "true");
+
                         // start a background timer to identify when specialization happens
                         // specialization usually happens via an http request (e.g. scale controller
                         // ping) but this timer is started as well to handle cases where we
--- a/src/WebJobs.Script.WebHost/Standby/StandbyManager.cs
+++ b/src/WebJobs.Script.WebHost/Standby/StandbyManager.cs
@@ -164,10 +164,10 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
                 {
                     try
                     {
-                        await CreateStandbyWarmupFunctions();
-
                         // Flag to indicate a function was initialized from placeholder mode
-                        _environment.SetEnvironmentVariable(EnvironmentSettingNames.InitializedFromPlaceholder, "true");
+                        _environment.SetEnvironmentVariable(EnvironmentSettingNames.InitializedFromPlaceholder, bool.TrueString);
+
+                        await CreateStandbyWarmupFunctions();
 
                         // start a background timer to identify when specialization happens
                         // specialization usually happens via an http request (e.g. scale controller
--- a/test/WebJobs.Script.Tests/StandbyManagerTests.cs
+++ b/test/WebJobs.Script.Tests/StandbyManagerTests.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
+using System;
 using System.Linq;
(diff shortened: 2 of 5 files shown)


fabiocav (approved):

fabiocav (commented):
Followed up with @liliankasem on the test failures

test/WebJobs.Script.Tests/StandbyManagerTests.cs
```diff
@@ -120,6 +121,20 @@ public async Task Specialize_StandbyManagerInitialize_EmitsExpectedMetric()
             Assert.True(metricsLogger.EventsBegan.Contains(MetricEventNames.SpecializationStandbyManagerInitialize) && metricsLogger.EventsEnded.Contains(MetricEventNames.SpecializationStandbyManagerInitialize));
         }
 
+        [Fact]
+        public async Task Specialize_StandbyManagerInitialize_SetsInitializedFromPlaceholderFlag()
+        {
+            TestMetricsLogger metricsLogger = new TestMetricsLogger();
+            _testEnvironment.SetEnvironmentVariable(EnvironmentSettingNames.AzureWebsiteHostName, "placeholder.azurewebsites.net");
```
jviau, line 128: This probably has no impact, but I make a habit of always using `.test` domain in unit tests. This TLD is specifically reserved for testing software, and no website actually sits behind it. There have been instances of exploiting unit tests by someone using a placeholder domain, then a bad actor actually spins up that domain and a unit test runs against it, and then a malicious payload is sent via the domain.

No impact here I imagine because you are not actually connecting to it. But thought I'd recommend a habit I have.

surgupta-msft (approved):

jviau (approved):

test/WebJobs.Script.Tests/StandbyManagerTests.cs
liliankasem: I didn't know that was a thing, thank you for sharing! Should be fine here as it's a unit test/not actually used. But will defo use this information for future tests!

src/WebJobs.Script.WebHost/Standby/StandbyManager.cs
```diff
@@ -164,6 +164,9 @@ public async Task InitializeAsync()
                 {
                     try
                     {
+                        // Flag to indicate a function was initialized from placeholder mode
+                        _environment.SetEnvironmentVariable(EnvironmentSettingNames.InitializedFromPlaceholder, bool.TrueString);
+
```
kshyju, line 169: Q: Is there value in logging this info to our logs?
Q: Do we know what all code path should be updated to use this new env variable to make things reliable?

kshyju (approved):

src/WebJobs.Script.WebHost/Standby/StandbyManager.cs
liliankasem: 1. I don't think so, we kind of already know when specialization is happening, we have logs that show up in kusto to let us know this. The only reason this is introduced is because the placeholder mode variable is removed by DWAS after specialization, and the java team needs a variable to let them know this app was started from placeholder. 
2. No host code path needs this variable, it's just for the java team that is writing a worker profile condition that only activates when an application is started from placeholder

kshyju: Thank you!

PR #8695, opened 2022-09-01
Don't call TypeReference.Resolve for system attributes

### LINKED ISSUE #8597: ExtensionMetadataGenerator hangs when calling TypeReference.Resolve on some system attributes
This line of code hangs when running extension metadata generator on certain system attributes.

https://github.com/Azure/azure-functions-host/blob/b5e21770905343a3a58a6614c0a32ec3d80f1860/tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator.Console/ExtensionsMetadataGenerator.cs#L95

#### Repro
Call ExtensionsMetadataGenerator.Generate on the assemblies in [drop.zip](https://github.com/Azure/azure-functions-host/files/9238646/drop.zip)

soninaren:

resolves #8597

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator.Console/ExtensionsMetadataGenerator.cs
+++ b/tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator.Console/ExtensionsMetadataGenerator.cs
@@ -90,6 +90,11 @@ namespace ExtensionsMetadataGenerator
                     return true;
                 }
 
+                //if (currentAttributeType.FullName.StartsWith("System", StringComparison.OrdinalIgnoreCase))
+                //{
+                //    return false;
+                //}
+
                 try
                 {
                     currentAttributeType = currentAttributeType.Resolve()?.BaseType;
--- a/tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator.Console/ExtensionsMetadataGenerator.cs
+++ b/tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator.Console/ExtensionsMetadataGenerator.cs
@@ -90,10 +90,10 @@ namespace ExtensionsMetadataGenerator
                     return true;
                 }
 
-                //if (currentAttributeType.FullName.StartsWith("System", StringComparison.OrdinalIgnoreCase))
-                //{
-                //    return false;
-                //}
+                if (currentAttributeType.FullName.StartsWith("System", StringComparison.OrdinalIgnoreCase))
+                {
+                    return false;
+                }
 
                 try
                 {


liliankasem (approved):

tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator.Console/ExtensionsMetadataGenerator.cs
```diff
@@ -90,6 +90,11 @@ public static bool IsWebJobsStartupAttributeType(TypeReference attributeType, Co
                     return true;
                 }
 
+                if (currentAttributeType.FullName.StartsWith("System", StringComparison.OrdinalIgnoreCase))
```
jviau, line 93: Are we okay with only checking the name? Anyone can write a type in the `System` namespace. Will that cause any issues?

I am not that familiar with Mono, but I wonder if there is a stricter way to verify the assembly? But also, Is that effort even worth it when your current check will get the job done.

jviau (approved):

tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator.Console/ExtensionsMetadataGenerator.cs
soninaren: Checking for System namespace is fine, because all we are looking for is webjobs attributes. We should be fine as long as we don't write to system or webjobs namespace for webjobs attributes.

brettsam, line 93: Which assembly, and type did you hit this with? I'm curious whether there's some class of types that we cannot call `Resolve()` on. There may be a better check than just whether they begin with "System".

plukawski: Good day, I have faced exactly same problem. For me it hanged on `System.ComponentModel.DataAnnotations.KeyAttribute` (project is based on .Net Framework 4.8). It took me a while to figure out what is going on. Is there any workaround for this until this fix is released?
What I have found is that at least for me this is hanging on every attribute from `System.ComponentModel.DataAnnotations` namespace.

plukawski (commented):
I want also to notice that in line 135:
https://github.com/Azure/azure-functions-host/blob/2312529abfd6773f6c18ff4708ab37c036a3a123/tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator.Console/ExtensionsMetadataGenerator.cs#L135

Is missing null check - the line below that line for me on some project throws Null Reference exception. Because the Value property is sometimes null.

soninaren:
Closing this one in favor of #8772

PR #8708, opened 2022-09-04
Add log to understand why background synctrigger is skipped

pragnagopa:

resolves #issue_for_this_pr

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/FunctionsSyncService.cs
+++ b/src/WebJobs.Script.WebHost/FunctionsSyncService.cs
@@ -42,8 +42,10 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
         {
             get
             {
-                return _primaryHostStateProvider.IsPrimary &&
+                var shouldSyncTrigger = _primaryHostStateProvider.IsPrimary &&
                        (_scriptHostManager.State == ScriptHostState.Running);
+                _logger.LogDebug("ShouldSyncTriggers: '{shouldSyncTrigger}',PrimaryHostStateProvider.IsPrimary: '{primaryHostStateProvider.IsPrimary}',ScriptHostState: '{scriptHostManager.State}'", shouldSyncTrigger, _primaryHostStateProvider.IsPrimary, _scriptHostManager.State);
+                return shouldSyncTrigger;
             }
         }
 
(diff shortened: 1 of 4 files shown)


src/WebJobs.Script.WebHost/FunctionsSyncService.cs
```diff
@@ -42,8 +42,10 @@ internal bool ShouldSyncTriggers
         {
             get
             {
-                return _primaryHostStateProvider.IsPrimary &&
+                var shouldSyncTrigger = _primaryHostStateProvider.IsPrimary &&
                        (_scriptHostManager.State == ScriptHostState.Running);
+                _logger.LogDebug("ShouldSyncTriggers: '{shouldSyncTrigger}',PrimaryHostStateProvider.IsPrimary: '{primaryHostStateProvider.IsPrimary}',ScriptHostState: '{scriptHostManager.State}'", shouldSyncTrigger, _primaryHostStateProvider.IsPrimary, _scriptHostManager.State);
```
kshyju, line 47: Should we use LoggerMessage.Define approach for this log entry? looks like this version causes boxing for all 3 params passed.

kshyju (approved):

liliankasem (approved):

src/WebJobs.Script.WebHost/FunctionsSyncService.cs
mathewc, line 47: I don't like that we're writing a log each time this property is accessed. For example, currently it's accessed in a timer callback. Perhaps now that's only invoked once, but in the future this property may be used in other ways, and we don't want a log entry each time it's called. Do we really need this log - do you think it will help? There are lots of other reasons a background sync might not be done, e.g.

- Instance isn’t primary
- Instance isn’t in a running state after the startup delay elapses
- IsSyncTriggersEnvironment returns false
- No storage account configured for hash storage

So the utility of this log isn't clear to me.

kshyju: As of now, the caller of this prop is also in the same class. So all the information is also available for the caller method (`OnSyncTimerTick`). The logging can be moved from property get to the caller(`OnSyncTimerTick`).

PR #8721, opened 2022-09-08
moving BlobStorageSecretsRepository to use UploadAsync instead of OpenWriteAsync

brettsam:
Using `OpenWriteAsync` immediately clears the blob while blocks are written. This can lead to any reads of this blob seeing an empty file while a write is in progress. In this case, the SecretManager will believe, incorrectly, that there are no secrets and immediately regenerate them.

`UploadAsync` is an atomic operation, meaning that any reads will continue to see the last fully uploaded blob.

See here: https://github.com/Azure/azure-sdk-for-net/issues/29518

### DIFF
--- a/src/WebJobs.Script.WebHost/Security/KeyManagement/BlobStorageSecretsRepository.cs
+++ b/src/WebJobs.Script.WebHost/Security/KeyManagement/BlobStorageSecretsRepository.cs
@@ -5,9 +5,9 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Threading.Tasks;
+using Azure;
 using Azure.Storage.Blobs;
 using Azure.Storage.Blobs.Models;
-using Azure.Storage.Blobs.Specialized;
 using Microsoft.Azure.WebJobs.Host.Storage;
 using Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics.Extensions;
 using Microsoft.Extensions.Logging;
@@ -173,11 +173,22 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
 
         private async Task WriteToBlobAsync(string blobPath, string secretsContent)
         {
-            BlockBlobClient secretBlobClient = Container.GetBlockBlobClient(blobPath);
-            using (StreamWriter writer = new StreamWriter(await secretBlobClient.OpenWriteAsync(true)))
+            BlobClient secretBlobClient = Container.GetBlobClient(blobPath);
+            BlobUploadOptions uploadOptions = new BlobUploadOptions();
+
+            if (await secretBlobClient.ExistsAsync())
             {
-                await writer.WriteAsync(secretsContent);
+                // Return a 412 if another write beats us to updating the file.
+                BlobProperties properties = await secretBlobClient.GetPropertiesAsync();
+                uploadOptions.Conditions = new BlobRequestConditions { IfMatch = properties.ETag };
             }
+            else
+            {
+                // Return a 409 if another write beats us to creating the file.
+                uploadOptions.Conditions = new BlobRequestConditions { IfNoneMatch = ETag.All };
+            }
+
+            await secretBlobClient.UploadAsync(BinaryData.FromString(secretsContent), uploadOptions);
         }
 
         protected virtual void LogErrorMessage(string operation, Exception exception)
--- a/test/WebJobs.Script.Tests.Integration/Host/SecretsRepositoryTests.cs
+++ b/test/WebJobs.Script.Tests.Integration/Host/SecretsRepositoryTests.cs
@@ -10,7 +10,6 @@ using System.Threading.Tasks;
 using Azure;
 using Azure.Identity;
 using Azure.Security.KeyVault.Secrets;
-using Microsoft.Azure.Services.AppAuthentication;
 using Microsoft.Azure.Storage;
 using Microsoft.Azure.Storage.Blob;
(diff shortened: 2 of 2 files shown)


soninaren (approved):

test/WebJobs.Script.Tests.Integration/Host/SecretsRepositoryTests.cs
```diff
@@ -370,6 +369,133 @@ public async Task GetSecretSnapshots_ReturnsExpected(SecretsRepositoryType repos
             }
         }
 
+        [Theory]
+        [InlineData(SecretsRepositoryType.BlobStorage)]
+        [InlineData(SecretsRepositoryType.BlobStorageSas)]
+        public async Task BlobRepository_WriteAsync_DoesNot_ClearBlobContents(SecretsRepositoryType repositoryType)
```
brettsam, line 375: With the previous code in `WriteToBlobAsync()`, this test failed 100% of the time locally because it would find null secrets when Reading.

test/WebJobs.Script.Tests.Integration/Host/SecretsRepositoryTests.cs
```diff
@@ -370,6 +369,133 @@ public async Task GetSecretSnapshots_ReturnsExpected(SecretsRepositoryType repos
...
+                Assert.Equal(secretsContent.FunctionKeys[0].Value, "test");
+                Assert.Equal(secretsContent.SystemKeys[0].Name, KeyName);
+                Assert.Equal(secretsContent.SystemKeys[0].Value, "test");
+            }
+        }
+
+        [Theory]
+        [InlineData(SecretsRepositoryType.BlobStorage)]
+        [InlineData(SecretsRepositoryType.BlobStorageSas)]
+        public async Task BlobRepository_SimultaneousWrites_Throws_PreconditionFailed(SecretsRepositoryType repositoryType)
```
brettsam, line 415: With the previous code in `WriteToBlobAsync()`, this test behaved the same. I wanted to ensure no change in behavior.

test/WebJobs.Script.Tests.Integration/Host/SecretsRepositoryTests.cs
```diff
@@ -370,6 +369,133 @@ public async Task GetSecretSnapshots_ReturnsExpected(SecretsRepositoryType repos
...
+                Assert.Equal(412, ex.Status);
+                Assert.True(writeTask1.IsCompletedSuccessfully || writeTask2.IsCompletedSuccessfully,
+                    "One of the write operations should have completed successfully.");
+            }
+        }
+
+        [Theory]
+        [InlineData(SecretsRepositoryType.BlobStorage)]
+        [InlineData(SecretsRepositoryType.BlobStorageSas)]
+        public async Task BlobRepository_SimultaneousCreates_Throws_Conflict(SecretsRepositoryType repositoryType)
```
brettsam, line 460: With the previous code in `WriteToBlobAsync()`, this test behaved slightly different -- rather than throw a 409, it threw a 412. I don't see this as a problem, just wanted to call it out.

test/WebJobs.Script.Tests.Integration/Host/SecretsRepositoryTests.cs
```diff
@@ -370,6 +369,133 @@ public async Task GetSecretSnapshots_ReturnsExpected(SecretsRepositoryType repos
...
+                };
+
+                string testFunctionName = "host";
+
+                var target = _fixture.GetNewSecretRepository();
+
+                // Set up initial secrets.
+                await _fixture.WriteSecret(testFunctionName, testSecrets);
+
+                // Perform a write and read similtaneously. Previously, our usage of OpenWriteAsync
```
kshyju, line 395: nit: typo

kshyju (approved):

mathewc (approved):

src/WebJobs.Script.WebHost/Security/KeyManagement/BlobStorageSecretsRepository.cs
```diff
@@ -173,11 +173,22 @@ private string GetSecretsBlobPath(ScriptSecretsType secretsType, string function
...
         private async Task WriteToBlobAsync(string blobPath, string secretsContent)
         {
-            BlockBlobClient secretBlobClient = Container.GetBlockBlobClient(blobPath);
-            using (StreamWriter writer = new StreamWriter(await secretBlobClient.OpenWriteAsync(true)))
+            BlobClient secretBlobClient = Container.GetBlobClient(blobPath);
+            BlobUploadOptions uploadOptions = new BlobUploadOptions();
+
+            if (await secretBlobClient.ExistsAsync())
+            {
+                // Return a 412 if another write beats us to updating the file.
```
fabiocav, line 181: It would be good to track work to detect those conditions and better handle the conflict scenarios. This is a huge improvement, but will still cause an exception that will bubble up here and impact an instance of the host, or at least an API call. 

That enhancement would be outside of the scope of this PR, but it would be great if we handle the conflicts and cause the manager to load the new secrets instead of throwing. Would you mind opening an issue to discuss?

fabiocav (approved):

src/WebJobs.Script.WebHost/Security/KeyManagement/BlobStorageSecretsRepository.cs
brettsam: Good point. I filed https://github.com/Azure/azure-functions-host/issues/8731 to track.

PR #8722, opened 2022-09-08
Worker profile cleanup

liliankasem:
Just tidying up as I work on the active profile bug fix, figured better to check this in separately

### DIFF
--- a/src/WebJobs.Script/Workers/Profiles/Conditions/HostPropertyCondition.cs
+++ b/src/WebJobs.Script/Workers/Profiles/Conditions/HostPropertyCondition.cs
@@ -6,10 +6,9 @@ using System.ComponentModel.DataAnnotations;
 using System.Linq;
 using System.Text.RegularExpressions;
 using Microsoft.Azure.WebJobs.Script.Config;
-using Microsoft.Azure.WebJobs.Script.Workers.Profiles;
 using Microsoft.Extensions.Logging;
 
-namespace Microsoft.Azure.WebJobs.Script.Workers
+namespace Microsoft.Azure.WebJobs.Script.Workers.Profiles
 {
     /// <summary>
     /// An implementation of an <see cref="IWorkerProfileCondition"/> that checks if different host properties
@@ -68,27 +67,27 @@ namespace Microsoft.Azure.WebJobs.Script.Workers
                 return false;
             }
 
-            _logger.LogDebug($"Evaluating HostPropertyCondition with name: {Name}, value: {value} and expression {Expression}");
+            _logger.LogDebug("Evaluating HostPropertyCondition with name: {name}, value: {value} and expression {expression}", Name, value, Expression);
 
             return _regex.IsMatch(value);
         }
 
-        // Validates if condition parametrs meet expected values, fail if they don't
+        // Validates if condition parameters meet expected values, fail if they don't
         private void Validate()
         {
             if (string.IsNullOrEmpty(Name))
             {
-               throw new ValidationException($"HostPropertyCondition {nameof(Name)} cannot be empty.");
+                throw new ValidationException($"HostPropertyCondition {nameof(Name)} cannot be empty.");
             }
 
             if (!Enum.GetNames(typeof(HostProperty)).Any(x => x.ToLower().Contains(Name.ToLower())))
             {
-               throw new ValidationException($"HostPropertyCondition {nameof(Name)} is not a valid host property name.");
+                throw new ValidationException($"HostPropertyCondition {nameof(Name)} is not a valid host property name.");
             }
 
             if (string.IsNullOrEmpty(Expression))
             {
-               throw new ValidationException($"HostPropertyCondition {nameof(Expression)} cannot be empty.");
+                throw new ValidationException($"HostPropertyCondition {nameof(Expression)} cannot be empty.");
             }
 
             try
@@ -97,7 +96,7 @@ namespace Microsoft.Azure.WebJobs.Script.Workers
             }
             catch
             {
-              throw new ValidationException($"HostPropertyCondition {nameof(Expression)} must be a valid regular expression.");
+                throw new ValidationException($"HostPropertyCondition {nameof(Expression)} must be a valid regular expression.");
             }
         }
     }
--- a/src/WebJobs.Script/Workers/Profiles/Conditions/HostPropertyCondition.cs
+++ b/src/WebJobs.Script/Workers/Profiles/Conditions/HostPropertyCondition.cs
@@ -38,7 +38,7 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Profiles
             Validate();
         }
 
-        public enum HostProperty
+        private enum HostProperty
         {
             Sku,
             Platform,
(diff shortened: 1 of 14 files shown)


src/WebJobs.Script/Workers/Profiles/Conditions/HostPropertyCondition.cs
```diff
@@ -68,27 +67,27 @@ public bool Evaluate()
                 return false;
             }
 
-            _logger.LogDebug($"Evaluating HostPropertyCondition with name: {Name}, value: {value} and expression {Expression}");
+            _logger.LogDebug("Evaluating HostPropertyCondition with name: {name}, value: {value} and expression {expression}", Name, value, Expression);
```
kshyju, line 70: Should HostProperty enum be public ?

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
```diff
@@ -286,31 +298,33 @@ internal bool ShouldAddWorkerConfig(string workerDescriptionLanguage)
...
                 // After specialization only create worker provider for the language set by FUNCTIONS_WORKER_RUNTIME env variable
-                _logger.LogInformation($"{RpcWorkerConstants.FunctionWorkerRuntimeSettingName} set to {_workerRuntime}. Skipping WorkerConfig for language:{workerDescriptionLanguage}");
+                _logger.LogInformation("{RpcWorkerConstants.FunctionWorkerRuntimeSettingName} set to {workerRuntime}. Skipping WorkerConfig for language: {}", RpcWorkerConstants.FunctionWorkerRuntimeSettingName, _workerRuntime, workerDescriptionLanguage);
                 return false;
             }
+
             return true;
         }
 
         internal void ReadLanguageWorkerFile(string workerPath)
```
kshyju: Should this method be private?

kshyju (approved):

src/WebJobs.Script/Workers/Profiles/Conditions/HostPropertyCondition.cs
liliankasem: I don't see why it is, I can make it private!

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
```diff
@@ -165,21 +171,25 @@ internal void AddProvider(string workerDir)
...
                         };
+
                         _workerDescriptionDictionary[workerDescription.Language] = rpcWorkerConfig;
                         ReadLanguageWorkerFile(arguments.WorkerPath);
-                        _logger.LogDebug($"Added WorkerConfig for language: {workerDescription.Language}");
+
+                        _logger.LogDebug("Added WorkerConfig for language: {language}", workerDescription.Language);
                     }
                 }
                 catch (Exception ex)
```
jviau: Can we add a `when (!ex.IsFatal())` here?

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
```diff
@@ -286,31 +298,33 @@ internal bool ShouldAddWorkerConfig(string workerDescriptionLanguage)
...
-                _logger.LogDebug($"EnvironmentVariable {RpcWorkerConstants.FunctionWorkerRuntimeSettingName}: {_workerRuntime}");
+                _logger.LogDebug("EnvironmentVariable {functionWorkerRuntimeSettingName}: {workerRuntime}", RpcWorkerConstants.FunctionWorkerRuntimeSettingName, _workerRuntime);
                 if (_workerRuntime.Equals(workerDescriptionLanguage, StringComparison.OrdinalIgnoreCase))
                 {
                     return true;
                 }
+
                 // After specialization only create worker provider for the language set by FUNCTIONS_WORKER_RUNTIME env variable
-                _logger.LogInformation($"{RpcWorkerConstants.FunctionWorkerRuntimeSettingName} set to {_workerRuntime}. Skipping WorkerConfig for language:{workerDescriptionLanguage}");
+                _logger.LogInformation("{RpcWorkerConstants.FunctionWorkerRuntimeSettingName} set to {workerRuntime}. Skipping WorkerConfig for language: {}", RpcWorkerConstants.FunctionWorkerRuntimeSettingName, _workerRuntime, workerDescriptionLanguage);
```
jviau: Missing name for 3rd param:
```suggestion
                _logger.LogInformation("{RpcWorkerConstants.FunctionWorkerRuntimeSettingName} set to {workerRuntime}. Skipping WorkerConfig for language: {workerDescriptionLanguage}", RpcWorkerConstants.FunctionWorkerRuntimeSettingName, _workerRuntime, workerDescriptionLanguage);
```

src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
```diff
@@ -144,34 +144,36 @@ public async Task SpecializeAsync()
...
-                    || string.Equals(workerRuntime, RpcWorkerConstants.PythonLanguageWorkerName, StringComparison.OrdinalIgnoreCase))
-                {
-                    // Use if readonly and not v2 compatible on ~3 extension
-                    return _applicationHostOptions.CurrentValue.IsFileSystemReadOnly && !_environment.IsV2CompatibileOnV3Extension();
-                }
-                return true;
+            // Special case: node and PowerShell apps must be read-only to use the placeholder mode channel
+            // Also cannot use placeholder worker that is targeting ~3 but has backwards compatibility with V2 enabled
+            // TODO: Remove special casing when resolving https://github.com/Azure/azure-functions-host/issues/4534
+            if (string.Equals(workerRuntime, RpcWorkerConstants.NodeLanguageWorkerName, StringComparison.OrdinalIgnoreCase)
```
jviau: Not sure if you are interested in pattern matching, but you could accomplish this that way. Only drawback is it has to be constants and only ordinal works. Then you get into the issue of normalizing to upper case over lower case (Turkish i issue, but that _really_ isn't applicable here - so lowercase would be safe).

``` CSharp
if (workerRuntime.ToUpperInvariant() is "NODE" or "POWERSHELL" or "PYTHON")
```

jviau (dismissed):
Always love seeing cleanup like this. Just the missing log param value (which I think will cause exceptions if not addressed).

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
liliankasem: good catch!

src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
liliankasem: Doing it this way should still be okay I assume?

```
if (workerRuntime.ToLowerInvariant() is RpcWorkerConstants.NodeLanguageWorkerName
                or RpcWorkerConstants.PowerShellLanguageWorkerName
                or RpcWorkerConstants.PythonLanguageWorkerName)
```

satvu (approved):

src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
fabiocav: Won't the call to `ToUpperInvariant` or `ToLowerInvariant` here for the pattern matching approach cause a string copy/allocation and overall perform worse?

liliankasem: I believe Fabio did a benchmark on this and the existing approach is more performant - will revert this change in the next PR

PR #8732, opened 2022-09-12
[v3.x] moving BlobStorageSecretsRepository to use UploadAsync instead of OpenWriteAsync

brettsam:
v3.x version of #8721


liliankasem (dismissed):

jviau:
I don't know if there is any way to get this race condition 100% in an integration test. I personally would not bother much with that. I would rather write a unit test with a mock and verify we call the client with appropriate options (IfMatch / IfNoneMatch here). At some point we just have to trust the BlobClient is implemented correctly and will behave as expected.

liliankasem (approved):

PR #8754, opened 2022-09-15
HOTFIXv3 - Introduce feature flag to Execution Logger. Opt-in for FunctionExecutionCount 

CooperLink:
Regression in latest host versions due to https://github.com/Azure/azure-functions-host/pull/8315. The number of execution emitted by high load functions overwhelms the metering service and causes Function Apps to fail with memory exceptions.

This PR introduced a feature flag to make FEC reporting optional until we are able to release a metering service fix targetting ANT100. For apps without high continuous load, they can enable FEC reporting.

Related ICMs:
https://portal.microsofticm.com/imp/v3/incidents/details/334835894/home
https://portal.microsofticm.com/imp/v3/incidents/details/333462646/home

### DIFF
--- a/src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
@@ -69,12 +69,15 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
         {
             var logger = _loggerFactory.GetOrCreate(FunctionsExecutionEventsCategory);
             string currentUtcTime = DateTime.UtcNow.ToString();
-            string log = currentUtcTime;
             if (FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagEnableLinuxEPExecutionCount))
             {
                 log = string.Join(",", executionId, siteName, concurrency.ToString(), functionName, invocationId, executionStage, executionTimeSpan.ToString(), success.ToString(), currentUtcTime);
+                WriteEvent(logger, log);
             }
-            WriteEvent(logger, log);
+            else
+            {
+                WriteEvent(logger, currentUtcTime);
+            }
         }
 
         private static void WriteEvent(LinuxAppServiceFileLogger logger, string evt)
--- a/src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
@@ -68,7 +68,14 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
         {
             var logger = _loggerFactory.GetOrCreate(FunctionsExecutionEventsCategory);
             string currentUtcTime = DateTime.UtcNow.ToString();
-            string log = string.Join(",", executionId, siteName, concurrency.ToString(), functionName, invocationId, executionStage, executionTimeSpan.ToString(), success.ToString(), currentUtcTime);
+            if (FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagEnableLinuxEPExecutionCount)) 
+            {
+                string log = string.Join(",", executionId, siteName, concurrency.ToString(), functionName, invocationId, executionStage, executionTimeSpan.ToString(), success.ToString(), currentUtcTime);
+            }
+            else 
+            {
+                string log = currentUtcTime;
+            }
             WriteEvent(logger, log);
         }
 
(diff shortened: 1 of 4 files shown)


src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
```diff
@@ -68,7 +69,11 @@ public LinuxAppServiceEventGenerator(LinuxAppServiceFileLoggerFactory loggerFact
         {
             var logger = _loggerFactory.GetOrCreate(FunctionsExecutionEventsCategory);
             string currentUtcTime = DateTime.UtcNow.ToString();
-            string log = string.Join(",", executionId, siteName, concurrency.ToString(), functionName, invocationId, executionStage, executionTimeSpan.ToString(), success.ToString(), currentUtcTime);
+            string log = currentUtcTime;
+            if (FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagEnableLinuxEPExecutionCount))
+            {
+                log = string.Join(",", executionId, siteName, concurrency.ToString(), functionName, invocationId, executionStage, executionTimeSpan.ToString(), success.ToString(), currentUtcTime);
+            }
             WriteEvent(logger, log);
```
liliankasem: I think this might be cleaner since log is just getting assigned to another string on 72, we really only need the log string if the feature flag is on

```suggestion
            if (FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagEnableLinuxEPExecutionCount))
            {
                string log = string.Join(",", executionId, siteName, concurrency.ToString(), functionName, invocationId, executionStage, executionTimeSpan.ToString(), success.ToString(), currentUtcTime);
                WriteEvent(logger, log);
                return;
            }

            WriteEvent(logger, currentUtcTime);
```

liliankasem: Or else-statement instead of return - either way

test/WebJobs.Script.Tests/Diagnostics/LinuxAppServiceEventGeneratorTests.cs
```diff
@@ -175,31 +175,31 @@ public void ParseAzureMonitoringEvents(LogLevel level, string resourceId, string
...
-                p => Assert.Equal(executionStage, p),
-                p => Assert.Equal(executionTimeSpan.ToString(), p),
-                p => Assert.True(Convert.ToBoolean(p)),
-                p => Assert.True(DateTime.TryParse(p, out DateTime dt)));
-        }
+        // [Theory]
+        // [MemberData(nameof(LinuxEventGeneratorTestData.GetFunctionExecutionEvents), MemberType = typeof(LinuxEventGeneratorTestData))]
+        // public void ParseFunctionExecutionEvents(string executionId, string siteName, int concurrency, string functionName, string invocationId,
+        //     string executionStage, long executionTimeSpan, bool success)
+        // {
```
liliankasem: You can test by setting the feature flag and seeing what the outcome is (with and without the flag)

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
```diff
@@ -68,8 +69,15 @@ public LinuxAppServiceEventGenerator(LinuxAppServiceFileLoggerFactory loggerFact
         {
             var logger = _loggerFactory.GetOrCreate(FunctionsExecutionEventsCategory);
             string currentUtcTime = DateTime.UtcNow.ToString();
-            string log = string.Join(",", executionId, siteName, concurrency.ToString(), functionName, invocationId, executionStage, executionTimeSpan.ToString(), success.ToString(), currentUtcTime);
-            WriteEvent(logger, log);
+            if (FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagEnableLinuxEPExecutionCount))
+            {
+                log = string.Join(",", executionId, siteName, concurrency.ToString(), functionName, invocationId, executionStage, executionTimeSpan.ToString(), success.ToString(), currentUtcTime);
```
liliankasem:
```suggestion
                string log = string.Join(",", executionId, siteName, concurrency.ToString(), functionName, invocationId, executionStage, executionTimeSpan.ToString(), success.ToString(), currentUtcTime);
```

test/WebJobs.Script.Tests/Diagnostics/LinuxAppServiceEventGeneratorTests.cs
liliankasem: Example of setting the feature flag in a test

https://github.com/Azure/azure-functions-host/blob/9a776f22ed6b85fc026a2799f3b0d4b8dd87a9ad/test/WebJobs.Script.Tests/UtilityTests.cs#L866

fabiocav: Indeed. The test should be updated, but similar to the comment in the other PR, we should not be checking commented out code in.

fabiocav (changes requested):

test/WebJobs.Script.Tests/Diagnostics/LinuxAppServiceEventGeneratorTests.cs
CooperLink: Updated the tests in line with other PR

fabiocav (approved):

PR #8763, opened 2022-09-16
Register the profile manager at the WebHost scope only

### LINKED ISSUE #8516: Worker Profiles - Active profile not being tracked

liliankasem:
resolves #8516

Change to DI WorkerProfileManager so there is only one instance

### DIFF
--- a/src/WebJobs.Script/ScriptHostBuilderExtensions.cs
+++ b/src/WebJobs.Script/ScriptHostBuilderExtensions.cs
@@ -26,6 +26,7 @@ using Microsoft.Azure.WebJobs.Script.ManagedDependencies;
 using Microsoft.Azure.WebJobs.Script.Scale;
 using Microsoft.Azure.WebJobs.Script.Workers;
 using Microsoft.Azure.WebJobs.Script.Workers.Http;
+using Microsoft.Azure.WebJobs.Script.Workers.Profiles;
 using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
--- a/src/WebJobs.Script/ScriptHostBuilderExtensions.cs
+++ b/src/WebJobs.Script/ScriptHostBuilderExtensions.cs
@@ -326,6 +326,7 @@ namespace Microsoft.Azure.WebJobs.Script
             services.AddSingleton<HostIdValidator>();
             services.AddSingleton<IHostIdProvider, ScriptHostIdProvider>();
             services.TryAddSingleton<IScriptEventManager, ScriptEventManager>();
+            services.AddSingleton<IWorkerProfileManager, WorkerProfileManager>();
 
             // Add Language Worker Service
             // Need to maintain the order: Add RpcInitializationService before core script host services
--- a/src/WebJobs.Script/Workers/Profiles/WorkerProfileManager.cs
+++ b/src/WebJobs.Script/Workers/Profiles/WorkerProfileManager.cs
@@ -14,8 +14,6 @@ namespace Microsoft.Azure.WebJobs.Script.Workers
     /// </summary>
     internal class WorkerProfileManager : IWorkerProfileManager
     {
-        private const string LogCategory = "Host.WorkerProfiles";
-
         private readonly ILogger _logger;
         private readonly IEnvironment _environment;
         private readonly IEnumerable<IWorkerProfileConditionProvider> _conditionProviders;
@@ -23,16 +21,11 @@ namespace Microsoft.Azure.WebJobs.Script.Workers
         private string _activeProfile = string.Empty;
         private Dictionary<string, List<WorkerDescriptionProfile>> _profiles;
 
-        public WorkerProfileManager(ILoggerFactory loggerFactory, IEnvironment environment)
+        public WorkerProfileManager(ILogger<WorkerProfileManager> logger, IEnvironment environment)
         {
-            if (loggerFactory is null)
-            {
-                throw new ArgumentNullException(nameof(loggerFactory));
-            }
-
             _environment = environment ?? throw new ArgumentNullException(nameof(environment));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
 
-            _logger = loggerFactory.CreateLogger(LogCategory);
             _profiles = new Dictionary<string, List<WorkerDescriptionProfile>>();
             _conditionProviders = new List<IWorkerProfileConditionProvider>
             {
--- a/src/WebJobs.Script/Workers/Profiles/WorkerProfileManager.cs
+++ b/src/WebJobs.Script/Workers/Profiles/WorkerProfileManager.cs
@@ -1,5 +1,5 @@
 // Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+// Licensed under the MIT License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
 using System.Collections.Generic;
@@ -72,9 +72,11 @@ namespace Microsoft.Azure.WebJobs.Script.Workers
                 _logger?.LogInformation($"Worker initialized with profile - {profile.Name}, Profile ID {profile.ProfileId} from worker config.");
                 _activeProfile = profile.ProfileId;
                 workerDescription = profile.ApplyProfile(defaultWorkerDescription);
-                return;
             }
-            workerDescription = defaultWorkerDescription;
+            else
+            {
+                workerDescription = defaultWorkerDescription;
+            }
         }
 
         /// <inheritdoc />
--- a/src/WebJobs.Script/Workers/Profiles/WorkerProfileManager.cs
+++ b/src/WebJobs.Script/Workers/Profiles/WorkerProfileManager.cs
@@ -14,17 +14,30 @@ namespace Microsoft.Azure.WebJobs.Script.Workers
     /// </summary>
     internal class WorkerProfileManager : IWorkerProfileManager
     {
-        private readonly ILogger _logger;
-        private readonly IEnumerable<IWorkerProfileConditionProvider> _conditionProviders;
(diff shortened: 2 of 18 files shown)


src/WebJobs.Script/Workers/Profiles/WorkerProfileManager.cs
```diff
@@ -14,17 +14,30 @@ namespace Microsoft.Azure.WebJobs.Script.Workers
     /// </summary>
     internal class WorkerProfileManager : IWorkerProfileManager
     {
+        private const string LogCategory = "Host.WorkerProfiles";
```
brettsam: We're trying not to use "string" categories anymore and instead use `ILogger<WorkerProfileManager>` in the constructor, which will use that type name as the category...

test/WebJobs.Script.Tests.Shared/TestHelpers.cs
```diff
@@ -24,8 +24,8 @@
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Options;
-using Microsoft.WebJobs.Script.Tests;
 using Moq;
+using Newtonsoft.Json.Linq;
```
brettsam: what happened here? was this using statement really missing?

brettsam (commented):
just a couple minor comments...

test/WebJobs.Script.Tests.Shared/TestHelpers.cs
liliankasem: Ooh nice catch! I made a change that was using json, but removed those changes and missed this import

src/WebJobs.Script/Workers/Profiles/WorkerProfileManager.cs
```diff
@@ -14,17 +14,30 @@ namespace Microsoft.Azure.WebJobs.Script.Workers
...
+        private string _activeProfile = string.Empty;
         private Dictionary<string, List<WorkerDescriptionProfile>> _profiles;
-        private string _activeProfile;
 
-        public WorkerProfileManager(ILogger logger, IEnumerable<IWorkerProfileConditionProvider> conditionProviders)
+        public WorkerProfileManager(ILoggerFactory loggerFactory, IEnvironment environment)
         {
-            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
-            _conditionProviders = conditionProviders ?? throw new ArgumentNullException(nameof(conditionProviders));
+            if (loggerFactory is null)
```
jviau: Are we on a new enough .NET version to use `ArgumentNullException.ThrowIfNull(loggerFactory)`?

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
```diff
@@ -32,22 +32,19 @@ internal class RpcWorkerConfigFactory
...
-            {
-                new WorkerProfileConditionProvider(_logger, _environment)
-            };
+            _metricsLogger = metricsLogger ?? throw new ArgumentNullException(nameof(metricsLogger));
+            _profileManager = workerProfileManager ?? throw new ArgumentNullException(nameof(workerProfileManager));
 
-            _profileManager = new WorkerProfileManager(_logger, conditionProviders);
+            _workerRuntime = _environment.GetEnvironmentVariable(RpcWorkerConstants.FunctionWorkerRuntimeSettingName);
 
+            string assemblyLocalPath = Path.GetDirectoryName(new Uri(typeof(RpcWorkerConfigFactory).Assembly.CodeBase).LocalPath);
```
jviau: Is this used for anything?

src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
```diff
@@ -159,9 +157,9 @@ private bool UsePlaceholderChannel(string workerRuntime)
             // Special case: node and PowerShell apps must be read-only to use the placeholder mode channel
             // Also cannot use placeholder worker that is targeting ~3 but has backwards compatibility with V2 enabled
             // TODO: Remove special casing when resolving https://github.com/Azure/azure-functions-host/issues/4534
-            if (workerRuntime.ToLowerInvariant() is RpcWorkerConstants.NodeLanguageWorkerName
-                or RpcWorkerConstants.PowerShellLanguageWorkerName
-                or RpcWorkerConstants.PythonLanguageWorkerName)
+            if (string.Equals(workerRuntime, RpcWorkerConstants.NodeLanguageWorkerName, StringComparison.OrdinalIgnoreCase)
```
jviau, line 160: Ah yeah forgot about the perf. This might be solved with C# 11 where you can make the invariant via a span and then pattern match against that (no heap allocation, `.ToSpan().ToLowerInvariant()`).

jviau (approved):

src/WebJobs.Script/Workers/Profiles/WorkerProfileManager.cs
liliankasem: I get "ThrowIfNull" is not available on `ArgumentNullException`. However I am removing the factory as per Bretts comment

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
liliankasem: No, I actually don't even remember adding this line at all ☠️ thanks for catching

src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
liliankasem: I think I'm going to leave this out for now - if we'd like to make changes around this we can create a PR for it with benchmark stats just to be sure!

jviau: Oh yeah of course. I don't think we are on C# 11 anyways?

liliankasem: Ah yes, I don't believe we are yet :D

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -326,6 +327,7 @@ public static void AddCommonServices(IServiceCollection services)
             services.AddSingleton<HostIdValidator>();
             services.AddSingleton<IHostIdProvider, ScriptHostIdProvider>();
             services.TryAddSingleton<IScriptEventManager, ScriptEventManager>();
+            services.AddSingleton<IWorkerProfileManager, WorkerProfileManager>();
```
shreyas-gopalakrishna, line 330: Just as a precaution, checking if this service registration would impact cold start of languages that do not use profiles. 
Bringing this to attention as previously we had decided to avoid service registration. 
@safihamid could you please let us know your opinion on this #Resolved

safihamid: Oh yes, this will have a cold start impact. we reviewed and measured a similar PR related to this and Fabio's suggestion fixed it. @shreyas-gopalakrishna can you point to that PR please?

shreyas-gopalakrishna: Initial PR - https://github.com/Azure/azure-functions-host/pull/8365
PR where we removed this - https://github.com/Azure/azure-functions-host/pull/8502

liliankasem: @fabiocav and @brettsam  discussed this and we decided this was the best way to unblock the java team for now. In future this will need a full refactor but we don't have the capacity for that right now.

liliankasem: I did have a quick sync with @fabiocav about this and he mentioned that this should improve cold start as we are now only registering once at the WebHost scope whereas before it was getting registered multiple times at the WebHost and JobHost scope

safihamid: yes that's right. I see they are being added in AddCommonServices. Didn't notice it initially.

brettsam (approved):

test/WebJobs.Script.Tests.Shared/TestHelpers.cs
```diff
@@ -25,6 +25,7 @@
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Options;
 using Moq;
+using Newtonsoft.Json.Linq;
```
fabiocav: What prompted this change?

fabiocav (approved):
Small question. No need to block

test/WebJobs.Script.Tests.Shared/TestHelpers.cs
liliankasem: I removed this but must have accidentally come back during the rebase - thanks for catching!

PR #8766, opened 2022-09-17
Fix incorrect license header

liliankasem:
License should be MIT


jviau:
We also have a `LICENSE_APACHE.txt` in the repo root. Is that to remain?

liliankasem:
> We also have a `LICENSE_APACHE.txt` in the repo root. Is that to remain?

Ah good point, @fabiocav can we remove that file do you think?

fabiocav:
> > We also have a `LICENSE_APACHE.txt` in the repo root. Is that to remain?
> 
> Ah good point, @fabiocav can we remove that file do you think?

That should be removed.

fabiocav:
Responded too quickly. We actually need to maintain that file. 

Some of the work done by @alrod here used code from other repositories that would need to maintain their license (https://github.com/Azure/azure-functions-host/commit/00e62f4c3d6dfb1247a261b6e9ceeb42201883c0). The header in that code does require the presence of the license file (by default, at the root) for reference. 

I do believe the changes went too far, though, as not all of the types defined in that PR came from that implementation (tests certainly didn't). Those files would also need to remain untouched in your PR (their license would need to remain the same), and perhaps we can add a comment so others are aware. 

@alrod might be able to assist in identifying what files came from the external repo.

liliankasem:
> Responded too quickly. We actually need to maintain that file.
> 
> Some of the work done by @alrod here used code from other repositories that would need to maintain their license ([00e62f4](https://github.com/Azure/azure-functions-host/commit/00e62f4c3d6dfb1247a261b6e9ceeb42201883c0)). The header in that code does require the presence of the license file (by default, at the root) for reference.
> 
> I do believe the changes went too far, though, as not all of the types defined in that PR came from that implementation (tests certainly didn't). Those files would also need to remain untouched in your PR (their license would need to remain the same), and perhaps we can add a comment so others are aware.
> 
> @alrod might be able to assist in identifying what files came from the external repo.

Keeping the `LICENSE_APACHE.txt`, but looking through Alexey's PR there, I didn't change anything that was added during that PR under apache so I think we're good to go with merging this

satvu (approved):
lgtm - compared these changes to Alexey's PR as well and there's no crossover.

alrod (approved):

PR #8772, opened 2022-09-21
[pack] [EMG] fixing infinite loop issue with typeforwarders

brettsam:
It is rare, but some customers can hit an infinite loop with circular TypeForwarders. The issue in Cecil is here: https://github.com/jbevain/cecil/issues/706

This fix short-circuits our search for extensions if we hit a TPA as there's no need to go further. This breaks that circular logic as we don't keep searching for base types and assemblies once we make it this far.

### DIFF
--- a/tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator.Console/ExtensionsMetadataGenerator.cs
+++ b/tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator.Console/ExtensionsMetadataGenerator.cs
@@ -186,7 +186,7 @@ namespace ExtensionsMetadataGenerator
 
         public class FunctionsAssemblyResolver : DefaultAssemblyResolver
         {
-            private static HashSet<string> _trustedPlatformAssemblies = GetTrustedPlatformAssemblies();
+            private static readonly HashSet<string> _trustedPlatformAssemblies = GetTrustedPlatformAssemblies();
 
             private static HashSet<string> GetTrustedPlatformAssemblies()
             {
--- a/tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator.Console/ExtensionsMetadataGenerator.cs
+++ b/tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator.Console/ExtensionsMetadataGenerator.cs
@@ -190,13 +190,16 @@ namespace ExtensionsMetadataGenerator
 
             private static HashSet<string> GetTrustedPlatformAssemblies()
             {
-                var tpas = new HashSet<string>();
                 var data = AppContext.GetData("TRUSTED_PLATFORM_ASSEMBLIES");
-                foreach (var tpa in data.ToString().Split(";"))
+                var tpaArr = data.ToString().Split(Path.PathSeparator);
+                var tpaHash = new HashSet<string>(tpaArr.Length, StringComparer.OrdinalIgnoreCase);
+
+                foreach (var tpa in tpaArr)
                 {
-                    tpas.Add(tpa);
+                    tpaHash.Add(tpa);
                 }
-                return tpas;
+
+                return tpaHash;
             }
 
             public override AssemblyDefinition Resolve(AssemblyNameReference name)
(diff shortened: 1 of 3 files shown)


tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator.Console/ExtensionsMetadataGenerator.cs
```diff
@@ -183,5 +183,35 @@ public static string GetReflectionFullName(TypeReference typeRef)
...
+        {
+            private static HashSet<string> _trustedPlatformAssemblies = GetTrustedPlatformAssemblies();
+
+            private static HashSet<string> GetTrustedPlatformAssemblies()
+            {
+                var tpas = new HashSet<string>();
+                var data = AppContext.GetData("TRUSTED_PLATFORM_ASSEMBLIES");
+                foreach (var tpa in data.ToString().Split(";"))
+                {
+                    tpas.Add(tpa);
```
kshyju: Should we switch to the HashSet constructor which takes capacity value? That will prevent the auto-grow behavior(copy & allocation). I ran a mini benchmark (.NET 6 though): https://gist.github.com/kshyju/026fbce09291bbe2fe55d12b45c6b00e

kshyju: Oh, this project is .NET6.

tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator.Console/ExtensionsMetadataGenerator.cs
```diff
@@ -183,5 +183,35 @@ public static string GetReflectionFullName(TypeReference typeRef)
...
+                return tpas;
+            }
+
+            public override AssemblyDefinition Resolve(AssemblyNameReference name)
+            {
+                var assemblyDef = base.Resolve(name);
+
+                // As soon as we get to a TPA we can stop. This helps prevent the odd circular reference
+                // with type forwarders as well.
+                if (_trustedPlatformAssemblies.Contains(assemblyDef.MainModule.FileName))
```
kshyju, line 211: Should we worry about supporting case insensitive Contains check?

```
var tpas = new HashSet<string>(StringComparer.OrdinalIgnoreCase);

Comment (brettsam) [reply tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator.Console/ExtensionsMetadataGenerator.cs:211]: yeah good call...

File: tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator.Console/ExtensionsMetadataGenerator.cs [hunk 3]
```diff
@@ -111,7 +111,7 @@ public static bool IsWebJobsStartupAttributeType(TypeReference attributeType, Co
 
         public static IEnumerable<ExtensionReference> GenerateExtensionReferences(string fileName, ConsoleLogger logger)
         {
-            BaseAssemblyResolver resolver = new DefaultAssemblyResolver();
+            var resolver = new FunctionsAssemblyResolver();
```
jviau, line 114: Do we prefer the var style or `FunctionsAssemblyResolver resolver = new();`?

tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator.Console/ExtensionsMetadataGenerator.cs
```diff
@@ -183,5 +183,38 @@ public static string GetReflectionFullName(TypeReference typeRef)
         {
             return typeRef.FullName.Replace("/", "+");
         }
+
+        public class FunctionsAssemblyResolver : DefaultAssemblyResolver
+        {
+            private static HashSet<string> _trustedPlatformAssemblies = GetTrustedPlatformAssemblies();
```
jviau: nit: can be marked `readonly`.

jviau (approved):

fabiocav (approved):

tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator.Console/ExtensionsMetadataGenerator.cs
brettsam: I prefer the newer style... but we didn't really want to mix styles in these existing projects, so we are sticking with the 'var' style unless it's a new project. At least I think that's our philosophy :-)

satvu (approved):

liliankasem (approved):

PR #8776, opened 2022-09-22
Ensure fatal exceptions do not cause key regeneration

### LINKED ISSUE #8775: Fatal exceptions cause secret regeneration
The host incorrectly regenerates secrets when the CryptographicException thrown is caused by a fatal exception. This causes problems with secrets being regenerated when they shouldn't and leaving the application in a broken state.

fabiocav:
resolves #8775

This is a tactical initial fix. Additional enhancements for this flow will come in follow up PRs

### DIFF
--- a/src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
+++ b/src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
@@ -106,7 +106,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
                             // so we read the secrets running them through the appropriate readers
                             hostSecrets = ReadHostSecrets(hostSecrets);
                         }
-                        catch (CryptographicException ex)
+                        catch (CryptographicException ex) when (!ex.InnerException.IsFatal())
                         {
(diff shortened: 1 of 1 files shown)


brettsam (approved):

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
```diff
@@ -106,7 +106,7 @@ public async virtual Task<HostSecretsInfo> GetHostSecretsAsync()
                             // so we read the secrets running them through the appropriate readers
                             hostSecrets = ReadHostSecrets(hostSecrets);
                         }
-                        catch (CryptographicException ex)
+                        catch (CryptographicException ex) when (!ex.InnerException.IsFatal())
```
mathewc, line 109: Now, if we have a failure here, hostSecrets will be null, which will lead to a null ref below on `hostSecrets.HasStaleKeys`. It seems we'll need special handling in these methods for the case where we can't read keys at all, and tests to ensure we handle the lack of keys (nulls) properly.

mathewc (approved):

PR #8778, opened 2022-09-22
Ensure fatal exceptions do not cause key regeneration

fabiocav:


src/WebJobs.Script.WebHost/Security/SecretManager.cs
```diff
@@ -95,7 +95,7 @@ public async virtual Task<HostSecretsInfo> GetHostSecretsAsync()
                         // so we read the secrets running them through the appropriate readers
                         hostSecrets = ReadHostSecrets(hostSecrets);
                     }
-                    catch (CryptographicException)
+                    catch (CryptographicException ex) when (!ex.InnerException.IsFatal())
```
jviau, line 98: No null ref concern with the `IsFatal()` call?

jviau (approved):

src/WebJobs.Script.WebHost/Security/SecretManager.cs
fabiocav: The `IsFatal` handles `null` parameters and iterates over the inner exceptions, so it's safe here.

brettsam (approved):

liliankasem (approved):

PR #8780, opened 2022-09-23
Target base scale support

### LINKED ISSUE #2909: Target base scale in runtime driven scale
Intension
In order to GA target base scale we need to add the feature into runtime driven scale, the feature was already implemented in `ScaleController`. This work intersects with SC redesign but should be released before as SC redesign is bigger item and it will take longer. SC redesign will use the change in the PR for target base scale(TBS).

Changes/Design
New interfaces:
`ITargetScaler` - Interface defining an Azure Functions scaler that makes scale decisions based on current event source metrics and function concurrency. Needs to be implemented in an extension.
`ITargetScalerManager` - Manager for registering and accessing `ITargetScaler` instances for `JobHost` instance. Implemented in WebJobs SDK.
`ITargetScalerProvider` - Provider interface for returning an `ITargetScaler` instance. Needs to be implemented in an extension.

Flow:
* An extension implements `ITargetScaler`
* `HostListenerFactory` in webjobs registers `ITargetScaler` using `ITargetScalerManager.Register`
* SC calls FunctionHost's api: `admin/host/scale/status`
* FunctionHost gets all the registered scalers `ITargetScalers` by `ITargetScalerManager.GetTargetScalers` 
* FunctionHost figures out `ITargetScalers` and `IScaleMonitors` to process. If an extension implements both and TBS is enabled only `ITargetScalers` is used.
* FunctionHost calls `ITargetScaler.GetScaleResultAsync` and `IScaleMonitor.GetScaleStatus` and return response to SC

The flow is similar to regular(incremental) runtime scale.

alrod:

resolves https://github.com/Azure/azure-webjobs-sdk/issues/2909

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script/Config/FunctionsHostingConfiguration.cs
+++ b/src/WebJobs.Script/Config/FunctionsHostingConfiguration.cs
@@ -77,10 +77,6 @@ namespace Microsoft.Azure.WebJobs.Script.Config
                             _configs = configs;
                             _logger.LogDebug($"Updaiting FunctionsHostingConfigurations '{settings}'");
                         }
-                        else
-                        {
-                            throw new InvalidOperationException("FunctionsHostingConfigurations does not exists");
-                        }
                     }
                 }
             }
--- a/src/WebJobs.Script/Config/FunctionsHostingConfiguration.cs
+++ b/src/WebJobs.Script/Config/FunctionsHostingConfiguration.cs
@@ -19,6 +19,7 @@ namespace Microsoft.Azure.WebJobs.Script.Config
 
         private DateTime _configTTL;
         private Dictionary<string, string> _configs;
+        private bool isConfigurationFileExists = true;
 
         public FunctionsHostingConfiguration(IEnvironment environment, ILoggerFactory loggerFactory)
         {
@@ -61,6 +62,12 @@ namespace Microsoft.Azure.WebJobs.Script.Config
 
         public string GetValue(string key, string defaultValue = null)
         {
+            // the config file will be parsed after functionn host restart
+            if (!isConfigurationFileExists)
+            {
+                return defaultValue;
+            }
+
             var configs = _configs;
             if (configs == null || DateTime.UtcNow > _configTTL)
             {
@@ -76,6 +83,12 @@ namespace Microsoft.Azure.WebJobs.Script.Config
                             configs = Parse(settings);
                             _configs = configs;
                             _logger.LogDebug($"Updaiting FunctionsHostingConfigurations '{settings}'");
+                        }
+                        else
+                        {
+                            // if config file does not exists we want to stop trying getting the value and return default value
+                            isConfigurationFileExists = false;
+                            _logger.LogDebug($"FunctionsHostingConfigurations file does not exist");
                         }
                     }
                 }
--- a/src/WebJobs.Script/Config/FunctionsHostingConfiguration.cs
+++ b/src/WebJobs.Script/Config/FunctionsHostingConfiguration.cs
@@ -19,7 +19,7 @@ namespace Microsoft.Azure.WebJobs.Script.Config
 
         private DateTime _configTTL;
         private Dictionary<string, string> _configs;
-        private bool isConfigurationFileExists = true;
+        private bool _configurationFileExists = true;
 
         public FunctionsHostingConfiguration(IEnvironment environment, ILoggerFactory loggerFactory)
         {
@@ -63,7 +63,7 @@ namespace Microsoft.Azure.WebJobs.Script.Config
         public string GetValue(string key, string defaultValue = null)
         {
             // the config file will be parsed after functionn host restart
-            if (!isConfigurationFileExists)
+            if (!_configurationFileExists)
             {
                 return defaultValue;
             }
@@ -87,7 +87,7 @@ namespace Microsoft.Azure.WebJobs.Script.Config
                         else
                         {
                             // if config file does not exists we want to stop trying getting the value and return default value
-                            isConfigurationFileExists = false;
+                            _configurationFileExists = false;
                             _logger.LogDebug($"FunctionsHostingConfigurations file does not exist");
                         }
                     }
--- a/src/WebJobs.Script/Config/IFunctionsHostingConfiguration.cs
+++ b/src/WebJobs.Script/Config/IFunctionsHostingConfiguration.cs
@@ -3,8 +3,10 @@
 
 namespace Microsoft.Azure.WebJobs.Script.Config
 {
-    internal interface IFunctionsHostingConfiguration
+    public interface IFunctionsHostingConfiguration
     {
         bool FunctionsWorkerDynamicConcurrencyEnabled { get; }
+
+        string GetValue(string key, string defaultValue = null);
     }
 }
--- a/src/WebJobs.Script/Environment/EnvironmentExtensions.cs
+++ b/src/WebJobs.Script/Environment/EnvironmentExtensions.cs
@@ -601,5 +601,10 @@ namespace Microsoft.Azure.WebJobs.Script
         {
             return environment.GetEnvironmentVariableOrDefault(FunctionsExtensionVersion, string.Empty);
         }
+
+        public static bool IsTargetBasedScalingEnabled(this IEnvironment environment)
+        {
+            return string.Equals(environment.GetEnvironmentVariable(TargetBaseScalingEnabled), "1");
+        }
     }
 }
--- a/src/WebJobs.Script/Scale/FunctionsScaleManager.cs
+++ b/src/WebJobs.Script/Scale/FunctionsScaleManager.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Azure.AppService.Proxy.Client;
 using Microsoft.Azure.WebJobs.Host.Scale;
 using Microsoft.Azure.WebJobs.Script.Config;
 using Microsoft.Extensions.Logging;
@@ -56,22 +57,22 @@ namespace Microsoft.Azure.WebJobs.Script.Scale
         /// <returns>The scale vote.</returns>
         public virtual async Task<ScaleStatusResult> GetScaleStatusAsync(ScaleStatusContext context)
         {
-            var scaleMonitors = _monitorManager.GetMonitors();
-            var targetScalers = _targetScalerManager.GetTargetScalers();
+            GetScaleStatusProviders(_monitorManager, _targetScalerManager, out IEnumerable<IScaleMonitor> scaleMonitors, out IEnumerable<ITargetScaler> targetScalers);
 
-            Utility.GetScaleInstancesToProcess(_environment, _functionsHostingConfiguration, scaleMonitors, targetScalers,
+            Utility.GetScalersToSample(_environment, _functionsHostingConfiguration, scaleMonitors, targetScalers,
                 out List<IScaleMonitor> scaleMonitorsToProcess, out List<ITargetScaler> targetScalersToProcess);
 
-            var targetScalerVotes = await GetTargetScalersResult(context, targetScalersToProcess);
+            var scaleMonitorVotes = await GetScaleMonitorsResultAsync(context, scaleMonitorsToProcess);
+            var targetScalerVotes = await GetTargetScalersResultAsync(context, targetScalersToProcess);
 
             return new ScaleStatusResult
             {
-                Vote = await GetScaleMonitorsResult(context, scaleMonitorsToProcess, targetScalerVotes.Select(x => x.Vote)),
-                TargetWorkerCount = targetScalerVotes.Any() ? targetScalerVotes.Select(x => x.TargetWorkerCount).Max() : null
+                Vote = GetAggregateScaleVote(scaleMonitorVotes.Union(targetScalerVotes.Select(x => x.Vote)), context, _logger),
+                TargetWorkerCount = targetScalerVotes.Any() ? targetScalerVotes.Max(x => x.TargetWorkerCount) : null
             };
         }
 
-        private async Task<ScaleVote> GetScaleMonitorsResult(ScaleStatusContext context, IEnumerable<IScaleMonitor> scaleMonitorsToProcess, IEnumerable<ScaleVote> targetScaleVotes)
+        private async Task<IEnumerable<ScaleVote>> GetScaleMonitorsResultAsync(ScaleStatusContext context, IEnumerable<IScaleMonitor> scaleMonitorsToProcess)
         {
             List<ScaleVote> votes = new List<ScaleVote>();
             if (scaleMonitorsToProcess.Any())
@@ -116,11 +117,10 @@ namespace Microsoft.Azure.WebJobs.Script.Scale
                 // this can happen if the host is offline
             }
 
-            ScaleVote vote = GetAggregateScaleVote(votes.Union(targetScaleVotes), context, _logger);
-            return vote;
+            return votes;
         }
 
-        private async Task<IEnumerable<TargetScalerVote>> GetTargetScalersResult(ScaleStatusContext context, IEnumerable<ITargetScaler> targetScalersToProcess)
+        private async Task<IEnumerable<TargetScalerVote>> GetTargetScalersResultAsync(ScaleStatusContext context, IEnumerable<ITargetScaler> targetScalersToProcess)
         {
             List<TargetScalerVote> targetScaleVotes = new List<TargetScalerVote>();
 
@@ -150,8 +150,10 @@ namespace Microsoft.Azure.WebJobs.Script.Scale
                                 _logger.LogDebug($"Snapshot dynamic concurrency for target scaler '{targetScaler.TargetScalerDescriptor.FunctionId}' is '{functionSnapshot.Concurrency}'");
                             }
                         }
+
                         TargetScalerResult result = await targetScaler.GetScaleResultAsync(targetScaleStatusContext);
                         _logger.LogDebug($"Target worker count for '{targetScaler.TargetScalerDescriptor.FunctionId}' is '{result.TargetWorkerCount}'");
+
                         ScaleVote vote = ScaleVote.None;
                         if (context.WorkerCount > result.TargetWorkerCount)
                         {
@@ -161,6 +163,7 @@ namespace Microsoft.Azure.WebJobs.Script.Scale
                         {
                             vote = ScaleVote.ScaleOut;
                         }
+
                         targetScaleVotes.Add(new TargetScalerVote
                         {
                             TargetWorkerCount = result.TargetWorkerCount,
@@ -174,8 +177,17 @@ namespace Microsoft.Azure.WebJobs.Script.Scale
                     }
                 }
             }
-            //int? targetWorkerCount = targetScaleVotes.Any() ? targetScaleVotes.Max() : null;
             return targetScaleVotes;
+        }
+
+        internal static void GetScaleStatusProviders(
+            IScaleMonitorManager scaleMonitorManager,
+            ITargetScalerManager targetScalerManager,
+            out IEnumerable<IScaleMonitor> scaleMonitors,
+            out IEnumerable<ITargetScaler> targetScalers)
+        {
+            scaleMonitors = scaleMonitorManager.GetMonitors();
+            targetScalers = targetScalerManager.GetTargetScalers();
         }
 
         internal static ScaleVote GetAggregateScaleVote(IEnumerable<ScaleVote> votes, ScaleStatusContext context, ILogger logger)
--- a/src/WebJobs.Script/Scale/FunctionsScaleManager.cs
+++ b/src/WebJobs.Script/Scale/FunctionsScaleManager.cs
@@ -57,9 +57,10 @@ namespace Microsoft.Azure.WebJobs.Script.Scale
         /// <returns>The scale vote.</returns>
         public virtual async Task<ScaleStatusResult> GetScaleStatusAsync(ScaleStatusContext context)
         {
-            GetScaleStatusProviders(_monitorManager, _targetScalerManager, out IEnumerable<IScaleMonitor> scaleMonitors, out IEnumerable<ITargetScaler> targetScalers);
+            var scaleMonitors = _monitorManager.GetMonitors();
+            var targetScalers = _targetScalerManager.GetTargetScalers();
 
-            Utility.GetScalersToSample(_environment, _functionsHostingConfiguration, scaleMonitors, targetScalers,
+            GetScalersToSample(_environment, _functionsHostingConfiguration, scaleMonitors, targetScalers,
                 out List<IScaleMonitor> scaleMonitorsToProcess, out List<ITargetScaler> targetScalersToProcess);
 
             var scaleMonitorVotes = await GetScaleMonitorsResultAsync(context, scaleMonitorsToProcess);
@@ -180,16 +181,6 @@ namespace Microsoft.Azure.WebJobs.Script.Scale
             return targetScaleVotes;
         }
 
-        internal static void GetScaleStatusProviders(
-            IScaleMonitorManager scaleMonitorManager,
-            ITargetScalerManager targetScalerManager,
-            out IEnumerable<IScaleMonitor> scaleMonitors,
-            out IEnumerable<ITargetScaler> targetScalers)
-        {
-            scaleMonitors = scaleMonitorManager.GetMonitors();
-            targetScalers = targetScalerManager.GetTargetScalers();
-        }
-
         internal static ScaleVote GetAggregateScaleVote(IEnumerable<ScaleVote> votes, ScaleStatusContext context, ILogger logger)
         {
             ScaleVote vote = ScaleVote.None;
@@ -217,6 +208,58 @@ namespace Microsoft.Azure.WebJobs.Script.Scale
             }
 
             return vote;
+        }
+
+        /// <summary>
+        /// Returns scale monitors and target scalers we want to use based on the configuration.
+        /// Scaler monitor will be ignored if a target scaler is defined in the same extensions assembly and TBS is enabled.
+        /// </summary>
+        /// <param name="environment">Environment variables.</param>
+        /// <param name="hostingConfiguration">Hosting configuration.This is used to enable TDS on stamp level for specific triggers.</param>
+        /// <param name="scaleMonitors">Registered scale monitors.</param>
+        /// <param name="targetScalers">Registered target scalers.</param>
+        /// <param name="scaleMonitorsToSample">Scale monitor to process.</param>
+        /// <param name="targetScalersToSample">Target scaler to process.</param>
+        internal static void GetScalersToSample(
+            IEnvironment environment,
+            IFunctionsHostingConfiguration hostingConfiguration,
+            IEnumerable<IScaleMonitor> scaleMonitors,
+            IEnumerable<ITargetScaler> targetScalers,
+            out List<IScaleMonitor> scaleMonitorsToSample,
+            out List<ITargetScaler> targetScalersToSample)
+        {
+            scaleMonitorsToSample = new List<IScaleMonitor>();
+            targetScalersToSample = new List<ITargetScaler>();
+
+            // Check if TBS enabled on app level
+            if (environment.IsTargetBasedScalingEnabled())
+            {
+                HashSet<string> targetScalerAssemblies = new HashSet<string>();
+                foreach (var scaler in targetScalers)
+                {
+                    string assemblyName = scaler.GetType().Assembly.GetName().Name;
+                    string flag = hostingConfiguration.GetValue(assemblyName, null);
+                    if (flag == "1")
+                    {
+                        targetScalersToSample.Add(scaler);
+                        targetScalerAssemblies.Add(assemblyName);
+                    }
+                }
+
+                foreach (var monitor in scaleMonitors)
+                {
+                    string monitorAssemblyName = monitor.GetType().Assembly.GetName().Name;
+                    // Check if there are scale monitor and target scaler defined in the same assembly
+                    if (!targetScalerAssemblies.Contains(monitorAssemblyName))
+                    {
+                        scaleMonitorsToSample.Add(monitor);
+                    }
+                }
+            }
+            else
+            {
+                scaleMonitorsToSample.AddRange(scaleMonitors);
+            }
         }
     }
 }
--- a/src/WebJobs.Script/Scale/FunctionsScaleManager.cs
+++ b/src/WebJobs.Script/Scale/FunctionsScaleManager.cs
@@ -4,8 +4,10 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Host.Scale;
+using Microsoft.Azure.WebJobs.Script.Config;
 using Microsoft.Extensions.Logging;
 
 namespace Microsoft.Azure.WebJobs.Script.Scale
@@ -17,6 +19,10 @@ namespace Microsoft.Azure.WebJobs.Script.Scale
     {
         private readonly IScaleMonitorManager _monitorManager;
         private readonly IScaleMetricsRepository _metricsRepository;
+        private readonly ITargetScalerManager _targetScalerManager;
+        private readonly IConcurrencyStatusRepository _concurrencyStatusRepository;
+        private readonly IFunctionsHostingConfiguration _functionsHostingConfiguration;
+        private readonly IEnvironment _environment;
         private readonly ILogger _logger;
 
         // for mock testing only
@@ -24,10 +30,21 @@ namespace Microsoft.Azure.WebJobs.Script.Scale
         {
         }
 
-        public FunctionsScaleManager(IScaleMonitorManager monitorManager, IScaleMetricsRepository metricsRepository, ILoggerFactory loggerFactory)
+        public FunctionsScaleManager(
+            IScaleMonitorManager monitorManager,
+            IScaleMetricsRepository metricsRepository,
+            ITargetScalerManager targetScalerManager,
+            IConcurrencyStatusRepository concurrencyStatusRepository,
+            IFunctionsHostingConfiguration functionsHostingConfiguration,
+            IEnvironment environment,
+            ILoggerFactory loggerFactory)
         {
             _monitorManager = monitorManager;
             _metricsRepository = metricsRepository;
+            _targetScalerManager = targetScalerManager;
+            _concurrencyStatusRepository = concurrencyStatusRepository;
+            _functionsHostingConfiguration = functionsHostingConfiguration;
+            _environment = environment;
             _logger = loggerFactory.CreateLogger<FunctionsScaleManager>();
         }
 
@@ -39,13 +56,17 @@ namespace Microsoft.Azure.WebJobs.Script.Scale
         /// <returns>The scale vote.</returns>
         public virtual async Task<ScaleStatusResult> GetScaleStatusAsync(ScaleStatusContext context)
         {
-            var monitors = _monitorManager.GetMonitors();
+            var scaleMonitors = _monitorManager.GetMonitors();
+            var targetScalers = _targetScalerManager.GetTargetScalers();
+
+            Utility.GetScaleInstancesToProcess(_environment, _functionsHostingConfiguration, scaleMonitors, targetScalers,
+                out List<IScaleMonitor> scaleMonitorsToProcess, out List<ITargetScaler> targetScalersToProcess);
 
             List<ScaleVote> votes = new List<ScaleVote>();
-            if (monitors.Any())
+            if (scaleMonitorsToProcess.Any())
             {
                 // get the collection of current metrics for each monitor
-                var monitorMetrics = await _metricsRepository.ReadMetricsAsync(monitors);
+                var monitorMetrics = await _metricsRepository.ReadMetricsAsync(scaleMonitorsToProcess);
 
                 _logger.LogDebug($"Computing scale status (WorkerCount={context.WorkerCount})");
                 _logger.LogDebug($"{monitorMetrics.Count} scale monitors to sample");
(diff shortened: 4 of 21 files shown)


src/WebJobs.Script/Scale/FunctionsScaleManager.cs
```diff
@@ -84,16 +94,62 @@ public virtual async Task<ScaleStatusResult> GetScaleStatusAsync(ScaleStatusCont
...
+            var targetScalers = _targetScalerManager.GetTargetScalers();
+            List<int> targetScaleVotes = new List<int>();
+
+            if (targetScalers.Any())
+            {
+                _logger.LogDebug($"{targetScalers.Count()} target scalers to sample");
+                HostConcurrencySnapshot shapshot = null;
+                try
+                {
+                    shapshot = await _concurrencyStatusRepository.ReadAsync(CancellationToken.None);
```
chiangvincent: nit: typo for `snapshot`?

src/WebJobs.Script/Scale/FunctionsScaleManager.cs
```diff
@@ -84,16 +94,62 @@ public virtual async Task<ScaleStatusResult> GetScaleStatusAsync(ScaleStatusCont
...
+                }
+
+                foreach (var targetScaler in targetScalers)
+                {
+                    try
+                    {
+                        TargetScaleStatusContext targetScaleStatusContext = new TargetScaleStatusContext();
+                        if (shapshot != null)
+                        {
+                            if (shapshot.FunctionSnapshots.TryGetValue(targetScaler.TargetScalerDescriptor.FunctionId, out var functionShaphsot))
```
chiangvincent: nit: typo `functionSnapshot`

mathewc: shapshot as well :)

src/WebJobs.Script/Scale/ScaleStatusResult.cs
```diff
@@ -15,6 +15,6 @@ namespace Microsoft.Azure.WebJobs.Script.Scale
     /// </remarks>
     public class ScaleStatusResult
     {
-        public ScaleVote Vote { get; set; }
+        public int Vote { get; set; }
```
mathewc: I don't think we should change this public protocol contract. We can still set the enum value correctly based on current worker count and the new target worker count in the case of TBS, and just add a new TargetWorkerCount property to ScaleStatusResult for TBS, which is only set if TBS is being used.

src/WebJobs.Script/Scale/FunctionsScaleManager.cs
```diff
@@ -84,16 +94,62 @@ public virtual async Task<ScaleStatusResult> GetScaleStatusAsync(ScaleStatusCont
...
+                            }
+                        }
+                        TargetScalerResult result = await targetScaler.GetScaleResultAsync(targetScaleStatusContext);
+                        _logger.LogDebug($"Target scaler '{targetScaler.TargetScalerDescriptor.Id}' voted '{result.WorkerCountDifference}'");
+                        targetScaleVotes.Add(result.WorkerCountDifference);
+                    }
+                    catch (Exception exc) when (!exc.IsFatal())
+                    {
+                        // if a particular target scaler fails, log and continue
+                        _logger.LogError(exc, $"Failed to query scale result for scale monitor '{targetScaler.TargetScalerDescriptor.Id}'.");
```
mathewc: not a "monitor"

seilorjunior:
@seilorjunior

src/WebJobs.Script/Scale/ScaleStatusResult.cs
chiangvincent: I do think it is cleaner if this contract changes to an `int`, especially in the future if we add additional scaling algorithms. The current ScaleVote is essentially returns an int changeWorkerCount (-1, 0 or 1) -- if we were to add new scaling algorithms in the future, does that mean adding a different attribute in this object for each scaling type?

alrod: This is also related to @mathewc's comments:
https://github.com/Azure/azure-webjobs-sdk/pull/2918#discussion_r980318906
https://github.com/Azure/azure-webjobs-sdk/pull/2918#discussion_r980319382

I am calculating the worker count difference in an extension. If we are introducing the new property the SC will be responsible to determinate the difference in workers.

chiangvincent: We can mark this as resolved, discussed offline and we should be keeping the current contract for ScaleVote, and add a separate attribute `TargetWorkerCount`

src/WebJobs.Script/Utility.cs
```diff
@@ -996,6 +997,52 @@ public static FunctionAppContentEditingState GetFunctionAppContentEditingState(I
...
+            out List<IScaleMonitor> scaleMonitrsToProcess,
+            out List<ITargetScaler> targetScalersToProcess)
+        {
+            scaleMonitrsToProcess = new List<IScaleMonitor>();
+            targetScalersToProcess = new List<ITargetScaler>();
+
+            string value = environment.GetEnvironmentVariableOrDefault(EnvironmentSettingNames.TargetBaseScalingEnabled, "0");
+
+            // Check if TBS enabled on stamp level
+            if (value == "1")
```
chiangvincent: `TARGET_BASED_SCALING_ENABLED` accepts `true` (case-insensitive), or `1` in scale controller -- please add this here!

alrod: In [function](https://learn.microsoft.com/en-us/azure/azure-functions/functions-app-settings) for we use "1" as equivalent of "true".

for example:
https://learn.microsoft.com/en-us/azure/azure-functions/functions-app-settings#website_run_from_package

should we also use "true". TARGET_BASED_SCALING_ENABLED is not documented yet.

src/WebJobs.Script/Utility.cs
```diff
@@ -996,6 +997,56 @@ public static FunctionAppContentEditingState GetFunctionAppContentEditingState(I
...
+            {
+                string flags = hostingConfiguration.GetValue(ScriptConstants.ScaleControllerFeatureFlags, null).ToLower();
```
chiangvincent: I don't think this will work with the existing `Parse` method in `FunctionsHostingConfiguration.cs` -- the first step in that method is to split the app setting entries using "," or ";". This will cause issues with ScaleControllerFeatureFlags, since it itself contains a comma-separated list of enabled TargetScalers

src/WebJobs.Script/Utility.cs
```diff
@@ -996,6 +997,56 @@ public static FunctionAppContentEditingState GetFunctionAppContentEditingState(I
...
+                    // we made the assumption that if a listener implements both IScaleMonitor and ITargetScaler
+                    // IScaleMonitor.Descriptor.Id must start with "{functionId}-". For example: "{functionId}-QueueTrigger-{queueName}"
+                    var array = monitor.Descriptor.Id.Split("-");
+                    string functionId = array.Length > 0 ? array[0] : null;
+                    if (!string.IsNullOrEmpty(functionId))
+                    {
+                        ITargetScaler scaler = targetScalersToProcess.SingleOrDefault(x => x.TargetScalerDescriptor.FunctionId == functionId);
+                        if (scaler == null)
+                        {
+                            scaleMonitrsToProcess.Add(monitor);
```
chiangvincent: nit: typo scaleMonitorsToProcess

src/WebJobs.Script/Utility.cs
```diff
@@ -996,6 +997,56 @@ public static FunctionAppContentEditingState GetFunctionAppContentEditingState(I
...
+            IEnumerable<ITargetScaler> targetScalers,
+            out List<IScaleMonitor> scaleMonitrsToProcess,
+            out List<ITargetScaler> targetScalersToProcess)
+        {
+            scaleMonitrsToProcess = new List<IScaleMonitor>();
+            targetScalersToProcess = new List<ITargetScaler>();
+
+            string value = environment.GetEnvironmentVariableOrDefault(EnvironmentSettingNames.TargetBaseScalingEnabled, "0");
+
+            // Check if TBS enabled on stamp level
```
chiangvincent: nit: this is an app setting on a per-app level

src/WebJobs.Script/Utility.cs
```diff
@@ -996,6 +997,56 @@ public static FunctionAppContentEditingState GetFunctionAppContentEditingState(I
...
+                        {
+                            targetScalersToProcess.Add(scaler);
+                        }
+                    }
+                }
+
+                foreach (var monitor in scaleMonitors)
+                {
+                    // we made the assumption that if a listener implements both IScaleMonitor and ITargetScaler
+                    // IScaleMonitor.Descriptor.Id must start with "{functionId}-". For example: "{functionId}-QueueTrigger-{queueName}"
```
mathewc: This isn't a valid assumption. Note that the definition of the interface [here](https://github.com/Azure/azure-webjobs-sdk/blob/919d2d63419691166a6e36f1c1785f4c2c31aeb3/src/Microsoft.Azure.WebJobs.Host/Scale/ScaleMonitorDescriptor.cs#L17) indicates that this is just an opaque ID the format of which is completely determined by the extension.

Instead of using FunctionId as a match key, what about enabling based on the extension assembly name (e.g. Microsoft.WebJobs.Sdk.Extensions.ServiceBus). Then we simply add all scalers for an enabled extension assembly, and skip adding monitors for enabled assemblies. That way we don't have to do a match based on monitor/scaler ID.

src/WebJobs.Script/Utility.cs
```diff
@@ -996,6 +997,56 @@ public static FunctionAppContentEditingState GetFunctionAppContentEditingState(I
             }
         }
 
+        public static void GetScaleInstancesToProcess(
```
mathewc: It would be helpful to have xml doc for methods indicating what they're for. Otherwise, readers (like me) have to grok through the code to figure it all out.

src/WebJobs.Script/Utility.cs
```diff
@@ -996,6 +997,56 @@ public static FunctionAppContentEditingState GetFunctionAppContentEditingState(I
...
+            IEnvironment environment,
+            IFunctionsHostingConfiguration hostingConfiguration,
+            IEnumerable<IScaleMonitor> scaleMonitors,
+            IEnumerable<ITargetScaler> targetScalers,
+            out List<IScaleMonitor> scaleMonitrsToProcess,
```
mathewc: spelling

src/WebJobs.Script/Config/FunctionsHostingConfiguration.cs
```diff
@@ -77,10 +77,6 @@ public string GetValue(string key, string defaultValue = null)
                             _configs = configs;
                             _logger.LogDebug($"Updaiting FunctionsHostingConfigurations '{settings}'");
                         }
-                        else
-                        {
-                            throw new InvalidOperationException("FunctionsHostingConfigurations does not exists");
-                        }
```
mathewc, line 83: ?

src/WebJobs.Script/Config/IFunctionsHostingConfiguration.cs
```diff
@@ -3,8 +3,10 @@
 
 namespace Microsoft.Azure.WebJobs.Script.Config
 {
-    internal interface IFunctionsHostingConfiguration
+    public interface IFunctionsHostingConfiguration
     {
         bool FunctionsWorkerDynamicConcurrencyEnabled { get; }
+
+        string GetValue(string key, string defaultValue = null);
```
mathewc, line 10: Previously this interface provided a programming model on top of raw config. Why do we need to expose the raw config now? I.e. why not just expose nice properties for the values you want to expose (e.g. ScaleControllerFeatureFlags), or perhaps expose booleans for the flags we care about?

src/WebJobs.Script/Config/IFunctionsHostingConfiguration.cs
```diff
@@ -3,8 +3,10 @@
 
 namespace Microsoft.Azure.WebJobs.Script.Config
 {
-    internal interface IFunctionsHostingConfiguration
+    public interface IFunctionsHostingConfiguration
```
mathewc, line 6: why the change here?

src/WebJobs.Script/Utility.cs
chiangvincent: Resolving this conversation via offline conversation with @pragnagopa. Ok to check if value == "1" only.

src/WebJobs.Script/Utility.cs
```diff
@@ -996,6 +997,56 @@ public static FunctionAppContentEditingState GetFunctionAppContentEditingState(I
...
+
+            // Check if TBS enabled on stamp level
+            if (value == "1")
+            {
+                string flags = hostingConfiguration.GetValue(ScriptConstants.ScaleControllerFeatureFlags, null).ToLower();
+                if (!string.IsNullOrEmpty(flags))
+                {
+                    foreach (var scaler in targetScalers)
+                    {
+                        if (flags.Contains(scaler.GetType().Name.ToLower()))
```
mathewc: To ensure we don't have casing issues, recommend doing a case insensitive search. That way we don't have to lowercase the feature flags which makes them less readable

src/WebJobs.Script/Scale/FunctionsScaleManager.cs
```diff
@@ -39,13 +56,17 @@ public FunctionsScaleManager(IScaleMonitorManager monitorManager, IScaleMetricsR
         /// <returns>The scale vote.</returns>
         public virtual async Task<ScaleStatusResult> GetScaleStatusAsync(ScaleStatusContext context)
         {
-            var monitors = _monitorManager.GetMonitors();
+            var scaleMonitors = _monitorManager.GetMonitors();
+            var targetScalers = _targetScalerManager.GetTargetScalers();
+
+            Utility.GetScaleInstancesToProcess(_environment, _functionsHostingConfiguration, scaleMonitors, targetScalers,
```
mathewc: Can we make this an instance method on FunctionsScaleManager named GetScaleStatusProviders that doesn't take any inputs, and returns the scaleMonitors/targetScalers as output parameters? FunctionsScaleMonitorService can take FunctionsScaleManager as a ctor param and call the same method.

This will allow us to initialize target scalers as Enumerable.Empty<ITargetScaler>() and avoid instantiation of target scalers when TBS isn't enabled.

src/WebJobs.Script/Scale/FunctionsScaleManager.cs
```diff
@@ -84,11 +105,52 @@ public virtual async Task<ScaleStatusResult> GetScaleStatusAsync(ScaleStatusCont
...
+            {
+                _logger.LogDebug($"{targetScalersToProcess.Count()} target scalers to sample");
+                HostConcurrencySnapshot snapshot = null;
+                try
+                {
+                    snapshot = await _concurrencyStatusRepository.ReadAsync(CancellationToken.None);
+                }
+                catch (Exception exc) when (!exc.IsFatal())
+                {
+                    _logger.LogError(exc, $"Failed to read concurrency status repository");
```
mathewc, line 138: If we have a failure here, is it really the right thing to do to silently fail and continue to compute scale status based on incomplete info?

src/WebJobs.Script/Scale/FunctionsScaleManager.cs
```diff
@@ -84,11 +105,52 @@ public virtual async Task<ScaleStatusResult> GetScaleStatusAsync(ScaleStatusCont
                 // this can happen if the host is offline
             }
 
-            var vote = GetAggregateScaleVote(votes, context, _logger);
+            List<int> targetScaleVotes = new List<int>();
+
+            if (targetScalersToProcess.Any())
```
mathewc, line 128: Let's break these two blocks out into private instance helpers GetScaleMonitorsResult/GetTargetScalersResult. That'll keep the flow of this method clean.

src/WebJobs.Script/Scale/FunctionsScaleManager.cs
```diff
@@ -84,11 +105,52 @@ public virtual async Task<ScaleStatusResult> GetScaleStatusAsync(ScaleStatusCont
...
+                    }
+                    catch (Exception exc) when (!exc.IsFatal())
+                    {
+                        // if a particular target scaler fails, log and continue
+                        _logger.LogError(exc, $"Failed to query scale result for target scaler '{targetScaler.TargetScalerDescriptor.FunctionId}'.");
+                    }
+                }
+            }
 
+            var vote = GetAggregateScaleVote(votes, context, _logger);
```
mathewc: If there are only target scalers and no votes, this will result in a ScaleIn vote, which really isn't consistent if we end up returning a worker count that indicates scale out. One thing we could do is have GetTargetScalersResult return an IEnumerable of TargetScalerVote where TargetScalerVote contains a ScaleVote and TargetWorkerCount. As these votes are added, ScaleVote can be set correctly based on comparing TargetWorkerCount and WorkerCount. Then, we can pass in the monitorVotes.Union(targetScalerVotes) to GetAggregateScaleVote and it'll aggregate correctly into a single ScaleVote.

test/WebJobs.Script.Tests/Controllers/Admin/HostControllerTests.cs
```diff
@@ -149,7 +149,7 @@ public async Task GetScaleStatus_RuntimeScaleModeEnabled_Succeeds()
                 WorkerCount = 5
             };
             var scaleManagerMock = new Mock<FunctionsScaleManager>(MockBehavior.Strict);
-            var scaleStatusResult = new ScaleStatusResult { Vote = ScaleVote.ScaleOut };
+            var scaleStatusResult = new ScaleStatusResult { Vote = ScaleVote.ScaleOut, TargetWorkerCount = 2 };
```
mathewc, line 152: Rather than modify an existing test, for mixed mode you should add new tests as needed

test/WebJobs.Script.Tests/Scale/FunctionsScaleMonitorServiceTests.cs
```diff
@@ -192,6 +204,25 @@ public async Task OnTimer_MonitorFailuresAreHandled()
             var metricsWritten = _metricsRepository.Metrics[monitor2].Take(5);
             Assert.Equal(testMetrics2, metricsWritten);
         }
+
+        [Fact]
+        public async Task TakeMetricsSamplesAsync_Skips_TakingMertrics()
```
mathewc: spelling. Also the test isn't clear to me - looks like you're adding a monitor below, so shouldn't that monitor be sampled?

kshyju:
@alrod Could you please update the PR description with more context on changes this PR is bringing? This will help future readers and current reviewers like me who does not fully understand the scaling scenarios and impact of this change. Thank you!

src/WebJobs.Script/Utility.cs
alrod: We decided to make it:
ScmHostingConfiguration = "ENABLE_FEATUREX=1,A=B,TimeOut=123,eventhubtargetscaler=1,servicebustargetscaler=1,queuetargetscaler=1"

src/WebJobs.Script/Utility.cs
alrod: I was thinking it's not very good idea to use string keys. Type assembly name is much better.

src/WebJobs.Script/Config/FunctionsHostingConfiguration.cs
alrod: We do not want to throw exception if the file does not exists.

src/WebJobs.Script/Config/IFunctionsHostingConfiguration.cs
alrod: `FunctionsScaleManager`,`FunctionsScaleMonitorService`, `Utility` are public classes.

src/WebJobs.Script/Config/IFunctionsHostingConfiguration.cs
alrod: For TBS we need just read a value by a key. Format:
"ENABLE_FEATUREX=1,A=B,TimeOut=123,eventhubtargetscaler=1,servicebustargetscaler=1,queuetargetscaler=1"

Also the method will be useful  to enable other features using `FunctionsHostingConfiguration`.

src/WebJobs.Script/Utility.cs
alrod: The internal dictionary is case insensitive so we do need to lowercase.

https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Config/FunctionsHostingConfiguration.cs#L100

src/WebJobs.Script/Scale/FunctionsScaleManager.cs
alrod: yes, as a `TargetScaler` has default hardcoded concurrency. Also is looks like a exception will be thrown [here ](https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host.Storage/BlobStorageConcurrencyStatusRepository.cs#L61) if `snapshotPersistenceEnabled": true`

test/WebJobs.Script.Tests/Controllers/Admin/HostControllerTests.cs
alrod: This test is checking that host controller returns 200 if runtime scale is enabled, it does not make sense to add a new test.

test/WebJobs.Script.Tests/Scale/FunctionsScaleMonitorServiceTests.cs
alrod: We testing that if there a target scaler in the extension we do not do sampling.

src/WebJobs.Script/Scale/FunctionsScaleManager.cs
alrod: I do not see any benefits doing this as we have underlying interfaces in place, also it will require refactoring `FunctionsScaleMonitorServiceTests` to initialize `FunctionsScaleManager` or introducing `IFunctionsScaleManager` to mock.

src/WebJobs.Script/Scale/TargetScalerVote.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Azure.WebJobs.Host.Scale;
+
+namespace Microsoft.Azure.WebJobs.Script.Scale
+{
+    internal class TargetScalerVote
```
chiangvincent, line 8: This will be used by the scale controller redesign directly, so this needs to be public (like ScaleVote)

chiangvincent: Ah, I see how this is used. Why have a separate class with the same attributes as `ScaleStatusResult`? If it is to purely distinguish which type of scaling algorithm is used, why not have a bool attribute `isTargetScaler` or something similar? `TargetScalerVote` appears redundant the way it's used right now

alrod: > @alrod Could you please update the PR description with more context on changes this PR is bringing? This will help future readers and current reviewers like me who does not fully understand the scaling scenarios and impact of this change. Thank you!

@kshyju,
Added a link to the issue in the PR description.

src/WebJobs.Script/ScriptConstants.cs
```diff
@@ -215,5 +215,7 @@ public static class ScriptConstants
 
         public static readonly string FunctionMetadataDirectTypeKey = "DirectType";
         public static readonly string LiveLogsSessionAIKey = "#AzFuncLiveLogsSessionId";
+
+        public static readonly string ScaleControllerFeatureFlags = "ScaleControllerFeatureFlags";
```
chiangvincent: This can be removed, as the format of ScmHostingConfiguration should be `feature1=1,eventhubtargetscaler=1,servicebustargetscaler=1`, etc. There is no longer a feature called `ScaleControllerFeatureFlags`

alrod:
> @alrod Could you please update the PR description with more context on changes this PR is bringing? This will help future readers and current reviewers like me who does not fully understand the scaling scenarios and impact of this change. Thank you!

@kshyju, added  a link to the issue in PR description.

src/WebJobs.Script/Scale/TargetScalerVote.cs
alrod: This is was based on @mathewc's comment:
https://github.com/Azure/azure-functions-host/pull/8780#discussion_r987194976
@mathewc wanted consistency in the response if there are no ScaleVotes at all.

@chiangvincent, if `TargetWorkerCount` is set in the response SC is not going to take into account `Vote`, right?

chiangvincent: Yes, `TargetWorkerCount` takes precedence.

src/WebJobs.Script/Utility.cs
```diff
@@ -996,6 +997,59 @@ public static FunctionAppContentEditingState GetFunctionAppContentEditingState(I
...
+            targetScalersToProcess = new List<ITargetScaler>();
+
+            string value = environment.GetEnvironmentVariableOrDefault(EnvironmentSettingNames.TargetBaseScalingEnabled, "0");
+
+            // Check if TBS enabled on app level
+            if (value == "1")
+            {
+                foreach (var scaler in targetScalers)
+                {
+                    string flag = hostingConfiguration.GetValue(scaler.GetType().Name, null);
```
alrod: need to check if scaler is `null`

src/WebJobs.Script/Config/FunctionsHostingConfiguration.cs
```diff
@@ -19,6 +19,7 @@ internal class FunctionsHostingConfiguration : IFunctionsHostingConfiguration
 
         private DateTime _configTTL;
         private Dictionary<string, string> _configs;
+        private bool isConfigurationFileExists = true;
```
mathewc: fields should have an _ prefix. Also, recommend renaming to configurationFileExists

src/WebJobs.Script/Scale/FunctionsScaleManager.cs
mathewc: All right. You might consider removing lines 59/60, and pass the monitorManager and targetScalerManager into the method so it'll do everything internally. That way we don't have the 4 collections floating around at call sites. Also, lets move the static method to FunctionsScaleManager.

src/WebJobs.Script/Utility.cs
```diff
@@ -996,6 +997,59 @@ public static FunctionAppContentEditingState GetFunctionAppContentEditingState(I
...
+            IFunctionsHostingConfiguration hostingConfiguration,
+            IEnumerable<IScaleMonitor> scaleMonitors,
+            IEnumerable<ITargetScaler> targetScalers,
+            out List<IScaleMonitor> scaleMonitorsToProcess,
+            out List<ITargetScaler> targetScalersToProcess)
+        {
+            scaleMonitorsToProcess = new List<IScaleMonitor>();
+            targetScalersToProcess = new List<ITargetScaler>();
+
+            string value = environment.GetEnvironmentVariableOrDefault(EnvironmentSettingNames.TargetBaseScalingEnabled, "0");
```
mathewc: Let's add a bool environment.IsTargetBasedScalingEnabled helper like we do for many other settings.

src/WebJobs.Script/Utility.cs
mathewc: I really think the hosting config for targetScalers should be at the assembly level, to match the assembly level check we're doing below. That keeps things in sync with the below, and keeps us out of the business of depending on implementation class names which technically could change since they're an internal implementation detail.

src/WebJobs.Script/Utility.cs
```diff
@@ -996,6 +997,59 @@ public static FunctionAppContentEditingState GetFunctionAppContentEditingState(I
...
+                    {
+                        targetScalersToProcess.Add(scaler);
+                    }
+                }
+
+                foreach (var monitor in scaleMonitors)
+                {
+                    string monitorAssemblyName = monitor.GetType().Assembly.FullName;
+                    // Check if there are scale monitor and target scaler defined in the same assembly
+                    ITargetScaler scaler = targetScalersToProcess.FirstOrDefault(x => x.GetType().Assembly.FullName == monitorAssemblyName);
```
mathewc: Recommend building a HashSet<Assembly> in the loop above then this is just a simple existence check based on assembly.

src/WebJobs.Script/Utility.cs
```diff
@@ -996,6 +997,59 @@ public static FunctionAppContentEditingState GetFunctionAppContentEditingState(I
...
+
+                    if (scaler == null)
+                    {
+                        scaleMonitorsToProcess.Add(monitor);
+                    }
+                }
+            }
+            else
+            {
+                scaleMonitorsToProcess = new List<IScaleMonitor>(scaleMonitors);
```
mathewc: Don't new up another list here - just AddRange on the one you already created

src/WebJobs.Script/Utility.cs
```diff
@@ -996,6 +997,59 @@ public static FunctionAppContentEditingState GetFunctionAppContentEditingState(I
             }
         }
 
+        /// <summary>
+        /// Returns scale monitors and target scalers we want to use based on the configuration.
+        /// Scaler monitor will be ignored if a target scaler is defined in the same extensions assembly and TBS is enabled.
+        /// </summary>
+        /// <param name="environment">Environment variables.</param>
+        /// <param name="hostingConfiguration">Housing configuration.This is used to enable TDS on stamp level for specific triggers.</param>
```
mathewc: Spelling

src/WebJobs.Script/Utility.cs
```diff
@@ -996,6 +997,59 @@ public static FunctionAppContentEditingState GetFunctionAppContentEditingState(I
...
+        /// <param name="scaleMonitors">Registered scale monitors.</param>
+        /// <param name="targetScalers">Registered target scalers.</param>
+        /// <param name="scaleMonitorsToProcess">Scale monitor to process.</param>
+        /// <param name="targetScalersToProcess">Target scaler to process.</param>
+        public static void GetScaleInstancesToProcess(
```
mathewc: nit: Can we rename this to GetScalersToSample?

src/WebJobs.Script/Scale/FunctionsScaleManager.cs
```diff
@@ -84,15 +116,69 @@ public virtual async Task<ScaleStatusResult> GetScaleStatusAsync(ScaleStatusCont
                 // this can happen if the host is offline
             }
 
-            var vote = GetAggregateScaleVote(votes, context, _logger);
+            ScaleVote vote = GetAggregateScaleVote(votes.Union(targetScaleVotes), context, _logger);
```
mathewc: Let's move this aggregation out of this method - this method should only be about monitors, and shouldn't be taking any targetScaleVotes as a parameter. This method should just return it's collection of ScaleVotes

src/WebJobs.Script/Scale/FunctionsScaleManager.cs
```diff
@@ -84,15 +116,69 @@ public virtual async Task<ScaleStatusResult> GetScaleStatusAsync(ScaleStatusCont
...
+                        });
+                    }
+                    catch (Exception exc) when (!exc.IsFatal())
+                    {
+                        // if a particular target scaler fails, log and continue
+                        _logger.LogError(exc, $"Failed to query scale result for target scaler '{targetScaler.TargetScalerDescriptor.FunctionId}'.");
+                    }
+                }
+            }
+            //int? targetWorkerCount = targetScaleVotes.Any() ? targetScaleVotes.Max() : null;
```
mathewc: remove this commented out line of code

src/WebJobs.Script/Scale/FunctionsScaleManager.cs
```diff
@@ -39,13 +56,28 @@ public FunctionsScaleManager(IScaleMonitorManager monitorManager, IScaleMetricsR
...
 
+            Utility.GetScaleInstancesToProcess(_environment, _functionsHostingConfiguration, scaleMonitors, targetScalers,
+                out List<IScaleMonitor> scaleMonitorsToProcess, out List<ITargetScaler> targetScalersToProcess);
+
+            var targetScalerVotes = await GetTargetScalersResult(context, targetScalersToProcess);
+
+            return new ScaleStatusResult
+            {
+                Vote = await GetScaleMonitorsResult(context, scaleMonitorsToProcess, targetScalerVotes.Select(x => x.Vote)),
+                TargetWorkerCount = targetScalerVotes.Any() ? targetScalerVotes.Select(x => x.TargetWorkerCount).Max() : null
```
mathewc: nit: simplify the max operation to targetScalerVotes.Max(p => p.TargetWorkerCount)

src/WebJobs.Script/Scale/FunctionsScaleManager.cs
```diff
@@ -84,15 +116,69 @@ public virtual async Task<ScaleStatusResult> GetScaleStatusAsync(ScaleStatusCont
                 // this can happen if the host is offline
             }
 
-            var vote = GetAggregateScaleVote(votes, context, _logger);
+            ScaleVote vote = GetAggregateScaleVote(votes.Union(targetScaleVotes), context, _logger);
+            return vote;
+        }
 
-            return new ScaleStatusResult
+        private async Task<IEnumerable<TargetScalerVote>> GetTargetScalersResult(ScaleStatusContext context, IEnumerable<ITargetScaler> targetScalersToProcess)
```
mathewc: Follow the Async naming suffix pattern for this and any other async methods

src/WebJobs.Script/Scale/FunctionsScaleManager.cs
```diff
@@ -84,15 +116,69 @@ public virtual async Task<ScaleStatusResult> GetScaleStatusAsync(ScaleStatusCont
...
+                        TargetScalerContext targetScaleStatusContext = new TargetScalerContext();
+                        if (snapshot != null)
+                        {
+                            if (snapshot.FunctionSnapshots.TryGetValue(targetScaler.TargetScalerDescriptor.FunctionId, out var functionSnapshot))
+                            {
+                                targetScaleStatusContext.InstanceConcurrency = functionSnapshot.Concurrency;
+                                _logger.LogDebug($"Snapshot dynamic concurrency for target scaler '{targetScaler.TargetScalerDescriptor.FunctionId}' is '{functionSnapshot.Concurrency}'");
+                            }
+                        }
+                        TargetScalerResult result = await targetScaler.GetScaleResultAsync(targetScaleStatusContext);
```
mathewc, line 155: nit: please space this code out for readability. E.g. space on lines 153, 155, 164. Basically space apart logical blocks

src/WebJobs.Script/Utility.cs
alrod: @mathewc , you are saying that we just should use assembly name as the key in the hosting configuration? It's a good  point.

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -601,5 +601,10 @@ public static string GetFunctionsExtensionVersion(this IEnvironment environment)
         {
             return environment.GetEnvironmentVariableOrDefault(FunctionsExtensionVersion, string.Empty);
         }
+
+        public static bool IsTargetBasedScalingEnabled(this IEnvironment environment)
+        {
+            return string.Equals(environment.GetEnvironmentVariable(TargetBaseScalingEnabled), "1");
```
pragnagopa, line 607: This should check for both App Setting and Hosting Configuration?

src/WebJobs.Script/Utility.cs
```diff
@@ -996,6 +997,58 @@ public static FunctionAppContentEditingState GetFunctionAppContentEditingState(I
...
+            targetScalersToSample = new List<ITargetScaler>();
+
+            // Check if TBS enabled on app level
+            if (environment.IsTargetBasedScalingEnabled())
+            {
+                HashSet<string> targetScalerAssemblies = new HashSet<string>();
+                foreach (var scaler in targetScalers.Where(x => x != null))
+                {
+                    string assemblyName = scaler.GetType().Assembly.GetName().Name;
+                    string flag = hostingConfiguration.GetValue(assemblyName, null);
```
pragnagopa: I recommend moving this check into the Environment extensions - so `IsTargetBasedScalingEnabled` will have logic to check both app setting and hosting configuration

pragnagopa (commented):
added minor comments. Thanks!

src/WebJobs.Script/Utility.cs
alrod: We decided to implement the checks (app setting and hosting config) is in the function host to avoid duplication of the code in all the extensions - extensions just need to implement the required interface to support TBS.

Also SC will have his own hosting configuration check implementation as it reads hosting config directly from SQL.

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
alrod: HostingConfig is checked here:
https://github.com/Azure/azure-functions-host/pull/8780/files#diff-70d41caa568fa2a59c53884cf200b3f224fae1d27962367f9b6b151cd7135685R1028

src/WebJobs.Script/Scale/TargetScalerVote.cs
alrod: `TargetScalerVote` is added to calculate correct value for `Vote` if there TBS is enabled. Even if SC does not use we are adding the logic for consistency in the response.

src/WebJobs.Script/Utility.cs
mathewc: Yes

src/WebJobs.Script/Scale/FunctionsScaleManager.cs
```diff
@@ -39,13 +57,28 @@ public FunctionsScaleManager(IScaleMonitorManager monitorManager, IScaleMetricsR
         /// <returns>The scale vote.</returns>
         public virtual async Task<ScaleStatusResult> GetScaleStatusAsync(ScaleStatusContext context)
         {
-            var monitors = _monitorManager.GetMonitors();
+            GetScaleStatusProviders(_monitorManager, _targetScalerManager, out IEnumerable<IScaleMonitor> scaleMonitors, out IEnumerable<ITargetScaler> targetScalers);
```
mathewc: This GetScaleStatusProviders method isn't providing any value. Let's remove it and just pass in the managers to GetScalersToSample and have it do everything internally. Also, move the utility method GetScalersToSample from Utility class to FunctionsScaleManager please.

src/WebJobs.Script/Utility.cs
```diff
@@ -996,6 +997,58 @@ public static FunctionAppContentEditingState GetFunctionAppContentEditingState(I
...
+            out List<ITargetScaler> targetScalersToSample)
+        {
+            scaleMonitorsToSample = new List<IScaleMonitor>();
+            targetScalersToSample = new List<ITargetScaler>();
+
+            // Check if TBS enabled on app level
+            if (environment.IsTargetBasedScalingEnabled())
+            {
+                HashSet<string> targetScalerAssemblies = new HashSet<string>();
+                foreach (var scaler in targetScalers.Where(x => x != null))
```
mathewc: The targetScalers collection should never contain nulls - remove this null check

src/WebJobs.Script/Utility.cs
```diff
@@ -996,6 +997,58 @@ public static FunctionAppContentEditingState GetFunctionAppContentEditingState(I
...
+                    string assemblyName = scaler.GetType().Assembly.GetName().Name;
+                    string flag = hostingConfiguration.GetValue(assemblyName, null);
+                    if (flag == "1")
+                    {
+                        targetScalersToSample.Add(scaler);
+                        targetScalerAssemblies.Add(assemblyName);
+                    }
+                }
+
+                foreach (var monitor in scaleMonitors.Where(x => x != null))
```
mathewc: Same here - this null check should be removed

test/WebJobs.Script.Tests/Configuration/FunctionsHostingConfigurationTest.cs
```diff
@@ -43,30 +41,24 @@ public void FunctionsHostingConfiguration_Registered_ByDefault()
...
         [Fact]
-        public void FunctionsWorkerDynamicConcurrencyEnabled_Throws_InvalidOperationException()
-        {
-            FunctionsHostingConfiguration conf = new FunctionsHostingConfiguration(_environment, _loggerFactory, Path.Combine("C:\\", "test.txt"), DateTime.Now.AddMilliseconds(1), TimeSpan.FromMinutes(5));
-            Assert.Throws<InvalidOperationException>(() => conf.FunctionsWorkerDynamicConcurrencyEnabled);
-        }
-
-        [Fact]
-        public async Task FunctionsWorkerDynamicConcurrencyEnabled_UpdatesSettings()
+        public async Task GetValue_UpdatesSettings()
```
mathewc: nit: GetValue doesn't update any settings - it just returns a value, right. Maybe the test should be named GetValue_ReturnsExpectedValue?

test/WebJobs.Script.Tests/Configuration/FunctionsHostingConfigurationTest.cs
```diff
@@ -134,5 +127,12 @@ public void ReturnsExpectedConfig(string config, int configCount)
                 Assert.True(conf.Config.Count == configCount);
             }
         }
+
+        [Fact]
+        public void GetValue_ConfigurationFileDoesNotExists()
```
mathewc: nit: in this test, the configuration file DOES actually exist right? Should it be named something like Constructor_ConfigFileExists_DoesNotLog? What about the other test - Constructor_ConfigFileDoesNotExist_Logs

test/WebJobs.Script.Tests/Configuration/FunctionsHostingConfigurationTest.cs
alrod: The idea of the test is checking that the settings are updated if the config file was rewritten. 
In the test we write to the config file twice and set short cache time for the settings.

mathewc: I see - that's why descriptive test names are important :) if the test were named GetValue_ReloadsConfig_OnUpdate I'd have understood :)

mathewc (approved):

chiangvincent (approved):

PR #8796, opened 2022-09-28
 Adding telemetry to analyze failing workers via Kusto logs

### LINKED ISSUE #7292: Recycle failing language workers
Have seen a few scenarios where the functions host is running multiple language workers and some of them are consistently failing, while others are working fine. Often this involves some user code component going into a bad state.

In most cases this is resolved with a restart on the function app, but this is a heavy handed approach that requires manual intervention.  The functions host should detect this scenario and attempt to resolve it by recycling the language workers that are failing. It will still be up to the user to analyze their logs and fix their problems, but the platform can reduce the impact until that happens by automatically recycling failing workers.

The single language worker case is a bit trickier. There's no obvious signal that a restart will fix the problem. I'd suggest the first version of this improvement should only kick in when there are multiple language workers.

surgupta-msft:

This is part-1 of issue - recycling failing language workers (#7292)

Design - [doc](https://dev.azure.com/msazure/One/_git/AAPT-Antares-Docs?path=/TeamDocs/FunctionTeamDocs/Design/OOPWorkers/%5BIn%20Progress%5D%20RecycleFailingLanguageWorkers.md&version=GBsurgupta/recycle-failing-workers&_a=preview) 

**Update -** Based on recent discussions, the first part of this feature (#7292) to have only additional logs and metrics to see how many production apps are currently impacted with this issue and how many workers will be recycled once the feature is in. Once we analyze Prod apps and decide the value for failureThreshold, we can implement part-2 with logic to identify and recycle failing workers.

Once this PR is released, a kusto query like below will give the needed information.

```kusto
let FailureThreshold = 60;
let TotalInvocations = FunctionsMetrics
| where PreciseTimeStamp >= datetime(2022-10-12 03:00:0)
| where AppName == "surg-net-test64"
| where EventName endswith "worker.invoke"
| project EventName, WorkerId = trim(" ",substring(EventName, 0, 36))
| summarize TotalInvocationsCount=count() by EventName, WorkerId;
let SuccessfulInvocations = FunctionsMetrics
| where PreciseTimeStamp >= datetime(2022-10-12 03:00:0)
| where AppName == "surg-net-test64"
| where EventName endswith "worker.invoke.succeeded"
| project EventName, WorkerId = trim(" ",substring(EventName, 0, 36))
| summarize SuccessfulInvocationsCount=count() by EventName, WorkerId;
TotalInvocations
| join kind = fullouter SuccessfulInvocations on WorkerId
| extend FailedInvocations=TotalInvocationsCount-SuccessfulInvocationsCount
| extend FailureRatePercentage = (FailedInvocations * 100)/TotalInvocationsCount
| extend WorkerToRecycle = iff(FailureRatePercentage > FailureThreshold,"Yes","No")
| order by FailureRatePercentage desc
| project WorkerId, TotalInvocationsCount, SuccessfulInvocationsCount, FailedInvocations, FailureRatePercentage, WorkerToRecycle
```

image


src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -1219,6 +1252,65 @@ internal async void OnTimer(object sender, System.Timers.ElapsedEventArgs e)
...
+                // Specifically ignore this race - we're exiting and that's okay
+            }
+        }
+
+        internal void CheckAndRecycleFaultyLanguageWorker()
+        {
+            _workerChannelLogger.LogInformation("Checking and recycling faulty language workers");
+
+            Dictionary<string, int> failureRatePerWorkerId = new Dictionary<string, int>();
+            var failureThreshold = 30; // ToDo: Moving this constant once finalized
```
surgupta-msft: @mathewc thoughts on `failureThreshold` value? Right now, this code will just **log** the workerIds that are identified as faulty workers to be recycled. Does `30` look good to help us examine Prod apps?

src/WebJobs.Script.Grpc/WorkerInvocationMetrics.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        public double AverageInvocationLatency { get; set; }
+
+        public static int IncrementTotalInvocationsOfWorker(string workerId, IDictionary<string, WorkerInvocationMetrics> invocationMetricsPerWorkerId)
+        {
+            invocationMetricsPerWorkerId.TryGetValue(workerId, out WorkerInvocationMetrics workerInvocationMetrics);
+            workerInvocationMetrics = workerInvocationMetrics ?? new WorkerInvocationMetrics();
+
+            workerInvocationMetrics.TotalInvocations = workerInvocationMetrics.TotalInvocations + 1;
+            invocationMetricsPerWorkerId[workerId] = workerInvocationMetrics;
```
kshyju: Is this code thread safe?  I see we are using Dictionary type which is not thread safe. Should we use ConcurrentDictionary?

src/WebJobs.Script.Grpc/WorkerInvocationMetrics.cs
```diff
@@ -0,0 +1,50 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Azure.WebJobs.Script.Grpc
+{
+    public class WorkerInvocationMetrics
```
kshyju: Should this type have `public` visibility? Would `internal` work?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -52,6 +52,8 @@ internal class GrpcWorkerChannel : IRpcWorkerChannel, IDisposable
         private readonly Dictionary<MsgType, Queue<PendingItem>> _pendingActions = new ();
         private readonly ChannelWriter<OutboundGrpcEvent> _outbound;
         private readonly ChannelReader<InboundGrpcEvent> _inbound;
+        private IDictionary<string, WorkerInvocationMetrics> _invocationMetricsPerWorkerId = new Dictionary<string, WorkerInvocationMetrics>();
```
kshyju: _Feel free to correct me if my understanding below is incorrect._

Question: There will be one instance of `GrpcWorkerChannel` for each worker instance. So each of those instances is for that specific worker instance. So, do we really need a dictionary in that case?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -787,6 +795,15 @@ internal async Task InvokeResponse(InvocationResponse invokeResponse)
             if (_executingInvocations.TryRemove(invokeResponse.InvocationId, out ScriptInvocationContext context)
                 && invokeResponse.Result.IsInvocationSuccess(context.ResultSource, capabilityEnabled))
             {
+                var currentInvocationLatency = _workerInvocationStopWatch.GetElapsedTime();
+                _workerChannelLogger.LogDebug($"[HostMonitor] Worker {Id} invocation request took {currentInvocationLatency.TotalMilliseconds}ms");
+
+                int successfulInvocations = WorkerInvocationMetrics.IncrementSuccessfulInvocationsOfWorker(Id, _invocationMetricsPerWorkerId);
+                double averageInvocationLatency = WorkerInvocationMetrics.UpdateAverageInvocationLatency(Id, _invocationMetricsPerWorkerId, currentInvocationLatency);
```
kshyju: How would this data (invocation latency) differ from the `function.invoke.latency` metric/event which we already capture and store in FunctionsMetrics?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
kshyju: See my other comment about the thread safety question. If we decide to not use a dictionary, and decided to keep int variable to store the numbers, we could use the methods from [Interlocked](https://learn.microsoft.com/en-us/dotnet/api/system.threading.interlocked?view=net-6.0) to keep the operation atomic.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -1219,6 +1252,65 @@ internal async void OnTimer(object sender, System.Timers.ElapsedEventArgs e)
...
+            {
+                _invocationTimer.Start();
+            }
+            catch (ObjectDisposedException)
+            {
+                // Specifically ignore this race - we're exiting and that's okay
+            }
+        }
+
+        internal void CheckAndRecycleFaultyLanguageWorker()
```
kshyju: Sorry If I am comprehending this incorrectly. :) 

I wonder this method belongs to this class? Recycling means killing the instance of this (GrpcWorkerChannel) class. Correct? If so, shouldn't we do this operation from another class which is still alive ? Perhaps some class which is monitoring the metric data for each worker and then initiate the recycle operation form there.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -1219,6 +1252,65 @@ internal async void OnTimer(object sender, System.Timers.ElapsedEventArgs e)
...
+            {
+                return;
+            }
+
+            try
+            {
+                CheckAndRecycleFaultyLanguageWorker();
+            }
+            catch
+            {
```
kshyju: Shouldn't we log the Exception details here?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
jviau: I agree - I do not think we need a dictionary here. The key is `workerId`, which as far as I can tell is read-only for the life of a `GrpcWorkerChannel`. So this dictionary would have at most 1 entry in it - so I think we can safely remove the dictionary and hold this values directly.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -840,6 +857,21 @@ internal async Task InvokeResponse(InvocationResponse invokeResponse)
                         SendCloseSharedMemoryResourcesForInvocationRequest(outputMaps);
                     }
                 }
+
+                string data = $"WorkerId: {Id}, TotalInvocations : {_invocationMetricsPerWorkerId[Id].TotalInvocations}, SuccessfulInvocations : {_invocationMetricsPerWorkerId[Id].SuccessfulInvocations}, AverageInvocationLatency : {_invocationMetricsPerWorkerId[Id].AverageInvocationLatency}";
+                _metricsLogger.LogEvent(string.Format(MetricEventNames.WorkerInvocationStatus, Id), data: data);
+            }
+
+            if (_invocationTimer == null)
```
jviau: This is not thread safe - two concurrent invocations may both initialize a timer. This will lead to double callbacks of `OnInvocationTimer` and also a failure to dispose one of them. Can we initialize the timer in the ctor, then just start it here?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -1219,6 +1252,65 @@ internal async void OnTimer(object sender, System.Timers.ElapsedEventArgs e)
...
+            var workerIdToRecycle = sortedFailureRatePerWorkerId.First().Key;
+            var failureRate = failureRatePerWorkerId[workerIdToRecycle];
+
+            if (failureRate > failureThreshold)
+            {
+                // ToDo: in next iteration of this feature => recycle the workerId in variable: workerIdToRecycle
+                // ToDo: Other criteria for recycle => if invocation latency is too high
+
+                _metricsLogger.LogEvent(string.Format(MetricEventNames.WorkerRecycled, Id), data: $"WorkerId to be recycled: {workerIdToRecycle}");
+                _workerChannelLogger.LogDebug($"WorkerId to be Recycled: {workerIdToRecycle}, Failure rate: {failureRate}");
```
jviau: Please us structured logging.
```suggestion
                _workerChannelLogger.LogDebug("WorkerId to be Recycled: {workerId}, Failure rate: {failureRate}", workerIdToRecycle, failureRate);
```

jviau (changes requested):
I'd like to discuss a bit about our metrics approach. This seems odd to me for a couple reasons:

1. Metrics is a measurement _over an interval of time_.
    - This PR is measuring failure rate for the lifetime of a worker. The interval is fixed to the application lifetime and not the measurement itself.
1. We are performing in-app aggregation.
    - Metrics should just emit the measurements, then the aggregation / projection happens server side.
 
If we where to add a metric for `function.invoke.count` and have a `success = true/false` dimension on it, can we not perform an aggregation over time on where the telemetry goes and determine health of the worker? Or we could just add the `success` dimension (with an appropriate name) to the `function.invoke.latency` metric mentioned by @kshyju (if it is not already there).

Now, if we really want to make sure we have an in-worker hook for the health of the worker, we should look into [Microsoft.Extensions.Diagnostics.HealthChecks](https://www.nuget.org/packages/Microsoft.Extensions.Diagnostics.HealthChecks/)

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -787,6 +795,15 @@ internal async Task InvokeResponse(InvocationResponse invokeResponse)
...
                 && invokeResponse.Result.IsInvocationSuccess(context.ResultSource, capabilityEnabled))
             {
+                var currentInvocationLatency = _workerInvocationStopWatch.GetElapsedTime();
+                _workerChannelLogger.LogDebug($"[HostMonitor] Worker {Id} invocation request took {currentInvocationLatency.TotalMilliseconds}ms");
+
+                int successfulInvocations = WorkerInvocationMetrics.IncrementSuccessfulInvocationsOfWorker(Id, _invocationMetricsPerWorkerId);
+                double averageInvocationLatency = WorkerInvocationMetrics.UpdateAverageInvocationLatency(Id, _invocationMetricsPerWorkerId, currentInvocationLatency);
+
+                string successData = $"WorkerId: {Id}, TotalInvocations : {_invocationMetricsPerWorkerId[Id].TotalInvocations}, SuccessfulInvocations : {successfulInvocations}, AverageInvocationLatency : {averageInvocationLatency}";
+                _metricsLogger.LogEvent(string.Format(MetricEventNames.WorkerInvokeSucceeded, Id), data: successData);
```
mathewc: For the initial round of changes, I was thinking we could just leverage the existing support metric events have for aggregation, and simply log count/latency events for each function completed here. We'd have event formats that include the worker Id, e.g.:

WorkerInvocationSucceeded = {0}worker.{1}.invoke.succeeded
WorkerInvocationFailed = {0}worker.{1}.invoke.failed
WorkerInvocationLatency = {0}worker.{1}.invoke.latency

For the latency event, we use BeginEvent/EndEvent as we already do for FunctionInvokeLatency. Then, MetricsEventManager will aggregate/flush these regularly. With this per worker ID data in the metrics table, we can then write Kusto queries that aggregate all this by worker ID, and we can have queries that look for apps with > 1 worker. However, I realize one problem with that is that we wouldn't be able to aggregate those events by host instance, unless we were to also include the HostInstanceId as part of the event name as well.

Using metrics events like this would give us production data we need, however I realize that if/once we decide we want to actually implement worker recycles on top of this, we'd then have to actually programmatically capture the relevant data as you're doing now.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -787,6 +795,15 @@ internal async Task InvokeResponse(InvocationResponse invokeResponse)
             if (_executingInvocations.TryRemove(invokeResponse.InvocationId, out ScriptInvocationContext context)
                 && invokeResponse.Result.IsInvocationSuccess(context.ResultSource, capabilityEnabled))
             {
+                var currentInvocationLatency = _workerInvocationStopWatch.GetElapsedTime();
+                _workerChannelLogger.LogDebug($"[HostMonitor] Worker {Id} invocation request took {currentInvocationLatency.TotalMilliseconds}ms");
```
mathewc: Why this new log? Doesn't seem related to the goal of these changes. We have to be careful adding per invocation logs like this because the have a big impact on performance as we've found over time.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -840,6 +857,21 @@ internal async Task InvokeResponse(InvocationResponse invokeResponse)
                         SendCloseSharedMemoryResourcesForInvocationRequest(outputMaps);
                     }
                 }
+
+                string data = $"WorkerId: {Id}, TotalInvocations : {_invocationMetricsPerWorkerId[Id].TotalInvocations}, SuccessfulInvocations : {_invocationMetricsPerWorkerId[Id].SuccessfulInvocations}, AverageInvocationLatency : {_invocationMetricsPerWorkerId[Id].AverageInvocationLatency}";
+                _metricsLogger.LogEvent(string.Format(MetricEventNames.WorkerInvocationStatus, Id), data: data);
```
mathewc: Regarding the data payload on these events. If you look at LogEvent [here](https://github.com/Azure/azure-functions-host/blob/dcfd7fc1458da020de2c3f49b1d9672b1e1bc229/src/WebJobs.Script.WebHost/Diagnostics/MetricsEventManager.cs#L118) you'll see that aggregations are done based on event name / function name only, so in this case the resulting data from the aggregated event will be from the first event. It's confusing and funky but that's the way it works - for a particular event, if data is being used, that data should be constant for all the events. So this won't actually do what you want it to do.

Instead, probably what you want instead is to periodically log these metrics from your timer, in a format that makes them easily queryable for our purposes. And we'd log them not as metrics events (MetricsLogger) but as regular logs at Debug level.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
mathewc: Her intent was that this is a placeholder for the work we're ultimately planning. However, because we're only logging this first round and not recycling anything, I think we can simplify, remove all the bits about recycling, thresholds, etc. and just log. Rename this method to LogInvocationMetrics

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
mathewc: To add, any time we're doing a catch all like this, we need to be using the IsFatal extension method to ensure we're not trapping OOM, etc. level exceptions. You can see other examples in the code.

src/WebJobs.Script.Grpc/WorkerInvocationMetrics.cs
surgupta-msft: Done

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
surgupta-msft: Removed unnecessary log line.

src/WebJobs.Script.Grpc/WorkerInvocationMetrics.cs
surgupta-msft: Updated to use concurrentDict

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
surgupta-msft: Done

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
surgupta-msft: Removing latency logic for now as it was not the performance efficient approach and anyway was not using it for faulty worker calculation. Will investigate the correct approach there and then update.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
surgupta-msft: Updated code here. Please check if this looks correct.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
surgupta-msft: Renamed method to `IdentifyFaultyLanguageWorker()` if this looks okay.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
surgupta-msft: Done

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
surgupta-msft: Corrected code here

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
surgupta-msft: Let's discuss this one offline. I did deploy this code to a function app and looks like the dictionary is getting updated properly - 

```kusto
https://wawscus.kusto.windows.net/wawsprod?query=H4sIAAAAAAAEAHWQT0vDQBDF74LfYchJD6GbKB4KCj1YCFgpJl9g3Z2mq9k%2f7sy2BPzwbgRtC%2b115vfmvTfL5BQb7%2bjF93R99Q37LUaErlk9t91itYanR9CSkY3Fm1rUdVmJsqpB3M2FmIvbg2QRwqu0CMo7lsYRFJRiXzrkkpH44b44sG2yVsbxiI2oRjVgAT6e2S6lGVLWxZzkAtJ5lkPjdl7J3z7ZbTY7S7botHE9mH8YIn6lnBH2hrdH80ZfMHtDhWaHGigphUSbNJyeo5AjIGyy%2buTc9AIfNUZ4H2GdSxvCLr%2b2ZWkDSFITEKL%2fQMV%2fvtOI5SdCJX4ALraBGa4BAAA%3d&web=0

FunctionsLogs
| where TIMESTAMP >= datetime(2022-10-12 03:00:0)
| where AppName contains "surg-net-test64"
| where Summary contains "recycle" or Summary contains "Failure rate" or Summary contains "TotalInvocations"
//or Summary contains "Sending invocation request with invocationId" or Summary contains "Received successful invocation response for invocationId"
| order by PreciseTimeStamp asc
| project Summary
| take 10


Summary
--
WorkerId: 320f5298-7802-4aad-ab2a-603a539ceae8, Failure rate: 20, TotalInvocations : 5, SuccessfulInvocations : 4, AverageInvocationLatency : 0
WorkerId: 3921ad56-794d-4e8d-afc6-b92b625b48dd, Failure rate: 80, TotalInvocations : 5, SuccessfulInvocations : 1, AverageInvocationLatency : 0
Faulty WorkerId to be Recycled: 3921ad56-794d-4e8d-afc6-b92b625b48dd, Failure rate: 80
WorkerId: 320f5298-7802-4aad-ab2a-603a539ceae8, Failure rate: 0, TotalInvocations : 3, SuccessfulInvocations : 3, AverageInvocationLatency : 0
WorkerId: 3921ad56-794d-4e8d-afc6-b92b625b48dd, Failure rate: 100, TotalInvocations : 3, SuccessfulInvocations : 0, AverageInvocationLatency : 0
Faulty WorkerId to be Recycled: 3921ad56-794d-4e8d-afc6-b92b625b48dd, Failure rate: 100
WorkerId: 320f5298-7802-4aad-ab2a-603a539ceae8, Failure rate: 0, TotalInvocations : 3, SuccessfulInvocations : 3, AverageInvocationLatency : 0
WorkerId: 3921ad56-794d-4e8d-afc6-b92b625b48dd, Failure rate: 100, TotalInvocations : 3, SuccessfulInvocations : 0, AverageInvocationLatency : 0
Faulty WorkerId to be Recycled: 3921ad56-794d-4e8d-afc6-b92b625b48dd, Failure rate: 100
WorkerId: 320f5298-7802-4aad-ab2a-603a539ceae8, Failure rate: 0, TotalInvocations : 4, SuccessfulInvocations : 4, AverageInvocationLatency : 0
```

kshyju: I think we do not need the ConcurrentDictionary. This instance of `GrpcWorkerChannel` is unique for that worker (see the `Id` property on this class). So in that case, we really do not need a dictionary.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -127,6 +133,36 @@ internal class GrpcWorkerChannel : IRpcWorkerChannel, IDisposable
             _startLatencyMetric = metricsLogger?.LatencyEvent(string.Format(MetricEventNames.WorkerInitializeLatency, workerConfig.Description.Language, attemptCount));
 
             _state = RpcWorkerChannelState.Default;
+
+            if (_invocationTimer == null)
```
kshyju: nit: I think you can remove this null check because you know it is null.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -52,7 +53,11 @@ internal class GrpcWorkerChannel : IRpcWorkerChannel, IDisposable
         private readonly Dictionary<MsgType, Queue<PendingItem>> _pendingActions = new ();
         private readonly ChannelWriter<OutboundGrpcEvent> _outbound;
         private readonly ChannelReader<InboundGrpcEvent> _inbound;
-
+        private Meter _workerInvocationMeter;
```
kshyju: Who/What is the consumer of metric data collected by this Meter instance?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -52,7 +53,11 @@ internal class GrpcWorkerChannel : IRpcWorkerChannel, IDisposable
...
+        private Meter _workerInvocationSucceededMeter;
```
kshyju: Meter implements `IDisposable`. Let's make sure we dispose that.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -633,6 +633,9 @@ internal async Task SendInvocationRequest(ScriptInvocationContext context)
                     context.CancellationToken.Register(() => SendInvocationCancel(invocationRequest.InvocationId));
                 }
 
+                _metricsLogger.LogEvent(string.Format(MetricEventNames.WorkerInvoked, Id), functionName: context.FunctionMetadata.Name);
+                _workerChannelLogger.LogInformation("Sending invocation request with invocationId: {invocationId} on workerId: {workerId} for function: {functionName}", invocationRequest.InvocationId, Id, context.FunctionMetadata.Name);
```
jviau: small nit: how important are these log statements? if not that important, I recommend going with `trace` level, maybe `debug`.
```suggestion
                _workerChannelLogger.LogTrace("Sending invocation request with invocationId: {invocationId} on workerId: {workerId} for function: {functionName}", invocationRequest.InvocationId, Id, context.FunctionMetadata.Name);
```

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -788,6 +791,9 @@ internal async Task InvokeResponse(InvocationResponse invokeResponse)
             if (_executingInvocations.TryRemove(invokeResponse.InvocationId, out ScriptInvocationContext context)
                 && invokeResponse.Result.IsInvocationSuccess(context.ResultSource, capabilityEnabled))
             {
+                _metricsLogger.LogEvent(string.Format(MetricEventNames.WorkerInvokeSucceeded, Id));
+                _workerChannelLogger.LogInformation("Received successful invocation response for invocationId: {invocationId} and workerId: {workerId}", invokeResponse.InvocationId, Id);
```
jviau: same question / nit here:
```suggestion
                _workerChannelLogger.LogTrace("Received successful invocation response for invocationId: {invocationId} and workerId: {workerId}", invokeResponse.InvocationId, Id);
```

jviau (approved):

kshyju (approved):

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
mathewc: I think we should remove these 2 new log statements. They're not needed for the new metrics you're logging, and will have a performance impact in high throughput scenarios. In general on the execution hotpath we want to minimize the per invocation logging we do since past performance work has shown it has a big impact.

src/WebJobs.Script/Diagnostics/MetricEventNames.cs
```diff
@@ -25,6 +25,8 @@ public static class MetricEventNames
 
         // language worker level events
         public const string WorkerInitializeLatency = "host.startup.outofproc.{0}worker.initialize.attempt{1}.latency";
+        public const string WorkerInvoked = "{0}worker.invoke";
```
mathewc, line 28: I'm thinking you should have a pair of worker.invoke.succeeded/worker.invoke.failed events similar to what we do for function invocations here https://github.com/Azure/azure-functions-host/blob/7054c51ad9666d6ae0f30da57054d9d8c457789b/src/WebJobs.Script.WebHost/Diagnostics/FunctionInstanceLogger.cs#L103.

That allows you to query the exact number of failures/successes in a given time window. With the events you have now, in a time window you would be incorrectly interpreting invocations which may be in progress but not completed as failures, right?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -785,63 +787,71 @@ internal async Task InvokeResponse(InvocationResponse invokeResponse)
...
+                        IList<string> outputMaps = GetOutputMaps(invokeResponse.OutputData);
+                        if (outputMaps.Count > 0)
+                        {
+                            // If this invocation was using any shared memory maps produced by the worker, close them to free memory
+                            SendCloseSharedMemoryResourcesForInvocationRequest(outputMaps);
+                        }
                     }
                 }
             }
+            else
```
mathewc: Looks like this else is at the wrong level - it should be for "if (invokeResponse.Result.IsInvocationSuccess"

mathewc: There should be a an existing test somewhere that we can add to to ensure these new metrics are emitted correctly. I'm assuming we have tests for worker invocations already.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -633,6 +633,8 @@ internal async Task SendInvocationRequest(ScriptInvocationContext context)
                     context.CancellationToken.Register(() => SendInvocationCancel(invocationRequest.InvocationId));
                 }
 
+                _metricsLogger.LogEvent(string.Format(MetricEventNames.WorkerInvoked, Id), functionName: context.FunctionMetadata.Name);
```
mathewc, line 636: Do we need this anymore, since we're now logging success/failure metrics?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
surgupta-msft: Done

src/WebJobs.Script/Diagnostics/MetricEventNames.cs
surgupta-msft: Done

PR #8803, opened 2022-09-29
Legion container assignment

### LINKED ISSUE #8616: Support FunctionsRuntime on Legion Host

Tonewall:
Alternate design for Legion container assignment. Instead of having a separate LegionInstanceManager and duplicating the code from the InstanceManager, this PR includes a IAssignmentHandler (AtlasAssignmentHandler for current Linux Consumption and LeigonAssignmentHandler for Legion). The LegionAssignmentHandler's methods are no-ops as of now because hosts in Legion will not be responsible for specializing msi, downloading, or mounting app files.
[Other PR](https://github.com/Azure/azure-functions-host/pull/8634)

resolves #8616 

### Pull request checklist

    * [x] Otherwise: Backport tracked by issue/PR #8616 

### Additional information

Additional PR information


src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
```diff
@@ -135,6 +135,32 @@ public static void AddWebJobsScriptHost(this IServiceCollection services, IConfi
             services.AddTransient<VirtualFileSystem>();
             services.AddTransient<VirtualFileSystemMiddleware>();
 
+            services.AddSingleton<IAssignmentHandler>(p =>
```
fabiocav, line 138: Please check the condition before the service registration and let DI create the instance and inject the services instead.

Example: https://github.com/Azure/azure-functions-host/blob/ec3c2d1049351870bbf611760520cf1e48075c8f/src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs#L145-L152

fabiocav (changes requested):
Had a few comments here but backed them out as my main question became more about the approach. 

Do we need to introduce a new interface here? With this changes, `IAssignmentHandler` and `IInstanceManager` are almost identical, and there's very little common logic left in the `InstanceManager` after the changes.

We should consider avoiding the new interface and nested services and have different implementations of the existing `IInstanceManager` interface for your scenarios (one for Atlas, one for Legion). Any common logic and specific methods required for this scenarios (i.e., `Download`, preferably `DownloadAsync`) could be defined by base abstract class shared by them (e.g., `LinuxInstanceManager`) if required.

src/WebJobs.Script.WebHost/Management/InstanceManager.cs
```diff
@@ -25,82 +16,25 @@ public class InstanceManager : IInstanceManager
...
         {
-            _client = httpClientFactory?.CreateClient() ?? throw new ArgumentNullException(nameof(httpClientFactory));
             _webHostEnvironment = webHostEnvironment ?? throw new ArgumentNullException(nameof(webHostEnvironment));
             _logger = logger ?? throw new ArgumentNullException(nameof(logger));
-            _metricsLogger = metricsLogger;
-            _meshServiceClient = meshServiceClient;
-            _runFromPackageHandler = runFromPackageHandler ?? throw new ArgumentNullException(nameof(runFromPackageHandler));
-            _packageDownloadHandler = packageDownloadHandler ?? throw new ArgumentNullException(nameof(packageDownloadHandler));
+            _meshServiceClient = meshServiceClient ?? throw new ArgumentNullException(nameof(meshServiceClient));
+            _assignmentHandler = assignmentHandler;
```
jviau, line 31: Verify arguments:

```suggestion
            _assignmentHandler = assignmentHandler ?? throw new ArgumentNullException(nameof(assignmentHandler));
```

src/WebJobs.Script.WebHost/Management/InstanceManager.cs
```diff
@@ -25,82 +16,25 @@ public class InstanceManager : IInstanceManager
...
         private readonly IEnvironment _environment;
-        private readonly IOptionsFactory<ScriptApplicationHostOptions> _optionsFactory;
-        private readonly HttpClient _client;
         private readonly IScriptWebHostEnvironment _webHostEnvironment;
+        private readonly IAssignmentHandler _assignmentHandler;
 
-        public InstanceManager(IOptionsFactory<ScriptApplicationHostOptions> optionsFactory, IHttpClientFactory httpClientFactory, IScriptWebHostEnvironment webHostEnvironment,
-            IEnvironment environment, ILogger<InstanceManager> logger, IMetricsLogger metricsLogger, IMeshServiceClient meshServiceClient, IRunFromPackageHandler runFromPackageHandler,
-            IPackageDownloadHandler packageDownloadHandler)
+        public InstanceManager(IScriptWebHostEnvironment webHostEnvironment,
```
jviau, line 24: ~This type is `public` - would this not be a breaking change for anyone external consuming this type? Or are we confident customers do not consume this type directly?~ I have been informed we don't expect customers to use any types in this assembly.

src/WebJobs.Script.WebHost/Management/LinuxSpecialization/IAssignmentHandler.cs
```diff
@@ -0,0 +1,19 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Script.WebHost.Models;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Management.LinuxSpecialization
+{
+    public interface IAssignmentHandler
+    {
+        Task<string> ValidateContext(HostAssignmentContext assignmentContext);
```
jviau, line 11: Do we follow `*Async` naming for task returning methods?

```suggestion
        Task<string> ValidateContextAsync(HostAssignmentContext assignmentContext);
```

src/WebJobs.Script.WebHost/Management/LinuxSpecialization/LegionAssignmentHandler.cs
```diff
@@ -0,0 +1,44 @@
...
+        private readonly ILogger<LegionAssignmentHandler> _logger;
+
+        public LegionAssignmentHandler(ILogger<LegionAssignmentHandler> logger)
+        {
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        }
+
+        public Task<string> ValidateContext(HostAssignmentContext assignmentContext)
+        {
+            _logger.LogInformation($"Skipping host assignment context (SiteId: {assignmentContext.SiteId}, SiteName: '{assignmentContext.SiteName}'. IsWarmup: '{assignmentContext.IsWarmupRequest}')");
```
jviau, line 22: nit: use structured logging.

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
```diff
@@ -135,6 +135,32 @@ public static void AddWebJobsScriptHost(this IServiceCollection services, IConfi
...
+            {
+                if (SystemEnvironment.Instance.IsLinuxConsumptionOnAtlas())
```
jviau, line 140: Are you able to perform this check outside of this lambda? Like so:

``` CSharp
if (SystemEnvironment.Instance.IsLinuxConsumptionOnAtlas())
{
    services.AddSingleton<IAssignmentHandler, AtlasAssignmentHandler>();
}
else if (SystemEnvironment.Instance.IsLinuxConsumptionOnLegion())
{
    services.AddSingleton<IAssignmentHandler, LegionAssignmentHandler>();
}
else
{
    // probably throw here? We are in a bad state if no assignment handler can be resolved I imagine.
}
```

src/WebJobs.Script.WebHost/Management/LinuxSpecialization/IAssignmentHandler.cs
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Script.WebHost.Models;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Management.LinuxSpecialization
+{
+    public interface IAssignmentHandler
```
jviau, line 9: While I like the direction you are going with reducing duplicate code, I am still conflicted regarding this abstraction layer. I _think_ what would be a better fit would be to derive a `InstanceManagerBase` class which has the common logic shared between the two, then appropriate `protected abstract` methods for the hooks where logic needs to be customized. The legion version can have almost all of those hooks be a no-op, and the atlas version can hold its current logic.

jviau (changes requested):
Consider the base class route - I think it will yield a more OOP result. Other than that and some minor comments, looking good for reducing code duplication!

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
fabiocav: Provided similar feedback above.

jviau: I see that now, my bad for missing it.

src/WebJobs.Script.WebHost/Management/LinuxSpecialization/IAssignmentHandler.cs
jviau: Realize now @fabiocav suggested essentially the same thing.

src/WebJobs.Script.WebHost/Management/LinuxSpecialization/IAssignmentHandler.cs
fabiocav: We do. Left the same comment, but backed out as it wouldn't be applicable if the recommendation is followed. I believe @Tonewall was just copying what was in `IInstanceManager`, which unfortunately, did not follow the convention.

src/WebJobs.Script.WebHost/Management/LinuxSpecialization/IAssignmentHandler.cs
Tonewall: @fabiocav @jviau I think your suggestion is a better approach since the InstanceManager class doesn't have much logic after introducing a new IAssignmentHandler interface. I have [another PR that utilizes a different InstanceManager](https://github.com/Azure/azure-functions-host/pull/8634) (one for Atlas and one for Legion) like you guys have suggested.
I was looking through the InstanceManager class, and I found that _GetInstanceInfo()_ and _Reset()_ (Internal method) are the only methods with the exact same implementation. So, I don't think an abstract class is necessary for this case. Should I just keep the changes in the [other PR](https://github.com/Azure/azure-functions-host/pull/8634) as is?

Thank you for your suggestions!

Tonewall:
Closing this PR for [this PR implementation](https://github.com/Azure/azure-functions-host/pull/8634)

PR #8808, opened 2022-09-30
[Linux Consumption]HostId generation should be cloud specific

### LINKED ISSUE #8788: [Linux Consumption] HostId generation logic should be cloud specific
https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Host/ScriptHostIdProvider.cs#L68

Code assumes the dns suffix will always be azurewebsites.net. Breaks on other clouds with different suffixes.

Exception Stack trace for Python Functions App + Blob Trigger + Linux Consumption running in fairfax - ICM https://portal.microsofticm.com/imp/v3/incidents/details/319380431/home

```
Azure.RequestFailedException : The specifed resource name contains invalid characters. RequestId:7ffb25bd-0003-0073-7d7c-ad4f51000000 Time:2022-08-11T12:18:30.7949544Z Status: 400 (The specifed resource name contains invalid characters.) ErrorCode: InvalidResourceName  Content: <?xml version='1.0' encoding='utf-8'?><Error><Code>InvalidResourceName</Code><Message>The specifed resource name contains invalid characters. RequestId:7ffb25bd-0003-0073-7d7c-ad4f51000000 Time:2022-08-11T12:18:30.7949544Z</Message></Error>  Headers: Server: Windows-Azure-Queue/1.0,Microsoft-HTTPAPI/2.0 x-ms-request-id: 7ffb25bd-0003-0073-7d7c-ad4f51000000 x-ms-version: 2018-11-09 x-ms-error-code: InvalidResourceName Date: Thu, 11 Aug 2022 12:18:30 GMT Content-Length: 243 Content-Type: application/xml
     at async Azure.Storage.Queues.QueueRestClient.GetPropertiesAsync(Nullable`1 timeout,CancellationToken cancellationToken)
    at async Azure.Storage.Queues.QueueClient.GetPropertiesInternal(Boolean async,CancellationToken cancellationToken,String operationName)
    at async Azure.Storage.Queues.QueueClient.ExistsInternal(Boolean async,CancellationToken cancellationToken)
    at async Azure.Storage.Queues.QueueClient.ExistsAsync(CancellationToken cancellationToken)
    at async Microsoft.Azure.WebJobs.Extensions.Storage.Common.Listeners.QueueListener.ExecuteAsync(CancellationToken cancellationToken)
    at async Microsoft.Azure.WebJobs.Extensions.Storage.Common.Timers.TaskSeriesTimer.RunAsync(CancellationToken cancellationToken)
```

balag0:
Logic doesnt account for non-Azure clouds resulting in failures across different scenarios.

resolves https://github.com/Azure/azure-functions-host/issues/8788

### Pull request checklist

* [ X] My changes **do not** require documentation changes
    * [X ] Otherwise: Backport tracked by issue/PR  (https://github.com/Azure/azure-functions-host/issues/8788)
* [ X] My changes **do not** require diagnostic events changes

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script/Environment/EnvironmentExtensions.cs
+++ b/src/WebJobs.Script/Environment/EnvironmentExtensions.cs
@@ -470,7 +470,7 @@ namespace Microsoft.Azure.WebJobs.Script
                 || string.IsNullOrEmpty(environment.GetEnvironmentVariable(MeshInitURI));
         }
 
-        public static CloudName GetCloudName(this IEnvironment environment)
+        public static CloudName GetCloudName(this IEnvironment environment, bool throwOnError = false)
         {
             var cloudName = environment.GetEnvironmentVariable(EnvironmentSettingNames.CloudName);
             if (Enum.TryParse(cloudName, true, out CloudName cloud))
@@ -478,6 +478,10 @@ namespace Microsoft.Azure.WebJobs.Script
                 return cloud;
             }
 
+            if (throwOnError)
+            {
+                throw new ArgumentException(nameof(GetCloudName), cloudName);
+            }
             return CloudName.Azure;
         }
 
@@ -499,6 +503,28 @@ namespace Microsoft.Azure.WebJobs.Script
                     return CloudConstants.USSecStorageSuffix;
                 default:
                     return CloudConstants.AzureStorageSuffix;
+            }
+        }
+
+        public static string GetDnsSuffixWithDotPrefix(this IEnvironment environment)
+        {
+            var cloudName = GetCloudName(environment, true);
+            switch (cloudName)
+            {
+                case CloudName.Azure:
+                    return CloudConstants.AzureDnsSuffixWithDotPrefix;
+                case CloudName.Blackforest:
+                    return CloudConstants.BlackforestDnsSuffixWithDotPrefix;
+                case CloudName.Fairfax:
+                    return CloudConstants.FairfaxDnsSuffixWithDotPrefix;
+                case CloudName.Mooncake:
+                    return CloudConstants.MooncakeDnsSuffixWithDotPrefix;
+                case CloudName.USNat:
+                    return CloudConstants.USNatDnsSuffixWithDotPrefix;
+                case CloudName.USSec:
+                    return CloudConstants.USSecDnsSuffixWithDotPrefix;
+                default:
+                    throw new ArgumentException(nameof(GetDnsSuffixWithDotPrefix), cloudName.ToString());
             }
         }
 
--- a/src/WebJobs.Script/Host/ScriptHostIdProvider.cs
+++ b/src/WebJobs.Script/Host/ScriptHostIdProvider.cs
@@ -65,7 +65,7 @@ namespace Microsoft.Azure.WebJobs.Script
             {
                 // The hostid is derived from the hostname for Linux consumption.
                 string hostName = environment.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebsiteHostName);
-                hostId = hostName?.Replace(".azurewebsites.net", string.Empty);
+                hostId = hostName?.Replace(environment.GetDnsSuffixWithDotPrefix(), string.Empty);
             }
             else
             {
--- a/src/WebJobs.Script/CloudConstants.cs
+++ b/src/WebJobs.Script/CloudConstants.cs
@@ -5,6 +5,8 @@ namespace Microsoft.Azure.WebJobs.Script
 {
     public static class CloudConstants
     {
+        public const string AzureCloudName = "Azure";
+
         public const string AzureStorageSuffix = "core.windows.net";
         public const string BlackforestStorageSuffix = "core.cloudapi.de";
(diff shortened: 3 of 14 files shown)


src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -470,14 +470,18 @@ public static bool IsMountDisabled(this IEnvironment environment)
...
         {
             var cloudName = environment.GetEnvironmentVariable(EnvironmentSettingNames.CloudName);
             if (Enum.TryParse(cloudName, true, out CloudName cloud))
             {
                 return cloud;
             }
 
+            if (throwOnError)
+            {
+                throw new ArgumentException(nameof(GetCloudName), cloudName);
```
fabiocav, line 483: Consider a different exception type that better conveys the problem. The usage of `ArgumentException` here is not passing the expected parameters

src/WebJobs.Script/Host/ScriptHostIdProvider.cs
```diff
@@ -65,7 +65,7 @@ internal static HostIdResult GetDefaultHostId(IEnvironment environment, ScriptAp
             {
                 // The hostid is derived from the hostname for Linux consumption.
                 string hostName = environment.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebsiteHostName);
-                hostId = hostName?.Replace(".azurewebsites.net", string.Empty);
+                hostId = hostName?.Replace(environment.GetDnsSuffixWithDotPrefix(), string.Empty);
```
fabiocav, line 68: For existing clouds with the new handling, this will cause host IDs to change and break a number of features for those sites. Everything the host scopes, including HTTP API keys, would be invalidated.

fabiocav (changes requested):

src/WebJobs.Script/Host/ScriptHostIdProvider.cs
mathewc, line 68: Is it because the value we need to extract may itself contain periods that we can't take a parsing approach here (e.g. everything up to the first period) rather than a list of cloud suffixes that we'll have to ensure stays up to date over time?

pragnagopa: @balag0 - as this is a known issue with blob triggers apps on linux consumption - lets scope this to Linux Consumption ? Also, do you know if this works on Linux Dedicated Sku?

Another consideration - provide opt-in flag for Cx to ensure we do not break any existing sceanarios that are currently running in Linux Consumption - non-public cloud

mathewc: Also, once an app has ran and established/persisted an ID value, changing it later is very problematic. Won't these changes result in a changing ID for some apps at this point?

mathewc: One more point - there's code in the ScaleController (GetHostIdFromSiteName) that computes the ID and that logic needs to be kept in sync with this logic. That logic doesn't do any of this suffix parsing - it's just using the site name.

balag0: Removing the change to HostNameProvider would scope it to just Linux consumption.
It doesnt break on other skus.

@fabiocav mentioned making this change as part of vNext which I think also works since things have been broken for a while.

balag0: Yes, would need some discussion on how to roll out this change to reduce breaks. Current options are appsetting to opt in to this behavior or make this change part of vnext

Regarding SC, it seems like the logic there is already out of sync with what we do here (both before and after this PR)
So I assume it is already broken there and we dont have to worry about that getting out of sync with host?

pragnagopa: @chiangvincent  - please open tracking item to keep SC logic that looks at HostId in sync with the host

> One more point - there's code in the ScaleController (GetHostIdFromSiteName) that computes the ID and that logic needs to be kept in sync with this logic. That logic doesn't do any of this suffix parsing - it's just using the site name.

pragnagopa: @balag0 / @mathewc - thinking more about this - we have another option 

- Provide opt-in flag for existing Cx
- Default the flag to true for new applications only through control plane changes in Antares

This will help make the change now without breaking existing apps but also will help address the broken scenarios without waiting for vNext. 

thoughts?

fabiocav: Another option, and perhaps the safest, would be a fix in the storage layer to deal with the host IDs that have invalid names. Even the second option about having the flag defaulted to true could impact customer flows where they delete and recreate resources through automation.

pragnagopa: Yes you are right - recreated apps option 2 would not work. So, Opt-in flag for now and long term updates to storage layer?

PR #8815, opened 2022-10-04
Support model-type bindings for out-of-proc workers

liliankasem:
Add support for model-type bindings for out-of-proc workers
- Epic: https://github.com/Azure/azure-functions-dotnet-worker/issues/1081
- Related issue: https://github.com/Azure/azure-functions-dotnet-worker/issues/1035

Part of the work to support SDK-type bindings for out-of-proc workers is for the host to bind to `ParameterBindingData` if a function supports deferred binding. Deferred binding means that instead of resolving binding payloads host-side, it will be deferred so that the worker resolves binding when it gets the invocation request. For example, if a function binds to a BlobClient, instead of creating and passing around a BlobClient, we will use `ParameterBindingData` to gather all the relevant information the worker would need to create the BlobClient itself (worker-side).

Things happening in this PR:

- Support binding to ParameterBindingData. If a function indicates that it supports deferred binding, we can use `ParameterBindingData` to get the relevant information from the extension. By setting this as a functions defaultType, we can check extension converters to see if they can provide `ParameterBindingData` which enables us to bind functions to SDK-types on the worker side.

- A limitation here is that we only support `ParameterBindingData` and deferred binding for trigger bindings **if** none of the function's input or output bindings contain an expression (you can read more about why this limitation exists in this [issue](https://github.com/Azure/azure-functions-dotnet-worker/issues/1080)). Because of this, we first check if the trigger binding "SupportsDeferredBinding". If it does, we check all the attributes for the input and output bindings to see if there are any expressions being used. If an expressions is being used, we set the "SkipDeferredBinding" flag in the function's metadata `Properties`. We use `SkipDeferredBinding` later down the line to skip deferred binding for the trigger, meaning we set `SupportsDeferredBinding` to false. So instead of using ParameterBindingData, we use the defaultType that is evaluated by the WebJobs sdk.

- Update ScriptBindingContext to include `Properties` prop and a `SupportsDeferredBinding` prop. In the protobuf, the [Binding info was updated](https://github.com/Azure/azure-functions-language-worker-protobuf/blob/27a28ca180e6a34b8aea3349e3ffe46d2ff36ce7/src/proto/FunctionRpc.proto#L524-L525) to include a `Properties` map. This map is used by workers to add additional information about a function binding. For this specific feature, the worker will add a `SupportsDeferredBinding` prop to the `Properties` map to indicate to the host that a given function supports deferred binding.

- Convert ParameterBindingData to ModelBindingData so that we can send this information of gRPC to workers. `ModelBindingData` is the gRPC message representation of the WebJobs `ParameterBindingData`; you can find the protobuf definition for this [here](https://github.com/Azure/azure-functions-language-worker-protobuf/blob/27a28ca180e6a34b8aea3349e3ffe46d2ff36ce7/src/proto/FunctionRpc.proto#L656-L671)

The https://github.com/Azure/azure-functions-dotnet-worker/issues/1081#issuecomment-1276566098 has a list of all the PRs related to this feature


src/WebJobs.Script.Abstractions/Description/Binding/DataType.cs
```diff
@@ -11,6 +11,7 @@ public enum DataType
         Undefined,
         String,
         Binary,
-        Stream
+        Stream,
+        Reference
```
fabiocav: Since we renamed the type, should we also rename this to `BindingData`?

src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs
```diff
@@ -62,6 +65,24 @@ public static ValueTask<TypedData> ToRpc(this object value, ILogger logger, Grpc
...
         }
 
+        internal static TypedData ToBindingData(this ParameterBindingData data, ILogger logger)
+        {
+            var bindingData = new BindingData
+            {
+                Version = data.Version,
+                ContentType = data.ContentType,
+                Source = data.Source,
+                Content = data.Content.ToString()
```
fabiocav: Do you expect the conversion here to change once we update the content to be `BinaryData`? Are we going to keep the proto definition as a string?

src/WebJobs.Script/Binding/FunctionBinding.cs
```diff
@@ -221,6 +222,12 @@ internal static async Task BindStringAsync(BindingContext context)
             context.Value = str;
         }
 
+        internal static async Task BindParameterBindingData(BindingContext context)
+        {
+            var parameterBindingData = await context.Binder.BindAsync<ParameterBindingData>(context.Attributes);
+            context.Value = parameterBindingData;
+        }
```
fabiocav:
```suggestion
        internal static Task BindParameterBindingData(BindingContext context)
        {
            return context.Binder.BindAsync<ParameterBindingData>(context.Attributes);
        }
```

src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs
liliankasem: We can do string or bytes - there aren't any other types that fit.

If we use bytes, we can do this `ByteString.CopyFrom(data.Content)`

And on the worker side to get this back to BinaryData we would need to do something like `BinaryData IBindingData.Content => BinaryData.FromBytes(Content.ToByteArray());`

With string we do `BinaryData IBindingData.Content => BinaryData.FromString(Content);`

It feels like an unnecessary amount of conversions to me, but bytes are supposed to be more efficient

src/WebJobs.Script/Binding/FunctionBinding.cs
liliankasem: This breaks the workflow, `context.Value` needs to be set

src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs
liliankasem: @fabiocav Currently using bytes, let me know what you think

src/WebJobs.Script/WebJobs.Script.csproj
```diff
@@ -79,6 +78,9 @@
     <PackageReference Include="System.Reactive.Linq" Version="5.0.0" />
     <PackageReference Include="System.Runtime.Loader" Version="4.3.0" />
   </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\WebJobs.Script.Abstractions\WebJobs.Script.Abstractions.csproj" />
```
liliankasem: Need to revert this change once a new version of `Microsoft.Azure.WebJobs.Script.Abstractions` is published

src/WebJobs.Script.WebHost/WebJobs.Script.WebHost.csproj
```diff
@@ -75,7 +75,7 @@
     <PackageReference Include="Microsoft.Azure.Cosmos.Table" Version="1.0.8" />
     <PackageReference Include="Microsoft.Azure.Storage.File" Version="11.1.7" />
 
-    <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.34-11938" />
+    <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.34-dev638006900649255460" />
```
liliankasem: Will update once the new SDK is published

src/WebJobs.Script.Abstractions/Description/Binding/DataType.cs
```diff
@@ -11,6 +11,7 @@ public enum DataType
         Undefined,
         String,
         Binary,
-        Stream
+        Stream,
+        BindingData
     }
 }
```
liliankasem: Do we need to publish a new Abstractions package

src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs
```diff
@@ -53,6 +55,7 @@ public static ValueTask<TypedData> ToRpc(this object value, ILogger logger, Grpc
                     JObject jobj => new TypedData() { Json = jobj.ToString(Formatting.None) },
                     string str => new TypedData() { String = str },
                     double dbl => new TypedData() { Double = dbl },
+                    ParameterBindingData bindingData => bindingData.ToBindingData(logger),
```
surgupta-msft: We are not using logger in the method implementation. We can remove this if not needed.

src/WebJobs.Script/Binding/ExtensionBinding.cs
```diff
@@ -111,6 +112,10 @@ public override async Task BindAsync(BindingContext context)
             {
                 await BindStringAsync(context);
             }
+            else if (_binding.DefaultType == typeof(ParameterBindingData))
+            {
+                await BindParameterBindingData(context);
```
surgupta-msft: We can add Async suffix in method name.

src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs
```diff
@@ -62,6 +65,24 @@ public static ValueTask<TypedData> ToRpc(this object value, ILogger logger, Grpc
             }
         }
 
+        internal static TypedData ToBindingData(this ParameterBindingData data, ILogger logger)
```
surgupta-msft: We can add tests for this in `GrpcMessageConversionExtensionsTests.cs`

src/WebJobs.Script.Abstractions/Description/Binding/DataType.cs
liliankasem: Yes, moving this to a new PR

src/WebJobs.Script.WebHost/WebJobs.Script.WebHost.csproj
```diff
@@ -75,7 +75,7 @@
     <PackageReference Include="Microsoft.Azure.Cosmos.Table" Version="1.0.8" />
     <PackageReference Include="Microsoft.Azure.Storage.File" Version="11.1.7" />
 
-    <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.34-11938" />
+    <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.1.0-dev638022334275693870" />
```
liliankasem: Pending WebJobs release

src/WebJobs.Script.Grpc/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
```diff
@@ -32,7 +32,7 @@ message StreamingMessage {
     WorkerInitRequest worker_init_request = 17;
```
liliankasem: Remove this change and create separate proto PR once we finalize the protobuf

src/WebJobs.Script/Description/FunctionDescriptorProvider.cs
```diff
@@ -74,6 +82,36 @@ public virtual async Task<(bool, FunctionDescriptor)> TryCreate(FunctionMetadata
             }
         }
 
+        private bool FunctionBindingContainsExpression(Collection<FunctionBinding> bindings)
```
liliankasem: `BindingAttributeContainsExpression` is probably more fitting

src/WebJobs.Script/Description/FunctionDescriptorProvider.cs
```diff
@@ -110,8 +139,14 @@ public void VerifyResolvedBindings(FunctionMetadata functionMetadata, IEnumerabl
 
             ApplyMethodLevelAttributes(functionMetadata, triggerMetadata, methodAttributes);
 
+            Type parameterType = null;
+            if (BindingAttributeContainsExpression(inputBindings.Union(outputBindings)))
+            {
+                parameterType = typeof(ParameterBindingData);
```
liliankasem: since we're setting `parameterType` here, we don't ever go through the GetRequestedType path

src/WebJobs.Script/Description/FunctionDescriptorProvider.cs
```diff
@@ -74,6 +74,35 @@ public virtual async Task<(bool, FunctionDescriptor)> TryCreate(FunctionMetadata
...
+            {
+                if (binding.Metadata.IsTrigger)
+                {
+                    // skip triggers, we only care if input bindings contain expressions
+                    continue;
+                }
+
+                foreach (var attribute in binding.Attributes)
+                {
+                    if (bindingExpressionRegex.IsMatch(attribute.ToString()))
```
liliankasem: this doesn't work, need to figure out how to get `BlobPath` from this attribute. We do know the type via TypeId

attribute.ToString() -> "Microsoft.Azure.WebJobs.BlobAttribute"

liliankasem: Actually I just found GetAttributeData - I think this will solve the issue

src/WebJobs.Script/Description/FunctionDescriptorProvider.cs
```diff
@@ -110,8 +142,14 @@ public void VerifyResolvedBindings(FunctionMetadata functionMetadata, IEnumerabl
 
             ApplyMethodLevelAttributes(functionMetadata, triggerMetadata, methodAttributes);
 
+            Type parameterType = null;
+            if (BindingAttributeContainsExpression(inputBindings.Union(outputBindings)))
+            {
+                parameterType = typeof(object); // what do we set this to? we still need to evaluate DataType
```
liliankasem: Here we're saying "if an input or output binding contains an expression, set `parameterType` so we can skip the `GetRequestedType` flow"

But what do we set the parameterType to? We still need to evaluate DataType to figure that out. So would it be better to manipulate the the `SupportsDeferredBinding` flag instead?

src/WebJobs.Script.Abstractions/Description/Binding/BindingMetadata.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
```
liliankasem: Abstractions project changes will be their own PR

src/WebJobs.Script/Description/Workers/WorkerFunctionDescriptorProvider.cs
```diff
@@ -52,10 +54,20 @@ protected override IFunctionInvoker CreateFunctionInvoker(string scriptFilePath,
...
             BindingMetadata triggerMetadata, Collection<CustomAttributeBuilder> methodAttributes, Collection<FunctionBinding> inputBindings, Collection<FunctionBinding> outputBindings)
         {
-            var parameters = await base.GetFunctionParametersAsync(functionInvoker, functionMetadata, triggerMetadata, methodAttributes, inputBindings, outputBindings);
-
             var bindings = inputBindings.Union(outputBindings);
 
+            // If an input or output binding uses expressions, we do not want to use ParameterBindingData for the trigger bindings
+            if (triggerMetadata.SupportsDeferredBinding())
+            {
+                bool skipDeferBinding = BindingAttributeContainsExpression(bindings);
```
mathewc: nit: skipDeferredBinding - to match the naming elsewhere

src/WebJobs.Script/Description/Workers/WorkerFunctionDescriptorProvider.cs
```diff
@@ -52,10 +54,20 @@ protected override IFunctionInvoker CreateFunctionInvoker(string scriptFilePath,
...
+                if (skipDeferBinding)
+                {
+                    triggerMetadata.Properties.Add(BindingMetadataExtensions.SkipDeferredBindingKey, true);
```
mathewc: So we're already mutating the BindingMetadata here by adding this new property. I wonder if it'd be any worse to just remove the "supports" property from the collection in this case, that way we don't need the complexity of 2 different flags?

src/WebJobs.Script/Description/Workers/WorkerFunctionDescriptorProvider.cs
```diff
@@ -90,5 +102,37 @@ protected override IFunctionInvoker CreateFunctionInvoker(string scriptFilePath,
...
         }
+
+        private bool BindingAttributeContainsExpression(IEnumerable<FunctionBinding> bindings)
+        {
+            if (bindings.Count() <= 0)
+            {
+                return false;
+            }
+
+            var bindingExpressionRegex = new Regex(@"{(.*?)\}");
```
mathewc: Rather than new this up each time, we should declare it as a field, and compile it, e.g. https://github.com/Azure/azure-functions-host/blob/852eed9ceef6ef56b431428a8eb31f1bd9c97f3b/src/WebJobs.Script/Description/DotNet/LocalSharedAssemblyProvider.cs#L28

src/WebJobs.Script/Description/Workers/WorkerFunctionDescriptorProvider.cs
```diff
@@ -90,5 +102,37 @@ protected override IFunctionInvoker CreateFunctionInvoker(string scriptFilePath,
...
+            {
+                if (binding.Metadata.IsTrigger)
+                {
+                    // skip triggers, we only care if input and output bindings contain expressions
+                    continue;
+                }
+
+                foreach (var attribute in binding.Attributes)
+                {
+                    foreach (var property in attribute.GetType().GetProperties())
```
mathewc: Should use the same binding flags that AttributeCloner uses for binding attributes https://github.com/Azure/azure-webjobs-sdk/blob/3f4ec78be9f43bb041937425ced00b341883aa42/src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs#L42. We should also filter these properties to just string properties, rather than doing ToString below. We wouldn't want to run this on any non-string properties.

src/WebJobs.Script/Description/Workers/WorkerFunctionDescriptorProvider.cs
```diff
@@ -90,5 +102,37 @@ protected override IFunctionInvoker CreateFunctionInvoker(string scriptFilePath,
 
             return parameters;
         }
+
+        private bool BindingAttributeContainsExpression(IEnumerable<FunctionBinding> bindings)
+        {
+            if (bindings.Count() <= 0)
```
mathewc: can remove this check once you make the regex a field

src/WebJobs.Script.Abstractions/Description/Binding/BindingMetadata.cs
```diff
@@ -81,6 +87,7 @@ public static BindingMetadata Create(JObject raw)
             string bindingDirectionValue = (string)raw["direction"];
             string connection = (string)raw["connection"];
             string bindingType = (string)raw["type"];
+            var properties = raw["properties"].ToObject<IDictionary<string, object>>();
```
mathewc: Wondering about casing on the lookups we're doing in this dictionary. Can we make this case insensitive? Though I see the FunctionMetadata Properties dictionary is also declared case sensitive which was an oversight.

src/WebJobs.Script/Description/Workers/WorkerFunctionDescriptorProvider.cs
```diff
@@ -90,5 +102,37 @@ protected override IFunctionInvoker CreateFunctionInvoker(string scriptFilePath,
 
             return parameters;
         }
+
+        private bool BindingAttributeContainsExpression(IEnumerable<FunctionBinding> bindings)
```
mathewc: You'll want to add tests for this helper function

src/WebJobs.Script/Description/Workers/WorkerFunctionDescriptorProvider.cs
liliankasem: We could just remove the supports property, but then we're only removing it from this instance of triggerMetadata- we have things like the script context and other instances of the inputBindings where we are not removing the "supports" flag from "Properties". I guess the same can be argued here since this will be the only binding instance where we add the "skip" flag 🤔 Something about removing the original data is iffy to me, but the original is always available in .Raw anyway

src/WebJobs.Script/Description/Workers/WorkerFunctionDescriptorProvider.cs
liliankasem: I'm not sure I follow; what does the regex field have to do with checking the binding.Count?

src/WebJobs.Script/Description/Workers/WorkerFunctionDescriptorProvider.cs
liliankasem: I keep ended up with a object array, so this is what I managed to get working but still seems heavy? Open to further suggestions


```csharp
var properties = attribute.GetType()
                    .GetProperties(flags)
                    .Where(prop => prop.PropertyType == typeof(string))
                    .Select(p => p.GetValue(attribute))
                    .ToArray();

foreach (var property in properties)
{
    if (_expressionRegex.IsMatch((string)property))
```

src/WebJobs.Script.Abstractions/Description/Binding/BindingMetadata.cs
```diff
@@ -81,6 +87,10 @@ public static BindingMetadata Create(JObject raw)
             string bindingDirectionValue = (string)raw["direction"];
             string connection = (string)raw["connection"];
             string bindingType = (string)raw["type"];
+            var properties = raw["properties"] != null
+                                ? raw["properties"].ToObject<IDictionary<string, object>>()
+                                : new Dictionary<string, object>();
```
fabiocav: Use `ImmutableDictionary<string, object>.Empty` instead of `new Dictionary<string, object>();`

src/WebJobs.Script.Grpc/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
```diff
@@ -647,3 +652,20 @@ message RpcHttp {
...
+// Used for hydrating SDK-type bindings in out-of-proc workers
+message ModelBindingData
+{
+    // The version of the binding data content
+    string version = 1;
+
+    // The extension source of the binding data
+    string source = 2;
+
+    // The binding data content
```
fabiocav: nit: the comments for content and content_type seem to be swapped.

src/WebJobs.Script/Binding/Extensibility/ScriptBindingContext.cs
```diff
@@ -116,7 +129,7 @@ public TValue GetMetadataValue<TValue>(string name, TValue defaultValue = defaul
             {
                 if (Metadata.TryGetValue(name, StringComparison.OrdinalIgnoreCase, out JToken value))
                 {
-                    return value.Value<TValue>();
+                    return value.ToObject<TValue>();
```
fabiocav: Was this failing with the dictionary?

src/WebJobs.Script/Binding/GeneralScriptBindingProvider.cs
```diff
@@ -61,7 +62,16 @@ public override bool TryResolveAssembly(string assemblyName, AssemblyLoadContext
         // Read the properties and convert that into a System.Type.
         internal static Type GetRequestedType(ScriptBindingContext context)
         {
-            Type type = ParseDataType(context);
+            Type type;
```
fabiocav: nit: can we make this a bit more succinct with a ternary conditional here?

src/WebJobs.Script/Binding/Extensibility/ScriptBindingContext.cs
liliankasem: Yup it was failing, but since I'm not using this method for the dictionary anymore, I'm going to reverse this change

src/WebJobs.Script.Abstractions/Description/Binding/BindingMetadata.cs
liliankasem: I get an error when I try this, doesn't seem happy with the expression 🤔

```
Type of conditional expression cannot be determined because there is no implicit conversion between 'System.Collections.Generic.Dictionary<string, object>' and 'System.Collections.Immutable.ImmutableDictionary<string, object>' 
```

src/WebJobs.Script/Description/Workers/WorkerFunctionDescriptorProvider.cs
```diff
@@ -90,5 +104,43 @@ protected override IFunctionInvoker CreateFunctionInvoker(string scriptFilePath,
...
+
+            foreach (ExtensionBinding binding in bindings)
+            {
+                if (binding.Metadata.IsTrigger)
+                {
+                    // skip triggers, we only care if input and output bindings contain expressions
+                    continue;
+                }
+
+                foreach (var attribute in binding.Attributes)
```
fabiocav, line 128: Got sidetracked and forgot to send this. But we should be able to remove the `foreach` below and replace the where/select/toarray calls with one `Any` with the predicate and Regex evaluation. That should reduce a lot of the work being done here.

src/WebJobs.Script/Description/Workers/WorkerFunctionDescriptorProvider.cs
liliankasem: Narrowed it down to this:

```
return attribute.GetType()
    .GetProperties(flags)
    .Any(prop => prop.PropertyType == typeof(string) && _expressionRegex.IsMatch((string)prop.GetValue(attribute)));
```

src/WebJobs.Script/Description/Workers/WorkerFunctionDescriptorProvider.cs
mathewc: I was just pointing out that you don't need to do a .Count() here at all - the below foreach will only execute if there are actually any bindings. As it stands, you're doing an unnecessary double enumeration. I had mentioned the regex because previously you were initializing it here before the foreach, and I thought you were trying to avoid that instantiation.

src/WebJobs.Script/Description/Workers/WorkerFunctionDescriptorProvider.cs
```diff
@@ -90,5 +104,33 @@ protected override IFunctionInvoker CreateFunctionInvoker(string scriptFilePath,
 
             return parameters;
         }
+
+        internal bool BindingAttributeContainsExpression(IEnumerable<FunctionBinding> bindings)
+        {
+            BindingFlags flags = BindingFlags.Instance | BindingFlags.Public;
```
mathewc: Just inline this below where it's used

src/WebJobs.Script/Description/Workers/WorkerFunctionDescriptorProvider.cs
liliankasem: Ahhh gotcha, that makes sense - thanks!

src/WebJobs.Script/Extensions/BindingMetadataExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Azure.WebJobs.Script.Description;
+
+namespace Microsoft.Azure.WebJobs.Script.Extensions
+{
+    public static class BindingMetadataExtensions
+    {
+        internal const string SupportsDeferredBindingKey = "supportsDeferredBinding";
```
mathewc: nit: Use Pascal casing for these property values

src/WebJobs.Script/Description/Workers/WorkerFunctionDescriptorProvider.cs
mathewc: What you have is OK. The two flags are a little confusing, and I was trying to see if we could improve, but I don't have a solid proposal.

src/WebJobs.Script/Description/Workers/WorkerFunctionDescriptorProvider.cs
```diff
@@ -100,5 +114,26 @@ protected override IFunctionInvoker CreateFunctionInvoker(string scriptFilePath,
 
             return parameters;
         }
+
+        internal bool BindingAttributeContainsExpression(IEnumerable<FunctionBinding> bindings)
+        {
+            foreach (ExtensionBinding binding in bindings)
```
mathewc, line 120: Any reason why you're using the derived ExtensionBinding type here rather than FunctionBinding? I think you should just use FunctionBinding, since there can be other subclasses, e.g. HttpBinding

liliankasem: no reason, probably a typo! Fixed

liliankasem: Ah I lied, there was a reason that I forgot about! 'FunctionBinding' does not contain a definition for 'Attributes' so I had to use ExtensionBinding here

mathewc (approved):

surgupta-msft (approved):

PR #8821, opened 2022-10-05
[pack] Increasing the perf of extensions

shrohilla:

resolves #issue_for_this_pr

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script/Binding/FunctionBinding.cs
+++ b/src/WebJobs.Script/Binding/FunctionBinding.cs
@@ -147,6 +147,7 @@ namespace Microsoft.Azure.WebJobs.Script.Binding
             IEnumerable values = ReadAsEnumerable(context.Value);
 
             // convert values as necessary and add to the collector
+            List<Task> tasks = new List<Task>();
             foreach (var value in values)
             {
                 object converted = null;
@@ -196,8 +197,10 @@ namespace Microsoft.Azure.WebJobs.Script.Binding
                     throw new ArgumentException("Unsupported collection type.");
                 }
 
-                await collector.AddAsync((T)converted);
+                tasks.Add(collector.AddAsync((T)converted));
             }
+
+            await Task.WhenAll(tasks);
         }
 
         internal static async Task BindJTokenAsync<T>(BindingContext context, FileAccess access)


fabiocav:
@shrohilla I realize this is still in draft, but the title made me curious :)

To better describe the changes here, it would be good to remove Kafka from the description as this is not Kafka specific.

Have you done any measurements here?

shrohilla:
> @shrohilla I realize this is still in draft, but the title made me curious :)
> 
> To better describe the changes here, it would be good to remove Kafka from the description as this is not Kafka specific.
> 
> Have you done any measurements here?

@fabiocav for Kafka out of proc apps it worked well, while no change for perf for eventhub, but I only tested for kafka and Eventhub that too for python app.

src/WebJobs.Script/Binding/FunctionBinding.cs
```diff
@@ -196,8 +197,10 @@ internal static async Task BindAsyncCollectorAsync<T>(BindingContext context)
                     throw new ArgumentException("Unsupported collection type.");
                 }
 
-                await collector.AddAsync((T)converted);
+                tasks.Add(collector.AddAsync((T)converted));
```
mathewc, line 200: I think parallelizing this will cause problems for some existing IAsyncCollector implementations. For example, look at the CosmosDB collector [here](https://github.com/Azure/azure-webjobs-sdk-extensions/blob/a0fb2011acfa88126d9eecc285d7a7f8283ccbe5/src/WebJobs.Extensions.CosmosDB/Bindings/CosmosDBAsyncCollector.cs#L28). It's doing an optimistic upsert, and if the DB or collection doesn't exist, it'll handle the error, create the collection and try the upsert again. If all the Adds were parallelized, they will likely all fail into the Create DB/Collection case and result in a bunch of retries/failures. @ealsur what are your thoughts here? Likely there are other existing collectors that will have similar issues.

I think rather than making this change, extensions should just implement IAsyncCollector.Flush batch support, so AddAsync just accumulates, and when Flush is invoked by the binding pipeline it can perform efficient batched service operations.

ealsur: Also, if the customer wants parallelization, they can parallelize the AddAsync on the IAsyncCollector, doing the same WhenAll pattern.

shrohilla:
No longer required synced offline with @mathewc and found the solution in kafka extension

PR #8828, opened 2022-10-10
Track ApplicationInsights ingestion service response by subscribing t…

### LINKED ISSUE #2649: Detect if Application Insights is not able to connect to the Telemetry and Live Metrics Endpoints.
Customers run their Functions in a number of environments.
In some of these environments like Isolated and\or Functions Premium it is possible that they may have configured the Virtual Networks in such a way as to block access to the endpoints.
E.G DNS, Route Tables, NSG's etc.

https://docs.microsoft.com/en-us/azure/azure-monitor/app/ip-addresses 

The customer may also choose to block ingestion from the public internet.

https://docs.microsoft.com/en-us/azure/azure-monitor/platform/private-link-security#configure-application-insights

#### Repro steps

Provide the steps required to reproduce the problem

1. Configure a New Function App and enable Application Insights. 

2. Turn off Public Network Access ingestion on the Application Insights resource. 

image

#### Expected behavior

That the Functions runtime would let you know when it is unable to send data to Application Insights. 

#### Actual behavior

No monitor data is available and there's no way to know why.

#### Known workarounds

No workarounds exist.

RohitRanjanMS:
Track ApplicationInsights ingestion service response by subscribing to transmission event. It will only log in case of non-sucess response or if the debug trace feature flag (EnableDebugTracing) is enabled, this will log all the request.

e.g. 
```sh
Diagnostic source 'Microsoft.Azure.Functions.Host.ApplicationInsights' emitted event 'TransmissionStatus': {"statusCode":400,"statusDescription":"Invalid instrumentation key","id":"vkZu1wA1u3c=","items":[{"type":"AppTraces","count":3},{"type":"AppRequests","count":1},{"type":"AppMetrics","count":12}],"responseTimeInMs":65,"timeout":"00:01:40"}
```

resolves [this](https://github.com/Azure/azure-webjobs-sdk/issues/2649) issue.

### Pull request checklist

* [x] have added all required tests (Unit tests, E2E tests)


src/WebJobs.Script/Config/TransmissionStatusTelemetryModule.cs
```diff
@@ -0,0 +1,131 @@
...
+        /// </summary>
+        /// <param name="configuration">Telemetry configuration to use for initialization.</param>
+        public void Initialize(TelemetryConfiguration configuration)
+        {
+            // Prevent the telemetry module from being initialized multiple times.
+            if (isInitialized)
+            {
+                return;
+            }
+            if (configuration.TelemetryChannel is ServerTelemetryChannel)
```
kshyju: nit: pattern matching approach can make this even more concise.
````
if (configuration.TelemetryChannel is ServerTelemetryChannel channel)
{
   // use channel now.
}
````

File: src/WebJobs.Script/Config/TransmissionStatusTelemetryModule.cs [hunk 2]
```diff
@@ -0,0 +1,131 @@
...
+
+namespace Microsoft.Azure.WebJobs.Script.Config
+{
+    /// <summary>
+    /// Transmission Status Telemetry Module
+    /// </summary>
+    internal class TransmissionStatusTelemetryModule : ITelemetryModule, IDisposable
+    {
+        private bool isInitialized = false;
+        private DiagnosticListener source = new DiagnosticListener(ScriptConstants.HostDiagnosticSourcePrefix + "ApplicationInsights");
```
kshyju: Use the `_`  prefix to be consistent with the rest of the code base.

```suggestion
        private readonly DiagnosticListener _source
```

kshyju: Same comment for other private variables.

src/WebJobs.Script/Config/TransmissionStatusTelemetryModule.cs
```diff
@@ -0,0 +1,131 @@
...
+                        .Select(n => new
+                        {
+                            type = n.Key,
+                            count = n.Count()
+                        });
+
+            BackendResponse backendResponse = null;
+            if (!string.IsNullOrWhiteSpace(args?.Response?.Content))
+            {
+                backendResponse = JsonConvert.DeserializeObject<BackendResponse>(args.Response.Content, options);
```
kshyju: I see that `System.Text.Json.JsonSerializer` is available to use in this project and our target TFM is net6.0. Should we consider using the Deserialize method from that here? S.T.J will use source generation feature to generate the type information and we can use[ this overload](https://docs.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializer.deserialize?view=net-6.0#system-text-json-jsonserializer-deserialize-1(system-string-system-text-json-serialization-metadata-jsontypeinfo((-0)))) for deserialization. That should perform better than the Deserialize (from Newtonsoft.Json) we currently have in this PR.

https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/source-generation?pivots=dotnet-6-0

src/WebJobs.Script/Config/TransmissionStatusTelemetryModule.cs
```diff
@@ -0,0 +1,131 @@
...
+                errorCode = topStatusCode,
+                responseTimeInMs = args?.ResponseDurationInMs,
+                retryAfterHeader = args?.Response?.RetryAfterHeader,
+                timeout = transmission?.Timeout,
+            };
+            return JsonConvert.SerializeObject(log, options);
+        }
+    }
+
+    internal class BackendResponse
```
kshyju: Consider improving the naming to be a bit more specific/meaningful/readable.

src/WebJobs.Script/Config/TransmissionStatusTelemetryModule.cs
kshyju: Same comment applies for the other Serialize method call at 107

kshyju: The change would be like this.
```
// Update the options as needed.
[JsonSourceGenerationOptions(WriteIndented = true)]
[JsonSerializable(typeof(BackendResponse))]
internal partial class SourceGenerationContext : JsonSerializerContext
{
}
```
and
```
var backendResponse = System.Text.Json.JsonSerializer.Deserialize(
                    args.Response.Content, SourceGenerationContext.Default.BackendResponse);
```

src/WebJobs.Script/Config/TransmissionStatusTelemetryModule.cs
```diff
@@ -0,0 +1,131 @@
...
+            if (!string.IsNullOrWhiteSpace(args?.Response?.Content))
+            {
+                backendResponse = JsonConvert.DeserializeObject<BackendResponse>(args.Response.Content, options);
+            }
+
+            string topErrorMessage = null;
+            int? topStatusCode = null;
+            if (backendResponse?.Errors != null && backendResponse.Errors.Length > 0)
+            {
+                topErrorMessage = backendResponse.Errors[0].Message;
```
kshyju: We only log the first error message?. Would logging all the error messages be useful? May be build a comma separated string?

src/WebJobs.Script/Config/TransmissionStatusTelemetryModule.cs
```diff
@@ -0,0 +1,131 @@
...
+using Microsoft.ApplicationInsights.Extensibility.Implementation;
+using Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.WebJobs.Script.Config
+{
+    /// <summary>
+    /// Transmission Status Telemetry Module
+    /// </summary>
+    internal class TransmissionStatusTelemetryModule : ITelemetryModule, IDisposable
```
mathewc: Do we need a new module for this? Can't we just subscribe to the event on TelemetryConfiguration.TelemetryChannel in the config action we're already specifying [here](https://github.com/Azure/azure-functions-host/blob/54941dba8cece0c93744c9bc6a23ea469d0873b7/src/WebJobs.Script/ScriptHostBuilderExtensions.cs#L390)?

RohitRanjanMS: yes, we can subscribe to the event. I thought telemetry module would be a cleaner approach. Do you see/anticipate any issue with an additional module?

src/WebJobs.Script/Config/TransmissionStatusTelemetryModule.cs
RohitRanjanMS: The way we use AI SDK, the error message is going to be same for all the telemetry items. I am trying to keep the log message as small it can be.

src/WebJobs.Script/Config/TransmissionStatusTelemetryModule.cs
kshyju: Cold start impact? Now one more service needs to be resolved by the DI container.

mathewc: Yes, in addition there'll be a perf impact (however small) of AI having to dispatch to an additional module. If we can do this with a lot less code, I don't know why we wouldn't. The less code the better :) @brettsam may have thoughts on this as well.

brettsam: My only thought is that besides being more natural, the Module gives us a nice Dispose where we can unregister the Handler, which I think we'd need to do. 

I'm not sure what the right approach is if we don't use the Module.

mathewc: I don't see us unregistering the handler in Dispose below? Anyhow, if you don't see any big issues with use of a new module I'm fine with it. I don't have much experience with modules, and was looking for an easier way. If we too the approach I mentioned above and registered the handler, it's just a one liner to subscribe to the event. It's not clear to me whether we need to worry about unsubscribe - what's the lifetime of the telemetry channel? Perhaps when the host is disposed it gets disposed as well.

src/WebJobs.Script/Config/TransmissionStatusTelemetryModule.cs
RohitRanjanMS: Made the change.

src/WebJobs.Script/Config/TransmissionStatusTelemetryModule.cs
RohitRanjanMS: Removed the DI

src/WebJobs.Script/Config/TransmissionStatusHandler.cs
```diff
@@ -0,0 +1,147 @@
...
+    internal partial class LogMessageContext : JsonSerializerContext
+    {
+    }
+
+    internal class IngestionServiceResponse
+    {
+        [JsonPropertyName("itemsReceived")]
+        public int? ItemsReceived { get; set; }
+
+        [JsonPropertyName("itemsAccepted")]
```
brettsam, line 103: Do we need all these `JsonPropertyName` attributes if we're setting CamelCase on the serializer itself?

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -387,12 +388,18 @@ internal static void ConfigureApplicationInsights(HostBuilderContext context, IL
                 {
                     o.InstrumentationKey = appInsightsInstrumentationKey;
                     o.ConnectionString = appInsightsConnectionString;
-                }, t => t.TelemetryProcessorChainBuilder.Use(next => new ScriptTelemetryProcessor(next)));
+                }, t =>
+                {
+                    if (t.TelemetryChannel is ServerTelemetryChannel channel)
+                    {
+                        channel.TransmissionStatusEvent += new TransmissionStatusHandler().Handler;
```
brettsam: should TransmissionStatusHandler.Handler be static?

src/WebJobs.Script/Config/TransmissionStatusHandler.cs
RohitRanjanMS: IngestionServiceResponse is only used for Deserialize.

brettsam (approved):

src/WebJobs.Script/Config/TransmissionStatusHandler.cs
```diff
@@ -0,0 +1,147 @@
...
+using System.Linq;
+using System.Text.Json.Serialization;
+using Microsoft.ApplicationInsights.Channel;
+using Microsoft.ApplicationInsights.Extensibility.Implementation;
+using JsonSerializer = System.Text.Json.JsonSerializer;
+
+namespace Microsoft.Azure.WebJobs.Script.Config
+{
+    /// <summary>
+    /// Transmission Status Event Handler
```
liliankasem: Do you mind expanding on the documentation here for folks not familiar with that this is. What is a transmission status event handler? How is it used? What is the purpose of this handler etc.

RohitRanjanMS: Added comment.

src/WebJobs.Script/Config/TransmissionStatusHandler.cs
```diff
@@ -0,0 +1,148 @@
...
+                response = JsonSerializer.Deserialize(
+                    args.Response.Content, IngestionServiceResponseContext.Default.IngestionServiceResponse);
+            }
+
+            string topErrorMessage = null;
+            int? topStatusCode = null;
+            if (response?.Errors != null && response.Errors.Length > 0)
+            {
+                topErrorMessage = response.Errors[0].Message;
+                topStatusCode = response?.Errors[0].StatusCode;
```
liliankasem, line 59: a) because response? has already been checked in the if-statement and b) for consistency with line 58(either they should both have ? or not not have it 

```suggestion
                topStatusCode = response.Errors[0].StatusCode;
```

src/WebJobs.Script/Config/TransmissionStatusHandler.cs
```diff
@@ -0,0 +1,148 @@
...
+            }
+
+            var log = new LogMessage
+            {
+                StatusCode = args?.Response?.StatusCode,
```
liliankasem, line 64: Do we want to have a check for` if args?.Response == null` and `if response == null` early on instead of checking them multiple times?

src/WebJobs.Script/Config/TransmissionStatusHandler.cs
```diff
@@ -0,0 +1,148 @@
...
+                StatusDescription = args?.Response?.StatusDescription,
+                Id = transmission?.Id,
+                Items = items,
+                Received = response?.ItemsReceived,
+                Accepted = response?.ItemsAccepted,
```
liliankasem, line 69: What do we want to do if response is null? Is this property okay to be null or empty?

RohitRanjanMS: yes, response is null in case of success response. "Received" and "Accepted" won't be logged.

soninaren (approved):

PR #8867, opened 2022-10-26
Remove BindingData from DataType and update BindingMetadata

liliankasem:
- We're no longer using the DataType for sdk-type bindings, so the PR is removing the unused enum value "BindingData".

- Update BindingMetadata to include `Properties` prop. In the protobuf, the [Binding info was updated](https://github.com/Azure/azure-functions-language-worker-protobuf/blob/27a28ca180e6a34b8aea3349e3ffe46d2ff36ce7/src/proto/FunctionRpc.proto#L524-L525) to include a `Properties` map. We need to reflect this change by updating out representation of that message here.

### DIFF
--- a/src/WebJobs.Script.Abstractions/Description/Binding/BindingMetadata.cs
+++ b/src/WebJobs.Script.Abstractions/Description/Binding/BindingMetadata.cs
@@ -24,7 +24,16 @@ namespace Microsoft.Azure.WebJobs.Script.Description
         [JsonConstructor]
         private BindingMetadata(Dictionary<string, object> properties)
         {
-            Properties = new Dictionary<string, object>(properties, StringComparer.OrdinalIgnoreCase);
+            if (properties is null)
+            {
+                Properties = new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase);
+            }
+            else
+            {
+                Properties = properties.Count > 0
+                                ? new Dictionary<string, object>(properties, StringComparer.OrdinalIgnoreCase)
+                                : properties;
+            }
         }
 
         /// <summary>
--- a/src/WebJobs.Script.Abstractions/Description/Binding/BindingMetadata.cs
+++ b/src/WebJobs.Script.Abstractions/Description/Binding/BindingMetadata.cs
@@ -87,7 +87,10 @@ namespace Microsoft.Azure.WebJobs.Script.Description
             string bindingDirectionValue = (string)raw["direction"];
             string connection = (string)raw["connection"];
             string bindingType = (string)raw["type"];
-            var properties = raw["properties"].ToObject<IDictionary<string, object>>();
+            var properties = raw["properties"] != null
+                                ? raw["properties"].ToObject<IDictionary<string, object>>()
+                                : new Dictionary<string, object>();
+
             BindingDirection bindingDirection = default(BindingDirection);
 
             if (!string.IsNullOrEmpty(bindingDirectionValue) &&
--- a/src/WebJobs.Script.Abstractions/Description/Binding/BindingMetadata.cs
+++ b/src/WebJobs.Script.Abstractions/Description/Binding/BindingMetadata.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Globalization;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Converters;
@@ -87,9 +88,9 @@ namespace Microsoft.Azure.WebJobs.Script.Description
             string bindingDirectionValue = (string)raw["direction"];
             string connection = (string)raw["connection"];
             string bindingType = (string)raw["type"];
-            var properties = raw["properties"] != null
-                                ? raw["properties"].ToObject<IDictionary<string, object>>()
-                                : new Dictionary<string, object>();
+            IDictionary<string, object> properties = raw.TryGetValue("properties", StringComparison.OrdinalIgnoreCase, out JToken value)
+                            ? new Dictionary<string, object>(value.ToObject<IDictionary<string, object>>(), StringComparer.OrdinalIgnoreCase)
+                            : ImmutableDictionary<string, object>.Empty;
 
             BindingDirection bindingDirection = default(BindingDirection);
 
--- a/src/WebJobs.Script.Abstractions/Description/Binding/BindingMetadata.cs
+++ b/src/WebJobs.Script.Abstractions/Description/Binding/BindingMetadata.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Collections.Immutable;
 using System.Globalization;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Converters;
@@ -18,9 +17,12 @@ namespace Microsoft.Azure.WebJobs.Script.Description
     {
         private const string _systemReturnParameterBindingName = "$return";
 
-        public BindingMetadata()
+        public BindingMetadata() : this(new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase)) { }
+
+        [JsonConstructor]
+        private BindingMetadata(Dictionary<string, object> properties)
         {
-            Properties = new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase);
+            Properties = new Dictionary<string, object>(properties, StringComparer.OrdinalIgnoreCase);
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


surgupta-msft (approved):

fabiocav (approved):

mathewc (approved):

kashimiz (approved):

surgupta-msft (approved):

src/WebJobs.Script.Abstractions/Description/Binding/BindingMetadata.cs
```diff
@@ -81,6 +88,10 @@ public static BindingMetadata Create(JObject raw)
             string bindingDirectionValue = (string)raw["direction"];
             string connection = (string)raw["connection"];
             string bindingType = (string)raw["type"];
+            IDictionary<string, object> properties = raw.TryGetValue("properties", StringComparison.OrdinalIgnoreCase, out JToken value)
+                            ? new Dictionary<string, object>(value.ToObject<IDictionary<string, object>>(), StringComparer.OrdinalIgnoreCase)
+                            : ImmutableDictionary<string, object>.Empty;
```
jviau: Is `ImmutableDictionary` correct here? It implements `IDictionary` sure, but will it behave as expected? What will happen if someone tries to mutate it? `Properties` is technically writeable after all.

I worry we will have an inconsistent experience: if properties _starts_ with at least one value, I can mutate it. If it starts with _zero_ values, writing may fail. But I may be misunderstanding how `ImmutableDictionary` works when accessed through `IDictionary` contract.

kashimiz:
Approved with a question - do we have tests around this that need to be updated?

src/WebJobs.Script.Abstractions/Description/Binding/BindingMetadata.cs
liliankasem: That is a good question, Fabio originally made this suggestion, but maybe not realizing Properties has a setter? Maybe in this case just `new Dictionary<string, object>` make sense?

@fabiocav thoughts?

jviau: > Properties has a setter

It isn't that this has a setter or not, it is that this type is `IDictionary<>`, which is a _mutable_ contract. While `ImmutableDictionary` does implement it, it throws `NotSupportedException` anytime you try to actually use a [mutable API from it](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableDictionary_2.cs#L264). The concern is this could cause very obscure bugs due to the concrete type, and behavior, behind `Properties` changing significantly based on some pre-conditions.

`Dictionary<string, object>` would be more consistent (as you suggested). Or if we don't expect anyone to ever need to write to this dictionary, we should change the type to `IReadOnlyDictionary`. Then using `ImmutableDictionary` is completely fine.

liliankasem: Yup, that makes sense! Thanks for the details :) We are writing to the properties dictionary in another PR, but only if the dictionary contains a value, which in that case the type will be Dictionary and not IImmutableDictionary. But I can see where this will cause issues down the line - I'm happy to change this back to Dictionary<string, object>!

mathewc: You might consider making Properties behave the same way FunctionMetadata.Properties does https://github.com/Azure/azure-functions-host/blob/6bfbe194b317b2e20835f9c5ed8f9f6ab8d78765/src/WebJobs.Script.Abstractions/Description/FunctionMetadata.cs#L17. We initialize it in the default ctor, and the property itself has no public setter.

liliankasem: Ooo, I like that idea!

liliankasem:
> Approved with a question - do we have tests around this that need to be updated?

There aren't any that need an update, but I can add some new unit tests for fun

src/WebJobs.Script.Abstractions/Description/Binding/BindingMetadata.cs
```diff
@@ -95,6 +108,11 @@ public static BindingMetadata Create(JObject raw)
             bindingMetadata.Connection = connection;
             bindingMetadata.Raw = raw;
 
+            if (raw.TryGetValue("properties", StringComparison.OrdinalIgnoreCase, out JToken value))
+            {
+                bindingMetadata.Properties = new Dictionary<string, object>(value.ToObject<IDictionary<string, object>>(), StringComparer.OrdinalIgnoreCase);
```
jviau: Instead of assigning a new dictionary entirely, can you add the range of values from the `JToken` to the existing dictionary? Then the `private set` can be removed.

Alternatively, add a private ctor which takes in `Dictionary<>` and call that here:

``` CSharp
public BindingMetadata() : this(new Dictionary<string, object>()) { }

[JsonConstructor]
private BindingMetadata(Dictionary<string, object> properties)
{
    Properties = properties;
}

public static BindingMetadata Create(JObject raw)
{
    // verify raw is not null
    BindingMetadata bindingMetadata = raw.ToObject<BindingMetadata>();
    bindingMetadata.Raw = raw;

     // validate enums
    return bindingMetadata;
}
```

src/WebJobs.Script.Abstractions/Description/Binding/BindingMetadata.cs
```diff
@@ -81,6 +93,7 @@ public static BindingMetadata Create(JObject raw)
             string bindingDirectionValue = (string)raw["direction"];
```
jviau: This is an issue that exists before this PR, but I am actually a bit confused by the goal of this function. It seems to do a mix of manual and automatic deserialization? Am I correct in that there is some redundant assignments? The `raw.ToObject<BindingMetadata>()` along with the `string bindingType = (string)raw["type"];` and similar lines seem redundant? Won't the `ToObject<>()` call already set those properties?

I am also curious if a custom JSON type converter could work?

jviau: My other comment has a suggestion on how to refactor this method. I'll leave it up to you if you want to bother with that or not.

liliankasem: Yeah it is confusing, I can give your suggestion a quick try and if there's too much friction we can opt to address this in a different PR

liliankasem: @jviau was able to make your refactor suggestion work, will all the tests green - any additional feedback?

src/WebJobs.Script.Abstractions/Description/Binding/BindingMetadata.cs
```diff
@@ -78,24 +92,29 @@ public bool IsReturn
...
+            {
+                string bindingDirectionValue = (string)raw["direction"];
+                if (!string.IsNullOrEmpty(bindingDirectionValue) &&
+                    !Enum.TryParse<BindingDirection>(bindingDirectionValue, true, out BindingDirection bindingDirection))
+                {
+                    throw new FormatException(string.Format(CultureInfo.InvariantCulture, "'{0}' is not a valid binding direction.", bindingDirectionValue));
+                }
 
-            return bindingMetadata;
+                throw ex;
```
jviau:
```suggestion
                throw;
```

`throw ex;` will reset the stack trace on that exception. `throw;` will preserve the stack trace.

src/WebJobs.Script.Abstractions/Description/Binding/BindingMetadata.cs
```diff
@@ -78,24 +92,29 @@ public bool IsReturn
...
+
+                return bindingMetadata;
+            }
+            catch (Exception ex)
+            {
+                string bindingDirectionValue = (string)raw["direction"];
+                if (!string.IsNullOrEmpty(bindingDirectionValue) &&
+                    !Enum.TryParse<BindingDirection>(bindingDirectionValue, true, out BindingDirection bindingDirection))
+                {
+                    throw new FormatException(string.Format(CultureInfo.InvariantCulture, "'{0}' is not a valid binding direction.", bindingDirectionValue));
```
jviau: Can you include `ex` as an inner exception?

src/WebJobs.Script.Abstractions/Description/Binding/BindingMetadata.cs
```diff
@@ -16,6 +17,14 @@ public class BindingMetadata
     {
         private const string _systemReturnParameterBindingName = "$return";
 
+        public BindingMetadata() : this(new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase)) { }
```
jviau: My code suggestion had sloppy (lazy) formatting. Not sure if that matters to you much, but I would format this as:
```suggestion
        public BindingMetadata()
            : this(new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase))
        {
        }
```

jviau (approved):

PR #8883, opened 2022-11-02
Create new test project for Abstractions

liliankasem:
The reference to the Abstractions project in the Script tests is causing issues as the WebHost project references the Abstractions nuget package - this is causing some sort of conflict. This PR separates the Abstractions unit tests into their own project to avoid any conflict

### DIFF
--- a/test/WebJobs.Script.Abstractions/BindingMetadataTests.cs
+++ b/test/WebJobs.Script.Abstractions/BindingMetadataTests.cs
@@ -21,8 +21,8 @@ namespace Microsoft.Azure.WebJobs.Script.Tests
             Assert.Equal("AzureWebJobsStorage", result.Connection);
             Assert.Equal(triggerBinding, result.Raw);
             Assert.Equal(BindingDirection.In, result.Direction);
-            Assert.True(result.Properties.TryGetValue("SupportsDeferredBinding", out bool supportsDeferredBinding));
-            Assert.False(supportsDeferredBinding);
+            Assert.True(result.Properties.TryGetValue("SupportsDeferredBinding", out var supportsDeferredBinding));
+            Assert.False((bool)supportsDeferredBinding);
             Assert.True(result.IsTrigger);
             Assert.False(result.IsReturn);
         }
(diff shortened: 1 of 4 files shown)


surgupta-msft (approved):

test/WebJobs.Script.Abstractions/BindingMetadataTests.cs
```diff
@@ -21,8 +21,8 @@ public void BindingMetadata_Create_TriggerBinding_Success()
             Assert.Equal("AzureWebJobsStorage", result.Connection);
             Assert.Equal(triggerBinding, result.Raw);
             Assert.Equal(BindingDirection.In, result.Direction);
-            Assert.True(result.Properties.TryGetValue("SupportsDeferredBinding", out bool supportsDeferredBinding));
-            Assert.False(supportsDeferredBinding);
+            Assert.True(result.Properties.TryGetValue("SupportsDeferredBinding", out var supportsDeferredBinding));
```
jviau, line 24: Intrigued by this change. Did this parameter type change?

test/WebJobs.Script.Abstractions/xunit.runner.json
```diff
@@ -0,0 +1,3 @@
+﻿{
+    "parallelizeTestCollections": false
```
jviau: Any reason we don't want to parallelize these?

jviau (approved):

test/WebJobs.Script.Abstractions/xunit.runner.json
liliankasem: Ah good catch, copy/paste error - no reason not to parallelize

test/WebJobs.Script.Abstractions/BindingMetadataTests.cs
liliankasem: No the type is the same, always has been object - I think maybe the version of the language changed between the two test projects? Was also surprised to see this come up as an error - let me change this to net6.0 and see if this stays the same

liliankasem: Ah nope, that didn't do it xD I'm not sure what it could be

PR #8891, opened 2022-11-04
[LinuxDedicated] Emit Function Execution Logs faster

CooperLink:
This PR extends abandoned prior PR https://github.com/Azure/azure-functions-host/pull/8443.  It encompasses two logging changes that will only affect Linux App Service applications (Dedicated/EP). The change introduces an exponential backoff delay starting at 1 second and finishing at 32 seconds for Linux App Service FunctionExecutionCount events.  The logger uses the backoff delay to determine time between logging attempts. Platform takes dependency on the FunctionExecutionCount logger to determine if a Function is executing. If the host missed logging its initial execution for whatever reason, it would wait 30 seconds before attempting again. Therefore, the 30 second delay was causing platform to take a long time to determine when a function app started processing.

The other change present in this PR updates the Function Execution Event logs for Linux Dedicated and Elastic Premium to provide a more verbose explanation of the execution. 

This PR also puts the above feature behind a SCM Hosting Configuration flag to allow batched testing in production environments for both the LinuxFastLog feature above and Linux App Service FunctionExecutionCount metric logging. 

resolves #issue_for_this_pr

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
@@ -2,23 +2,47 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using System.Runtime.CompilerServices;
 using Microsoft.Azure.WebJobs.Script.Config;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
 
 namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 {
     internal class LinuxAppServiceEventGenerator : LinuxEventGenerator
     {
         private readonly Action<string> _writeEvent;
-        private readonly LinuxAppServiceFileLoggerFactory _loggerFactory;
         private readonly HostNameProvider _hostNameProvider;
+        private readonly IOptions<FunctionsHostingConfigOptions> _functionsHostingConfigOptions;
+        private readonly Lazy<ILinuxAppServiceFileLogger> _functionsExecutionEventsCategoryLogger;
+        private readonly Lazy<ILinuxAppServiceFileLogger> _functionsLogsCategoryLogger;
+        private readonly Lazy<ILinuxAppServiceFileLogger> _functionsMetricsCategoryLogger;
+        private readonly Lazy<ILinuxAppServiceFileLogger> _functionsDetailsCategoryLogger;
 
-        public LinuxAppServiceEventGenerator(LinuxAppServiceFileLoggerFactory loggerFactory, HostNameProvider hostNameProvider, Action<string> writeEvent = null)
+        public LinuxAppServiceEventGenerator(
+            ILinuxAppServiceFileLoggerFactory loggerFactory,
+            HostNameProvider hostNameProvider,
+            IOptions<FunctionsHostingConfigOptions> functionsHostingConfigOptions,
+            Action<string> writeEvent = null)
         {
             _writeEvent = writeEvent ?? WriteEvent;
-            _loggerFactory = loggerFactory;
             _hostNameProvider = hostNameProvider ?? throw new ArgumentNullException(nameof(hostNameProvider));
+            _functionsHostingConfigOptions = functionsHostingConfigOptions;
+            bool executionLogBackoffEnabled = !functionsHostingConfigOptions.Value.DisableLinuxAppServiceLogBackoff;
+            _functionsExecutionEventsCategoryLogger = loggerFactory.Create(FunctionsExecutionEventsCategory, backoffEnabled: executionLogBackoffEnabled);
+            _functionsLogsCategoryLogger = loggerFactory.Create(FunctionsLogsCategory, backoffEnabled: false);
+            _functionsMetricsCategoryLogger = loggerFactory.Create(FunctionsMetricsCategory, backoffEnabled: false);
+            _functionsDetailsCategoryLogger = loggerFactory.Create(FunctionsDetailsCategory, backoffEnabled: false);
         }
+
+        //Exposing Loggers for Unit tests
+        internal Lazy<ILinuxAppServiceFileLogger> FunctionsExecutionEventsCategoryLogger => _functionsExecutionEventsCategoryLogger;
+
+        internal Lazy<ILinuxAppServiceFileLogger> FunctionsLogsCategoryLogger => _functionsLogsCategoryLogger;
+
+        internal Lazy<ILinuxAppServiceFileLogger> FunctionsMetricsCategoryLogger => _functionsMetricsCategoryLogger;
+
+        internal Lazy<ILinuxAppServiceFileLogger> FunctionsDetailsCategoryLogger => _functionsDetailsCategoryLogger;
 
         public static string TraceEventRegex { get; } = "(?<Level>[0-6]),(?<SubscriptionId>[^,]*),(?<HostName>[^,]*),(?<AppName>[^,]*),(?<FunctionName>[^,]*),(?<EventName>[^,]*),(?<Source>[^,]*),\"(?<Details>.*)\",\"(?<Summary>.*)\",(?<HostVersion>[^,]*),(?<EventTimestamp>[^,]+),(?<ExceptionType>[^,]*),\"(?<ExceptionMessage>.*)\",(?<FunctionInvocationId>[^,]*),(?<HostInstanceId>[^,]*),(?<ActivityId>[^,\"]*)";
 
@@ -39,8 +63,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
             var hostName = _hostNameProvider.Value;
             using (FunctionsSystemLogsEventSource.SetActivityId(activityId))
             {
-                var logger = _loggerFactory.GetOrCreate(FunctionsLogsCategory);
-                WriteEvent(logger, $"{(int)ToEventLevel(level)},{subscriptionId},{hostName},{appName},{functionName},{eventName},{source},{NormalizeString(details)},{NormalizeString(summary)},{hostVersion},{formattedEventTimestamp},{exceptionType},{NormalizeString(exceptionMessage)},{functionInvocationId},{hostInstanceId},{activityId}");
+                WriteEvent(_functionsLogsCategoryLogger.Value, $"{(int)ToEventLevel(level)},{subscriptionId},{hostName},{appName},{functionName},{eventName},{source},{NormalizeString(details)},{NormalizeString(summary)},{hostVersion},{formattedEventTimestamp},{exceptionType},{NormalizeString(exceptionMessage)},{functionInvocationId},{hostInstanceId},{activityId}");
             }
         }
 
@@ -48,15 +71,13 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
             long minimum, long maximum, long count, DateTime eventTimestamp, string data, string runtimeSiteName, string slotName)
         {
             var hostVersion = ScriptHost.Version;
-            var logger = _loggerFactory.GetOrCreate(FunctionsMetricsCategory);
-            WriteEvent(logger, $"{subscriptionId},{appName},{functionName},{eventName},{average},{minimum},{maximum},{count},{hostVersion},{eventTimestamp.ToString(EventTimestampFormat)},{data}");
+            WriteEvent(_functionsMetricsCategoryLogger.Value, $"{subscriptionId},{appName},{functionName},{eventName},{average},{minimum},{maximum},{count},{hostVersion},{eventTimestamp.ToString(EventTimestampFormat)},{data}");
         }
 
         public override void LogFunctionDetailsEvent(string siteName, string functionName, string inputBindings, string outputBindings,
             string scriptType, bool isDisabled)
         {
-            var logger = _loggerFactory.GetOrCreate(FunctionsDetailsCategory);
-            WriteEvent(logger, $"{siteName},{functionName},{NormalizeString(inputBindings)},{NormalizeString(outputBindings)},{scriptType},{(isDisabled ? 1 : 0)}");
+            WriteEvent(_functionsDetailsCategoryLogger.Value, $"{siteName},{functionName},{NormalizeString(inputBindings)},{NormalizeString(outputBindings)},{scriptType},{(isDisabled ? 1 : 0)}");
         }
 
         public override void LogFunctionExecutionAggregateEvent(string siteName, string functionName, long executionTimeInMs,
@@ -67,20 +88,20 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
         public override void LogFunctionExecutionEvent(string executionId, string siteName, int concurrency, string functionName,
             string invocationId, string executionStage, long executionTimeSpan, bool success)
         {
-            var logger = _loggerFactory.GetOrCreate(FunctionsExecutionEventsCategory);
             string currentUtcTime = DateTime.UtcNow.ToString();
-            if (FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagEnableLinuxEPExecutionCount))
+            bool detailedExecutionEventsDisabled = _functionsHostingConfigOptions.Value.DisableLinuxAppServiceExecutionDetails;
+            if (!detailedExecutionEventsDisabled)
             {
                 string log = string.Join(",", executionId, siteName, concurrency.ToString(), functionName, invocationId, executionStage, executionTimeSpan.ToString(), success.ToString(), currentUtcTime);
-                WriteEvent(logger, log);
+                WriteEvent(_functionsExecutionEventsCategoryLogger.Value, log);
             }
             else
             {
-                WriteEvent(logger, currentUtcTime);
+                WriteEvent(_functionsExecutionEventsCategoryLogger.Value, currentUtcTime);
             }
         }
 
-        private static void WriteEvent(LinuxAppServiceFileLogger logger, string evt)
+        private static void WriteEvent(ILinuxAppServiceFileLogger logger, string evt)
         {
             logger.Log(evt);
         }
--- a/src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
@@ -2,7 +2,6 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
-using System.Runtime.CompilerServices;
 using Microsoft.Azure.WebJobs.Script.Config;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
@@ -14,10 +13,10 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
         private readonly Action<string> _writeEvent;
         private readonly HostNameProvider _hostNameProvider;
         private readonly IOptions<FunctionsHostingConfigOptions> _functionsHostingConfigOptions;
-        private readonly Lazy<ILinuxAppServiceFileLogger> _functionsExecutionEventsCategoryLogger;
-        private readonly Lazy<ILinuxAppServiceFileLogger> _functionsLogsCategoryLogger;
-        private readonly Lazy<ILinuxAppServiceFileLogger> _functionsMetricsCategoryLogger;
-        private readonly Lazy<ILinuxAppServiceFileLogger> _functionsDetailsCategoryLogger;
+        private ILinuxAppServiceFileLogger _functionsExecutionEventsCategoryLogger;
+        private ILinuxAppServiceFileLogger _functionsLogsCategoryLogger;
+        private ILinuxAppServiceFileLogger _functionsMetricsCategoryLogger;
+        private ILinuxAppServiceFileLogger _functionsDetailsCategoryLogger;
 
         public LinuxAppServiceEventGenerator(
             ILinuxAppServiceFileLoggerFactory loggerFactory,
@@ -28,21 +27,19 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
             _writeEvent = writeEvent ?? WriteEvent;
             _hostNameProvider = hostNameProvider ?? throw new ArgumentNullException(nameof(hostNameProvider));
             _functionsHostingConfigOptions = functionsHostingConfigOptions;
-            bool executionLogBackoffEnabled = !functionsHostingConfigOptions.Value.DisableLinuxAppServiceLogBackoff;
-            _functionsExecutionEventsCategoryLogger = loggerFactory.Create(FunctionsExecutionEventsCategory, backoffEnabled: executionLogBackoffEnabled);
+            _functionsExecutionEventsCategoryLogger = loggerFactory.Create(FunctionsExecutionEventsCategory, backoffEnabled: !_functionsHostingConfigOptions.Value.DisableLinuxAppServiceLogBackoff);
             _functionsLogsCategoryLogger = loggerFactory.Create(FunctionsLogsCategory, backoffEnabled: false);
-            _functionsMetricsCategoryLogger = loggerFactory.Create(FunctionsMetricsCategory, backoffEnabled: false);
-            _functionsDetailsCategoryLogger = loggerFactory.Create(FunctionsDetailsCategory, backoffEnabled: false);
+            _functionsMetricsCategoryLogger = loggerFactory.Create(FunctionsMetricsCategory, false);
+            _functionsDetailsCategoryLogger = loggerFactory.Create(FunctionsDetailsCategory, false);
         }
 
-        //Exposing Loggers for Unit tests
-        internal Lazy<ILinuxAppServiceFileLogger> FunctionsExecutionEventsCategoryLogger => _functionsExecutionEventsCategoryLogger;
+        internal ILinuxAppServiceFileLogger FunctionsExecutionEventsCategoryLogger => _functionsExecutionEventsCategoryLogger;
 
-        internal Lazy<ILinuxAppServiceFileLogger> FunctionsLogsCategoryLogger => _functionsLogsCategoryLogger;
+        internal ILinuxAppServiceFileLogger FunctionsLogsCategoryLogger => _functionsLogsCategoryLogger;
 
-        internal Lazy<ILinuxAppServiceFileLogger> FunctionsMetricsCategoryLogger => _functionsMetricsCategoryLogger;
+        internal ILinuxAppServiceFileLogger FunctionsMetricsCategoryLogger => _functionsMetricsCategoryLogger;
 
-        internal Lazy<ILinuxAppServiceFileLogger> FunctionsDetailsCategoryLogger => _functionsDetailsCategoryLogger;
+        internal ILinuxAppServiceFileLogger FunctionsDetailsCategoryLogger => _functionsDetailsCategoryLogger;
 
         public static string TraceEventRegex { get; } = "(?<Level>[0-6]),(?<SubscriptionId>[^,]*),(?<HostName>[^,]*),(?<AppName>[^,]*),(?<FunctionName>[^,]*),(?<EventName>[^,]*),(?<Source>[^,]*),\"(?<Details>.*)\",\"(?<Summary>.*)\",(?<HostVersion>[^,]*),(?<EventTimestamp>[^,]+),(?<ExceptionType>[^,]*),\"(?<ExceptionMessage>.*)\",(?<FunctionInvocationId>[^,]*),(?<HostInstanceId>[^,]*),(?<ActivityId>[^,\"]*)";
 
@@ -63,7 +60,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
             var hostName = _hostNameProvider.Value;
             using (FunctionsSystemLogsEventSource.SetActivityId(activityId))
             {
-                WriteEvent(_functionsLogsCategoryLogger.Value, $"{(int)ToEventLevel(level)},{subscriptionId},{hostName},{appName},{functionName},{eventName},{source},{NormalizeString(details)},{NormalizeString(summary)},{hostVersion},{formattedEventTimestamp},{exceptionType},{NormalizeString(exceptionMessage)},{functionInvocationId},{hostInstanceId},{activityId}");
+                WriteEvent(_functionsLogsCategoryLogger, $"{(int)ToEventLevel(level)},{subscriptionId},{hostName},{appName},{functionName},{eventName},{source},{NormalizeString(details)},{NormalizeString(summary)},{hostVersion},{formattedEventTimestamp},{exceptionType},{NormalizeString(exceptionMessage)},{functionInvocationId},{hostInstanceId},{activityId}");
             }
         }
 
@@ -71,13 +68,13 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
             long minimum, long maximum, long count, DateTime eventTimestamp, string data, string runtimeSiteName, string slotName)
         {
             var hostVersion = ScriptHost.Version;
-            WriteEvent(_functionsMetricsCategoryLogger.Value, $"{subscriptionId},{appName},{functionName},{eventName},{average},{minimum},{maximum},{count},{hostVersion},{eventTimestamp.ToString(EventTimestampFormat)},{data}");
+            WriteEvent(_functionsMetricsCategoryLogger, $"{subscriptionId},{appName},{functionName},{eventName},{average},{minimum},{maximum},{count},{hostVersion},{eventTimestamp.ToString(EventTimestampFormat)},{data}");
         }
 
         public override void LogFunctionDetailsEvent(string siteName, string functionName, string inputBindings, string outputBindings,
             string scriptType, bool isDisabled)
         {
-            WriteEvent(_functionsDetailsCategoryLogger.Value, $"{siteName},{functionName},{NormalizeString(inputBindings)},{NormalizeString(outputBindings)},{scriptType},{(isDisabled ? 1 : 0)}");
+            WriteEvent(_functionsDetailsCategoryLogger, $"{siteName},{functionName},{NormalizeString(inputBindings)},{NormalizeString(outputBindings)},{scriptType},{(isDisabled ? 1 : 0)}");
         }
 
         public override void LogFunctionExecutionAggregateEvent(string siteName, string functionName, long executionTimeInMs,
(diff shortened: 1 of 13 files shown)


src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceFileLogger.cs
```diff
@@ -87,7 +92,11 @@ public virtual async Task ProcessLogQueue(object state)
             while (!_cancellationTokenSource.IsCancellationRequested)
             {
                 await InternalProcessLogQueue();
-                await Task.Delay(TimeSpan.FromSeconds(FlushFrequencySeconds), _cancellationTokenSource.Token).ContinueWith(task => { });
+                await Task.Delay(TimeSpan.FromSeconds(_flushFrequencySeconds), _cancellationTokenSource.Token).ContinueWith(task => { });
+                if ( _currentFlushFrequencySeconds < MaxFlushFrequencySeconds)
```
mathewc: The way you have this structured now we're never resetting back to 1, so once you increment to 32 you'll stay there. I think instead you want InternalProcessLogQueue to return a bool indicating whether anything was written. Then, you only back off if there are no events written, otherwise you reset to 1 (meaning we're active again and need to be more responsive). Looks like you're not using this new _currentFlushFrequencySeconds value above? I think you'll want to set this value prior then use it in the delay.

Also, you'd want to use _currentFlushFrequencySeconds = Min(MaxFlushFrequencySeconds, _currentFlushFrequencySeconds * FlushFrequencyGrowthFactor) so we don't go over max.

In practice, you should see how far you actually end up backing off. Because all our system logs go through this logger, I'd expect even if no functions are executing, some of our other periodic logs (like lease acquisitions, etc.) would keep things hot, so the question is whether we'd end up always using 1s or a really low number in most cases.

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceFileLogger.cs
```diff
@@ -46,7 +48,10 @@ public LinuxAppServiceFileLogger(string logFileName, string logFileDirectory, IF
         public int MaxFileSizeMb { get; set; } = 10;
 
         // Maximum time between successive flushes (seconds)
-        public int FlushFrequencySeconds { get; set; } = 30;
+         public int MaxFlushFrequencySeconds { get; set; } = 30;
+
+        // Flush Frequency Growth Factor
+        public int FlushFrequencyGrowthFactor { get; set; } = 2;
```
mathewc: I think you can just remove this and hardcode the factor to 2 inline where it's used.

src/WebJobs.Script/ScriptConstants.cs
```diff
@@ -124,7 +124,8 @@ public static class ScriptConstants
         public const string FeatureFlagEnableDebugTracing = "EnableDebugTracing";
         public const string FeatureFlagEnableMultiLanguageWorker = "EnableMultiLanguageWorker";
         public const string FeatureFlagEnableProxies = "EnableProxies";
-        public const string FeatureFlagEnableLinuxEPExecutionCount = "EnableLinuxFEC";
+        public const string HostingConfigEnableLinuxEPExecutionCount = "EnableLinuxFEC";
+        public const string HostingConfigEnableLinuxEPFastLog = "EnableLinuxEPFastLog";
```
mathewc: nit: I'd remove the EP bit - it's not really specific to EP. Also, might rename to EnableLinuxLogBackoff?

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
```diff
@@ -111,6 +111,7 @@ public static void AddWebJobsScriptHost(this IServiceCollection services, IConfi
                 else if (SystemEnvironment.Instance.IsLinuxAppService())
                 {
                     var hostNameProvider = p.GetService<HostNameProvider>();
+                    var functionsHostingConfiguration = p.GetService<IOptions<FunctionsHostingConfiguration>>().Value;
```
mathewc: Not seeing this being used?

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
```diff
@@ -12,12 +12,14 @@ internal class LinuxAppServiceEventGenerator : LinuxEventGenerator
         private readonly Action<string> _writeEvent;
         private readonly LinuxAppServiceFileLoggerFactory _loggerFactory;
         private readonly HostNameProvider _hostNameProvider;
+        private readonly IFunctionsHostingConfiguration _functionsHostingConfiguration;
 
-        public LinuxAppServiceEventGenerator(LinuxAppServiceFileLoggerFactory loggerFactory, HostNameProvider hostNameProvider, Action<string> writeEvent = null)
+        public LinuxAppServiceEventGenerator(LinuxAppServiceFileLoggerFactory loggerFactory, HostNameProvider hostNameProvider, Action<string> writeEvent = null, FunctionsHostingConfiguration functionsHostingConfiguration)
```
mathewc: Change parameter type to IFunctionsHostingConfiguration

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
```diff
@@ -67,9 +69,11 @@ public LinuxAppServiceEventGenerator(LinuxAppServiceFileLoggerFactory loggerFact
         public override void LogFunctionExecutionEvent(string executionId, string siteName, int concurrency, string functionName,
             string invocationId, string executionStage, long executionTimeSpan, bool success)
         {
-            var logger = _loggerFactory.GetOrCreate(FunctionsExecutionEventsCategory);
+            bool fastLinuxLoggingEnabled = !string.IsNullOrEmpty(_functionsHostingConfiguration.GetValue(ScriptConstants.HostingConfigEnableLinuxEPFastLog));
+            var logger = _loggerFactory.GetOrCreate(FunctionsExecutionEventsCategory, fastLinuxLoggingEnabled);
             string currentUtcTime = DateTime.UtcNow.ToString();
-            if (FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagEnableLinuxEPExecutionCount))
```
mathewc, line 72: So this was previously mistakenly made a feature flag? I agree that it wouldn't make sense for a user to configure this flag.

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
```diff
@@ -67,9 +69,11 @@ public LinuxAppServiceEventGenerator(LinuxAppServiceFileLoggerFactory loggerFact
         public override void LogFunctionExecutionEvent(string executionId, string siteName, int concurrency, string functionName,
             string invocationId, string executionStage, long executionTimeSpan, bool success)
         {
-            var logger = _loggerFactory.GetOrCreate(FunctionsExecutionEventsCategory);
+            bool fastLinuxLoggingEnabled = !string.IsNullOrEmpty(_functionsHostingConfiguration.GetValue(ScriptConstants.HostingConfigEnableLinuxEPFastLog));
```
mathewc, line 72: Add tests for this new mode/logic to LinuxAppServiceEventGeneratorTests

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceFileLogger.cs
```diff
@@ -22,13 +22,15 @@ public class LinuxAppServiceFileLogger
         private readonly IFileSystem _fileSystem;
         private readonly CancellationTokenSource _cancellationTokenSource;
         private Task _outputTask;
+        private int _currentFlushFrequencySeconds;
 
-        public LinuxAppServiceFileLogger(string logFileName, string logFileDirectory, IFileSystem fileSystem, bool startOnCreate = true)
+        public LinuxAppServiceFileLogger(string logFileName, string logFileDirectory, IFileSystem fileSystem, bool startOnCreate = true, bool linuxFastLog = false)
```
mathewc: Add tests for this new mode to LinuxAppServiceEventGeneratorTests

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
CooperLink: I was not aware of HostingConfig flags until the demo. When FunctionsHost started emitting FunctionExecutionCount logs with fields, it caused a memory leak in the metering service.  To disable it, I used the FeatureFlags.  HostingConfig flags will be much better for toggling the change if it causes any further issues.

fabiocav (commented):
A couple of things on this PR:
 - Are tests still pending? Would be good to have validation
 - This is taking a dependency on the new hosting configuration API, which is an API we're a bit concerned about and would be nice to avoid expanding usage until fixed. (@pragnagopa  for info as well). For that reason, it would be good to hold on this change

fabiocav (changes requested):
Flagging as changes requested until we close on the question above.

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceFileLogger.cs
CooperLink: Thank you for your review Mathew. In this case, my goal is to emit the first FunctionExecutionCount log as fast as possible.  After the first log is emitted, I don't mind waiting 30 seconds between emissions. Platform picks up the any available FunctionExecutionCount log and reports to the DataService that a worker is performing and can be scaled. 

I have incorporated lots of your suggestions. I appreciate the thorough review. I don't think I have a complete grasp of how the File Loggers are working. My current understanding is that this 30 seconds delay is not actually a maximum. I may have that incorrect. But, in private testing I found a marked increase in scaling speed when I implemented the backoff. Therefore, I think that the loggers are only filing every `FlushFrequency` seconds. Atleast it seems that way for the very first log emission.

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceFileLogger.cs
CooperLink: Coded inline thanks

src/WebJobs.Script/ScriptConstants.cs
CooperLink: Removed EP portions. While the Platform functionality is EP specific (scaling), the host func is not.

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
CooperLink: Fixed this. Thank you for calling it out. Didn't get saved before I committed.

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceFileLogger.cs
mathewc: If that's the case, then I guess I don't understand these changes. This logger is used for all logs, not only function execution logs. What is the "FunctionExecutionCount" log you're referring to? There's no guarantees on what logs will happen first after startup, so how do you know what ends up in the first batch to be flushed? Its unclear what you're trying to accomplish here.

The current logger will only flush logs every 30s. It accumulates them and flushes them in batches. With the changes you're making, it'll flush at after 1s, then 2s, then 4s, ... 30s, 30s. Why would we want to do that?

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceFileLogger.cs
```diff
@@ -46,7 +48,7 @@ public LinuxAppServiceFileLogger(string logFileName, string logFileDirectory, IF
         public int MaxFileSizeMb { get; set; } = 10;
 
         // Maximum time between successive flushes (seconds)
-        public int FlushFrequencySeconds { get; set; } = 30;
+         public int MaxFlushFrequencySeconds { get; set; } = 30;
```
mathewc: nit: looks like you've got an extra space here. I see the same on line 93 below. Maybe your editor is doing this? Please check other places

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceFileLogger.cs
```diff
@@ -22,13 +22,15 @@ public class LinuxAppServiceFileLogger
...
+        private int _currentFlushFrequencySeconds;
 
-        public LinuxAppServiceFileLogger(string logFileName, string logFileDirectory, IFileSystem fileSystem, bool startOnCreate = true)
+        public LinuxAppServiceFileLogger(string logFileName, string logFileDirectory, IFileSystem fileSystem, bool startOnCreate = true, bool logBackoffEnabled = false)
         {
             _logFileName = logFileName;
             _logFileDirectory = logFileDirectory;
             _logFilePath = Path.Combine(_logFileDirectory, _logFileName + ".log");
             _buffer = new BlockingCollection<string>(new ConcurrentQueue<string>());
+            _currentFlushFrequencySeconds = logBackoffEnabled ? 1 : 30;
```
mathewc: Rather than duplicate 30 here, I believe the intent is to use MaxFlushFrequencySeconds? I.e. if MaxFlushFrequencySeconds was set to a differnet value, you'd want to use that here as well. To simplify this, you could just remove this field and ctor logic and define a local in ProcessLogQueue to manage this.

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceFileLogger.cs
mathewc: I see what you're trying to accomplish now. When "logBackoffEnabled" is specified on construction, the logger should log its first batch fast then progressively back off, to effect more responsive startup time logging. You're not trying to implement a activity based backoff that resets itself. At steady state after startup, we're fine with the 30s loop.

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceFileLoggerFactory.cs
```diff
@@ -17,10 +17,10 @@ public LinuxAppServiceFileLoggerFactory()
             _logRootPath = Environment.GetEnvironmentVariable(EnvironmentSettingNames.FunctionsLogsMountPath);
         }
 
-        public virtual LinuxAppServiceFileLogger GetOrCreate(string category)
+        public virtual LinuxAppServiceFileLogger GetOrCreate(string category, bool logBackoffEnabled = false)
         {
             return Loggers.GetOrAdd(category,
-                static (c, path) => new Lazy<LinuxAppServiceFileLogger>(() => new LinuxAppServiceFileLogger(c, path, new FileSystem())), _logRootPath).Value;
```
mathewc, line 23: The caching here is technically incorrect now. For example if we create 2 different loggers for category "foo" with logBackoffEnabled=true/false respectively, we'll cache the first and use it in all cases.

src/WebJobs.Script/ScriptConstants.cs
```diff
@@ -124,7 +124,8 @@ public static class ScriptConstants
         public const string FeatureFlagEnableDebugTracing = "EnableDebugTracing";
         public const string FeatureFlagEnableMultiLanguageWorker = "EnableMultiLanguageWorker";
         public const string FeatureFlagEnableProxies = "EnableProxies";
-        public const string FeatureFlagEnableLinuxEPExecutionCount = "EnableLinuxFEC";
+        public const string HostingConfigEnableLinuxAppServiceExecutionCount = "EnableLinuxFEC";
+        public const string HostingConfigEnableLinuxLogBackoff = "EnableLinuxLogBackoff";
```
mathewc: The backoff logic you added is scoped only to function execution events, not all linux logging in general. So it'd be best for this config flag to have a more specific name. E.g. EnableLinuxExecutionEventLogBackoff or something.

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
```diff
@@ -67,9 +69,11 @@ public LinuxAppServiceEventGenerator(LinuxAppServiceFileLoggerFactory loggerFact
         public override void LogFunctionExecutionEvent(string executionId, string siteName, int concurrency, string functionName,
             string invocationId, string executionStage, long executionTimeSpan, bool success)
         {
-            var logger = _loggerFactory.GetOrCreate(FunctionsExecutionEventsCategory);
+            bool logBackoffEnabled = !string.IsNullOrEmpty(_functionsHostingConfiguration.GetValue(ScriptConstants.HostingConfigEnableLinuxLogBackoff));
+            var logger = _loggerFactory.GetOrCreate(FunctionsExecutionEventsCategory, logBackoffEnabled);
             string currentUtcTime = DateTime.UtcNow.ToString();
-            if (FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagEnableLinuxEPExecutionCount))
+            bool executionCountMetricEnabled = !string.IsNullOrEmpty(_functionsHostingConfiguration.GetValue(ScriptConstants.HostingConfigEnableLinuxAppServiceExecutionCount));
```
mathewc: Why is the naming here referencing "execution count"? We're not logging any execution count. The behavior being switched is "detailed execution events", right?

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
```diff
@@ -67,9 +69,11 @@ public LinuxAppServiceEventGenerator(LinuxAppServiceFileLoggerFactory loggerFact
         public override void LogFunctionExecutionEvent(string executionId, string siteName, int concurrency, string functionName,
             string invocationId, string executionStage, long executionTimeSpan, bool success)
         {
-            var logger = _loggerFactory.GetOrCreate(FunctionsExecutionEventsCategory);
```
mathewc, line 70: I want to point out that LogFunctionExecutionEvent is called by MetricsEventManager at two different points in time. Immediately when function executions complete, and every 5 seconds when longer running activities are flushed. So for long running functions, there'll be a 5s delay from MetricsEventManager before it's even flushed to your backoff code.

That'll be much better than the 30+ second delay we can see now, but if we wanted to avoid a 5second delay we'd have to consider other designs.

src/WebJobs.Script/ScriptConstants.cs
```diff
@@ -124,7 +124,8 @@ public static class ScriptConstants
         public const string FeatureFlagEnableDebugTracing = "EnableDebugTracing";
         public const string FeatureFlagEnableMultiLanguageWorker = "EnableMultiLanguageWorker";
         public const string FeatureFlagEnableProxies = "EnableProxies";
-        public const string FeatureFlagEnableLinuxEPExecutionCount = "EnableLinuxFEC";
+        public const string HostingConfigEnableLinuxAppServiceExecutionCount = "EnableLinuxFEC";
```
mathewc: based on my comments above, we may want to rename this as well. E.g. EnableLinuxDetailedExecutionEvents

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
CooperLink: Thank you for bringing this up.  Just re-iterating to make sure that I understand you
When this PR is merged and deployed, the time to first execution event log will be Min(ExecutionTime, 5s).  This could delay scale out slightly for functions with longer execution times.

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
CooperLink: Updated this to be ExecutionDetails instead of Count.  We use the metric on Platform to derive the FunctionExecutionCount metric.  But, on Host these are just execution events :)

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceFileLogger.cs
CooperLink: I replaced 30 with MaxFlushFrequencySeconds. Thank you for the callout. I am not sure that I understand how moving to ProcessLogQueue would simplify the code here.

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceFileLoggerFactory.cs
CooperLink: I took note of this. The limitation of caching the logger is that hosting config for these settings only get propagated on host startup.  For fastlogging this matters less as it should only affect the first execution that occurs when a host starts its workers I believe.  This setting would only update on host start/restart. 
To avoid the state where the Host picks up a config change but, cannot apply it to the logger due to caching. I will try to make the FastLogging configuration static. Then whatever Host's state is should be applied to its logger when its instantiated and neither can be changed without a host restart. 

ExecutionDetails can make use of the HostingConfigMonitor. So in the case that detailed logs need to be shut off, we should be able to affect all running Functions.

src/WebJobs.Script/ScriptConstants.cs
CooperLink: I did update this to HostingConfigEnableLinuxAppServiceDetailedExecutionEvents.  
This code does not run for Linux Consumption so i wanted to try to make that clear.

src/WebJobs.Script/ScriptConstants.cs
CooperLink: Provided more detailed names for these vars. Thank you for your review Mathew. 
I am working on applying the changes to hosting config that Alexey made. Then I am going to work on testing.

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceFileLogger.cs
CooperLink: Thank you for calling this out. I have taken a look through the changed files for leading or trailing spaces.

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
```diff
@@ -67,9 +74,11 @@ public LinuxAppServiceEventGenerator(LinuxAppServiceFileLoggerFactory loggerFact
...
+            bool logBackoffEnabled = !_functionsHostingConfigOptions.Value?.DisableLinuxLogBackoff ?? true;
```
alrod: _functionsHostingConfigOptions.Value is always should have value, can. be simplified to:

`bool logBackoffEnabled = !_functionsHostingConfigOptions.Value.DisableLinuxLogBackoff`

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceFileLogger.cs
```diff
@@ -46,7 +48,7 @@ public LinuxAppServiceFileLogger(string logFileName, string logFileDirectory, IF
         public int MaxFileSizeMb { get; set; } = 10;
 
         // Maximum time between successive flushes (seconds)
-        public int FlushFrequencySeconds { get; set; } = 30;
+        public int MaxFlushFrequencySeconds { get; set; } = 30;
```
alrod: Does it need to be public?

CooperLink: I believe that these vars are public for use in Unit Tests. The MaxFlushFrequencySeconds is not currently used anywhere from what I can see. I will change the access to private on this one, 
Edit : I set it to private and removed the setter

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
CooperLink: Thank you for the review. Changed this

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
```diff
@@ -112,7 +112,8 @@ public static void AddWebJobsScriptHost(this IServiceCollection services, IConfi
                 else if (SystemEnvironment.Instance.IsLinuxAppService())
                 {
                     var hostNameProvider = p.GetService<HostNameProvider>();
-                    return new LinuxAppServiceEventGenerator(new LinuxAppServiceFileLoggerFactory(), hostNameProvider);
+                    IOptions<FunctionsHostingConfigOptions> functionsHostingConfigOptions = p.GetService<IOptions<FunctionsHostingConfigOptions>>();
```
alrod, line 115: We need to validate setup is already executed for `IOptions<FunctionsHostingConfigOptions>` here

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceFileLogger.cs
```diff
@@ -87,7 +89,11 @@ public virtual async Task ProcessLogQueue(object state)
             while (!_cancellationTokenSource.IsCancellationRequested)
             {
                 await InternalProcessLogQueue();
-                await Task.Delay(TimeSpan.FromSeconds(FlushFrequencySeconds), _cancellationTokenSource.Token).ContinueWith(task => { });
+                await Task.Delay(TimeSpan.FromSeconds(_currentFlushFrequencySeconds), _cancellationTokenSource.Token).ContinueWith(task => { });
+                if (_currentFlushFrequencySeconds < MaxFlushFrequencySeconds)
+                {
+                    _currentFlushFrequencySeconds = Math.Min(MaxFlushFrequencySeconds, _currentFlushFrequencySeconds * 2);
```
alrod: why do we multiplying 2 here?

CooperLink: This is the backoff mechanism for fast file logging. So the delay on the Task is set to the following sequence: 1, 2, 4, 8, 16, 30 and maxes out at 30

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceFileLogger.cs
CooperLink: I updated the LinuxAppServiceEventGeneratorTests.cs FunctionExecutionEvents test.  It takes in a boolean to determine whether to set the HostingConfig. And then uses that boolean to verify the FileLogger is emitting the expected schema given feature on/off

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceFileLoggerFactory.cs
CooperLink: With this in mind, I chose to use the static definition of HostingConfiguration tied to lifecycle. There should hopefully be no confusion on the cached file logger now.  Whichever logger type is assigned to these events to begin with will persist. 

ExecutionDetails relies on the static hosting configuration as well.  So if this feature needs to be disabled, only new hosts after the HostingConfig is pushed will affect a change.

src/WebJobs.Script/ScriptConstants.cs
CooperLink: This has been renamed to the following flags 
DisableLinuxAppServiceLogBackoff and DisableLinuxAppServiceExecutionDetails

For more prescriptive names,.

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
CooperLink: Validated that HostingConfigOptions gets loaded up before the IEventGenerator is resolved. Discussed offline. Resolving

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
mathewc: Yes

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceFileLoggerFactory.cs
```diff
@@ -20,7 +20,15 @@ public LinuxAppServiceFileLoggerFactory()
...
+                static (c, path) => new Lazy<LinuxAppServiceFileLogger>(() => new LinuxAppServiceFileLogger(c, path, new FileSystem(), false)), _logRootPath).Value;
+        }
+
+        public virtual LinuxAppServiceFileLogger GetOrCreateBackoff(string category)
+        {
+            // If logbackoff is enabled, we want to create a new logger instance
+            // so that we can backoff flush frequency and emit the first log faster. This benefits EP scaling
+            return Loggers.GetOrAdd(category,
```
mathewc: I still don't like the fact that we're using the same underlying Loggers cache dictionary for these two methods which means we can run into the same incorrect caching I mentioned above. E.g. if someone calls GetOrCreate("foo") first, then GetOrCreateBackoff("foo") - the latter won't return a backoff logger.

Instead, I'm thinking that you can just have a new factory method like this that is called from the LinuxAppServiceEventGenerator constructor:

```csharp
public virtual Lazy<LinuxAppServiceFileLogger> Create(string category, bool backoffEnabled)
{
    return new Lazy<LinuxAppServiceFileLogger>(() => new LinuxAppServiceFileLogger(category, _logRootPath, new FileSystem(), backoffEnabled: backoffEnabled));
}
```

This sidesteps the caching issues in the shared Logger cache.

CooperLink: I don't think that I understand where the create call would go in this case. Cache is confusing and would like to fix that. But,  the fileLoggerFactory Loggers dictionary is how we reference the EventGenerators during every logging call. If we don't add the FileLogger to that dictionary I am not sure I understand how to update the eventGenerator code.

mathewc: You'd simply create it in the LinuxAppServiceEventGenerator constructor:

```csharp
_executionEventsLogger = loggerFactory.Create(FunctionsExecutionEventsCategory, backoffEnabled: true);
```

Then in LogFunctionExecutionEvent you'd use this logger via _executionEventsLogger.Value.

CooperLink: Oh alright! I was thinking that GetOrCreate was required in case the Logger failed or exited or something. Then GetOrCreate would instantiate a new logger. But, I understand what you are saying I think I can add that now

pragnagopa: Thanks @mathewc ! It does make it simpler to remove the cache altogether as there are only 4 log categories --> only 4 loggers need to be created.

@CooperLink - Addressed this comment in https://github.com/Azure/azure-functions-host/pull/9066 --> please take a look and see if it makes sense, if yes, go ahead and merge - check if unit test coverage if good!

CooperLink: Udpated the code to remove cache. Need to wait for a VS update but, then I will update the Unit Testing as well.

test/WebJobs.Script.Tests/Diagnostics/LinuxAppServiceEventGeneratorTests.cs
```diff
@@ -45,13 +48,16 @@ public LinuxAppServiceEventGeneratorTests()
 
             var loggerFactoryMock = new Mock<LinuxAppServiceFileLoggerFactory>(MockBehavior.Strict);
             loggerFactoryMock.Setup(f => f.GetOrCreate(It.IsAny<string>())).Returns<string>(s => _loggers[s]);
+            loggerFactoryMock.Setup(f => f.GetOrCreateBackoff(It.IsAny<string>())).Returns<string>(s => _loggers[s]);
```
CooperLink: @pragnagopa  with the latest updates to remove the cache of file loggers I can't replace this code in the unit tests. 
Prior to the changes, loggers were never directly referenced.  They were passed as an object from the cache into the WriteEvent function.  Since they were behind a dictionary they weren't directly referenced other than as a var. 

I don't really understand what I need to do to replace the file loggers with Mock File Loggers now that the cache is no longer present.

pragnagopa: @CooperLink - see https://github.com/Azure/azure-functions-host/pull/9070/files - had to introduce interfaces to make the code more unit test friendly , if looks good,

1. Merged the PR into this branch
2. Rebase your current PR on latest in dev
3. Verify new code has unit test coverage

CooperLink: Rebased and verified.

balag0 (approved):

CooperLink:
Failing the Nodee2e test due to Fixture.Host.GetScriptHostLogMessages appears to have changed. I didn't do anything related to the code here I don't think. Locally I am failing many more E2e tests but succeeding at my added unit tests.  Seeking some guidance from host people about this test
https://github.com/Azure/azure-functions-host/blob/dev/test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/NodeEndToEndTests.cs#L208

pragnagopa (approved):

pragnagopa:
@fabiocav / @mathewc / @alrod - I have reviewed and approved this PR. Please take a final pass.

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
```diff
@@ -67,20 +85,20 @@ public LinuxAppServiceEventGenerator(LinuxAppServiceFileLoggerFactory loggerFact
         public override void LogFunctionExecutionEvent(string executionId, string siteName, int concurrency, string functionName,
             string invocationId, string executionStage, long executionTimeSpan, bool success)
         {
-            var logger = _loggerFactory.GetOrCreate(FunctionsExecutionEventsCategory);
             string currentUtcTime = DateTime.UtcNow.ToString();
-            if (FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagEnableLinuxEPExecutionCount))
+            bool detailedExecutionEventsDisabled = _functionsHostingConfigOptions.Value.DisableLinuxAppServiceExecutionDetails;
```
fabiocav, line 89: nit: Once resolved, this value won't changed, so storing this result would be more efficient here as it would avoid the overhead of the method calls and dictionary lookup.

This assuming we're OK with this not changing after initialization. If we want to react to hosting configuration changes, we'd need to change from `IOptions<FunctionsHostingConfigOptions>` to `IOptionsMonitor<FunctionsHostingConfigOptions>`

fabiocav (approved):
Small nit

Outside of the scope of this PR, but it would be good to consider an enhancement to the logger so it's more efficient and auto-detects when it needs to process events.

mathewc (approved):

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
pragnagopa: Thanks @fabiocav ! Tagging @CooperLink to address this and 

> Outside of the scope of this PR, but it would be good to consider an enhancement to the logger so it's more efficient and auto-detects when it needs to process events.

in a separate PR!

PR #8895, opened 2022-11-07
Capture the property bag from metadata request and pass it to Function load request

### LINKED ISSUE #8302: Add 'programming model type' configuration setting to 'function.json'

The Python Functions team is in the process of bringing the new programming model to public preview, and it will require support with the Azure portal. We had a chat with Portal folks to enable support for showing PyStein code on the portal, and one of the things that came out from the discussion was to differentiate legacy vs PyStein code patterns (and directory structures).

The requirement is to make it easier to identify new vs old apps by using a configuration setting. Our hope is when _handle_functions_metadata_request, and it is returned successfully, the host adds it to the functions config that is sent back to GetFunctions rest call. This change will help the portal to switch the showing mechanism to other different languages as well, which won't have a function.json to get to, and this the vfs​ calls would fail as well.

We have been brainstorming and collaborating with the Portal team on other solutions, and felt this solution may be a straightforward way to go, and potentially reusable for other scenarios.

### LINKED ISSUE #8898: Provide a way to pass data between multiple worker process for stein
As part of work for project stein. The function host send a metadata request to the worker process to get the function metadata. That worker would index functions and send the function metadata response. This request is only sent to one worker, however there is indexing related data that needs to be shared between workers. 

Proposed solution.
1. Worker sends the data that it needs to share between workers as part of the function metadata response with a prefix "worker."
2. The host reads the respective properties and includes the properties with prefix "worker." in the function load request.

soninaren:
resolves #8302
resolves #8898

### Pull request checklist

### Additional information

Additional PR information
As part of work for project stein. The function host send a metadata request to the worker process to get the function metadata. That worker would index functions and send the function metadata response. This request is only sent to one worker, however there is indexing related data that needs to be shared between workers. 

Proposed solution.
1. Worker sends the data that it needs to share between workers as part of the function metadata response with a prefix "worker."
2. The host reads the properties from the RpcFunctionMetadata and includes the properties in the function load request.

### DIFF
--- a/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
@@ -528,10 +528,9 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
 
         internal FunctionLoadRequest GetFunctionLoadRequest(FunctionMetadata metadata, ManagedDependencyOptions managedDependencyOptions)
         {
-            var functionId = metadata.GetFunctionId();
             FunctionLoadRequest request = new FunctionLoadRequest()
             {
-                FunctionId = functionId,
+                FunctionId = metadata.GetFunctionId(),
                 Metadata = new RpcFunctionMetadata()
                 {
                     Name = metadata.Name,
@@ -558,7 +557,7 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
             foreach (var property in metadata.Properties)
             {
                 // worker properties are expected to be string values
-                request.Metadata.Properties.Add(property.Key, property.Value.ToString());
+                request.Metadata.Properties.Add(property.Key, property.Value?.ToString());
             }
 
             return request;
@@ -733,7 +732,7 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
                     {
                         if (!functionMetadata.Properties.TryAdd(property.Key, property.Value.ToString()))
                         {
-                            _workerChannelLogger?.LogDebug("{metadataPropertyKey} is already a part of metadata properties", property.Key);
+                            _workerChannelLogger?.LogDebug("{metadataPropertyKey} is already a part of metadata properties for {functionId}", property.Key, metadata.FunctionId);
                         }
                     }
 
--- a/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
@@ -730,7 +730,7 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
 
                     foreach (var property in metadata.Properties)
                     {
-                        if (!functionMetadata.Properties.TryAdd(property.Key, property.Value.ToString()))
+                        if (!functionMetadata.Properties.TryAdd(property.Key, property.Value?.ToString()))
                         {
                             _workerChannelLogger?.LogDebug("{metadataPropertyKey} is already a part of metadata properties for {functionId}", property.Key, metadata.FunctionId);
                         }
--- a/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
@@ -561,8 +561,6 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
                 request.Metadata.Properties.Add(property.Key, property.Value.ToString());
             }
 
-            _workerChannelLogger?.LogDebug($"Adding {request.Metadata.Properties.Count} worker properties for {functionId}");
-
             return request;
         }
 
@@ -733,7 +731,10 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
 
                     foreach (var property in metadata.Properties)
                     {
-                        functionMetadata.Properties.TryAdd(property.Key, property.Value.ToString());
+                        if (!functionMetadata.Properties.TryAdd(property.Key, property.Value.ToString()))
+                        {
+                            _workerChannelLogger?.LogDebug("{metadataPropertyKey} is already a part of metadata properties", property.Key);
+                        }
                     }
 
                     var bindings = new List<string>();
--- a/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
@@ -528,9 +528,10 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
 
         internal FunctionLoadRequest GetFunctionLoadRequest(FunctionMetadata metadata, ManagedDependencyOptions managedDependencyOptions)
         {
+            var functionId = metadata.GetFunctionId();
             FunctionLoadRequest request = new FunctionLoadRequest()
             {
-                FunctionId = metadata.GetFunctionId(),
+                FunctionId = functionId,
                 Metadata = new RpcFunctionMetadata()
                 {
                     Name = metadata.Name,
@@ -553,6 +554,15 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
 
                 request.Metadata.Bindings.Add(binding.Name, bindingInfo);
             }
+
+            foreach (var property in metadata.Properties)
+            {
+                // worker properties are expected to be string values
+                request.Metadata.Properties.Add(property.Key, property.Value.ToString());
+            }
+
+            _workerChannelLogger?.LogDebug($"Adding {request.Metadata.Properties.Count} worker properties for {functionId}");
+
             return request;
         }
 
@@ -720,6 +730,11 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
                     };
 
                     functionMetadata.SetFunctionId(metadata.FunctionId);
+
+                    foreach (var property in metadata.Properties)
+                    {
+                        functionMetadata.Properties.TryAdd(property.Key, property.Value.ToString());
+                    }
 
                     var bindings = new List<string>();
                     foreach (string binding in metadata.RawBindings)
--- a/src/WebJobs.Script.Grpc/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
+++ b/src/WebJobs.Script.Grpc/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
@@ -326,7 +326,7 @@ message RpcFunctionMetadata {
   // Properties for function metadata
   // They're usually specific to a worker and largely passed along to the controller API for use
   // outside the host
-   map<string,string> Properties = 16;
+   map<string,string> properties = 16;
 }
 
 // Host tells worker it is ready to receive metadata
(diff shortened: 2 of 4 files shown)


mathewc:
PR description doesn't include any details or reference any bug for context. Please provide this info so reviewers know what you're trying to accomplish. I.e. fill out the template.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -553,6 +554,19 @@ internal FunctionLoadRequest GetFunctionLoadRequest(FunctionMetadata metadata, M
 
                 request.Metadata.Bindings.Add(binding.Name, bindingInfo);
             }
+
+            foreach (var property in metadata.Properties)
+            {
+                // Only add worker properties
```
mathewc: Why only worker properties? What is the source of the properties you're trying to flow?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -553,6 +554,19 @@ internal FunctionLoadRequest GetFunctionLoadRequest(FunctionMetadata metadata, M
...
+                if (property.Key.StartsWith("worker.", StringComparison.CurrentCultureIgnoreCase))
+                {
+                    // worker properties are expected to be string values
+                    request.Metadata.Properties.Add(property.Key, property.Value.ToString());
+                }
+            }
+
+            _workerChannelLogger?.LogDebug($"Adding {request.Metadata.Properties.Count} worker properties for {functionId}");
```
mathewc: Why is this log is needed. We don't log for all the other bits we're adding like bindings, etc.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
soninaren: Apologies for missing description in the PR. Going to add detailed description and the issue i am trying to address.
Worker properties originate from language workers in the metadata response. The purpose of this PR is to pass those properties to all instances of language workers via load request.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
soninaren: I wanted to have a trace of properties carried from metadata response to the load request. One way to identify the data passed and have some sort of logging.

soninaren:
> PR description doesn't include any details or reference any bug for context. Please provide this info so reviewers know what you're trying to accomplish. I.e. fill out the template.

Apologies for not adding the details early on. I just wanted to get a build going. However, I waited too long to add the details. Going to add them ASAP.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
mathewc: Why not just include all metadata properties - why do we need filtering?

soninaren: I think the idea is to be deliberate on what properties the worker wants to pass to the load request. The worker would be adding the prefix `worker.` to the properties it wants to pass on the other workers via load request. 

cc: @madelinegordon , @fabiocav, was there another reason to only include properties with `worker.` prefix?

madelinegordon: I think the reason it's only for worker properties is that the requirement in #8302 is from the portal team, for the purpose of being able to switch the view to other languages. But I'm not 100% sure that's the reason.

fabiocav: I don't think we need to filter those, but I believe the original ask was to have the prefix removed before sending it back. As-is, this is something the worker needs to be aware of, as if the worker adds a property named `foo` it would roundtrip with the name `worker.foo`

soninaren: @fabiocav I am a little confused by your comment. We are not adding any prefix to the properties. Just roundtripping the properties one with prefix included.

In any case based on the comments it looks like we don't really need to filter the properties with prefix. I had a chat with @gavin-aguiar, looks like they were not expecting the prefix either. I am going to make a change so that we round-trip all properties As-Is.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -553,6 +554,15 @@ internal FunctionLoadRequest GetFunctionLoadRequest(FunctionMetadata metadata, M
...
                 request.Metadata.Bindings.Add(binding.Name, bindingInfo);
             }
+
+            foreach (var property in metadata.Properties)
+            {
+                // worker properties are expected to be string values
+                request.Metadata.Properties.Add(property.Key, property.Value.ToString());
+            }
+
+            _workerChannelLogger?.LogDebug($"Adding {request.Metadata.Properties.Count} worker properties for {functionId}");
```
jviau: Please use structured logging

src/WebJobs.Script.Grpc/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
```diff
@@ -326,7 +326,7 @@ message RpcFunctionMetadata {
   // Properties for function metadata
   // They're usually specific to a worker and largely passed along to the controller API for use
   // outside the host
-   map<string,string> Properties = 16;
+   map<string,string> properties = 16;
```
jviau, line 329: Any breaking change concern here? I am not sure what release state this is already in or gRPC specifics.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
mathewc: In my mind, this log would just be noise, of limited/no utility. Why would logging the count of metadata properties help us diagnose anything? My vote would be to just remove this.

src/WebJobs.Script.Grpc/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
soninaren: I actually pulled this change from protobuf repo. It is added as a subtree in the repo. This is technically a breaking change. The change was actually submitted by the python team and was the only team using at the moment. We were able to pick up this fix this because of the nature of the change and limited exposure. I consulted with fabio before pulling in this change.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
soninaren: done

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
soninaren: Have removed the log statement as it wasn't adding much value

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -553,6 +554,13 @@ internal FunctionLoadRequest GetFunctionLoadRequest(FunctionMetadata metadata, M
 
                 request.Metadata.Bindings.Add(binding.Name, bindingInfo);
             }
+
+            foreach (var property in metadata.Properties)
+            {
+                // worker properties are expected to be string values
+                request.Metadata.Properties.Add(property.Key, property.Value.ToString());
```
jviau: Can the value be `null` here?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -528,9 +528,10 @@ internal void SendFunctionLoadRequest(FunctionMetadata metadata, ManagedDependen
 
         internal FunctionLoadRequest GetFunctionLoadRequest(FunctionMetadata metadata, ManagedDependencyOptions managedDependencyOptions)
         {
+            var functionId = metadata.GetFunctionId();
             FunctionLoadRequest request = new FunctionLoadRequest()
             {
-                FunctionId = metadata.GetFunctionId(),
+                FunctionId = functionId,
```
jviau: Any reason for this change? Or just style preference?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -528,9 +528,10 @@ internal void SendFunctionLoadRequest(FunctionMetadata metadata, ManagedDependen
 
         internal FunctionLoadRequest GetFunctionLoadRequest(FunctionMetadata metadata, ManagedDependencyOptions managedDependencyOptions)
         {
+            var functionId = metadata.GetFunctionId();
```
jviau: nit: if this is `string`, please us language types over `var`
```suggestion
            string functionId = metadata.GetFunctionId();
```

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -721,6 +729,14 @@ internal void ProcessFunctionMetadataResponses(FunctionMetadataResponse function
 
                     functionMetadata.SetFunctionId(metadata.FunctionId);
 
+                    foreach (var property in metadata.Properties)
+                    {
+                        if (!functionMetadata.Properties.TryAdd(property.Key, property.Value.ToString()))
+                        {
+                            _workerChannelLogger?.LogDebug("{metadataPropertyKey} is already a part of metadata properties", property.Key);
```
jviau: Your original log had `functionId` as a property also - that does seem fairly useful to me. Do you think it is worth including?

jviau (approved):
Minor comments

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
soninaren: it is possible for it to be null, will add a null check

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
soninaren: i had done this originally for the log statement, but with that removed this became irrelevant. Going to revert this.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
soninaren: good point, let me include that as well

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -721,6 +728,14 @@ internal void ProcessFunctionMetadataResponses(FunctionMetadataResponse function
 
                     functionMetadata.SetFunctionId(metadata.FunctionId);
 
+                    foreach (var property in metadata.Properties)
+                    {
+                        if (!functionMetadata.Properties.TryAdd(property.Key, property.Value.ToString()))
```
mathewc: in the code above you check Value for null. Do you need one here as well?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
soninaren: have reverted this change

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
soninaren: yes, have added one here as well

mathewc (approved):

PR #8900, opened 2022-11-09
Implementing selective loading for stein

### LINKED ISSUE #8825: Investigate the ordering of Init request and Metadata request in multi-worker scenarios
Ideally, it is expected that worker will receive `WorkerInitRequest` first and then `FunctionMetadataRequest` (in case of Stein) or `FunctionLoadRequest` (legacy scenario). However, by random chance, it seems the second worker may receive a function load request or a metadata request as its first request. 

Need to investigate this. Ideally, these messages should be in predictable order (where `WorkerInitRequest` will be the first one).

#### Case A
 
Worker 1 gets an init request and then metadata request (as expected). However, worker 2 receives a function load request before an init request.
 
[2022-10-06T02:42:39.129Z] Worker 71780151-7b5a-4152-af87-95168dc0aacf received WorkerInitRequest
[2022-10-06T02:42:39.464Z] Worker 71780151-7b5a-4152-af87-95168dc0aacf received FunctionsMetadataRequest
[2022-10-06T02:42:39.514Z] Worker 71780151-7b5a-4152-af87-95168dc0aacf received FunctionLoadRequest
[2022-10-06T02:42:39.878Z] Worker bffcf551-f64f-48c7-808d-fba0659b37a1 received FunctionLoadRequest
[2022-10-06T02:42:39.886Z] Worker bffcf551-f64f-48c7-808d-fba0659b37a1 received WorkerInitRequest
 
#### Case B
 
Worker 1 is initialized, but worker 2 receives the metadata request before it is initialized.
 
[2022-10-06T02:36:12.135Z] Worker 554e325b-3cef-4db0-8626-6a32a4a67e7e received WorkerInitRequest
[2022-10-06T02:36:12.737Z] Worker d43a5b45-cb3e-45a8-ba89-a53548271186 received FunctionsMetadataRequest
[2022-10-06T02:36:12.741Z] Worker d43a5b45-cb3e-45a8-ba89-a53548271186 received WorkerInitRequest

### LINKED ISSUE #8870: Implement selective loading of extensions in Host in case of Stein (WorkerIndexing)
This issue tracks implementation of selective loading of extensions in case of WorkerIndexing - [Design doc](https://dev.azure.com/msazure/One/_git/AAPT-Antares-Docs?version=GBmaster&path=/TeamDocs/FunctionTeamDocs/Design/Host/SelectiveLoadingOfExtensions.md) 

Corresponding design issue for reference - https://github.com/Azure/azure-functions-host/issues/8614

soninaren:
On a high level, This PR allows us to perform selective loading of extensions where extension bundle is being used along with stein programming model. 

More detailed description to follow

resolves #8870
resolves #8825

### Pull request checklist

#### WithindexingEnabled = true

- [x] Stein payload
- [x] Stein payload with specialization
- [x] Stein payload worker restart happening as expected due to timeout *
- [x] Stein payload with a specific trigger (should load only that extension)
- [x] Stein payload without bundles in Host.json
- [x] Stein payload with multiple workers processes
- [x] Non-Stein payload
- [x] Non-Stein payload with specialization
- [x] Non-Stein payload with multiple workers processes
- [x] Non-Stein payload worker restart happening as expected due to timeout
- [x] Non-Stein payload with a specific trigger (should load only that extension)
- [x] Non-Stein payload without bundles in Host.json

#### WithindexingEnabled = false

- [x] Non-Stein payload
- [x] Non-Stein payload with specialization
- [x] Non-Stein payload with multiple workers processes
- [x] Non-Stein payload worker restart happening as expected due to timeout
- [x] Non-Stein payload with a specific trigger (should load only that extension)
- [x] Non-Stein payload without bundles in Host.json


src/WebJobs.Script/Host/ScriptHost.cs
```diff
@@ -371,19 +363,12 @@ private void LogHostFunctionErrors()
...
-                _logger.LogInformation("Worker indexing is enabled");
-                functionMetadata = _functionMetadataManager.GetFunctionMetadata(forceRefresh: false, dispatcher: _functionDispatcher);
-            }
-            else
-            {
-                functionMetadata = _functionMetadataManager.GetFunctionMetadata(false);
-                _workerRuntime = _workerRuntime ?? Utility.GetWorkerRuntime(functionMetadata);
-            }
+
+            functionMetadata = _functionMetadataManager.GetFunctionMetadata(forceRefresh: false);
```
surgupta-msft, line 370: Should it be `True` as we discussed in design sync?

test/WebJobs.Script.Tests/AggregateFunctionMetadataProviderTests.cs
```diff
@@ -1,326 +0,0 @@
...
-            var ex = Assert.Throws<FormatException>(() =>
-            {
-                AggregateFunctionMetadataProvider.ValidateBindings(rawBindings, functionMetadata);
-            });
-
-            Assert.Equal("At least one binding must be declared.", ex.Message);
-        }
-
-        [Fact]
-        public void ValidateBindings_DuplicateBindingNames_Throws()
```
surgupta-msft, line 54: Do we have plans to add these tests in a different class? It seems we are removing all these tests.

src/WebJobs.Script/Host/FunctionMetadataManager.cs
```diff
@@ -132,9 +133,22 @@ internal ImmutableArray<FunctionMetadata> LoadFunctionMetadata(bool forceRefresh
...
 
-            IFunctionMetadataProvider metadataProvider = new AggregateFunctionMetadataProvider(_loggerFactory.CreateLogger<AggregateFunctionMetadataProvider>(), dispatcher, _functionMetadataProvider, _scriptOptions);
+            bool workerIndexing = Utility.CanWorkerIndex(workerConfigs, SystemEnvironment.Instance);
+            IFunctionMetadataProvider functionMetadataProvider = null;
 
-            immutableFunctionMetadata = metadataProvider.GetFunctionMetadataAsync(workerConfigs, SystemEnvironment.Instance, forceRefresh).GetAwaiter().GetResult();
+            if (workerIndexing)
+            {
+                functionMetadataProvider = _functionMetadataProviders.Where(p => p.GetType() == typeof(WorkerFunctionMetadataProvider)).FirstOrDefault();
+                immutableFunctionMetadata = functionMetadataProvider.GetFunctionMetadataAsync(workerConfigs, SystemEnvironment.Instance, forceRefresh).GetAwaiter().GetResult();
```
surgupta-msft: This will only work when specialization has happened before this step, so we have Customer payload available at this point. Is this correct? How can we make sure that it is happening in code?

src/WebJobs.Script/Host/ScriptHost.cs
soninaren: We can leave this as false as ScriptStartupTypeLocator has set it to true. That piece of code runs everytime script host starts

src/WebJobs.Script/Host/FunctionMetadataManager.cs
soninaren: It will actually work before or after specialization. The only difference is that `workerIndexing` will only return true after specialization. ( `false` once we flip the feature flag). It ends up getting called at both times, before and after specialization. We clear the cache after specialization, which gives us actual customer payload.

src/WebJobs.Script/Host/FunctionMetadataResult.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Collections.Immutable;
+using Microsoft.Azure.WebJobs.Script.Description;
+
+namespace Microsoft.Azure.WebJobs.Script
+{
+    public class FunctionMetadataResult
```
jviau: Is this assembly referenced directly by customers? If so, I recommend starting with `internal` for all types and then only going to `public` if absolutely necessary.

src/WebJobs.Script/Host/IHostFunctionMetadataProvider.cs
```diff
@@ -0,0 +1,18 @@
...
+
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Script.Description;
+using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
+
+namespace Microsoft.Azure.WebJobs.Script
+{
+    public interface IHostFunctionMetadataProvider
```
jviau: Are you intending for `DefaultFunctionMetadataProvider` to take in this interface and not the exact type?

src/WebJobs.Script/Host/IWorkerFunctionMetadataProvider.cs
```diff
@@ -0,0 +1,17 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
+
+namespace Microsoft.Azure.WebJobs.Script
+{
+    public interface IWorkerFunctionMetadataProvider
```
jviau: Same with the other comment - recommend making this and other types `internal` if possible.

jviau (approved):
Looks good! Recommend minimizing public API surface as much as possible.

test/WebJobs.Script.Tests/AggregateFunctionMetadataProviderTests.cs
soninaren: yes, going to add these tests back.

src/WebJobs.Script/Host/IHostFunctionMetadataProvider.cs
soninaren: yes, had forgotten to make that change. its done now.

src/WebJobs.Script/Host/FunctionMetadataResult.cs
soninaren: done

surgupta-msft:
As discussed offline, it will be helpful if you can update PR description with the specific scenarios you tested manually like timeouts, etc as this PR is touching lot of important components.

Based on my experience and previously seen issues, below are some important scenarios to validate -

- Placeholder/Specialization - using appsetting `"WEBSITE_PLACEHOLDER_MODE" : 1`
- LinuxConsumption path
- Coretools - using appsetting `"FUNCTIONS_CORETOOLS_ENVIRONMENT": "true"`
- Timeouts 
- Extension bundles - with and without bundles in Host.json
- Deploying a test stamp using private site extension and check for any exceptions/noise/timeouts, etc

soninaren:
> As discussed offline, it will be helpful if you can update PR description with the specific scenarios you tested manually like timeouts, etc as this PR is touching lot of important components.

Adding a list in the description. Will update the status as i finish running through the scenarios

src/WebJobs.Script/Host/DefaultFunctionMetadataProvider.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Collections.Immutable;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Script.Description;
+using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost
+{
+    internal class DefaultFunctionMetadataProvider : IFunctionMetadataProvider
```
surgupta-msft: What do we mean by `default` in the naming here? This class is doing same thing as `AggregateFunctionMetadataProvider.cs`  class we had earlier which meant to working with both objects - `IWorkerFunctionMetadataProvider` and `IHostFunctionMetadataProvider` same as this class is doing. But, I feel _Aggregate_ makes more sense than _default_ to refer to combination of both providers.

We introduced the term `UseDefaultMetadataIndexing` to refer to Host indexing as default.

src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
```diff
@@ -7,90 +7,108 @@
...
-                        throw new InvalidOperationException(nameof(_dispatcher));
-                    }
-
-                    // start up GRPC channels
-                    await _dispatcher.InitializeAsync(new List<FunctionMetadata>());
+                    throw new InvalidOperationException(nameof(_channelManager));
+                }
 
-                    // get function metadata from worker, then validate it
-                    rawFunctions = await _dispatcher.GetWorkerMetadata();
```
surgupta-msft, line 68: Looks like we are bypassing the dispatcher completely and getting Metadata directly from Channel. In that case, we can remove GetFunctionMetadata() from RPCFunctionInvocationDispatcher - https://github.com/Azure/azure-functions-host/blob/960fcfcada10c237acd044e0190afdc73ba1b747/src/WebJobs.Script/Workers/IFunctionInvocationDispatcher.cs#L30

as it was specifically introduced in Anjana's PR to get Metadata from Dispatcher - https://github.com/Azure/azure-functions-host/commit/7e52075f6377530ddc0970e38cff1625816f2bed

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -384,38 +358,6 @@ public async Task<IEnumerable<RawFunctionMetadata>> GetWorkerMetadata()
             return (channels != null && channels.Length > 0) ? await channels.First().GetFunctionMetadata() : null;
```
surgupta-msft: Remove this method `GetWorkerMetadata()` from Dispatcher as it is not getting used anywhere.

src/WebJobs.Script/Host/IHostFunctionMetadataProvider.cs
```diff
@@ -0,0 +1,18 @@
...
+
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Script.Description;
+using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
+
+namespace Microsoft.Azure.WebJobs.Script
+{
+    internal interface IHostFunctionMetadataProvider
```
surgupta-msft, line 15: Any reason behind having 3 interfaces with very similar code? Is it possible to combine them? 

`IFunctionMetadataProvider`
`IHostFunctionMetadataProvider`
`IWorkerFunctionMetadataProvider`

src/WebJobs.Script/Host/HostFunctionMetadataProvider.cs
```diff
@@ -38,6 +38,8 @@ public HostFunctionMetadataProvider(IOptionsMonitor<ScriptApplicationHostOptions
         public ImmutableDictionary<string, ImmutableArray<string>> FunctionErrors
            => _functionErrors.ToImmutableDictionary(kvp => kvp.Key, kvp => kvp.Value.ToImmutableArray());
 
+        public bool UseDefaultMetadataIndexing => true;
```
surgupta-msft: Do we need this line here? It does not have any references.

src/WebJobs.Script/Workers/IFunctionInvocationDispatcher.cs
```diff
@@ -25,8 +25,6 @@ public interface IFunctionInvocationDispatcher : IDisposable
 
         Task<bool> RestartWorkerWithInvocationIdAsync(string invocationId);
 
-        Task FinishInitialization(IEnumerable<FunctionMetadata> functions, CancellationToken cancellationToken = default);
-
         Task<IEnumerable<RawFunctionMetadata>> GetWorkerMetadata();
```
surgupta-msft: Remove `GetWorkerMetadata()` as it is not being used.

src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
```diff
@@ -7,90 +7,108 @@
...
+                // start up GRPC channels
+                await _channelManager.InitializeChannelAsync(_workerRuntime);
```
surgupta-msft: My understanding from the design discussion is that we will use an existing channel created during placeholder mode for getting metadata. But, looks like we are initializing a new channel here. Feel free to update the [design doc](https://dev.azure.com/msazure/One/_git/AAPT-Antares-Docs?version=GBmaster&path=/TeamDocs/FunctionTeamDocs/Design/Host/SelectiveLoadingOfExtensions.md&_a=preview) if there is any change in design.

In `ScriptHost.cs`, previously we were not initializing channel if workerIndexing is true so we were doing `dispatcher.initialize()` here. But now, ScriptHost initializes channel for both legacy and Stein.

src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
```diff
@@ -7,90 +7,108 @@
...
 
-                    if (IsDefaultIndexingRequired(rawFunctions))
-                    {
-                        _logger.LogDebug("Fallback to host indexing as worker denied indexing");
-                        functions = await _hostFunctionMetadataProvider.GetFunctionMetadataAsync(workerConfigs, environment, forceRefresh);
-                    }
-                    else if (!IsNullOrEmpty(rawFunctions))
+                var channels = _channelManager.GetChannels(_workerRuntime);
```
surgupta-msft: At this point, do we always have 1 channel in `var channels`? or can there be scenarios with more than 1 channel here?

src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
```diff
@@ -7,90 +7,108 @@
...
+                foreach (string workerId in channels.Keys.ToList())
```
surgupta-msft, line 69: If there can be more than 1 channel, then will this loop run multiple times unnecessarily, even if we got FunctionsMetadata with the first iteration?

src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
```diff
@@ -7,90 +7,108 @@
...
+                {
+                    if (channels.TryGetValue(workerId, out TaskCompletionSource<IRpcWorkerChannel> initializedLanguageWorkerChannelTask))
                     {
-                        functions = ValidateMetadata(rawFunctions);
+                        _logger.LogDebug("Found initialized language worker channel for runtime: {workerRuntime} workerId:{workerId}", _workerRuntime, workerId);
+                        try
+                        {
+                            IRpcWorkerChannel channel = await initializedLanguageWorkerChannelTask.Task;
```
surgupta-msft: Since we are starting this channel here, do we need to do anything in this class to dispose the channel properly?

src/WebJobs.Script/Host/DefaultFunctionMetadataProvider.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        private readonly IEnvironment _environment;
+        private readonly ILogger<DefaultFunctionMetadataProvider> _logger;
+        private IWorkerFunctionMetadataProvider _workerFunctionMetadataProvider;
+        private IHostFunctionMetadataProvider _hostFunctionMetadataProvider;
+
+        public DefaultFunctionMetadataProvider(ILogger<DefaultFunctionMetadataProvider> logger, IWorkerFunctionMetadataProvider workerFunctionMetadataProvider, IHostFunctionMetadataProvider hostFunctionMetadataProvider)
+        {
+            _logger = logger;
+            _workerFunctionMetadataProvider = workerFunctionMetadataProvider;
```
surgupta-msft: Not null checks for `workerFunctionMetadataProvider` and `hostFunctionMetadataProvider`

src/WebJobs.Script/Host/DefaultFunctionMetadataProvider.cs
soninaren: Aggregation generally means we are combining more than one thing. However, the AggregateFunctionMetadataProvider wasn't a true aggregator. It was instead a selector. For something to be an aggregator it needs to combine results from multiple sources into one. I believe we have aggregateException that behaves in a similar manner. I do get your point though default in this case can be misleading. We can probably name it `FunctionMetadataProvider`

src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
soninaren: Done

src/WebJobs.Script/Workers/IFunctionInvocationDispatcher.cs
soninaren: done

src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
soninaren: Yes, you are right. I found a bug there in my testing. I have updated to only initialize if we don't have a single functioning worker channel.

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
soninaren: done

src/WebJobs.Script/Host/IHostFunctionMetadataProvider.cs
soninaren: We can certainly combine the interfaces into one however this is done deliberately to model the right expectations from the interfaces. Having the same interface implemented between host and worker provider would mean that we have different behavior expectations from different implementation of the same method between different types. For example, we have a type that has a fallback and another one that does not. Additionally, if we combine all of them into a single interface, then we would register 3 or at the minimum 2 types for the same interface in DI. In order to get a specific type (worker or host metadata) to be used in a class we would have to do either request IEnumerable and perform reflection to identify and that specific type. We might be able to do it in a clean way without reflection. but the key thing here is that when we do this, we are indirectly communicating that we can expect more providers to be registered in this way. However, that is not the case. Additionally doing it this way allows us to contain the logic within the default metadata provider and allows us to cleanup the overloaded MetadataManager type as well.

src/WebJobs.Script/Host/HostFunctionMetadataProvider.cs
soninaren: we don't need it. good catch.

src/WebJobs.Script/Host/DefaultFunctionMetadataProvider.cs
soninaren: done.

src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
soninaren: it doesn't look like we need to do that. I don't see us doing that in other places where we start a channel. Also, we have events that watch and manage channel lifetime as expected,

src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
soninaren: I don't think we have scenarios where we expect more than 1 channel here. For our scenario we need atleast 1 channel.

src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
soninaren: This was a great catch. We need to break the loop once we have valid metadata. I think this was copy paste error. i have fixed the logic here.

soninaren:
> As discussed offline, it will be helpful if you can update PR description with the specific scenarios you tested manually like timeouts, etc as this PR is touching lot of important components.
> 
> Based on my experience and previously seen issues, below are some important scenarios to validate -
> 
> * Placeholder/Specialization - using appsetting `"WEBSITE_PLACEHOLDER_MODE" : 1`
> * LinuxConsumption path
> * Coretools - using appsetting `"FUNCTIONS_CORETOOLS_ENVIRONMENT": "true"`
> * Timeouts
> * Extension bundles - with and without bundles in Host.json
> * Deploying a test stamp using private site extension and check for any exceptions/noise/timeouts, etc

Yes, absolutely. I have combined the scenarios i have tested and included the ones you suggested as well. Going to run the scenarios again after PR feedback commits.

src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
```diff
@@ -7,90 +7,117 @@
...
+            }
+
+            return new FunctionMetadataResult(useDefaultMetadataIndexing: false, _functions);
+        }
+
+        private bool IsDefaultIndexingRequired(IEnumerable<RawFunctionMetadata> functions)
+        {
+            foreach (RawFunctionMetadata function in functions)
+            {
+                if (function.UseDefaultMetadataIndexing == true)
```
madelinegordon: nit: could simplify to 
```suggestion
                if (function.UseDefaultMetadataIndexing)
```
or 
```suggestion
                return functions.Any(f => f.UseDefaultMetadataIndexing);
```
?

src/WebJobs.Script/Host/FunctionMetadataProvider.cs
```diff
@@ -0,0 +1,60 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Script.Description;
+using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
```
madelinegordon, line 11: nit: Is this using statement necessary?

src/WebJobs.Script/Host/IHostFunctionMetadataProvider.cs
madelinegordon: To clarify my understanding - does each metadata provider correspond to a different metadata type/class? 

If keeping separate interfaces, would it make sense to extend the specification in expectations to the methods in the interfaces too instead of just keeping the difference at the interface names? e.g. in IHostFunctionMetadataProvider, GetHostFunctionMetadataAsync instead of GetFunctionMetadataAsync?

src/WebJobs.Script/Host/FunctionMetadataProvider.cs
soninaren: `using Microsoft.Extensions.Options;` is being used

`using Microsoft.CodeAnalysis.Scripting;` is not being used, have removed that.

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
```diff
@@ -160,6 +159,14 @@ public static void AddWebJobsScriptHost(this IServiceCollection services, IConfi
             // Language Worker Hosted Services need to be intialized before WebJobsScriptHostService
             ScriptHostBuilderExtensions.AddCommonServices(services);
 
+            services.AddSingleton<IFunctionMetadataProvider>(sp =>
+            {
+                return new FunctionMetadataProvider(
+                    ActivatorUtilities.GetServiceOrCreateInstance<ILogger<FunctionMetadataProvider>>(sp),
```
jviau: `ILogger<T>` is known to be in the container, so you can use:
```suggestion
                    sp.GetRequiredService<ILogger<FunctionMetadataProvider>>(),
```

src/WebJobs.Script/Host/FunctionMetadataProvider.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        public FunctionMetadataProvider(ILogger<FunctionMetadataProvider> logger, IWorkerFunctionMetadataProvider workerFunctionMetadataProvider, IHostFunctionMetadataProvider hostFunctionMetadataProvider)
+        {
+            _logger = logger;
+            _workerFunctionMetadataProvider = workerFunctionMetadataProvider;
+            _hostFunctionMetadataProvider = hostFunctionMetadataProvider;
+            _environment = SystemEnvironment.Instance;
+        }
+
+        public ImmutableDictionary<string, ImmutableArray<string>> FunctionErrors { get; private set; }
```
jviau, line 30: Not sure what the normal behavior of this interface is, but should this be initialized to empty immutable dictionary?

I personally prefer having a never-null approach whenever possible.

src/WebJobs.Script/Host/IHostFunctionMetadataProvider.cs
jviau: The interfaces exist only for unit testing - being able to provide mocks. Other than that, this is a very targeted scenario for the host vs worker function indexing logic, which necessitates slight deviations in the contract between the two. This prevents re-use of the `IFunctionMetadataProvider` for the worker logic. The host could use it sure - and I am not against that. The worker logic cannot use it (it needs a different return type). But overall, we have a targeted scenario here and not a general extensibility point. We are weighing the cost vs benefit of creating a contract which satisfies the scenario, but also does not betray the implementation details.

The multiple interfaces may look excessive, and they may honestly be just that. But there is not too much of a benefit to solving it differently. Although it may be worth making all of these very explicit. @soninaren - can you add comments explaining this? Or alternatively, you could even nest these interfaces within `DefaultFunctionMetadataProvider` to show that they are to be used **only** for that implementation.

But also, if we feel that we should not be introducing one-off interfaces like this as a matter of principal, then I also am not against that.

src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
```diff
@@ -7,90 +7,116 @@
...
         private ImmutableArray<FunctionMetadata> _functions;
-        private IOptions<ScriptJobHostOptions> _scriptOptions;
-        private IFunctionMetadataProvider _hostFunctionMetadataProvider;
-
-        public AggregateFunctionMetadataProvider(
-            ILogger logger,
-            IFunctionInvocationDispatcher invocationDispatcher,
-            IFunctionMetadataProvider hostFunctionMetadataProvider,
-            IOptions<ScriptJobHostOptions> scriptOptions)
+        private string _workerRuntime;
```
jviau: can this be readonly?

src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
```diff
@@ -7,90 +7,116 @@
...
+                    throw new InvalidOperationException(nameof(_channelManager));
+                }
 
-                    // start up GRPC channels
-                    await _dispatcher.InitializeAsync(new List<FunctionMetadata>());
+                var channels = _channelManager.GetChannels(_workerRuntime);
 
-                    // get function metadata from worker, then validate it
-                    rawFunctions = await _dispatcher.GetWorkerMetadata();
+                if (channels == null || channels.Count() == 0)
```
jviau: alternative:
```suggestion
                if (channels?.Any() != true)
```

src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
```diff
@@ -226,18 +252,6 @@ internal static FunctionMetadata ValidateBindings(IEnumerable<string> rawBinding
...
-            {
-                if (function.UseDefaultMetadataIndexing == true)
-                {
-                    return true;
-                }
-            }
-            return false;
-        }
-
         private bool IsNullOrEmpty(IEnumerable<RawFunctionMetadata> functions)
```
jviau, line 243: If you want, this can be removed and replaced with `functions?.Any() != true` in all places.

src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
```diff
@@ -7,90 +7,116 @@
...
                 }
-                else
+            }
+
+            return new FunctionMetadataResult(useDefaultMetadataIndexing: false, _functions);
+        }
+
+        private bool IsDefaultIndexingRequired(IEnumerable<RawFunctionMetadata> functions)
+        {
+            foreach (RawFunctionMetadata function in functions)
```
jviau: I am fine with the `foreach` (technically more performant than LINQ), but you can also use `functions.Any(x => x.UseDefaultMetadataIndexing)`

jviau (approved):

src/WebJobs.Script/Workers/Http/HttpFunctionInvocationDispatcher.cs
```diff
@@ -213,14 +213,6 @@ public async Task<bool> RestartWorkerWithInvocationIdAsync(string invocationId)
...
 
-        public Task<IEnumerable<RawFunctionMetadata>> GetWorkerMetadata()
-        {
-            // Throw exception because this method is meant to support the worker indexing pipeline.
-            // This is needed in RpcFunctionInvocationDispatcher to start the GRPC process of
-            // requesting metadata from the worker and receiving metadata from the worker.
-            throw new NotSupportedException("This method is meant to support the worker indexing pipeline.");
-        }
-
         public Task FinishInitialization(IEnumerable<FunctionMetadata> functions, CancellationToken cancellationToken = default)
```
surgupta-msft: We can remove `FinishInitialization` from HostDispatcher.

src/WebJobs.Script/Host/IWorkerFunctionMetadataProvider.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
+
+namespace Microsoft.Azure.WebJobs.Script
```
surgupta-msft, line 9: Having a summary to provide class/interface description will be useful - specially for the new ones being added in the PR.

src/WebJobs.Script/Host/FunctionMetadataProvider.cs
```diff
@@ -0,0 +1,60 @@
...
+        public async Task<ImmutableArray<FunctionMetadata>> GetFunctionMetadataAsync(IEnumerable<RpcWorkerConfig> workerConfigs, IEnvironment environment, bool forceRefresh = false)
+        {
+            bool workerIndexing = Utility.CanWorkerIndex(workerConfigs, _environment);
+
+            if (!workerIndexing)
+            {
+                return await GetMetadataFromHostProvider(workerConfigs, environment, forceRefresh);
+            }
+
+            FunctionMetadataResult functionMetadataResult = await _workerFunctionMetadataProvider.GetFunctionMetadataAsync(workerConfigs, SystemEnvironment.Instance, forceRefresh);
```
surgupta-msft: We will need a null check here for `_workerFunctionMetadataProvider`.

src/WebJobs.Script/Host/FunctionMetadataProvider.cs
```diff
@@ -0,0 +1,60 @@
...
+                _logger.LogDebug("Fallback to host indexing as worker denied indexing");
+                return await GetMetadataFromHostProvider(workerConfigs, environment, forceRefresh);
+            }
+
+            return functionMetadataResult.Functions;
+        }
+
+        private async Task<ImmutableArray<FunctionMetadata>> GetMetadataFromHostProvider(IEnumerable<RpcWorkerConfig> workerConfigs, IEnvironment environment, bool forceRefresh = false)
+        {
+            var functions = await _hostFunctionMetadataProvider.GetFunctionMetadataAsync(workerConfigs, environment, forceRefresh);
```
surgupta-msft: null check for `_hostFunctionMetadataProvider`

src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
```diff
@@ -7,90 +7,116 @@
...
+                        _logger.LogDebug("Found initialized language worker channel for runtime: {workerRuntime} workerId:{workerId}", _workerRuntime, workerId);
+                        try
+                        {
+                            IRpcWorkerChannel channel = await languageWorkerChannelTask.Task;
+                            rawFunctions = await channel.GetFunctionMetadata();
+
+                            if (IsDefaultIndexingRequired(rawFunctions))
+                            {
+                                _functions.Clear();
+                                return new FunctionMetadataResult(useDefaultMetadataIndexing: true, _functions);
```
surgupta-msft, line 82: This is the only line that is causing difference in the code of two interfaces `IHostFunctionMetadataProvider` and `IWorkerFunctionMetadataProvider`, otherwise it's exactly the same code. And we are artificially making this object to return `FunctionMetadataResult`. We can do the check `IsDefaultIndexingRequired(rawFunctions)` directly in `FunctionMetadataProvider` and then there is no difference in the interfaces.

src/WebJobs.Script/Host/IHostFunctionMetadataProvider.cs
surgupta-msft: Below is my proposed structuring which I feel will solve all of the concerns above.

1. IFunctionMetadataProvider (base interface)
        - Method - `GetMetadataAsync()`
2. IHostFunctionMetadataProvider : IFunctionMetadataProvider
        - No additional methods
3. IWorkerFunctionMetadataProvider : IFunctionMetadataProvider
        - Additional Method - `IsFallbackNeededToHostIndexing(functions)`

@jviau I don't think that difference in contract is mandatory. See my comment here - https://github.com/Azure/azure-functions-host/pull/8900/files#r1030908969

madelinegordon: @jviau that makes sense! Synced with @soninaren offline and also think it makes sense to keep them separate/that comments sound like a good idea!

test/WebJobs.Script.Tests/DefaultFunctionMetadataProviderTests.cs
```diff
@@ -0,0 +1,126 @@
...
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.Extensions.Options;
+using Microsoft.WebJobs.Script.Tests;
+using Moq;
+using Xunit;
+
+namespace Microsoft.Azure.WebJobs.Script.Tests
+{
+    public class DefaultFunctionMetadataProviderTests
```
surgupta-msft: Renaming it to `FunctionMetadataProviderTests`

madelinegordon (approved):

surgupta-msft (approved):

src/WebJobs.Script/Host/FunctionMetadataProvider.cs
surgupta-msft: My main point here is - if _workerFunctionMetadataProvider is null, we should throw an exception. We used to do that previously in `AggregateFunctionMetadataProvider` for null check on dispatcher

https://github.com/Azure/azure-functions-host/blob/02b161d931e5eeae1d14dc363cafedb4be876c36/src/WebJobs.Script/Host/AggregateFunctionMetadataProvider.cs#L59-L61

soninaren: Sorry i caught this very late. The code flow there was actually very different that warranted a null check. We have a lot of code paths calling the method, some with dispatcher being null others not. In this case though it is populated via DI and we almost never expect null here. I will take a look at it regardless

PR #8902, opened 2022-11-10
Add unit test for the SDK binding feature

liliankasem:
Adding unit for the SDK-binding feature. Related PRs can be found in the [epic](https://github.com/Azure/azure-functions-dotnet-worker/issues/1081#issuecomment-1276566098).

### DIFF
--- a/test/WebJobs.Script.Tests/Binding/GeneralScriptBindingProviderTests.cs
+++ b/test/WebJobs.Script.Tests/Binding/GeneralScriptBindingProviderTests.cs
@@ -21,27 +21,33 @@ namespace Microsoft.Azure.WebJobs.Script.Tests
         [InlineData(null, null, typeof(object))]
         [InlineData(null, "many", typeof(object[]))]
         [InlineData("string", null, typeof(string))]
-        [InlineData("StRing", null, typeof(string))] // case insenstive
-        [InlineData("string", "mANy", typeof(string[]))] // case insensitve
+        [InlineData("StRing", null, typeof(string))] // case insensitive
+        [InlineData("string", "mANy", typeof(string[]))] // case insensitive
         [InlineData("binary", null, typeof(byte[]))]
         [InlineData("binary", "many", typeof(byte[][]))]
         [InlineData("stream", null, typeof(Stream))]
         [InlineData("stream", "many", typeof(Stream[]))] // nonsense?
-        public void Validate(string dataType, string cardinality, Type expectedType)
+        [InlineData("string", null, typeof(ParameterBindingData), true)]
+        [InlineData("string", "many", typeof(ParameterBindingData[]), true)]
+        [InlineData(null, null, typeof(ParameterBindingData), true)]
+        [InlineData(null, "many", typeof(ParameterBindingData[]), true)]
+        public void Validate(string dataType, string cardinality, Type expectedType, bool supportsDeferredBinding = false)
         {
-            var ctx = New(dataType, cardinality);
+            var ctx = New(dataType, cardinality, supportsDeferredBinding);
             var type = GeneralScriptBindingProvider.GetRequestedType(ctx);
             Assert.Equal(expectedType, type);
         }
 
-        private static ScriptBindingContext New(string dataType, string cardinality)
+        private static ScriptBindingContext New(string dataType, string cardinality, bool supportsDeferredBinding)
         {
-            var jobj = new JObject();
-            jobj["type"] = "test";
-            jobj["direction"] = "in";
-            jobj["datatype"] = dataType;
-            jobj["cardinality"] = cardinality;
-            return new ScriptBindingContext(jobj);
+            var bindingMetadataJObject = new JObject();
+            bindingMetadataJObject["type"] = "test";
+            bindingMetadataJObject["direction"] = "in";
+            bindingMetadataJObject["datatype"] = dataType;
+            bindingMetadataJObject["cardinality"] = cardinality;
+            bindingMetadataJObject["properties"] = new JObject { { "supportsDeferredBinding", supportsDeferredBinding } };
+
+            return new ScriptBindingContext(bindingMetadataJObject);
         }
 
         [Fact]
--- a/tools/ExtensionsMetadataGenerator/test/ExtensionsMetadataGeneratorTests/ExtensionsMetadataGeneratorTests.csproj
+++ b/tools/ExtensionsMetadataGenerator/test/ExtensionsMetadataGeneratorTests/ExtensionsMetadataGeneratorTests.csproj
@@ -11,7 +11,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.5" />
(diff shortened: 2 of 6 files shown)


test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SdkTypeBindings/DotNetIsolated_BlobStorage.cs
```diff
@@ -0,0 +1,241 @@
...
+
+            var response = await SamplesTestHelpers.InvokeHttpTrigger(Fixture, "BlobInputBookFunction");
+            var responseContent = await response.Content.ReadAsStringAsync();
+
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+            Assert.Equal("Golden Compass", responseContent);
+        }
+
+        [Fact]
+        public async Task BlobTrigger_BlobClientFunction_Invoke_Succeeds()
```
liliankasem: All of the BlobTrigger tests pass when run individually but are flaky when the entire class runs all the tests together - the issue being that no logs are found. Looking for some suggestions to either make these 5 tests run one after the other or an alternative way to validate these BlobTrigger functions

sample/DotNetIsolated/Blob/BlobTriggerBindingSamples.cs
```diff
@@ -0,0 +1,64 @@
...
+using System.IO;
+using System.Text;
+using System.Threading.Tasks;
+using Azure.Storage.Blobs;
+using Microsoft.Azure.Functions.Worker;
+using Microsoft.Extensions.Logging;
+
+namespace SampleApp
+{
+    public static class BlobTriggerBindingSamples
```
madelinegordon: Are these sample functions covering different scenarios from the ones in BlobInputBindingSamples.cs?

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SdkTypeBindings/DotNetIsolated_BlobStorage.cs
madelinegordon: Are they failing when just running the entire class locally or from inside the integration test pipeline?

tools/ExtensionsMetadataGenerator/test/ExtensionsMetadataGeneratorTests/ExtensionsMetadataGeneratorTests.csproj
```diff
@@ -11,7 +11,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.5" />
+    <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.35" />
```
madelinegordon: Is there a reason for this referenced version downgrade?

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SdkTypeBindings/DotNetIsolated_BlobStorage.cs
madelinegordon: Are they failing when just running the entire class of tests locally or from inside the integration tests pipeline?

tools/ExtensionsMetadataGenerator/test/ExtensionsMetadataGeneratorTests/ExtensionsMetadataGeneratorTests.csproj
liliankasem: I read that as 3.0.05, so I thought this would be an upgrade. I don't think WebJobs 3.0.5  as in (50) exists?

sample/DotNetIsolated/Blob/BlobTriggerBindingSamples.cs
liliankasem: Yup here we're testing the BlobTrigger binding, the other ones are testing the input binding

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SdkTypeBindings/DotNetIsolated_BlobStorage.cs
liliankasem: Locally, but could also be an issue in the pipelines - trying to get the pipeline to work to check there too. Either way I want to make sure I don't merge any flaky tests

surgupta-msft: Looks like xunit supports something like below for sequential execution. Source - [SO Link](https://stackoverflow.com/questions/1408175/execute-unit-tests-serially-rather-than-in-parallel), [Link](https://medium.com/selectstarfromweb/xunit-tests-in-parallel-ad32788ce1bd). There are more suggestions in SO link.

```csharp
namespace IntegrationTests
{
    [Collection("Sequential")]
    public class Class1
    {
        [Fact]
        public void Test1() { }

        [Fact]
        public void Test2() { }
    }
}

```

liliankasem: My understanding from that SO discussion is that all of the individual tests in the the class will still run in parallel, but the classes will run in sequence - meaning I would need to create a class for each test for them to run in sequence?

Edit: Hmmm, I think I misunderstood. "Each test class is a unique test collection and tests under it will run in sequence" - so in theory if I add collection it will turn that class into a sequence run test class - let me try it out!

test/WebJobs.Script.Tests/Binding/GeneralScriptBindingProviderTests.cs
```diff
@@ -21,27 +21,33 @@ public class GeneralScriptBindingProviderTests
...
         [InlineData(null, "many", typeof(object[]))]
         [InlineData("string", null, typeof(string))]
-        [InlineData("StRing", null, typeof(string))] // case insenstive
-        [InlineData("string", "mANy", typeof(string[]))] // case insensitve
+        [InlineData("StRing", null, typeof(string))] // case insensitive
+        [InlineData("string", "mANy", typeof(string[]))] // case insensitive
         [InlineData("binary", null, typeof(byte[]))]
         [InlineData("binary", "many", typeof(byte[][]))]
         [InlineData("stream", null, typeof(Stream))]
         [InlineData("stream", "many", typeof(Stream[]))] // nonsense?
```
madelinegordon, line 29: nit - what does the question in the comment  (line 29) indicate? not sure how to interpret just first-pass reading over it

madelinegordon (approved):

test/WebJobs.Script.Tests/Binding/GeneralScriptBindingProviderTests.cs
liliankasem: I didn't write the original test so I'm not sure what the original author intended here? Potentially an example of a nonsense case?

surgupta-msft (approved):

PR #8915, opened 2022-11-16
Distributed Tracing for Java Azure Functions - Duplicate logs in AI issue

RohitRanjanMS:
Java worker has an option to use an ApplicationInsights agent to send logs and traces to AI. If the worker agent is enabled (APPLICATIONINSIGHTS_ENABLE_AGENT) then the host won't send logs to AI and Quick pulse service. 
[Distributed Tracing for Java Azure Functions](https://github.com/Azure/azure-functions-java-worker/wiki/Distributed-Tracing-for-Java-Azure-Functions)

### Pull request checklist

* [] I have added all required tests (Unit tests, E2E tests)

### DIFF
--- a/src/WebJobs.Script/ScriptHostBuilderExtensions.cs
+++ b/src/WebJobs.Script/ScriptHostBuilderExtensions.cs
@@ -419,7 +419,7 @@ namespace Microsoft.Azure.WebJobs.Script
                     {
                         options.AddFilter<ApplicationInsightsLoggerProvider>((category, logLevel) =>
                         {
-                            // match Function.<FunctionName>.User
+                            // match Function.<FunctionName>.User category
                             if (!string.IsNullOrEmpty(category) && category.Length > 14 && category.EndsWith(".User", StringComparison.Ordinal))
                             {
                                 return false;
--- a/src/WebJobs.Script/ScriptHostBuilderExtensions.cs
+++ b/src/WebJobs.Script/ScriptHostBuilderExtensions.cs
@@ -420,7 +420,7 @@ namespace Microsoft.Azure.WebJobs.Script
                         options.AddFilter<ApplicationInsightsLoggerProvider>((category, logLevel) =>
                         {
                             // match Function.<FunctionName>.User
-                            if (category.Length > 14 && category.EndsWith(".User", StringComparison.Ordinal))
+                            if (!string.IsNullOrEmpty(category) && category.Length > 14 && category.EndsWith(".User", StringComparison.Ordinal))
                             {
                                 return false;
                             }
--- a/src/WebJobs.Script/ScriptHostBuilderExtensions.cs
+++ b/src/WebJobs.Script/ScriptHostBuilderExtensions.cs
@@ -405,22 +405,24 @@ namespace Microsoft.Azure.WebJobs.Script
                     });
                 }
 
-                if (SystemEnvironment.Instance.IsApplicationInsightsAgentEnabled())
+                builder.Services.AddOptions<LoggerFilterOptions>().Configure<IEnvironment>((options, environment) =>
                 {
-                    // Skip sending user generated logs to AI and QuickPulse if worker AI agent is configured, worker will send these logs to AI and Quickpulse service.
-                    builder.Services.Configure<LoggerFilterOptions>(options => options.AddFilter<ApplicationInsightsLoggerProvider>((category, logLevel) =>
+                    if (environment.IsApplicationInsightsAgentEnabled())
                     {
-                        // match Function.<FunctionName>.User
-                        if (category.Length > 14 && category.EndsWith(".User", StringComparison.Ordinal))
+                        options.AddFilter<ApplicationInsightsLoggerProvider>((category, logLevel) =>
                         {
-                            return false;
-                        }
-                        else
-                        {
-                            return true;
-                        }
-                    }));
-                }
+                            // match Function.<FunctionName>.User
+                            if (category.Length > 14 && category.EndsWith(".User", StringComparison.Ordinal))
+                            {
+                                return false;
+                            }
+                            else
+                            {
+                                return true;
+                            }
+                        });
+                    }
+                });
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -404,6 +404,23 @@ internal static void ConfigureApplicationInsights(HostBuilderContext context, IL
                         o.EnableDependencyTracking = false;
                     });
                 }
+
+                if (SystemEnvironment.Instance.IsApplicationInsightsAgentEnabled())
```
brettsam: For the dotnet-isolated devs (and others) -- I've talked with @RohitRanjanMS and we can re-use this logic, but instead of using a Java-specific environment variable, we'd use a capability that the worker can send to the host. The logic to filter will be the same.

I can follow up with that work. I think that'll be more general (and not require an app setting, etc), but we don't really need to block this PR on that.

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -404,6 +404,23 @@ internal static void ConfigureApplicationInsights(HostBuilderContext context, IL
...
+                {
+                    // Skip sending user generated logs to AI and QuickPulse if worker AI agent is configured, worker will send these logs to AI and Quickpulse service.
+                    builder.Services.Configure<LoggerFilterOptions>(options => options.AddFilter<ApplicationInsightsLoggerProvider>((category, logLevel) =>
+                    {
+                        // match Function.<FunctionName>.User
+                        if (category.Length > 14 && category.EndsWith(".User", StringComparison.Ordinal))
```
brettsam: I'm really curious about the perf of doing this vs lazily filling and checking a HashSet. This is fine for now as it won't impact any language but Java... and even then, a subset of scenarios, I believe. We can re-evaluate later if needed.

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -411,6 +411,23 @@ internal static void ConfigureApplicationInsights(HostBuilderContext context, IL
...
```
RohitRanjanMS: Benchmarked hashset and regex as well, it seems EndsWith is very efficient in this simple scenario.

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -411,6 +411,23 @@ internal static void ConfigureApplicationInsights(HostBuilderContext context, IL
                         o.EnableDependencyTracking = false;
                     });
                 }
+
+                if (SystemEnvironment.Instance.IsApplicationInsightsAgentEnabled())
+                {
+                    // Skip sending user generated logs to AI and QuickPulse if worker AI agent is configured, worker will send these logs to AI and Quickpulse service.
+                    builder.Services.Configure<LoggerFilterOptions>(options => options.AddFilter<ApplicationInsightsLoggerProvider>((category, logLevel) =>
+                    {
```
brettsam, line 419: There should be a way to get an `IEnvironment` inside here so you don't have to use `SystemEnvironment.Instance` -- that'd make things more testable.

I think it's like this, although I didn't confirm...

```csharp
builder.Services.AddOptions<LoggerFilterOptions>()
    .Configure<IEnvironment>((options, environment) => // environment comes from DI now
    {
        if (environment.IsApplicationInsightsAgentEnabled())
        {
            options.AddFilter<ApplicationInsightsLoggerProvider>((category, logLevel) =>
            {
                // etc...
            });
        }
    });
```

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
brettsam: You just snuck in before me with your benchmark comment! :-)

brettsam (approved):

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -411,6 +411,26 @@ internal static void ConfigureApplicationInsights(HostBuilderContext context, IL
                         o.EnableDependencyTracking = false;
                     });
                 }
+
+                builder.Services.AddOptions<LoggerFilterOptions>().Configure<IEnvironment>((options, environment) =>
+                {
+                    // Skip sending user generated logs to AI and QuickPulse if worker AI agent is configured, worker will send these logs to AI and Quickpulse service.
+                    if (environment.IsApplicationInsightsAgentEnabled())
+                    {
+                        options.AddFilter<ApplicationInsightsLoggerProvider>((category, logLevel) =>
```
jviau, line 420: Nothing against the filter, but curious what the reasoning for filtering here vs in a `ITelemetryProcessor` - especially since this is AppInsights specific.

RohitRanjanMS: ITelemetryProcessor processes the telemetry information before it is sent to Application Insights whereas AddFilter will filter this earlier in the pipeline. This should save some compute.

jviau (approved):

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -411,6 +411,26 @@ internal static void ConfigureApplicationInsights(HostBuilderContext context, IL
...
+                        {
+                            // match Function.<FunctionName>.User
+                            if (category.Length > 14 && category.EndsWith(".User", StringComparison.Ordinal))
```
jviau: Is `category` guaranteed to be non-null here?

RohitRanjanMS: yes, this is what my understanding is. I will double check with Brett otherwise I will add a null check to be safe.

RohitRanjanMS: Added null check.

PR #8917, opened 2022-11-16
catching exceptions due to race with worker process exit

### LINKED ISSUE #8916: Race condition on function timeout can lead to "Did not find any initialized language workers"
We have a race in [this chunk of code](https://github.com/Azure/azure-functions-host/blob/960fcfcada10c237acd044e0190afdc73ba1b747/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs#L1116-L1121) during a function timeout:
- one place handles the process's `Exited` event, then calls [`process.Close()`](https://github.com/Azure/azure-functions-host/blob/960fcfcada10c237acd044e0190afdc73ba1b747/src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs#L155)
- another calls [`process.WaitForExit()`](https://github.com/Azure/azure-functions-host/blob/960fcfcada10c237acd044e0190afdc73ba1b747/src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs#L211).

If `Close()` happens first, `WaitForExit()` will throw (silently in some cases -- it depends on the Trigger) and we'll just never restart the worker.

Leads to "Did not find any initialized language workers" message forever until site is restarted.

Wrote a sample console app to show:

```csharp
Process p = new Process();
// This app immediately exits
p.StartInfo.FileName = @"D:\projects\ConsoleApp2\bin\Debug\net6.0\ConsoleApp2.exe";
p.StartInfo.ArgumentList.Add("exit");
p.EnableRaisingEvents = true;
p.Exited += P_Exited;

void P_Exited(object? sender, EventArgs e)
{
    p.WaitForExit();
    p.Close();
}

p.Start();

// introduce a slight delay
await Task.Delay(1000);

try
{
    bool exited = p.WaitForExit(5000);
}
catch (Exception ex)
{
    // throws "No process is associated with this object."
}
```

brettsam:
Fixes #8916

### DIFF
--- a/src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs
+++ b/src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs
@@ -11,7 +11,6 @@ using Microsoft.Azure.WebJobs.Host.Scale;
 using Microsoft.Azure.WebJobs.Logging;
 using Microsoft.Azure.WebJobs.Script.Diagnostics;
 using Microsoft.Azure.WebJobs.Script.Eventing;
-using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 
@@ -140,12 +139,13 @@ namespace Microsoft.Azure.WebJobs.Script.Workers
 
         private void OnProcessExited(object sender, EventArgs e)
         {
+            _workerProcessLogger.LogDebug("Process {processId} has exited with code {exitCode}.", Process?.Id, Process?.ExitCode);
+
             if (Disposing)
             {
                 // No action needed
                 return;
             }
-            string exceptionMessage = string.Join(",", _processStdErrDataQueue.Where(s => !string.IsNullOrEmpty(s)));
 
             try
             {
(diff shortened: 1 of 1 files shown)


soninaren (approved):

src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs
```diff
@@ -208,7 +209,14 @@ internal void BuildAndLogConsoleLog(string msg, LogLevel level)
...
         public void WaitForProcessExitInMilliSeconds(int waitTime)
         {
-            Process.WaitForExit(waitTime);
+            try
+            {
+                Process.WaitForExit(waitTime);
+            }
+            catch (Exception ex)
+            {
+                _workerProcessLogger.LogDebug(ex, "An exception was thrown while waiting for process {processId} to exit. It is possible that the process had already exited and this can be ignored.", Process?.Id);
```
brettsam, line 221: catching this exception is the fix; others are just log improvements to help debug in the future

jviau (approved):

fabiocav (approved):

src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs
```diff
@@ -208,7 +209,14 @@ internal void BuildAndLogConsoleLog(string msg, LogLevel level)
 
         public void WaitForProcessExitInMilliSeconds(int waitTime)
         {
-            Process.WaitForExit(waitTime);
+            try
+            {
+                Process.WaitForExit(waitTime);
```
fabiocav: Can we also check if the process has exited here first (in addition to the exception handling) in an attempt to avoid this condition?

brettsam: yes, that's easy. coming soon...

PR #8935, opened 2022-11-23
Filter the secret for preventing logging

TsuyoshiUshio:
# Description

The place I changed could potentially share the SAS token secret on logging system. 
This is the idea of filtering out the logging.  

I don't fully understand the pattern which parameter has comes, so that review required. 
@pragnagopa @balag0 

I'd like to discuss if this approach works first. 

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/Management/LinuxSpecialization/BashCommandHandler.cs
+++ b/src/WebJobs.Script.WebHost/Management/LinuxSpecialization/BashCommandHandler.cs
@@ -43,7 +43,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management.LinuxSpecialization
                             CreateNoWindow = true
                         }
                     };
-                    _logger.LogInformation($"Running: {process.StartInfo.FileName} {process.StartInfo.Arguments}");
+                    _logger.LogInformation(ReplaceStorageSASTokenSecret($"Running: {process.StartInfo.FileName} {process.StartInfo.Arguments}"));
                     process.Start();
                     var output = process.StandardOutput.ReadToEnd().Trim();
                     var error = process.StandardError.ReadToEnd().Trim();
--- a/src/WebJobs.Script.WebHost/Management/LinuxSpecialization/BashCommandHandler.cs
+++ b/src/WebJobs.Script.WebHost/Management/LinuxSpecialization/BashCommandHandler.cs
@@ -40,7 +40,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management.LinuxSpecialization
                             CreateNoWindow = true
                         }
                     };
-                    _logger.LogInformation(Sanitizer.Sanitize($"Running: {process.StartInfo.FileName} {process.StartInfo.Arguments}"));
+
                     process.Start();
                     var output = process.StandardOutput.ReadToEnd().Trim();
                     var error = process.StandardError.ReadToEnd().Trim();
--- a/src/WebJobs.Script.WebHost/Management/LinuxSpecialization/BashCommandHandler.cs
+++ b/src/WebJobs.Script.WebHost/Management/LinuxSpecialization/BashCommandHandler.cs
@@ -5,6 +5,7 @@ using System;
 using System.Diagnostics;
 using System.IO;
 using System.IO.Compression;
+using System.Text.RegularExpressions;
 using Microsoft.Azure.WebJobs.Script.Diagnostics;
 using Microsoft.Extensions.Logging;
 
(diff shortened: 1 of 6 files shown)


src/WebJobs.Script.WebHost/Management/LinuxSpecialization/BashCommandHandler.cs
```diff
@@ -41,7 +43,7 @@ public BashCommandHandler(IMetricsLogger metricsLogger, ILogger<BashCommandHandl
                             CreateNoWindow = true
                         }
                     };
-                    _logger.LogInformation($"Running: {process.StartInfo.FileName} {process.StartInfo.Arguments}");
+                    _logger.LogInformation(ReplaceStorageSASTokenSecret($"Running: {process.StartInfo.FileName} {process.StartInfo.Arguments}"));
```
mathewc: We already have an existing component [Sanitizer](https://github.com/Azure/azure-functions-host/blob/259db7282f347522e8550f1262cc8fb07e583b0f/src/WebJobs.Script/Sanitizer.cs#L12) that we use in other places for this purpose. Recommend you use that and verify that it removes secrets in your scenario as well. It should.

However, depending on where these arguments are coming from, perhaps safest is to just not log the arguments at all - why do we need to?

TsuyoshiUshio: Hi @mathewc  
Thanks! Sanitizer is what I wanted!!!  I'll use it!
The argument can't be share in here. I'll add you on the e-mail thread.

TsuyoshiUshio: I tested locally, and found this case, Sanitizer doesn't detect it. 
I added pattern on Sanitizer.

surgupta-msft (approved):

src/WebJobs.Script.WebHost/Management/LinuxSpecialization/BashCommandHandler.cs
pragnagopa: @balag0 - do you see any issues if arguments are not logged?

TsuyoshiUshio: We talked on an e-mail thread. @mathewc recommended us to do the following approach. 
* Remove the centralized logging.
* Do the logging on the caller side.

I updated by this commit.  Test might fail. (I'll keep on update it) since Integration testing looking at the logs sometime. Could you review if it is ok for you? Since we might use logs for automation.

https://github.com/Azure/azure-functions-host/pull/8935/commits/8ffe3fc430414317c9d01008f720bbe26ed95e28

pragnagopa (approved):

src/WebJobs.Script.WebHost/Management/LinuxSpecialization/PackageDownloadHandler.cs
```diff
@@ -133,6 +135,8 @@ private void AriaDownload(string directory, string fileName, Uri zipUri, bool is
                 _logger.LogError(msg);
                 throw new InvalidOperationException(msg);
             }
+            _logger.LogInformation($"Executed: {Aria2CExecutable} --allow-overwrite -x12 -d {directory} -o {fileName} '{Sanitizer.Sanitize(zipUri?.AbsoluteUri)}");
```
mathewc: Here and below, don't copy this line, since things can get out of sync in the future. Instead capture the command line in a local and use in both places, running the entire log line through sanitizer.

src/WebJobs.Script.WebHost/Management/LinuxSpecialization/BashCommandHandler.cs
```diff
@@ -41,7 +40,7 @@ public BashCommandHandler(IMetricsLogger metricsLogger, ILogger<BashCommandHandl
                             CreateNoWindow = true
                         }
                     };
-                    _logger.LogInformation($"Running: {process.StartInfo.FileName} {process.StartInfo.Arguments}");
+                    _logger.LogInformation($"Running: {process.StartInfo.FileName}");
```
mathewc: Recommend either removing this line completely at this point, or updating it to make it clear that arguments are omitted, e.g. "Running: bash.exe (arguments omitted)"

src/WebJobs.Script.WebHost/Management/LinuxSpecialization/PackageDownloadHandler.cs
TsuyoshiUshio: Will do it.

src/WebJobs.Script.WebHost/Management/LinuxSpecialization/BashCommandHandler.cs
TsuyoshiUshio: Thanks. I'll go with "Running: bash.exe (arguments omitted) approach. 

I noticed the logging comes after this section, so that these logs became inconsistent. Also this logging is used for Integration testing. So that keep the Running logs with your recommendation will be great.

mathewc (approved):

PR #8948, opened 2022-12-01
Support collection of model-type bindings for out-of-proc workers

surgupta-msft:

This PR enables support for Collection of [model-type bindings](https://github.com/Azure/azure-functions-host/pull/8815)

Issue - https://github.com/Azure/azure-functions-dotnet-worker/issues/1156
Epic - https://github.com/Azure/azure-functions-dotnet-worker/issues/1081

### Pull request checklist

### Additional information

Additional PR information


src/WebJobs.Script.Grpc/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
```diff
@@ -431,6 +431,7 @@ message TypedData {
     CollectionDouble collection_double = 10;
     CollectionSInt64 collection_sint64 = 11;
     ModelBindingData model_binding_data = 12;
+    CollectionModelBindingData collection_model_binding_data = 13;
```
surgupta-msft: Will remove changes from proto file once [this](https://github.com/Azure/azure-functions-host/pull/8946) PR is merged.

test/WebJobs.Script.Tests/Workers/Rpc/GrpcMessageConversionExtensionsTests.cs
```diff
@@ -711,5 +711,34 @@ public void ToModelBindingData_Creates_Valid_BindingData()
...
+            var modelBindingData = new ModelBindingData
+            {
+                Version = parameterBindingData.Version,
+                ContentType = parameterBindingData.ContentType,
+                Source = parameterBindingData.Source,
+                Content = ByteString.CopyFrom(parameterBindingData.Content)
+            };
+
+            var collectionModelBindingData = new CollectionModelBindingData();
+            collectionModelBindingData.ModelBindingData.Add(modelBindingData);
```
liliankasem, line 736: potentially a little pedantic but can we test with more than one item in the collection?

probably also worth having a test for an empty collection and null

src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs
```diff
@@ -84,6 +85,23 @@ internal static TypedData ToModelBindingData(this ParameterBindingData data)
             return typedData;
         }
 
+        internal static TypedData ToModelBindingDataArray(this ParameterBindingData[] dataArray)
+        {
+            var collectionModelBindingData = new CollectionModelBindingData();
+
+            foreach (ParameterBindingData element in dataArray)
+            {
+                collectionModelBindingData.ModelBindingData.Add(element.ToModelBindingData().ModelBindingData);
```
liliankasem: What if element is null? or dataArray is empty?

src/WebJobs.Script/Binding/GeneralScriptBindingProvider.cs
```diff
@@ -72,7 +72,7 @@ internal static Type GetRequestedType(ScriptBindingContext context)
             {
                 // arrays are supported for both trigger input as well
                 // as output bindings
-                type = type.GetType() == typeof(ParameterBindingData) ? typeof(IEnumerable<>).MakeGenericType(type) : type.MakeArrayType();
+                type = context.SupportsDeferredBinding ? typeof(IEnumerable<ParameterBindingData>) : type.MakeArrayType();
```
liliankasem: Need to be careful with this - supportsDeferredBinding could be set to true, but then skipDeferredBinding might be added in the expression case so `supportsDeferredBinding` might be misleading - May need to check both flags. Theres also an extension util for this

src/WebJobs.Script/Binding/GeneralScriptBindingProvider.cs
```diff
@@ -71,7 +72,7 @@ internal static Type GetRequestedType(ScriptBindingContext context)
             {
                 // arrays are supported for both trigger input as well
                 // as output bindings
-                type = type.MakeArrayType();
+                type = context.SupportsDeferredBinding ? typeof(IEnumerable<ParameterBindingData>) : type.MakeArrayType();
```
surgupta-msft: @mathewc and @fabiocav looking for feedback on the change in this line. 

`MakeArrayType()` converts an object to its Array type - for example `string` becomes `string[]` and `ParameterBindingData` becomes `ParameterBindingData[]`. This array type is not able to bind to anything in extension code - [link](https://github.com/Azure/azure-sdk-for-net/blob/126ed7e63f99e305a962ac43613d55a25ab5fbfe/sdk/storage/Microsoft.Azure.WebJobs.Extensions.Storage.Blobs/src/Config/BlobsExtensionConfigProvider.cs#L110)

We wanted Array to bind to below line but it gives error that `expected IEnumerable<> but found array []`. Hence I made this change. I can also totally remove `MakeArrayType()` and convert everything to `IEnumerable<>` but wanted to get feedback first if it looks good and it won't break anything if changed to `IEnumerable<>`

```csharp
rule.AddOpenConverter<MultiBlobContext, IEnumerable<BlobCollectionType>>(typeof(BlobCollectionConverter<>), this);
```

src/WebJobs.Script/Binding/GeneralScriptBindingProvider.cs
surgupta-msft: I agree. We will also need similar change on line 63 above, right?

```csharp
Type type = context.SupportsDeferredBinding ? typeof(ParameterBindingData) : ParseDataType(context);
```

src/WebJobs.Script/Binding/GeneralScriptBindingProvider.cs
mathewc: Yeah, something seems wrong here - we shouldn't have to special case `IEnumerable<ParameterBindingData>` above. What OOP bindings does this array code currently work for? For example, the EventHub binding supports binding to EventData[] (e.g. [here](IEnumerable<ParameterBindingData>) in code, and [here](https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-event-hubs-trigger?tabs=in-process%2Cfunctionsv2%2Cextensionv5&pivots=programming-language-javascript#example) in doc). We should verify array works for EventHub OOP and figure out the difference. Perhaps we just need to update the blob extension to support binding to arrays. At the WebJobs SDK level, we have enumerable tests for the trigger (e.g. [here](https://github.com/Azure/azure-sdk-for-net/blob/4162f6fa2445b2127468b9cfd080f01c9da88eba/sdk/storage/Microsoft.Azure.WebJobs.Extensions.Storage.Blobs/samples/readmes/BlobExtensionSamples.cs#L261)) - does a test binding to `BlobClient[]` work as well?

surgupta-msft: On EventHub's side, below are the files doing specific handling for Array type `EventData[]` -

- [EventHubTriggerBindingStrategy](https://github.com/Azure/azure-sdk-for-net/blob/abed7d9c80073dde103af4a3b1f1daa7bccb9cdb/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Triggers/EventHubTriggerBindingStrategy.cs#L40) - ```public EventData[] BindMultiple(EventHubTriggerInput value)```
- [EventHubTriggerInput](https://github.com/Azure/azure-sdk-for-net/blob/abed7d9c80073dde103af4a3b1f1daa7bccb9cdb/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Triggers/EventHubTriggerInput.cs#L20)
- [EventHubExtensionConfigProvider](https://github.com/Azure/azure-sdk-for-net/blob/abed7d9c80073dde103af4a3b1f1daa7bccb9cdb/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Config/EventHubExtensionConfigProvider.cs#L66)


On Blob extensions side, currently `BlobClient[]` does not work because there is no rule in [BlobsExtensionConfigProvider](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/storage/Microsoft.Azure.WebJobs.Extensions.Storage.Blobs/src/Config/BlobsExtensionConfigProvider.cs) that can bind to array type (error - `expected IEnumerable<> but found array []`). Also, their [documentation](https://github.com/Azure/azure-sdk-for-net/blob/4162f6fa2445b2127468b9cfd080f01c9da88eba/sdk/storage/Microsoft.Azure.WebJobs.Extensions.Storage.Blobs/samples/readmes/BlobExtensionSamples.cs#L261) says that `IEnumerable<BlobClient>` works and they have a rule to bind to that in [BlobsExtensionConfigProvider](https://github.com/Azure/azure-sdk-for-net/blob/abed7d9c80073dde103af4a3b1f1daa7bccb9cdb/sdk/storage/Microsoft.Azure.WebJobs.Extensions.Storage.Blobs/src/Config/BlobsExtensionConfigProvider.cs#L102)

```csharp
rule.AddOpenConverter<MultiBlobContext, IEnumerable<BlobCollectionType>>(typeof(BlobCollectionConverter<>), this);
```

So, in summary, Host is converting to `Array type` but blob extension supports only `Enumerable type` while EventHub supports `array type`. Based on this, we have 2 options now -

1. Either, support `IEnumerable<>` in Host which I did above. One advantage of this approach is that we need only a single line change to support `IEnumerable<>`.
2. Or, support `Array` in Blob extensions. For this, I have prototyped changes below - 
          - On Host side, keep the line 74 above as it is: `type = type.MakeArrayType();`. 
          - On Blob Extensions side - [full changes in draft mode](https://github.com/surgupta-msft/azure-sdk-for-net/compare/sdk-bindings/support-reference-type...surgupta-msft:azure-sdk-for-net:sdk-bindings/collection-changes?expand=1)


``` csharp
// Add new rule for Array []. Rule for IEnumerable is already there in this class.
rule.AddOpenConverter<MultiBlobContext, BlobCollectionType[]>(typeof(BlobCollectionArrayConverter<>), this);

// Add new converter for Array. Converter for IEnumerable is already there in this class.
 private class BlobCollectionArrayConverter<T> : IAsyncConverter<MultiBlobContext, T[]>
```

With prototyped changes, option 2 is causing code duplication between `IEnumerable<BlobClient>` and `BlobClient[]` as there are 2 different converters for them, but if option-2 makes more sense then we can discuss how to minimize this duplication.

One concern from option-2 could be that because of it Blob extension will support `BlobClient[]` which is not mentioned in their documentation as of now. It only says about `IEnumerable<BlobClient>`. Can there be any issue because of this new support?

surgupta-msft: Updated code based on discussion offline. Also added changes in [blob extension PR](https://github.com/Azure/azure-sdk-for-net/pull/31757/commits/58c9eecd0d1f8525816ad24fd7590708cd6f36b9) to support array type

src/WebJobs.Script/Binding/GeneralScriptBindingProvider.cs
surgupta-msft: no change is needed in this class for collection support. Let's discuss offline about correcting line 63. Will create new PR if needed.

liliankasem (approved):

test/WebJobs.Script.Tests/Workers/Rpc/GrpcMessageConversionExtensionsTests.cs
```diff
@@ -711,5 +711,54 @@ public void ToModelBindingData_Creates_Valid_BindingData()
...
+            collectionModelBindingData.ModelBindingData.Add(modelBindingData);
+
+            TypedData typedData = new TypedData();
+            typedData.CollectionModelBindingData = collectionModelBindingData;
+
+            Assert.Equal(returned_typedData.CollectionModelBindingData.ModelBindingData.Count, 2);
```
kshyju: first param of `Equal` is the expected value.

test/WebJobs.Script.Tests/Workers/Rpc/GrpcMessageConversionExtensionsTests.cs
```diff
@@ -711,5 +711,54 @@ public void ToModelBindingData_Creates_Valid_BindingData()
...
+            Assert.Equal(typedData.CollectionModelBindingData.ModelBindingData.First(), returned_typedData.CollectionModelBindingData.ModelBindingData.First());
```
kshyju, line 743: Does this do a value comparison?

kshyju (approved):

test/WebJobs.Script.Tests/Workers/Rpc/GrpcMessageConversionExtensionsTests.cs
surgupta-msft: yes, validated again.

PR #8965, opened 2022-12-07
modifying approach to ignoring app insights coming from workers

brettsam:
#8915 introduced a way to ignore logs coming from App Insights based on an app setting. This takes that approach and modifies it so that it will also work for a worker that declares this behavior via a capability at startup.

Rather than relying strictly on the app setting, this now looks at a ~custom property on the current `Activity`~. This value is set in the grpc channel and is based on that channel's capabilities (or on the app setting as before).

Edit: Approach changed to using an `AsyncLocal`.

### DIFF
--- a/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
@@ -54,7 +54,6 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
         private readonly ChannelWriter<OutboundGrpcEvent> _outbound;
         private readonly ChannelReader<InboundGrpcEvent> _inbound;
         private readonly IOptions<FunctionsHostingConfigOptions> _hostingConfigOptions;
-
         private IDisposable _functionLoadRequestResponseEvent;
         private bool _disposed;
         private bool _disposing;
@@ -80,6 +79,7 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
         private TimeSpan _functionLoadTimeout = TimeSpan.FromMinutes(1);
         private bool _isSharedMemoryDataTransferEnabled;
         private bool _cancelCapabilityEnabled;
+        private bool _isWorkerApplicationInsightsLoggingEnabled;
 
         private System.Timers.Timer _timer;
 
@@ -94,7 +94,6 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
             IEnvironment environment,
             IOptionsMonitor<ScriptApplicationHostOptions> applicationHostOptions,
             ISharedMemoryManager sharedMemoryManager,
-            IFunctionDataCache functionDataCache,
             IOptions<WorkerConcurrencyOptions> workerConcurrencyOptions,
             IOptions<FunctionsHostingConfigOptions> hostingConfigOptions)
         {
@@ -392,7 +391,9 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
             }
 
             _state = _state | RpcWorkerChannelState.Initialized;
-            _workerCapabilities.UpdateCapabilities(_initMessage.Capabilities);
+
+            UpdateCapabilities(_initMessage.Capabilities);
+
             _isSharedMemoryDataTransferEnabled = IsSharedMemoryDataTransferEnabled();
             _cancelCapabilityEnabled = !string.IsNullOrEmpty(_workerCapabilities.GetCapabilityState(RpcWorkerConstants.HandlesInvocationCancelMessage));
 
@@ -402,7 +403,20 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
                 ScriptHost.IsFunctionDataCacheEnabled = false;
             }
 
+            if (_environment.IsApplicationInsightsAgentEnabled() ||
+                (bool.TryParse(_workerCapabilities.GetCapabilityState(RpcWorkerConstants.WorkerApplicationInsightsLoggingEnabled), out bool appInsightsWorkerEnabled) &&
+                appInsightsWorkerEnabled))
+            {
+                _isWorkerApplicationInsightsLoggingEnabled = true;
+            }
+
             _workerInitTask.TrySetResult(true);
+        }
+
+        // Allow tests to add capabilities, even if not directly supported by the worker.
+        internal virtual void UpdateCapabilities(IDictionary<string, string> fields)
+        {
+            _workerCapabilities.UpdateCapabilities(fields);
         }
 
         public void SetupFunctionInvocationBuffers(IEnumerable<FunctionMetadata> functions)
@@ -882,7 +896,7 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
         }
 
         /// <summary>
-        /// Request to free memory allocated by the worker (for output bindings)
+        /// Request to free memory allocated by the worker (for output bindings).
         /// </summary>
         /// <param name="outputMaps">List of names of shared memory maps to close from the worker.</param>
         internal void SendCloseSharedMemoryResourcesForInvocationRequest(IList<string> outputMaps)
@@ -905,12 +919,18 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
         internal void Log(GrpcEvent msg)
         {
             var rpcLog = msg.Message.RpcLog;
-            LogLevel logLevel = (LogLevel)rpcLog.Level;
             if (_executingInvocations.TryGetValue(rpcLog.InvocationId, out ScriptInvocationContext context))
             {
                 // Restore the execution context from the original invocation. This allows AsyncLocal state to flow to loggers.
-                System.Threading.ExecutionContext.Run(context.AsyncExecutionContext, (s) =>
+                System.Threading.ExecutionContext.Run(context.AsyncExecutionContext, static (state) =>
                 {
+                    var stateTuple = ((ScriptInvocationContext Context, RpcLog RpcLog, bool AppInsightsEnabledOnWorker))state;
+
+                    var rpcLog = stateTuple.RpcLog;
+                    LogLevel logLevel = (LogLevel)rpcLog.Level;
+
+                    var context = stateTuple.Context;
+
                     if (rpcLog.LogCategory == RpcLogCategory.CustomMetric)
                     {
                         if (rpcLog.PropertiesMap.TryGetValue(LogConstants.NameKey, out var metricName)
@@ -925,18 +945,27 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
                     }
                     else
                     {
-                        if (rpcLog.Exception != null)
+                        try
                         {
-                            // TODO fix RpcException catch all https://github.com/Azure/azure-functions-dotnet-worker/issues/370
-                            var exception = new Workers.Rpc.RpcException(rpcLog.Message, rpcLog.Exception.Message, rpcLog.Exception.StackTrace);
-                            context.Logger.Log(logLevel, new EventId(0, rpcLog.EventId), rpcLog.Message, exception, (state, exc) => state);
+                            WorkerTraceFilterTelemetryProcessor.FilterApplicationInsightsFromWorker.Value = stateTuple.AppInsightsEnabledOnWorker;
+
+                            if (rpcLog.Exception != null)
+                            {
+                                // TODO fix RpcException catch all https://github.com/Azure/azure-functions-dotnet-worker/issues/370
+                                var exception = new Workers.Rpc.RpcException(rpcLog.Message, rpcLog.Exception.Message, rpcLog.Exception.StackTrace);
+                                context.Logger.Log(logLevel, new EventId(0, rpcLog.EventId), rpcLog.Message, exception, (state, exc) => state);
+                            }
+                            else
+                            {
+                                context.Logger.Log(logLevel, new EventId(0, rpcLog.EventId), rpcLog.Message, null, (state, exc) => state);
+                            }
                         }
-                        else
+                        finally
                         {
-                            context.Logger.Log(logLevel, new EventId(0, rpcLog.EventId), rpcLog.Message, null, (state, exc) => state);
+                            WorkerTraceFilterTelemetryProcessor.FilterApplicationInsightsFromWorker.Value = false;
                         }
                     }
-                }, null);
+                }, (context, rpcLog, _isWorkerApplicationInsightsLoggingEnabled));
             }
         }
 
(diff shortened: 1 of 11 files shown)


liliankasem (approved):

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -49,7 +49,7 @@ internal class GrpcWorkerChannel : IRpcWorkerChannel, IDisposable
         private readonly IOptions<WorkerConcurrencyOptions> _workerConcurrencyOptions;
         private readonly WaitCallback _processInbound;
         private readonly object _syncLock = new object();
-        private readonly Dictionary<MsgType, Queue<PendingItem>> _pendingActions = new ();
+        private readonly Dictionary<MsgType, Queue<PendingItem>> _pendingActions = new Dictionary<MsgType, Queue<PendingItem>>();
```
jviau: This change necessary? I personally like the `new();` syntax (because the type is clear from the left side).

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -916,18 +930,38 @@ internal void Log(GrpcEvent msg)
...
-                        if (rpcLog.Exception != null)
+                        bool cleanup = false;
+                        try
                         {
-                            // TODO fix RpcException catch all https://github.com/Azure/azure-functions-dotnet-worker/issues/370
-                            var exception = new Workers.Rpc.RpcException(rpcLog.Message, rpcLog.Exception.Message, rpcLog.Exception.StackTrace);
-                            context.Logger.Log(logLevel, new EventId(0, rpcLog.EventId), rpcLog.Message, exception, (state, exc) => state);
+                            if (stateTuple.AppInsightsEnabledOnWorker)
+                            {
+                                if (Activity.Current != null)
```
jviau: I am not super familiar with our `Activity` usage, but I know `Activity` has optimizations when using `ActivitySource` to _not create_ the Activity at all when there are no listeners. So, do we risk `Activity` _sometimes_ being `null` here and thus this will work only some of the time?

I see that above we have a comment that "this allows AsyncLocal state to flow to loggers". What do you think of using our own `AsyncLocal<T>` here instead? Might not need the cleanup logic as well then. Or if you do, the `T` could be a small `IDisposable` implementation which performs the cleanup on dispose.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
brettsam: What settings do I need to build this locally? Something that I have configured automatically removes that space, then it fails to build, then I re-add it, etc. But it only happens if I ever open this file. I literally cannot run locally with this like this, and this was the easiest solution.

I suspect others would hit this as well... if there's some editorconfig I can use to make this work, I'd gladly do it.

I tried moving to a newer StyleCop, but that opened up a whole new can of worms with dozens of newer violations that it wanted me to address.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
brettsam: None of this uses ActivitySource today -- it's all manually created by App Insights and/or WebJobs.

Your comment about `AsyncLocal` is what I had originally proposed to @fabiocav (wrapped in something like `IHttpContextAccessor`... where another class can access some saved state of the current invocation). Is that something like what you were thinking?

I'm thinking about moving to that now after your comment. Any odd dependencies we take on Activity like this may be troublesome to decouple in the future. For example, if no one is listening, then stuff starts to fall apart b/c we don't have an Activity to store this state on.

jviau: You could do some more robust accessor similar to `IHttpContextAccessor` if you want to put in that effort. I was thinking more a pointed `AsyncLocal` which is for _just_ this use case. It can be a static and directly accessed by both places that set and get it. So long as we keep it `internal` I am not worried too much about the robustness of it. It is just as you said - relying on `Activity` may not be the most stable approach. Especially if we make any changes to our tracing in the future.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
jviau: That space _should_ be removed. The style is typically `new()` and not `new ()`. What error are you getting for the failed build? I see we don't have a `global.json` to lock the dotnet SDK in place, so this could be a case of you having an older dotnet SDK?

src/WebJobs.Script/Config/ApplicationInsightsLoggerFilterOptionsSetup.cs
```diff
@@ -0,0 +1,32 @@
...
+            _environment = environment;
+        }
+
+        public void Configure(LoggerFilterOptions options)
+        {
+            // Out-of-proc workers have the option of handling App Insights by themselves. If this is the case, they can
+            // set this AsyncLocal to true to indicate the log should be skipped.
+            if (!_environment.IsDotNetInProc())
+            {
+                options.AddFilter<ApplicationInsightsLoggerProvider>((_, _) => !FilterApplicationInsightsFromWorker.Value);
```
jviau: Given the pains we have had with provider-specific filters, I do wonder if this will also prevent the non-provider filters from being applied?

Can we verify if when this is `True`, that other filters will still be applied?

src/WebJobs.Script/ScriptConstants.cs
```diff
@@ -39,6 +39,8 @@ public static class ScriptConstants
         public const string LogPropertyProcessIdKey = "ProcessId";
         public const string LogPropertyActivityIdKey = "MS_ActivityId";
 
+        public const string IgnoreApplicationInsightsKey = "MS_IgnoreApplicationInisghts";
```
mathewc: Is this used anywhere? I don't see a reference.

mathewc:
What tests can we add for this code? As it stands, looks like there are no tests, which means we won't know if we regress this?

src/WebJobs.Script/Config/ApplicationInsightsLoggerFilterOptionsSetup.cs
RohitRanjanMS: we can remove one of the "_". Use the overload with just level filter.

src/WebJobs.Script/ScriptConstants.cs
brettsam: Not any more; i've removed it in my next commit  -- also spelled wrong!

brettsam:
@mathewc -- adding tests now. Required some shuffling around of how the GrpcCapabilities are created so I can inject this capability into an existing node test (even though node doesn't actually support it).

src/WebJobs.Script/Config/ApplicationInsightsLoggerFilterOptionsSetup.cs
brettsam: Done.

brettsam: Yes, I can add a test for this. To be specific, is your concern that b/c this filter is specific, it gets picked and other filters are ignored so if a worker does not have a capability, suddenly something generic like this in host.json would be ignored?

```
{
  "Logging": {
    "LogLevel": {
      "Function.Function1.User": "None" 
    }
  }
}
```

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
brettsam: This was fixed in #8975. Have rebased on that change.

src/WebJobs.Script/Config/ApplicationInsightsLoggerFilterOptionsSetup.cs
brettsam: Okay -- good call here as it did completely interact with the other filters in non-intuitive ways. I've gone ahead and moved the check to an `ITelemetryProcessor` now, meaning that only App Insights will see this and all filters remain as-is.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -112,7 +113,7 @@ internal class GrpcWorkerChannel : IRpcWorkerChannel, IDisposable
             _processInbound = state => ProcessItem((InboundGrpcEvent)state);
             _hostingConfigOptions = hostingConfigOptions;
 
-            _workerCapabilities = new GrpcCapabilities(_workerChannelLogger);
+            _grpcCapabilitiesFactory = grpcCapabilitiesFactory;
```
brettsam: This change isn't great as some locations check capabilities before they're set, so now they're null... I'm going to think through this. The behavior of this capabilities object is pretty fragile (and I've just realized that we check/parse the values a lot... may have to refactor this first).

brettsam:
My new PR backs some things out:
- @jviau's comment about testing filters was a great one. It highlighted some issues. I've moved this into an `ITelemetryProcessor`, which means now it definitely will only be touched by App Insights and will not be affected by other filters.
- Simplified the testing from other PRs. Rather than overhauling `GrpcCapabilities` here, I've filed an issue for some refactoring in the future: https://github.com/Azure/azure-functions-host/issues/8983.

jviau (approved):

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
brettsam: I ended up simplifying this behavior by adding a virtual method I can override to intercept these capabilities in a test. It's still not beautiful, but we have plans already to overhaul a bit of this (capabilities included) and I felt like I was getting too far into the weeds.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannelFactory.cs
```diff
@@ -25,12 +25,11 @@ public class GrpcWorkerChannelFactory : IRpcWorkerChannelFactory
         private readonly IEnvironment _environment = null;
         private readonly IOptionsMonitor<ScriptApplicationHostOptions> _applicationHostOptions = null;
         private readonly ISharedMemoryManager _sharedMemoryManager = null;
-        private readonly IFunctionDataCache _functionDataCache = null;
         private readonly IOptions<WorkerConcurrencyOptions> _workerConcurrencyOptions;
         private readonly IOptions<FunctionsHostingConfigOptions> _hostingConfigOptions;
 
         public GrpcWorkerChannelFactory(IScriptEventManager eventManager, IEnvironment environment, IRpcServer rpcServer, ILoggerFactory loggerFactory, IOptionsMonitor<LanguageWorkerOptions> languageWorkerOptions,
-            IOptionsMonitor<ScriptApplicationHostOptions> applicationHostOptions, IRpcWorkerProcessFactory rpcWorkerProcessManager, ISharedMemoryManager sharedMemoryManager, IFunctionDataCache functionDataCache,
+            IOptionsMonitor<ScriptApplicationHostOptions> applicationHostOptions, IRpcWorkerProcessFactory rpcWorkerProcessManager, ISharedMemoryManager sharedMemoryManager,
```
kshyju, line 32: Should we also remove `languageWorkerOptions` and `rpcServer` params? They appears to be unused.

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -606,5 +604,13 @@ public static bool IsTargetBasedScalingEnabled(this IEnvironment environment)
         {
             return string.Equals(environment.GetEnvironmentVariable(TargetBaseScalingEnabled), "1");
         }
+
+        public static bool IsDotNetInProc(this IEnvironment environment)
```
kshyju: Do we use this method? Don't see anywhere being called.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -402,9 +403,22 @@ internal void WorkerInitResponse(GrpcEvent initEvent)
                 ScriptHost.IsFunctionDataCacheEnabled = false;
             }
 
+            if (_environment.IsApplicationInsightsAgentEnabled() ||
+                (bool.TryParse(_workerCapabilities.GetCapabilityState(RpcWorkerConstants.WorkerApplicationInsightsLoggingEnabled), out bool appInsightsWorkerEnabled) &&
+                appInsightsWorkerEnabled))
```
kshyju, line 408: For other places, we don't really check the capability value is truthy.(ex: line 398). Should we stick to that approach for being consistent?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -402,9 +403,22 @@ internal void WorkerInitResponse(GrpcEvent initEvent)
                 ScriptHost.IsFunctionDataCacheEnabled = false;
             }
 
+            if (_environment.IsApplicationInsightsAgentEnabled() ||
```
kshyju, line 406: Question about the `||` condition here. This means the app setting cannot override the worker behavior (_worker sent a capability indicating ai logging enabled_). But customer wants to override it for some reason without redeploying worker payload and this _or_ condition will not allow that. Is that intentional behavior?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
brettsam: I really dislike that we only check for existence of the key in a lot of places -- it means if you explicitly set something to "false", it'll still be evaluated as "true". I find that really confusing.

I didn't want to touch all that, though -- but I do think we should address it at some point. I just didn't want to add to the confusion.

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
brettsam: yeah, this likely doesn't really matter much anymore. will pull it out.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
brettsam: Yeah, it was intentional. My thinking was that if either of these are set, we need to respect that and use it. In reality, most people won't even know about this and it's up to the workers/infrastructure to set these.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannelFactory.cs
brettsam: I didn't even notice that -- yeah, let me clean that up while I'm making these changes.

kshyju (approved):

PR #8975, opened 2022-12-13
bumping stylecop and addressing all errors

brettsam:
A small issue with StyleCop (not allowing `... = new();` caused me to investigate bumping it up for newer C# language support. This resulted in a bunch of very similar, boilerplate errors that I went through and addressed. They were:
- Tuples with no element names.
- Tuples with lowercase element names.
- No `.` at the end of doc comments.
- Changing `new ();` -> `new();`

### DIFF
--- a/src/WebJobs.Script/Description/Workers/ScriptInvocationContext.cs
+++ b/src/WebJobs.Script/Description/Workers/ScriptInvocationContext.cs
@@ -20,7 +20,7 @@ namespace Microsoft.Azure.WebJobs.Script.Description
 
         public IEnumerable<KeyValuePair<string, string>> Attributes { get; set; }
 
-        public IEnumerable<(string name, DataType type, object val)> Inputs { get; set; }
+        public IEnumerable<(string Name, DataType Type, object Val)> Inputs { get; set; }
 
         public Dictionary<string, object> BindingData { get; set; }
 
--- a/src/WebJobs.Script/Workers/Http/DefaultHttpWorkerService.cs
+++ b/src/WebJobs.Script/Workers/Http/DefaultHttpWorkerService.cs
@@ -51,9 +51,9 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Http
 
         private static HttpClient CreateHttpClient(IOptions<HttpWorkerOptions> httpWorkerOptions)
         {
-            HttpClientHandler handler = new ();
+            HttpClientHandler handler = new();
             handler.AllowAutoRedirect = !httpWorkerOptions.Value.EnableForwardingHttpRequest;
-            return new (handler);
+            return new(handler);
         }
 
         public Task InvokeAsync(ScriptInvocationContext scriptInvocationContext)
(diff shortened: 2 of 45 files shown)


src/WebJobs.Script/Description/Workers/ScriptInvocationContext.cs
```diff
@@ -20,7 +20,7 @@ public class ScriptInvocationContext
 
         public IEnumerable<KeyValuePair<string, string>> Attributes { get; set; }
 
-        public IEnumerable<(string name, DataType type, object val)> Inputs { get; set; }
+        public IEnumerable<(string Name, DataType Type, object Val)> Inputs { get; set; }
```
jviau, line 23: This might be a breaking change. Is this assembly publicly used?

src/WebJobs.Script/Workers/Http/DefaultHttpWorkerService.cs
```diff
@@ -79,12 +79,12 @@ internal async Task ProcessHttpInAndOutInvocationRequest(ScriptInvocationContext
                 Outputs = new Dictionary<string, object>()
             };
 
-            (string name, DataType type, object request) input = scriptInvocationContext.Inputs.First();
+            (string Name, DataType Type, object Request) input = scriptInvocationContext.Inputs.First();
```
jviau: Do you know why we have both the deconstructor and the `input` variable? Typically, for deconstructors you only have that and then refer to them directly:

``` CSharp
(string name, DataType type, object request) = scriptInvocationContext.Inputs.First();
HttpRequest httpRequest = request as HttpRequest;
....
```

Alternatively, you use `var` and let the signature naming take hold:

``` CSharp
var input = scriptInvocationContext.Inputs.First();
HttpRequest httpRequest = input.Request as HttpRequest;
```

mathewc (approved):

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
```diff
@@ -251,7 +251,7 @@ internal WorkerProcessCountOptions GetWorkerProcessCount(JObject workerConfig)
             // Validate
             if (workerProcessCount.ProcessCount <= 0)
             {
-                throw new ArgumentOutOfRangeException($"{nameof(workerProcessCount.ProcessCount)}",  "ProcessCount must be greater than 0.");
+                throw new ArgumentOutOfRangeException($"{nameof(workerProcessCount.ProcessCount)}", "ProcessCount must be greater than 0.");
```
jviau: The string interpolation here is unnecessary? Realize this is not from your changes though.
```suggestion
                throw new ArgumentOutOfRangeException(nameof(workerProcessCount.ProcessCount), "ProcessCount must be greater than 0.");
```

jviau (commented):
Only concerned about that one possible breaking change.

src/WebJobs.Script/Description/Workers/ScriptInvocationContext.cs
mathewc: It is a public type exposed from the Microsoft.Azure.WebJobs.Script package, so technically you're right. However in practice, this isn't a library customers are building on, just us. So I think this is OK. We've made similar "breaks" in the past.

brettsam: Yeah, what @mathewc said. I believe this is only consumed by Core Tools via an internal feed, and we typically just absorb any API break there when we update the reference. In this case, though, I'm see no reference to this type anyway, so we likely don't even need to adjust anything downstream.

src/WebJobs.Script/Workers/Http/DefaultHttpWorkerService.cs
brettsam: Not sure -- and in fact, in this case, it's odd b/c the third element in the tuple doesn't actually match the name of the one defined in the method (`Request` vs `Val`). I'm going to change this to `var` to make this more straightforward.

fabiocav (approved):

PR #9001, opened 2022-12-23
Add SignalR heartbeat as an approved service

MatthewSteeples:

I haven't created an issue for this Pull Request, but the problem it is solving is that when you register Azure SignalR service to use in your own code (rather than what is passed in as a function) you receive an error both in the local emulator and the Functions runtime indicating that the Heartbeat service is not permitted.

```
Invalid host services. Microsoft.Azure.WebJobs.Script.WebHost: The following service registrations did not match the expected services:
[Invalid] ServiceType: Microsoft.Extensions.Hosting.IHostedService, Lifetime: Singleton, ImplementationType: Microsoft.Azure.SignalR.HeartBeat, Microsoft.Azure.SignalR, Version = 1.19.2.0, Culture = neutral, PublicKeyToken = adb9793829ddae60.
```

This change adds that class to the approved list.

### Pull request checklist

I don't know if this is a big enough change to be included in the release notes, or if it needs a specific test (it's akin to a configuration change as far as I can see).

* [?] My changes **should not** be added to the release notes for the next release
* [?] I have added all required tests (Unit tests, E2E tests)


brettsam:
We have the approved list to prevent the running of Services that don't compose well with Functions and may result in a bad experience or supportability issues. 

I don't know enough about this scenario to know whether it's a good candidate to add here. I see a comment like this, which suggests that there's a different approach that works with Functions: https://github.com/Azure/azure-signalr/issues/1329#issuecomment-869375720

Is there a reason that this is required over using the SignalR bindings? https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-signalr-service?tabs=in-process&pivots=programming-language-csharp

MatthewSteeples:
As part of our migration away from WebJobs and towards Functions we weren't making use of the bindings, which is why this was required. Bits of our code were responsible only for sending SignalR messages (and not receiving them). We were fairly new to the Azure SignalR service and it looks like the referenced "Management" library is the way forward with this, so we'll try that for now.

PR #9027, opened 2023-01-11
Skip deferred binding for CosmosDB trigger

### LINKED ISSUE #9026: Make sure not to use ParameterBindingData for Cosmos DB triggers
CosmosDB triggers will not work for ParameterBindingData (and therefore sdk-type bindings). We need to make sure that when a CosmosDB trigger is being used in dotnet-isolated, we should not be using `ParameterBindingData` / `supportsDeferredBinding` should be `False`

liliankasem:

resolves #9026

With the way the cosmos db sdk is implemented it is not possible to use ParameterBindingData/DeferredBinding for cosmos DB triggers with out-of-proc workers. We need to make sure we continue to handle triggers the same way we do today, hence the skipDeferredBinding flag.

See this [issue](https://github.com/Azure/azure-webjobs-sdk-extensions/issues/800) discussion for reference.

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script/Description/Workers/WorkerFunctionDescriptorProvider.cs
+++ b/src/WebJobs.Script/Description/Workers/WorkerFunctionDescriptorProvider.cs
@@ -65,11 +65,15 @@ namespace Microsoft.Azure.WebJobs.Script.Description
         {
             var bindings = inputBindings.Union(outputBindings);
 
-            // If an input or output binding uses expressions, we do not want to use ParameterBindingData for the trigger bindings
             if (triggerMetadata.SupportsDeferredBinding())
             {
-                bool skipDeferredBinding = BindingAttributeContainsExpression(bindings);
-                if (skipDeferredBinding)
+                // If an input or output binding uses expressions, we do not want to use ParameterBindingData for the trigger bindings
+                bool containsExpression = BindingAttributeContainsExpression(bindings);
+
+                // If the trigger is a CosmosDBTrigger, we do not want to use ParameterBindingData as it is not supported
+                bool isCosmosTrigger = IsCosmosDBTrigger(triggerMetadata);
+
+                if (containsExpression || isCosmosTrigger)
                 {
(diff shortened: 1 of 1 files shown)


src/WebJobs.Script/Description/Workers/WorkerFunctionDescriptorProvider.cs
```diff
@@ -65,11 +65,15 @@ protected override IFunctionInvoker CreateFunctionInvoker(string scriptFilePath,
...
-            // If an input or output binding uses expressions, we do not want to use ParameterBindingData for the trigger bindings
             if (triggerMetadata.SupportsDeferredBinding())
             {
-                bool skipDeferredBinding = BindingAttributeContainsExpression(bindings);
-                if (skipDeferredBinding)
+                // If an input or output binding uses expressions, we do not want to use ParameterBindingData for the trigger bindings
+                bool containsExpression = BindingAttributeContainsExpression(bindings);
+
+                // If the trigger is a CosmosDBTrigger, we do not want to use ParameterBindingData as it is not supported
+                bool isCosmosTrigger = IsCosmosDBTrigger(triggerMetadata);
```
jviau, line 74: It doesn't feel right to me to insert single extension specific logic into this core layer. Can this be pulled into a formal extensibility behavior? Can we delegate the supported decision to the `BindingMetadata` itself, and extenders can control it?

liliankasem: Yup agree, was trying to have a quick patch in place for if we release Cosmos soon.

However, I did sync with Fabio about this today and since Cosmos is not in any rush to be released we can focus on refactoring the supportsDeferredBinding attribute for extensions and the function metadata generator on the worker to have more granular information about what bindings support deferred binding rather than the blanket "all or nothing" approach we have in place today. 

As a result, I'll abandon this PR and create an issue for the refactor work that needs to happen in the worker

PR #9028, opened 2023-01-12
Added telemetry to indicate when ParameterBindingData is being used

### LINKED ISSUE #9021: Add logging and telemetry to indicate when ParameterBindingData is being used
With the introduction of the SDK-type binding feature, we would like to have telemetry in place to help us identify when this feature is being used. This will help us:

a) build a detector to identify when ParameterBindingData is being used in a customer app
b) gather telemetry for adoption of this new feature

- [ ] Add telemetry log in host to indicate ParameterBindingData is being used

surgupta-msft:

resolves #9021 

This PR adds telemetry in host to indicate when ParameterBindingData is being used to help track adoption of this new feature

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
@@ -674,7 +674,7 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
 
                 foreach ((string, DataType, object) element in context.Inputs)
                 {
-                    if (element.Item3.GetType() == typeof(ParameterBindingData))
+                    if (element.Item3.GetType() == typeof(ParameterBindingData) || element.Item3.GetType() == typeof(ParameterBindingData[]))
                     {
                         _workerChannelLogger.LogInformation($"Binding to ParameterBindingData:{element}");
                         _metricsLogger.LogEvent(string.Format(MetricEventNames.BindToParameterBindingData, context.FunctionMetadata.Name, element.Item1));
--- a/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
@@ -14,6 +14,7 @@ using System.Threading;
 using System.Threading.Channels;
 using System.Threading.Tasks;
 using System.Threading.Tasks.Dataflow;
+using System.Xml.Linq;
 using Google.Protobuf.Collections;
 using Google.Protobuf.WellKnownTypes;
 using Microsoft.Azure.WebJobs.Logging;
@@ -590,6 +591,23 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
                 BindingInfo bindingInfo = binding.ToBindingInfo();
 
                 request.Metadata.Bindings.Add(binding.Name, bindingInfo);
+
+                binding.Properties.TryGetValue(ScriptConstants.SupportsDeferredBindingKey, out bool supportsDeferredBindingFlag);
+
+                if (supportsDeferredBindingFlag)
+                {
+                    binding.Properties.TryGetValue(ScriptConstants.SkipDeferredBindingKey, out bool skipDeferredBindingFlag);
+
+                    if (skipDeferredBindingFlag)
+                    {
+                        _workerChannelLogger.LogInformation("Function '{functionName}' skipping binding to ParameterBindingData for binding name '{name}' and type '{type}' as '{SkipDeferredBinding}' set to '{value}'", metadata.Name, binding.Name, binding.Type, ScriptConstants.SkipDeferredBindingKey, skipDeferredBindingFlag);
+                    }
+                    else
+                    {
+                        _workerChannelLogger.LogInformation("Function '{functionName}' binding to ParameterBindingData for binding name '{name}' and type '{type}' as '{supportsDeferredBinding}' set to '{value}'", metadata.Name, binding.Name, binding.Type, ScriptConstants.SupportsDeferredBindingKey, supportsDeferredBindingFlag);
+                        _metricsLogger.LogEvent(string.Format(MetricEventNames.BindToParameterBindingData, metadata.Name));
+                    }
+                }
             }
 
             foreach (var property in metadata.Properties)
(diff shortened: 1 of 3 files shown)


src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -672,6 +672,15 @@ internal async Task SendInvocationRequest(ScriptInvocationContext context)
                     return;
                 }
 
+                foreach ((string, DataType, object) element in context.Inputs)
+                {
+                    if (element.Item3.GetType() == typeof(ParameterBindingData) || element.Item3.GetType() == typeof(ParameterBindingData[]))
```
kshyju: Question: Do we care about collecting this information for every function invocation at runtime? If we are interested in usage of sdk binding usage in function apps, could we not infer that from the function metadata during worker startup/init?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -672,6 +672,15 @@ internal async Task SendInvocationRequest(ScriptInvocationContext context)
...
+                    {
+                        _workerChannelLogger.LogInformation("Function '{functionName}' binding to ParameterBindingData for parameter '{name}', datatype '{datatype}' and binding '{binding}'", context.FunctionMetadata.Name, element.Item1, element.Item2, element.Item3);
```
kshyju: Question: Do we need to log this to functionslogs as well? Is logging to FunctionsMetrics not enough? We probably are going to run aggregate queries against the metrics table. Don't we?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
liliankasem: Yeah I think we might be able to log when parsing the metadata to determine the supports deferred binding flag and use that as an indication that this function should be using ParameterBindingData. Keeping in mind that not every function in the Function App will be using `ParameterBindingData`, just the ones who use the new blob ext so we will need to capture for each function.

I think that should be enough on the telemetry front. But for logging/debug purposes, we can log different parts of the process to see whats happening with the data as we handle it

mathewc: I agree with Shyju that we shouldn't be logging this on each invocation, since that has performance implications. It should done at index/load type time. Also we shouldn't need both a log statement and a metric event right? Can't the metric event suffice?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -672,6 +672,15 @@ internal async Task SendInvocationRequest(ScriptInvocationContext context)
                     return;
                 }
 
+                foreach ((string, DataType, object) element in context.Inputs)
```
jviau: nit: use deconstruction:

```suggestion
                foreach ((string name, DataType dataType, object binding) in context.Inputs)
```

Then instead of `element.ItemX`, you can use appropriate variable from above.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
jviau: Curious why `Information` and not `Debug`?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
surgupta-msft: Change here is not needed now

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
surgupta-msft: Checking with Mathew in another comment and will remove log if not needed

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
surgupta-msft: Moved telemetry to Load request instead of per invocation.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -590,6 +591,23 @@ internal FunctionLoadRequest GetFunctionLoadRequest(FunctionMetadata metadata, M
...
+
+                binding.Properties.TryGetValue(ScriptConstants.SupportsDeferredBindingKey, out bool supportsDeferredBindingFlag);
+
+                if (supportsDeferredBindingFlag)
+                {
+                    binding.Properties.TryGetValue(ScriptConstants.SkipDeferredBindingKey, out bool skipDeferredBindingFlag);
+
+                    if (skipDeferredBindingFlag)
+                    {
+                        _workerChannelLogger.LogInformation("Function '{functionName}' skipping binding to ParameterBindingData for binding name '{name}' and type '{type}' as '{SkipDeferredBinding}' set to '{value}'", metadata.Name, binding.Name, binding.Type, ScriptConstants.SkipDeferredBindingKey, skipDeferredBindingFlag);
```
surgupta-msft, line 603: @mathewc I have moved telemetry to Load request instead of per invocation. Wanted to double check if we need these log statements or only metric is sufficient. I have added log in case we will need information such as Binding type, name, but totally fine to remove them if not needed.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -590,6 +591,11 @@ internal FunctionLoadRequest GetFunctionLoadRequest(FunctionMetadata metadata, M
                 BindingInfo bindingInfo = binding.ToBindingInfo();
 
                 request.Metadata.Bindings.Add(binding.Name, bindingInfo);
+
+                if (binding.SupportsDeferredBinding() && !binding.SkipDeferredBinding())
+                {
+                    _metricsLogger.LogEvent(string.Format(MetricEventNames.BindToParameterBindingData, metadata.Name));
+                }
```
surgupta-msft, line 598: @mathewc let me know if we need any log line here giving more information such as - binding name/type and values of `supportsDeferredBinding` and `SkipDeferredBinding`.

liliankasem: It might help to have a trace/debug log here

liliankasem: For `SkipDeferredBinding`, can you add a trace log [here](https://github.com/Azure/azure-functions-host/blob/afafa87da101d8ed2fe1fa7aee317692db252786/src/WebJobs.Script/Description/FunctionDescriptorProvider.cs#L150) to indicate that we are changing `supportsDeferredBinding` to false because the `SkipDeferredBinding` flag was set

liliankasem: It might also be worth a trace log [here](https://github.com/Azure/azure-functions-host/blob/afafa87da101d8ed2fe1fa7aee317692db252786/src/WebJobs.Script/Description/Workers/WorkerFunctionDescriptorProvider.cs#L74) to let us know that an input or output binding contains an expression

src/WebJobs.Script/Diagnostics/MetricEventNames.cs
```diff
@@ -52,6 +52,7 @@ public static class MetricEventNames
         // function level events
         public const string FunctionInvokeLatency = "function.invoke.latency";
         public const string FunctionBindingTypeFormat = "function.binding.{0}";
+        public const string BindToParameterBindingData = "function{0}.binding.parameterbindingdata";
```
mathewc: The format parameter is the function name? I think this metric name should just be function.binding.parameterbindingdata or maybe better function.binding.deferred, and the function name is handled by the LogEvent overload that takes a function name as a parameter.

surgupta-msft: The name function.binding.deferred looks good. Updated.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
surgupta-msft: We have few questions to discuss on logging part. I will start discussion and another PR for logging. We can keep this PR for metrics part.

mathewc (approved):

liliankasem (approved):

PR #9041, opened 2023-01-18
Adding an API to get worker process info

### LINKED ISSUE #8840: Create an ARM API to provide process information for OOP workers
Related issue: https://github.com/Azure/azure-functions-host/issues/8611

soninaren:

resolves #8840

### Pull request checklist

### Additional information

Additional PR information


src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -107,6 +109,36 @@ public async Task<IActionResult> GetHostStatus([FromServices] IScriptHostManager
...
         }
 
+        [HttpGet]
+        [Route("admin/host/workerProcesses")]
+        [Authorize(Policy = PolicyNames.AdminAuthLevelOrInternal)]
+        public async Task<IActionResult> GetWorkerProcesses([FromServices] IServiceProvider serviceProvider = null)
+        {
+            var processes = new List<WorkerProcessInfo>();
+            // When making changes to HostStatus, ensure that you update the HostStatus class in AAPT-Antares-Websites repo as well.
+            var rpcFunctionInvocationDispatcher = serviceProvider.GetService<RpcFunctionInvocationDispatcher>();
```
liliankasem: Do we want to use something like:

```
if (Utility.TryGetHostService(serviceProvider, out RpcFunctionInvocationDispatcher rpcFunctionInvocationDispatcher))
```

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -107,6 +109,36 @@ public async Task<IActionResult> GetHostStatus([FromServices] IScriptHostManager
...
+            if (rpcFunctionInvocationDispatcher != null)
```
liliankasem: nit: 
```suggestion
            if (rpcFunctionInvocationDispatcher is not null)
```

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -107,6 +109,36 @@ public async Task<IActionResult> GetHostStatus([FromServices] IScriptHostManager
...
+            {
+                var channels = await rpcFunctionInvocationDispatcher.GetAllWorkerChannelsAsync();
+                if (channels != null)
```
liliankasem:
```suggestion
                if (channels is not null)
```

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -107,6 +109,36 @@ public async Task<IActionResult> GetHostStatus([FromServices] IScriptHostManager
...
+                {
+                    foreach (var channel in channels)
+                    {
+                        var processInfo = new WorkerProcessInfo()
+                        {
+                            ProcessId = channel.Id,
+                            ExeName = channel.WorkerProcess.Process.StartInfo.FileName,
+                            DebugEngine = "python",
```
liliankasem: Can we use a constant here?

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -107,6 +109,36 @@ public async Task<IActionResult> GetHostStatus([FromServices] IScriptHostManager
...
+                            IsEligibleForOpenInBrowser = false
+                        };
+                        processes.Add(processInfo);
+                    }
+                }
+            }
+
+            return Ok(processes);
```
liliankasem: Could something go wrong here where we might need to be able to return a different http code?

src/WebJobs.Script.WebHost/Models/WorkerProcessInfo.cs
```diff
@@ -0,0 +1,38 @@
...
+        public string ExeName { get; set; }
+
+        /// <summary>
+        /// Gets or sets the debug engine string assosiated with the worker process.
+        /// </summary>
+        [JsonProperty(PropertyName = "debugEngine", DefaultValueHandling = DefaultValueHandling.Ignore)]
+        public string DebugEngine { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether the worker process Is eligible to be opened in browser.
```
satvu: nit: lowercase "Is" in this comment

src/WebJobs.Script.WebHost/Controllers/HostController.cs
soninaren: Updated the code to return 503 if we are not able to get required services. Otherwise we would simply return empty collection

src/WebJobs.Script/Utility.cs
```diff
@@ -101,6 +101,23 @@ internal static Attribute GetHierarchicalAttributeOrNull(MethodInfo method, Type
             return null;
         }
 
+        internal static string GetDebugEngineInfo(RpcWorkerConfig workerConfig, string runtime)
+        {
+            if (runtime == RpcWorkerConstants.DotNetIsolatedLanguageWorkerName)
+            {
+                if (workerConfig.Description.DefaultRuntimeVersion.Contains(RpcWorkerConstants.NetFramework))
+                {
+                    return RpcWorkerConstants.NetCoreDebugEngine;
```
liliankasem: Shouldn't this be `NetFrameworkDebugEngine` and the else statement should be NetCore?

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -107,6 +110,51 @@ public async Task<IActionResult> GetHostStatus([FromServices] IScriptHostManager
...
+                foreach (var pair in webhostChannelDictionary)
+                {
+                    var workerChannel = await pair.Value.Task;
+                    channels.Add(workerChannel);
+                }
+            }
+
+            if (Utility.TryGetHostService(scriptHostManager, out IJobHostRpcWorkerChannelManager jobHostLanguageWorkerChannelManager))
+            {
+                channels.AddRange(jobHostLanguageWorkerChannelManager.GetChannels(workerRuntime));
```
liliankasem: Would there be any concern at all if we **don't** get `jobHostLanguageWorkerChannelManager` and the channels from it? Is that normal?

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -107,6 +110,51 @@ public async Task<IActionResult> GetHostStatus([FromServices] IScriptHostManager
             return Ok(status);
         }
 
+        [HttpGet]
+        [Route("admin/host/workerProcesses")]
```
mathewc: Similar to the Kudu API that exists /api/processes [here](https://github.com/projectkudu/kudu/blob/1df1ce7dad1d941b2658ed0a66f9d09ee6f96a8f/Kudu.Services.Web/App_Start/NinjectServices.cs#L538).

Wondering if our API shouldn't be admin/host/processes and also include the host process?

soninaren: This API is mainly there to assist with remote debugging for dotnet isolated workers. Not sure if adding the host process itself would be of use.

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -107,6 +110,51 @@ public async Task<IActionResult> GetHostStatus([FromServices] IScriptHostManager
...
+            }
+
+            if (Utility.TryGetHostService(scriptHostManager, out IJobHostRpcWorkerChannelManager jobHostLanguageWorkerChannelManager))
+            {
+                channels.AddRange(jobHostLanguageWorkerChannelManager.GetChannels(workerRuntime));
+            }
+
+            foreach (var channel in channels)
+            {
+                var processInfo = new WorkerProcessInfo()
```
mathewc: There's already an IHostProcessMonitor (defined in the WebJobs SDK [here](https://github.com/Azure/azure-webjobs-sdk/blob/7bfcd07e520210c9cc092e6c1ac9413fc5c3d178/src/Microsoft.Azure.WebJobs.Host/Scale/DefaultHostProcessMonitor.cs#L15)) that we use for scale status that is all wired up in the Functions Host to monitor both the host and all child processes. Since all this bookeeping is already handled by that component, perhaps it makes sense to just have it also implement a new **IHostProcessProvider** interface that can be queried for all the process info. That way this is centralized. It internally has a ProcessMonitor for all processes, and we could define method(s) to return that info.

src/WebJobs.Script.WebHost/Controllers/HostController.cs
soninaren: It should be okay. In some cases the jobHostLanguageWorkerChannelManager may not have channels any as they may have initialized at webhost level. Also, in case script host is restarting when the API call is made it is possible that we may not get jobHostLanguageWorkerChannelManager.

src/WebJobs.Script/Utility.cs
soninaren: yes it should be

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -107,6 +110,51 @@ public async Task<IActionResult> GetHostStatus([FromServices] IScriptHostManager
             return Ok(status);
         }
 
+        [HttpGet]
```
mathewc, line 117: Regarding security, having this be GET will allow Readers to list this info, which is OK because the Kudu APIs for listing processes are also allowed for Readers as well. Just pointing this out so we keep this in mind that we don't extend this API to include any secrets - if we did, we'd have to lock it down to only Contributors.

soninaren: That is a good point. Let me add a comment to the method outlining this information..

src/WebJobs.Script.WebHost/Controllers/HostController.cs
mathewc: That's the current use case, but since we're adding a new API my vote would be to make it general and return host process as well, so in the future if we decide we need to get host process also we don't have to add yet another separate API for that.

src/WebJobs.Script.WebHost/Controllers/HostController.cs
soninaren: It would be neat to get the `IHostProcessProvider` implementation working. I think we would not get all the information we need from the proposed `IHostProcessProvider` though, mainly the debug engine info. I am getting some information from the WorkerMetadata object received as part of worker init response. Also, since the current implementation is not introducing any new bookkeeping components, we can easily swap the underlying components without impacting the API if we decide to have `IHostProcessProvider` implementation. Given the sensitive nature of this workitem i would like to merge the current implementation if there are no major red flags.

src/WebJobs.Script.WebHost/Controllers/HostController.cs
soninaren: That makes sense. Let me try and include the host process information as well. Do you know we have record that anywhere?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -139,7 +139,7 @@ internal class GrpcWorkerChannel : IRpcWorkerChannel, IDisposable
 
         public IWorkerProcess WorkerProcess => _rpcWorkerProcess;
 
-        internal RpcWorkerConfig Config => _workerConfig;
+        public RpcWorkerConfig WorkerConfig => _workerConfig;
```
madelinegordon, line 142: just wondering - what is the reason for the change in the modifier?

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -107,6 +110,56 @@ public async Task<IActionResult> GetHostStatus([FromServices] IScriptHostManager
             return Ok(status);
         }
 
+        /// <summary>
+        ///  Regarding security, GET will allow Readers to list this info, which is OK.
```
madelinegordon:
```suggestion
        ///  Currently, anyone in Reader role can access this information. 
        /// If this API is extended to include any secrets, it will need to be 
        /// locked down to only Contributor roles.  
```

src/WebJobs.Script.WebHost/Models/WorkerProcessInfo.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// Gets or sets the worker process id.
+        /// </summary>
+        [JsonProperty(PropertyName = "processId", DefaultValueHandling = DefaultValueHandling.Ignore)]
+        public int ProcessId { get; set; }
+
+        /// <summary>
+        /// Gets or sets the name of the worker process executable.
+        /// </summary>
+        [JsonProperty(PropertyName = "exeName", DefaultValueHandling = DefaultValueHandling.Ignore)]
+        public string ExeName { get; set; }
```
madelinegordon: nit - naming suggestion: either `ProcessName` or `ExecutableName`

src/WebJobs.Script/Workers/Rpc/RpcWorkerConstants.cs
```diff
@@ -73,5 +73,9 @@ public static class RpcWorkerConstants
         public const string PythonThreadpoolThreadCount = "PYTHON_THREADPOOL_THREAD_COUNT";
         public const string PSWorkerInProcConcurrencyUpperBound = "PSWorkerInProcConcurrencyUpperBound";
         public const string DefaultConcurrencyLimit = "1000";
+
+        public const string NetCoreDebugEngine = ".NETCore";
```
madelinegordon:
```suggestion
        public const string DotNetCoreDebugEngine = ".NETCore";
```

src/WebJobs.Script/Workers/Rpc/RpcWorkerConstants.cs
```diff
@@ -73,5 +73,9 @@ public static class RpcWorkerConstants
...
+        public const string NetFrameworkDebugEngine = ".NETFramework";
```
madelinegordon:
```suggestion
        public const string DotNetFrameworkDebugEngine = ".NETFramework";
```

src/WebJobs.Script/Workers/Rpc/RpcWorkerConstants.cs
```diff
@@ -73,5 +73,9 @@ public static class RpcWorkerConstants
...
+        public const string NetFramework = "Framework";
```
madelinegordon:
```suggestion
        public const string DotNetFramework = "Framework";
```

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
soninaren: Needed to expose worker config to the hostController. For that i had to add a property in the interface `IRpcWorkerChannel`.

src/WebJobs.Script/Utility.cs
```diff
@@ -101,6 +101,23 @@ internal static Attribute GetHierarchicalAttributeOrNull(MethodInfo method, Type
             return null;
         }
 
+        internal static string GetDebugEngineInfo(RpcWorkerConfig workerConfig, string runtime)
+        {
+            if (runtime == RpcWorkerConstants.DotNetIsolatedLanguageWorkerName)
```
kshyju: Should this be a case insensitive check?

src/WebJobs.Script.WebHost/Models/WorkerProcessInfo.cs
```diff
@@ -0,0 +1,38 @@
...
+        public int ProcessId { get; set; }
+
+        /// <summary>
+        /// Gets or sets the name of the worker process executable.
+        /// </summary>
+        [JsonProperty(PropertyName = "executableName", DefaultValueHandling = DefaultValueHandling.Ignore)]
+        public string ExecutableName { get; set; }
+
+        /// <summary>
+        /// Gets or sets the debug engine string assosiated with the worker process.
```
kshyju: nit: typo

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -107,6 +110,56 @@ public async Task<IActionResult> GetHostStatus([FromServices] IScriptHostManager
...
+            }
+
+            foreach (var channel in channels)
+            {
+                var processInfo = new WorkerProcessInfo()
+                {
+                    ProcessId = channel.WorkerProcess.Process.Id,
+                    ExecutableName = channel.WorkerProcess.Process.ProcessName,
+                    DebugEngine = Utility.GetDebugEngineInfo(channel.WorkerConfig, workerRuntime),
+                    IsEligibleForOpenInBrowser = false
```
kshyju, line 170: Confused about this property. How this will be used when value of this prop is true/false ?

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -107,6 +110,56 @@ public async Task<IActionResult> GetHostStatus([FromServices] IScriptHostManager
...
+            {
+                channels.AddRange(jobHostLanguageWorkerChannelManager.GetChannels(workerRuntime));
+            }
+
+            foreach (var channel in channels)
+            {
+                var processInfo = new WorkerProcessInfo()
+                {
+                    ProcessId = channel.WorkerProcess.Process.Id,
+                    ExecutableName = channel.WorkerProcess.Process.ProcessName,
```
kshyju: Should this property be called `ProcessName` ?

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -107,6 +110,56 @@ public async Task<IActionResult> GetHostStatus([FromServices] IScriptHostManager
...
+        [Route("admin/host/workerProcesses")]
+        [Authorize(Policy = PolicyNames.AdminAuthLevelOrInternal)]
+        public async Task<IActionResult> GetWorkerProcesses([FromServices] IScriptHostManager scriptHostManager)
+        {
+            if (!Utility.TryGetHostService(scriptHostManager, out IWebHostRpcWorkerChannelManager webHostLanguageWorkerChannelManager))
+            {
+                return StatusCode(StatusCodes.Status503ServiceUnavailable);
+            }
+
+            var processes = new List<WorkerProcessInfo>();
```
kshyju: Can move this to line 147 with capacity in the constructor. 

```c#
var processes = new List<WorkerProcessInfo>(channels.Count);
```

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -107,6 +110,56 @@ public async Task<IActionResult> GetHostStatus([FromServices] IScriptHostManager
...
+            string workerRuntime = _environment.GetFunctionsWorkerRuntime();
+
+            var webhostChannelDictionary = webHostLanguageWorkerChannelManager.GetChannels(workerRuntime);
+
+            List<IRpcWorkerChannel> channels = new List<IRpcWorkerChannel>();
+            if (webhostChannelDictionary is not null)
+            {
+                foreach (var pair in webhostChannelDictionary)
+                {
+                    var workerChannel = await pair.Value.Task;
```
kshyju: Do we need to await in this sequential loop? Should we consider adding the task to a collection and do `await Task.WhenAll` outside the loop?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -360,6 +360,9 @@ internal void FunctionEnvironmentReloadResponse(FunctionEnvironmentReloadRespons
             _workerChannelLogger.LogDebug("Received FunctionEnvironmentReloadResponse from WorkerProcess with Pid: '{0}'", _rpcWorkerProcess.Id);
 
             LogWorkerMetadata(res.WorkerMetadata);
+
+            _workerConfig.Description.DefaultRuntimeVersion = _workerConfig.Description.DefaultRuntimeVersion ?? res.WorkerMetadata.RuntimeName;
```
kshyju: I [see ](https://github.com/Azure/azure-functions-host/blob/0266f360d578f06901b5021152ac7fb18d0ec160/test/WebJobs.Script.Tests/Workers/Rpc/RpcWorkerConfigTests.cs#L514)the `DefaultRuntimeVersion` prop stores the version value (ex:3.7) for the language (ex: python). But this assignment is setting it to the runtime name (ex: python). Is that expected?

src/WebJobs.Script.WebHost/Controllers/HostController.cs
soninaren: This property is required by the VS, it is so far false for all languages.

soninaren: I think there is a different type of debugger attachment action that would happen if it is true.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
soninaren: Have added a new runtime name property to resolve this

src/WebJobs.Script.WebHost/Controllers/HostController.cs
soninaren: done

src/WebJobs.Script.WebHost/Controllers/HostController.cs
kshyju: Got it. Thanks!

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -360,6 +360,10 @@ internal void FunctionEnvironmentReloadResponse(FunctionEnvironmentReloadRespons
             _workerChannelLogger.LogDebug("Received FunctionEnvironmentReloadResponse from WorkerProcess with Pid: '{0}'", _rpcWorkerProcess.Id);
 
             LogWorkerMetadata(res.WorkerMetadata);
+
+            _workerConfig.Description.DefaultRuntimeVersion = _workerConfig.Description.DefaultRuntimeVersion ?? res.WorkerMetadata.RuntimeVersion;
+            _workerConfig.Description.DefaultRuntimeName = _workerConfig.Description.DefaultRuntimeName ?? res.WorkerMetadata.RuntimeName;
```
kshyju, line 365: Question: So we are only updating these properties if they are null. Env reload response will have the correct runtime value (after specialization). In that case, shouldn't we just override irrespective of existing value being null or not?

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -107,6 +109,77 @@ public async Task<IActionResult> GetHostStatus([FromServices] IScriptHostManager
...
+
+            var hostProcess = Process.GetCurrentProcess();
+            List<FunctionProcesses.FunctionProcessInfo> processes = new()
+            {
+                new FunctionProcesses.FunctionProcessInfo()
+                {
+                    ProcessId = hostProcess.Id,
+                    DebugEngine = RpcWorkerConstants.DotNetCoreDebugEngine,
+                    IsEligibleForOpenInBrowser = false,
+                    ProcessName = hostProcess.ProcessName
```
kshyju, line 135: Is the process name going to be dotnet.exe or the dll name? In that case, how does the caller know it is the host process or the oop worker process (in the case of dotnet isolated/powershell)? Is that a concern?

kshyju (approved):

src/WebJobs.Script.WebHost/Controllers/HostController.cs
soninaren: This would likely be dotnet.exe. At the moment there is no differentiation between host and worker process. My assumption is that VS will provide the user with the prompt to select a process in case there is ambiguity. We can update the API incase a clear distinction is needed.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
soninaren: We don't know for sure if all workers are sending the version and runtime name as part of metadata response. If we override this info and worker is only providing this info as part of workerConfig then we would be overriding it with null.

satvu (approved):

Issue #9037, opened 2023-01-17
Add logs for host startup cancellation


liliankasem:
In a recent CRI (355435073), we saw a host startup cancellation log but there was no exception or other details/information about what might have caused this cancellation. In this specific case, it was most likely due to to customer code startup issues which resulted in the host taking too long to start and therefore being cancelled.

![Screenshot 2023-01-17 at 2 13 58 PM](https://user-images.githubusercontent.com/2198905/213024181-b14ec3dc-0f9c-48f6-90c4-8689f867ff28.png)

Looking at the host codebase, we see that this is most likely not a host cancellation as we did not see this log `Initialization cancellation requested by runtime` at all in the Function Logs. Therefore, it looks like the exception is being swallowed as we're not logging or throwing at this point, which could potentially lead to a scenario where the host is running but not working (in a state that cannot be recovered from without a FunctionApp restart).

```csharp
try
{
    await StartHostAsync(tokenSource.Token);
}
catch (OperationCanceledException)
{
    if (cancellationToken.IsCancellationRequested)
    {
        _logger.ScriptHostServiceInitCanceledByRuntime(); // "Initialization cancellation requested by runtime"
        throw;
    }

    // If the exception was triggered by our loop cancellation token, just ignore as
    // it doesn't indicate an issue.
}
```

Code: [StartAsync](https://github.com/Azure/azure-functions-host/blob/901d8b6d5859aad426c65053567e1f6fd5b98275/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs#L177-L191)

- [x] Add logging to the above code so we can see what exception might be happening
- [x] Discuss if we should throw the exception here / see if we should be handling this differently

fabiocav:
Done. Resolved by #9091

PR #9046, opened 2023-01-24
Add a log for the StartAsync cancellation exception 

liliankasem:

Part 1 of #9037

In a recent CRI (355435073), we saw a host startup cancellation log but there was no exception or other details/information about what might have caused this cancellation. In this specific case, it was most likely due to to customer code startup issues which resulted in the host taking too long to start and therefore being cancelled.

Screenshot 2023-01-17 at 2 13 58 PM

Looking at the host codebase, we see that this is most likely not a host cancellation as we did not see this log `Initialization cancellation requested by runtime` at all in the Function Logs. Therefore, it looks like the exception is being swallowed as we're not logging or throwing at this point, which could potentially lead to a scenario where the host is running but not working (in a state that cannot be recovered from without a FunctionApp restart).

In this PR we are adding logs to the above code so we can see what exception might be happening

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
+++ b/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
@@ -178,8 +178,10 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
             {
                 await StartHostAsync(tokenSource.Token);
             }
-            catch (OperationCanceledException)
+            catch (OperationCanceledException ex)
             {
+                _logger.LogWarning("Host startup operation has been cancelled", ex);
+
                 if (cancellationToken.IsCancellationRequested)
                 {
                     _logger.ScriptHostServiceInitCanceledByRuntime();


satvu (approved):

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -178,8 +178,10 @@ public async Task StartAsync(CancellationToken cancellationToken)
             {
                 await StartHostAsync(tokenSource.Token);
             }
-            catch (OperationCanceledException)
+            catch (OperationCanceledException ex)
             {
+                _logger.LogWarning("Host startup operation has been cancelled", ex);
```
kshyju, line 183: The comments at 191 indicates that getting into this catch block may not be an issue in certain cases. I wonder this warning messages causes confusion.

kshyju: looks like @fabiocav is the author of that comment :)

liliankasem: Yup - Fabio and I recently discussed this and he disagrees with that comment xD This is why part 2 of this issue is to have a design discussion if we should be handling this differently. For now we want more data on what is causing the cancellation

kshyju (approved):

PR #9061, opened 2023-01-31
Emit diagnostic event upon Host Id collision

### LINKED ISSUE #8949: Host ID collision error should be a diagnostic event
For background on hostID collision see:
https://github.com/Azure/azure-functions-host/wiki/Host-IDs

#### Repro steps

1. Create two function apps with a deliberate host ID collision, e.g. make them share a storage account and have a name that is longer than 32 characters.
2. Restart the apps once they are both created, to force the collision check to occur

#### Expected behavior

One or both the apps should fail to start with the collision error e.g. A collision for Host ID 'asdf' was detected in the configured storage account. For more information, see https://aka.ms/functions-hostid-collision.

This error should be displayed in the portal, as an error banner. Most importantly, the link in the above error message needs to be rendered so that the user can follow it and take the recommended actions. For this to happen, it needs to be logged as a diagnostic event.

#### Actual behavior

No mention of host ID truncation appears in the portal experience. The user has to go to detailed logs such as app insights to find the details of the truncation error. They may see other error noise that is just a symptom, of the host shutting down to the collision, such as the below message, but this won't lead them to understanding the root of the problem.

```
2022-12-01T23:26:41.158 [Error] Failed to start a new language worker for runtime: dotnet-isolated.
System.Threading.Tasks.TaskCanceledException : A task was canceled.
   at async Microsoft.Azure.WebJobs.Script.Grpc.GrpcWorkerChannel.StartWorkerProcessAsync(CancellationToken cancellationToken) at /_/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs : 271
```

madelinegordon:

resolves #8949 

### Pull request checklist

    Added documentation [here](https://learn.microsoft.com/en-us/azure/azure-functions/errors-diagnostics/diagnostic-events/azfd0004)

### DIFF
--- a/src/WebJobs.Script/Host/HostIdValidator.cs
+++ b/src/WebJobs.Script/Host/HostIdValidator.cs
@@ -133,7 +133,7 @@ namespace Microsoft.Azure.WebJobs.Script
                 // so anything other than Warning is treated as Error.
                 _logger.LogError(message);
 
-                DiagnosticEventLoggerExtensions.LogDiagnosticEventError(_logger, HostIdCollisionErrorCode, message, HostIdCollisionHelpLink, new Exception(message));
+                DiagnosticEventLoggerExtensions.LogDiagnosticEventError(_logger, HostIdCollisionErrorCode, message, HostIdCollisionHelpLink, new ArgumentException(message));
 
                 _applicationLifetime.StopApplication();
             }
--- a/src/WebJobs.Script/Host/HostIdValidator.cs
+++ b/src/WebJobs.Script/Host/HostIdValidator.cs
@@ -29,6 +29,7 @@ namespace Microsoft.Azure.WebJobs.Script
     {
         public const string BlobPathFormat = "ids/usage/{0}";
         private const string HostIdCollisionErrorCode = "AZFD004";
+        private const string HostIdCollisionHelpLink = "https://go.microsoft.com/fwlink/?linkid=2224100";
         private const LogLevel DefaultLevel = LogLevel.Error;
 
         private readonly IEnvironment _environment;
@@ -124,7 +125,7 @@ namespace Microsoft.Azure.WebJobs.Script
             {
                 _logger.LogWarning(message);
 
-                DiagnosticEventLoggerExtensions.LogDiagnosticEventInformation(_logger, HostIdCollisionErrorCode, message, "https://aka.ms/functions-hostid-collision");
+                DiagnosticEventLoggerExtensions.LogDiagnosticEventInformation(_logger, HostIdCollisionErrorCode, message, HostIdCollisionHelpLink);
             }
             else
             {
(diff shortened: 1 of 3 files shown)


src/WebJobs.Script/Host/HostIdValidator.cs
```diff
@@ -117,15 +120,21 @@ private void HandleCollision(string hostId)
...
             }
             else
             {
-                // we only allow Warning/Error levels to be specified, so anything other than
-                // Warning is treated as Error
+                // We only allow Warning/Error levels to be specified,
+                // so anything other than Warning is treated as Error.
                 _logger.LogError(message);
+
+                DiagnosticEventLoggerExtensions.LogDiagnosticEventError(_logger, HostIdCollisionErrorCode, message, HostIdCollisionHelpLink, new Exception(message));
```
liliankasem: Can we use a more specific exception here? We can either create a custom one (if one doesn't already exist) or use an existing exception type i.e. argument exception or invalid operation

https://www.tutorialsteacher.com/csharp/csharp-exception

liliankasem (approved):

test/WebJobs.Script.Tests.Integration/Host/HostIdValidatorTests.cs
```diff
@@ -127,10 +127,13 @@ public async Task ValidateHostIdUsageAsync_Collision_WarningLevel_Logs()
             await _hostIdValidator.ValidateHostIdUsageAsync(_testHostId);
 
             var logs = _loggerProvider.GetAllLogMessages();
-            var log = logs.Single();
+            var log = logs[0];
             Assert.Equal(LogLevel.Warning, log.Level);
             Assert.Equal(string.Format(Resources.HostIdCollisionFormat, _testHostId), log.FormattedMessage);
 
+            var diagLog = logs[1];
```
surgupta-msft, line 135: Should we assert on `logs` length before accessing values?

satvu (approved):

test/WebJobs.Script.Tests.Integration/Host/HostIdValidatorTests.cs
kashimiz: Seconding this suggestion.

test/WebJobs.Script.Tests.Integration/Host/HostIdValidatorTests.cs
```diff
@@ -127,10 +127,13 @@ public async Task ValidateHostIdUsageAsync_Collision_WarningLevel_Logs()
...
+            Assert.Contains("collision", diagLog.FormattedMessage);
```
kashimiz: Is there a reason we're not validating against the whole error message, including the host ID?

madelinegordon: We can validate against that too. It is the same message as in the log above; I will add that validation.

kashimiz (approved):

PR #9084, opened 2023-02-08
Update active profile when multiple runtimes use worker description profiles 

### LINKED ISSUE #9052: Profile manager code sets incorrect value for _activeProfile variable causing specialization code to fail for dotnet-isolated
Specialization code path fails for dotnet-isolated function apps when the workerconfig does not have any entries in the "profiles" section.

Current behavior observed during specialization.

1. [WebHostRpcWorkerChannelManager.SpecializeAsync](https://github.com/Azure/azure-functions-host/blob/7570b6daba86ae58931f116694e78700f4e86cd4/src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs#L118) method is invoked.
2. [UsePlaceholderChannel](https://github.com/Azure/azure-functions-host/blob/7570b6daba86ae58931f116694e78700f4e86cd4/src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs#L143) method gets called with "dotnet-isolated" as `workerRuntime` parameter.
3. [WorkerProfileManager.IsCorrectProfileLoaded](https://github.com/Azure/azure-functions-host/blob/v4.15.1/src/WebJobs.Script/Workers/Profiles/WorkerProfileManager.cs#L93) gets called and it returns `false`
   
image

4. `UsePlaceholderChannel` returns `false` and the caller code shuts down the placeholder channel.
https://github.com/Azure/azure-functions-host/blob/7570b6daba86ae58931f116694e78700f4e86cd4/src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs#L132-L137

Synced with @shreyas-gopalakrishna  and he thinks the value of `_activeProfile` is incorrect here. It should be ideally null/empty because the worker config for dotnet-isolated did not have any profiles. Further debugging shown that the guid stored in that variable is the id from the java worker config.

The `LoadWorkerDescriptionFromProfiles` method sets the value for  `_activeProfile` every time it is called. In my case, it was initially called for java and that did set the GUID and then later called for dotnet-isolated. The `GetEvaluatedProfile` method returned `false`, so the code path to set/reset the value of `_activeProfile` did not get executed. The value from java stayed there for the variable.

https://github.com/Azure/azure-functions-host/blob/7570b6daba86ae58931f116694e78700f4e86cd4/src/WebJobs.Script/Workers/Profiles/WorkerProfileManager.cs#L61-L73

shreyas-gopalakrishna:
### Update active profile when multiple runtimes use worker description profiles 

resolves #9052

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script/Workers/Profiles/WorkerProfileManager.cs
+++ b/src/WebJobs.Script/Workers/Profiles/WorkerProfileManager.cs
@@ -18,7 +18,7 @@ namespace Microsoft.Azure.WebJobs.Script.Workers
         private readonly IEnvironment _environment;
         private readonly IEnumerable<IWorkerProfileConditionProvider> _conditionProviders;
 
-        private string _activeProfile = string.Empty;
+        private Dictionary<string, string> _activeProfiles;
         private Dictionary<string, List<WorkerDescriptionProfile>> _profiles;
 
         public WorkerProfileManager(ILogger<WorkerProfileManager> logger, IEnvironment environment)
@@ -26,6 +26,7 @@ namespace Microsoft.Azure.WebJobs.Script.Workers
             _environment = environment ?? throw new ArgumentNullException(nameof(environment));
             _logger = logger ?? throw new ArgumentNullException(nameof(logger));
 
+            _activeProfiles = new Dictionary<string, string>();
             _profiles = new Dictionary<string, List<WorkerDescriptionProfile>>();
             _conditionProviders = new List<IWorkerProfileConditionProvider>
             {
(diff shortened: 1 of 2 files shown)


kshyju (approved):

src/WebJobs.Script/Workers/Profiles/WorkerProfileManager.cs
```diff
@@ -97,7 +98,13 @@ public bool IsCorrectProfileLoaded(string workerRuntime)
             {
                 profileId = profile.ProfileId;
             }
-            return _activeProfile.Equals(profileId);
+
+            var activeProfile = string.Empty;
+            if (!_activeProfiles.TryGetValue(workerRuntime, out activeProfile))
+            {
+                activeProfile = string.Empty;
```
liliankasem, line 105: it's already being set to an empty string on 102, we shouldn't need to do that again? You could probably not use the if statement here if `TryGetValue` returning false doesn't lead to a different action

shreyas-gopalakrishna: When TryGetValue returns false, activeProfile is set to null. So setting it to empty. I have corrected the initialization.

liliankasem (approved):

PR #9091, opened 2023-02-10
improving handling and logging of exceptions thrown during external startup

### LINKED ISSUE #9030: Detect if Host Startup errors are related to custom DI
We should be able to show a clear message that the failure from startup was from the host, extensions or it may be from customer's code issue. This will allow debug process to be more targetted. 

 the startup class.

            var config = new ConfigurationBuilder()

                           .AddJsonFile("local.settings.json", optional: true, reloadOnChange: true)

                           .AddEnvironmentVariables()

                           .Build();

at Microsoft.Commerce.Agreements.ElectronicSigning.DocuSign.DocuSignClient..ctor(String accountId,String clientId,String clientSecret,String clientUserId,Byte[] privateKey,String restApiUri,String oAuthUri,Int32 tokenExpiryInHours) at E:\agent_work\12\s\DocuSignClient\DocuSignClient.cs : 108   at Microsoft.Commerce.Agreements.ElectronicSigning.Factory.CreateESignClient(ESignClientConfiguration configuration) at E:\agent_work\12\s\SignedDocumentOrchestratorFunction\Factory.cs : 24   at Microsoft.Commerce.Agreements.ElectronicSigning.Startup.Configure(IFunctionsHostBuilder builder) at E:\agent_work\12\s\SignedDocumentOrchestratorFunction\Startup.cs : 36   at Microsoft.Azure.Functions.Extensions.DependencyInjection.FunctionsStartup.Configure(IWebJobsBuilder builder)

For issues in production, please check for a solution to common issues in the Azure portal before opening a bug. In the Azure portal, navigate to your function app, select `Diagnose and solve problems` from the left, and view relevant dashboards before opening your issue.

Please provide the following:

- Timestamp:
- Function App version:
- Function App name:
- Function name(s) (as appropriate):
- Invocation ID:
- Region:

#### Repro steps

#### Expected behavior

#### Actual behavior

Provide a description of the actual behavior observed.

#### Known workarounds

Provide a description of any known workarounds.

#### Related information 

* Programming language used 
* Links to source
* Bindings used

brettsam:
Fixes #9030, #9037

Today we let exceptions from startup classes flow through to the `WebJobsScriptHostService`. Because the customer's host has not been created, they have no logger and no direct way to surface errors other than via the portal's "LastError" check.

This now adds two more ways to send errors:
- via the DiagnosticEvent logging infrastructure
- via App Insights (if there's a connection string present)

By wrapping the exception in our own type, it also eliminates the special handling of the OperationCanceledExcpetion that was causing us to sit idle if a startup class threw this excpetion. Now we'll mark the host as in Error and retry (with backoff) as we would with any other startup exception.

The experience after this is in:
- Exceptions are logged as a `Microsoft.Azure.WebJobs.Script.ExternalStartupException`.
- These exceptions will be retried in case they are transient. It's hard for us to know, but something like pulling config from a db in your startup may be transient, so the host will retry with a backoff forever just in case.
- If Application Insights is enabled, we will attempt to create a `TelemetryClient` and send the exception there.
- A diagnostic exception will be logged, which means you will get a banner like this:

image

which then leads to this

image

which ultimately links to https://learn.microsoft.com/en-us/azure/azure-functions/errors-diagnostics/diagnostic-events/azfd0005

Hopefully this helps diagnose previously silent startup problems...

### DIFF
--- a/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventLogger.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventLogger.cs
@@ -6,19 +6,38 @@ using System.Collections.Generic;
 using System.Linq;
 using Microsoft.Azure.WebJobs.Script.Config;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
 
 namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 {
-    public class DiagnosticEventLogger : ILogger
+    public class DiagnosticEventLogger : ILogger, IDisposable
     {
         private readonly IDiagnosticEventRepositoryFactory _diagnosticEventRepositoryFactory;
         private readonly IEnvironment _environment;
-        private IDiagnosticEventRepository _diagnosticEventRepository;
+        private readonly IDisposable _standbyChangeHandler;
 
-        public DiagnosticEventLogger(IDiagnosticEventRepositoryFactory diagnosticEventRepositoryFactory, IEnvironment environment)
+        private IDiagnosticEventRepository _diagnosticEventRepository;
+        private bool _isEnabled;
+
+        public DiagnosticEventLogger(IDiagnosticEventRepositoryFactory diagnosticEventRepositoryFactory, IEnvironment environment,
+            IOptionsMonitor<StandbyOptions> standbyOptions)
         {
             _diagnosticEventRepositoryFactory = diagnosticEventRepositoryFactory;
             _environment = environment;
+
+            SetEnabledState(standbyOptions.CurrentValue);
+
+            // Re-evaluate whether this is enabled after specialization
+            if (standbyOptions.CurrentValue.InStandbyMode)
+            {
+                _standbyChangeHandler = standbyOptions.OnChange(o => SetEnabledState(o));
+            }
+        }
+
+        private void SetEnabledState(StandbyOptions options)
+        {
+            _isEnabled = !options.InStandbyMode &&
+                !FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagDisableDiagnosticEventLogging, _environment);
         }
 
         public IDisposable BeginScope<TState>(TState state)
@@ -26,10 +45,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
             return null;
         }
 
-        public bool IsEnabled(LogLevel logLevel)
-        {
-            return !_environment.IsPlaceholderModeEnabled();
-        }
+        public bool IsEnabled(LogLevel logLevel) => _isEnabled;
 
         private bool IsDiagnosticEvent(IDictionary<string, object> state)
         {
@@ -52,6 +68,12 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
                 }
                 _diagnosticEventRepository.WriteDiagnosticEvent(DateTime.UtcNow, stateInfo[ScriptConstants.ErrorCodeKey].ToString(), logLevel, message, stateInfo[ScriptConstants.HelpLinkKey].ToString(), exception);
             }
+        }
+
+        public void Dispose()
+        {
+            _standbyChangeHandler?.Dispose();
+            (_diagnosticEventRepository as IDisposable)?.Dispose();
         }
     }
 }
--- a/src/WebJobs.Script.WebHost/Program.cs
+++ b/src/WebJobs.Script.WebHost/Program.cs
@@ -2,9 +2,7 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
-using System.Collections;
 using System.Linq;
-using System.Runtime.InteropServices;
 using System.Threading;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
@@ -79,6 +77,13 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
 
                     loggingBuilder.AddDefaultWebJobsFilters();
                     loggingBuilder.AddWebJobsSystem<WebHostSystemLoggerProvider>();
+
+                    if (!FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagDisableDiagnosticEventLogging))
+                    {
+                        loggingBuilder.Services.AddSingleton<ILoggerProvider, DiagnosticEventLoggerProvider>();
+                        loggingBuilder.Services.TryAddSingleton<IDiagnosticEventRepository, DiagnosticEventTableStorageRepository>();
+                        loggingBuilder.Services.TryAddSingleton<IDiagnosticEventRepositoryFactory, DiagnosticEventRepositoryFactory>();
+                    }
                 })
                 .UseStartup<Startup>();
         }
--- a/src/WebJobs.Script.WebHost/Program.cs
+++ b/src/WebJobs.Script.WebHost/Program.cs
@@ -77,13 +77,6 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
 
                     loggingBuilder.AddDefaultWebJobsFilters();
                     loggingBuilder.AddWebJobsSystem<WebHostSystemLoggerProvider>();
-
-                    if (!FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagDisableDiagnosticEventLogging))
-                    {
-                        loggingBuilder.Services.AddSingleton<ILoggerProvider, DiagnosticEventLoggerProvider>();
-                        loggingBuilder.Services.TryAddSingleton<IDiagnosticEventRepository, DiagnosticEventTableStorageRepository>();
-                        loggingBuilder.Services.TryAddSingleton<IDiagnosticEventRepositoryFactory, DiagnosticEventRepositoryFactory>();
-                    }
                 })
                 .UseStartup<Startup>();
         }
--- a/src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
+++ b/src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
@@ -157,6 +157,13 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
                     services.AddSingleton(hostingConfigOptions);
                     services.AddSingleton(hostingConfigOptionsMonitor);
 
+                    if (!FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagDisableDiagnosticEventLogging))
+                    {
+                        services.AddSingleton<ILoggerProvider, DiagnosticEventLoggerProvider>();
+                        services.TryAddSingleton<IDiagnosticEventRepository, DiagnosticEventTableStorageRepository>();
+                        services.TryAddSingleton<IDiagnosticEventRepositoryFactory, DiagnosticEventRepositoryFactory>();
+                    }
+
                     ConfigureRegisteredBuilders(services, rootServiceProvider);
                 });
 
--- a/src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
+++ b/src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
@@ -93,9 +93,8 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
 
                     if (!FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagDisableDiagnosticEventLogging))
                     {
+                        // Services that this logger depends on are registered at the WebHost level.
                         loggingBuilder.Services.AddSingleton<ILoggerProvider, DiagnosticEventLoggerProvider>();
-                        loggingBuilder.Services.TryAddSingleton<IDiagnosticEventRepository, DiagnosticEventTableStorageRepository>();
-                        loggingBuilder.Services.TryAddSingleton<IDiagnosticEventRepositoryFactory, DiagnosticEventRepositoryFactory>();
                     }
 
                     ConfigureRegisteredBuilders(loggingBuilder, rootServiceProvider);
--- a/src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
+++ b/src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
@@ -93,7 +93,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
 
                     if (!FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagDisableDiagnosticEventLogging))
                     {
-                        // Services that this logger depends on are registered at the WebHost level.
+                        // services that this depends on are registered at the webhost level
                         loggingBuilder.Services.AddSingleton<ILoggerProvider, DiagnosticEventLoggerProvider>();
                     }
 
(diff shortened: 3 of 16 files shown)


src/WebJobs.Script.WebHost/Program.cs
```diff
@@ -79,6 +77,13 @@ public static IWebHostBuilder CreateWebHostBuilder(string[] args = null)
 
                     loggingBuilder.AddDefaultWebJobsFilters();
                     loggingBuilder.AddWebJobsSystem<WebHostSystemLoggerProvider>();
+
+                    if (!FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagDisableDiagnosticEventLogging))
```
brettsam: @soninaren -- these were only available from the ScriptHost before (I think) -- any objection to me also registering them at the WebHost level? It doesn't seem like there'd be any conflicts.

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -47,6 +51,9 @@ public static class ScriptHostBuilderExtensions
         private const string DelayedConfigurationActionKey = "MS_DelayedConfigurationAction";
         private const string ConfigurationSnapshotKey = "MS_ConfigurationSnapshot";
 
+        private const string ExternalStartupErrorDiagnosticId = "AZFD0005";
```
mathewc: I think we only have one other place in host now (HostIdValidator) that is emitting diagnostic logs? I wonder if we shouldn't move the diagnostic IDs and help links into a central file so they're all in one place.

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -209,7 +216,15 @@ public static IHostBuilder AddScriptHostCore(this IHostBuilder builder, ScriptAp
                     // Only set our external startup if we're not suppressing host initialization
                     // as we don't want to load user assemblies otherwise.
                     var locator = context.Properties.GetAndRemove<ScriptStartupTypeLocator>(StartupTypeLocatorKey);
-                    webJobsBuilder.UseExternalStartup(locator, webJobsBuilderContext, loggerFactory);
+
+                    try
+                    {
+                        webJobsBuilder.UseExternalStartup(locator, webJobsBuilderContext, loggerFactory);
```
kshyju, line 219: Question for my learning: What if the external startup class/method has some infinite loop or similar thing (Similar to that recent CRI you helped with)? I looked at the `UseExternalStartup` method and do not see any kind of timeout mechanism we implemented there to throw an exception if the customer startup code did not finish execution in X number of seconds.

src/WebJobs.Script.WebHost/Program.cs
kshyju: Unrelated to change in this PR, But I noticed we [call string.split ](https://github.com/Azure/azure-functions-host/blob/28e9550320b13137f4eaf16483ebcdb630699506/src/WebJobs.Script/Config/FeatureFlags.cs#L22-L23) and `Contains` every time we call the `IsEnabled` method. Wonder there is a way for us to cache that to may be a hashset?  Also should we be reading that env variable every time that method is called?

kshyju: I see we are doing explicit disable via feature flag. Should we be doing explicit enable? Not all customers use customer startup classes. So Why not those small number of customers explicitly enable this feature flag as needed, so that no cold start impact for everyone.

brettsam: Couple responses to set up the explanation:
- this diagnostic event logging already exists and is used to surface any errors to the customer directly via the portal (the portal can ask us if there are any events, then pops them up)
- today, this is only supported in the ScriptHost. It's set up here: https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs#L94-L99
- but I need this in the WebHost -- specifically to surface issues where the ScriptHost cannot be built b/c of exceptions
- this won't affect cold start -- everything will already be constructed and warmed up by the time this happens

All that being said -- this feature flag check won't even work today b/c of placeholder mode. We'll need this to be *always* registered, then filter it away if it's explicitly disabled. I need to refactor all of this a bit. I'm working on that now.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventLogger.cs
```diff
@@ -6,31 +6,47 @@
...
+        public DiagnosticEventLogger(IDiagnosticEventRepositoryFactory diagnosticEventRepositoryFactory, IEnvironment environment,
+            IOptionsMonitor<StandbyOptions> standbyOptions)
         {
             _diagnosticEventRepositoryFactory = diagnosticEventRepositoryFactory;
             _environment = environment;
+
+            SetEnabledState(standbyOptions.CurrentValue);
+
+            // Re-evaluate whether this is enabled after specialization
+            if (standbyOptions.CurrentValue.InStandbyMode)
```
brettsam, line 31: We use this pattern elsewhere in the host to re-evaluate state after specialization (I swear I didn't just make it up here): https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticListenerService.cs#L31-L36

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventLogger.cs
```diff
@@ -53,5 +69,11 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
                 _diagnosticEventRepository.WriteDiagnosticEvent(DateTime.UtcNow, stateInfo[ScriptConstants.ErrorCodeKey].ToString(), logLevel, message, stateInfo[ScriptConstants.HelpLinkKey].ToString(), exception);
             }
         }
+
+        public void Dispose()
+        {
+            _standbyChangeHandler?.Dispose();
+            (_diagnosticEventRepository as IDisposable)?.Dispose();
```
brettsam, line 76: noticed we were never disposing this -- so taking care of that

src/WebJobs.Script.WebHost/Program.cs
brettsam: I just updated this in my recent commit -- no longer check it every time; instead we cache and only re-evaluate if specialization has occurred (which is really the only way it can change)

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
brettsam: Yeah, let me do that.

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
brettsam: I don't think we explicitly handle this ourselves anywhere. It's worth investigating whether we can add something here (in a follow-up).

src/WebJobs.Script.WebHost/Program.cs
soninaren: No objections on my end.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventLogger.cs
jviau: This seems like a perfect use of options `OnChange`. However, you could alternatively just evaluate this within `IsEnabled()`:

``` CSharp
public bool IsEnabled(LogLevel logLevel) => !_options.CurrentValue.InStandbyMode &&
    !FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagDisableDiagnosticEventLogging, _environment);
```

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
brettsam: Done.

src/WebJobs.Script.WebHost/Program.cs
brettsam: We could probably come back to this and improve that `FeatureFlags` thing entirely to cache internally. But it is fairly scarcely used, I believe so probably not a huge priority.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventLogger.cs
brettsam: That was my initial approach, but here's what I went through in my head:
- `FeatureFlags.IsEnabled()` does some string parsing -- which I wanted to avoid doing on every log message.
- So I started trying to refactor `FeatureFlags` so it cached internally in a `HashSet<string>` for fast lookup and no parsing.
- But `IEnvironment` values can change (tests can do it; specialization can do it), which means that we'd need a way to also store/compare that original string and invalidate the cache if it ever changed.
- You can also technically pass in different `IEnviroment` objects on every call -- so we'd need to maintain a cache for each one I'd imagine.
- Would we need to introduce locks to make sure it was thread-safe? It looks like we only access this in about a dozen places, so maybe I could see if it mattered to any of them. (note we've had a *ton* of threading issues during specialization in the past so we're very cautious about stuff changing mid-specialization).

Then I felt like I was way too into the weeds and went back to a pattern we already use elsewhere.

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
brettsam: looks like i've already filed that here: https://github.com/Azure/azure-functions-host/issues/8703

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventLogger.cs
```diff
@@ -6,31 +6,47 @@
...
-        public IDisposable BeginScope<TState>(TState state)
+        private void SetEnabledState(StandbyOptions options)
         {
-            return null;
+            _isEnabled = !options.InStandbyMode &&
+                !FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagDisableDiagnosticEventLogging, _environment);
         }
 
-        public bool IsEnabled(LogLevel logLevel)
+        public IDisposable BeginScope<TState>(TState state)
```
liliankasem, line 43: Do we need this method if all it does is return null? Or is there more to it?

liliankasem (approved):

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventLogger.cs
kshyju: I think we need it since we are implementing `ILogger`

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventLogger.cs
```diff
@@ -6,31 +6,47 @@
...
+                _standbyChangeHandler = standbyOptions.OnChange(o => SetEnabledState(o));
+            }
         }
 
-        public IDisposable BeginScope<TState>(TState state)
+        private void SetEnabledState(StandbyOptions options)
         {
-            return null;
+            _isEnabled = !options.InStandbyMode &&
+                !FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagDisableDiagnosticEventLogging, _environment);
```
kshyju, line 40: Q: Do we need the feature flag check here? Looks like we do this check before registering this type to DI container (in `WebHostServiceCollectionExtension.AddWebJobsScriptHost`).

src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
```diff
@@ -93,9 +93,8 @@ public static class WebScriptHostBuilderExtension
 
                     if (!FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagDisableDiagnosticEventLogging))
                     {
+                        // services that this depends on are registered at the webhost level
                         loggingBuilder.Services.AddSingleton<ILoggerProvider, DiagnosticEventLoggerProvider>();
```
kshyju, line 97: DiagnosticEventLoggerProvider is also registered in `WebHostServiceCollectionExtension.AddWebJobsScriptHost`. Do we need to register it here as well ?

src/WebJobs.Script/Diagnostics/DiagnosticEventConstants.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Microsoft.Azure.WebJobs.Script
+{
+    internal static class DiagnosticEventConstants
+    {
+        public const string HostIdCollisionErrorCode = "AZFD004";
+        public const string HostIdCollisionHelpLink = "https://go.microsoft.com/fwlink/?linkid=2224100";
+
+        public const string ExternalStartupErrorCode = "AZFD0005";
+        public const string ExternalStartupErrorHelpLink = "https://go.microsoft.com/fwlink/?linkid=2224847";
```
kshyju, line 12: this link gives 404. Do we need to setup something somewhere else for this to point to the right resource?

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -476,5 +498,33 @@ private static void AddProcessRegistry(IServiceCollection services)
...
+        private static void RecordAndThrowExternalStartupException(string message, Exception ex, ILoggerFactory loggerFactory)
+        {
+            var startupEx = new ExternalStartupException(message, ex);
+
+            var logger = loggerFactory.CreateLogger(LogCategories.Startup);
+            logger.LogDiagnosticEventError(DiagnosticEventConstants.ExternalStartupErrorCode, message, DiagnosticEventConstants.ExternalStartupErrorHelpLink, startupEx);
+
+            // Send the error to App Insights if possible. This is happening during ScriptHost construction so we
+            // have no existing TelemetryClient to use. Create a one-off client and flush it ASAP.
+            string appInsightsConnStr = Environment.GetEnvironmentVariable(EnvironmentSettingNames.AppInsightsConnectionString);
```
kshyju: I remember Jacob asking similar question before. Don't remember the answer. But should we use `IConfiguration` abstraction to read the env variable/app setting value?  
I recently chose to do that for a change I am working on instead of reading from Environment directly. Wondering I should change back to this approach.

kshyju (approved):

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventLogger.cs
brettsam: it has to be checked here as well:
- Ideally, we'd never even register this, which is what the DI container checks for. This works in non-placeholder cases.
- However, in placeholder mode in the webhost, we have to always register it, then "disable" it after specialization and the env vars have changed. There's currently no way to adjust the registered WebHost services after we specialize... they need to all be registered at startup.

src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
brettsam: Yeah -- `ILoggerProviders` don't get shared between WebHost and JobHost. The way it works is:
- If the JobHost asks for service `IFoo` and there aren't any, it checks the parent WebHost for its version of `IFoo`. This is why a bunch of services can be shared between the two DI containers.
- However, logging requests an `IEnumerable<ILoggerProvider>` -- this exists in JobHost already, so it will not ask the WebHost for its version. `IOptions<T>` has the same limitation -- b/c .NET will always create you an `IOptions<T>`, it doesn't flow through to the parent at all. This has bitten us a bunch in the past. So there's certain types (loggers, options, etc) that don't flow from parent -> child like other types.

src/WebJobs.Script/Diagnostics/DiagnosticEventConstants.cs
brettsam: PR for this is here -- https://github.com/MicrosoftDocs/azure-docs-pr/pull/227355. Should be released by the time this goes out.

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
brettsam: You almost always want to use `IConfiguration`. In some places we use the `IEnvironment` abstraction b/c we don't want to allow you to override some system env vars -- we want to go "straight to the source". But this case is an odd one b/c it's happening on startup so we don't have any DI container, and we may not have an `IConfiguration`.

There's 2 scenarios where this can get called b/c of an exception:
1. During `ConfigureAppConfiguration` -- at this point, I don't think we can trust `IConfiguration` as it hasn't been fully built. That's originally why I went straight to `Environment`.
2. During `ConfigureServices` -- at this point, at least, `IConfiguration` has been built, so we can use it. 

I made a change to do this -- at least this way, if someone is populating that connection string from KeyVault or something, they'll resolve correctly in `ConfigureServices`.

This is mostly a "best effort" logging attempt -- something's gone horribly wrong at startup and we just want to send it wherever we can.

PR #9094, opened 2023-02-14
Add error details to RunFromPackage failure log

### LINKED ISSUE #8922: Include failure details in log for Run From Package startup failures
When an app using "run from package" fails due to problems downloading the zip package, the platform writes a file "FAILED TO INITIALIZE RUN FROM PACKAGE.txt", which the host looks for, and when found will refuse to start (code [here](https://github.com/Azure/azure-functions-host/blob/a68b37b1d667e5d9a3152da935c3b80384da51da/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs#L203)). The content of this log file actually includes reasons for the failure. We should log those details as well to aid in issue diagnosis.

We should also confirm that this error message gets written to the user's Application Insights (if it isn't already) to reduce support cases.

madelinegordon:

resolves #8922 

### Pull request checklist

### Additional information

Added file contents to RunFromPackage failure log.

### DIFF
--- a/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
+++ b/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
@@ -200,9 +200,11 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
                 // Check for marker file indicating a zip package failure, and if found stop the application.
                 // We never want to run with an incorrect file system.
                 string path = Path.Combine(_applicationHostOptions.CurrentValue.ScriptPath, ScriptConstants.RunFromPackageFailedFileName);
+
                 if (File.Exists(path))
                 {
-                    _logger.LogError($"Shutting down host due to presence of {path}");
+                    string errorMessage = File.ReadAllText(path);
+                    _logger.LogError($"Shutting down host due to presence of {path}. Error details: {errorMessage}");
                     _applicationLifetime.StopApplication();
                 }
             }
--- a/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
+++ b/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
@@ -203,14 +203,20 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
 
                 if (File.Exists(path))
                 {
+                    string errorPrefix = $"Shutting down host due to presence of {path}.";
+                    string errorSuffix = string.Empty;
                     try
                     {
                         string fileContent = File.ReadAllText(path);
-                        _logger.LogError($"Shutting down host due to presence of {path}. File content: {fileContent}");
+                        errorSuffix = $" File content: {fileContent}";
                     }
                     catch (Exception ex)
                     {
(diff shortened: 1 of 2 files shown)


src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -200,9 +200,11 @@ private void CheckFileSystem()
                 // Check for marker file indicating a zip package failure, and if found stop the application.
                 // We never want to run with an incorrect file system.
                 string path = Path.Combine(_applicationHostOptions.CurrentValue.ScriptPath, ScriptConstants.RunFromPackageFailedFileName);
+
                 if (File.Exists(path))
                 {
-                    _logger.LogError($"Shutting down host due to presence of {path}");
+                    string errorMessage = File.ReadAllText(path);
+                    _logger.LogError($"Shutting down host due to presence of {path}. Error details: {errorMessage}");
```
kshyju: Should we use "File content" instead of Error details to be more explicit ?

kshyju (approved):

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -200,9 +200,11 @@ private void CheckFileSystem()
                 // Check for marker file indicating a zip package failure, and if found stop the application.
                 // We never want to run with an incorrect file system.
                 string path = Path.Combine(_applicationHostOptions.CurrentValue.ScriptPath, ScriptConstants.RunFromPackageFailedFileName);
+
                 if (File.Exists(path))
                 {
-                    _logger.LogError($"Shutting down host due to presence of {path}");
+                    string errorMessage = File.ReadAllText(path);
```
kshyju: What happens if the `ReadAllText` throws any of the documented exception here? Do we have an outer catch which logs that exception?

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -200,9 +200,11 @@ private void CheckFileSystem()
...
+                    _logger.LogError($"Shutting down host due to presence of {path}. Error details: {errorMessage}");
                     _applicationLifetime.StopApplication();
                 }
```
kshyju, line 222: If ZipDeploymentAppSettingsExist is truthy and this file does not exist, should we log a warning?

liliankasem: yes i think we should

liliankasem (approved):

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
madelinegordon: From my understanding, `ZipDeploymentAppSettingsExist = true` and this file not existing is the expected behavior/happy path

madelinegordon: And we would only stop the host if `FAILED TO INITIALIZE RUN FROM PACKAGE.txt` exists.

mathewc (approved):

PR #9098, opened 2023-02-15
fixing exception when accessing a disposed Process

brettsam:
The exception handling introduced by #8917 can inadvertently re-throw the exception by accessing the disposed `Process?.Id` in the exception message.

I've also added a test that failed before fixing the message.


test/WebJobs.Script.Tests/Workers/Rpc/RpcWorkerProcessTests.cs
```diff
@@ -214,5 +214,29 @@ public void CreateWorkerProcess_AddsHostingConfiguration()
...
+            startInfo.WindowStyle = ProcessWindowStyle.Hidden;
+            startInfo.FileName = "cmd.exe";
+            startInfo.Arguments = $"ls";
+            process.StartInfo = startInfo;
+            process.Start();
+
+            _rpcWorkerProcess.Process = process;
+            _rpcWorkerProcess.Dispose();
+
+            _rpcWorkerProcess.WaitForProcessExitInMilliSeconds(1);
```
brettsam, line 232: This was throwing an exception before my change.

liliankasem (approved):

kshyju (approved):

PR #9139, opened 2023-03-07
Adding null check to fix the issue with input/output binding for the sdk type scenario

### LINKED ISSUE #1362: [SDK Binding - Blob] Output binding does not work with trigger binding using an SDK-type
Using an output binding with a trigger bindings whilst binding to an SDK-type (stream/blobClient etc.) leads to an error.

```c#
        [Function(nameof(BlobTriggerStreamTest))]
        [BlobOutput("test-output-stream-dotnet-isolated/output.txt")]
        public async Task<string> BlobTriggerStreamTest(
            [BlobTrigger("test-trigger-stream-dotnet-isolated")] Stream stream,
            FunctionContext context)
        {
            using var blobStreamReader = new StreamReader(stream);
            string content = await blobStreamReader.ReadToEndAsync();
            _logger.LogInformation(".NET Blob trigger function processed a blob.\n Content: " + content);
            return content;
        }
```

Error:

`The 'BlobTriggerStreamTest' function is in error: Value cannot be null. (Parameter 'input').`

surgupta-msft:

resolves https://github.com/Azure/azure-functions-dotnet-worker/issues/1362

Can repro the issue with InputBinding as well, hence added tests for both input and output binding

```csharp
        [Function(nameof(SampleFunction))]
        public void SampleFunction(
            [BlobTrigger("stream-trigger/sample1.txt")] string content,
            [BlobInput("input-container/file.txt")] string myBlob,
            FunctionContext context)
        {
            _logger.LogInformation("Trigger content: {content}", content);
            _logger.LogInformation("Blob content: {content}", myBlob);
        }
```

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script/Description/Workers/WorkerFunctionDescriptorProvider.cs
+++ b/src/WebJobs.Script/Description/Workers/WorkerFunctionDescriptorProvider.cs
@@ -129,11 +129,16 @@ namespace Microsoft.Azure.WebJobs.Script.Description
                 {
                     return attribute.GetType()
                                     .GetProperties(BindingFlags.Instance | BindingFlags.Public)
-                                    .Any(prop => prop.PropertyType == typeof(string) && _expressionRegex.IsMatch((string)prop.GetValue(attribute)));
+                                    .Any(prop => prop.PropertyType == typeof(string) && IsMatch((string)prop.GetValue(attribute)));
                 }
             }
 
             return false;
+        }
+
+        private bool IsMatch(string value)
+        {
+            return string.IsNullOrEmpty(value) ? false : _expressionRegex.IsMatch(value);
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/WebJobs.Script/Description/Workers/WorkerFunctionDescriptorProvider.cs
```diff
@@ -129,11 +129,16 @@ internal bool BindingAttributeContainsExpression(IEnumerable<FunctionBinding> bi
...
+                                    .Any(prop => prop.PropertyType == typeof(string) && IsMatch((string)prop.GetValue(attribute)));
                 }
             }
 
             return false;
         }
+
+        private bool IsMatch(string value)
+        {
+            return string.IsNullOrEmpty(value) ? false : _expressionRegex.IsMatch(value);
```
surgupta-msft, line 141: The specific scenario for the bug is - if there is no expression in `BlobPath` and the connection is null then the regex.IsMatch(connection) will throw null exception. If there is an expression in `BlobPath` then it gets returned as true and the empty connection is not evaluated so we won't see the error.

liliankasem (commented):
Nice! Thanks for fixing. Will do some local E2E tests before I approve but will get this done by EOD.

Edit: tested this E2E and works as expected :)

liliankasem (approved):

satvu (approved):
lgtm

PR #9145, opened 2023-03-10
First pieces of native placeholder support

brettsam:
Some changes:
- Adds the native worker preview package. If in placeholder mode, the host will now start up this placeholder worker.
- Only specializes if `WEBSITE_USE_PLACEHOLDER_DOTNETISOLATED` is set to `1`. Otherwise, kills the placeholder worker and spins up a new one. This is for the opt-in preview only -- eventually this won't be required.
- Passes a new environment variable `FUNCTIONS_APPLICATION_DIRECTORY` to the worker process -- this points to the ScriptRoot. This was previously only available in grpc messages, but the dotnet-isolated placeholder needs this information earlier.

### DIFF
--- a/src/WebJobs.Script/Host/FunctionMetadataManager.cs
+++ b/src/WebJobs.Script/Host/FunctionMetadataManager.cs
@@ -52,7 +52,10 @@ namespace Microsoft.Azure.WebJobs.Script
             // services that change with the scope of the script host.
             scriptHostManager.ActiveHostChanged += (s, e) =>
             {
-                InitializeServices();
+                if (e.NewHost is not null)
+                {
+                    InitializeServices();
+                }
             };
         }
 
--- a/src/WebJobs.Script/Host/FunctionMetadataManager.cs
+++ b/src/WebJobs.Script/Host/FunctionMetadataManager.cs
@@ -14,7 +14,6 @@ using Microsoft.Azure.WebJobs.Script.Workers;
 using Microsoft.Azure.WebJobs.Script.Workers.Http;
 using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
 using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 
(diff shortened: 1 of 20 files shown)


fabiocav (approved):

src/WebJobs.Script/Host/FunctionMetadataManager.cs
```diff
@@ -51,9 +50,12 @@ public class FunctionMetadataManager : IFunctionMetadataManager
 
             // Every time script host is re-intializing, we also need to re-initialize
             // services that change with the scope of the script host.
-            scriptHostManager.HostInitializing += (s, e) =>
+            scriptHostManager.ActiveHostChanged += (s, e) =>
```
brettsam, line 53: For future reference -- I changed this due to a race during specialization while I was testing this. The point here is to refresh services on the new host, and ActiveHostChanged is much more appropriate for this and removes the race. The ActiveHostChanged API was added after this code, so it wasn't an option at the time.

Also, we have another component that does something similar, which makes me more comfortable (https://github.com/Azure/azure-functions-host/blob/8ceb05a89a4337f07264d4991545538a3e8b58a0/src/WebJobs.Script/Config/ActiveHostConfigurationSource.cs#L32)

src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
```diff
@@ -124,7 +124,7 @@ public async Task SpecializeAsync()
 
             if (_workerRuntime != null && rpcWorkerChannel != null)
             {
-                if (UsePlaceholderChannel(_workerRuntime))
+                if (UsePlaceholderChannel(rpcWorkerChannel))
```
kshyju, line 127: When `WEBSITE_USE_PLACEHOLDER_DOTNETISOLATED` is not truthy and specialization happens. Don't we need to update the worker config so that the executable path is set to the customer payload dll? Currently it looks like it is still going to use the worker config which points to the FunctionsNetHost.exe.

kshyju (approved):

src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
brettsam: I think I mentioned offline -- but will capture it here as well. Once specialization occurs, the host will load the worker's FunctionMetadataProvider assembly from its payload. This has code to overwrite the configuration for dotnet-isolated to point to the customer's app directly. So, we don't actually need any code here -- it just happens automatically because of this:

https://github.com/Azure/azure-functions-host/blob/6b3e19463055d44cae61cff32c93eb58f0cf790e/src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs#L105-L106

PR #9149, opened 2023-03-14
[Enhanced Http] Move Http Proxying logic to its own service

satvu:

resolves #issue_for_this_pr

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
@@ -31,7 +31,6 @@ using Microsoft.Azure.WebJobs.Script.ManagedDependencies;
 using Microsoft.Azure.WebJobs.Script.Workers;
 using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
 using Microsoft.Azure.WebJobs.Script.Workers.SharedMemoryDataTransfer;
-using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Yarp.ReverseProxy.Forwarder;
@@ -85,7 +84,7 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
         private bool _isSharedMemoryDataTransferEnabled;
         private bool? _cancelCapabilityEnabled;
         private bool _isWorkerApplicationInsightsLoggingEnabled;
-        private IHttpForwarder _httpForwarder;
+        private IHttpProxyService _httpProxyService;
 
         private System.Timers.Timer _timer;
 
@@ -102,7 +101,7 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
             ISharedMemoryManager sharedMemoryManager,
             IOptions<WorkerConcurrencyOptions> workerConcurrencyOptions,
             IOptions<FunctionsHostingConfigOptions> hostingConfigOptions,
-            IHttpForwarder httpForwarder)
+            IHttpProxyService httpProxyService)
         {
             _workerId = workerId;
             _eventManager = eventManager;
@@ -118,7 +117,7 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
             _processInbound = state => ProcessItem((InboundGrpcEvent)state);
             _hostingConfigOptions = hostingConfigOptions;
 
-            _httpForwarder = httpForwarder;
+            _httpProxyService = httpProxyService;
             _workerCapabilities = new GrpcCapabilities(_workerChannelLogger);
 
             if (!_eventManager.TryGetGrpcChannels(workerId, out var inbound, out var outbound))
@@ -738,27 +737,7 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
 
                 if (!string.IsNullOrEmpty(_workerCapabilities.GetCapabilityState(RpcWorkerConstants.EnableHttpProxying)))
                 {
-                    var handler = new SocketsHttpHandler();
-                    var invoker = new HttpMessageInvoker(handler);
-                    var options = new ForwarderRequestConfig();
-                    HttpRequest httpRequest = context.Inputs.FirstOrDefault(i => i.Val is HttpRequest).Val as HttpRequest;
-                    HttpContext httpContext = httpRequest.HttpContext;
-
-                    // add IsHttpProxying to HttpContext Items (verify this is internal or not?)
-                    httpContext.Items.Add("IsHttpProxying", bool.TrueString);
-
-                    // add invocation id as correlation id
-                    // TODO: add "invocation-id" as a constant somewhere / maybe find a better name
-                    httpRequest.Headers.TryAdd("invocation-id", context.ExecutionContext.InvocationId.ToString());
-
-                    var port = Environment.GetEnvironmentVariable("Azure_Functions_HttpProxyingPort");
-
-                    if (port is null)
-                    {
-                        port = "5555";
-                    }
-
-                    var aspNetTask = _httpForwarder.SendAsync(httpContext, "http://localhost:" + port, invoker, options);
+                    var aspNetTask = _httpProxyService.Forward(context);
 
                     context.Properties.Add("HttpProxyingTask", aspNetTask);
                 }
--- a/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
@@ -735,7 +735,7 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
                     context.CancellationToken.Register(() => SendInvocationCancel(invocationRequest.InvocationId));
                 }
 
-                if (!string.IsNullOrEmpty(_workerCapabilities.GetCapabilityState(RpcWorkerConstants.EnableHttpProxying)))
+                if (!string.IsNullOrEmpty(_workerCapabilities.GetCapabilityState(RpcWorkerConstants.EnableHttpProxying)) && context.FunctionMetadata.IsHttpInFunction())
                 {
                     var aspNetTask = _httpProxyService.Forward(context);
 
--- a/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
@@ -735,7 +735,7 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
                     context.CancellationToken.Register(() => SendInvocationCancel(invocationRequest.InvocationId));
                 }
 
-                if (!string.IsNullOrEmpty(_workerCapabilities.GetCapabilityState(RpcWorkerConstants.EnableHttpProxying)) && context.FunctionMetadata.IsHttpInFunction())
+                if (!string.IsNullOrEmpty(_workerCapabilities.GetCapabilityState(RpcWorkerConstants.EnableHttpProxying)) && context.FunctionMetadata.IsHttpTriggerFunction())
                 {
                     var aspNetTask = _httpProxyService.Forward(context);
 
--- a/src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
+++ b/src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
@@ -35,7 +35,23 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
 
         public ValueTask<ForwarderError> Forward(ScriptInvocationContext context)
         {
+            if (context is null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            if (context.Inputs is null)
+            {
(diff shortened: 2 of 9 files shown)


src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
```diff
@@ -0,0 +1,52 @@
...
+
+            // TODO: Update this logic. Port should come through configuration.
+            var port = Environment.GetEnvironmentVariable("AZURE_FUNCTIONS_HTTP_PROXY_PORT") ?? "5555";
+
+            _proxyEndpoint = "http://localhost:" + port;
+        }
+
+        public ValueTask<ForwarderError> Forward(ScriptInvocationContext context)
+        {
+            HttpRequest httpRequest = context.Inputs.FirstOrDefault(i => i.Val is HttpRequest).Val as HttpRequest;
```
liliankasem, line 55: Do we need to check if content or context.Inputs is null

src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
```diff
@@ -0,0 +1,52 @@
...
+            HttpContext httpContext = httpRequest.HttpContext;
+
+            httpContext.Items.Add("IsHttpProxying", bool.TrueString);
```
liliankasem, line 64: nit: make `IsHttpProxying` a const string

satvu: noted in [todo comment](https://github.com/Azure/azure-functions-host/blob/7efb10cc1db55dbb319b6f16e889550d4612e2d5/src/WebJobs.Script.WebHost/Middleware/FunctionInvocationMiddleware.cs#L67) at the other location it's used - haven't proceeded yet since this was a "hack" design decision but if we decide to keep it will do.

src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
satvu: Let me check what this would mean in our scenario to make sure any error message makes sense / is actionable. 

edit: I don't know when `context` would be null - we get it from [here](https://github.com/Azure/azure-functions-host/blob/b13595e6fab9051fc5603a38495ff062b5d33912/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs#L693) in the `GrpcWorkerChannel` when a function is invoked. I also don't believe we would have a context without an input (a Function would have at least a trigger). 

Following up on what behavior we want if an HttpRequest isn't found in inputs.

src/WebJobs.Script/Extensions/FunctionMetadataExtensions.cs
```diff
@@ -33,6 +33,16 @@ public static bool IsHttpInAndOutFunction(this FunctionMetadata metadata)
             return false;
         }
 
+        public static bool IsHttpInFunction(this FunctionMetadata metadata)
+        {
+            if (metadata.InputBindings.Any(b => string.Equals("httptrigger", b.Type, StringComparison.OrdinalIgnoreCase)))
```
satvu: **note**: follows same logic as the above method `IsHttpInAndOutFunction`. I can turn this string into a constant for just this file (like "IsDirect" and "IsCodeless" in the vars above), but we prob want this cleanup for dev too, not just this feature branch.

brettsam: Why does the `HttpInAndOut` only accept a single input and a single output binding? Is it trying to only filter for a "pure" http trigger?

src/WebJobs.Script/Extensions/FunctionMetadataExtensions.cs
```diff
@@ -33,6 +33,16 @@ public static bool IsHttpInAndOutFunction(this FunctionMetadata metadata)
...
+            {
+                return true;
+            }
+
+            return false;
```
brettsam: nit -- i think you could simplify this to 
```
return metadata.InputBindings.Any(b => string.Equals("httptrigger", b.Type, StringComparison.OrdinalIgnoreCase));
```

src/WebJobs.Script/Extensions/FunctionMetadataExtensions.cs
```diff
@@ -33,6 +33,16 @@ public static bool IsHttpInAndOutFunction(this FunctionMetadata metadata)
             return false;
         }
 
+        public static bool IsHttpInFunction(this FunctionMetadata metadata)
```
brettsam: i know you're following the other method naming -- but it seems more accurate to call this `IsHttpTriggerFunction`...

src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
brettsam: I agree with @liliankasem -- you should be defensive here and check nulls before accessing.

brettsam (approved):

src/WebJobs.Script/Extensions/FunctionMetadataExtensions.cs
satvu: @brettsam It's only used in one location [here](https://github.com/Azure/azure-functions-host/blob/6eea6da0952857b4cc64339f329cdf61432b5815/src/WebJobs.Script/Workers/Http/DefaultHttpWorkerService.cs#L62) and I think it's for custom handlers? 🤷‍♀️

src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
satvu: Added null checks but the only one I have a clear error message for is the case of `HttpRequest` not being found - I think the case of `context` being non-null and `context.Inputs` being null would have some message similar to "Function context must have valid inputs/input bindings, but none were found" or something easy to understand. 

Less certain about `context` being null but currently throwing an `ArgumentNullException`. Lmk if y'all have suggestions @liliankasem @brettsam @fabiocav

src/WebJobs.Script/Extensions/FunctionMetadataExtensions.cs
brettsam: ah that would make sense -- it's only for "http only" scenarios.

src/WebJobs.Script/Extensions/FunctionMetadataExtensions.cs
```diff
@@ -24,15 +26,20 @@ public static bool IsHttpInAndOutFunction(this FunctionMetadata metadata)
 
             BindingMetadata inputBindingMetadata = metadata.InputBindings.ElementAt(0);
             BindingMetadata outputBindingMetadata = metadata.OutputBindings.ElementAt(0);
-            if (string.Equals("httptrigger", inputBindingMetadata.Type, StringComparison.OrdinalIgnoreCase) &&
-                string.Equals("http", outputBindingMetadata.Type, StringComparison.OrdinalIgnoreCase))
+            if (string.Equals(HttpTriggerKey, inputBindingMetadata.Type, StringComparison.OrdinalIgnoreCase) &&
```
satvu, line 29: Adding the constants b/c this was bothering me, but since it does touch unrelated components (custom handler code), I can revert this section and keep the constants just for `IsHttpTriggerFunction`.

src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
```diff
@@ -0,0 +1,68 @@
...
+        private HttpMessageInvoker _messageInvoker;
+        private ForwarderRequestConfig _forwarderRequestConfig;
+        private string _proxyEndpoint;
+
+        public DefaultHttpProxyService(IHttpForwarder httpForwarder)
+        {
+            _httpForwarder = httpForwarder;
+
+            var handler = new SocketsHttpHandler();
+            _messageInvoker = new HttpMessageInvoker(handler);
```
kshyju: `HttpMessageInvoker` implements `IDisposable`. We need to properly dispose it.

kshyju: Same for `SocketsHttpHandler` as well.

src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
```diff
@@ -0,0 +1,68 @@
...
+                throw new InvalidOperationException($"Cannot proxy an HttpTrigger Function without an input of type {nameof(HttpRequest)}.");
+            }
+
+            HttpContext httpContext = httpRequest.HttpContext;
+
+            httpContext.Items.Add("IsHttpProxying", bool.TrueString);
+
+            // add invocation id as correlation id
+            // TODO: add "invocation-id" as a constant somewhere / maybe find a better name
+            httpRequest.Headers.TryAdd("invocation-id", context.ExecutionContext.InvocationId.ToString());
```
kshyju, line 68: Should we use a prefix here for headers we add to the request? For ex: AFD adds headers with "x-" prefix. https://learn.microsoft.com/en-us/azure/frontdoor/front-door-http-headers-protocol#from-the-front-door-to-the-backend

src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
```diff
@@ -0,0 +1,68 @@
...
+        public ValueTask<ForwarderError> Forward(ScriptInvocationContext context)
+        {
+            if (context is null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            if (context.Inputs is null)
+            {
+                throw new ArgumentNullException(nameof(context.Inputs));
```
kshyju: TBH, I feel ArgumentNullException here for the sub property (of the argument passed in) is not the best.  May be InvalidOperationException?

src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
```diff
@@ -0,0 +1,68 @@
...
+        public DefaultHttpProxyService(IHttpForwarder httpForwarder)
+        {
+            _httpForwarder = httpForwarder;
+
+            var handler = new SocketsHttpHandler();
+            _messageInvoker = new HttpMessageInvoker(handler);
+            _forwarderRequestConfig = new ForwarderRequestConfig();
+
+            // TODO: Update this logic. Port should come through configuration.
+            var port = Environment.GetEnvironmentVariable("AZURE_FUNCTIONS_HTTP_PROXY_PORT") ?? "5555";
```
kshyju, line 32: Should we read this from `IConfiguraiton` / `IOptions<ConfigForThisType>` ? Do we have a configuration source which we populate from the host.json? That allows user to override this value via host.json if needed?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -736,29 +735,9 @@ internal async Task SendInvocationRequest(ScriptInvocationContext context)
                     context.CancellationToken.Register(() => SendInvocationCancel(invocationRequest.InvocationId));
                 }
 
-                if (!string.IsNullOrEmpty(_workerCapabilities.GetCapabilityState(RpcWorkerConstants.EnableHttpProxying)))
+                if (!string.IsNullOrEmpty(_workerCapabilities.GetCapabilityState(RpcWorkerConstants.EnableHttpProxying)) && context.FunctionMetadata.IsHttpTriggerFunction())
```
kshyju, line 738: Does this change depend on any change on the language worker side? Or is it safe to deploy this host changes when we are ready?

kshyju (commented):
I do not see a build from this PR changes.

src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
satvu: https://github.com/Azure/azure-functions-dotnet-worker/issues/1419 - wip, this implementation is for prototyping purposes.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
satvu: The changes on the host side and the code paths we are adding shouldn't be hit unless a worker has the capability set, but I think we're still far from deploying any of these host changes. These are just for the feature branch.

src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
satvu: Haven't gotten around to finalizing this name since I wasn't sure we were keeping this logic. At this point I think we are keeping it, so will take ur suggestion and try to finalize a name at the http sync.

Requires PR to change on worker side [here](https://github.com/Azure/azure-functions-dotnet-worker/blob/befaead487e044a3e391587faf6d7f6397595ab4/extensions/Worker.Extensions.Http.AspNetCore/src/AspNetMiddleware.cs#L26) as well.

satvu: A mix of headers use "X-" and "X-Functions" but we will revisit this before merging from feature branch to dev.

PR #9157, opened 2023-03-16
Adding new HostId generation logic for CV2

mathewc:
Introducing a new algorithm for host ID generation that doesn't truncate. Also updating sync triggers to include the HostId along with the sync triggers payload (initially for CV2 only). ScaleController has already been updated to use the payload HostId when specified by the host.

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
@@ -22,6 +22,7 @@ using Microsoft.Azure.WebJobs.Script.WebHost.Extensions;
 using Microsoft.Azure.WebJobs.Script.WebHost.Models;
 using Microsoft.Azure.WebJobs.Script.WebHost.Security;
 using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Newtonsoft.Json;
@@ -315,23 +316,22 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
             var triggersArray = new JArray(triggers);
             int count = triggersArray.Count;
 
+            // Form the base minimal result
+            string hostId = await _hostIdProvider.GetHostIdAsync(CancellationToken.None);
+            JObject result = GetMinimalPayload(hostId, triggersArray);
+
             if (!ArmCacheEnabled)
             {
-                // extended format is disabled - just return triggers
+                // extended format is disabled - just return minimal results
                 return new SyncTriggersPayload
                 {
-                    Content = JsonConvert.SerializeObject(triggersArray),
+                    Content = JsonConvert.SerializeObject(result),
                     Count = count
                 };
             }
 
-            // Add triggers to the payload
-            JObject result = new JObject();
-            result.Add("triggers", triggersArray);
-
             // Add all listable functions details to the payload
             JObject functions = new JObject();
-            string routePrefix = await WebFunctionsManager.GetRoutePrefix(hostOptions.RootScriptPath);
             var listableFunctions = _functionMetadataManager.GetFunctionMetadata().Where(m => !m.IsCodeless());
             var functionDetails = await WebFunctionsManager.GetFunctionMetadataResponse(listableFunctions, hostOptions, _hostNameProvider);
             result.Add("functions", new JArray(functionDetails.Select(p => JObject.FromObject(p))));
@@ -395,15 +395,12 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
 
             if (json.Length > ScriptConstants.MaxTriggersStringLength && !_environment.IsKubernetesManagedHosting())
             {
-                // The settriggers call to the FE enforces a max request size
-                // limit. If we're over limit, revert to the minimal triggers
-                // format.
+                // The settriggers call to the FE enforces a max request size limit.
+                // If we're over limit, revert to the minimal triggers format.
                 _logger.LogWarning($"SyncTriggers payload of length '{json.Length}' exceeds max length of '{ScriptConstants.MaxTriggersStringLength}'. Reverting to minimal format.");
-                return new SyncTriggersPayload
-                {
-                    Content = JsonConvert.SerializeObject(triggersArray),
-                    Count = count
-                };
+
+                var minimalResult = GetMinimalPayload(hostId, triggersArray);
+                json = JsonConvert.SerializeObject(minimalResult);
             }
 
             return new SyncTriggersPayload
@@ -411,6 +408,23 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
                 Content = json,
                 Count = count
             };
+        }
+
+        private JObject GetMinimalPayload(string hostId, JArray triggersArray)
+        {
+            JObject result = new JObject
+            {
+                { "triggers", triggersArray }
+            };
+
+            if (_environment.IsFlexConsumptionSku())
+            {
+                // Currently we're only sending the HostId for Flex Consumption. Eventually we'll do this for all SKUs.
+                // When the HostId is sent, ScaleController will use it directly rather than compute it itself.
+                result["hostId"] = hostId;
+            }
+
+            return result;
         }
 
         internal static async Task<JObject> GetHostJsonExtensionsAsync(IOptionsMonitor<ScriptApplicationHostOptions> applicationHostOptions, ILogger logger)
--- a/src/WebJobs.Script/Host/ScriptHostIdProvider.cs
+++ b/src/WebJobs.Script/Host/ScriptHostIdProvider.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Linq;
+using System.Security.Cryptography;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
@@ -33,6 +34,7 @@ namespace Microsoft.Azure.WebJobs.Script
             string hostId = _config[ConfigurationSectionNames.HostIdPath];
             if (hostId == null)
             {
+                // if the user hasn't configured an explicit ID, we generate the default ID.
                 HostIdResult result = GetDefaultHostId(_environment, _options.CurrentValue);
                 hostId = result.HostId;
                 if (result.IsTruncated && !result.IsLocal)
@@ -48,9 +50,22 @@ namespace Microsoft.Azure.WebJobs.Script
         {
             HostIdResult result = new HostIdResult();
 
-            // We're setting the default here on the newly created configuration
-            // If the user has explicitly set the HostID via host.json, it will overwrite
-            // what we set here
+            if (environment.IsFlexConsumptionSku())
+            {
+                // in Flex Consumption, we use a Guid based host ID without truncation.
+                string uniqueSlotName = environment?.GetAzureWebsiteUniqueSlotName();
+                byte[] hash;
+                using (MD5 md5 = MD5.Create())
+                {
+                    hash = md5.ComputeHash(Encoding.UTF8.GetBytes(uniqueSlotName));
+                }
+
+                result.HostId = new Guid(hash).ToString().Replace("-", string.Empty).ToLowerInvariant();
+
+                return result;
+            }
+
+            // Note: HostIds must conform to the rules documented here: https://learn.microsoft.com/en-us/azure/azure-functions/functions-app-settings#azurefunctionswebhost__hostid
             string hostId = null;
             if (environment.IsAppService() || environment.IsAnyKubernetesEnvironment())
             {
--- a/test/WebJobs.Script.Tests.Integration/Management/FunctionsSyncManagerTests.cs
+++ b/test/WebJobs.Script.Tests.Integration/Management/FunctionsSyncManagerTests.cs
@@ -139,6 +139,7 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.Managment
             _mockEnvironment.Setup(p => p.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebsiteName)).Returns((string)null);
             _mockEnvironment.Setup(p => p.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebsiteSlotName)).Returns((string)null);
             _mockEnvironment.Setup(p => p.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebJobsFeatureFlags)).Returns((string)null);
+            _mockEnvironment.Setup(p => p.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebsiteSku)).Returns((string)ScriptConstants.DynamicSku);
 
             _hostNameProvider = new HostNameProvider(_mockEnvironment.Object);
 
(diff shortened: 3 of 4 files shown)


src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -393,15 +397,17 @@ public async Task<SyncTriggersPayload> GetSyncTriggersPayload()
...
             {
-                // The settriggers call to the FE enforces a max request size
-                // limit. If we're over limit, revert to the minimal triggers
-                // format.
+                // The settriggers call to the FE enforces a max request size limit.
+                // If we're over limit, revert to the minimal triggers format.
                 _logger.LogWarning($"SyncTriggers payload of length '{json.Length}' exceeds max length of '{ScriptConstants.MaxTriggersStringLength}'. Reverting to minimal format.");
-                return new SyncTriggersPayload
+
+                var minimalResult = new JObject
```
mathewc: To ensure that we always return the HostID, we update the minimal format to not simply be an array. Consumers of the payload already know how to handle both object and array formats.

src/WebJobs.Script/ScriptConstants.cs
```diff
@@ -107,6 +107,7 @@ public static class ScriptConstants
         public const string XIdentityHeader = "X-IDENTITY-HEADER";
         public const string ContainerInstanceHeader = "fx-current-instance";
         public const string DynamicSku = "Dynamic";
+        public const string DynamicV2Sku = "DynamicV2";
```
mathewc: @balag0 - is this the right SKU constant?

test/WebJobs.Script.Tests/ScriptHostIdProviderTests.cs
```diff
@@ -94,6 +94,29 @@ public void GetDefaultHostId_AzureHost_ReturnsExpectedResult(string siteName, st
             Assert.False(result.IsLocal);
         }
 
+        [Theory]
+        [InlineData("myfunctionstestapp", "6606e225f163a70190e4f4357d3dad78")]
+        [InlineData("TEST-FUNCTIONS-XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX", "9dd2d6d54f1135ee6db6116f084b29bd")]
+        [InlineData("TEST-FUNCTIONS-XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXA", "c842e18afa2a84eac2818a956687a72e")]
```
mathewc, line 100: Cases like these would previously result in truncation and collisions

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -323,8 +323,12 @@ public async Task<SyncTriggersPayload> GetSyncTriggersPayload()
                 };
             }
 
-            // Add triggers to the payload
+            // Add HostId
             JObject result = new JObject();
+            string hostId = await _hostIdProvider.GetHostIdAsync(CancellationToken.None);
+            result.Add("hostId", hostId);
```
paulbatum: Assuming the scale controller will use this, are there extra things we need to check, such as whether sync triggers runs correctly after a slot swap?

mathewc: Can you detail the scenario you're concerned about not working? ScaleController is already dependent on sync trigger updates being correct/timely after swaps, otherwise it'd be incorrectly monitoring/scaling.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -323,8 +323,12 @@ public async Task<SyncTriggersPayload> GetSyncTriggersPayload()
...
+
+            // Add triggers to the payload
```
kshyju: Do we use the result of `await WebFunctionsManager.GetRoutePrefix(hostOptions.RootScriptPath);` in line 336? Can that be removed?

src/WebJobs.Script/Host/ScriptHostIdProvider.cs
```diff
@@ -48,11 +50,23 @@ internal static HostIdResult GetDefaultHostId(IEnvironment environment, ScriptAp
...
+                using (MD5 md5 = MD5.Create())
+                {
+                    hash = md5.ComputeHash(Encoding.UTF8.GetBytes(uniqueSlotName));
+                }
+
+                result.HostId = new Guid(hash).ToString().Replace("-", string.Empty).ToLowerInvariant();
+
+                return result;
+            }
+            else if (environment.IsAppService() || environment.IsAnyKubernetesEnvironment())
```
kshyju: nit: `else` can be removed.

src/WebJobs.Script/Host/ScriptHostIdProvider.cs
```diff
@@ -48,11 +50,23 @@ internal static HostIdResult GetDefaultHostId(IEnvironment environment, ScriptAp
...
-            // We're setting the default here on the newly created configuration
-            // If the user has explicitly set the HostID via host.json, it will overwrite
-            // what we set here
+            // Note: HostIds must conform to the rules documented here: https://learn.microsoft.com/en-us/azure/azure-functions/functions-app-settings#azurefunctionswebhost__hostid
             string hostId = null;
-            if (environment.IsAppService() || environment.IsAnyKubernetesEnvironment())
+            if (environment.IsConsumptionV2Sku())
+            {
+                // in CV2, we use a Guid based host ID without truncation.
+                string uniqueSlotName = environment?.GetAzureWebsiteUniqueSlotName();
```
kshyju, line 56: with the method name, I thought it was just slot name unique to the app. But when I looked at the implementation I learned it is the website name (which is unique across azure)+slot.

Question: Is it possible for 2 customers to create/use same site name in azurewebsites.net and may be azurewebsites.cn (I don't know that even exist. Remembering we(my prev team) had a msn.cn for china and msn.com for rest of the world)?  that situation will create the same value for uniqueSlotName and hence the guid generated from that. Is that a concern?

kshyju (approved):

src/WebJobs.Script/Host/ScriptHostIdProvider.cs
mathewc: No, site names must be globally unique.

src/WebJobs.Script/ScriptConstants.cs
pragnagopa: tagging @Tonewall as well .

pragnagopa: @mathewc  - @Tonewall - shared this offline -  https://github.com/Azure/azure-functions-host/blob/1f75688392491c60bae3411a2f50542dcbcf5b27/src/WebJobs.Script/Environment/EnvironmentExtensions.cs#L300 - please see if you can use the same helper

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
mathewc: Good catch - removed the dead code :)

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -201,14 +201,26 @@ public static bool IsWindowsConsumption(this IEnvironment environment)
...
+        /// Gets a value indicating whether the application is running in one of our consumption SKUs (e.g.  Windows or Linux Consumption (Dynamic),
+        /// or Consumption V2).
         /// App Service environment.
         /// </summary>
         /// <param name="environment">The environment to verify.</param>
         /// <returns><see cref="true"/> if running in a Windows or Linux Consumption App Service app; otherwise, false.</returns>
         public static bool IsConsumptionSku(this IEnvironment environment)
         {
-            return IsWindowsConsumption(environment) || IsAnyLinuxConsumption(environment);
+            return IsWindowsConsumption(environment) || IsAnyLinuxConsumption(environment) || IsConsumptionV2Sku(environment);
```
mathewc: Update in the latest revision - I realized the this new consumption v2 Sku needs to be accounted for in our existing IsConsumptionSku helper.

src/WebJobs.Script/ScriptConstants.cs
mathewc: Franklin confirmed to me offline that DynamicV2 is the SKU value that we'll see in WEBSITE_SKU and that's what I want to test for in my case.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -313,23 +313,26 @@ public async Task<SyncTriggersPayload> GetSyncTriggersPayload()
             var triggersArray = new JArray(triggers);
             int count = triggersArray.Count;
 
+            // Form the base minimal result
+            string hostId = await _hostIdProvider.GetHostIdAsync(CancellationToken.None);
```
mathewc, line 320: We're now ALWAYs sending the object format on syncs. Previously when cache was disabled or we were otherwise returning a "minmal" format, we were only sending an array. FE/ScaleController handles both formats of course so this isn't a problem.

test/WebJobs.Script.Tests/ScriptHostIdProviderTests.cs
```diff
@@ -94,6 +94,29 @@ public void GetDefaultHostId_AzureHost_ReturnsExpectedResult(string siteName, st
...
+        [InlineData("TEST-FUNCTIONS-XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXB", "b595edd5b794d34e79c81361c823487c")]
+        public void GetDefaultHostId_ConsumptionV2_ReturnsExpectedResult(string siteName, string expected)
```
nzthiago:
```suggestion
        public void GetDefaultHostId_FlexConsumption_ReturnsExpectedResult(string siteName, string expected)
```

test/WebJobs.Script.Tests/Extensions/EnvironmentExtensionsTests.cs
```diff
@@ -156,11 +156,12 @@ public void GetAntaresVersion_ReturnsExpectedResult(bool isLinuxConsumption, boo
...
-        [InlineData(false, true, true)]
-        [InlineData(false, false, false)]
-        [InlineData(true, true, true)]
-        public void IsConsumption_ReturnsExpectedResult(bool isLinuxConsumption, bool isWindowsConsumption, bool expectedValue)
+        [InlineData(true, false, false, true)]
+        [InlineData(false, false, true, true)]
+        [InlineData(false, true, false, true)]
+        [InlineData(false, false, false, false)]
+        [InlineData(true, true, false, true)]
+        public void IsConsumptionSku_ReturnsExpectedResult(bool isLinuxConsumption, bool isWindowsConsumption, bool isConsumptionV2, bool expectedValue)
```
nzthiago:
```suggestion
        public void IsConsumptionSku_ReturnsExpectedResult(bool isLinuxConsumption, bool isWindowsConsumption, bool isFlexConsumption, bool expectedValue)
```

test/WebJobs.Script.Tests/Extensions/EnvironmentExtensionsTests.cs
```diff
@@ -173,9 +174,26 @@ public void IsConsumption_ReturnsExpectedResult(bool isLinuxConsumption, bool is
                 env.SetEnvironmentVariable(EnvironmentSettingNames.AzureWebsiteSku, ScriptConstants.DynamicSku);
             }
 
+            if (isConsumptionV2)
```
nzthiago:
```suggestion
            if (isFlexConsumption)
```

nzthiago (changes requested):
Minor recommendations

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
chiangvincent, line 320: Is the hostId regenerated every time the `GetHostIdAsync` method is called? Double checking the following case: background sync triggers refresh called -> hostId regenerates -> sync triggers hash changes but triggers payload remains the same -> sync triggers payload written to hosting db-> site restarted unnecessarily.

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -201,14 +201,26 @@ public static bool IsWindowsConsumption(this IEnvironment environment)
...
-        /// <returns><see cref="true"/> if running in a Windows or Linux Consumption App Service app; otherwise, false.</returns>
+        /// <returns><see cref="true"/> if running in one of our Consumption SKUs, false otherwise.</returns>
         public static bool IsConsumptionSku(this IEnvironment environment)
         {
-            return IsWindowsConsumption(environment) || IsAnyLinuxConsumption(environment);
+            return IsWindowsConsumption(environment) || IsAnyLinuxConsumption(environment) || IsFlexConsumptionSku(environment);
+        }
+
+        /// <summary>
+        /// Gets a value indicating whether the application is running in the Flex Consumption Sku.
```
chiangvincent: Curious is Flex Consumption a new SKU?

src/WebJobs.Script/Host/ScriptHostIdProvider.cs
```diff
@@ -33,6 +34,7 @@ public Task<string> GetHostIdAsync(CancellationToken cancellationToken)
             string hostId = _config[ConfigurationSectionNames.HostIdPath];
             if (hostId == null)
             {
+                // if the user hasn't configured an explicit ID, we generate the default ID.
```
chiangvincent, line 37: Ah this answers the question I had earlier -- so if a user doesn't configure explicit ID, then we'd return a new hash each time? 

If the uniqueSlotName is doesn't change during the lifecycle of the ScriptHostIdProvider, and hashes to the same value each time, I'd consider storing the hash result in an attribute as an optimization

chiangvincent (changes requested):

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
mathewc: Yes, the ID is recomputed each time ScriptHostIdProvider.GetHostIdAsync is called (code [here](https://github.com/Azure/azure-functions-host/blob/016d6b03fc115da5e2ebea6caa7e4ad37140a7f7/src/WebJobs.Script/Host/ScriptHostIdProvider.cs#L31)). You're pointing out that when these changes are first released, the sync triggers payload will actually change because we're including the new HostID value. That's fine and expected - we want the DB to have the latest/updated payload with the ID. A site shouldn't restart when it's sync triggers payload changes.

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
mathewc: Yes - it's CV2

src/WebJobs.Script/Host/ScriptHostIdProvider.cs
mathewc: For a given site name, the host ID we return here will of course remain constant. Putting caching in place is beyond the scope of this PR, and we'd have to be careful. This ScriptHostIDProvider instance lives across host instances so we have to be careful with caching. From a layering point of view, it's always possible for upstream consumers of this API to cache the value if they want to.

chiangvincent (approved):

nzthiago (approved):

test/WebJobs.Script.Tests.Integration/Management/FunctionsSyncManagerTests.cs
```diff
@@ -139,6 +139,7 @@ public FunctionsSyncManagerTests()
             _mockEnvironment.Setup(p => p.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebsiteName)).Returns((string)null);
             _mockEnvironment.Setup(p => p.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebsiteSlotName)).Returns((string)null);
             _mockEnvironment.Setup(p => p.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebJobsFeatureFlags)).Returns((string)null);
+            _mockEnvironment.Setup(p => p.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebsiteSku)).Returns((string)ScriptConstants.DynamicSku);
```
nzthiago, line 142: Should there be a line like this but for ScriptConstants.FlexConsumptionSku as well?

mathewc: This is just the default for all the existing tests, when not explicitly overridden by individual tests. I had to add this because the sync codepaths now check sku where they didn't before, so a mock had to be set up.

I have added some Flex specific tests to verify the new logic - see below.

nzthiago: I see, thanks for clarifying!

PR #9158, opened 2023-03-16
Improve error handling for invocation request cancellations

### LINKED ISSUE #8986: Improve error handling for invocation request cancellations
As a follow up from issue #8978:

Seeing a "host unhandled error" when there has been a cancellation (after invocation but prior to sending the invocation request gRPC message to the worker) is confusing and doesn't help identify what may have happened.  Further more, the incoming invocation that we cancel fails with a TaskCanceledException.

Screenshot 2022-12-14 at 2 57 11 PM

#### Suggestion

a) We handle the `TaskCanceledException` that happens when we cancel the ResultSource:

```csharp
if (context.CancellationToken.IsCancellationRequested)
{
    _workerChannelLogger.LogDebug("Cancellation has been requested. The invocation request for {functionName} with invocation id '{invocationId}' is being cancelled.", context.FunctionMetadata.Name, invocationId);
    context.ResultSource.SetCanceled();
    return;
}
```

b) What might be more intuitive here is for us to report the invocation as cancelled instead of failed (as a warning level log) i.e.

`Executed 'Functions.MyHttpFunction' (Cancelled, Id=13456, Duration=50ms)`

With the details showing an appropriate error message such as "A cancellation was requested. Invocation has not been sent to the worker". 

#### Tasks

- [x] Reproduce getting a `TaskCanceledException` and investigate why this is not a `FunctionInvocationException`
- [x] Implement a better error handling approach so that we don't see a "host unhandled error" log message in this use case

#### How to repro

1.  Update `SendInvocationRequest` so that [this if-statement](https://github.com/Azure/azure-functions-host/blob/83081e19b68768f5cb8da360c3b4f6e879e49e67/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs#L650-L655) always hits i.e. `if (true) { context.ResultSource.TrySetCanceled }

liliankasem:

Error handling to catch the TaskCanceledException caused by cancelling the result source:

```csharp
if (context.CancellationToken.IsCancellationRequested)
{
    _workerChannelLogger.LogDebug("Cancellation has been requested. The invocation request for {functionName} with invocation id '{invocationId}' is being canceled.", context.FunctionMetadata.Name, invocationId);
    context.ResultSource.SetCanceled(); // leads to TaskCanceledException that is never caught/handled
    return;
}
```

resolves Azure/azure-functions-host#8986

Screenshot 2023-03-22 at 3 57 54 PM

### Pull request checklist

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -32,7 +32,7 @@ jobs:
 
 - job: BuildArtifacts
   dependsOn: InitializePipeline
-  condition: and(succeeded(), or(ne(variables['Build.Reason'], 'PullRequest'), eq(dependencies.InitializePipeline.outputs['Initialize.BuildArtifacts'], true)))
+  # condition: and(succeeded(), or(ne(variables['Build.Reason'], 'PullRequest'), eq(dependencies.InitializePipeline.outputs['Initialize.BuildArtifacts'], true)))
   variables:
     ${{ if or( eq( variables['Build.Reason'], 'PullRequest' ), and( not( contains( variables['Build.SourceBranch'], 'release/4.' ) ), not( contains( variables['Build.SourceBranch'], 'release/ExtensionsMetadataGenerator/' ) ) ) ) }}:
       suffixTemp: $(buildNumber)
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -32,7 +32,7 @@ jobs:
 
 - job: BuildArtifacts
   dependsOn: InitializePipeline
-  # condition: and(succeeded(), or(ne(variables['Build.Reason'], 'PullRequest'), eq(dependencies.InitializePipeline.outputs['Initialize.BuildArtifacts'], true)))
+  condition: and(succeeded(), or(ne(variables['Build.Reason'], 'PullRequest'), eq(dependencies.InitializePipeline.outputs['Initialize.BuildArtifacts'], true)))
   variables:
     ${{ if or( eq( variables['Build.Reason'], 'PullRequest' ), and( not( contains( variables['Build.SourceBranch'], 'release/4.' ) ), not( contains( variables['Build.SourceBranch'], 'release/ExtensionsMetadataGenerator/' ) ) ) ) }}:
       suffixTemp: $(buildNumber)
--- a/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
@@ -716,28 +716,29 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
                 }
 
                 // do not send an invocation request if cancellation has been requested
-                if (context.CancellationToken.IsCancellationRequested)
+                if (true)
                 {
+                    await Task.Delay(100);
                     _workerChannelLogger.LogWarning("Cancellation has been requested. The invocation request with id '{invocationId}' is canceled and will not be sent to the worker.", invocationId);
                     context.ResultSource.TrySetCanceled();
                     return;
                 }
 
-                var invocationRequest = await context.ToRpcInvocationRequest(_workerChannelLogger, _workerCapabilities, _isSharedMemoryDataTransferEnabled, _sharedMemoryManager);
-                AddAdditionalTraceContext(invocationRequest.TraceContext.Attributes, context);
-                _executingInvocations.TryAdd(invocationRequest.InvocationId, context);
+                // var invocationRequest = await context.ToRpcInvocationRequest(_workerChannelLogger, _workerCapabilities, _isSharedMemoryDataTransferEnabled, _sharedMemoryManager);
+                // AddAdditionalTraceContext(invocationRequest.TraceContext.Attributes, context);
+                // _executingInvocations.TryAdd(invocationRequest.InvocationId, context);
 
-                _metricsLogger.LogEvent(string.Format(MetricEventNames.WorkerInvoked, Id), functionName: context.FunctionMetadata.Name);
+                // _metricsLogger.LogEvent(string.Format(MetricEventNames.WorkerInvoked, Id), functionName: context.FunctionMetadata.Name);
 
-                await SendStreamingMessageAsync(new StreamingMessage
-                {
-                    InvocationRequest = invocationRequest
-                });
+                // await SendStreamingMessageAsync(new StreamingMessage
+                // {
+                //     InvocationRequest = invocationRequest
+                // });
 
-                if (_cancelCapabilityEnabled != null && _cancelCapabilityEnabled.Value)
-                {
-                    context.CancellationToken.Register(() => SendInvocationCancel(invocationRequest.InvocationId));
-                }
+                // if (_cancelCapabilityEnabled != null && _cancelCapabilityEnabled.Value)
+                // {
+                //     context.CancellationToken.Register(() => SendInvocationCancel(invocationRequest.InvocationId));
+                // }
             }
             catch (Exception invokeEx)
             {
--- a/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
@@ -718,7 +718,7 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
                 // do not send an invocation request if cancellation has been requested
                 if (context.CancellationToken.IsCancellationRequested)
                 {
-                    _workerChannelLogger.LogWarning("Cancellation has been requested. The invocation request with id '{invocationId}' is cancelled and will not be sent to the worker.", invocationId);
+                    _workerChannelLogger.LogWarning("Cancellation has been requested. The invocation request with id '{invocationId}' is canceled and will not be sent to the worker.", invocationId);
                     context.ResultSource.TrySetCanceled();
                     return;
                 }
--- a/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
@@ -716,29 +716,28 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
                 }
 
                 // do not send an invocation request if cancellation has been requested
-                if (true)
+                if (context.CancellationToken.IsCancellationRequested)
                 {
-                    await Task.Delay(100);
                     _workerChannelLogger.LogWarning("Cancellation has been requested. The invocation request with id '{invocationId}' is canceled and will not be sent to the worker.", invocationId);
                     context.ResultSource.TrySetCanceled();
                     return;
                 }
 
-                // var invocationRequest = await context.ToRpcInvocationRequest(_workerChannelLogger, _workerCapabilities, _isSharedMemoryDataTransferEnabled, _sharedMemoryManager);
-                // AddAdditionalTraceContext(invocationRequest.TraceContext.Attributes, context);
-                // _executingInvocations.TryAdd(invocationRequest.InvocationId, context);
+                var invocationRequest = await context.ToRpcInvocationRequest(_workerChannelLogger, _workerCapabilities, _isSharedMemoryDataTransferEnabled, _sharedMemoryManager);
+                AddAdditionalTraceContext(invocationRequest.TraceContext.Attributes, context);
+                _executingInvocations.TryAdd(invocationRequest.InvocationId, context);
 
-                // _metricsLogger.LogEvent(string.Format(MetricEventNames.WorkerInvoked, Id), functionName: context.FunctionMetadata.Name);
+                _metricsLogger.LogEvent(string.Format(MetricEventNames.WorkerInvoked, Id), functionName: context.FunctionMetadata.Name);
 
-                // await SendStreamingMessageAsync(new StreamingMessage
-                // {
-                //     InvocationRequest = invocationRequest
-                // });
+                await SendStreamingMessageAsync(new StreamingMessage
+                {
+                    InvocationRequest = invocationRequest
+                });
 
-                // if (_cancelCapabilityEnabled != null && _cancelCapabilityEnabled.Value)
-                // {
-                //     context.CancellationToken.Register(() => SendInvocationCancel(invocationRequest.InvocationId));
-                // }
+                if (_cancelCapabilityEnabled != null && _cancelCapabilityEnabled.Value)
+                {
+                    context.CancellationToken.Register(() => SendInvocationCancel(invocationRequest.InvocationId));
+                }
             }
             catch (Exception invokeEx)
             {
--- a/src/WebJobs.Script.WebHost/Middleware/ExceptionMiddleware.Log.cs
+++ b/src/WebJobs.Script.WebHost/Middleware/ExceptionMiddleware.Log.cs
@@ -11,19 +11,14 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Middleware
         private static class Logger
         {
             private static readonly Action<ILogger, Exception> _responseStarted =
-                LoggerMessage.Define(LogLevel.Debug, new EventId(1, nameof(ResponseStarted)), "The response has already started, the status code will not be modified.");
+               LoggerMessage.Define(LogLevel.Debug, new EventId(1, nameof(ResponseStarted)), "The response has already started, the status code will not be modified.");
 
             private static readonly Action<ILogger, Exception> _unhandledHostError =
                 LoggerMessage.Define(LogLevel.Error, new EventId(2, nameof(UnhandledHostError)), "An unhandled host error has occurred.");
 
-            private static readonly Action<ILogger, Exception> _invocationCanceled =
-                LoggerMessage.Define(LogLevel.Warning, new EventId(3, nameof(InvocationCanceled)), "An invocation request was canceled before it was sent to the worker.");
-
             public static void ResponseStarted(ILogger logger) => _responseStarted(logger, null);
 
             public static void UnhandledHostError(ILogger logger, Exception ex) => _unhandledHostError(logger, ex);
-
-            public static void InvocationCanceled(ILogger logger, Exception ex) => _invocationCanceled(logger, ex);
         }
     }
 }
--- a/src/WebJobs.Script.WebHost/Middleware/ExceptionMiddleware.Log.cs
+++ b/src/WebJobs.Script.WebHost/Middleware/ExceptionMiddleware.Log.cs
@@ -16,14 +16,14 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Middleware
             private static readonly Action<ILogger, Exception> _unhandledHostError =
                 LoggerMessage.Define(LogLevel.Error, new EventId(2, nameof(UnhandledHostError)), "An unhandled host error has occurred.");
 
-            private static readonly Action<ILogger, Exception> _invocationCancelled =
-                LoggerMessage.Define(LogLevel.Warning, new EventId(3, nameof(InvocationCancelled)), "An invocation request was cancelled before it was sent to the worker.");
+            private static readonly Action<ILogger, Exception> _invocationCanceled =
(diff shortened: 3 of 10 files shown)


src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -718,7 +718,7 @@ internal async Task SendInvocationRequest(ScriptInvocationContext context)
                 // do not send an invocation request if cancellation has been requested
                 if (context.CancellationToken.IsCancellationRequested)
                 {
-                    _workerChannelLogger.LogWarning("Cancellation has been requested. The invocation request with id '{invocationId}' is cancelled and will not be sent to the worker.", invocationId);
+                    _workerChannelLogger.LogWarning("Cancellation has been requested. The invocation request with id '{invocationId}' is canceled and will not be sent to the worker.", invocationId);
```
satvu, line 721: :eagle:

satvu (approved):

src/WebJobs.Script.WebHost/Middleware/ExceptionMiddleware.cs
```diff
@@ -28,11 +28,18 @@ public async Task Invoke(HttpContext context)
...
+                // exceptions throw by function code are handled/logged elsewhere
+                // our goal here is to log exceptions coming from our own runtime
                 if (!(ex is FunctionInvocationException))
                 {
-                    // exceptions throw by function code are handled/logged elsewhere
-                    // our goal here is to log exceptions coming from our own runtime
-                    Logger.UnhandledHostError(_logger, ex);
+                    if (ex is TaskCanceledException)
+                    {
+                        Logger.InvocationCancelled(_logger, ex);
```
satvu: 🦅  need to americanize this one too

src/WebJobs.Script.WebHost/Middleware/ExceptionMiddleware.Log.cs
```diff
@@ -11,14 +11,19 @@ internal partial class ExceptionMiddleware
         private static class Logger
         {
             private static readonly Action<ILogger, Exception> _responseStarted =
-               LoggerMessage.Define(LogLevel.Debug, new EventId(1, nameof(ResponseStarted)), "The response has already started, the status code will not be modified.");
+                LoggerMessage.Define(LogLevel.Debug, new EventId(1, nameof(ResponseStarted)), "The response has already started, the status code will not be modified.");
```
madelinegordon:
```suggestion
                LoggerMessage.Define(LogLevel.Debug, new EventId(1, nameof(ResponseStarted)), "The response has already started. The status code will not be modified.");
```

src/WebJobs.Script.WebHost/Middleware/ExceptionMiddleware.cs
```diff
@@ -28,11 +28,18 @@ public async Task Invoke(HttpContext context)
             }
             catch (Exception ex)
             {
+                // exceptions throw by function code are handled/logged elsewhere
```
madelinegordon:
```suggestion
                // Exceptions thrown by function code are handled/logged elsewhere.
```

madelinegordon (approved):

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
liliankasem: 🇬🇧☠️😢

src/WebJobs.Script.WebHost/Middleware/ExceptionMiddleware.Log.cs
liliankasem: I don't want to edit the log message since I didn't make any changes here

src/WebJobs.Script.WebHost/Middleware/ExceptionMiddleware.Log.cs
```diff
@@ -11,14 +11,19 @@ internal partial class ExceptionMiddleware
...
 
             private static readonly Action<ILogger, Exception> _unhandledHostError =
                 LoggerMessage.Define(LogLevel.Error, new EventId(2, nameof(UnhandledHostError)), "An unhandled host error has occurred.");
 
+            private static readonly Action<ILogger, Exception> _invocationCanceled =
+                LoggerMessage.Define(LogLevel.Warning, new EventId(3, nameof(InvocationCanceled)), "An invocation request was canceled before it was sent to the worker.");
```
liliankasem: @fabiocav we already log something about this in GrpcWorkerChannel as a warning, do we need to log this again? And should we log one of them as info or debug?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
fabiocav: We do standardize on `en-US` in the codebase, so we should have this reverted :)

We also want to avoid log verbiage changes (unless absolutely necessary) as those could be breaking.

src/WebJobs.Script.WebHost/Middleware/ExceptionMiddleware.Log.cs
fabiocav: If we already have a log that covers this, as long as it is sufficient, no need for this additional one.

src/WebJobs.Script.WebHost/Middleware/ExceptionMiddleware.cs
```diff
@@ -28,11 +28,18 @@ public async Task Invoke(HttpContext context)
...
             catch (Exception ex)
             {
+                // Exceptions thrown by function code are handled/logged elsewhere.
+                // our goal here is to log exceptions coming from our own runtime
                 if (!(ex is FunctionInvocationException))
                 {
-                    // exceptions throw by function code are handled/logged elsewhere
-                    // our goal here is to log exceptions coming from our own runtime
-                    Logger.UnhandledHostError(_logger, ex);
+                    if (ex is TaskCanceledException)
```
fabiocav: We we certain here that every instance of `TaskCancelledException` would be an invocation cancellation? There are other flows in the Invoke path that could lead to this being thrown, right? Are we trying to differentiate that in an y way, or is it accurate to interpret as an invocation cancellation?

We know when an invocation is cancelled. Would it be more accurate to have a specific exception that communicates that so there's no ambiguity? That could actually be a `FunctionInvocationException` derived type. Thoughts?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
liliankasem: `canceled` is US and `cancelled` is UK, so I was updating to follow the en-US standard, these are all logs I've added and I don't think they're used anywhere so shouldn't be breaking. Happy to revert if we're concerned though

src/WebJobs.Script.WebHost/Middleware/ExceptionMiddleware.cs
liliankasem: Yeah I was thinking the same thing, I'm not 100% sure that we can be sure that is is specifically from an invocation cancellation in this code path. 

I like the idea of introducing an exception for this, but we would have to catch that [here](https://github.com/Azure/azure-functions-host/blob/9fc7fa583bafe4f2f6e591a1ad401235884a03b8/src/WebJobs.Script/Description/Workers/WorkerFunctionInvoker.cs#L96) and we should be able to check the ResultSource state. Or do you think it should be somewhere else?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
fabiocav: I completely flipped your update and somehow saw this going the other way around 🤦🏼 

The concern about the verbiage change is still valid, but I agree with you that those are unlikely to cause issues. 

I'm good with the change.

azure-pipelines.yml
```diff
@@ -32,7 +32,7 @@ jobs:
 
 - job: BuildArtifacts
   dependsOn: InitializePipeline
-  condition: and(succeeded(), or(ne(variables['Build.Reason'], 'PullRequest'), eq(dependencies.InitializePipeline.outputs['Initialize.BuildArtifacts'], true)))
+  # condition: and(succeeded(), or(ne(variables['Build.Reason'], 'PullRequest'), eq(dependencies.InitializePipeline.outputs['Initialize.BuildArtifacts'], true)))
```
liliankasem: temp change for private site testing

src/WebJobs.Script/Description/Workers/WorkerFunctionInvoker.cs
```diff
@@ -91,11 +91,23 @@ protected override async Task<object> InvokeCore(object[] parameters, FunctionIn
...
+            try
+            {
+                _logger.LogTrace("Sending invocation id: '{id}", invocationId);
+
+                await _functionDispatcher.InvokeAsync(invocationContext);
+                result = await invocationContext.ResultSource.Task;
+
+                await BindOutputsAsync(triggerValue, context.Binder, result);
+            }
+            catch (Exception ex) when (invocationContext.ResultSource.Task.IsCanceled)
```
liliankasem, line 105: Do we want to check for `invocationContext.CancellationToken.IsCancellationRequested` instead (or as well)?

I think at this stage (in this code path), if the cancellation token is cancelled and there is an exception, it is during a function invocation path

src/WebJobs.Script/Description/Workers/FunctionInvocationCanceledException.cs
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Azure.WebJobs.Host;
+
+namespace Microsoft.Azure.WebJobs.Script.Description
+{
+    public class FunctionInvocationCanceledException : FunctionInvocationException
```
jviau: Have we considered using or deriving from `OperationCanceledException`? I recommend we consider it because it is an established pattern to use that exception type (or something derived from it) to represent cancellation. In fact, many libraries and tools - even the BCL - has special case handling in places for this exception type. For example, a `Task` representing work which throws this exception becomes "canceled" and not "failed".

src/WebJobs.Script/Description/Workers/WorkerFunctionInvoker.cs
jviau: Any reason we don't use `catch (OperationCanceledException ex)`? Can you comment why we do it that way?

If we are looking for handling only if a _specific_ `await` statement above is canceled, should we wrap it with its own `try/catch (OperationCanceledException ex)` statement? It might be a bit clearer to the reader what our intentions are then.

liliankasem: I want to make sure we're only catching this when we cancel the ResultSource. This only happens at one point in our code base (GrpcWorkerChancel -> InvocationRequest). I'm trying to avoid messing with any other flow we have here.

Yup it is only for a specific await statement so I can update the code for the try/catch. I think the exception thrown in TaskCanceledException and I don't know if something else might cause it or not so I feel like having the resultsource check doesn't hurt?

src/WebJobs.Script/Description/Workers/FunctionInvocationCanceledException.cs
liliankasem: I think we want to use `FunctionInvocationException` because a) we want to link this to a function invocation and make that clear to users and b) how the exception middleware handles exceptions that are / are not `FunctionInvocationException` (though we can always change this if there's a strong enough argument)

- https://github.com/Azure/azure-functions-host/blob/cacd18bd7c208cf0cad4912b920186c3397ea527/src/WebJobs.Script.WebHost/Middleware/ExceptionMiddleware.cs#L31

fabiocav: For this flow, using something derived from `FunctionInvocationException` makes sense. We want this to be handled by the WebJobs/Host and using `OperationCancelledException` , or deriving from it, could cause other code paths to prevent that from flowing back up.

src/WebJobs.Script/Description/Workers/WorkerFunctionInvoker.cs
```diff
@@ -91,11 +91,23 @@ protected override async Task<object> InvokeCore(object[] parameters, FunctionIn
...
-            _logger.LogTrace($"Sending invocation id:{invocationId}");
+            ScriptInvocationResult result = new();
+
+            _logger.LogTrace("Sending invocation id: '{id}", invocationId);
             await _functionDispatcher.InvokeAsync(invocationContext);
-            var result = await invocationContext.ResultSource.Task;
+
+            try
+            {
+                result = await invocationContext.ResultSource.Task;
```
kshyju, line 101: Is accessing the ResultSource.Task enough? Is it possible for line 97 to throw a TaskCancelledException/OperationCancelled exception ?

src/WebJobs.Script/Description/Workers/WorkerFunctionInvoker.cs
```diff
@@ -91,11 +91,23 @@ protected override async Task<object> InvokeCore(object[] parameters, FunctionIn
...
+            }
+            catch (TaskCanceledException ex)
```
kshyju: Should we use the parent type `OperationCanceledException` ? I wonder some APIs we use down the line throws that exception instead of TaskCancelled.

src/WebJobs.Script/Description/Workers/WorkerFunctionInvoker.cs
liliankasem: I don't think so, the invocation task is the `ResultSource.Task` and that's what we're trying to handle, I can go back to including line 97 if we need to

src/WebJobs.Script/Description/Workers/WorkerFunctionInvoker.cs
liliankasem: I think we should be clear for the specific scenarios we're trying to handle here, but I'm fine to use `OperationCanceledException` instead. My personal preference is to actually check `invocationContext.ResultSource.Task.IsCanceled`

src/WebJobs.Script/Description/Workers/FunctionInvocationCanceledException.cs
fabiocav: I'm thinking about where we want to have this defined, in case we want this to flow all the way back up to WebJobs.

fabiocav (commented):
The changes look good, but wanted to discuss whether we should consider defining the new exception in WebJobs.

liliankasem:
Going to merge for now as folks are waiting on this change, but keeping a note to have the exception moved to WebJobs in case it helps with other scenarios.

Edit: @fabiocav @kshyju @jviau  - need one more approval to merge :) I'll be OOF so feel free to merge on my behalf if there are no blocking comments

fabiocav (approved):

RichardBurns1982:
Will this merge fix the issue we have raised here: https://github.com/Azure/azure-functions-durable-extension/issues/2454

Also, when will this fix be published and in production on the functions?

liliankasem:
> Will this merge fix the issue we have raised here: Azure/azure-functions-durable-extension#2454
> 
> Also, when will this fix be published and in production on the functions?

Not necessarily; this fix is to avoid us going down the code path that leads to a "An unhandled host error has occurred." error (which results in a host shut down). We're handling that by introducing a new exception type `FunctionInvocationCanceledException` - this is to classify what is happening when the platform cancellation token is cancelled and we have to stop any **new** invocations before we send them to the worker.

Unfortunately, when an invocation request occurs and the cancellation token has been signaled to cancel, it's not possible to just get rid of that invocation - that invocation **request** has still come through and we have to do something with it. In the cancellation scenario, we don't process new invocations and we "cancel" them before we send the work to the worker. Our only options here are to fail or succeeded the invocation as there is no "cancelled" state. We choose "fail" because we never send the invocation to the worker and a "successful" invocation at this point would be misleading.

With this change, the invocation will still fail (expected) but you'll still see a clearer error message like this instead:

`FunctionInvocationCanceledException: The invocation request with id '{invocationId}' was canceled before the request was sent to the worker.`

cc: @fabiocav

RichardBurns1982:
> > Will this merge fix the issue we have raised here: Azure/azure-functions-durable-extension#2454
> > Also, when will this fix be published and in production on the functions?
> 
> Not necessarily; this fix is to avoid us going down the code path that leads to a "An unhandled host error has occurred." error (which results in a host shut down). We're handling that by introducing a new exception type `FunctionInvocationCanceledException` - this is to classify what is happening when the platform cancellation token is cancelled and we have to stop any **new** invocations before we send them to the worker.
> 
> Unfortunately, when an invocation request occurs and the cancellation token has been signaled to cancel, it's not possible to just get rid of that invocation - that invocation **request** has still come through and we have to do something with it. In the cancellation scenario, we don't process new invocations and we "cancel" them before we send the work to the worker. Our only options here are to fail or succeeded the invocation as there is no "cancelled" state. We choose "fail" because we never send the invocation to the worker and a "successful" invocation at this point would be misleading.
> 
> With this change, the invocation will still fail (expected) but you'll still see a clearer error message like this instead:
> 
> `FunctionInvocationCanceledException: The invocation request with id '{invocationId}' was canceled before the request was sent to the worker.`
> 
> cc: @fabiocav

Thanks for the response, I appreciate it.

Do you know why these are getting cancelled?  This problem has plagued us since we moved to net7 isolated functions from net6 in-process.  There is nothing in our code that is cancelling anything and we can find no evidence of any issues within our code that would be causing it.  We are intermittently losing orchestration functions mid flow and retry logic not working.

liliankasem:
> Thanks for the response, I appreciate it.
> 
> Do you know why these are getting cancelled? This problem has plagued us since we moved to net7 isolated functions from net6 in-process. There is nothing in our code that is cancelling anything and we can find no evidence of any issues within our code that would be causing it. We are intermittently losing orchestration functions mid flow and retry logic not working.

Typically, platform cancellations happen when we need to restart or shutdown the host, or a function has reached the timeout limit. E.g. scale down event occurs, a worker gets recycled, we request to shutdown the host, the platform cancellation token gets cancelled. When this happens, if you have inflight invocations, you should be able to handle the cancellation token. If you get new invocations during this period, they're not handled. But, you mentioned an orchestrator function, I don't know that we have cancellation support with durable functions so that might be different case entirely.  @jviau @davidmrdavid FYI in case you have more info around that.

davidmrdavid:
I'll defer the specifics of cancellation-token support in DF isolated to @jviau. 
@RichardBurns1982: can you please clarify what you mean by "We are intermittently losing orchestration functions mid flow"? Does that mean that the orchestration is erroring out in some way? Does it recover or is the error permanent?

jviau:
DF isolated does not integrate with the invocation request cancellation of Azure Functions. The DF code does not listen to that cancellation token nor anything requests cancellation of said token (from the DF perspective). I don't see how this would be related to an orchestration function terminating mid-flow.

RichardBurns1982:
> I'll defer the specifics of cancellation-token support in DF isolated to @jviau. @RichardBurns1982: can you please clarify what you mean by "We are intermittently losing orchestration functions mid flow"? Does that mean that the orchestration is erroring out in some way? Does it recover or is the error permanent?

I've raised the issue here: https://github.com/Azure/azure-functions-durable-extension/issues/2454, this has only started happening since moving to net7 isolated.  It sometimes happens on the orchestration, sometimes on the activity when it tries to run but it fails the orchestration function and doesn't even hit our new exception handling middleware, either way there is no way for us to recover, particularly if it happens on the orchestration function.  We can try to put retries and more resilience around the activity within the orchestration function but that won't help us when it is the Orchestration function that raises a TaskCancelledException from within the host code.

RichardBurns1982:
> DF isolated does not integrate with the invocation request cancellation of Azure Functions. The DF code does not listen to that cancellation token nor anything requests cancellation of said token (from the DF perspective). I don't see how this would be related to an orchestration function terminating mid-flow.

Likely nothing to do with Orchestration functions, we run approx. 50k function instances on any given environment a day, approximately 90% of which are orchestration functions/activities so that just happens to be where we are seeing it.

Happy to provide any invocation details so they can be investigated.

PR #9159, opened 2023-03-17
Support for HttpContext.RequestAborted cancellation token for Http requests

### LINKED ISSUE #8903: Support for HttpContext.RequestAborted in the HttpTrigger cancellation token
Using Http triggers in Azure Functions is a common use case. Having longer running functions is also quite normal. Therefore I'm actually quite surprised that cancelling a request (fx. in Postman) affect the function running. IMHO that is what developers would expect (at least it's what I did, and what the author did [here](https://stackoverflow.com/questions/60202259/azure-functions-using-cancellation-token-with-http-trigger/))

Something like what's described [here](https://stackoverflow.com/questions/60202259/azure-functions-using-cancellation-token-with-http-trigger/63439515#63439515) - when the client cancels the http request the Azure Function should be notified through the cancellation token by default

Manually handling the linking of the host cancellation token and HttpContext.RequestAborted as described [here](https://stackoverflow.com/questions/60202259/azure-functions-using-cancellation-token-with-http-trigger/63439515#63439515)

#### Additional context
None

liliankasem:

resolves #8903

**Use case**: InvocationCancel should be sent to the worker whether the host cancellation token is cancelled or if if the `RequestAborted` cancellation token is cancelled (in the case where we have a http trigger).

**Solution**: Update the cancellation token being used to be the RequestAborted on. This cancellation token is linked to the lifetime source so the cancellation token should trigger for both request aborted events, and lifetime events.

Verified that the `RequestAborted` CT works with application lifetime cancellations.

### Pull request checklist

* [ ] I have added all required tests (Unit tests, E2E tests) // TODO

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/Middleware/FunctionInvocationMiddleware.cs
+++ b/src/WebJobs.Script.WebHost/Middleware/FunctionInvocationMiddleware.cs
@@ -128,7 +128,8 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Middleware
                 using (logger.BeginScope(scopeState))
                 {
                     var applicationLifetime = context.RequestServices.GetService<IApplicationLifetime>();
-                    CancellationToken cancellationToken = applicationLifetime != null ? applicationLifetime.ApplicationStopping : CancellationToken.None;
+                    CancellationToken cancellationToken = context.RequestAborted;
+
                     await functionExecution.ExecuteAsync(context.Request, cancellationToken);
 
                     if (context.Items.TryGetValue(ScriptConstants.AzureFunctionsDuplicateHttpHeadersKey, out object value))
--- a/release_notes.md
+++ b/release_notes.md
@@ -16,6 +16,7 @@
 - Update PowerShell Worker 7.2 to 4.0.2719 [Release Note](https://github.com/Azure/azure-functions-powershell-worker/releases/tag/v4.0.2719)
 - Update Node.js Worker Version to [3.6.0](https://github.com/Azure/azure-functions-nodejs-worker/releases/tag/v3.6.0)
 - Upgrade PowerShell language worker 7.0 to 4.0.2733 [Release Note](https://github.com/Azure/azure-functions-powershell-worker/releases/tag/v4.0.2733)
+- Support for HttpContext.RequestAborted in the HttpTrigger cancellation token (#9159)
 
 **Release sprint:** Sprint 139
 [ [bugs](https://github.com/Azure/azure-functions-host/issues?q=is%3Aissue+milestone%3A%22Functions+Sprint+139%22+label%3Abug+is%3Aclosed) | [features](https://github.com/Azure/azure-functions-host/issues?q=is%3Aissue+milestone%3A%22Functions+Sprint+139%22+label%3Afeature+is%3Aclosed) ]
--- a/release_notes.md
+++ b/release_notes.md
@@ -16,7 +16,7 @@
 - Update PowerShell Worker 7.2 to 4.0.2719 [Release Note](https://github.com/Azure/azure-functions-powershell-worker/releases/tag/v4.0.2719)
 - Update Node.js Worker Version to [3.6.0](https://github.com/Azure/azure-functions-nodejs-worker/releases/tag/v3.6.0)
 - Upgrade PowerShell language worker 7.0 to 4.0.2733 [Release Note](https://github.com/Azure/azure-functions-powershell-worker/releases/tag/v4.0.2733)
-- Support for HttpContext.RequestAborted in the HttpTrigger cancellation token (#9159)
+- Support for HttpContext.RequestAborted cancellation token for Http requests (#9159)
 
 **Release sprint:** Sprint 139
 [ [bugs](https://github.com/Azure/azure-functions-host/issues?q=is%3Aissue+milestone%3A%22Functions+Sprint+139%22+label%3Abug+is%3Aclosed) | [features](https://github.com/Azure/azure-functions-host/issues?q=is%3Aissue+milestone%3A%22Functions+Sprint+139%22+label%3Afeature+is%3Aclosed) ]


src/WebJobs.Script/Description/Workers/WorkerFunctionInvoker.cs
```diff
@@ -187,14 +194,23 @@ private object TransformInput(object input, Dictionary<string, object> bindingDa
...
                 return CancellationToken.None;
             }
 
-            return (CancellationToken)input;
+            CancellationToken cancellationToken = (CancellationToken)input;
+
+            if (httpRequest is not null)
+            {
+                // TODO: dispose cancellationSource
+                var cancellationSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, httpRequest.HttpContext.RequestAborted);
```
liliankasem, line 209: Is this the best approach for allowing cancellation for both the host cancellation token and the http one? 

Suggestions for disposal?

jviau:
I am wondering if there is a way we can avoid introducing HTTP specific logic for cancellation here. Hopefully there is also an existing way for HTTP extension to link the cancellation token on its own? What I am thinking is that there must be _some_ piece of code which is bridging the incoming HTTP request in the host to a function invocation. At the point in time it shims this HTTP call to a function invocation, why is it not supplying the `RequestAborted` cancellation token as a cancellation token for the function invocation?

If we figure this out then we can also ensure any extension can do similarly. If an extension has triggers and dispatches function calls, make sure they can provide a cancellation token as part of that dispatch and trust it will eventually get passed on to the function implementation. This would also be important for Durable Functions as we _eventually_ want to support orchestration recursive cancellation.

I _think_ the relevant piece of code for HTTP function dispatching is [ScriptRouteHandler](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.WebHost/Routing/ScriptRouteHandler.cs). But to be honest, I couldn't really follow what is going on there - it looks to be the hook point for handling function routes from the WebJobs HTTP extension?

liliankasem:
> I am wondering if there is a way we can avoid introducing HTTP specific logic for cancellation here. Hopefully there is also an existing way for HTTP extension to link the cancellation token on its own? What I am thinking is that there must be _some_ piece of code which is bridging the incoming HTTP request in the host to a function invocation. At the point in time it shims this HTTP call to a function invocation, why is it not supplying the `RequestAborted` cancellation token as a cancellation token for the function invocation?
> 
> If we figure this out then we can also ensure any extension can do similarly. If an extension has triggers and dispatches function calls, make sure they can provide a cancellation token as part of that dispatch and trust it will eventually get passed on to the function implementation. This would also be important for Durable Functions as we _eventually_ want to support orchestration recursive cancellation.
> 
> I _think_ the relevant piece of code for HTTP function dispatching is [ScriptRouteHandler](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.WebHost/Routing/ScriptRouteHandler.cs). But to be honest, I couldn't really follow what is going on there - it looks to be the hook point for handling function routes from the WebJobs HTTP extension?

That's a great point, let me look into if we can do something like this!

Edit: @jviau - some findings to share:

The current cancellation token that we bind to is the one we get from the "_cancellationToken" parameter  which we get via `CancellationTokenValueProvider` ([link](https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Bindings/Cancellation/CancellationTokenValueProvider.cs))

The HttpRequest comes from the "_req" parameter which we get from `HttpRequestValueBinder` ([link](https://github.com/Azure/azure-webjobs-sdk-extensions/blob/2de8907944d2820c267b25ae1e49221392eb3a5b/src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs#L489)). This request is the one we can get the RequestAborted ct from. 

This is where we evaluate the parameters using value providers [link](https://github.com/Azure/azure-webjobs-sdk/blob/c1e3886e6b4664871a0a91c83b03d0ed7689cb6b/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs#L905). At this point we have access to both `_req` and the host `_cancellationToken` in the WebJobs SDK.

The entry point for when we trigger a HTTP function is [here](https://github.com/Azure/azure-webjobs-sdk-extensions/blob/2de8907944d2820c267b25ae1e49221392eb3a5b/src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs#L133). This is also where we create the `HttpRequestValueBinder` that gets used when the host tries to get the `_req` parameter:

`valueProvider = new HttpRequestValueBinder(_parameter, request, invokeString);`

I think we need to find a way to link the `RequestAborted` cancellation token with the one provided by `CancellationTokenValueProvider`? Might need to do more digging here but not certain on how to tie these two together. Important note: we still want the host cancellation token (this is used for timeout and host shutdown scenarios).

src/WebJobs.Script.WebHost/Middleware/FunctionInvocationMiddleware.cs
```diff
@@ -128,7 +128,8 @@ private async Task<IActionResult> GetResultAsync(HttpContext context, IFunctionE
                 using (logger.BeginScope(scopeState))
                 {
                     var applicationLifetime = context.RequestServices.GetService<IApplicationLifetime>();
-                    CancellationToken cancellationToken = applicationLifetime != null ? applicationLifetime.ApplicationStopping : CancellationToken.None;
+                    CancellationToken cancellationToken = context.HttpContext.RequestAborted ?? CancellationToken.None;
```
jviau: Is `RequestAborted` nullable? If not, you don't need to null-coalesce, as it will default to `CancellationToken.None` as it is a struct.

jviau (approved):

src/WebJobs.Script.WebHost/Middleware/FunctionInvocationMiddleware.cs
liliankasem: Good call out; will double check and update as necessary

satvu (approved):

PR #9168, opened 2023-03-21
[V4] Allow site auth JWT tokens to be passed via x-ms-site-token header.

mathewc:
In addition to accepting admin JWT tokens as we do now via the standard Authorization header, we want to allow them to be passed via our own custom x-ms-site-token header. This avoids conflicts in scenarios where an application may be using the Authorization header themselves. Addresses https://github.com/Azure/azure-functions-host/issues/6886.

### Pull request checklist

    * [x] v3 https://github.com/Azure/azure-functions-host/pull/9175
    * [x] v1 https://github.com/Azure/azure-functions-host/pull/9176

### DIFF
--- a/src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
+++ b/src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
@@ -1,19 +1,18 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
-using System;
+using System.Linq;
 using System.Security.Claims;
-using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Authentication;
 using Microsoft.AspNetCore.Authentication.JwtBearer;
-using Microsoft.Azure.Web.DataProtection;
 using Microsoft.Azure.WebJobs.Extensions.Http;
 using Microsoft.Azure.WebJobs.Script;
 using Microsoft.Azure.WebJobs.Script.Config;
 using Microsoft.Azure.WebJobs.Script.WebHost;
 using Microsoft.Azure.WebJobs.Script.WebHost.Security.Authentication;
+using Microsoft.Extensions.Primitives;
 using Microsoft.IdentityModel.Tokens;
 using static Microsoft.Azure.WebJobs.Script.EnvironmentSettingNames;
 using static Microsoft.Azure.WebJobs.Script.ScriptConstants;
@@ -31,8 +30,16 @@ namespace Microsoft.Extensions.DependencyInjection
                             {
                                 OnMessageReceived = c =>
                                 {
+                                    // By default, tokens are passed via the standard Authorization Bearer header. However we also support
+                                    // passing tokens via the x-ms-site-token header.
+                                    if (c.Request.Headers.TryGetValue(ScriptConstants.SiteTokenHeaderName, out StringValues values))
+                                    {
+                                        // the token we set here will be the one used - Authorization header won't be checked.
+                                        c.Token = values.FirstOrDefault();
+                                    }
+
                                     // Temporary: Tactical fix to address specialization issues. This should likely be moved to a token validator
-                                    // TODO: DI (FACAVAL) This will be fixed once the permanent fix is in plance
+                                    // TODO: DI (FACAVAL) This will be fixed once the permanent fix is in place
                                     if (_specialized == 0 && !SystemEnvironment.Instance.IsPlaceholderModeEnabled() && Interlocked.CompareExchange(ref _specialized, 1, 0) == 0)
                                     {
                                         o.TokenValidationParameters = CreateTokenValidationParameters();
@@ -55,7 +62,7 @@ namespace Microsoft.Extensions.DependencyInjection
 
                             o.TokenValidationParameters = CreateTokenValidationParameters();
 
-                            // TODO: DI (FACAVAL) Remove this once th work above is completed.
+                            // TODO: DI (FACAVAL) Remove this once the work above is completed.
                             if (!SystemEnvironment.Instance.IsPlaceholderModeEnabled())
                             {
                                 // We're not in standby mode, so flag as specialized
@@ -65,17 +72,23 @@ namespace Microsoft.Extensions.DependencyInjection
 
         private static TokenValidationParameters CreateTokenValidationParameters()
         {
-            string defaultKey = Util.GetDefaultKeyValue();
-
             var result = new TokenValidationParameters();
-            if (defaultKey != null)
+            if (SecretsUtility.TryGetEncryptionKey(out byte[] key))
             {
                 // TODO: Once ScriptSettingsManager is gone, Audience and Issuer shouold be pulled from configuration.
-                result.IssuerSigningKey = new SymmetricSecurityKey(Encoding.UTF8.GetBytes(defaultKey));
+                result.IssuerSigningKey = new SymmetricSecurityKey(key);
                 result.ValidateAudience = true;
                 result.ValidateIssuer = true;
-                result.ValidAudience = string.Format(AdminJwtValidAudienceFormat, ScriptSettingsManager.Instance.GetSetting(AzureWebsiteName));
-                result.ValidIssuer = string.Format(AdminJwtValidIssuerFormat, ScriptSettingsManager.Instance.GetSetting(AzureWebsiteName));
+                result.ValidAudiences = new string[]
+                {
+                    string.Format(AdminJwtValidAudienceFormat, ScriptSettingsManager.Instance.GetSetting(AzureWebsiteName)),
+                    AdminJwtAppServiceIssuer
+                };
+                result.ValidIssuers = new string[]
+                {
+                    string.Format(AdminJwtValidIssuerFormat, ScriptSettingsManager.Instance.GetSetting(AzureWebsiteName)),
+                    AdminJwtAppServiceIssuer
+                };
             }
 
             return result;
--- a/src/WebJobs.Script.WebHost/Security/SecretsUtility.cs
+++ b/src/WebJobs.Script.WebHost/Security/SecretsUtility.cs
@@ -2,11 +2,12 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
-using System.IO;
+using System.Linq;
+using Microsoft.Azure.Web.DataProtection;
 
 namespace Microsoft.Azure.WebJobs.Script.WebHost
 {
-    internal class SecretsUtility
+    internal static class SecretsUtility
     {
         public static string GetNonDecryptableName(string secretsPath)
         {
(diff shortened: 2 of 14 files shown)


src/WebJobs.Script/ScriptConstants.cs
```diff
@@ -101,6 +101,7 @@ public static class ScriptConstants
         public const string AntaresScaleOutHeaderName = "X-FUNCTION-SCALEOUT";
         public const string AntaresColdStartHeaderName = "X-MS-COLDSTART";
         public const string SiteTokenHeaderName = "x-ms-site-restricted-token";
+        public const string FunctionsTokenHeaderName = "x-functions-token";
```
mathewc, line 104: The naming chosen here matches the scheme we use for key based auth, where we use "x-functions-key".

liliankasem (approved):

src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
```diff
@@ -74,8 +82,16 @@ private static TokenValidationParameters CreateTokenValidationParameters()
                 result.IssuerSigningKey = new SymmetricSecurityKey(Encoding.UTF8.GetBytes(defaultKey));
                 result.ValidateAudience = true;
                 result.ValidateIssuer = true;
-                result.ValidAudience = string.Format(AdminJwtValidAudienceFormat, ScriptSettingsManager.Instance.GetSetting(AzureWebsiteName));
-                result.ValidIssuer = string.Format(AdminJwtValidIssuerFormat, ScriptSettingsManager.Instance.GetSetting(AzureWebsiteName));
+                result.ValidAudiences = new string[]
+                {
+                    string.Format(AdminJwtValidAudienceFormat, ScriptSettingsManager.Instance.GetSetting(AzureWebsiteName)),
+                    AdminJwtAppServiceIssuer
```
mathewc, line 85: Supporting a more general aud/iss naming scheme.

src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
```diff
@@ -31,8 +30,16 @@ public static AuthenticationBuilder AddScriptJwtBearer(this AuthenticationBuilde
                             {
                                 OnMessageReceived = c =>
                                 {
+                                    // By default, tokens are passed via the standard Authorization Bearer header. However we also support
+                                    // passing tokens via the x-ms-site-token header.
+                                    if (c.Request.Headers.TryGetValue(ScriptConstants.SiteTokenHeaderName, out StringValues values))
```
mathewc, line 35: This is the key change. The rest of the changes are refactorings

src/WebJobs.Script.WebHost/Security/SecretsUtility.cs
```diff
@@ -17,5 +18,71 @@ public static string GetNonDecryptableName(string secretsPath)
             }
             return secretsPath + $".{ScriptConstants.Snapshot}.{timeStamp}.json";
         }
+
+        public static bool TryGetEncryptionKey(out byte[] key, IEnvironment environment = null)
```
mathewc, line 22: For our two token auth mechanisms, you can see that we were using separate logic to get the key. I want to centralize this logic to discrepancies don't crop up. As you can see, we already had different resolution logic for the key in the two token paths.

src/WebJobs.Script.WebHost/Security/SimpleWebTokenHelper.cs
```diff
@@ -90,19 +87,7 @@ public static string Decrypt(byte[] encryptionKey, string value)
 
         public static string Decrypt(string value, IEnvironment environment = null)
         {
-            if ((environment != null) && environment.IsKubernetesManagedHosting())
```
mathewc, line 93: Moved this to the new shared helper, maintaining precedence

src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
```diff
@@ -65,17 +72,23 @@ public static AuthenticationBuilder AddScriptJwtBearer(this AuthenticationBuilde
 
         private static TokenValidationParameters CreateTokenValidationParameters()
         {
-            string defaultKey = Util.GetDefaultKeyValue();
-
             var result = new TokenValidationParameters();
-            if (defaultKey != null)
+            if (SecretsUtility.TryGetEncryptionKey(out byte[] key))
```
mathewc, line 76: Previously we were relying solely on the DataProtection resolution, which only supported a few of the key environment settings (AzureWebEncryptionKey and MACHINEKEY_DecryptionKey). The shared helper puts this into alignment with the other token handling.

src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
```diff
@@ -31,8 +30,16 @@ public static AuthenticationBuilder AddScriptJwtBearer(this AuthenticationBuilde
                             {
                                 OnMessageReceived = c =>
                                 {
+                                    // By default, tokens are passed via the standard Authorization Bearer header. However we also support
+                                    // passing tokens via the x-ms-site-token header.
+                                    if (c.Request.Headers.TryGetValue(ScriptConstants.SiteTokenHeaderName, out StringValues values))
+                                    {
+                                        // the token we set here will be the one used - Authorization header won't be checked.
+                                        c.Token = values.FirstOrDefault();
```
jviau, line 38: Does this have any impact of lines 43 to 46? The `CreateTokenValidationParameters();`. I don't know the precise behavior of auth middleware, but I am guessing by us setting `c.Token` here, we will short-circuit the default behavior of extracting a token from `Authorization` header. I want to make sure what looks like validation customization below still operates as expected in this scenario.

jviau, line 38: For my own understanding: do we expect this token to be the JWT directly, or will it be like `Authorization` header where it is `Authorization: Bearer {token}`?

src/WebJobs.Script/ScriptConstants.cs
```diff
@@ -100,7 +100,8 @@ public static class ScriptConstants
         public const string AntaresLogIdHeaderName = "X-ARR-LOG-ID";
         public const string AntaresScaleOutHeaderName = "X-FUNCTION-SCALEOUT";
         public const string AntaresColdStartHeaderName = "X-MS-COLDSTART";
-        public const string SiteTokenHeaderName = "x-ms-site-restricted-token";
+        public const string SiteRestrictedTokenHeaderName = "x-ms-site-restricted-token";
+        public const string SiteTokenHeaderName = "x-ms-site-token";
```
jviau, line 104: Curious how we decided on this header name? Are there existing examples of this header name being used? If not, are there any examples out there of an alt-auth header that we should consider re-using?

If we arbitrarily decided on this name and don't think it is worth spending much time on picking the best name I am fine with that - so long as we at least acknowledge this is the route we are choosing.

src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
```diff
@@ -65,17 +72,23 @@ public static AuthenticationBuilder AddScriptJwtBearer(this AuthenticationBuilde
...
+            if (SecretsUtility.TryGetEncryptionKey(out byte[] key))
             {
                 // TODO: Once ScriptSettingsManager is gone, Audience and Issuer shouold be pulled from configuration.
-                result.IssuerSigningKey = new SymmetricSecurityKey(Encoding.UTF8.GetBytes(defaultKey));
+                result.IssuerSigningKey = new SymmetricSecurityKey(key);
                 result.ValidateAudience = true;
                 result.ValidateIssuer = true;
-                result.ValidAudience = string.Format(AdminJwtValidAudienceFormat, ScriptSettingsManager.Instance.GetSetting(AzureWebsiteName));
-                result.ValidIssuer = string.Format(AdminJwtValidIssuerFormat, ScriptSettingsManager.Instance.GetSetting(AzureWebsiteName));
+                result.ValidAudiences = new string[]
```
jviau, line 82: tiny performance nit: if we don't expect these arrays to change over the lifetime of a host, can we cache these arrays instead of new'ing them up for each request?

src/WebJobs.Script.WebHost/Security/SecretsUtility.cs
```diff
@@ -17,5 +18,83 @@ public static string GetNonDecryptableName(string secretsPath)
...
+            {
+                var podEncryptionKey = environment.GetEnvironmentVariable(EnvironmentSettingNames.PodEncryptionKey);
+                if (!string.IsNullOrEmpty(podEncryptionKey))
+                {
+                    key = Convert.FromBase64String(podEncryptionKey);
+                    return true;
+                }
+            }
+
+            // Use WebSiteAuthEncryptionKey if available else fallback to ContainerEncryptionKey.
```
madelinegordon:
```suggestion
            // Use WebSiteAuthEncryptionKey if available else fall back to ContainerEncryptionKey.
```

src/WebJobs.Script.WebHost/Security/SecretsUtility.cs
```diff
@@ -17,5 +18,83 @@ public static string GetNonDecryptableName(string secretsPath)
...
+        }
+
+        public static byte[] ToKeyBytes(this string hexOrBase64)
+        {
+            // only support 32 bytes (256 bits) key length
+            if (hexOrBase64.Length == 64)
+            {
+                return Enumerable.Range(0, hexOrBase64.Length)
+                    .Where(x => x % 2 == 0)
+                    .Select(x => Convert.ToByte(hexOrBase64.Substring(x, 2), 16))
```
jviau, line 74: A note for a potential change in a different PR. This is fairly expensive, especially for what I think could be in a hot path.

1. use `foreach` instead of LINQ, it is less GC pressure
2. use `byte.Parse` or with a `ReadOnlySpan<char>` ([Convert.ToByte](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Convert.cs#L781) calls `byte.Parse` anyways).

This way we go from string to byte[] without any extra allocations.

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/JwtTokenAuthTests.cs
```diff
@@ -0,0 +1,136 @@
...
+        {
+            _fixture = fixture;
+        }
+
+        [Theory]
+        [InlineData(nameof(HttpRequestHeader.Authorization))]
+        [InlineData(nameof(HttpRequestHeader.Authorization), ScriptConstants.AdminJwtAppServiceIssuer, ScriptConstants.AdminJwtAppServiceIssuer)]
+        [InlineData(ScriptConstants.SiteTokenHeaderName)]
+        [InlineData(ScriptConstants.SiteTokenHeaderName, ScriptConstants.AdminJwtAppServiceIssuer, ScriptConstants.AdminJwtAppServiceIssuer)]
+        public async Task InvokeAdminApi_ValidToken_Succeeds(string headerName, string aud = null, string iss = null)
```
madelinegordon: nit - Not sure if I would have understood `aud` and `iss` without looking elsewhere in the PR; could name the params `audience` and `issuer` for easier reading?

src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
jviau, line 76: another potential perf optimization for a later PR: can this key be cached? Do we expect it to change over the lifetime of an app?

If we do expect it to change, would be ideal to move it into `IOptions` pattern so that we only regenerate it when its underlying configuration changes.

jviau (approved):

src/WebJobs.Script.WebHost/Security/SecretsUtility.cs
```diff
@@ -17,5 +18,83 @@ public static string GetNonDecryptableName(string secretsPath)
...
+            return Convert.FromBase64String(hexOrBase64);
+        }
+
+        private static bool TryGetEncryptionKey(IEnvironment environment, string keyName, out byte[] encryptionKey, bool throwIfFailed = true)
+        {
+            encryptionKey = null;
+            var hexOrBase64 = environment.GetEnvironmentVariable(keyName);
+            if (string.IsNullOrEmpty(hexOrBase64))
+            {
+                if (throwIfFailed)
```
madelinegordon: Just wondering for understanding - why is throwing optional here?

src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
mathewc: The authenticate method for this middleware checks whether a token value has been specified [here](https://github.com/aspnet/Security/blob/26d27d871b7992022c082dc207e3d126e1d9d278/src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs#L63) and uses it if so. We're using this extensibility as designed.

src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
mathewc: Yes, I thought of the same thing as well, but didn't want to include that here. Yes, we could have been caching this all along for the lifetime of a host. We can do that in the future.

src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
mathewc: The AddJwtBearer options config delegate above makes it hard to see, but these options are only created once when the host is built, not on each request.

src/WebJobs.Script.WebHost/Security/SecretsUtility.cs
mathewc: Yeah, I just moved this code as is :) Agree with your suggestion though that we should fix this.

src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
mathewc: It's the token directly, as per contract with handler code [here](https://github.com/aspnet/Security/blob/26d27d871b7992022c082dc207e3d126e1d9d278/src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs#L75).

src/WebJobs.Script.WebHost/Security/SecretsUtility.cs
mathewc: Fixed

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/JwtTokenAuthTests.cs
mathewc: Done!

src/WebJobs.Script.WebHost/Security/SecretsUtility.cs
mathewc: I remove this code.

liliankasem (approved):

jviau:
Looking at this more, the existing `x-ms-restricted-site-token` will be authenticated as the `ArmToken` authentication scheme and the `x-ms-site-token` as `Bearer` authentication scheme (shares the same policy with the `Authentication` header for other JWT tokens). Is this something we want? Or should we have the `x-ms-site-token` be part of `ArmToken` scheme?

PR #9186, opened 2023-03-29
Add support for additional key encodings

mathewc:
In my PR https://github.com/Azure/azure-functions-host/pull/9168 I inadvertently changed the encoding being used for the key signing key when creating TokenValidationParameters. The previous logic before my change is [here](https://github.com/Azure/azure-functions-host/blob/e7e3df5805f1417c0b2b06f1e795edfff820f039/src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs#L74) and used UTF8. This PR retains support for the previous UTF8 encoding while also supporting hex/base64 encodings.

I'm also adjusting the list of valid audience/issuers, while retaining back compat.

### DIFF
--- a/test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/JwtTokenAuthTests.cs
+++ b/test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/JwtTokenAuthTests.cs
@@ -6,6 +6,7 @@ using System.Linq;
 using System.Net;
 using System.Net.Http;
 using System.Net.Http.Headers;
+using System.Text;
 using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Script.WebHost;
 using Microsoft.Azure.WebJobs.Script.WebHost.Management;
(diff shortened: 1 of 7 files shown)


liliankasem (approved):

kshyju (approved):

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/JwtTokenAuthTests.cs
```diff
@@ -72,6 +73,18 @@ public async Task InvokeAdminApi_InvalidToken_Fails(string headerName)
             Assert.Equal(HttpStatusCode.Unauthorized, response.StatusCode);
         }
 
+        [Fact]
+        public async Task InvokeAdminApi_ValidToken_UTF8Encoding_Succeeds()
```
mathewc, line 77: This is just a sanity test since I removed a Convert.FromBase64String conversion above on line 31 and we now use ToKeyBytes in all cases which handles both hex and base64.

test/WebJobs.Script.Tests/Helpers/SecretsUtilityTest.cs
```diff
@@ -0,0 +1,25 @@
...
+using System;
+using Microsoft.Azure.WebJobs.Script.WebHost;
+using Xunit;
+
+namespace Microsoft.Azure.WebJobs.Script.Tests.Helpers
+{
+    public class SecretsUtilityTest
+    {
+        [Fact]
+        public void ToKeyBytes_ReturnsExpectedValue()
```
mathewc, line 13: This is just a sanity test since I removed a Convert.FromBase64String conversion above on line 31 and we now use ToKeyBytes in all cases which handles both hex and base64.

eliaslopezgt (approved):

jviau (approved):

PR #9193, opened 2023-03-31
[Enhanced Http] Support for http proxying scenarios

satvu:

Add initial components for the enhanced HTTP model (read more [here](https://github.com/Azure/azure-functions-dotnet-worker/issues/1387)). These components are hidden behind a capability and this code path should only light up when a worker implementing the HTTP model provides the necessary info.

These are minimal changes needed to make it easier for partner teams and stakeholders to use and provide feedback as we continue to develop this feature.

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
@@ -731,8 +731,23 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
 
                 var invocationRequest = await context.ToRpcInvocationRequest(_workerChannelLogger, _workerCapabilities, _isSharedMemoryDataTransferEnabled, _sharedMemoryManager);
 
+                AddAdditionalTraceContext(invocationRequest.TraceContext.Attributes, context);
+                _executingInvocations.TryAdd(invocationRequest.InvocationId, context);
+
+                _metricsLogger.LogEvent(string.Format(MetricEventNames.WorkerInvoked, Id), functionName: context.FunctionMetadata.Name);
+
+                var grpcTask = SendStreamingMessageAsync(new StreamingMessage
+                {
+                    InvocationRequest = invocationRequest
+                });
+
+                if (_cancelCapabilityEnabled)
+                {
+                    context.CancellationToken.Register(() => SendInvocationCancel(invocationRequest.InvocationId));
+                }
+
                 // hard code this for prototyping - automatically forward http requests for http trigger functions
-                if (context.FunctionMetadata.Trigger.Type.Contains("httpTrigger"))
+                if (!string.IsNullOrEmpty(_workerCapabilities.GetCapabilityState(RpcWorkerConstants.EnableHttpProxying)))
                 {
                     var handler = new SocketsHttpHandler();
                     var invoker = new HttpMessageInvoker(handler);
@@ -741,28 +756,16 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
                     HttpContext httpContext = httpRequest.HttpContext;
 
                     // add invocation id as correlation id
+                    // TODO: add "invocation-id" as a constant somewhere / maybe find a better name
                     httpRequest.Headers.TryAdd("invocation-id", context.ExecutionContext.InvocationId.ToString());
 
-                    // so http request comes in as an asp.net type (3rd input).
-                    _ = _httpForwarder.SendAsync(httpContext, "http://localhost:5555/", invoker, options, static (context, request) =>
-                    {
-                        return ValueTask.CompletedTask;
-                    });
+                    var aspNetTask = _httpForwarder.SendAsync(httpContext, "http://localhost:5555/", invoker, options);
+
+                    await Task.WhenAll(grpcTask.AsTask(), aspNetTask.AsTask());
                 }
-
-                AddAdditionalTraceContext(invocationRequest.TraceContext.Attributes, context);
-                _executingInvocations.TryAdd(invocationRequest.InvocationId, context);
-
-                _metricsLogger.LogEvent(string.Format(MetricEventNames.WorkerInvoked, Id), functionName: context.FunctionMetadata.Name);
-
-                await SendStreamingMessageAsync(new StreamingMessage
+                else
                 {
-                    InvocationRequest = invocationRequest
-                });
-
-                if (_cancelCapabilityEnabled != null && _cancelCapabilityEnabled.Value)
-                {
-                    context.CancellationToken.Register(() => SendInvocationCancel(invocationRequest.InvocationId));
+                    await grpcTask;
                 }
             }
             catch (Exception invokeEx)
--- a/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
@@ -10,6 +10,7 @@ using System.IO;
 using System.Linq;
 using System.Net.Http;
 using System.Reactive.Linq;
+using System.Reflection.Metadata;
 using System.Text;
 using System.Threading;
 using System.Threading.Channels;
@@ -85,6 +86,8 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
         private bool? _cancelCapabilityEnabled;
         private bool _isWorkerApplicationInsightsLoggingEnabled;
         private IHttpProxyService _httpProxyService;
+        private Uri _httpProxyEndpoint;
+        private bool _isHttpProxyingWorker = false;
 
         private System.Timers.Timer _timer;
 
@@ -426,6 +429,14 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
                 _isWorkerApplicationInsightsLoggingEnabled = true;
             }
 
+            // If http proxying is enabled, we need to get the proxying endpoint of this worker
+            var httpUri = _workerCapabilities.GetCapabilityState(RpcWorkerConstants.HttpUri);
+            if (!string.IsNullOrEmpty(httpUri))
+            {
+                _httpProxyEndpoint = new Uri(httpUri);
+                _isHttpProxyingWorker = true;
+            }
+
             _workerInitTask.TrySetResult(true);
         }
 
@@ -745,9 +756,9 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
                     context.CancellationToken.Register(() => SendInvocationCancel(invocationRequest.InvocationId));
                 }
 
-                if (!string.IsNullOrEmpty(_workerCapabilities.GetCapabilityState(RpcWorkerConstants.EnableHttpProxying)) && context.FunctionMetadata.IsHttpTriggerFunction())
+                if (_isHttpProxyingWorker && context.FunctionMetadata.IsHttpTriggerFunction())
                 {
-                    var aspNetTask = _httpProxyService.Forward(context);
+                    var aspNetTask = _httpProxyService.Forward(context, _httpProxyEndpoint);
 
                     context.Properties.Add("HttpProxyingTask", aspNetTask);
                 }
--- a/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
@@ -740,11 +740,11 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
                     HttpRequest httpRequest = context.Inputs.FirstOrDefault(i => i.Val is HttpRequest).Val as HttpRequest;
                     HttpContext httpContext = httpRequest.HttpContext;
 
-                    // add function id to headers as a correlation id
-                    // httpRequest.Headers.Add("function-id", context.FunctionMetadata.GetFunctionId());
+                    // add invocation id as correlation id
+                    httpRequest.Headers.TryAdd("invocation-id", context.ExecutionContext.InvocationId.ToString());
 
                     // so http request comes in as an asp.net type (3rd input).
-                    await _httpForwarder.SendAsync(httpContext, "http://localhost:5555/", invoker, options, static (context, request) =>
+                    _ = _httpForwarder.SendAsync(httpContext, "http://localhost:5555/", invoker, options, static (context, request) =>
                     {
                         return ValueTask.CompletedTask;
                     });
--- a/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
@@ -764,7 +764,7 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
                 {
                     var aspNetTask = _httpProxyService.Forward(context, _httpProxyEndpoint);
 
-                    context.Properties.Add("HttpProxyingTask", aspNetTask);
+                    context.Properties.Add(ScriptConstants.HttpProxyTask, aspNetTask);
                 }
             }
             catch (Exception invokeEx)
@@ -930,7 +930,7 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
             {
                 if (invokeResponse.Result.IsInvocationSuccess(context.ResultSource, capabilityEnabled))
                 {
-                    if (context.Properties.TryGetValue("HttpProxyingTask", out ValueTask<ForwarderError> httpProxyTask))
+                    if (context.Properties.TryGetValue(ScriptConstants.HttpProxyTask, out ValueTask<ForwarderError> httpProxyTask))
                     {
                         ForwarderError httpProxyTaskResult = await httpProxyTask;
 
--- a/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
@@ -31,7 +31,6 @@ using Microsoft.Azure.WebJobs.Script.ManagedDependencies;
 using Microsoft.Azure.WebJobs.Script.Workers;
 using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
 using Microsoft.Azure.WebJobs.Script.Workers.SharedMemoryDataTransfer;
-using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Yarp.ReverseProxy.Forwarder;
@@ -85,7 +84,7 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
         private bool _isSharedMemoryDataTransferEnabled;
         private bool? _cancelCapabilityEnabled;
         private bool _isWorkerApplicationInsightsLoggingEnabled;
-        private IHttpForwarder _httpForwarder;
+        private IHttpProxyService _httpProxyService;
 
         private System.Timers.Timer _timer;
 
@@ -102,7 +101,7 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
             ISharedMemoryManager sharedMemoryManager,
             IOptions<WorkerConcurrencyOptions> workerConcurrencyOptions,
             IOptions<FunctionsHostingConfigOptions> hostingConfigOptions,
-            IHttpForwarder httpForwarder)
+            IHttpProxyService httpProxyService)
         {
             _workerId = workerId;
             _eventManager = eventManager;
@@ -118,7 +117,7 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
             _processInbound = state => ProcessItem((InboundGrpcEvent)state);
             _hostingConfigOptions = hostingConfigOptions;
 
-            _httpForwarder = httpForwarder;
+            _httpProxyService = httpProxyService;
             _workerCapabilities = new GrpcCapabilities(_workerChannelLogger);
 
             if (!_eventManager.TryGetGrpcChannels(workerId, out var inbound, out var outbound))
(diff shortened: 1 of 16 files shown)


satvu:
Working on tests but the rest is ready for review.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -420,6 +426,14 @@ internal void WorkerInitResponse(GrpcEvent initEvent)
                 _isWorkerApplicationInsightsLoggingEnabled = true;
             }
 
+            // If http proxying is enabled, we need to get the proxying endpoint of this worker
+            var httpUri = _workerCapabilities.GetCapabilityState(RpcWorkerConstants.HttpUri);
+            if (!string.IsNullOrEmpty(httpUri))
+            {
+                _httpProxyEndpoint = new Uri(httpUri);
+                _isHttpProxyingWorker = true;
```
jviau: Can this be a property based on `_httpProxyEndpoint`?

``` CSharp
private bool IsHttpProxyingWorker => _httpProxyEndpoint is not null;
```

src/WebJobs.Script.Grpc/Server/IHttpProxyService.cs
```diff
@@ -0,0 +1,15 @@
...
+using System;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Script.Description;
+using Yarp.ReverseProxy.Forwarder;
+
+namespace Microsoft.Azure.WebJobs.Script.Grpc
+{
+    public interface IHttpProxyService
+    {
+        public ValueTask<ForwarderError> Forward(ScriptInvocationContext context, Uri httpUri);
```
jviau: nit: 
```suggestion
        public ValueTask<ForwarderError> ForwardAsync(ScriptInvocationContext context, Uri httpUri);
```

Also, should this take in a `CancellationToken`? Or is one available on the context object?

src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
```diff
@@ -0,0 +1,69 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.Azure.WebJobs.Script.Description;
+using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
+using Yarp.ReverseProxy.Forwarder;
+
+namespace Microsoft.Azure.WebJobs.Script.Grpc
+{
+    internal class DefaultHttpProxyService : IHttpProxyService, IDisposable
+    {
+        private SocketsHttpHandler _handler;
```
jviau: nit: please name all of these fields `readonly` (unless they need re-assignment).

src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
```diff
@@ -0,0 +1,69 @@
...
+        private IHttpForwarder _httpForwarder;
+        private HttpMessageInvoker _messageInvoker;
+        private ForwarderRequestConfig _forwarderRequestConfig;
+
+        public DefaultHttpProxyService(IHttpForwarder httpForwarder)
+        {
+            _httpForwarder = httpForwarder;
```
jviau: nit: check for null. I think being .net6 you have these two ways to do it:

``` csharp
ArgumentNullException.ThrowIfNull(httpForwarder);
_httpForwarder = httpForwarder
```

or

``` csharp
_httpForwader = httpForwarder ?? throw new ArgumentNullException(nameof(httpForwarder);
```

src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        public void Dispose()
+        {
+            _handler?.Dispose();
+            _messageInvoker?.Dispose();
+        }
+
+        public ValueTask<ForwarderError> Forward(ScriptInvocationContext context, Uri httpUri)
+        {
+            if (context is null)
```
jviau: I think we are .net6 in this project right? If so, you can shorten this to:

``` csharp
ArgumentNullException.ThrowIfNull(context);
```

src/WebJobs.Script/ScriptConstants.cs
```diff
@@ -225,5 +225,9 @@ public static class ScriptConstants
         public static readonly string LiveLogsSessionAIKey = "#AzFuncLiveLogsSessionId";
 
         public static readonly string FunctionsHostingConfigSectionName = "FunctionsHostingConfig";
+
+        // HTTP Proxying constants
+        public static readonly string IsHttpProxyingEnabled = "IsHttpProxying";
+        public static readonly string HttpProxyCorrelationHeader = "x-invocation-id";
```
jviau: Going by other extension headers, I think we want `x-ms-` as the prefix.
```suggestion
        public static readonly string HttpProxyCorrelationHeader = "x-ms-invocation-id";
```

test/WebJobs.Script.Tests/Microsoft.Azure.WebJobs.Script.WebHost.deps.json
```diff
@@ -2190,11 +2190,11 @@
           "System.Runtime": "4.3.1"
         }
       },
-      "System.IO.Hashing/6.0.0": {
```
satvu, line 2193: https://www.fuget.org/packages/System.IO.Hashing/7.0.0/lib/netstandard2.0/diff/6.0.0/

Reviewed the above w/ @fabiocav we believe this is a safe change.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -420,6 +426,14 @@ internal void WorkerInitResponse(GrpcEvent initEvent)
                 _isWorkerApplicationInsightsLoggingEnabled = true;
             }
 
+            // If http proxying is enabled, we need to get the proxying endpoint of this worker
+            var httpUri = _workerCapabilities.GetCapabilityState(RpcWorkerConstants.HttpUri);
+            if (!string.IsNullOrEmpty(httpUri))
+            {
+                _httpProxyEndpoint = new Uri(httpUri);
```
surgupta-msft: We should check if `httpUri` has the correct format and also handle cases where `new Uri(httpUri)` can throw exceptions.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
surgupta-msft: Do we need these 2 variables at global level and in this function? Can I not directly check for capability and assign _httpProxyEndpoint in `SendInvocationRequest()`?

Nit - If variable `_isHttpProxyingWorker` is needed - not sure if "proxying" is appropriate to use. In case we want to change it, some alternatives are - `_WorkerSupportsHttpProxy` or `_IsWorkerHttpProxyEnabled` or something similar.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -420,6 +426,14 @@ internal void WorkerInitResponse(GrpcEvent initEvent)
                 _isWorkerApplicationInsightsLoggingEnabled = true;
             }
 
+            // If http proxying is enabled, we need to get the proxying endpoint of this worker
+            var httpUri = _workerCapabilities.GetCapabilityState(RpcWorkerConstants.HttpUri);
```
surgupta-msft, line 430: nit - Are we checking somewhere else if HttpProxying is enabled and then fetching `RpcWorkerConstants.HttpUri`? The comment above is bit confusing for me. If `HttpUri` is acting as a capability, then may be we can rename it to `SupportsHttpUri` or `HandlesHttpUri` or `SupportsHttpProxing` or something similar.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -738,6 +752,13 @@ internal async Task SendInvocationRequest(ScriptInvocationContext context)
                 {
                     context.CancellationToken.Register(() => SendInvocationCancel(invocationRequest.InvocationId));
                 }
+
+                if (_isHttpProxyingWorker && context.FunctionMetadata.IsHttpTriggerFunction())
+                {
+                    var aspNetTask = _httpProxyService.Forward(context, _httpProxyEndpoint);
+
+                    context.Properties.Add("HttpProxyingTask", aspNetTask);
```
surgupta-msft: Defining constant for string literal `HttpProxyingTask`

src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
```diff
@@ -0,0 +1,69 @@
...
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            if (context.Inputs is null)
```
surgupta-msft: Should we also add Empty check here?

src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
```diff
@@ -0,0 +1,69 @@
...
+            {
+                throw new InvalidOperationException($"Cannot proxy the HttpTrigger function {context.FunctionMetadata.Name} without an input of type {nameof(HttpRequest)}.");
+            }
+
+            HttpContext httpContext = httpRequest.HttpContext;
+
+            httpContext.Items.Add(ScriptConstants.IsHttpProxyingEnabled, bool.TrueString);
+
+            // add invocation id as correlation id
+            httpRequest.Headers.TryAdd(ScriptConstants.HttpProxyCorrelationHeader, context.ExecutionContext.InvocationId.ToString());
```
surgupta-msft, line 59: We will need null check if `ExecutionContext` or `InvocationId` can be null here. Ignore if it's not necessary or being already checked somewhere else in the flow.

src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
```diff
@@ -0,0 +1,69 @@
...
+
+            var aspNetTask = _httpForwarder.SendAsync(httpContext, httpUri.ToString(), _messageInvoker, _forwarderRequestConfig);
```
surgupta-msft, line 61: This line can throw exception if arguments (especially httpUri) are not in the expected format. We should also add tests for such unhappy scenarios.

src/WebJobs.Script.WebHost/Middleware/FunctionInvocationMiddleware.cs
```diff
@@ -52,6 +53,15 @@ public async Task Invoke(HttpContext context)
 
                 int nestedProxiesCount = GetNestedProxiesCount(context, functionExecution);
                 IActionResult result = await GetResultAsync(context, functionExecution);
+
+                if (context.Items.TryGetValue(ScriptConstants.IsHttpProxyingEnabled, out var value))
+                {
+                    if (value.ToString() == bool.TrueString)
```
surgupta-msft: can `value` be null here? Thinking of a case where it is not added in ScriptConstants but being checked here.

src/WebJobs.Script/Extensions/FunctionMetadataExtensions.cs
```diff
@@ -24,15 +26,20 @@ public static bool IsHttpInAndOutFunction(this FunctionMetadata metadata)
...
             {
                 return true;
             }
 
             return false;
         }
 
+        public static bool IsHttpTriggerFunction(this FunctionMetadata metadata)
+        {
+            return metadata.InputBindings.Any(b => string.Equals(HttpTriggerKey, b.Type, StringComparison.OrdinalIgnoreCase));
```
surgupta-msft, line 40: null check for metadata and InputBindings.

src/WebJobs.Script/ScriptConstants.cs
```diff
@@ -225,5 +225,9 @@ public static class ScriptConstants
         public static readonly string LiveLogsSessionAIKey = "#AzFuncLiveLogsSessionId";
 
         public static readonly string FunctionsHostingConfigSectionName = "FunctionsHostingConfig";
+
+        // HTTP Proxying constants
+        public static readonly string IsHttpProxyingEnabled = "IsHttpProxying";
```
surgupta-msft: nit - `HttpProxyEnabled` or `SupportsHttpProxy` since it's a string

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -81,6 +82,9 @@ internal class GrpcWorkerChannel : IRpcWorkerChannel, IDisposable
         private bool _isSharedMemoryDataTransferEnabled;
         private bool? _cancelCapabilityEnabled;
         private bool _isWorkerApplicationInsightsLoggingEnabled;
+        private IHttpProxyService _httpProxyService;
+        private Uri _httpProxyEndpoint;
+        private bool _isHttpProxyingWorker = false;
```
soninaren: This would be cached when the worker is running under placeholder mode. Likely may not matter for dotnet worker. However, if there is customer config that controls the behavior then we may run into caching issues.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
satvu: Regarding names it was discussed in the PR into the feature branch (which is now this PR into dev) and we will change them once we decide on final names. Mention of "proxy" is probably only useful internally so this is up for debate still.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
satvu: We are calling it this because the working sends a Uri (string form) over instead of `bool.TrueString`

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -420,6 +426,21 @@ internal void WorkerInitResponse(GrpcEvent initEvent)
                 _isWorkerApplicationInsightsLoggingEnabled = true;
             }
 
+            // If http proxying is enabled, we need to get the proxying endpoint of this worker
+            var httpUri = _workerCapabilities.GetCapabilityState(RpcWorkerConstants.HttpUri);
+            if (!string.IsNullOrEmpty(httpUri) && FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagEnableHttpProxying))
```
soninaren, line 431: `FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagEnableHttpProxying)` will always be false for worker started in placeholder mode

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -738,6 +759,13 @@ internal async Task SendInvocationRequest(ScriptInvocationContext context)
                 {
                     context.CancellationToken.Register(() => SendInvocationCancel(invocationRequest.InvocationId));
                 }
+
+                if (_isHttpProxyingWorker && context.FunctionMetadata.IsHttpTriggerFunction())
```
soninaren: `IsHttpTriggerFunction` explicitly looks at httpTrigger functions only. This does not include any webhooks from extensions such as durable or trigger that use http such as eventGrid. I am assuming this is by design.

src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
satvu: Since `httpUri` is passed in as a paremter (of type Uri), the result of `toString()` should be valid.

src/WebJobs.Script.Grpc/GrpcServiceCollectionsExtensions.cs
```diff
@@ -16,6 +16,9 @@ public static IServiceCollection AddScriptGrpc(this IServiceCollection services)
 
             services.AddSingleton<IRpcServer, AspNetCoreGrpcServer>();
 
+            services.AddHttpForwarder();
```
satvu, line 19: Looking for a clean way to avoid adding this without a feature flag.

src/WebJobs.Script.Grpc/GrpcServiceCollectionsExtensions.cs
```diff
@@ -16,6 +16,9 @@ public static IServiceCollection AddScriptGrpc(this IServiceCollection services)
...
+            services.AddSingleton<IHttpProxyService, DefaultHttpProxyService>();
```
satvu, line 20: Unfortunately, we can't get this behind the feature flag (can't access `FeatureFlags` yet at this point, host will crash) so these services will be added but not used.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -902,6 +930,17 @@ internal async Task InvokeResponse(InvocationResponse invokeResponse)
             {
                 if (invokeResponse.Result.IsInvocationSuccess(context.ResultSource, capabilityEnabled))
```
soninaren, line 930: What happens if `IsInvocationSuccess` is false but the proxy request made it through. Also what happens when `IsInvocationSuccess` is false and proxy request has failed.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -738,6 +759,13 @@ internal async Task SendInvocationRequest(ScriptInvocationContext context)
                 {
                     context.CancellationToken.Register(() => SendInvocationCancel(invocationRequest.InvocationId));
                 }
+
+                if (_isHttpProxyingWorker && context.FunctionMetadata.IsHttpTriggerFunction())
+                {
+                    var aspNetTask = _httpProxyService.Forward(context, _httpProxyEndpoint);
```
soninaren: Are there any down sides of awaiting the task here? From what i am understanding the proxy request is already made here and awaiting the task simply provides with the forward errors if any. we can store the proxy result in the context and log it along with invocation success of failure.

satvu: Original design awaited the task here - I believe it changed because this task often completes after function invocation due to logic on the worker side (if there are no errors, this awaits until the http response is written back) and we decided to just send it off and unify if the function invocation succeeds.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
satvu: The design for these cases hasn't been discussed but for now if `IsInvocationSuccess` fails at all we will ignore any proxying results and log an error.

test/WebJobs.Script.Tests/Microsoft.Azure.WebJobs.Script.WebHost.deps.json
jviau: Curious why this file is included at all in our code base? Being a build/restore artifact, shouldn't it be excluded?

src/WebJobs.Script.Grpc/Server/IHttpProxyService.cs
satvu: @jviau Yes `CancellationToken` implementation is something on the to-do list across the entire flow between host and worker.

satvu:
Closing and postponing to next deadline instead.

src/WebJobs.Script.Grpc/Server/IHttpProxyService.cs
```diff
@@ -0,0 +1,15 @@
...
+using System;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Script.Description;
+using Yarp.ReverseProxy.Forwarder;
+
+namespace Microsoft.Azure.WebJobs.Script.Grpc
+{
+    public interface IHttpProxyService
+    {
+        public ValueTask<ForwarderError> ForwardAsync(ScriptInvocationContext context, Uri httpUri);
```
jviau: small nit: don't need to include `public` on interfaces, they are public by default.
```suggestion
        ValueTask<ForwarderError> ForwardAsync(ScriptInvocationContext context, Uri httpUri);
```

soninaren (approved):

satvu:
Noting again for the record this is for _private preview_ and reopening again because we want to reemphasize the main concerns:

- are we concerned about reliability?
- are we concerned about our ability to iterate

If both are no, then this work should be fine for now. We want to put this in the hands of some PMs and ASP.NET partners for further validation.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -420,6 +426,20 @@ internal void WorkerInitResponse(GrpcEvent initEvent)
                 _isWorkerApplicationInsightsLoggingEnabled = true;
             }
 
+            // If http proxying is enabled, we need to get the proxying endpoint of this worker
+            var httpUri = _workerCapabilities.GetCapabilityState(RpcWorkerConstants.HttpUri);
```
jviau, line 430: Should the feature flag check be moved around _all_ of the HTTP proxy logic?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -902,6 +929,17 @@ internal async Task InvokeResponse(InvocationResponse invokeResponse)
             {
                 if (invokeResponse.Result.IsInvocationSuccess(context.ResultSource, capabilityEnabled))
                 {
+                    if (context.Properties.TryGetValue(ScriptConstants.HttpProxyTask, out ValueTask<ForwarderError> httpProxyTask))
```
jviau: Storing value task like this is _slightly_ uncomfortable to me. ValueTask is meant to be a short-lifetime task that **does not support** awaiting multiple times. Depending on the `ValueTask` itself, it might not even be safe to set aside and unwrap later (not entirely sure about this, but looking at `IValueTaskSource` I could see that being potentially true). Bit of a read here: https://devblogs.microsoft.com/dotnet/understanding-the-whys-whats-and-whens-of-valuetask/

You _might_ have to call `ValueTask.AsTask()` and store that in properties, just to be safe. Although if we can review if the two stage invoke and later await is necessary in this path would be best (but we can also review that later).

jviau (approved):

jviau:
So long as it is clear this is private preview and we are okay with making breaking changes later I am fine with this.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
satvu: The other logic is accessed through the _IsHttpProxyingWorker_ var which can only be set to true here when the feature flag is checked - I can do that as well but I was concerned about excessive checking since it would occur in every invocation + invocation response.

satvu: Discussed w/ Naren we think it's cheap enough and the extra protection is good

test/WebJobs.Script.Tests/Microsoft.Azure.WebJobs.Script.WebHost.deps.json
satvu: Skipped this comment earlier but I am only editing it as it's needed to pass tests - it's a file in the test repo to make sure all dependency changes are caught and approval is asked from specific code owners (the test error message says so, you can see in earlier pipeline failures).

fabiocav (approved):

PR #9217, opened 2023-04-19
capping the bundle version to 3.19.0 and 4.2.0

soninaren:
Setting the max version of extension bundle v3 runtime can use to 3.19.0 and 4.2.0

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script/Config/FunctionsHostingConfigOptions.cs
+++ b/src/WebJobs.Script/Config/FunctionsHostingConfigOptions.cs
@@ -32,6 +32,28 @@ namespace Microsoft.Azure.WebJobs.Script.Config
         }
 
         /// <summary>
+        /// Gets the highest version of extension bundle v4 supported
+        /// </summary>
+        public string MaximumSupportedBundleV3Version
+        {
+            get
+            {
+                return GetFeature(ScriptConstants.MaximumSupportedBundleV3Version) ?? "3.19.0";
+            }
+        }
+
+        /// <summary>
+        /// Gets the highest version of extension bundle v3 supported
+        /// </summary>
+        public string MaximumSupportedBundleV4Version
+        {
+            get
+            {
+                return GetFeature(ScriptConstants.MaximumSupportedBundleV4Version) ?? "4.2.0";
+            }
+        }
+
+        /// <summary>
         /// Gets feature by name.
         /// </summary>
         /// <param name="name">Feature name.</param>
--- a/src/WebJobs.Script/Config/FunctionsHostingConfigOptions.cs
+++ b/src/WebJobs.Script/Config/FunctionsHostingConfigOptions.cs
@@ -32,7 +32,7 @@ namespace Microsoft.Azure.WebJobs.Script.Config
         }
 
         /// <summary>
-        /// Gets the highest version of extension bundle v4 supported
+        /// Gets the highest version of extension bundle v3 supported
         /// </summary>
         public string MaximumSupportedBundleV3Version
         {
@@ -43,7 +43,7 @@ namespace Microsoft.Azure.WebJobs.Script.Config
         }
 
         /// <summary>
-        /// Gets the highest version of extension bundle v3 supported
+        /// Gets the highest version of extension bundle v4 supported
         /// </summary>
         public string MaximumSupportedBundleV4Version
         {
--- a/src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
+++ b/src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
@@ -10,10 +10,12 @@ using System.Linq;
 using System.Net.Http;
 using System.Resources;
 using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Script.Config;
 using Microsoft.Azure.WebJobs.Script.Configuration;
 using Microsoft.Azure.WebJobs.Script.Diagnostics.Extensions;
 using Microsoft.Azure.WebJobs.Script.Models;
 using Microsoft.Extensions.Logging;
+using Microsoft.WindowsAzure.Storage.Shared.Protocol;
 using Newtonsoft.Json;
 using NuGet.Versioning;
 
@@ -23,16 +25,18 @@ namespace Microsoft.Azure.WebJobs.Script.ExtensionBundle
     {
         private readonly IEnvironment _environment;
         private readonly ExtensionBundleOptions _options;
+        private readonly FunctionsHostingConfigOptions _configOption;
         private readonly ILogger _logger;
         private readonly string _cdnUri;
         private string _extensionBundleVersion;
 
-        public ExtensionBundleManager(ExtensionBundleOptions options, IEnvironment environment, ILoggerFactory loggerFactory)
+        public ExtensionBundleManager(ExtensionBundleOptions options, IEnvironment environment, ILoggerFactory loggerFactory, FunctionsHostingConfigOptions configOption)
         {
             _environment = environment ?? throw new ArgumentNullException(nameof(environment));
             _logger = loggerFactory.CreateLogger<ExtensionBundleManager>() ?? throw new ArgumentNullException(nameof(loggerFactory));
             _cdnUri = _environment.GetEnvironmentVariable(EnvironmentSettingNames.ExtensionBundleSourceUri) ?? ScriptConstants.ExtensionBundleDefaultSourceUri;
             _options = options ?? throw new ArgumentNullException(nameof(options));
+            _configOption = configOption ?? throw new ArgumentNullException(nameof(configOption));
         }
 
         public async Task<ExtensionBundleDetails> GetExtensionBundleDetails()
@@ -128,7 +132,7 @@ namespace Microsoft.Azure.WebJobs.Script.ExtensionBundle
                 if (FileUtility.DirectoryExists(path))
                 {
                     var bundleDirectories = FileUtility.EnumerateDirectories(path);
-                    string version = FindBestVersionMatch(_options.Version, bundleDirectories);
+                    string version = FindBestVersionMatch(_options.Version, bundleDirectories, _options.Id, _configOption);
 
                     if (!string.IsNullOrEmpty(version))
                     {
@@ -236,7 +240,8 @@ namespace Microsoft.Azure.WebJobs.Script.ExtensionBundle
 
             var content = await response.Content.ReadAsStringAsync();
             var bundleVersions = JsonConvert.DeserializeObject<IEnumerable<string>>(content);
-            var matchingBundleVersion = FindBestVersionMatch(_options.Version, bundleVersions);
+
+            var matchingBundleVersion = FindBestVersionMatch(_options.Version, bundleVersions, _options.Id, _configOption);
 
             if (string.IsNullOrEmpty(matchingBundleVersion))
             {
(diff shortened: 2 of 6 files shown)


src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -246,8 +246,13 @@ private async Task<string> GetLatestMatchingBundleVersionAsync(HttpClient httpCl
             return matchingBundleVersion;
         }
 
-        private static string FindBestVersionMatch(VersionRange versionRange, IEnumerable<string> versions)
+        internal static string FindBestVersionMatch(VersionRange versionRange, IEnumerable<string> versions)
         {
+            versionRange = new VersionRange(versionRange.MinVersion,
```
liliankasem: Where is the versionRange parameter coming in from (caller of this method)? Would it make more sense to do this sooner?

liliankasem (dismissed):
just 1 question about if we can change the range sooner otherwise lgtm

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
soninaren: This method is called by two flows, locating extension bundle on local machine and location extension bundle on cdn for download. Doing it sooner would mean we would have to make a change in 2 places.

surgupta-msft (dismissed):

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
liliankasem: thanks for the info!

liliankasem:
Do we also need to add this to v4 as well?

soninaren:
> Do we also need to add this to v4 as well?

No, doing that to v4 would prevent us from using new versions of extension bundle in v4.

liliankasem:
> > Do we also need to add this to v4 as well?
> 
> No, doing that to v4 would prevent us from using new versions of extension bundle in v4.

Then do we need to check for the v4 major version if this will never go into v4?

soninaren:
> > > Do we also need to add this to v4 as well?
> > 
> > 
> > No, doing that to v4 would prevent us from using new versions of extension bundle in v4.
> 
> Then do we need to check for the v4 major version if this will never go into v4?

Bundle v4 is not related to runtime v4. They are independent components. Though I think we should prevent people from using other major versions of extension bundle as well. Will make a small change to not allow using a bundle version higher than v4.

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -23,23 +25,25 @@ public class ExtensionBundleManager : IExtensionBundleManager
...
         private string _extensionBundleVersion;
 
-        public ExtensionBundleManager(ExtensionBundleOptions options, IEnvironment environment, ILoggerFactory loggerFactory)
+        public ExtensionBundleManager(ExtensionBundleOptions options, IEnvironment environment, ILoggerFactory loggerFactory, FunctionsHostingConfigOptions configOption)
         {
             _environment = environment ?? throw new ArgumentNullException(nameof(environment));
             _logger = loggerFactory.CreateLogger<ExtensionBundleManager>() ?? throw new ArgumentNullException(nameof(loggerFactory));
             _cdnUri = _environment.GetEnvironmentVariable(EnvironmentSettingNames.ExtensionBundleSourceUri) ?? ScriptConstants.ExtensionBundleDefaultSourceUri;
             _options = options ?? throw new ArgumentNullException(nameof(options));
+            _configOption = configOption;
```
liliankasem: Do we also want to do this just to be safe ` ?? throw new ArgumentNullException(nameof(configOption));`

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -112,7 +116,7 @@ private async Task<string> GetBundle(HttpClient httpClient)
             return bundlePath;
         }
 
-        internal bool TryLocateExtensionBundle(out string bundlePath)
+        internal bool TryLocateExtensionBundle(out string bundlePath, FunctionsHostingConfigOptions configOption)
```
liliankasem: Do we need this to be passed as a function parameter if it's a global variable in this class?

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -246,7 +251,7 @@ private async Task<string> GetLatestMatchingBundleVersionAsync(HttpClient httpCl
             return matchingBundleVersion;
         }
 
-        private static string FindBestVersionMatch(VersionRange versionRange, IEnumerable<string> versions)
+        internal static string FindBestVersionMatch(VersionRange versionRange, IEnumerable<string> versions, string bundleId, FunctionsHostingConfigOptions configOption)
```
liliankasem, line 254: Same thing here, I feel like you should be able to just reference `configOption` without making this change - if I'm missing some context please let me know :)

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -259,7 +264,26 @@ private static string FindBestVersionMatch(VersionRange versionRange, IEnumerabl
                 return version;
             }).Where(v => v != null);
 
-            return bundleVersions.OrderByDescending(version => version.Version).FirstOrDefault()?.ToString();
+            var matchingVersion = bundleVersions.OrderByDescending(version => version.Version).FirstOrDefault();
+            if (bundleId == ScriptConstants.DefaultExtensionBundleId)
```
liliankasem: nit: if it makes sense, reverse if statement and return early?

liliankasem (dismissed):
some minor comments otherwise lgtm

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
soninaren: It does not, for some reason i thought it was a static method. Have fixed this

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
soninaren: can't access configOption directly since FindBestVersionMatch a static method

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
soninaren: done

liliankasem:
@soninaren how come you have to downgrade the versions? We should also update the PR title

soninaren:
> @soninaren how come you have to downgrade the versions? We should also update the PR title

I think that was in error. I have fixed it.

src/WebJobs.Script/Config/FunctionsHostingConfigOptions.cs
```diff
@@ -31,6 +31,28 @@ public bool FunctionsWorkerDynamicConcurrencyEnabled
...
+        public string MaximumSupportedBundleV3Version
+        {
+            get
+            {
+                return GetFeature(ScriptConstants.MaximumSupportedBundleV3Version) ?? "3.19.0";
+            }
+        }
+
+        /// <summary>
+        /// Gets the highest version of extension bundle v3 supported
```
liliankasem: I think the comments are swapped on these two properties - this should say v4?

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -100,6 +100,7 @@ public static IHostBuilder AddScriptHost(this IHostBuilder builder, Action<Scrip
                 {
                     configBuilder.Add(new HostJsonFileConfigurationSource(applicationOptions, SystemEnvironment.Instance, loggerFactory, metricsLogger));
                 }
+                configBuilder.Add(new FunctionsHostingConfigSource(SystemEnvironment.Instance));
```
jviau, line 103: this is the new config source which leads to the test failure. Will need to update the test. Also will these config source file be available in v3?

And finally - not sure what this looks like in v4, but should we consider this a "Host" configuration? ie, we register it in `ConfigureHostConfiguration` - this way these values are available directly during `ConfigureAppConfiguration` via `HostContext.Configuration` - probably not worth addressing in this PR.

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -23,16 +25,18 @@ public class ExtensionBundleManager : IExtensionBundleManager
     {
         private readonly IEnvironment _environment;
         private readonly ExtensionBundleOptions _options;
+        private readonly FunctionsHostingConfigOptions _configOption;
         private readonly ILogger _logger;
         private readonly string _cdnUri;
         private string _extensionBundleVersion;
 
-        public ExtensionBundleManager(ExtensionBundleOptions options, IEnvironment environment, ILoggerFactory loggerFactory)
+        public ExtensionBundleManager(ExtensionBundleOptions options, IEnvironment environment, ILoggerFactory loggerFactory, FunctionsHostingConfigOptions configOption)
```
jviau, line 33: An alternative approach would also be to add the values you need directly to `ExtensionBundleOptions` and update its configuration helper to pull out the two features we need.

This is more in line with that `IOptions` pattern is intended for. Where the options object you consume is scenario-specific and the construction of it is separated from the usage.

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -259,7 +264,29 @@ private static string FindBestVersionMatch(VersionRange versionRange, IEnumerabl
...
 
-            return bundleVersions.OrderByDescending(version => version.Version).FirstOrDefault()?.ToString();
+            var matchingVersion = bundleVersions.OrderByDescending(version => version.Version).FirstOrDefault();
+
+            if (bundleId != ScriptConstants.DefaultExtensionBundleId)
+            {
+                return matchingVersion?.ToString();
+            }
+
+            var maximumBundleV3Version = NuGetVersion.Parse(configOption.MaximumSupportedBundleV3Version);
```
jviau, line 274: As part of option configuration, you could do the `NuGetVersion` parsing on the options object, and not need to do it here.

jviau (dismissed):
To verify, it is okay we are introducing the new config source to v3.x? Wondering if there will be any issues with the file not being there, or not being in the format we expect compared to v4.

test/CSharpPrecompiledTestProjects/WebJobsStartupTests/Function1.cs
```diff
@@ -93,7 +93,7 @@ private static bool ValidateConfig(IConfiguration _config)
         {
             if (_config is ConfigurationRoot root)
             {
-                if (root.Providers.Count() != 7)
+                if (root.Providers.Count() != 8)
```
jviau, line 96: Also need to adjust the return statement lines 103-110 to expect this new provider.

src/WebJobs.Script/Config/FunctionsHostingConfigOptions.cs
soninaren: Fixed

test/CSharpPrecompiledTestProjects/WebJobsStartupTests/Function1.cs
soninaren: done

soninaren:
> To verify, it is okay we are introducing the new config source to v3.x? Wondering if there will be any issues with the file not being there, or not being in the format we expect compared to v4.



> To verify, it is okay we are introducing the new config source to v3.x? Wondering if there will be any issues with the file not being there, or not being in the format we expect compared to v4.

I tested it with file not present. Every feature in hostfileConfig has a default value. So it assumes that value.

kshyju (dismissed):

kshyju (approved):

PR #9227, opened 2023-04-24
Fixing v3.x build

brettsam:
Something seems to have changed in the v3.x branch and caused our dotnet installations to not work. This change did a little re-arranging to make it easier to fix in the future, but the real fix is adding this to the UseDotNet task:

```
    installationPath: 'C:\Program Files\dotnet'
```

It's unclear why this broke suddenly; I'm looking into an RCA.

### DIFF
--- a/build/install-dotnet.yml
+++ b/build/install-dotnet.yml
@@ -1,0 +1,14 @@
+steps:
+- task: UseDotNet@2
+  inputs:
+    packageType: 'sdk'
+    version: '3.1.x'
+    installationPath: 'C:\Program Files\dotnet'
+    performMultiLevelLookup: true
+- task: PowerShell@2
+  displayName: 'dotnet --info'
+  inputs:
+    targetType: 'inline'
+    script: |
+      Write-Host "PATH: $Env:PATH"
+      & dotnet --info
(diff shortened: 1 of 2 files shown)


build/install-dotnet.yml
```diff
@@ -0,0 +1,14 @@
+steps:
+- task: UseDotNet@2
+  inputs:
+    packageType: 'sdk'
+    version: '3.1.x'
+    installationPath: 'C:\Program Files\dotnet'
+    performMultiLevelLookup: true
+- task: PowerShell@2
+  displayName: 'dotnet --info'
```
brettsam, line 9: added this task for future debugging... i couldn't tell if anything had changed wrt the install logic b/c there wasn't enough logging to tell me how it used to work

build/install-dotnet.yml
```diff
@@ -0,0 +1,14 @@
+steps:
+- task: UseDotNet@2
+  inputs:
+    packageType: 'sdk'
+    version: '3.1.x'
+    installationPath: 'C:\Program Files\dotnet'
```
brettsam, line 6: this line makes it work... previously it was installing to `C:\hostedtoolcache\windows\dotnet`, which suddenly stopped working

soninaren (approved):

PR #9231, opened 2023-04-25
installing .net core 3.1 in CI

brettsam:
Some of our tests rely on .NET Core 3.1, which was recently removed from the CI images. Re-adding it here.

### DIFF
--- a/build/install-dotnet.yml
+++ b/build/install-dotnet.yml
@@ -1,5 +1,12 @@
 steps:
 - task: PowerShell@2
+  displayName: 'Install .NET 3.1'
+  inputs:
+    targetType: 'inline'
+    script: |
+      Invoke-WebRequest 'https://dot.net/v1/dotnet-install.ps1' -OutFile 'dotnet-install.ps1'
+      ./dotnet-install.ps1 -InstallDir 'C:\Program Files\dotnet' -Verbose -Channel 3.1
+- task: PowerShell@2
   displayName: 'Install .NET 6'
   inputs:
     targetType: 'inline'
@@ -8,3 +15,4 @@ steps:
       # Official release versions can be found at: https://dotnet.microsoft.com/download/dotnet/6.0
       # Newer versions can be found at: https://github.com/dotnet/installer#installers-and-binaries
       ./dotnet-install.ps1 -InstallDir 'C:\Program Files\dotnet' -Verbose -Channel 6.0
+      & dotnet --info


build/install-dotnet.yml
```diff
@@ -8,3 +15,4 @@ steps:
       # Official release versions can be found at: https://dotnet.microsoft.com/download/dotnet/6.0
       # Newer versions can be found at: https://github.com/dotnet/installer#installers-and-binaries
       ./dotnet-install.ps1 -InstallDir 'C:\Program Files\dotnet' -Verbose -Channel 6.0
+      & dotnet --info
```
brettsam, line 18: Added this line for future debugging.

VpOfEngineering (approved):

build/install-dotnet.yml
```diff
@@ -1,4 +1,11 @@
 steps:
+- task: PowerShell@2
+  displayName: 'Install .NET 3.1'
+  inputs:
+    targetType: 'inline'
+    script: |
+      Invoke-WebRequest 'https://dot.net/v1/dotnet-install.ps1' -OutFile 'dotnet-install.ps1'
+      ./dotnet-install.ps1 -InstallDir 'C:\Program Files\dotnet' -Verbose -Channel 3.1
```
Francisco-Gamino, line 8: I think `-Channel` refers to the release type. The command should be something like this:
`./dotnet-install.ps1 -InstallDir "$env:ProgramFiles/dotnet" -Version '3.1.426' -Channel 'release'`

brettsam: We've used `-Channel 6.0` for several years now in the .net 6.0 installer (right below this change). From the docs (https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-install-script), a bunch of things are supported there:

> -Channel|--channel <CHANNEL>
> 
> Specifies the source channel for the installation. The possible values are:
> 
> STS - the most recent Standard Term Support release
> LTS - the most recent Long Term Support release
> Two-part version in A.B format, representing a specific release (for example, 3.1 or 6.0).
> Three-part version in A.B.Cxx format, representing a specific SDK release (for example, 6.0.1xx or 6.0.2xx). Available since the 5.0 release.
> The version parameter overrides the channel parameter when any version other than latest is used.
> 
> The default value is LTS. For more information on .NET support channels, see the [.NET Support Policy](https://dotnet.microsoft.com/platform/support/policy/dotnet-core) page.

Using `-Version` seems to make sense if you want a specific version (with the specific build) -- but in this case, we just want whatever is the latest version; we don't really care about the specific build.

Francisco-Gamino: Make sense. Thank you @brettsam  for the clarification.

PR #9248, opened 2023-05-03
Sending metadata request once per scriptHost instance

soninaren:
Currently we have a race condition inside of `SendFunctionMetadataRequest` where if more than one thread enters the code block before the execution is complete. Then second thread would run into an exception as it would try to complete an already completed task. This PR adds a lock around that code block so we have only one thread enter the code block at a time.

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
@@ -85,6 +85,7 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
         private IHttpProxyService _httpProxyService;
         private Uri _httpProxyEndpoint;
         private System.Timers.Timer _timer;
+        private bool _functionMetadataRequestSent = false;
 
         internal GrpcWorkerChannel(
             string workerId,
@@ -540,6 +541,8 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
         public Task SendFunctionEnvironmentReloadRequest()
         {
             _functionsIndexingTask = new TaskCompletionSource<List<RawFunctionMetadata>>(TaskCreationOptions.RunContinuationsAsynchronously);
+            _functionMetadataRequestSent = false;
+
             _workerChannelLogger.LogDebug("Sending FunctionEnvironmentReloadRequest to WorkerProcess with Pid: '{0}'", _rpcWorkerProcess.Id);
             IDisposable latencyEvent = _metricsLogger.LatencyEvent(MetricEventNames.SpecializationEnvironmentReloadRequestResponse);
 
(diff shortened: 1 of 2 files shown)


liliankasem (approved):

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -795,23 +795,26 @@ public Task<List<RawFunctionMetadata>> GetFunctionMetadata()
 
         internal Task<List<RawFunctionMetadata>> SendFunctionMetadataRequest()
         {
-            // reset indexing task when in case we need to send another request
-            _functionsIndexingTask = new TaskCompletionSource<List<RawFunctionMetadata>>(TaskCreationOptions.RunContinuationsAsynchronously);
+            lock (_syncLock)
```
fabiocav: Do we want to reuse the existing lock here? Is this synchronization with the other paths using this required? Have we reviewed the other usages to avoid synchronization/lock issues?

fabiocav: Discussed additional changes offline as the task would be overridden with multiple calls.

soninaren: Have fixed this

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -795,22 +798,27 @@ public Task<List<RawFunctionMetadata>> GetFunctionMetadata()
 
         internal Task<List<RawFunctionMetadata>> SendFunctionMetadataRequest()
         {
-            // reset indexing task when in case we need to send another request
-            _functionsIndexingTask = new TaskCompletionSource<List<RawFunctionMetadata>>(TaskCreationOptions.RunContinuationsAsynchronously);
+            _workerChannelLogger.LogDebug("Fetching worker metadata, FunctionMetadataReceived set to: {functionMetadataReceived}", _functionMetadataRequestSent);
 
-            RegisterCallbackForNextGrpcMessage(MsgType.FunctionMetadataResponse, _functionLoadTimeout, 1,
-                msg => ProcessFunctionMetadataResponses(msg.Message.FunctionMetadataResponse), HandleWorkerMetadataRequestError);
+            if (!_functionMetadataRequestSent)
```
kshyju, line 803: Curious to know what use case might be causing 2 threads to invoke `SendFunctionMetadataRequest` method simultaneously.

kshyju: Also, this approach here is not really thread safe since the boolean variable is declared outside of the method scope. If 2 threads are calling this method simultaneously, it is possible that the code between 805 and 817 could be executed more than once (Imagine thread 1 called this method & started executing the code (805-817), but has not set the boolean value to `true` yet and thread 2 called this method).

This is why I am curious to know what scenarios is causing this issue.

soninaren: There isn't a scenario where we should see 2 threads executing this at once. 
There was a bug in the hot reload scenario where we would reuse the same language worker between script host restarts.  If a user makes changes to function files in quick successions we would have multiple host restarts. That is where we would see multiple threads calling this at the same time as we were not shutting down the channel. 

The change in the PR below fixes it. However, i am adding this in case we have an unknown scenario calling this multiple times, since we saw customers, app running into that scenario.
https://github.com/Azure/azure-functions-host/pull/9244/files

soninaren: @kshyju have updated the code to be thread safe.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -795,22 +799,29 @@ public Task<List<RawFunctionMetadata>> GetFunctionMetadata()
 
         internal Task<List<RawFunctionMetadata>> SendFunctionMetadataRequest()
         {
-            // reset indexing task when in case we need to send another request
-            _functionsIndexingTask = new TaskCompletionSource<List<RawFunctionMetadata>>(TaskCreationOptions.RunContinuationsAsynchronously);
+            _workerChannelLogger.LogDebug("Fetching worker metadata, FunctionMetadataReceived set to: {functionMetadataReceived}", _functionMetadataRequestSent);
+            if (!_functionMetadataRequestSent)
+            {
+                lock (_metadataLock)
+                {
```
brettsam, line 806: You need to re-check `if (!_functionMetadataRequestSent)` here inside the lock as well. It's possible that there's a race between here and the last check. So you typically (always?) need to check once outside the lock as a potential short-circuit, then once again inside where you know you're the only thread.

brettsam (commented):
I'd also like to see a test here. Ideally one that used multiple threads to call this method more than once while the request was still waiting. We should see it fail w/ the race. I suspect what would happen is that some of the returned Tasks would never complete as they were overwritten by subsequent calls.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
soninaren: Have added the additional boolean check.

brettsam (approved):

PR #9251, opened 2023-05-03
Console logging optimizations

### LINKED ISSUE #9228: Investigate optimzations re: Console.WriteLine in Linux Consumption
I've been investigating runtime performance when running in Linux Consumption and it looks like there is an opportunity to improve the logging pipeline for our internal telemetry. All system logs in this configuration will eventually go through this code path:

https://github.com/Azure/azure-functions-host/blob/76591f722931f651caddeb0f6080d934545246dd/src/WebJobs.Script.WebHost/Diagnostics/LinuxContainerEventGenerator.cs#L99-L105

I quantified the cost of this path by doing tests with/without console logging (using CONSOLE_LOGGING_DISABLED to turn off).

With logging:
```
Statistics        Avg      Stdev        Max
  Reqs/sec       118.23      46.22     500.97
  Latency      169.14ms    28.52ms   364.83ms
  Latency Distribution
     50%   167.89ms
     75%   186.89ms
     90%   205.85ms
     95%   220.14ms
     99%   252.13ms
  HTTP codes:
    1xx - 0, 2xx - 35475, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:    38.32KB/s
```

Without logging:
```
Statistics        Avg     Stdev        Max
 Reqs/sec      280.71     168.92    1154.63
 Latency      73.83ms   217.55ms     17.59s
 Latency Distribution
    50%   69.46ms
    75%   81.84ms
    90%   95.47ms
    95%  106.92ms
    99%  137.67ms
 HTTP codes:
   1xx - 0, 2xx - 50506, 3xx - 0, 4xx - 0, 5xx - 0
   others - 0

 Throughput:   83.95KB/s
```

Now this is a hello world test - the function doesn't do anything other than return a hardcoded response, which is the worst possible case in terms of logging overhead. Real world applications are not seeing the same impact as above, but it shows there is room for improvement.

I am testing a branch with some optimizations, will share a PR soon.

paulbatum:
Resolves #9228

For Linux Consumption, system logs go through standard out. Analysis shows that there is a significant cost associated with this code path, and the cost can be reduced by:
1. Moving the Console.WriteLine call to a background thread
2. Reducing the total number of Console.WriteLine calls by appending log lines together first.

This PR implements both of these improvements, but they can be controlled with the following environment variables:
- CONSOLE_LOGGING_BUFFER_SIZE - this controls the size of the log buffer (how many log lines), defaults to 8000, set to 0 to disable, ~set to -1 for unbounded buffer size.~
~- CONSOLE_LOGGING_BUFFER_BATCHED - when enabled, uses a string builder to write many log lines to stdout with a single Console.WriteLine call, defaults to enabled, set to zero to disable the batching.~

Naming for these variables was based on the existing CONSOLE_LOGGING_DISABLED setting for consistency.

## Performance Test Results

All tests run with Python 3.9 function app running hello world,  median result from 3 runs.

### ACI (single instance, private network)

Baseline:
```
Bombarding http://py39-aci-noai-consolebuffer-baseline.app:80/api/helloworld for 1m0s using 30 connection(s)
Statistics        Avg      Stdev        Max
  Reqs/sec        94.88      63.33     949.95
  Latency      315.32ms    46.99ms   611.99ms
  Latency Distribution
     50%   315.00ms
     75%   343.34ms
     90%   371.79ms
     95%   389.36ms
     99%   450.89ms
  HTTP codes:
    1xx - 0, 2xx - 5715, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:    27.19KB/s
```

Buffer 1000, batched:
```
Bombarding http://py39-aci-noai-consolebuffer-buffer1000-batched.app:80/api/helloworld for 1m0s using 30 connection(s)
Statistics        Avg      Stdev        Max
  Reqs/sec       233.86     180.42    1526.60
  Latency      128.20ms    19.10ms   269.31ms
  Latency Distribution
     50%   129.00ms
     75%   143.78ms
     90%   157.98ms
     95%   167.23ms
     99%   188.00ms
  HTTP codes:
    1xx - 0, 2xx - 14050, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:    69.20KB/s
```

### Legion (single instance, private network)

Baseline:
```
Bombarding http://py39-noai-mesh-perfcollect-consolebuffer-baseline.app:80/api/helloworld for 1m0s using 30 connection(s)
Statistics        Avg      Stdev        Max
  Reqs/sec       342.15     233.37    1578.96
  Latency       88.11ms    64.55ms      0.98s
  Latency Distribution
     50%    73.80ms
     75%    83.00ms
     90%    93.00ms
     95%   102.52ms
     99%   693.74ms
  HTTP codes:
    1xx - 0, 2xx - 20538, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:   101.72KB/s
```
Buffer 1000, batched:
```
Bombarding http://py39-noai-mesh-perfcollect-consolebuffer-buffer1000-batched.app:80/api/helloworld for 1m0s using 30 connection(s)
Statistics        Avg      Stdev        Max
  Reqs/sec       393.31     248.12    1650.02
  Latency       76.23ms    33.30ms      0.86s
  Latency Distribution
     50%    71.22ms
     75%    81.00ms
     90%    90.00ms
     95%    97.29ms
     99%   233.16ms
  HTTP codes:
    1xx - 0, 2xx - 23628, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:   121.39KB/s
```

### Linux Consumption, EUAP (burst scale to 30 instances, tested over public internet)

1000 concurrent clients, burst scale test
```
Baseline: 1,183 RPS (average over 5 minutes)
Buffer 1000, batched: 1,603 RPS (average over 5 minutes)
```

### Test results summary
All runs with this PR had better results, though there is significant variability. I am not 100% sure about why my EUAP tests did not see the same gains as the single instance tests I ran on ACI, but I suspect that network travel time is a factor here, because the EUAP tests were conducted over the public internet. We'll wait to see whether we get similar numbers as part of the official release. Console.WriteLine performance on Legion is not a major issue, so the gains are smaller there, but still worthwhile.

## Functional testing checklist
* [x] Test in Linux Consumption EUAP and confirm that system logs appear in kusto as expected, with expected volume (i.e. requests completed during the test matches number of function completion events
* [x] Test in Legion and confirm system logs appear in kusto as expected

## Pull request checklist

### DIFF
--- a/src/WebJobs.Script.WebHost/Diagnostics/BufferedConsoleWriter.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/BufferedConsoleWriter.cs
@@ -9,13 +9,8 @@ using System.Threading.Tasks;
 
 namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 {
-    internal class ConsoleWriter : IDisposable
+    internal sealed class BufferedConsoleWriter : IDisposable
     {
-        // A typical out-of-proc function execution will generate 8 log lines.
-        // A single core container can potentially get around 1K RPS at the higher end, and a typical log line is around 300 bytes
-        // So in the extreme case, this is about 1 second of buffer and should be less than 3MB
-        private const int DefaultBufferSize = 8000;
-
         // Because we read the log lines in batches from the buffer and write them to the console in one go,
         // we can influence the latency distribution by controlling how much of the buffer we will process in one pass.
         // If we set this to 1, the P50 latency will be low, but the P99 latency will be high.
@@ -34,46 +29,22 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
         private Action<string> _writeEvent;
         private bool _disposed;
 
-        public ConsoleWriter(IEnvironment environment, Action<Exception> unhandledExceptionHandler)
-            : this(environment, unhandledExceptionHandler, consoleBufferTimeout: DefaultConsoleBufferTimeout, autoStart: true)
+        public BufferedConsoleWriter(int bufferSize, Action<Exception> exceptionHandler)
+            : this(bufferSize, exceptionHandler, consoleBufferTimeout: DefaultConsoleBufferTimeout, autoStart: true)
         {
         }
 
-        internal ConsoleWriter(IEnvironment environment, Action<Exception> exceptionHandler, TimeSpan consoleBufferTimeout, bool autoStart)
+        internal BufferedConsoleWriter(int bufferSize, Action<Exception> exceptionHandler, TimeSpan consoleBufferTimeout, bool autoStart)
         {
-            bool consoleEnabled = environment.GetEnvironmentVariable(EnvironmentSettingNames.ConsoleLoggingDisabled) != "1";
+            _consoleBuffer = Channel.CreateBounded<string>(new BoundedChannelOptions(bufferSize) { SingleReader = true, SingleWriter = false });
+            _writeEvent = WriteToConsoleBuffer;
+            _consoleBufferTimeout = consoleBufferTimeout;
+            _writeResetEvent = new ManualResetEvent(true);
+            _maxLinesPerWrite = bufferSize / SingleWriteBufferDenominator;
 
-            if (consoleEnabled)
+            if (autoStart)
             {
-                int maxBufferSize = environment.GetEnvironmentVariable(EnvironmentSettingNames.ConsoleLoggingBufferSize) switch
-                {
-                    var s when int.TryParse(s, out int i) && i >= 0 => i,
-                    var s when int.TryParse(s, out int i) && i < 0 => throw new ArgumentOutOfRangeException(nameof(EnvironmentSettingNames.ConsoleLoggingBufferSize), "Console buffer size cannot be negative"),
-                    _ => DefaultBufferSize,
-                };
-
-                if (maxBufferSize == 0)
-                {
-                    // buffer size was set to zero - disable it
-                    _writeEvent = Console.WriteLine;
-                }
-                else
-                {
-                    _consoleBuffer = Channel.CreateBounded<string>(new BoundedChannelOptions(maxBufferSize) { SingleReader = true, SingleWriter = false });
-                    _writeEvent = WriteToConsoleBuffer;
-                    _consoleBufferTimeout = consoleBufferTimeout;
-                    _writeResetEvent = new ManualResetEvent(true);
-                    _maxLinesPerWrite = maxBufferSize / SingleWriteBufferDenominator;
-
-                    if (autoStart)
-                    {
-                        StartProcessingBuffer();
-                    }
-                }
-            }
-            else
-            {
-                _writeEvent = (string s) => { };
+                StartProcessingBuffer();
             }
 
             _exceptionhandler = exceptionHandler;
@@ -86,12 +57,14 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 
         private void WriteToConsoleBuffer(string evt)
         {
-            if (_consoleBuffer.Writer.TryWrite(evt) == false)
+            if (!_consoleBuffer.Writer.TryWrite(evt))
             {
                 _writeResetEvent.Reset();
-                if (_writeResetEvent.WaitOne(_consoleBufferTimeout) == false || _consoleBuffer.Writer.TryWrite(evt) == false)
+                bool waitFailed = !_writeResetEvent.WaitOne(_consoleBufferTimeout);
+
+                // If the wait failed, write to the console. Otherwise, try the writing again - if that fails, write to the console.
+                if (waitFailed || !_consoleBuffer.Writer.TryWrite(evt))
                 {
-                    // We have either timed out or the buffer was full again, so just write directly to console
                     Console.WriteLine(evt);
                 }
             }
@@ -157,28 +130,20 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
             }
         }
 
-        protected virtual void Dispose(bool disposing)
+        public void Dispose()
         {
             if (!_disposed)
             {
-                if (disposing)
+                if (_consoleBuffer != null)
                 {
-                    if (_consoleBuffer != null)
-                    {
-                        _consoleBuffer.Writer.TryComplete();
-                        _consoleBufferReadLoop.Wait(DisposeTimeout);
-                    }
-
-                    _writeResetEvent?.Dispose();
+                    _consoleBuffer.Writer.TryComplete();
+                    _consoleBufferReadLoop.Wait(DisposeTimeout);
                 }
 
+                _writeResetEvent?.Dispose();
                 _disposed = true;
             }
-        }
 
-        public void Dispose()
-        {
-            Dispose(disposing: true);
             GC.SuppressFinalize(this);
         }
     }
--- a/src/WebJobs.Script.WebHost/Diagnostics/BufferedConsoleWriter.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/BufferedConsoleWriter.cs
@@ -70,6 +70,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
             }
         }
 
+        // internal only for testing - should only be used from the constructor, not intended for concurrent callers.
         internal void StartProcessingBuffer()
         {
             // intentional no-op if the task is already running
--- a/src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Text;
-using System.Threading;
 using System.Threading.Channels;
 using System.Threading.Tasks;
 
@@ -80,30 +79,8 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 
         private void WriteToConsoleBuffer(string evt)
         {
-            try
+            if (_consoleBuffer.Writer.TryWrite(evt) == false)
             {
-                if (_consoleBuffer.Writer.TryWrite(evt) == false)
-                {
-                    // Buffer is currently full, wait until writing is permitted.
-                    using var source = new CancellationTokenSource(_consoleBufferTimeout);
-                    var writeTask = _consoleBuffer.Writer.WriteAsync(evt, source.Token);
-
-                    // This is the downside of using channels, we are on a sync code path and so we have to block on this task if we want to wait for the buffer to clear.
-                    if (writeTask.IsCompleted)
-                    {
-                        writeTask.GetAwaiter().GetResult();
-                    }
-                    else
-                    {
-                        writeTask.AsTask().GetAwaiter().GetResult();
-                    }
-                }
-            }
-            catch (Exception ex)
-            {
-                // Most likely a task cancellation exception from the timeout expiring, but regardless we use the handler
-                // dump the raw exception and write the event to the console directly
-                _exceptionhandler(ex);
                 Console.WriteLine(evt);
             }
         }
--- a/src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Text;
+using System.Threading;
 using System.Threading.Channels;
 using System.Threading.Tasks;
 
@@ -15,11 +16,20 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
         // So in the extreme case, this is about 1 second of buffer and should be less than 3MB
         private const int DefaultBufferSize = 8000;
 
+        // Because we read the log lines in batches from the buffer and write them to the console in one go,
+        // we can influence the latency distribution by controlling how much of the buffer we will process in one pass.
+        // If we set this to 1, the P50 latency will be low, but the P99 latency will be high.
+        // If we set this to a large value, it keeps the P99 latency under control but the P50 degrades.
+        // In local testing with a console attached, processing 1/10th of the buffer size per iteration yields single digit P50 while keeping P99 under 100ms.
+        private const int SingleWriteBufferDenominator = 10;
+
         private static readonly TimeSpan DisposeTimeout = TimeSpan.FromSeconds(5);
         private static readonly TimeSpan DefaultConsoleBufferTimeout = TimeSpan.FromSeconds(1);
+        private readonly ManualResetEvent _writeResetEvent;
         private readonly Channel<string> _consoleBuffer;
         private readonly TimeSpan _consoleBufferTimeout;
         private readonly Action<Exception> _exceptionhandler;
+        private readonly int _maxLinesPerWrite;
         private Task _consoleBufferReadLoop;
         private Action<string> _writeEvent;
         private bool _disposed;
@@ -35,32 +45,29 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 
             if (consoleEnabled)
             {
-                // We are going to used stdout, but do we write directly or use a buffer?
-                _consoleBuffer = environment.GetEnvironmentVariable(EnvironmentSettingNames.ConsoleLoggingBufferSize) switch
+                int maxBufferSize = environment.GetEnvironmentVariable(EnvironmentSettingNames.ConsoleLoggingBufferSize) switch
                 {
-                    "-1" => Channel.CreateUnbounded<string>(new UnboundedChannelOptions() { SingleReader = true, SingleWriter = false }),           // buffer size of -1 indicates that buffer should be enabled but unbounded
-                    "0" => null,                                                                                                                    // buffer size of 0 indicates that buffer should be disabled
-                    var s when int.TryParse(s, out int i) && i > 0 => Channel.CreateBounded<string>(i),
-                    _ => Channel.CreateBounded<string>(new BoundedChannelOptions(DefaultBufferSize) { SingleReader = true, SingleWriter = false }), // default behavior is to use buffer with default size
+                    var s when int.TryParse(s, out int i) && i >= 0 => i,
+                    var s when int.TryParse(s, out int i) && i < 0 => throw new ArgumentOutOfRangeException(nameof(EnvironmentSettingNames.ConsoleLoggingBufferSize), "Console buffer size cannot be negative"),
+                    _ => DefaultBufferSize,
                 };
 
-                if (_consoleBuffer == null)
+                if (maxBufferSize == 0)
                 {
+                    // buffer size was set to zero - disable it
                     _writeEvent = Console.WriteLine;
                 }
                 else
                 {
+                    _consoleBuffer = Channel.CreateBounded<string>(new BoundedChannelOptions(maxBufferSize) { SingleReader = true, SingleWriter = false });
                     _writeEvent = WriteToConsoleBuffer;
                     _consoleBufferTimeout = consoleBufferTimeout;
+                    _writeResetEvent = new ManualResetEvent(true);
+                    _maxLinesPerWrite = maxBufferSize / SingleWriteBufferDenominator;
+
                     if (autoStart)
                     {
-                        bool batched = environment.GetEnvironmentVariable(EnvironmentSettingNames.ConsoleLoggingBufferBatched) switch
-                        {
-                            "0" => false,     // disable batching by setting to 0
-                            _ => true,        // default behavior is batched
-                        };
-
-                        StartProcessingBuffer(batched);
+                        StartProcessingBuffer();
                     }
                 }
             }
@@ -81,30 +88,60 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
         {
             if (_consoleBuffer.Writer.TryWrite(evt) == false)
             {
-                Console.WriteLine(evt);
+                _writeResetEvent.Reset();
+                if (_writeResetEvent.WaitOne(_consoleBufferTimeout) == false || _consoleBuffer.Writer.TryWrite(evt) == false)
+                {
+                    // We have either timed out or the buffer was full again, so just write directly to console
+                    Console.WriteLine(evt);
+                }
             }
         }
 
-        internal void StartProcessingBuffer(bool batched)
+        internal void StartProcessingBuffer()
         {
             // intentional no-op if the task is already running
             if (_consoleBufferReadLoop == null || _consoleBufferReadLoop.IsCompleted)
             {
-                _consoleBufferReadLoop = ProcessConsoleBufferAsync(batched);
+                _consoleBufferReadLoop = ProcessConsoleBufferAsync();
             }
         }
 
-        private async Task ProcessConsoleBufferAsync(bool batched)
+        private async Task ProcessConsoleBufferAsync()
         {
             try
             {
-                if (batched)
+                var builder = new StringBuilder();
+
+                while (await _consoleBuffer.Reader.WaitToReadAsync())
                 {
-                    await ProcessConsoleBufferBatchedAsync();
-                }
-                else
-                {
-                    await ProcessConsoleBufferNonBatchedAsync();
+                    if (_consoleBuffer.Reader.TryRead(out string line1))
+                    {
+                        _writeResetEvent.Set();
+
+                        // Can we synchronously read multiple lines?
+                        // If yes, use the string builder to batch them together into a single write
+                        // If no, just write the single line without using the builder;
+                        if (_consoleBuffer.Reader.TryRead(out string line2))
+                        {
+                            builder.AppendLine(line1);
+                            builder.AppendLine(line2);
+                            int lines = 2;
+
+                            while (lines < _maxLinesPerWrite && _consoleBuffer.Reader.TryRead(out string nextLine))
+                            {
+                                builder.AppendLine(nextLine);
+                                lines++;
+                            }
+
+                            _writeResetEvent.Set();
+                            Console.Write(builder.ToString());
+                            builder.Clear();
+                        }
+                        else
+                        {
+                            Console.WriteLine(line1);
+                        }
+                    }
                 }
             }
             catch (Exception ex)
@@ -120,48 +157,6 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
             }
         }
 
-        private async Task ProcessConsoleBufferNonBatchedAsync()
-        {
-            await foreach (var line in _consoleBuffer.Reader.ReadAllAsync())
-            {
-                Console.WriteLine(line);
-            }
-        }
-
-        private async Task ProcessConsoleBufferBatchedAsync()
-        {
-            var builder = new StringBuilder();
-
-            while (await _consoleBuffer.Reader.WaitToReadAsync())
-            {
-                if (_consoleBuffer.Reader.TryRead(out string line1))
-                {
-                    // Can we synchronously read multiple lines?
-                    // If yes, use the string builder to batch them together into a single write
-                    // If no, just write the single line without using the builder;
-                    if (_consoleBuffer.Reader.TryRead(out string line2))
-                    {
-                        builder.AppendLine(line1);
-                        builder.AppendLine(line2);
-                        int lines = 2;
-
-                        while (lines < DefaultBufferSize && _consoleBuffer.Reader.TryRead(out string nextLine))
-                        {
-                            builder.AppendLine(nextLine);
-                            lines++;
-                        }
-
-                        Console.Write(builder.ToString());
-                        builder.Clear();
-                    }
-                    else
-                    {
-                        Console.WriteLine(line1);
-                    }
-                }
-            }
-        }
-
         protected virtual void Dispose(bool disposing)
         {
             if (!_disposed)
@@ -173,6 +168,8 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
                         _consoleBuffer.Writer.TryComplete();
                         _consoleBufferReadLoop.Wait(DisposeTimeout);
                     }
+
+                    _writeResetEvent?.Dispose();
                 }
 
                 _disposed = true;
--- a/src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
@@ -11,8 +11,8 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
     internal class ConsoleWriter
     {
         // A typical out-of-proc function execution will generate 8 log lines.
-        // A single core container can potentially get around 1K RPS at the higher end, so this is about 1 second of buffer in the extreme case.
-        // A typical log line is around 300 bytes, so this is about 2.4MB of buffer in the extreme case.
+        // A single core container can potentially get around 1K RPS at the higher end, and a typical log line is around 300 bytes
+        // So in the extreme case, this is about 1 second of buffer and should be less than 3MB
         private const int DefaultBufferSize = 8000;
 
         private static readonly TimeSpan DefaultConsoleBufferTimeout = TimeSpan.FromSeconds(1);
--- a/src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
@@ -20,7 +20,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
         // we can influence the latency distribution by controlling how much of the buffer we will process in one pass.
         // If we set this to 1, the P50 latency will be low, but the P99 latency will be high.
         // If we set this to a large value, it keeps the P99 latency under control but the P50 degrades.
-        // In local testing with a console attached, processing 1/10th of the buffer size per iteration yields single digit P50 while keeping P99 under 100ms.
+        // In local testing with a console attached, processing 1/10th of the buffer size per iteration yields P50 under 10ms with P99 under 100ms.
         private const int SingleWriteBufferDenominator = 10;
 
         private static readonly TimeSpan DisposeTimeout = TimeSpan.FromSeconds(5);
--- a/src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
@@ -3,24 +3,27 @@
 
 using System;
 using System.Text;
+using System.Threading;
 using System.Threading.Channels;
 using System.Threading.Tasks;
 
 namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 {
-    internal class ConsoleWriter
+    internal class ConsoleWriter : IDisposable
     {
         // A typical out-of-proc function execution will generate 8 log lines.
         // A single core container can potentially get around 1K RPS at the higher end, and a typical log line is around 300 bytes
         // So in the extreme case, this is about 1 second of buffer and should be less than 3MB
         private const int DefaultBufferSize = 8000;
 
+        private static readonly TimeSpan DisposeTimeout = TimeSpan.FromSeconds(5);
         private static readonly TimeSpan DefaultConsoleBufferTimeout = TimeSpan.FromSeconds(1);
         private readonly Channel<string> _consoleBuffer;
         private readonly TimeSpan _consoleBufferTimeout;
         private readonly Action<Exception> _exceptionhandler;
         private Task _consoleBufferReadLoop;
         private Action<string> _writeEvent;
+        private bool _disposed;
 
         public ConsoleWriter(IEnvironment environment, Action<Exception> unhandledExceptionHandler)
             : this(environment, unhandledExceptionHandler, consoleBufferTimeout: DefaultConsoleBufferTimeout, autoStart: true)
@@ -36,10 +39,10 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
                 // We are going to used stdout, but do we write directly or use a buffer?
                 _consoleBuffer = environment.GetEnvironmentVariable(EnvironmentSettingNames.ConsoleLoggingBufferSize) switch
                 {
-                    "-1" => Channel.CreateUnbounded<string>(),              // buffer size of -1 indicates that buffer should be enabled but unbounded
-                    "0" => null,                                            // buffer size of 0 indicates that buffer should be disabled
+                    "-1" => Channel.CreateUnbounded<string>(new UnboundedChannelOptions() { SingleReader = true, SingleWriter = false }),           // buffer size of -1 indicates that buffer should be enabled but unbounded
+                    "0" => null,                                                                                                                    // buffer size of 0 indicates that buffer should be disabled
                     var s when int.TryParse(s, out int i) && i > 0 => Channel.CreateBounded<string>(i),
-                    _ => Channel.CreateBounded<string>(DefaultBufferSize),  // default behavior is to use buffer with default size
+                    _ => Channel.CreateBounded<string>(new BoundedChannelOptions(DefaultBufferSize) { SingleReader = true, SingleWriter = false }), // default behavior is to use buffer with default size
                 };
 
                 if (_consoleBuffer == null)
@@ -79,17 +82,20 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
         {
             try
             {
-                while (_consoleBuffer.Writer.TryWrite(evt) == false)
+                if (_consoleBuffer.Writer.TryWrite(evt) == false)
                 {
                     // Buffer is currently full, wait until writing is permitted.
-                    // This is the downside of using channels, we are on a sync code path and so we have to block on this task
-                    var writeTask = _consoleBuffer.Writer.WaitToWriteAsync().AsTask();
-                    if (writeTask.WaitAsync(_consoleBufferTimeout).Result == false)
+                    using var source = new CancellationTokenSource(_consoleBufferTimeout);
+                    var writeTask = _consoleBuffer.Writer.WriteAsync(evt, source.Token);
+
+                    // This is the downside of using channels, we are on a sync code path and so we have to block on this task if we want to wait for the buffer to clear.
+                    if (writeTask.IsCompleted)
                     {
-                        // The buffer has been completed and does not allow further writes.
-                        // Currently this should not be possible, but the safest thing to do is just write directly to the console.
-                        Console.WriteLine(evt);
-                        break;
+                        writeTask.GetAwaiter().GetResult();
+                    }
+                    else
+                    {
+                        writeTask.AsTask().GetAwaiter().GetResult();
                     }
                 }
             }
@@ -133,7 +139,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
             {
                 // if this has failed for any reason, fall everything back to console
                 _writeEvent = Console.WriteLine;
-                _consoleBuffer.Writer.Complete();
+                _consoleBuffer.Writer.TryComplete();
             }
         }
 
@@ -149,15 +155,8 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
         {
             var builder = new StringBuilder();
 
-            while (true)
+            while (await _consoleBuffer.Reader.WaitToReadAsync())
             {
-                if (await _consoleBuffer.Reader.WaitToReadAsync() == false)
-                {
-                    // The buffer has been completed and does not allow further reads.
-                    // Currently this should not be possible, but safest thing to do is break out of the loop.
-                    break;
-                }
-
                 if (_consoleBuffer.Reader.TryRead(out string line1))
                 {
                     // Can we synchronously read multiple lines?
@@ -169,7 +168,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
                         builder.AppendLine(line2);
                         int lines = 2;
 
-                        while (_consoleBuffer.Reader.TryRead(out string nextLine) && lines < DefaultBufferSize)
+                        while (lines < DefaultBufferSize && _consoleBuffer.Reader.TryRead(out string nextLine))
                         {
                             builder.AppendLine(nextLine);
                             lines++;
@@ -184,6 +183,29 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
                     }
                 }
             }
+        }
+
+        protected virtual void Dispose(bool disposing)
+        {
+            if (!_disposed)
+            {
+                if (disposing)
+                {
+                    if (_consoleBuffer != null)
+                    {
+                        _consoleBuffer.Writer.TryComplete();
+                        _consoleBufferReadLoop.Wait(DisposeTimeout);
+                    }
+                }
+
+                _disposed = true;
+            }
+        }
+
+        public void Dispose()
+        {
+            Dispose(disposing: true);
+            GC.SuppressFinalize(this);
         }
     }
 }
--- a/src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
@@ -15,8 +15,8 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
         private readonly Channel<string> _consoleBuffer;
         private readonly TimeSpan _consoleBufferTimeout;
         private readonly Task _consoleBufferReadLoop;
-        private readonly Action<string> _writeEvent;
         private readonly Action<Exception> _exceptionhandler;
+        private Action<string> _writeEvent;
 
         public ConsoleWriter(IEnvironment environment, Action<Exception> unhandledExceptionHandler)
             : this(environment, unhandledExceptionHandler, consoleBufferTimeout: DefaultConsoleBufferTimeout, autoStart: true)
@@ -48,11 +48,13 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
                     _consoleBufferTimeout = consoleBufferTimeout;
                     if (autoStart)
                     {
-                        _consoleBufferReadLoop = environment.GetEnvironmentVariable(EnvironmentSettingNames.ConsoleLoggingBufferBatched) switch
+                        bool batched = environment.GetEnvironmentVariable(EnvironmentSettingNames.ConsoleLoggingBufferBatched) switch
                         {
-                            "0" => ProcessConsoleBufferAsync(),             // disable batching by setting to 0
-                            _ => ProcessConsoleBufferBatchedAsync(),        // default behavior is batched
+                            "0" => false,     // disable batching by setting to 0
+                            _ => true,        // default behavior is batched
                         };
+
+                        _consoleBufferReadLoop = ProcessConsoleBufferAsync(batched);
                     }
                 }
             }
@@ -64,7 +66,10 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
             _exceptionhandler = exceptionHandler;
         }
 
-        public Action<string> WriteHandler => _writeEvent;
+        public void WriteHandler(string evt)
+        {
+            _writeEvent(evt);
+        }
 
         private void WriteToConsoleBuffer(string evt)
         {
@@ -93,13 +98,17 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
             }
         }
 
-        internal async Task ProcessConsoleBufferAsync()
+        internal async Task ProcessConsoleBufferAsync(bool batched)
         {
             try
             {
-                await foreach (var line in _consoleBuffer.Reader.ReadAllAsync())
+                if (batched)
                 {
-                    Console.WriteLine(line);
+                    await ProcessConsoleBufferBatchedAsync();
+                }
+                else
+                {
+                    await ProcessConsoleBufferNonBatchedAsync();
                 }
             }
             catch (Exception ex)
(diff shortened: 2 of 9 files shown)


src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
```diff
@@ -0,0 +1,169 @@
...
+                    {
+                        // Can we synchronously read multiple lines?
+                        // If yes, use the string builder to batch them together into a single write
+                        // If no, just write the single line without using the builder;
+                        if (_consoleBuffer.Reader.TryRead(out string line2))
+                        {
+                            builder.AppendLine(line1);
+                            builder.AppendLine(line2);
+
+                            while (_consoleBuffer.Reader.TryRead(out string nextLine))
```
NickCraver: Should this have a cap per-outer-loop? The build could get very, very large under load if we get here.

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
```diff
@@ -0,0 +1,169 @@
...
+            }
+            catch (Exception ex)
+            {
+                // Not sure what can fail here, but there is nothing else monitoring this task so just log the exception
+                _exceptionhandler(ex);
+            }
+            finally
+            {
+                // if this has failed for any reason, complete the buffer so that everything falls back to console
+                _consoleBuffer.Writer.Complete();
```
NickCraver: I agree on these fallbacks ultimately using Console.Writeline...but if I understand correctly it'll be via a try/fail/retry pattern on getting there. We probably want to replace the delegate with a direct Console write in this failure case (common method for both?) so that the fallback isn't a net worse over current.

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
paulbatum: Makes sense, I added a cap of 200 log lines. Wasn't really sure what value is appropriate here, so feel free to suggest a different value.

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
paulbatum: Makes sense. I updated the implementation to have an extra level of indirection so that the delegate can be replaced on the fly without the consumer (LinuxContainerEventGenerator) knowing, and the finally block now does the replacement.

NickCraver: +1 - Minor: you'll want to swap the delegate swap and the .Complete() to reduce the race (ones trying to enter the pipe - we want to redirect new traffic before closing the door to optimize the failure path)

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
```diff
@@ -0,0 +1,177 @@
...
+using System;
+using System.Text;
+using System.Threading.Channels;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
+{
+    internal class ConsoleWriter
+    {
+        private const int DefaultBufferSize = 1000;
```
NickCraver: How many function calls is this? e.g. do we have 1 line per call, 5, 10? I'd say you'd want to aim for something like 1-10,000 * per-request in the count here (with comment indicating it). 1,000 just isn't a lot under load and I'm not sure this would behave near as well as it could in multi-core scenarios going so low - thoughts?

paulbatum: Good call! We generate 8 lines for a typical out of proc execution. I tested buffer sizes of 1K, 8K and 80K lines. The P99 latency definitely improves jumping from 1K to 8K but then it plateaus. There is some sign of RPS improvement when going from 8K to 80K but I am worried about the memory usage impact, especially in the context of containers that are allocated smaller amounts of memory. We could eventually optimize further by specifying different buffer sizes for different container sizes but I don't want to get into that right now. I'll send an update bumping this to 8K.

All results below are the median of 3 runs.

1K buffer, batched:
```
Bombarding http://py39-noai-mesh-perfcollect-consolebuffer-buffer1000-batched.app:80/api/helloworld for 1m0s using 30 connection(s)
Statistics        Avg      Stdev        Max
  Reqs/sec       393.59     254.27    1631.41
  Latency       76.23ms    32.45ms   784.71ms
  Latency Distribution
     50%    70.00ms
     75%    79.00ms
     90%    89.00ms
     95%    97.00ms
     99%   336.62ms
  HTTP codes:
    1xx - 0, 2xx - 23625, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:   121.39KB/s
```
8K buffer, batched:
```
Bombarding http://py39-noai-mesh-perfcollect-consolebuffer-buffer8000-batched.app:80/api/helloworld for 1m0s using 30 connection(s)
Statistics        Avg      Stdev        Max
  Reqs/sec       413.62     228.98    1578.89
  Latency       72.51ms     7.91ms   309.00ms
  Latency Distribution
     50%    72.00ms
     75%    80.73ms
     90%    89.00ms
     95%    94.94ms
     99%   110.59ms
  HTTP codes:
    1xx - 0, 2xx - 24837, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:   127.61KB/s
```

80K buffer, batched:
```
Bombarding http://py39-noai-mesh-perfcollect-consolebuffer-buffer80000-batched.app:80/api/helloworld for 1m0s using 30 connection(s)
Statistics        Avg      Stdev        Max
  Reqs/sec       427.36     250.94    1799.71
  Latency       70.21ms     8.27ms   186.00ms
  Latency Distribution
     50%    69.99ms
     75%    78.35ms
     90%    87.99ms
     95%    94.69ms
     99%   111.02ms
  HTTP codes:
    1xx - 0, 2xx - 25640, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:   132.23KB/s
```

This result makes sense to me, given that the current implementation will block for up to 1 second waiting on the buffer to become available before falling back to the console. In the 1K case, the P99 latency really looks like its impacted by waits on the buffer to be cleared.

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
```diff
@@ -0,0 +1,177 @@
...
+            catch (Exception ex)
+            {
+                // Most likely a task cancellation exception from the timeout expiring, but regardless we use the handler
+                // dump the raw exception and write the event to the console directly
+                _exceptionhandler(ex);
+                Console.WriteLine(evt);
+            }
+        }
+
+        internal async Task ProcessConsoleBufferAsync(bool batched)
```
mathewc: This is conceptually a "start" method, intended to be called once for the lifetime of the writer. I get that it's internal only for a single test and platform usages will use the public ctor w/autostart, but I wonder if you should prevent multiple calls to it to start up multiple consumer loops? Perhaps the Task shouldn't be returned, and instead it sets _consoleBufferReadLoop, and if already set it noops or errors?

Might also call this Start or StartProcessing to match your autoStart naming.

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
```diff
@@ -0,0 +1,177 @@
...
+                        // Currently this should not be possible, but the safest thing to do is just write directly to the console.
+                        Console.WriteLine(evt);
+                        break;
+                    }
+                }
+            }
+            catch (Exception ex)
+            {
+                // Most likely a task cancellation exception from the timeout expiring, but regardless we use the handler
+                // dump the raw exception and write the event to the console directly
```
mathewc: Potentially under load, we might get into this timeout situation often. Won't this mean that logs will be out of order with out of sync timestamps? E.g. the forced Console write jumps the queue and is logged before other pending logs?

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
```diff
@@ -0,0 +1,177 @@
...
+        private void WriteToConsoleBuffer(string evt)
+        {
+            try
+            {
+                while (_consoleBuffer.Writer.TryWrite(evt) == false)
+                {
+                    // Buffer is currently full, wait until writing is permitted.
+                    // This is the downside of using channels, we are on a sync code path and so we have to block on this task
+                    var writeTask = _consoleBuffer.Writer.WaitToWriteAsync().AsTask();
+                    if (writeTask.WaitAsync(_consoleBufferTimeout).Result == false)
```
mathewc: This sync over async (.Result) concerns me. For Linux Dedicated file logger, we also have buffering implemented, but it's using BlockingCollection (which also supports specifying a bound) (code [here](https://github.com/Azure/azure-functions-host/blob/e9637fc46830ccbea0107ae8adb89eec50dd2b57/src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceFileLogger.cs#L20)). Perhaps we could do the same thing here to avoid the async issue?

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
paulbatum: Yes, this implementation can result in out of order logs. The alternative is to fully block the writes. Do you have a strong preference?

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
paulbatum: See the results I posted as a comment on the PR for testing with BlockingCollection. The problem is that I don't want to allow an unbounded buffer size, and BlockingCollection performs poorly once a bound is set.

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
mathewc: Yeah, I'd say we should avoid out of order logs. If we're putting buffer in place for output I think we should honor the ordering, and block until we can write.

paulbatum: As we discussed yesterday, it seems better to have out of order logs than to be blocking the execution code path on waiting for the buffer to have space, which is reliant on the background task that is processing the buffer. This only occurs when the buffer is full and when we've waited for the buffer to be cleared (currently 1 second). We have the EventTimestamp column that can be used in cases where ordering is critical.

paulbatum:
I did some more testing based on the conversation about the fact that we occasionally have to block on a Task.Result due to the choice of using channels. I coded up some alternative versions:
1. BlockingCollection - similar what we do for linux dedicated, but with a bounded size.
2. Lock + Queue + Monitor for synchronization. This is based on the .NET console logging implementation: https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Logging.Console/src/ConsoleLoggerProcessor.cs

Neither of these compared favorably to the current implementation in this PR - all results below are the median of 3 runs. These tests were performed in ACI.

Baseline:
```
Bombarding http://py39-aci-noai-consolebuffer-baseline.app:80/api/helloworld for 1m0s using 30 connection(s)
Statistics        Avg      Stdev        Max
  Reqs/sec       103.53      74.82     847.87
  Latency      289.34ms    41.55ms   490.04ms
  Latency Distribution
     50%   289.47ms
     75%   316.62ms
     90%   341.94ms
     95%   359.11ms
     99%   397.00ms
  HTTP codes:
    1xx - 0, 2xx - 6236, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:    29.63KB/s
```

Blocking Collection, buffer size 8000:
```
Bombarding http://py39-noai-mesh-perfcollect-consolebuffer-blocking8000.app:80/api/helloworld for 1m0s using 30 connection(s)
Statistics        Avg      Stdev        Max
  Reqs/sec        18.06     131.03    1499.98
  Latency         1.67s      2.36s     10.03s
  Latency Distribution
     50%    31.00ms
     75%    51.00ms
     90%     10.02s
     95%     10.02s
     99%     10.03s
  HTTP codes:
    1xx - 0, 2xx - 921, 3xx - 0, 4xx - 0, 5xx - 0
    others - 180
  Errors:
    the server closed connection before returning the first response byte. Make sure the server returns 'Connection: close' response header before closing the connection - 180
  Throughput:     6.31KB/s
```

Lock + Queue + Monitor, buffer size 8000:
```
Bombarding http://py39-aci-noai-consolebuffer-monitor8000.app:80/api/helloworld for 1m0s using 30 connection(s)
Statistics        Avg      Stdev        Max
  Reqs/sec       118.50     106.08    1525.66
  Latency      252.98ms    44.96ms   478.40ms
  Latency Distribution
     50%   253.10ms
     75%   279.24ms
     90%   308.81ms
     95%   328.65ms
     99%   377.51ms
  HTTP codes:
    1xx - 0, 2xx - 7125, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:    34.25KB/s
```

Channel with buffer size of 8000, batched (current PR):
```
Bombarding http://py39-aci-noai-consolebuffer-channel8000-batched.app:80/api/helloworld for 1m0s using 30 connection(s)
Statistics        Avg      Stdev        Max
  Reqs/sec       146.76     120.72    1299.97
  Latency      204.24ms    34.40ms   365.00ms
  Latency Distribution
     50%   205.38ms
     75%   230.00ms
     90%   252.93ms
     95%   268.00ms
     99%   298.22ms
  HTTP codes:
    1xx - 0, 2xx - 8821, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:    43.56KB/s
```

The BlockingCollection results are particularly awful, and this matches up with some very inconsistent results I got when testing BlockingCollection outside of the context of the functions code base. Specifically the problems arise when it has a fixed bound that is being reached. Our current Linux Dedicated implementation does not set a bound on the collection, so it doesn't run into this problem.

Its worth noting that I didn't get quite get the same gains that I saw in my first round of testing on ACI. I did mention a similar discrepancy in my first post, the improvement I observed in EUAP was around 35%, while the single instance ACI testing on private network saw larger gains. This time round, my testing on ACI showed a 40% improvement. So this is pretty similar to what I saw in my EUAP results. I dont know why that initial ACI run with the channel implementation was so good.

I think the Channel based implementation is still the best option.

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
paulbatum: I upped this to a much higher value - the limit of how many lines we'll write to the builder before we flush and clear is now the same as the buffer size (currently 8000).

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
paulbatum: Makes sense, PR is updated with these changes.

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
paulbatum: One option that I forgot to mention is that we could immediately write to the console if the buffer is full. This would remove the WaitToWriteAsync call completely, so no blocking on Task.Result. It would increase the potential for out-of-order logs, but that might be an acceptable tradeoff.

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
paulbatum: Makes sense, done.

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
```diff
@@ -0,0 +1,189 @@
...
+        private readonly Action<Exception> _exceptionhandler;
+        private Task _consoleBufferReadLoop;
+        private Action<string> _writeEvent;
+
+        public ConsoleWriter(IEnvironment environment, Action<Exception> unhandledExceptionHandler)
+            : this(environment, unhandledExceptionHandler, consoleBufferTimeout: DefaultConsoleBufferTimeout, autoStart: true)
+        {
+        }
+
+        internal ConsoleWriter(IEnvironment environment, Action<Exception> exceptionHandler, TimeSpan consoleBufferTimeout, bool autoStart)
```
jviau: this class is trying to perform 3 different modes: writer to buffer, write to `System.Console`, or no-op. Instead, can this be split into a contract and different implementations? An interface `IConsoleWriter`, three (or four for bounded vs unbounded channel) different implementations, then a helper method `public static IConsole CreateConsole(...args)` which will create the appropriate one based on supplied arguments?

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
```diff
@@ -0,0 +1,189 @@
...
+
+        private void WriteToConsoleBuffer(string evt)
+        {
+            try
+            {
+                while (_consoleBuffer.Writer.TryWrite(evt) == false)
+                {
+                    // Buffer is currently full, wait until writing is permitted.
+                    // This is the downside of using channels, we are on a sync code path and so we have to block on this task
+                    var writeTask = _consoleBuffer.Writer.WaitToWriteAsync().AsTask();
```
jviau, line 86: If the goal is performance improvement, it may be valuable to remove the allocation that `.AsTask()` would cause.

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
```diff
@@ -0,0 +1,189 @@
...
+                    if (writeTask.WaitAsync(_consoleBufferTimeout).Result == false)
```
jviau: Use `GetAwaiter().GetResult()` - not `.Result`.

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
```diff
@@ -0,0 +1,189 @@
...
+        public void WriteHandler(string evt)
+        {
+            _writeEvent(evt);
+        }
+
+        private void WriteToConsoleBuffer(string evt)
+        {
+            try
+            {
+                while (_consoleBuffer.Writer.TryWrite(evt) == false)
```
jviau, line 82: This doesn't look like it will ever proceed beyond the first iteration?

```suggestion
                if (!_consoleBuffer.Writer.TryWrite(evt))
```

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
```diff
@@ -0,0 +1,189 @@
...
+                // dump the raw exception and write the event to the console directly
+                _exceptionhandler(ex);
+                Console.WriteLine(evt);
+            }
+        }
+
+        internal void StartProcessingBuffer(bool batched)
+        {
+            // intentional no-op if the task is already running
+            if (_consoleBufferReadLoop == null || _consoleBufferReadLoop.IsCompleted)
```
jviau, line 108:
```suggestion
            if (_consoleBufferReadLoop is { IsCompleted: true })
```

jviau: Alternative way via pattern matching:

```suggestion
            if (_consoleBufferReadLoop is { IsCompleted: true })
```

jviau (commented):
Do you have plans to flush out the buffer on process shutdown? If not, we will potentially lose console writes.

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
jviau: This cast is not ideal and may cause an extra allocation in the `IValueTaskSource` case, which would be counterproductive for the performance gains. However, not sure how to address this given the sync requirement. But I do suggest the following:

``` CSharp
try
{
    if (!_consoleBuffer.Writer.TryWrite(ext))
    {
        // WriteAsync will handle WaitToWriteAsync() -> TryWrite() loop for us
        ValueTask write = _consoleBuffer.Writer.WriteAsync(ext);
        if (write.IsCompleted)
        {
            write.GetAwaiter().GetResult(); // rethrow exceptions
        }
        else
        {
            write.AsTask().WaitAsync(_consoleBufferTimeout).GetAwaiter().GetResult();
        }
    }
}
catch (Exception ex)
{
}
```

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
```diff
@@ -0,0 +1,189 @@
...
+            {
+                Console.WriteLine(line);
+            }
+        }
+
+        private async Task ProcessConsoleBufferBatchedAsync()
+        {
+            var builder = new StringBuilder();
+
+            while (true)
```
jviau, line 152: nit: WaitToReadyAsync can be the directly in the outer `while` statement, and then an inner loop can be used to rapidly deque items without performing another wait:

``` CSharp
StringBuilder builder = new();
while (await reader.WaitToReadAsync())
{
    int lines = 0;
    string line = null;
    while (lines < DefaultBufferSize && reader.TryRead(out string item))
    {
        lines++;
        switch (lines)
        {
            case 0:
                line = item;
                break;
            case 1:
                builder.AppendLine(line);
                builder.AppendLine(item);
                break;
            default:
                builder.AppendLine(item);
                break;
        }
    }

    if (lines == 1)
    {
        Console.WriteLine(line);
    }
    else if (lines > 1)
    {
       Console.WriteLine(builder.ToString());
       builder.Clear();
    }
}
```

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
NickCraver: If structuring like this, you'd want to take advantage of `.IsCompletedSuccessfully` and use the `.Result` directly more cheaply, but note: this is the exception case in general where a write exceeded the buffer (by that I mean, it's only likely to actually trigger if feathering the limit, not during actual exhaustion).

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
```diff
@@ -0,0 +1,189 @@
...
+                    // Can we synchronously read multiple lines?
+                    // If yes, use the string builder to batch them together into a single write
+                    // If no, just write the single line without using the builder;
+                    if (_consoleBuffer.Reader.TryRead(out string line2))
+                    {
+                        builder.AppendLine(line1);
+                        builder.AppendLine(line2);
+                        int lines = 2;
+
+                        while (_consoleBuffer.Reader.TryRead(out string nextLine) && lines < DefaultBufferSize)
```
jviau: `lines < DefaultBufferSize` must go first, otherwise you may dequeue (`TryRead` is `true`), but never append it.

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
jviau: nit: but the wait to read directly in the while loop test.

```suggestion
            while (await _consoleBuffer.Reader.WaitToReadAsync())
```

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
jviau: There is no `.Result` on `ValueTask` (only on `ValueTask<T>`). But that is fine - `GetAwait().GetResult()` is 0 allocation. So this call here is only to rethrow exceptions, if any. One exception is the writer has completed before accepting your item.

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
jviau: Along with the contract change, we could put this in the DI container and have it injected into `LinuxContainerEventGenerator`. I would also then recommend we adopt `IOptions` pattern for this. Really going forward we should minimize any usage of `IEnvironment` or `IConfiguration` directly and opt for options patterns (only using `IEnvironment` or `IConfiguration` as part of options setup).

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
paulbatum: I think this might write the log entry twice? Because no cancellation token is passed to WriteAsync, so its still going to try to write even after we gave up waiting. I think to address that I'd need to stop using WaitAsync, and instead create a cancellation token source and use CancelAfter. I guess this is still slightly better than the current approach because it still prevents multiple iterations of waiting and then trying to write, so I'm switching it over now.

paulbatum: Assuming we want to avoid duplicate writes, what is the downside of this approach?

```csharp
            try
            {
                if (_consoleBuffer.Writer.TryWrite(evt) == false)
                {
                    // Buffer is currently full, wait until writing is permitted.
                    var source = new CancellationTokenSource(_consoleBufferTimeout);
                    var writeTask = _consoleBuffer.Writer.WriteAsync(evt, source.Token);

                    // This is the downside of using channels, we are on a sync code path and so we have to block on this task if we want to wait for the buffer to clear.
                    writeTask.GetAwaiter().GetResult();
                }
            }
            catch (Exception ex)
            {
                // Most likely a task cancellation exception from the timeout expiring, but regardless we use the handler
                // dump the raw exception and write the event to the console directly
                _exceptionhandler(ex);
                Console.WriteLine(evt);
            }
```

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
paulbatum: This is an OR, while I think your suggestion is for an AND?

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
paulbatum: great catch, will push a fix.

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
paulbatum: > Do you have plans to flush out the buffer on process shutdown? If not, we will potentially lose console writes.

At first I was skeptical about how necessary this is, given that the linux dedicated implemention already buffers and doesn't do a flush, and we don't have any reports of lost logs. But on second thought, its reasonably cheap to implement by relying on the DI behavior of calling dispose on all registered services that implement IDisposable, so I'll push an update that does this.

paulbatum: > this class is trying to perform 3 different modes: writer to buffer, write to `System.Console`, or no-op. Instead, can this be split into a contract and different implementations? An interface `IConsoleWriter`, three (or four for bounded vs unbounded channel) different implementations, then a helper method `public static IConsole CreateConsole(...args)` which will create the appropriate one based on supplied arguments?

Doesn't it seem kind of over engineered to have 3 different types given that the no op and the direct writing to console are one liners? Its not like each of the methods is having to check various flags to figure out what mode its in. It also makes the fallback implementation (start using console.write directly once buffer is completed) more complicated once these modes are separated into different types.

The bounded vs unbounded behavior are the same so I'm not sure why we would want different classes for those, the only difference is whether we call `CreateBounded` or `CreateUnbounded`.

paulbatum: > Along with the contract change, we could put this in the DI container and have it injected into `LinuxContainerEventGenerator`. I would also then recommend we adopt `IOptions` pattern for this. Really going forward we should minimize any usage of `IEnvironment` or `IConfiguration` directly and opt for options patterns (only using `IEnvironment` or `IConfiguration` as part of options setup).

Regarding registering ConsoleWriter - for types where there is literally just one consumer, I am not sure we want to clutter our DI registrations with them. For example, take a look at `WebHostMetricsLogger` - internally it uses a `MetricsEventManager` but because their lifetimes are tied together, the `MetricsEventManager` is not registered in the DI container and is instead constructed directly as part of the `WebHostMetricsLogger` constructor. I think this is better from an encapsulation perspective - otherwise the codepaths for registring services would continue to explode in size (they are already quite large).

Regarding configuration - I deliberately chose a style that is consistent with the related type - LinuxContainerEventGenerator, but I'll take another look.

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
jviau: I am not 100% sure how `Console` implementation works in C#, but is writing a point of contention? If the buffer is full, and this starts writing directly, would that slow down the ability to even drain the buffer since now more things are fighting for `Console`?

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
jviau: You don't have to separate into different types and use DI if you don't want to. However, I think while we only have 1 consumer right now, doesn't mean we won't have more in the future. I don't think someone in the future will take the time to move to DI when that 2nd consumer is added. Going the DI route also helps us follow the inversion-of-control pattern, and makes unit testing simpler. You could supply a test-implementation of `ConsoleWriter` - or the no-op implementation. You also don't need to think deeply about what values need to be in `IEnvironment`. 

As for the separate types - I like doing that for organization. I normally have the implementations as private nested classes. It helps me validate the behavior of each implementation (even the 1 liners), as I can consider the intended behavior of each implementation on their own.

But again - this is all up to you. I like it for code organization and testability.

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
jviau: Ah yes, I missed that. For an OR you can do `_consoleBufferReadLoop?.IsCompleted ?? true`

Or maybe (not sure if this compiles?)

``` CSharp
if (_consoleBufferReadLoop is { IsCompleted: true } or null)
```

leaving what you have is also completely fine.

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
```diff
@@ -0,0 +1,188 @@
...
+                    }
+                    else
+                    {
+                        Console.WriteLine(line1);
+                    }
+                }
+            }
+        }
+
+        protected virtual void Dispose(bool disposing)
```
jviau: You can implement `IAsyncDisposable` instead if you want access to async.

paulbatum:
I've ran some tests with large load and very small buffer sizes to determine how the buffer-full codepath behaves, and its not good - while there is no deadlock condition, the app shows signs of threadpool starvation, probably due to doing Task.Result on a sync codepath. 

```
Bombarding http://py39-perfcollect-channel100-batched.app:80/api/helloworld for 1m0s using 100 connection(s)
Statistics        Avg      Stdev        Max
  Reqs/sec        26.29     158.67    2818.03
  Latency         3.84s      3.02s     10.04s
  Latency Distribution
     50%   267.03ms
     75%     10.01s
     90%     10.02s
     95%     10.02s
     99%     10.03s
  HTTP codes:
    1xx - 0, 2xx - 1066, 3xx - 0, 4xx - 0, 5xx - 0
    others - 600
  Errors:
    the server closed connection before returning the first response byte. Make sure the server returns 'Connection: close' response header before closing the connection - 600
  Throughput:     9.48KB/s
```

```
[System.Runtime]
    ThreadPool Thread Count                                              233
```

Now with the planned buffer size of 8000 I don't think this condition is likely to be hit, but I consider this a deal breaker for the current implementation. I will do more testing and find an alternative, and I'll re-request reviews once I've pushed an update.

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
jviau: Do you intend for this class to be inherited? If not, you can ignore the dispose(true/false) pattern by adding `sealed` to your class declaration.

src/WebJobs.Script.WebHost/Diagnostics/LinuxContainerEventGenerator.cs
```diff
@@ -6,25 +6,39 @@
 
 namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 {
-    internal class LinuxContainerEventGenerator : LinuxEventGenerator
+    internal class LinuxContainerEventGenerator : LinuxEventGenerator, IDisposable
     {
         private const int MaxDetailsLength = 10000;
-        private static readonly Lazy<LinuxContainerEventGenerator> _Lazy = new Lazy<LinuxContainerEventGenerator>(() => new LinuxContainerEventGenerator(SystemEnvironment.Instance));
+        private static readonly Lazy<LinuxContainerEventGenerator> _Lazy = new Lazy<LinuxContainerEventGenerator>(() => new LinuxContainerEventGenerator(SystemEnvironment.Instance, Console.WriteLine));
```
jviau, line 14: Looking at our code base, I think we have a mix of  `PascalCase` and `_camelCase` for `private static readonly` naming. I don't think we use `_PascalCase` anywhere.

(personally, I like `s_camelCase` for this, but doesn't look like we use that)

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
```diff
@@ -0,0 +1,188 @@
...
+        }
+
+        public void WriteHandler(string evt)
+        {
+            _writeEvent(evt);
+        }
+
+        private void WriteToConsoleBuffer(string evt)
+        {
+            if (_consoleBuffer.Writer.TryWrite(evt) == false)
```
jviau: nit:

```suggestion
            if (!_consoleBuffer.Writer.TryWrite(evt))
```

jviau:
> > Do you have plans to flush out the buffer on process shutdown? If not, we will potentially lose console writes.
> 
> At first I was skeptical about how necessary this is, given that the linux dedicated implemention already buffers and doesn't do a flush, and we don't have any reports of lost logs. But on second thought, its reasonably cheap to implement by relying on the DI behavior of calling dispose on all registered services that implement IDisposable, so I'll push an update that does this.

We don't have any reports of lost logs? In my (limited) experience of diagnosing some function app issues, I have had several instances of what I chalked up to lost logs (but could not prove). These were all around a function app starting up, then being stopped. I was able to diagnose these issues without the logs, but still odd they weren't there. Not saying we do have lost logs, but some logs I expected weren't there.

It is also better safe than sorry - console might not be a bad for this, but something like app insights? I can easily reproduce lost logs locally by not flushing the `TelemetryClient` on shutdown in some test console apps I have written.

jviau:
Slip of the finger trying to hit "comment".

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
paulbatum: Further investigation revealed the tradeoff in this setting - you can shift the latency distribution with this value. For example a high value of 8000 will keep most percentiles low, but push the P99 out quite high:

Max lines - 8000:
```
  Latency Distribution
     50%     5.53ms
     75%     6.53ms
     90%     7.53ms
     95%     9.01ms
     99%   692.64ms
```     
While a lower value of 800 will push out the P90 and P95, but keep the P99 under control.

Max lines - 800:
```
  Latency Distribution
     50%     6.00ms
     75%     9.77ms
     90%    70.05ms
     95%    72.17ms
     99%    75.81ms
```

PR has been updated to prefer the latter.

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
paulbatum: Yes, I think that outcome is theoretically possible, but how it plays out depends a bit on the nature of the situation that causes the buffer to be full in the first place. For example, threadpool starvation can cause delays on the buffer consumer side, in which case Console.Write contention is not the problem.

The PR has been updated with a new synchronization approach that still follows the rule of "try to wait some time to maximize chance of preserving log ordering" but avoids the pitfalls of blocking on a `GetAwaiter().GetResult()`.

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
paulbatum: Good suggestion! The only implementor in the codebase currently is AspNetCoreGrpcServer, and it happens to implement both IDisposable and IAsyncDisposable. Do you know if this is required? The sync implementation is quite odd in that it calls the async implementation without blocking on the valuetask, so as far as I can tell it breaks the expected IDisposable contract (disposing might not be done by the time the method returns).
https://github.com/Azure/azure-functions-host/blob/4ec73a34dea76db91b6861db59d57d1ff00265f8/src/WebJobs.Script.Grpc/Server/AspNetCoreGrpcServer.cs#L60-L82

paulbatum: @jviau Unfortunately, I can't get this to work. I tested by using ctrl+c to shutdown a running host, and the Dispose implementation runs but the DisposeAync implementation does not. Either there is something wrong with my testing methodology, or our DI container doesn't have support for IAsyncDisposable.

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
paulbatum: I think I found a sweet spot - I factored the implementation into `BufferedConsoleWriter` that just takes simple parameters and doesnt know anything about environment, added a ConsoleLoggingOptions type that LinuxContainerEventGenerator depends on, and pushed the logic for handling the environment into a ConsoleLoggingOptionsSetup type.

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
paulbatum: Done

paulbatum:
> > > Do you have plans to flush out the buffer on process shutdown? If not, we will potentially lose console writes.
> > 
> > 
> > At first I was skeptical about how necessary this is, given that the linux dedicated implemention already buffers and doesn't do a flush, and we don't have any reports of lost logs. But on second thought, its reasonably cheap to implement by relying on the DI behavior of calling dispose on all registered services that implement IDisposable, so I'll push an update that does this.
> 
> We don't have any reports of lost logs? In my (limited) experience of diagnosing some function app issues, I have had several instances of what I chalked up to lost logs (but could not prove). These were all around a function app starting up, then being stopped. I was able to diagnose these issues without the logs, but still odd they weren't there. Not saying we do have lost logs, but some logs I expected weren't there.
> 
> It is also better safe than sorry - console might not be a bad for this, but something like app insights? I can easily reproduce lost logs locally by not flushing the `TelemetryClient` on shutdown in some test console apps I have written.

I agree, the implementation now has an explicit dispose, and I verified that the codepath runs when shutting down the runtime.

src/WebJobs.Script.WebHost/Diagnostics/ConsoleWriter.cs
paulbatum: Sticking with IDispose implementation for now.

paulbatum:
> I've ran some tests with large load and very small buffer sizes to determine how the buffer-full codepath behaves, and its not good - while there is no deadlock condition, the app shows signs of threadpool starvation, probably due to doing Task.Result on a sync codepath.
> 
> ```
> Bombarding http://py39-perfcollect-channel100-batched.app:80/api/helloworld for 1m0s using 100 connection(s)
> Statistics        Avg      Stdev        Max
>   Reqs/sec        26.29     158.67    2818.03
>   Latency         3.84s      3.02s     10.04s
>   Latency Distribution
>      50%   267.03ms
>      75%     10.01s
>      90%     10.02s
>      95%     10.02s
>      99%     10.03s
>   HTTP codes:
>     1xx - 0, 2xx - 1066, 3xx - 0, 4xx - 0, 5xx - 0
>     others - 600
>   Errors:
>     the server closed connection before returning the first response byte. Make sure the server returns 'Connection: close' response header before closing the connection - 600
>   Throughput:     9.48KB/s
> ```
> 
> ```
> [System.Runtime]
>     ThreadPool Thread Count                                              233
> ```
> 
> Now with the planned buffer size of 8000 I don't think this condition is likely to be hit, but I consider this a deal breaker for the current implementation. I will do more testing and find an alternative, and I'll re-request reviews once I've pushed an update.

Summarizing the newest changes - I still use a channel for the buffer, and the buffer is consumed using standard code for reading a channel. However on the write path, I don't take a dependency on any of the async functionality of the channel. If the channel is full, we block on a ManualResetEvent that is reset by the channel reader. As before, we apply a timeout policy so we give up on log ordering if the MRE is not reset.

Tested this new implementation with a tiny buffer, and it does not run into any thread starvation issues:

**Buffer size: 10 (deliberately very small)**
```
PS D:\code> d:\tools\bombardier-windows-amd64.exe -c 100 -d 60s -m GET --format=pt -p i,r -l http://localhost:5000/api/function1
Bombarding http://localhost:5000/api/function1 for 1m0s using 100 connection(s)
Statistics        Avg      Stdev        Max
  Reqs/sec       827.29      87.94    1280.83
  Latency      120.74ms   151.87ms      3.75s
  Latency Distribution
     50%   112.52ms
     75%   116.04ms
     90%   120.54ms
     95%   124.52ms
     99%   132.04ms
  HTTP codes:
    1xx - 0, 2xx - 49718, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:   206.22KB/s
```

paulbatum:
Results from testing latest revision (bb9526e2f) on ACI, median of 3 runs:

bb9526e2f with buffer disabled - should have similar perf characteristics to what is in prod today:
```
.\bombardier-windows-amd64.exe -c 30 -d 60s -m GET --format=pt -p i,r -l http://py39-aci-noai-consolebuffer-bb9526e2f-baseline.app/api/helloworld
Bombarding http://py39-aci-noai-consolebuffer-bb9526e2f-baseline.app:80/api/helloworld for 1m0s using 30 connection(s)
Statistics        Avg      Stdev        Max
  Reqs/sec       106.60      72.31     966.79
  Latency      280.87ms    42.80ms   511.59ms
  Latency Distribution
     50%   280.29ms
     75%   307.93ms
     90%   334.41ms
     95%   354.45ms
     99%   399.58ms
  HTTP codes:
    1xx - 0, 2xx - 6417, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:    31.57KB/s
```

bb9526e2f with default settings (buffer size 8000):
```
.\bombardier-windows-amd64.exe -c 30 -d 60s -m GET --format=pt -p i,r -l http://py39-aci-noai-consolebuffer-bb9526e2f-channel8000.app/api/helloworld
Bombarding http://py39-aci-noai-consolebuffer-bb9526e2f-channel8000.app:80/api/helloworld for 1m0s using 30 connection(s)
Statistics        Avg      Stdev        Max
  Reqs/sec       221.48     178.21    1450.01
  Latency      135.50ms    23.86ms   277.00ms
  Latency Distribution
     50%   134.65ms
     75%   153.68ms
     90%   172.81ms
     95%   184.52ms
     99%   212.01ms
  HTTP codes:
    1xx - 0, 2xx - 13301, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:    66.11KB/s
```

For some reason, I'm back to getting a full 2x perf increase from the buffer. Not that I'm complaining, but I do wish I understood what causes such significant swings when trying to quantify the upside of these changes.

paulbatum:
Results from testing latest revision (1f77a72) on ACI, median of 3 runs:

1f77a72 with buffer disabled - should have similar perf characteristics to what is in prod today:
```
Bombarding http://py39-aci-noai-consolebuffer-1f77a72-baseline.app:80/api/helloworld for 1m0s using 60 connection(s)
Statistics        Avg      Stdev        Max
  Reqs/sec       104.20     120.20    1668.76
  Latency      572.93ms    88.24ms      0.87s
  Latency Distribution
     50%   584.84ms
     75%   637.22ms
     90%   677.15ms
     95%   703.73ms
     99%   757.85ms
  HTTP codes:
    1xx - 0, 2xx - 6304, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:    30.70KB/s
```

1f77a72 with default settings (buffer size 8000):
```
Bombarding http://py39-aci-noai-consolebuffer-1f77a72-channel8000.app:80/api/helloworld for 1m0s using 60 connection(s)
Statistics        Avg      Stdev        Max
  Reqs/sec       151.36     185.57    2199.02
  Latency      395.61ms    76.41ms   698.17ms
  Latency Distribution
     50%   407.23ms
     75%   448.89ms
     90%   487.69ms
     95%   514.34ms
     99%   582.81ms
  HTTP codes:
    1xx - 0, 2xx - 9123, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:    44.94KB/s
```

This 45% performance improvement is similar to what I observed when testing in EUAP using the original channel base implementation, so it looks like these changes preserve the gains we measured earlier without introducing threadpool starvation issues in scenarios where the buffer is full.

Changes are ready for re-review.

src/WebJobs.Script.WebHost/Diagnostics/BufferedConsoleWriter.cs
```diff
@@ -0,0 +1,150 @@
...
+                {
+                    Console.WriteLine(evt);
+                }
+            }
+        }
+
+        internal void StartProcessingBuffer()
+        {
+            // intentional no-op if the task is already running
+            if (_consoleBufferReadLoop == null || _consoleBufferReadLoop.IsCompleted)
```
jviau, line 77: Does this need to be thread-safe? Technically we could have 2 calls here both kick off a `ProcessConsoleBufferAsync`. A `lock` section would fix it.

Usage of it in this PR shows no possible concurrent calls to this. Maybe at least add a comment saying not thread-safe, to hopefully avoid future refactoring introducing a concurrent call to this.

src/WebJobs.Script.WebHost/Diagnostics/LinuxContainerEventGenerator.cs
```diff
@@ -2,29 +2,64 @@
...
-        private readonly bool _consoleEnabled = true;
+        private readonly BufferedConsoleWriter _consoleWriter;
         private readonly IEnvironment _environment;
         private string _containerName;
         private string _stampName;
         private string _tenantId;
+        private bool _disposed;
 
-        public LinuxContainerEventGenerator(IEnvironment environment, Action<string> writeEvent = null)
+        public LinuxContainerEventGenerator(IEnvironment environment, IOptions<ConsoleLoggingOptions> consoleLoggingOptions)
```
jviau, line 23: You may need to decorate this with `[ActivatorUtilitiesConstructor]` so DI knows to use this ctor.

src/WebJobs.Script.WebHost/Diagnostics/LinuxContainerEventGenerator.cs
```diff
@@ -142,5 +170,23 @@ public static void LogEvent(string message, Exception e = null, LogLevel logLeve
                 slotName: SystemEnvironment.Instance.GetSlotName() ?? string.Empty,
                 eventTimestamp: DateTime.UtcNow);
         }
+
+        protected virtual void Dispose(bool disposing)
```
jviau, line 174: Is this class inherited anywhere? If not, can mark it `sealed` to avoid needing the dispose(true) pattern.

(did I leave this comment already and miss your answer? Getting deja-vu here, apologies if this is duplicate.)

src/WebJobs.Script/Config/ConsoleLoggingOptionsSetup.cs
```diff
@@ -0,0 +1,33 @@
...
+        {
+            options.LoggingDisabled = _configuration.GetValue<int>(EnvironmentSettingNames.ConsoleLoggingDisabled) == 1;
+
+            int bufferSize = _configuration.GetValue<int>(EnvironmentSettingNames.ConsoleLoggingBufferSize, ConsoleLoggingOptions.DefaultBufferSize);
+            if (bufferSize < 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(EnvironmentSettingNames.ConsoleLoggingBufferSize), "Console buffer size cannot be negative");
+            }
+
+            options.BufferEnabled = bufferSize > 0;
```
jviau: This feels a bit unintuitive to me: we will ignore any explicit `BufferEnabled = false` settings, and always set it based on `BufferSize`. Two ways I think this can be fixed:

1. Make `BufferEnabled` `internal`, so only `BufferSize` is exposed publicly and document that setting to `0` disables buffer.
2. Make `BufferEnabled` a `bool?` and only set it if it is still `null`.

src/WebJobs.Script/Config/ConsoleLoggingOptionsSetup.cs
```diff
@@ -0,0 +1,33 @@
...
+        public ConsoleLoggingOptionsSetup(IConfiguration configuration)
+        {
+            _configuration = configuration;
+        }
+
+        public void Configure(ConsoleLoggingOptions options)
+        {
+            options.LoggingDisabled = _configuration.GetValue<int>(EnvironmentSettingNames.ConsoleLoggingDisabled) == 1;
+
+            int bufferSize = _configuration.GetValue<int>(EnvironmentSettingNames.ConsoleLoggingBufferSize, ConsoleLoggingOptions.DefaultBufferSize);
```
jviau: While this is fine, it does make the value of `BufferSize` a bit subjective to the order of multiple `.Configure` calls. Some recommendations on improving it:

1. Instead of setting to default here, just set to default in the options object itself (`public int BufferSize { get; set; } = 8000;`), then would need a TryGet style approach for reading the value - setting only if it is present (and not mistakenly setting 0 when it is not present).
2. OR - make `BufferSize` an `int?` and set it to default only if it is still `null`.

jviau (approved):
Looks good! Can you verify if the buffer flushing will also occur on an unhandled exception / app crash?

I am sure there are some scenarios it won't be, which are unavoidable (StackOverflowException I believe, maybe some others). But nothing we can do about that.

src/WebJobs.Script.WebHost/Diagnostics/LinuxContainerEventGenerator.cs
paulbatum: You did leave a comment, and I incorporated your feedback:
- decided to make BufferedConsoleWriter sealed, and gave it a simplified a dispose implementation
- kept LinuxContainerEventGenerator unsealed, because its already part of a hierarchy that might get extended further

src/WebJobs.Script.WebHost/Diagnostics/LinuxContainerEventGenerator.cs
paulbatum: If that was necessary, the current functionality wouldn't work right? Also I searched the codebase and couldn't find any references to this type.

paulbatum: This type is instantiated explicitly, in the logic for the registration of implementations for IEventGenerator.

src/WebJobs.Script.WebHost/Diagnostics/BufferedConsoleWriter.cs
paulbatum: Sure, adding a comment being explicit about intended use

src/WebJobs.Script/Config/ConsoleLoggingOptionsSetup.cs
paulbatum: OK, making buffer size setter internal.

src/WebJobs.Script/Config/ConsoleLoggingOptionsSetup.cs
paulbatum: OK, added a test case that demonstrates the problem and made it pass by checking if the setting is present.

paulbatum:
> Looks good! Can you verify if the buffer flushing will also occur on an unhandled exception / app crash?
> 
> I am sure there are some scenarios it won't be, which are unavoidable (StackOverflowException I believe, maybe some others). But nothing we can do about that.

Yep, tested a few of these manually, by running both in direct console mode and buffered mode and comparing:

- Out of memory exception - logs are correctly flushed with exception details and the process terminates
- Environment.Exit(1) - logs are correctly flushed and the process terminates
- Environment.FailFast("kaboom") - we don't get a full flush in this case, just the termination message written to stdout:

```
MS_FUNCTION_LOGS 4,,,Function1,FunctionStarted,Function.Function1,"","Executing 'Functions.Function1' (Reason='This function was programmatically called via the host APIs.', Id=7302978d-a1ce-48db-b0ce-73bd39ea70bc)",4.21.0.0,2023-05-26T23:42:53.7145981Z,,"",7302978d-a1ce-48db-b0ce-73bd39ea70bc,345d2a1a-d0bf-4239-87e7-0228153fdf04,,FOO,,,,
Process terminated. kaboom
   at System.Environment.FailFast(System.String)
   at Microsoft.Azure.WebJobs.Script.Workers.Rpc.RpcFunctionInvocationDispatcher+<InvokeAsync>d__59.MoveNext()
   at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(System.Threading.Thread, System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object)
```

- StackOverflowException - as you expected, we don't get a full flush in this case, just the stackoverflow exception details:

```
MS_FUNCTION_LOGS 4,,,Function1,FunctionStarted,Function.Function1,"","Executing 'Functions.Function1' (Reason='This function was programmatically called via the host APIs.', Id=24211d31-8d26-487f-87ed-816858138616)",4.21.0.0,2023-05-26T23:40:34.4972600Z,,"",24211d31-8d26-487f-87ed-816858138616,7562dc7b-e903-413e-bc7b-bd51b6027ea9,,FOO,,,,
Stack overflow.
Repeat 19410 times:
--------------------------------
   at Microsoft.Azure.WebJobs.Script.Workers.Rpc.RpcFunctionInvocationDispatcher.Overflow2(Int32, Int32)
--------------------------------
   at Microsoft.Azure.WebJobs.Script.Workers.Rpc.RpcFunctionInvocationDispatcher+<InvokeAsync>d__59.MoveNext()
   at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(System.Threading.Thread, System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object)
```

I will note, that the two examples where we dont get a full flush also appear to have an issue where they are not formatted according to our rules - note the lack of  `MS_FUNCTION_LOGS` prefix. I think AppDomain.UnhandledException simply does not fire in these cases and so there is no opportunity for our formatting to run. I confirmed that this behavior is consistent regardless of whether the buffer is enabled, so it doesn't reflect a problem with this PR.

src/WebJobs.Script.WebHost/Diagnostics/BufferedConsoleWriter.cs
```diff
@@ -0,0 +1,151 @@
...
+                            builder.AppendLine(line2);
+                            int lines = 2;
+
+                            while (lines < _maxLinesPerWrite && _consoleBuffer.Reader.TryRead(out string nextLine))
+                            {
+                                builder.AppendLine(nextLine);
+                                lines++;
+                            }
+
+                            _writeResetEvent.Set();
```
mathewc, line 110: You chose to signal the event after writing potentially many lines, as opposed to signaling after each time you've successfully read a new line and made space in the channel. Probably doesn't make a huge difference, but wouldn't signaling after each read unblock the writer faster?

mathewc (approved):
Looks good! Just one question added.

src/WebJobs.Script.WebHost/Diagnostics/BufferedConsoleWriter.cs
paulbatum: Good question - when I was iterating on this code I did run a couple of quick tests with both variations, and I was not able to find a significant difference either way, but my testing wasn't particularly indepth.

Without having any useful data, I ended up going with the current approach. Here's my rough thinking, split into two cases - environments where we have multiple cores, and environments where we have a single core or fractional core.

Multiple cores - chances of the buffer getting full in a multicore environment are low, because it would require ~1000 function executions to occur without the threadpool dispatching the thread for the buffer consumer. So if the buffer is not getting full, why pay the cost of setting the MRE every iteration?

Single/fractional core - if we cannot run multiple threads concurrently, what is the benefit of waking up a writing thread while the consumer is still running?

PR #9253, opened 2023-05-04
fixing issue with worker indexing, dotnet-isolated, and placeholders

brettsam:

With "worker indexing" enabled, it means that all worker channels are created at the WebHost level. This works for every worker except dotnet-isolated, which relies on worker configuration coming from the customer payload at the JobHost level. This feature is not yet fully supported in dotnet-isolated, so it wasn't noticed originally.

The fix here is to propagate the worker configuration through to the initialization step of the WebHost grpc channel. Nearly everything was already wired up, but there was no API to accept the configuration during initialization. This meant the WebHostChannelManager was ignroing this, and always using the WebHost configuration.

In practice, this meant that during a "placeholder miss", the host kept restarting the dotnet-isolated native placeholder process. It should have been seeing the JobHost config and starting the customer's app directly.

I've also added a new dotnet-isolated application for some e2e tests around this.

### Pull request checklist


src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
```diff
@@ -52,26 +51,25 @@ public class WebHostRpcWorkerChannelManager : IWebHostRpcWorkerChannelManager
...
             _logger = loggerFactory.CreateLogger<WebHostRpcWorkerChannelManager>();
             _applicationHostOptions = applicationHostOptions;
-            _languageWorkerOptions = languageWorkerOptions;
 
             _shutdownStandbyWorkerChannels = ScheduleShutdownStandbyChannels;
             _shutdownStandbyWorkerChannels = _shutdownStandbyWorkerChannels.Debounce(milliseconds: 5000);
         }
 
-        public Task<IRpcWorkerChannel> InitializeChannelAsync(string runtime)
+        public Task<IRpcWorkerChannel> InitializeChannelAsync(IEnumerable<RpcWorkerConfig> workerConfigs, string runtime)
```
brettsam, line 59: This is the major change that caused a ripple throughout the other APIs. Rather than always using the WebHost-level workerConfigs (which could be wrong -- specifically for dotnet-isolated), it now will use the ones passed in.

src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
```diff
@@ -65,7 +65,7 @@ public async Task<FunctionMetadataResult> GetFunctionMetadataAsync(IEnumerable<R
 
                 if (channels?.Any() != true)
                 {
-                    await _channelManager.InitializeChannelAsync(_workerRuntime);
+                    await _channelManager.InitializeChannelAsync(workerConfigs, _workerRuntime);
```
brettsam, line 68: Here, we've been given the correct workerConfigs to use, but they were never being used for anything. This change now uses them for initialization, which allows a JobHost to override the WebHost, which is a requirement for dotnet-isolated.

test/DotNetIsolated60/Function1.cs
```diff
@@ -0,0 +1,30 @@
+using System.Net;
+using Microsoft.Azure.Functions.Worker;
+using Microsoft.Azure.Functions.Worker.Http;
+using Microsoft.Extensions.Logging;
+
+namespace DotNetIsolated60
```
brettsam, line 6: New test app explicitly added for e2e testing.

liliankasem (approved):

NuGet.config
```diff
@@ -9,6 +9,7 @@
     <add key="AzureFunctions@staging" value="https://azfunc.pkgs.visualstudio.com/e6a70c92-4128-439f-8012-382fe78d6396/_packaging/AzureFunctions%40staging/nuget/v3/index.json" />
     <add key="AzureFunctionsRelease" value="https://azfunc.pkgs.visualstudio.com/e6a70c92-4128-439f-8012-382fe78d6396/_packaging/AzureFunctionsRelease/nuget/v3/index.json" />
     <add key="AzureFunctionsPreRelease" value="https://azfunc.pkgs.visualstudio.com/e6a70c92-4128-439f-8012-382fe78d6396/_packaging/AzureFunctionsPreRelease/nuget/v3/index.json" />
+    <add key="AzureFunctionsTempStaging" value="https://azfunc.pkgs.visualstudio.com/e6a70c92-4128-439f-8012-382fe78d6396/_packaging/AzureFunctionsTempStaging/nuget/v3/index.json" />
```
brettsam, line 12: Testing also found the issue fixed in the worker with https://github.com/Azure/azure-functions-dotnet-worker/pull/1508. Added this feed so we can be sure to pick up pre-release packages if we need to.

src/WebJobs.Script/Workers/Rpc/RpcWorkerProcess.cs
```diff
@@ -49,13 +51,15 @@ internal class RpcWorkerProcess : WorkerProcess
...
             _hostingConfigOptions = hostingConfigOptions;
+            _environment = environment;
         }
 
         internal override Process CreateWorkerProcess()
         {
             var workerContext = new RpcWorkerContext(Guid.NewGuid().ToString(), RpcWorkerConstants.DefaultMaxMessageLengthBytes, _workerId, _workerProcessArguments, _scriptRootPath, _serverUri);
             workerContext.EnvironmentVariables.Add(WorkerConstants.FunctionsWorkerDirectorySettingName, _workerDirectory);
             workerContext.EnvironmentVariables.Add(WorkerConstants.FunctionsApplicationDirectorySettingName, _scriptRootPath);
+            workerContext.EnvironmentVariables.Add(RpcWorkerConstants.FunctionWorkerRuntimeVersionSettingName, _environment.GetEnvironmentVariable(RpcWorkerConstants.FunctionWorkerRuntimeVersionSettingName));
```
brettsam, line 62: In prod, we were relying on this value being set at the machine-level by Antares. Because of this, I wasn't able to properly test. I'm setting this explicitly now for testing purposes (and will also be required when we eventually abstract the process-start away to enable cross-machine workers).

soninaren (approved):

PR #9282, opened 2023-05-18
Comment out package reference for Azure.Functions.DotNetIsolatedNativeHost 

kshyju:
Comment out package reference for Azure.Functions.DotNetIsolatedNativeHost so that the dotnet-isolated directory won't be present in the "workers" directory. Without this directory, we will not be executing the new placeholder related code path for dotnet-isolated.


fabiocav (approved):

fabiocav:
This will trigger a number of test failures.

VpOfEngineering (approved):

PR #9285, opened 2023-05-22
Revert "Comment out package reference for Azure.Functions.DotNetIsolatedNativeHost  (#9282)"

brettsam:
This reverts commit 4ec73a34dea76db91b6861db59d57d1ff00265f8, which will re-add the dotnet-isolated placeholder.

Then also updates code to treat a null FUNCTIONS_WORKER_RUNTIME and "dll" scriptFile extension as "dotnet"... previously, this would choose "dotnet-isolated" and cause issues.

### DIFF
--- a/src/WebJobs.Script/Diagnostics/Extensions/ScriptHostLoggerExtension.cs
+++ b/src/WebJobs.Script/Diagnostics/Extensions/ScriptHostLoggerExtension.cs
@@ -12,111 +12,123 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics.Extensions
 
         private static readonly Action<ILogger, Exception> _hostIdIsSet =
             LoggerMessage.Define(
-            LogLevel.Warning,
-            new EventId(400, nameof(HostIdIsSet)),
-            "Host id explicitly set in configuration. This is not a recommended configuration and may lead to unexpected behavior.");
+                LogLevel.Warning,
+                new EventId(400, nameof(HostIdIsSet)),
+                "Host id explicitly set in configuration. This is not a recommended configuration and may lead to unexpected behavior.");
 
         private static readonly Action<ILogger, string, string, Exception> _functionError =
             LoggerMessage.Define<string, string>(
-            LogLevel.Error,
-            new EventId(402, nameof(FunctionError)),
-            "The '{functionName}' function is in error: {errorMessage}");
+                LogLevel.Error,
+                new EventId(402, nameof(FunctionError)),
+                "The '{functionName}' function is in error: {errorMessage}");
 
         private static readonly Action<ILogger, Exception> _hostIsInPlaceholderMode =
             LoggerMessage.Define(
-            LogLevel.Debug,
-            new EventId(403, nameof(HostIsInPlaceholderMode)),
-            "Host is in placeholdermode");
+                LogLevel.Debug,
+                new EventId(403, nameof(HostIsInPlaceholderMode)),
+                "Host is in placeholdermode");
 
         private static readonly Action<ILogger, string, Exception> _addingDescriptorProviderForLanguage =
             LoggerMessage.Define<string>(
-            LogLevel.Debug,
-            new EventId(404, nameof(AddingDescriptorProviderForLanguage)),
-            "Adding Function descriptor provider for language {workerRuntime}.");
+                LogLevel.Debug,
+                new EventId(404, nameof(AddingDescriptorProviderForLanguage)),
+                "Adding Function descriptor provider for language {workerRuntime}.");
 
         private static readonly Action<ILogger, Exception> _creatingDescriptors =
             LoggerMessage.Define(
-            LogLevel.Debug,
-            new EventId(405, nameof(CreatingDescriptors)),
-            "Creating function descriptors.");
+                LogLevel.Debug,
+                new EventId(405, nameof(CreatingDescriptors)),
+                "Creating function descriptors.");
 
         private static readonly Action<ILogger, Exception> _descriptorsCreated =
             LoggerMessage.Define(
-            LogLevel.Debug,
-            new EventId(406, nameof(DescriptorsCreated)),
-            "Function descriptors created.");
+                LogLevel.Debug,
+                new EventId(406, nameof(DescriptorsCreated)),
+                "Function descriptors created.");
 
         private static readonly Action<ILogger, Exception> _errorPurgingLogFiles =
             LoggerMessage.Define(
-            LogLevel.Error,
-            new EventId(407, nameof(ErrorPurgingLogFiles)),
-            "An error occurred while purging log files");
+                LogLevel.Error,
+                new EventId(407, nameof(ErrorPurgingLogFiles)),
+                "An error occurred while purging log files");
 
         private static readonly Action<ILogger, string, Exception> _deletingLogDirectory =
             LoggerMessage.Define<string>(
-            LogLevel.Debug,
-            new EventId(408, nameof(DeletingLogDirectory)),
-            "Deleting log directory '{logDir}'");
+                LogLevel.Debug,
+                new EventId(408, nameof(DeletingLogDirectory)),
+                "Deleting log directory '{logDir}'");
 
         private static readonly Action<ILogger, string, string, Exception> _failedToLoadType =
             LoggerMessage.Define<string, string>(
-            LogLevel.Warning,
-            new EventId(409, nameof(FailedToLoadType)),
-            "Failed to load type '{typeName}' from '{path}'");
+                LogLevel.Warning,
+                new EventId(409, nameof(FailedToLoadType)),
+                "Failed to load type '{typeName}' from '{path}'");
 
         private static readonly Action<ILogger, string, Exception> _configurationError =
             LoggerMessage.Define<string>(
-            LogLevel.Information,
-            new EventId(410, nameof(ConfigurationError)),
-            "{message}");
+                LogLevel.Information,
+                new EventId(410, nameof(ConfigurationError)),
+                "{message}");
 
         private static readonly Action<ILogger, string, string, Exception> _versionRecommendation =
             LoggerMessage.Define<string, string>(
-            LogLevel.Warning,
-            new EventId(411, nameof(VersionRecommendation)),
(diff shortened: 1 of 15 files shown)


src/WebJobs.Script/Diagnostics/Extensions/ScriptHostLoggerExtension.cs
```diff
@@ -12,111 +12,123 @@ internal static class ScriptHostLoggerExtension
...
+            LoggerMessage.Define<string>(
+                LogLevel.Debug,
+                new EventId(419, nameof(FunctionsWorkerRuntimeValue)),
+                $"{EnvironmentSettingNames.FunctionWorkerRuntime} value: '{{workerRuntime}}'");
+
+        private static readonly Action<ILogger, string, Exception> _resolvedWorkerRuntimeFromMetadata =
+            LoggerMessage.Define<string>(
+                LogLevel.Debug,
+                new EventId(420, nameof(ResolvedWorkerRuntimeFromMetadata)),
+                $"{EnvironmentSettingNames.FunctionWorkerRuntime} is null. Resolved worker runtime from function metadata: '{{workerRuntime}}'");
```
kshyju, line 131: Did you mean to have two `{{` ? I see all other entries not doing that.

kshyju (approved):

src/WebJobs.Script/Diagnostics/Extensions/ScriptHostLoggerExtension.cs
brettsam: yeah -- it's b/c this one is using string interpolation (`$`). So need to escape the `{}` around workerRuntime. It ends up being `FUNCTIONS_WORKER_RUNTIME is null. Resolved worker Resolved worker runtime from function metadata: '{workerRuntime}'`

src/WebJobs.Script/Host/ScriptHost.cs
```diff
@@ -277,17 +275,21 @@ public async Task InitializeAsync(CancellationToken cancellationToken = default)
                 PreInitialize();
                 HostInitializing?.Invoke(this, EventArgs.Empty);
 
-                _workerRuntime = _workerRuntime ?? _environment.GetEnvironmentVariable(EnvironmentSettingNames.FunctionWorkerRuntime);
```
brettsam, line 280: Note -- having `_workerRuntime` as a field was confusing as it was updated in multiple places. I've made it into a local here so this is the only flow where this can be determined.

src/WebJobs.Script/Host/ScriptHost.cs
```diff
@@ -277,17 +275,21 @@ public async Task InitializeAsync(CancellationToken cancellationToken = default)
...
                 IEnumerable<FunctionMetadata> functionMetadataList = GetFunctionsMetadata();
 
+                string workerRuntime = _environment.GetEnvironmentVariable(EnvironmentSettingNames.FunctionWorkerRuntime);
+                _logger.FunctionsWorkerRuntimeValue(workerRuntime);
+
+                if (workerRuntime is null)
+                {
+                    workerRuntime = Utility.GetWorkerRuntime(functionMetadataList);
+                    _logger.ResolvedWorkerRuntimeFromMetadata(workerRuntime);
+                }
```
brettsam, line 288: We now explicitly log how we've determined the worker runtime.

src/WebJobs.Script/Host/HostFunctionMetadataProvider.cs
```diff
@@ -179,10 +183,21 @@ internal static string ParseLanguage(string scriptFilePath, IEnumerable<RpcWorke
...
+            // - when FUNCTIONS_WORKER_RUNTIME is set to "dotnet", workerConfigs is empty so this correctly falls
+            //   through to choose DotNetAssembly with a "dll" extension
+            if (!(functionsWorkerRuntime is null && extension == "dll"))
             {
-                return workerConfig.Description.Language;
+                var workerConfig = workerConfigs.FirstOrDefault(config => config.Description.Extensions.Contains("." + extension));
+                if (workerConfig != null)
+                {
+                    return workerConfig.Description.Language;
+                }
```
brettsam: The fix is here -- hopefully this description is good enough to explain to us in the future why it works this way. Eventually we'll have a no-in-proc host so this flow won't matter as much.

src/WebJobs.Script/Host/HostFunctionMetadataProvider.cs
```diff
@@ -179,10 +183,21 @@ internal static string ParseLanguage(string scriptFilePath, IEnumerable<RpcWorke
 
             // determine the script type based on the primary script file extension
             string extension = Path.GetExtension(scriptFilePath).ToLowerInvariant().TrimStart('.');
-            var workerConfig = workerConfigs.FirstOrDefault(config => config.Description.Extensions.Contains("." + extension));
-            if (workerConfig != null)
+
+            // In the special case of a null FUNCTIONS_WORKER_RUNTIME and an extension of "dll",
```
fabiocav: There's an assumption made by this method that the value passed in the parameter is coming from reading `FUNCTIONS_WORKER_RUNTIME`, while the method uses what is passed in, which could eventually come from different sources. Would be good to update the comment to avoid confusion here in the future.

src/WebJobs.Script/Host/HostFunctionMetadataProvider.cs
```diff
@@ -179,10 +183,21 @@ internal static string ParseLanguage(string scriptFilePath, IEnumerable<RpcWorke
...
+            // we want to skip the worker config check (to prevent the dotnet-isolated worker from
+            // being chosen) and allow in-proc handling to be used.
+            // Other notes:
+            // - when dotnet-isolated, the FUNCTIONS_WORKER_RUNTIME is required to be set to "dotnet-isolated"
+            // - when FUNCTIONS_WORKER_RUNTIME is set to "dotnet", workerConfigs is empty so this correctly falls
+            //   through to choose DotNetAssembly with a "dll" extension
+            if (!(functionsWorkerRuntime is null && extension == "dll"))
```
fabiocav: Can we change the condition here to avoid the negation? 

For the extension, can you update this to use `string.Equals` and perform a case insensitive check?

src/WebJobs.Script/Host/HostFunctionMetadataProvider.cs
```diff
@@ -179,10 +183,21 @@ internal static string ParseLanguage(string scriptFilePath, IEnumerable<RpcWorke
...
             {
-                return workerConfig.Description.Language;
+                var workerConfig = workerConfigs.FirstOrDefault(config => config.Description.Extensions.Contains("." + extension));
```
fabiocav: I realize this is pre-existing code, and wouldn't expect this would ever be the case (and have no concrete scenario used today), but the current check could potentially lead to a false positive on a partial match (e.g., `.js` matching on `.jsx`).

Would also be good to make sure the check here is case insensitive.

src/WebJobs.Script/Host/ScriptHost.cs
```diff
@@ -368,7 +370,7 @@ private IEnumerable<FunctionMetadata> GetFunctionsMetadata()
             IEnumerable<FunctionMetadata> functionMetadata;
 
             functionMetadata = _functionMetadataManager.GetFunctionMetadata(forceRefresh: false, workerConfigs: _languageWorkerOptions.CurrentValue.WorkerConfigs);
-            _workerRuntime ??= Utility.GetWorkerRuntime(functionMetadata);
```
fabiocav, line 371: 👍🏼

src/WebJobs.Script/Host/ScriptHost.cs
```diff
@@ -67,16 +67,15 @@ public class ScriptHost : JobHost, IScriptJobHost
...
+        private readonly IFunctionInvocationDispatcher _functionDispatcher;
         private static readonly int _processId = Process.GetCurrentProcess().Id;
+        public static readonly string Version = GetAssemblyFileVersion(typeof(ScriptHost).Assembly);
 
         private ValueStopwatch _stopwatch;
-        private IPrimaryHostStateProvider _primaryHostStateProvider;
-        public static readonly string Version = GetAssemblyFileVersion(typeof(ScriptHost).Assembly);
         private ScriptSettingsManager _settingsManager;
-        private ILogger _logger = null;
-        private string _workerRuntime;
```
fabiocav, line 77: 👍🏼

test/WebJobs.Script.Tests/ScriptHostTests.cs
```diff
@@ -1513,9 +1513,11 @@ public async Task Initialize_LogsWarningForExplicitlySetHostId()
         }
 
         [Theory]
-        [InlineData("main.py", "python", "python")]
-        [InlineData("app.dll", "dotnet", DotNetScriptTypes.DotNetAssembly)] // if FUNCTIONS_WORKER_RUNTIME is missing, assume dotnet
-        public async Task Initialize_MissingWorkerRuntime_SetsCorrectRuntimeFromFunctionMetadata(string scriptFile, string expectedMetricLanguage, string expectedMetadataLanguage)
+        [InlineData("python", "main.py", "python", "python")]
+        [InlineData("dotnet-isolated", "app.dll", "dotnet-isolated", "dotnet-isolated")]
+        [InlineData("dotnet", "app.dll", "dotnet", DotNetScriptTypes.DotNetAssembly)]
+        [InlineData(null, "app.dll", "dotnet", DotNetScriptTypes.DotNetAssembly)] // if FUNCTIONS_WORKER_RUNTIME is missing, assume dotnet
```
fabiocav, line 1519: 👍🏼

src/WebJobs.Script/Host/HostFunctionMetadataProvider.cs
brettsam: i found it very difficult to read the non-negated condition... but let me add a small helper that may make it easier to read and parse.

src/WebJobs.Script/Host/HostFunctionMetadataProvider.cs
brettsam: Extensions is a `List<string>` so it's doing a direct comparison to strings in the list -- so .jsx won't match with .js.

However, I will add `OrdinalIgnoreCase` to this check.

src/WebJobs.Script/Host/HostFunctionMetadataProvider.cs
brettsam: actually -- i just short-circuited it -- i think it reads more nicely with the comment

PR #9292, opened 2023-05-24
Map WebJobs Extension RPC data sources onto the gRPC host endpoints

### LINKED ISSUE #9291: Map WebJobs RPC extensions onto the gRPC host
This issue is to track the work related to consuming the new WebJobs extensions that will be added as part of https://github.com/Azure/azure-webjobs-sdk/pull/2978. The necessary work for the functions host is to:

When the WebJobs script host is ready:
1. discover `WebJobsRpcEndpointDataSource` from the new script host's `IServiceProvider`.
2. map them onto the gRPC hosts available endpoints.

Additionally, as part of mapping the endpoints we should:

1. Ensure the gRPC service is resolved from a new scope of the script hosts `IServiceProvider`
2. Replace the `HttpContext.RequestServices` with the script hosts `IServiceProvider`.

jviau:
This PR is to a feature branch. The intention is to review, check in, build & release feature branch to internal feed. After testing and validation by partners, we will then merge feature branch into `dev`.

This PR consumes the new `Microsoft.Azure.WebJobs.Rpc.Core` package (from our ADO feed for now), which brings in a single contract `WebJobsRpcEndpointDataSource`. This type is used by WebJobs extensions via the `Microsoft.Azure.WebJobs.Extensions.Rpc` package to register and export their RPC extensions (only gRPC supported at this time). We then discover all of those data sources from the active script host and map them into the existing AspNetCore server we use for our own gRPC communication between host and worker. The end result here is that WebJobs extensions can easily hand off their gRPC service implementations to us for endpoint mapping. This enables custom RPC scenarios between worker and host.

resolves #9291

### Pull request checklist

    * [ ] Otherwise: Documentation issue linked to PR -- **TODO in another PR**


src/WebJobs.Script.Grpc/Server/ExtensionsCompositeEndpointDataSource.cs
```diff
@@ -0,0 +1,300 @@
...
+    /// Endpoint data source which composes all the WebJobs extension data sources together.
+    /// </summary>
+    /// <remarks>
+    /// Implementation is adapted from <see cref="CompositeEndpointDataSource"/>.
+    /// https://github.com/dotnet/aspnetcore/blob/main/src/Http/Routing/src/CompositeEndpointDataSource.cs.
+    /// </remarks>
+    internal sealed class ExtensionsCompositeEndpointDataSource : EndpointDataSource, IDisposable
+    {
+        private readonly object _lock = new();
+        private readonly List<EndpointDataSource> _dataSources = new();
```
liliankasem, line 27: Is this a good candidate for lazy initialization?

src/WebJobs.Script.Grpc/Server/ExtensionsCompositeEndpointDataSource.cs
```diff
@@ -0,0 +1,300 @@
...
+        private void CreateChangeTokenUnsynchronized(bool collectionChanged)
+        {
+            CancellationTokenSource cts = new();
+
+            if (collectionChanged)
+            {
+                _changeTokenRegistrations = new();
+                foreach (var dataSource in _dataSources)
+                {
+                    _changeTokenRegistrations.Add(ChangeToken.OnChange(
```
liliankasem, line 167: I'm not super familiar with this, but was wondering if using [CompositeChangeToken](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/change-tokens?view=aspnetcore-7.0#compositechangetoken-class) works here instead of manually managing a list of change token registrations? 

If this doesn't work for us, I'd love to learn more about why (just for my learning)

src/WebJobs.Script.Grpc/Server/ExtensionsCompositeEndpointDataSource.cs
```diff
@@ -0,0 +1,300 @@
...
+        /// <inheritdoc />
+        public override IChangeToken GetChangeToken()
+        {
+            ThrowIfDisposed();
+            EnsureChangeTokenInitialized();
+            return _consumerChangeToken;
+        }
+
+        /// <inheritdoc />
+        public void Dispose()
```
daviburg, line 63: Should this follow the dispose implementation pattern https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/dispose-pattern#basic-dispose-pattern
?

jviau: The purpose of that pattern is for inheritance. This class is `sealed`, so there are no potential inheritors, and that pattern is not needed. See [CA1063](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1063), which directly mentions using `sealed` as a fix for this.

src/WebJobs.Script.Grpc/Server/ExtensionsCompositeEndpointDataSource.cs
jviau: Theoretically we could, but I don't think there is much benefit for it.

1. This class is typically only ever 1 instance, so not much savings.
2. To get any savings, we can't use `Lazy<T>` (we would just be trading 1 object for another). We would need to remove `readonly` and then add more code to get back to the same level of nullability safety.

I don't think the work is worth it to save 1 list, which is not in a hot path.

src/WebJobs.Script.Grpc/Server/ExtensionsCompositeEndpointDataSource.cs
jviau: I agree that could be used here with a bit more effort. However, this class is a modified copy of [CompositeEndpointDataSource](https://github.com/dotnet/aspnetcore/blob/main/src/Http/Routing/src/CompositeEndpointDataSource.cs) from AspNetCore, and this change token pattern comes from there. Given this code is already proven by AspNetCore, I would rather not alter it more than I have to.

daviburg (approved):

liliankasem (approved):

PR #9297, opened 2023-05-31
fixing test that was crashing ci

brettsam:
A test was inadvertently trying to start a non-existent app due to partial mocking. This is now mocking out the process startup so the test won't fail with a crashing test process.


liliankasem (approved):

mathewc (approved):

brettsam:
actually completed this with #9285

PR #9304, opened 2023-06-01
Fixing key defaulting regression

mathewc:
Fixing a key ordering issue introduced by [this commit](https://github.com/Azure/azure-functions-host/commit/80f5ebcaba334a79ab049e2fe276dc15f257e554#diff-b407624f13720c64f179ca0ee3f592dc5317bd6e880e69431fdf88c397ee535a). The issue is that before that change, our JWT token handler used the DataProtection APIs to get the signing key. Those APIs would first use any user specified AzureWebEncryptionKey value, then fall back to the Antares platform provided key (machine key). The referenced commit changed that ordering so the platform key was used BEFORE the AzureWebEncryptionKey override.

The changes in this PR ensure that we allow both signing keys.

Backport:
- V3 https://github.com/Azure/azure-functions-host/pull/9307
- V1 https://github.com/Azure/azure-functions-host/pull/9309

### DIFF
--- a/src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
+++ b/src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
@@ -1,9 +1,9 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
+using System.Collections.Generic;
 using System.Linq;
 using System.Security.Claims;
-using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Authentication;
@@ -73,15 +73,11 @@ namespace Microsoft.Extensions.DependencyInjection
 
         private static TokenValidationParameters CreateTokenValidationParameters()
         {
+            var signingKeys = SecretsUtility.GetTokenIssuerSigningKeys();
             var result = new TokenValidationParameters();
-            if (SecretsUtility.TryGetEncryptionKey(out string key))
+            if (signingKeys.Length > 0)
             {
-                // TODO: Once ScriptSettingsManager is gone, Audience and Issuer should be pulled from configuration.
-                result.IssuerSigningKeys = new SecurityKey[]
-                {
-                    new SymmetricSecurityKey(key.ToKeyBytes()),
-                    new SymmetricSecurityKey(Encoding.UTF8.GetBytes(key))
-                };
+                result.IssuerSigningKeys = signingKeys;
                 result.ValidateAudience = true;
                 result.ValidateIssuer = true;
                 result.ValidAudiences = new string[]
--- a/test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SWAEndToEndTests.cs
+++ b/test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SWAEndToEndTests.cs
@@ -1,9 +1,6 @@
-﻿using Microsoft.Azure.WebJobs.Script.WebHost;
-using Microsoft.Azure.WebJobs.Script.WebHost.Management;
-using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
-using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.DependencyInjection.Extensions;
-using Newtonsoft.Json;
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
 using System;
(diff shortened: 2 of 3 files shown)


liliankasem (approved):

src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
```diff
@@ -73,15 +75,23 @@ public static AuthenticationBuilder AddScriptJwtBearer(this AuthenticationBuilde
 
         private static TokenValidationParameters CreateTokenValidationParameters()
         {
+            List<SymmetricSecurityKey> signingKeys = new List<SymmetricSecurityKey>();
+            string defaultKey = Util.GetDefaultKeyValue();
```
mathewc: This line is what we had before - previously this was the ONLY key used.

src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
```diff
@@ -73,15 +75,23 @@ public static AuthenticationBuilder AddScriptJwtBearer(this AuthenticationBuilde
...
+            if (defaultKey != null)
+            {
+                signingKeys.Add(new SymmetricSecurityKey(defaultKey.ToKeyBytes()));
+                signingKeys.Add(new SymmetricSecurityKey(Encoding.UTF8.GetBytes(defaultKey)));
+            }
+
+            if (SecretsUtility.TryGetEncryptionKey(out string key) && key != defaultKey)
```
mathewc: In addition to the above, we want to retain the new behavior introduced by the previous changes, since when generating tokens to send to the app, Antares doesn't use AzureWebEncryptionKey, it uses WEBSITE_AUTH_ENCRYPTION_KEY. See SiteTokenHelper in the Antares codebase for details.

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SWAEndToEndTests.cs
```diff
@@ -67,7 +70,7 @@ public async Task InvokeFunction_FunctionLevel_ValidToken_Succeeds(string header
         {
             // if an admin token is passed, the function invocation succeeds
             HttpRequestMessage request = new HttpRequestMessage(HttpMethod.Get, "api/HttpTrigger-FunctionAuth?code=test");
-            string token = _fixture.Host.GenerateAdminJwtToken();
+            string token = GetSWAAdminJwtToken();
```
mathewc, line 73: After updating the test fixture such that AzureWebEncryptionKey and WEBSITE_AUTH_ENCRYPTION_KEY were different, this existing test failed when sending a token signed with AzureWebEncryptionKey.

src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
fabiocav: Can we move this from `key != defaultKey` to `string.Equals` so we're explicit about casing?

src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
fabiocav: It's a bit confusing to have this in multiple places, as opposed to all consolidated in the `TryGetEncryptionKey` method. Are you concerned about the default probing done by our Data Protection layer?

mathewc: The issue is that SecretsUtility.TryGetEncryptionKey has an existing key ordering that is used for encryption/decryption of data, outside of JWT token use. It's also used by SimpleWebTokenHelper where the TryGetEncryptionKey logic previously existed before being refactored out into SecretsUtility for reuse in https://github.com/Azure/azure-functions-host/commit/80f5ebcaba334a79ab049e2fe276dc15f257e554. For the usages in SimpleWebTokenHelper, changing the ordering would result in breaking changes on that side.

mathewc: I've factored this out into a helper with some explanatory comments

fabiocav (approved):

src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
mathewc: Fixed. Using string.Equals default case sensitive

PR #9310, opened 2023-06-02
Fixing WebHost worker process shutdown race

### LINKED ISSUE #9288: Possible race with "Did not find any initialized language workers" during shutdown
It seems that we have the order-of-operations wrong when using the `WebHostRpcWorkerChannelManager`. This gets launched in placeholder mode during the `RpcInitializationService` call. This is an `IHostedService` -- as is the `WebJobsScriptHostService` -- which is responsible for stopping the JobHost and all Listeners during shutdown.

So what can happen during shutdown of a specialized worker (or any other worker started at WebHost level):
- we shut down all channels and all worker processes here: https://github.com/Azure/azure-functions-host/blob/a9e69d28e22188db35ec4c2956e1363681da8db2/src/WebJobs.Script/Workers/Rpc/RpcInitializationService.cs#LL66C10-L66C10
- and then we stop the JobHost, which stops all listeners: https://github.com/Azure/azure-functions-host/blob/a9e69d28e22188db35ec4c2956e1363681da8db2/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs#LL454C68-L454C68

This means that any messages coming from the JobHost will continue on to the channel manager, see there are no channels, and get hit with the "Did not find any initialized language workers" error.

It's not clear to me whether this was ever correct -- I don't see any recent changes that affected it, so it may have always been this way.

brettsam:
Fixes #9288

Test failed before I submitted the fix: https://azfunc.visualstudio.com/Azure%20Functions/_build/results?buildId=133472&view=ms.vss-test-web.build-test-results-tab

### DIFF
--- a/src/WebJobs.Script/Workers/Rpc/RpcInitializationService.cs
+++ b/src/WebJobs.Script/Workers/Rpc/RpcInitializationService.cs
@@ -60,14 +60,16 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
             }
         }
 
-        public async Task StopAsync(CancellationToken cancellationToken)
+        public Task StopAsync(CancellationToken cancellationToken)
         {
-            _logger.LogDebug("Shutting down Rpc Channels Manager");
-            await _webHostRpcWorkerChannelManager.ShutdownChannelsAsync();
+            return Task.CompletedTask;
         }
 
         public async Task OuterStopAsync(CancellationToken cancellationToken)
         {
+            _logger.LogDebug("Shutting down Rpc Channels Manager");
+            await _webHostRpcWorkerChannelManager.ShutdownChannelsAsync();
+
             _logger.LogDebug("Shutting down RPC server");
 
             try
--- a/src/WebJobs.Script/Workers/Rpc/RpcInitializationService.cs
+++ b/src/WebJobs.Script/Workers/Rpc/RpcInitializationService.cs
@@ -62,7 +62,7 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
 
         public async Task StopAsync(CancellationToken cancellationToken)
         {
-            _logger.LogDebug("Shuttingdown Rpc Channels Manager");
+            _logger.LogDebug("Shutting down Rpc Channels Manager");
             await _webHostRpcWorkerChannelManager.ShutdownChannelsAsync();
         }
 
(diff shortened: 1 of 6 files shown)


src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
```diff
@@ -294,25 +295,34 @@ public Task ShutdownChannelsAsync()
                 {
                     foreach (string workerId in standbyChannels.Keys)
                     {
-                        standbyChannels[workerId]?.Task.ContinueWith(channelTask =>
```
brettsam, line 297: This was all previously fire-and-forget with the way the continuation was wired up. While it wasn't a direct cause of what we were seeing, I realized that there was no guaranteed way to wait for channels to be shut down without refactoring this a bit.

I checked all callers and some were already debouncing or fire-and-forgetting the call -- so this should not affect behavior.

src/WebJobs.Script/Workers/Rpc/RpcInitializationService.cs
```diff
@@ -60,14 +60,16 @@ public async Task StartAsync(CancellationToken cancellationToken)
...
         {
-            _logger.LogDebug("Shuttingdown Rpc Channels Manager");
-            await _webHostRpcWorkerChannelManager.ShutdownChannelsAsync();
+            return Task.CompletedTask;
         }
 
         public async Task OuterStopAsync(CancellationToken cancellationToken)
         {
+            _logger.LogDebug("Shutting down Rpc Channels Manager");
+            await _webHostRpcWorkerChannelManager.ShutdownChannelsAsync();
```
brettsam, line 71: this is the fix -- previously, the channels (and processes) were shutting down before the host -- now they'll wait until all that is done before shutting down

src/WebJobs.Script/Workers/Rpc/RpcInitializationService.cs
```diff
@@ -60,14 +60,16 @@ public async Task StartAsync(CancellationToken cancellationToken)
             }
         }
 
-        public async Task StopAsync(CancellationToken cancellationToken)
+        public Task StopAsync(CancellationToken cancellationToken)
         {
-            _logger.LogDebug("Shuttingdown Rpc Channels Manager");
-            await _webHostRpcWorkerChannelManager.ShutdownChannelsAsync();
+            return Task.CompletedTask;
```
liliankasem, line 65: Just curious, this method isn't really doing anything anymore - does it need to exist?

liliankasem (approved):

src/WebJobs.Script/Workers/Rpc/RpcInitializationService.cs
brettsam: Yeah -- part of the interface, so it needs to be there.

kshyju (approved):

PR #9325, opened 2023-06-10
Update valid audience for legion applications during specialization

yojagad:

Expected valid audience modified based on SKU and based on whether it's during specialization. For legion, there will not be a `WEBSITE_SITE_NAME` populated while in placeholder mode, and therefore no sense in taking dependency on it.

Instead, it will look at `WEBSITE_POD_NAME`.

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
+++ b/src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
@@ -43,7 +43,7 @@ namespace Microsoft.Extensions.DependencyInjection
                         // TODO: DI (FACAVAL) This will be fixed once the permanent fix is in place
                         if (_specialized == 0 && !SystemEnvironment.Instance.IsPlaceholderModeEnabled() && Interlocked.CompareExchange(ref _specialized, 1, 0) == 0)
                         {
-                            o.TokenValidationParameters = CreateTokenValidationParameters(GetValidAudiences());
+                            o.TokenValidationParameters = CreateTokenValidationParameters();
                         }
 
                         return Task.CompletedTask;
@@ -61,7 +61,7 @@ namespace Microsoft.Extensions.DependencyInjection
                     }
                 };
 
-                o.TokenValidationParameters = CreateTokenValidationParameters(GetValidAudiences());
+                o.TokenValidationParameters = CreateTokenValidationParameters();
 
                 // TODO: DI (FACAVAL) Remove this once the work above is completed.
                 if (!SystemEnvironment.Instance.IsPlaceholderModeEnabled())
@@ -89,7 +89,7 @@ namespace Microsoft.Extensions.DependencyInjection
             };
         }
 
-        private static TokenValidationParameters CreateTokenValidationParameters(string[] validAudiences)
+        public static TokenValidationParameters CreateTokenValidationParameters()
         {
             var signingKeys = SecretsUtility.GetTokenIssuerSigningKeys();
             var result = new TokenValidationParameters();
@@ -98,7 +98,7 @@ namespace Microsoft.Extensions.DependencyInjection
                 result.IssuerSigningKeys = signingKeys;
                 result.ValidateAudience = true;
                 result.ValidateIssuer = true;
-                result.ValidAudiences = validAudiences;
+                result.ValidAudiences = GetValidAudiences();
                 result.ValidIssuers = new string[]
                 {
                     AppServiceCoreUri,
--- a/src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
+++ b/src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
@@ -26,52 +26,70 @@ namespace Microsoft.Extensions.DependencyInjection
 
         public static AuthenticationBuilder AddScriptJwtBearer(this AuthenticationBuilder builder)
             => builder.AddJwtBearer(o =>
+            {
+                o.Events = new JwtBearerEvents()
+                {
+                    OnMessageReceived = c =>
+                    {
+                        // By default, tokens are passed via the standard Authorization Bearer header. However we also support
+                        // passing tokens via the x-ms-site-token header.
+                        if (c.Request.Headers.TryGetValue(ScriptConstants.SiteTokenHeaderName, out StringValues values))
                         {
-                            o.Events = new JwtBearerEvents()
-                            {
-                                OnMessageReceived = c =>
-                                {
-                                    // By default, tokens are passed via the standard Authorization Bearer header. However we also support
-                                    // passing tokens via the x-ms-site-token header.
-                                    if (c.Request.Headers.TryGetValue(ScriptConstants.SiteTokenHeaderName, out StringValues values))
-                                    {
-                                        // the token we set here will be the one used - Authorization header won't be checked.
-                                        c.Token = values.FirstOrDefault();
-                                    }
+                            // the token we set here will be the one used - Authorization header won't be checked.
+                            c.Token = values.FirstOrDefault();
+                        }
 
-                                    // Temporary: Tactical fix to address specialization issues. This should likely be moved to a token validator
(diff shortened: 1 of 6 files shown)


src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -312,7 +312,8 @@ public static bool IsLinuxConsumptionOnAtlas(this IEnvironment environment)
         public static bool IsLinuxConsumptionOnLegion(this IEnvironment environment)
         {
             return !environment.IsAppService() &&
-                   !string.IsNullOrEmpty(environment.GetEnvironmentVariable(ContainerName)) &&
```
mathewc, line 315: There are existing tests for this helper in EnvironmentTests that should be updated to test this new behavior #Resolved

src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
```diff
@@ -26,52 +26,70 @@ public static class ScriptJwtBearerExtensions
...
+                if (!SystemEnvironment.Instance.IsPlaceholderModeEnabled())
+                {
+                    // We're not in standby mode, so flag as specialized
+                    _specialized = 1;
+                }
+            });
+
+        private static string[] GetValidAudiences()
+        {
+            if (SystemEnvironment.Instance.IsPlaceholderModeEnabled() &&
```
mathewc, line 76: There's a check on line 44 that tests whether we're specialized. I think you should factor that into a private method IsSpecialized() that can be used in both places, so we're doing the check consistently. #Resolved

src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
```diff
@@ -26,52 +26,70 @@ public static class ScriptJwtBearerExtensions
...
-                        });
+            return new string[]
+            {
+                string.Format(SiteAzureFunctionsUriFormat, ScriptSettingsManager.Instance.GetSetting(AzureWebsiteName)),
+                string.Format(SiteUriFormat, ScriptSettingsManager.Instance.GetSetting(AzureWebsiteName))
+            };
+        }
 
-        private static TokenValidationParameters CreateTokenValidationParameters()
+        private static TokenValidationParameters CreateTokenValidationParameters(string[] validAudiences)
```
mathewc: Rather than pass this in as a param, recommend just in-lining it in this method. #Resolved

src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
```diff
@@ -26,52 +26,70 @@ public static class ScriptJwtBearerExtensions
...
+            });
+
+        private static string[] GetValidAudiences()
+        {
+            if (SystemEnvironment.Instance.IsPlaceholderModeEnabled() &&
+                SystemEnvironment.Instance.IsLinuxConsumptionOnLegion())
+            {
+                return new string[]
+                {
+                    ScriptSettingsManager.Instance.GetSetting(WebsitePodName)
```
mathewc, line 81: Hmmm, wondering if we shouldn't use the SiteUriFormat as below in this scenario for consistency. Using this pod name instead of the site name ""https://<podname>.azurewebsites.net". #Resolved

src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
yojagad: `Line 44` looks like some special case tactical fix. 

`_specialized == 0` --> Not specialized according to this method

`!SystemEnvironment.Instance.IsPlaceholderModeEnabled()` --> SystemEnvironment says it's specialized though.

I don't think it would make sense to mix them both in a single method :D Also mixing them both in a single method seems very risky for future too.

yojagad: Comment makes it obvious too

`// Temporary: Tactical fix to address specialization issues. This should likely be moved to a token validator`

src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
mathewc: Also, it'd be ideal if you added a new unit test suite ScriptJwtBearerExtensionsTests to test this static method (make it internal for testing). That'd let you test this new conditional specialization logic. Looks like there wasn't an existing suite for you to add tests to, because previously the logic was pretty static.

src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
mathewc: Yeah, ok, makes sense.

src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
yojagad: I thought about this too. Althought I didn't have too much of a preference, I settled for this because

- No need to do `string.Format` when we don't have to.
- Other containers (existing and new) can just take things from `WEBSITE_POD_NAME`.
-  Simplifies things both in terms of implementation and explanation. (Imagine having to explain , why do you suffix ".azurewebsites.net" to `WEBSITE_POD_NAME` when it's not a real site? The answer is obvious to us but not everyone else..)

src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
yojagad: Made this public and added some tests.

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
yojagad: Updated `EnvironmentExtensionsTests`

liliankasem (approved):

src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
```diff
@@ -31,6 +31,7 @@ public static class EnvironmentSettingNames
         public const string FunctionsExtensionVersion = "FUNCTIONS_EXTENSION_VERSION";
         public const string FunctionWorkerRuntime = "FUNCTIONS_WORKER_RUNTIME";
         public const string ContainerName = "CONTAINER_NAME";
+        public const string WebsitePodName = "WEBSITE_POD_NAME";
```
mathewc, line 34: For making the switch for CV2 from CONTAINER_NAME to WEBSITE_POD_NAME, are there other helpers we need to update other than just IsLinuxConsumptionOnLegion? Looking at all the usages of ContainerName currently, I see other places like GetInstanceId, etc. Will we be sending both, or will we only send one? #Resolved

src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
mathewc: Use of a non URI value is to spec https://www.rfc-editor.org/rfc/rfc7519 as you're doing is to spec - the value just needs to be a StringOrUri. You're arguing that because the container isn't specialized yet, an App Service site doesn't really exist - it's just a non-externally addressable container. Fine I guess.

mathewc (approved):

src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
yojagad: We'll be sending both until we can completely decouple from `Container_Name`.

PR #9331, opened 2023-06-14
Patch Durable Extension version check

davidmrdavid:
Currently, when the host checks the durable extension version, it makes two errors:
1) It assumes all apps using extension bundles are using DF Extension V1.x
2) It assumes all apps using C# isolated are using DF Extension V1.x because it cannot find `extensions.json` in the `/bin` directory

This PR fixes this issue in the following ways:

For (1), given that only bundles V2 onwards is supported in Functions V4, we assume accordingly DF Extension V2.x is used.
For (2), we fix the location where `extensions.csproj` is looked to also include its location in isolated .NET apps.

**Implication:** During the SyncTriggers call, the Host will provide different metadata depending on whether DF Extension V2.x or V1.x is used. When testing target-based scaling, the Host was previously providing DF Extension V1.x-style metadata for all out of process apps. Now, we should be receiving the expected SyncTriggers metadata.

### Pull request checklist

### Additional information

Additional PR information


davidmrdavid:
FYI @pragnagopa, I suppose we can update the images after merge. Thank you

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -548,16 +548,26 @@ private async Task<string> GetDurableMajorVersionAsync(JObject hostJson, ScriptJ
...
             if (isUsingBundles)
             {
-                // TODO: As of 2019-12-12, there are no extension bundles for version 2.x of Durable.
-                // This may change in the future.
-                return "1";
+                // From Functions runtime V4 onwards, only bundles >= V2.x is supported, which implies the app should be using DF V2 or greater.
+                return "2";
             }
 
+            // There's two places where `extensions.json` may be located
```
fabiocav: The flow here should match what is used for the extension probing, using the same order of precedence:

1 - `bin` folder
2 - script root
3 - System path (`.azurefunctions`) which is the constant `AzureFunctionsSystemDirectoryName` inside of the script root path.

You can reference the logic here:
- https://github.com/Azure/azure-functions-host/blob/66731fa6715f02610c3d0842c56488e540e695e3/src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs#L128-L154

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -548,16 +548,26 @@ private async Task<string> GetDurableMajorVersionAsync(JObject hostJson, ScriptJ
             bool isUsingBundles = hostJson != null && hostJson.TryGetValue("extensionBundle", StringComparison.OrdinalIgnoreCase, out _);
             if (isUsingBundles)
             {
-                // TODO: As of 2019-12-12, there are no extension bundles for version 2.x of Durable.
-                // This may change in the future.
-                return "1";
+                // From Functions runtime V4 onwards, only bundles >= V2.x is supported, which implies the app should be using DF V2 or greater.
+                return "2";
```
fabiocav: Should we be getting the actual version from the metadata instead of hardcoding this here?

fabiocav (changes requested):

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
davidmrdavid: @fabiocav: you mean from the extension bundles version range? I thought of that, but worried that parsing the version range via regex might be dangerous and more error prone than it's worth, so I opted for just going with bundles V1 not being supported for Functions V4. Happy to be challenged on this though.

Is there code in the host that can already parse the bundles range? If so, I can re-use that

fabiocav: Here, do you care about the bundle or DF major version?

davidmrdavid: DF major version. Maybe I misunderstood your suggestion. Can you please elaborate?

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
davidmrdavid: Ah, good to know. I've added the check for the system path by referencing that logic. Let me know if that looks good to you

test/WebJobs.Script.Tests.Integration/Management/FunctionsSyncManagerTests.cs
```diff
@@ -474,7 +474,7 @@ public async Task TrySyncTriggers_NoDurableTaskHub_UsesBundles_V1DefaultsPosted(
                 Assert.True(syncResult.Success, "SyncTriggers should return success true");
                 Assert.True(string.IsNullOrEmpty(syncResult.Error), "Error should be null or empty");
 
-                VerifyResultWithCacheOn(expectedTaskHub: null, connection: null);
+                VerifyResultWithCacheOn(expectedTaskHub: "TestHubValue", connection: null);
```
davidmrdavid, line 477: This test assumed that bundles meant we're using DF V1, so I changed it accordingly.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -545,19 +546,60 @@ private async Task<DurableConfig> ReadDurableTaskConfig()
...
+            else
             {
-                return null;
+                if (isUsingBundles)
+                {
+                    // From Functions runtime V4 onwards, only bundles >= V2.x is supported, which implies the app should be using DF V2 or greater.
+                    return "2";
+                }
+
+                // There's 3 directories where `extensions.json` may be located
```
mathewc: Why are we duplicating this logic from ScriptStartupTypeLocator? They need to be kept in sync right - we want to use the same logic. Can't we factor the existing code into a shared helper that you can simply call here?

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -545,19 +546,60 @@ private async Task<DurableConfig> ReadDurableTaskConfig()
         // This is a stopgap approach to get the Durable extension version. It duplicates some logic in ExtensionManager.cs.
         private async Task<string> GetDurableMajorVersionAsync(JObject hostJson, ScriptJobHostOptions hostOptions)
```
mathewc, line 549: What is the plan for addressing https://github.com/Azure/azure-functions-host/issues/5390? We really need to get this logic out of the host. Durable is the only extension that has custom logic on this path.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -545,19 +546,60 @@ private async Task<DurableConfig> ReadDurableTaskConfig()
...
         {
+            string metadataFilePath;
             bool isUsingBundles = hostJson != null && hostJson.TryGetValue("extensionBundle", StringComparison.OrdinalIgnoreCase, out _);
-            if (isUsingBundles)
+            // This feature flag controls whether to opt out of the SyncTrigger metadata fix for OOProc DF apps from https://github.com/Azure/azure-functions-host/pull/9331
+            if (FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagEnableLegacyDurableVersionCheck))
```
mathewc, line 554: The new behavior will be on by default. Will this break existing apps that are currently running and potentially working around the existing behavior via config?

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
davidmrdavid: No plan yet, I was not aware of https://github.com/Azure/azure-functions-host/issues/5390 until today. I've replied to that thread to continue that conversation, but I would like to keep that separate from this PR. That's beyond the scope of this bug fix, I think.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
davidmrdavid: Our understanding, from discussing with @pragnagopa, is that it should not break existing apps. The consumer of SyncTriggers is the scale controller and ScaleController V2 can work both with the old buggy behavior (in that case it compensates for the missing SyncTriggers metadata) and with the new behavior introduced in this PR (in which case it doesn't need to compensate at all).

We've also added the opt-out flag `FeatureFlagEnableLegacyDurableVersionCheck` as a safety mechanism to opt out of this change in case we're wrong about this assessment.

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
```diff
@@ -72,6 +72,39 @@ public Type[] GetStartupTypes()
 
         internal bool HasExternalConfigurationStartups() => _startupTypes.Value.Any(p => typeof(IWebJobsConfigurationStartup).IsAssignableFrom(p));
 
+        public static string FindExtensionsMetadataPath(string rootScriptPath, out string baseProbingPath)
+        {
+            baseProbingPath = null;
```
davidmrdavid: Just refactoring this so it can be -reused as per: https://github.com/Azure/azure-functions-host/pull/9331#discussion_r1232543085

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
davidmrdavid: Good idea, I refactor the logic in ScriptStartupTypeLocator in my latest commit to do just this.

davidmrdavid: You can see the relevant diff here: https://github.com/Azure/azure-functions-host/pull/9331/commits/8b1fcd15ee4fa1cf287fcd16105df9f541e7d6b9

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
davidmrdavid: To clarify: this bug fix is motivated by ScV3 which does not have this compensating logic. That's how we discovered this bug

mathewc (approved):

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
fabiocav: That would be available (as the actual version) with the extensions resolved by the the bundle. I don't think we need to actually block on that, but lookin at the actual extension type would give you the exact major version. We already have code in the script startup type locator to parse the extensions once you get what is resolved by the bundles. 

Again, not a blocker.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
fabiocav: Thanks, that's what I was hoping for with my comment [here](https://github.com/Azure/azure-functions-host/pull/9331#discussion_r1230306366)

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
```diff
@@ -72,6 +72,39 @@ public Type[] GetStartupTypes()
 
         internal bool HasExternalConfigurationStartups() => _startupTypes.Value.Any(p => typeof(IWebJobsConfigurationStartup).IsAssignableFrom(p));
 
+        public static string FindExtensionsMetadataPath(string rootScriptPath, out string baseProbingPath)
```
fabiocav: nit (super nit): Can we rename this to `ResolveExtensionsMetadataPath`?

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -545,19 +547,40 @@ private async Task<DurableConfig> ReadDurableTaskConfig()
...
             {
-                return null;
+                if (isUsingBundles)
+                {
+                    // From Functions runtime V4 onwards, only bundles >= V2.x is supported, which implies the app should be using DF V2 or greater.
+                    return "2";
+                }
+
+                // If the app is not using bundles, we look for extensions.json
+                string metadataDirectoryPath = ScriptStartupTypeLocator.FindExtensionsMetadataPath(hostOptions.RootScriptPath, out _);
```
fabiocav: Questioning the use of the static method on `ScriptStartupTypeLocator` from here. The alternative would be to move it to `Utility` (or some other more appropriate type).

OK without the change too.

fabiocav (approved):

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
davidmrdavid: How about `TryResolveExtensionsMetadataPath` :) . I generalized this method a little bit now that it's been moved to `Utility.cs` so that it returns a `bool` indicating whether or not it's search was successful. Happy to be challenged on this

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
davidmrdavid: I had similar thoughts but I was trying to minimize changes unless explicitly requested :). Moved to `Utility`

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -447,6 +449,7 @@ internal async Task<IEnumerable<JObject>> GetFunctionTriggers(IEnumerable<Functi
                 .WhenAll())
                 .Where(t => t != null);
 
+            // TODO: We should remove extension-specific logic from the Host. See: https://github.com/Azure/azure-functions-host/issues/5390
```
davidmrdavid, line 452: comment requested by @mathewc

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
davidmrdavid: That would be a better approach, so I'd be happy to refine this logic later. However, as it's not a blocker, and given that Host v4 already fails fast when bundles V1 is loaded (just validated this locally), I think I'd rather keep this change as-is for simplicity and expediency's sake.

If all goes according to plan, we should be removing this DF specific logic from the Host anyways, so hopefully this is a short term solution for a codepath that is "soon" to be removed :)

fabiocav (approved):

PR #9356, opened 2023-06-28
bumping nuget project model reference

brettsam:


test/WebJobs.Script.Tests/Microsoft.Azure.WebJobs.Script.WebHost.deps.json
```diff
@@ -21,23 +21,24 @@
...
           "Microsoft.SourceLink.GitHub": "1.0.0",
           "StyleCop.Analyzers": "1.2.0-beta.435",
           "System.IdentityModel.Tokens.Jwt": "6.24.0",
           "System.Net.NameResolution": "4.3.0",
+          "System.Private.Uri": "4.3.2",
           "System.Security.Cryptography.Xml": "4.7.1",
           "System.Text.RegularExpressions": "4.3.1",
-          "Microsoft.Azure.WebJobs.Script.Reference": "4.17.0.0",
-          "Microsoft.Azure.WebJobs.Script.Grpc.Reference": "4.17.0.0"
+          "Microsoft.Azure.WebJobs.Script.Reference": "4.24.0.0",
```
surgupta-msft, line 40: One question for my understanding - where are all these changes in this file coming from? is this file auto-generated with the one line change we made in [WebJobs.Script.csproj](https://github.com/Azure/azure-functions-host/pull/9356/files#diff-6a49576770202827579aca50b0a820e13b37b1737a8e35a8b0da08d3f37a254d)?

surgupta-msft (approved):

test/WebJobs.Script.Tests/Microsoft.Azure.WebJobs.Script.WebHost.deps.json
brettsam: Yes, this is a little confusing. We have a test that compares a checked-in-version of deps.json (this one) with one generated during the build process. It will compare all assembly versions that we reference and the test will fail if one of them has been updated unexpectedly so that we have to manually review it to ensure we're not going to break anything. This was put in place b/c assembly bumps can break customers using in-proc -- specifically if we jump major versions. So we want to really avoid that and make sure it's not an accident.

However, there's a bunch of assemblies that change a lot (every release) and we don't really need to worry about them -- so they're excluded. 

So you get something like this -- where now, a few versions have gone by with no assembly bumps and this one now pulls in all of the latest ones that we typically ignore.

You can see that list of exclusions here -- https://github.com/Azure/azure-functions-host/blob/dev/test/WebJobs.Script.Tests/DependencyTests.cs#L18

surgupta-msft: makes sense!

brettsam: Oh and one other thing -- the one-line change in the csproj can cause ripples in the deps.json file.

This is a good example, where that Nuget.ProjectModel nuget package has a bunch of dependencies. So updating them will also update all of those dependency versions.

The deps.json file shows you the fully resolved dependencies and versions of the entire project, which is why we use it as our "source of truth" for this test.

liliankasem (approved):

test/WebJobs.Script.Tests/Microsoft.Azure.WebJobs.Script.WebHost.deps.json
```diff
@@ -4608,6 +4624,13 @@
       "path": "system.objectmodel/4.0.12",
       "hashPath": "system.objectmodel.4.0.12.nupkg.sha512"
     },
+    "System.Private.Uri/4.3.2": {
```
brettsam, line 4627: System.Private.Uri is already listed as `minorMatchOrLower`... but it's not clear what version we were previously referencing. It appears that we weren't referencing?

@fabiocav -- any ideas here? Should we move this to `private` if we weren't referencing it previously?

fabiocav: Followed up with @brettsam offline. This is fine as-is, as it's already a dependency of the host (and a framework reference), but now being overridden with the package reference.

fabiocav (approved):

PR #9357, opened 2023-06-29
cleaning up test env vars

brettsam:
PR #9325 introduced a test that does not reset environment variables when they get set. This can lead to subsequent tests failing because the env vars haven't been cleaned up.

### DIFF
--- a/test/WebJobs.Script.Tests/Extensions/ScriptJwtBearerExtensionsTests.cs
+++ b/test/WebJobs.Script.Tests/Extensions/ScriptJwtBearerExtensionsTests.cs
@@ -6,7 +6,6 @@ using System.Collections.Generic;
 using System.Linq;
 using Microsoft.Azure.WebJobs.Script.Config;
 using Microsoft.Extensions.DependencyInjection;
-using NuGet.Configuration;
 using Xunit;
 using static Microsoft.Azure.WebJobs.Script.EnvironmentSettingNames;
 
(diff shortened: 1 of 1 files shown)


test/WebJobs.Script.Tests/Extensions/ScriptJwtBearerExtensionsTests.cs
```diff
@@ -24,22 +23,21 @@ public void CreateTokenValidationParameters_HasExpectedAudience(bool isPlacehold
             var podName = "RandomPodName";
             var containerName = "RandomContainerName";
             var siteName = "RandomSiteName";
-            ScriptSettingsManager.Instance.SetSetting(AzureWebsiteName, siteName);
```
brettsam, line 27: behind-the-scenes, this is setting env vars that are not being cleaned up.

test/WebJobs.Script.Tests/Extensions/ScriptJwtBearerExtensionsTests.cs
```diff
@@ -54,17 +52,13 @@ public void CreateTokenValidationParameters_HasExpectedAudience(bool isPlacehold
...
-                    ScriptSettingsManager.Instance.GetSetting(ContainerName)
-                };
+                expectedWithContainerName = new string[] { containerName };
                 testData[AzureWebsiteInstanceId] = string.Empty;
                 testData[ContainerName] = containerName;
             }
 
             testData[ContainerEncryptionKey] = Convert.ToBase64String(TestHelpers.GenerateKeyBytes());
-            using (new TestScopedEnvironmentVariable(testData))
+            using (new TestScopedSettings(ScriptSettingsManager.Instance, testData))
```
brettsam, line 61: Wrapping it all in here ensures that it'll bet cleaned up on disposal.

brettsam:
@yojagad / @Tonewall -- can you double-check this looks okay? I basically just set up the test var dictionary then apply it wrapped in the `using` rather than doing it first...

yojagad (approved):
:shipit:

liliankasem (approved):

PR #9367, opened 2023-07-05
Apply capabilities correctly on reload

### LINKED ISSUE #9366: EnvironmentReloadResponse not updating all Capabilities
With #9020 we added the ability for a worker to send updated capabilities with the EnvironmentReloadResponse. However, the code that handles capabilities in the InitResponse does not do the same things in the EnvReloadResponse. We need to get these two matching.

Specifically, I don't believe that the memory mapping feature nor the http proxying feature will work if they are updated in the EnvReloadResponse.

satvu:

resolves #9366 

### Pull request checklist

### Additional Information 

7/6 blocked on design and standardization of how capabilities received in placeholder mode and/or after specialization are combined and applied.

### DIFF
--- a/src/WebJobs.Script.Grpc/Channel/GrpcCapabilities.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcCapabilities.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using Microsoft.Azure.WebJobs.Script.Grpc.Channel;
 using Microsoft.Extensions.Logging;
 
 namespace Microsoft.Azure.WebJobs.Script.Grpc
--- a/src/WebJobs.Script.Grpc/Channel/GrpcCapabilities.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcCapabilities.cs
@@ -33,28 +33,29 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
                 return;
             }
 
-            if (strategy == GrpcCapabilitiesUpdateStrategy.Merge)
-            {
-                _logger.LogDebug($"Updating capabilities using merge strategy: {capabilities.ToString()}");
+            _logger.LogDebug($"Updating capabilities using {strategy} strategy. Current: {_capabilities} Incoming: {capabilities}");
 
-                foreach (KeyValuePair<string, string> capability in capabilities)
-                {
-                    UpdateCapability(capability);
-                }
-            }
-            else if (strategy == GrpcCapabilitiesUpdateStrategy.Replace)
+            switch (strategy)
             {
-                _logger.LogDebug($"Updating capabilities using replace strategy: {capabilities.ToString()}");
-                _capabilities.Clear();
-                foreach (KeyValuePair<string, string> capability in capabilities)
-                {
-                    UpdateCapability(capability);
-                }
+                case GrpcCapabilitiesUpdateStrategy.Merge:
+                    foreach (KeyValuePair<string, string> capability in capabilities)
+                    {
+                        UpdateCapability(capability);
+                    }
+                    break;
+
+                case GrpcCapabilitiesUpdateStrategy.Replace:
+                    _capabilities.Clear();
+                    foreach (KeyValuePair<string, string> capability in capabilities)
+                    {
+                        UpdateCapability(capability);
+                    }
+                    break;
+                default:
+                    throw new InvalidOperationException($"Did not recognize the capability update strategy {strategy}.");
             }
-            else
-            {
-                throw new InvalidOperationException($"Did not recognize the capability update strategy {nameof(strategy)}.");
-            }
+
+            _logger.LogDebug($"Updated capabilities: {_capabilities}");
         }
 
         private void UpdateCapability(KeyValuePair<string, string> capability)
--- a/src/WebJobs.Script.Grpc/Channel/GrpcCapabilities.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcCapabilities.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using Microsoft.Azure.WebJobs.Script.Grpc.Channel;
 using Microsoft.Extensions.Logging;
 
 namespace Microsoft.Azure.WebJobs.Script.Grpc
@@ -26,18 +27,34 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
             return null;
         }
 
-        public void UpdateCapabilities(IDictionary<string, string> capabilities)
+        public void UpdateCapabilities(IDictionary<string, string> capabilities, GrpcCapabilitiesUpdateStrategy strategy)
         {
             if (capabilities == null)
             {
                 return;
             }
 
-            _logger.LogDebug($"Updating capabilities: {capabilities.ToString()}");
-
-            foreach (KeyValuePair<string, string> capability in capabilities)
+            if (strategy == GrpcCapabilitiesUpdateStrategy.Merge)
             {
-                UpdateCapability(capability);
+                _logger.LogDebug($"Updating capabilities using merge strategy: {capabilities.ToString()}");
+
+                foreach (KeyValuePair<string, string> capability in capabilities)
+                {
+                    UpdateCapability(capability);
+                }
+            }
+            else if (strategy == GrpcCapabilitiesUpdateStrategy.Replace)
+            {
+                _logger.LogDebug($"Updating capabilities using replace strategy: {capabilities.ToString()}");
+                _capabilities.Clear();
+                foreach (KeyValuePair<string, string> capability in capabilities)
+                {
+                    UpdateCapability(capability);
+                }
+            }
+            else
+            {
+                throw new InvalidOperationException($"Did not recognize the capability update strategy {nameof(strategy)}.");
             }
         }
 
--- a/src/WebJobs.Script.Grpc/Channel/GrpcCapabilities.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcCapabilities.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Text.Json;
 using Microsoft.Extensions.Logging;
 
 namespace Microsoft.Azure.WebJobs.Script.Grpc
@@ -33,7 +34,7 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
                 return;
             }
 
-            _logger.LogDebug($"Updating capabilities using {strategy} strategy. Current: {_capabilities} Incoming: {capabilities}");
+            _logger.LogDebug("Updating capabilities using {strategy} strategy. Current: {_capabilities} Incoming: {capabilities}", strategy, JsonSerializer.Serialize(_capabilities), JsonSerializer.Serialize(capabilities));
 
             switch (strategy)
             {
@@ -55,7 +56,7 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
                     throw new InvalidOperationException($"Did not recognize the capability update strategy {strategy}.");
             }
 
-            _logger.LogDebug($"Updated capabilities: {_capabilities}");
+            _logger.LogDebug("Updated capabilities: {capabilities}", JsonSerializer.Serialize(_capabilities));
         }
 
         private void UpdateCapability(KeyValuePair<string, string> capability)
--- a/src/WebJobs.Script.Grpc/Channel/GrpcCapabilitiesUpdateStrategy.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcCapabilitiesUpdateStrategy.cs
@@ -5,7 +5,11 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
 {
     internal enum GrpcCapabilitiesUpdateStrategy
     {
+        // overwrites existing values and appends new ones
+        // ex. worker init: {A: foo, B: bar} + env reload: {A:foo, B: foo, C: foo} -> {A: foo, B: foo, C: foo}
         Merge,
+        // existing capabilities are cleared and new capabilities are applied
+        // ex. worker init: {A: foo, B: bar} + env reload: {A:foo, C: foo} -> {A: foo, C: foo}
         Replace
     }
 }
--- a/src/WebJobs.Script.Grpc/Channel/GrpcCapabilitiesUpdateStrategy.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcCapabilitiesUpdateStrategy.cs
@@ -1,7 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
-namespace Microsoft.Azure.WebJobs.Script.Grpc.Channel
+namespace Microsoft.Azure.WebJobs.Script.Grpc
 {
     internal enum GrpcCapabilitiesUpdateStrategy
     {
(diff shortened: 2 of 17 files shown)


src/WebJobs.Script.Grpc/Channel/GrpcCapabilities.cs
```diff
@@ -26,18 +26,34 @@ public string GetCapabilityState(string capability)
...
         {
             if (capabilities == null)
             {
                 return;
             }
 
-            _logger.LogDebug($"Updating capabilities: {capabilities.ToString()}");
+            if (strategy == GrpcCapabilitiesUpdateStrategy.Merge)
+            {
+                _logger.LogDebug($"Updating capabilities using merge strategy: {capabilities.ToString()}");
```
brettsam: Couple of nits about the logging:

1. You can do this with only one message by using `strategy.ToString()` and moving the log up before we enter the if/else (makes it nicer if we add new strategies)
2. Can we be more explicit here? With something like `"Updating capabilities using {strategy} strategy. Current: {_capabilities} Incoming: {_capabilities}"`
3. You shouldn't need `ToString()` inside an interpolated string... it will do that automatically (I know you didn't add it, but let's clean it up a little :-))
4. And then we can have another message after another message `"Updated capabilities: {_capabilities.ToString()}"`

That would help debug some of this should there be issues.

src/WebJobs.Script.Grpc/Channel/GrpcCapabilities.cs
```diff
@@ -26,18 +26,34 @@ public string GetCapabilityState(string capability)
...
-        public void UpdateCapabilities(IDictionary<string, string> capabilities)
+        public void UpdateCapabilities(IDictionary<string, string> capabilities, GrpcCapabilitiesUpdateStrategy strategy)
         {
             if (capabilities == null)
             {
                 return;
             }
 
-            _logger.LogDebug($"Updating capabilities: {capabilities.ToString()}");
+            if (strategy == GrpcCapabilitiesUpdateStrategy.Merge)
```
brettsam: Can this be a switch statement instead? I think it could look a little nicer and makes it easier to add new statements if we add new strategies.

src/WebJobs.Script.Grpc/Channel/GrpcCapabilities.cs
```diff
@@ -26,18 +26,34 @@ public string GetCapabilityState(string capability)
...
+                _capabilities.Clear();
+                foreach (KeyValuePair<string, string> capability in capabilities)
+                {
+                    UpdateCapability(capability);
+                }
+            }
+            else
             {
-                UpdateCapability(capability);
+                throw new InvalidOperationException($"Did not recognize the capability update strategy {nameof(strategy)}.");
```
brettsam: You don't want `{nameof(strategy)}` here -- that'll write out "strategy". You want `{strategy}` which will write out the enum value.

src/WebJobs.Script.Grpc/Channel/GrpcCapabilities.cs
```diff
@@ -26,19 +26,36 @@ public string GetCapabilityState(string capability)
...
-        public void UpdateCapabilities(IDictionary<string, string> capabilities)
+        public void UpdateCapabilities(IDictionary<string, string> capabilities, GrpcCapabilitiesUpdateStrategy strategy)
         {
             if (capabilities == null)
             {
                 return;
             }
 
-            _logger.LogDebug($"Updating capabilities: {capabilities.ToString()}");
+            _logger.LogDebug($"Updating capabilities using {strategy} strategy. Current: {_capabilities} Incoming: {capabilities}");
```
kshyju: This log message will not print the content of the dictionary. It will print something like
 `Current: System.Collections.Generic.Dictionary2[System.String,System.String]`

src/WebJobs.Script.Grpc/Channel/GrpcCapabilitiesUpdateStrategy.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Script.Grpc
+{
+    internal enum GrpcCapabilitiesUpdateStrategy
+    {
+        Merge,
```
kshyju, line 10: nit: It may be a good idea to add the same comments you added in the proto file about the behavior.

src/WebJobs.Script.WebHost/WebJobsApplicationBuilderExtension.cs
```diff
@@ -27,6 +33,10 @@ public static IApplicationBuilder UseWebJobsScriptHost(this IApplicationBuilder
             IOptionsMonitor<HttpBodyControlOptions> httpBodyControlOptionsMonitor = builder.ApplicationServices.GetService<IOptionsMonitor<HttpBodyControlOptions>>();
             IServiceProvider serviceProvider = builder.ApplicationServices;
 
+            // Update the UseWebJobsScriptHost method with the following
+            _environment = environment;
+            _placeholderTimer = new Timer(OnTimedEvent, null, 60000, 60000);
```
kshyju: Did you mean to include this? We already have some code do this in `StandByManager` class. rite? Why do we need this change here?

test/DotNetIsolated60/DotNetIsolated60.csproj
```diff
@@ -8,10 +8,11 @@
     <FunctionsEnableWorkerIndexing>True</FunctionsEnableWorkerIndexing>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.Functions.Worker" Version="1.14.0" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker" Version="1.16.0" />
```
kshyju: 1.18.0 is the latest as of today.

test/DotNetIsolated60/DotNetIsolated60.csproj
```diff
@@ -8,10 +8,11 @@
...
     <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Http" Version="3.0.13" />
-    <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Storage" Version="4.0.4" />
-    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.10.0" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore" Version="1.0.0-preview2" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Storage" Version="5.1.2" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.11.0" />
```
kshyju: 1.12.0

src/WebJobs.Script.WebHost/WebJobsApplicationBuilderExtension.cs
satvu: Ah this was for testing some placeholder things manually - will remove it.

src/WebJobs.Script.Grpc/Channel/GrpcCapabilities.cs
liliankasem: We should also use the following logging pattern:

```suggestion
            _logger.LogDebug("Updating capabilities using {strategy} strategy. Current: {current} Incoming: {incoming}", strategy, _capabilities, capabilities);
```

src/WebJobs.Script.Grpc/Channel/GrpcCapabilities.cs
```diff
@@ -26,19 +26,36 @@ public string GetCapabilityState(string capability)
...
+                    foreach (KeyValuePair<string, string> capability in capabilities)
+                    {
+                        UpdateCapability(capability);
+                    }
+                    break;
+                default:
+                    throw new InvalidOperationException($"Did not recognize the capability update strategy {strategy}.");
             }
+
+            _logger.LogDebug($"Updated capabilities: {_capabilities}");
```
liliankasem:
```suggestion
            _logger.LogDebug("Updated capabilities: {capabilities}", _capabilities);
```

src/WebJobs.Script.WebHost/WebJobsApplicationBuilderExtension.cs
```diff
@@ -100,5 +110,15 @@ public static IApplicationBuilder UseWebJobsScriptHost(this IApplicationBuilder
 
             return builder;
         }
+
+        private static void OnTimedEvent(object state)
+        {
+            if (_environment.IsPlaceholderModeEnabled())
+            {
+                Console.WriteLine("Triggering specialization");
```
liliankasem, line 118: I don't think we should be using Console log anywhere

src/WebJobs.Script.WebHost/WebJobsApplicationBuilderExtension.cs
```diff
@@ -100,5 +110,15 @@ public static IApplicationBuilder UseWebJobsScriptHost(this IApplicationBuilder
 
             return builder;
         }
+
+        private static void OnTimedEvent(object state)
```
liliankasem: I'm guessing this is for testing only and will be removed?

satvu: Yeah this is for specializing the host locally and used for manual runs/debugging - will be removed.

test/DotNetIsolated60/HttpRequestDataFunction.cs
```diff
@@ -5,16 +5,16 @@
...
-    public class Function1
+    public class HttpRequestDataFunction
     {
         private readonly ILogger _logger;
 
-        public Function1(ILoggerFactory loggerFactory)
+        public HttpRequestDataFunction(ILoggerFactory loggerFactory)
         {
-            _logger = loggerFactory.CreateLogger<Function1>();
+            _logger = loggerFactory.CreateLogger<HttpRequestDataFunction>();
```
liliankasem: Since we're updating, let's not use the factory pattern and switch to this 

```csharp
private readonly ILogger<HttpRequestDataFunction> _logger;

        public HttpRequestDataFunction(ILogger<HttpRequestDataFunction> logger)
        {
            _logger = logger;
        }
```

test/DotNetIsolated60/HttpRequestFunction.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace DotNetIsolated60
+{
+    public class HttpRequestFunction
+    {
+        private readonly ILogger _logger;
+
+        public HttpRequestFunction(ILoggerFactory loggerFactory)
+        {
+            _logger = loggerFactory.CreateLogger<HttpRequestFunction>();
```
liliankasem: Same here, don't use loggerFactory

src/WebJobs.Script.Grpc/Channel/GrpcCapabilities.cs
```diff
@@ -26,19 +27,36 @@ public string GetCapabilityState(string capability)
...
             }
 
-            _logger.LogDebug($"Updating capabilities: {capabilities.ToString()}");
+            _logger.LogDebug("Updating capabilities using {strategy} strategy. Current: {_capabilities} Incoming: {capabilities}", strategy, JsonSerializer.Serialize(_capabilities), JsonSerializer.Serialize(capabilities));
 
-            foreach (KeyValuePair<string, string> capability in capabilities)
+            switch (strategy)
             {
-                UpdateCapability(capability);
+                case GrpcCapabilitiesUpdateStrategy.Merge:
```
surgupta-msft, line 41: Is it possible that `current` and `incoming` capabilities are exactly the same? In that case may be we can check that first and the apply merge/replace operation?

test/DotNetIsolated60/Program.cs
```diff
@@ -3,9 +3,23 @@
...
+
+if (useProxy)
+{
+    hostBuilder
+        .ConfigureFunctionsWebApplication()
+        .ConfigureGeneratedFunctionMetadataProvider();
+}
+else
+{
+    hostBuilder
```
surgupta-msft, line 19: Is this part of capabilities change?

src/WebJobs.Script.Grpc/Channel/GrpcCapabilities.cs
```diff
@@ -33,7 +34,7 @@ public void UpdateCapabilities(IDictionary<string, string> capabilities, GrpcCap
                 return;
             }
 
-            _logger.LogDebug($"Updating capabilities using {strategy} strategy. Current: {_capabilities} Incoming: {capabilities}");
+            _logger.LogDebug("Updating capabilities using {strategy} strategy. Current: {_capabilities} Incoming: {capabilities}", strategy, JsonSerializer.Serialize(_capabilities), JsonSerializer.Serialize(capabilities));
```
kshyju, line 37: Question: Do we care about logging these capability entries? We are calling `Serialize` twice here, which is in the cold start code path (env reload request)

satvu: I can definitely see this being helpful for investigations and we also use this to see how many people are using the aspnet integration feature (by checking for the capability)

cc: @brettsam

test/DotNetIsolated60/Program.cs
satvu: It's part of the test

src/WebJobs.Script.Grpc/Channel/GrpcCapabilities.cs
satvu: I don't think that case will appear often given the current use of this message. Looking forward, almost all the language worker owners have expressed an intent to use the "replace" strategy. 

Unless I'm missing a dotnet method that can do this efficiently, I think going through the key value pairs to check what's the same then only applying the operation if they are not is 2x the work of the current implementation.

src/WebJobs.Script.Grpc/Channel/GrpcCapabilities.cs
brettsam: this is okay -- there's lots of logging we do throughout the specialization flow and it's valuable. Plus, that `capabilities.ToString()` was actually serializing behind-the-scenes (`MapField<>.ToString()` serializes to json)

brettsam (approved):

test/DotNetIsolated60/DotNetIsolated60.csproj
```diff
@@ -8,10 +8,11 @@
     <FunctionsEnableWorkerIndexing>True</FunctionsEnableWorkerIndexing>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.Functions.Worker" Version="1.14.0" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker" Version="1.18.0" />
     <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Http" Version="3.0.13" />
-    <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Storage" Version="4.0.4" />
-    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.10.0" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore" Version="1.0.0-preview2" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Storage" Version="5.1.2" />
```
liliankasem:
```suggestion
    <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Storage" Version="6.0.0" />
```

liliankasem (approved):

PR #9381, opened 2023-07-10
Detect Sas expiration and emit warnings

### LINKED ISSUE #9358: Detect SAS expiration and emit warnings
Derived from issue reported in: #9113 

We should detect SAS URI expiration dates and emit warnings 30 days prior to expiration of the URI.

The warning should be emitted as a diagnostic event, linking to documentation that provides guidance to use identities instead.

We should use links to the documented guidance, like the one for the run from package connection [here](https://learn.microsoft.com/en-us/azure/azure-functions/run-functions-from-deployment-package#fetch-a-package-from-azure-blob-storage-using-a-managed-identity)

This should capture:

- `WEBSITE_RUN_FROM_PACKAGE`
- `AzureWebJobsStorage`
- Blob trigger/input URIs
- `WEBSITE_CONTENTAZUREFILECONNECTIONSTRING`

aishwaryabh:

resolves #9358 

This PR adds a diagnostic event warning if the SAS token within `WEBSITE_RUN_FROM_PACKAGE`, `AzureWebJobsStorage`, or `WEBSITE_CONTENTAZUREFILECONNECTIONSTRING` is set to expire within the next 30 days. The warning message specifies which variable has the SAS token that's set to expire and in how much time.

If the SAS token has already expired, it emits an error and specifies which variable has the expired SAS token.

PR for documentation changes: https://github.com/MicrosoftDocs/azure-docs-pr/pull/245063

### Pull request checklist

    * [x] Otherwise: Documentation issue linked to PR (working on getting documentation for the new error code page. For now it just links to the page for switching to managed identities)

### Additional information

Additional PR information


src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
```diff
@@ -0,0 +1,184 @@
...
+        private Task _analysisTask;
+        private bool _disposed;
+        private bool _analysisScheduled;
+        private ILogger _logger;
+
+        public TokenExpirationService(IEnvironment environment, WebJobsScriptHostService scriptHost, ILoggerFactory loggerFactory, IOptionsMonitor<StandbyOptions> standbyOptionsMonitor)
+        {
+            _environment = environment;
+            _scriptHost = scriptHost;
+            _loggerFactory = loggerFactory;
```
kshyju: Instead of injecting `ILoggerFactory`, you can inject `ILogger<TokenExpirationService> logger`

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
```diff
@@ -0,0 +1,184 @@
...
+            _standbyOptionsMonitor = standbyOptionsMonitor;
+            _logger = _loggerFactory.CreateLogger<TokenExpirationService>();
+        }
+
+        public Task StartAsync(CancellationToken cancellationToken)
+        {
+            try
+            {
+                if (!_environment.IsCoreTools())
```
kshyju: nit: Consider inverting if to reduce nesting (of `{`) which helps with readability.

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
```diff
@@ -0,0 +1,184 @@
...
+            EmitLogging(appSettings);
+        }
+
+        private void EmitLogging(Dictionary<string, string> appSettings)
+        {
+            foreach (var setting in appSettings)
+            {
+                var uri = setting.Value;
+                if (!string.IsNullOrEmpty(uri))
+                {
```
kshyju: nit: invert if to reduce nesting.

```csharp
if (string.IsNullOrEmpty(uri))
{
    continue;
}
```

src/WebJobs.Script/Utility.cs
```diff
@@ -879,6 +881,31 @@ public static bool IsResourceAzureBlobWithoutSas(Uri resourceUri)
...
+            {
+                return queryParams[SasTokenExpirationDate];
+            }
+
+            return string.Empty;
+        }
+
+        public static string GetSasTokenExpirationDateFromSasSignature(string azureWebJobsStorage)
+        {
+            var splitStorage = azureWebJobsStorage.Split(';');
```
kshyju: I think you can avoid the array allocation (`string.split` will allocate an array) with something like below:

```csharp
 var azureWebJobsStorageSpan = azureWebJobsStorage.AsSpan();
 var sasToken = azureWebJobsStorageSpan
                         .Slice(azureWebJobsStorageSpan.LastIndexOf(';') + 1).ToString();
```

src/WebJobs.Script/Utility.cs
```diff
@@ -879,6 +881,31 @@ public static bool IsResourceAzureBlobWithoutSas(Uri resourceUri)
...
+            var sasToken = splitStorage[splitStorage.Length - 1];
+            var queryParams = HttpUtility.ParseQueryString(sasToken);
+            if (queryParams != null && !string.IsNullOrEmpty(queryParams[SasTokenExpirationDate]))
```
kshyju: nit: I think the null check can be removed as the [method does not return null](https://learn.microsoft.com/en-us/dotnet/api/System.Web.HttpUtility.ParseQueryString?view=netstandard-2.1) ?

src/WebJobs.Script/Utility.cs
```diff
@@ -879,6 +881,31 @@ public static bool IsResourceAzureBlobWithoutSas(Uri resourceUri)
...
+                return queryParams[SasTokenExpirationDate];
+            }
+
+            return string.Empty;
+        }
+
+        public static string GetSasTokenExpirationDateFromSasSignature(string azureWebJobsStorage)
+        {
+            var splitStorage = azureWebJobsStorage.Split(';');
+            var sasToken = splitStorage[splitStorage.Length - 1];
```
kshyju: This method assumes that the "SharedAccessSignature" segment is the last item in the `;` separated string. Is it possible to get a string with different order or segments (SharedAccessSignature segment is before TableEndpoint segment) ?

src/WebJobs.Script/Utility.cs
```diff
@@ -879,6 +881,31 @@ public static bool IsResourceAzureBlobWithoutSas(Uri resourceUri)
             return true;
         }
 
+        public static string GetSasTokenExpirationDate(Uri resourceUri)
+        {
+            // Screen out URLs with an SAS token
+            var queryParams = HttpUtility.ParseQueryString(resourceUri.Query.ToLower());
```
kshyju: Do we need the "ToLower" call?

src/WebJobs.Script/Utility.cs
```diff
@@ -879,6 +881,31 @@ public static bool IsResourceAzureBlobWithoutSas(Uri resourceUri)
...
+            if (queryParams != null && !string.IsNullOrEmpty(queryParams[SasTokenExpirationDate]))
```
kshyju: I think you can remove the first null check.

`if (!string.IsNullOrEmpty(queryParams[SasTokenExpirationDate]))`

src/WebJobs.Script/Utility.cs
```diff
@@ -879,6 +881,31 @@ public static bool IsResourceAzureBlobWithoutSas(Uri resourceUri)
...
+        {
+            var splitStorage = azureWebJobsStorage.Split(';');
+            var sasToken = splitStorage[splitStorage.Length - 1];
+            var queryParams = HttpUtility.ParseQueryString(sasToken);
+            if (queryParams != null && !string.IsNullOrEmpty(queryParams[SasTokenExpirationDate]))
+            {
+                return queryParams[SasTokenExpirationDate];
+            }
+
+            return string.Empty;
```
kshyju: I think if the query string entry is not present, we should return null instead of `string.empty`.

kshyju: You may also consider parsing the value and returning a `DateTime` instance or `null` from this method since this method sounds like it returns a Date (instead of Date string).

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
```diff
@@ -0,0 +1,184 @@
...
+                    }
+                    else
+                    {
+                        sasTokenExpirationDate = Utility.GetSasTokenExpirationDate(new Uri(uri));
+                    }
+
+                    if (!string.IsNullOrEmpty(sasTokenExpirationDate))
+                    {
+                        var parsedDate = DateTime.Parse(sasTokenExpirationDate);
+                        var currentDate = DateTime.Now.Date;
```
kshyju: This can be moved outside the loop.

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
```diff
@@ -0,0 +1,184 @@
...
+                        // Log an error event if the token is already expired; otherwise log a warning event
+                        if (difference.TotalDays <= 0)
+                        {
+                            string message = string.Format(Resources.SasTokenExpiredFormat, setting.Key);
+                            DiagnosticEventLoggerExtensions.LogDiagnosticEventError(_logger, message, DiagnosticEventConstants.SasTokenExpiringErrorCode, DiagnosticEventConstants.SasTokenExpiringErrorHelpLink, new Exception(message));
+                        }
+                        if (difference.TotalDays <= 30)
+                        {
+                            string message = string.Format(Resources.SasTokenExpiringFormat, (int)difference.TotalDays, setting.Key);
+                            DiagnosticEventLoggerExtensions.LogDiagnosticEvent(_logger, Microsoft.Extensions.Logging.LogLevel.Warning, 0, DiagnosticEventConstants.SasTokenExpiringErrorCode, message, DiagnosticEventConstants.SasTokenExpiringErrorHelpLink, null);
```
kshyju: nit: consider using named argument for the `null` value (last param) to improve readability.

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
aishwaryabh: can you elaborate? I see there's if statement for checking if the env is not core tools, but not sure how I'd invert that to reduce the {} ?

aishwaryabh: ohhh i just read the next comment. I'll use continue :)

src/WebJobs.Script/Utility.cs
aishwaryabh: removed

src/WebJobs.Script/Utility.cs
aishwaryabh: Returned null! I decided to keep the return type as strings because we have to parse the queryParams[sasTokenExpirationDate] as a string explicitly first anyway before converting to a DateTime.

src/WebJobs.Script/Utility.cs
aishwaryabh: Discussed offline and the format of this string will always be the same according to this documentation: https://learn.microsoft.com/en-us/azure/storage/common/storage-configure-connection-string#create-a-connection-string-using-a-shared-access-signature

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
aishwaryabh: moved currentDate to be outside of loop

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
brettsam: I think he means to do something like this so everything isn't nested:

```csharp
if (_environment.IsCoreTools())
{
    return Task.CompletedTask;
}
```

src/WebJobs.Script/Utility.cs
```diff
@@ -879,6 +881,32 @@ public static bool IsResourceAzureBlobWithoutSas(Uri resourceUri)
...
+            var queryParams = HttpUtility.ParseQueryString(resourceUri.Query);
+            if (!string.IsNullOrEmpty(queryParams[SasTokenExpirationDate]))
+            {
+                return queryParams[SasTokenExpirationDate];
+            }
+
+            return null;
+        }
+
+        public static string GetSasTokenExpirationDateFromSasSignature(string azureWebJobsStorage)
```
brettsam: Locally (not in core tools) I run with `UseDevelopmentStorage=true` -- will this throw?

src/WebJobs.Script/Utility.cs
```diff
@@ -879,6 +881,32 @@ public static bool IsResourceAzureBlobWithoutSas(Uri resourceUri)
             return true;
         }
 
+        public static string GetSasTokenExpirationDate(Uri resourceUri)
```
brettsam: Can you combine these two new helpers together to take a string?

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
```diff
@@ -0,0 +1,184 @@
...
+                        sasTokenExpirationDate = Utility.GetSasTokenExpirationDateFromSasSignature(uri);
+                    }
+                    else
+                    {
+                        sasTokenExpirationDate = Utility.GetSasTokenExpirationDate(new Uri(uri));
+                    }
+
+                    if (!string.IsNullOrEmpty(sasTokenExpirationDate))
+                    {
+                        var parsedDate = DateTime.Parse(sasTokenExpirationDate);
```
brettsam: Should we use `TryParse()` here just in case it's malformed?

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
```diff
@@ -0,0 +1,184 @@
...
+                        {
+                            string message = string.Format(Resources.SasTokenExpiredFormat, setting.Key);
+                            DiagnosticEventLoggerExtensions.LogDiagnosticEventError(_logger, message, DiagnosticEventConstants.SasTokenExpiringErrorCode, DiagnosticEventConstants.SasTokenExpiringErrorHelpLink, new Exception(message));
+                        }
+                        if (difference.TotalDays <= 30)
+                        {
+                            string message = string.Format(Resources.SasTokenExpiringFormat, (int)difference.TotalDays, setting.Key);
+                            DiagnosticEventLoggerExtensions.LogDiagnosticEvent(_logger, Microsoft.Extensions.Logging.LogLevel.Warning, 0, DiagnosticEventConstants.SasTokenExpiringErrorCode, message, DiagnosticEventConstants.SasTokenExpiringErrorHelpLink, exception: null);
+                        }
+                    }
```
brettsam: Can we emit a log message here if we didn't find anything near-expiration, too? It'd be good to have the details even if we *did* find an expiration date. Something like "found SAS expiration date of 200 days; not warning" (but make it sound nicer :-)

That'd be good validation and we could have a detector perhaps to confirm.

src/WebJobs.Script/Utility.cs
aishwaryabh: Nope it'll return null!

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
```diff
@@ -0,0 +1,184 @@
...
+                { EnvironmentSettingNames.AzureWebsiteRunFromPackage, _environment.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebsiteRunFromPackage) },
+                { EnvironmentSettingNames.AzureFilesConnectionString, _environment.GetEnvironmentVariable(EnvironmentSettingNames.AzureFilesConnectionString) }
+            };
+
+            // We only want to parse AzureWebJobsStorage in the following format
+            // BlobEndpoint=<blob-endpoint-url>;QueueEndpoint=<queue-endpoint-url>;TableEndpoint=<table-endpoint-url>;SharedAccessSignature=<sas-token>
+            var azureWebJobsStorage = _environment.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebJobsSecretStorage);
+
+            // Check if the value starts with "BlobEndpoint=" and contains "SharedAccessSignature="
+            if (azureWebJobsStorage.StartsWith("BlobEndpoint=", StringComparison.OrdinalIgnoreCase)
```
Francisco-Gamino: Should we use  `StringComparison.OrdinalIgnoreCase` in the string contains check for `"SharedAccessSignature="` ?

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
```diff
@@ -0,0 +1,184 @@
...
+
+                        var difference = parsedDate.Subtract(currentDate);
+
+                        // Log an error event if the token is already expired; otherwise log a warning event
+                        if (difference.TotalDays <= 0)
+                        {
+                            string message = string.Format(Resources.SasTokenExpiredFormat, setting.Key);
+                            DiagnosticEventLoggerExtensions.LogDiagnosticEventError(_logger, message, DiagnosticEventConstants.SasTokenExpiringErrorCode, DiagnosticEventConstants.SasTokenExpiringErrorHelpLink, new Exception(message));
+                        }
+                        if (difference.TotalDays <= 30)
```
Francisco-Gamino: I am not sure if 30 days is enough warning time for users to react and make changes (like perform a new deployment). Would it make sense to increase this to 60 days or more? 
/cc @mattchenderson

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
Francisco-Gamino: This is a really good idea. Adding @FinVamp1 for FYI.

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
```diff
@@ -0,0 +1,184 @@
...
+            EmitLogging(appSettings);
+        }
+
+        private void EmitLogging(Dictionary<string, string> appSettings)
+        {
+            var currentDate = DateTime.Now.Date;
+            foreach (var setting in appSettings)
+            {
+                var uri = setting.Value;
+                if (!string.IsNullOrEmpty(uri))
```
Francisco-Gamino: We check the app setting value to ensure that it is not null or empty. Given that this app setting can be set by the user, should we check for white space (`string.IsNullOrWhiteSpace(...)`) as well?

test/WebJobs.Script.Tests/UtilityTests.cs
```diff
@@ -465,6 +465,24 @@ public void GetAccountNameFromDomain(string url, bool isUrlWithNoSas)
             Assert.Equal(isUrlWithNoSas, Utility.IsResourceAzureBlobWithoutSas(new Uri(url)));
         }
 
+        [Theory]
```
Francisco-Gamino, line 468: Should we add a test case that validates that the diagnostic event has been created?

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
aishwaryabh: Ok I won't prompt the user to generate a new SAS token or switch to using identities instead

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
aishwaryabh: that's fair I'll remove the check

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
liliankasem: Noticed this was resolved but still using ILoggerFactory - did the change accidentally get dropped?

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
mattchenderson: I don't have a sense for the impact. This is just informing when we start warning them, right? I'm generally a fan of communicating early, but my main concern would be actionability, how soon they might be looking at having to do this again, etc. Ultimately, that kind of depends on how long the expiration for the SAS is when it's created. I think either would be fine as a starting point, and we can readjust as we go. We could even split the difference and go 45.

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
```diff
@@ -0,0 +1,178 @@
...
+            _analysisTask = Task.Delay(TimeSpan.FromSeconds(5), _cancellationTokenSource.Token)
+               .ContinueWith(t => AnalyzeSasTokenInUri());
+        }
+
+        private void AnalyzeSasTokenInUri()
+        {
+            var jobHost = _scriptHost.GetService<IScriptJobHost>();
+
+            if (jobHost == null
+                || _cancellationTokenSource.IsCancellationRequested)
```
liliankasem: nit: this is short enough that is can on on the same line

```suggestion
            if (jobHost == null || _cancellationTokenSource.IsCancellationRequested)
```

And if `is` is available `jobHost is null` is cleaner

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
```diff
@@ -0,0 +1,178 @@
...
+            foreach (var setting in appSettings)
+            {
+                var uri = setting.Value;
+                if (!string.IsNullOrWhiteSpace(uri))
+                {
+                    var isAzureWebJobsStorage = setting.Key == EnvironmentSettingNames.AzureWebJobsSecretStorage;
+                    string sasTokenExpirationDate = Utility.GetSasTokenExpirationDate(uri, isAzureWebJobsStorage);
+
+                    if (!string.IsNullOrEmpty(sasTokenExpirationDate))
+                    {
```
liliankasem: invert this also 

```csharp
if (string.IsNullOrEmpty(sasTokenExpirationDate))
{
    continue;
}
```

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
```diff
@@ -0,0 +1,178 @@
...
+            }
+
+            return Task.CompletedTask;
+        }
+
+        public Task StopAsync(CancellationToken cancellationToken)
+        {
+            try
+            {
+                _cancellationTokenSource?.Cancel();
```
liliankasem: @brettsam @jviau would it make sense to check if _disposed is true before trying to cancel? Not sure if there is any potential for a race here that could be addressed with something like `lock (this) if (!_disposed) _cancellationTokenSource?.Cancel();` and a similar lock in Dispose()

src/WebJobs.Script/Diagnostics/DiagnosticEventConstants.cs
```diff
@@ -10,5 +10,8 @@ internal static class DiagnosticEventConstants
 
         public const string ExternalStartupErrorCode = "AZFD0005";
         public const string ExternalStartupErrorHelpLink = "https://go.microsoft.com/fwlink/?linkid=2224847";
+
+        public const string SasTokenExpiringErrorCode = "AZFD0006";
+        public const string SasTokenExpiringErrorHelpLink = "https://go.microsoft.com/fwlink/?linkid=2241431"; // TODO: update this link with new documentation page
```
liliankasem, line 15: Does this TODO need to happen after you merge the PR or before?

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
```diff
@@ -0,0 +1,178 @@
...
+        private readonly ILoggerFactory _loggerFactory;
+        private readonly IOptionsMonitor<StandbyOptions> _standbyOptionsMonitor;
+        private readonly WebJobsScriptHostService _scriptHost;
+        private CancellationTokenSource _cancellationTokenSource;
+        private Task _analysisTask;
+        private bool _disposed;
+        private bool _analysisScheduled;
+        private ILogger _logger;
+
+        public TokenExpirationService(IEnvironment environment, WebJobsScriptHostService scriptHost, ILoggerFactory loggerFactory, IOptionsMonitor<StandbyOptions> standbyOptionsMonitor)
```
liliankasem: Can we write a unit and/or integration test for this service?

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
aishwaryabh: My bad I fixed it now!

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
aishwaryabh: Changed!

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
aishwaryabh: Changed the time period to be 45 days

test/WebJobs.Script.Tests/UtilityTests.cs
aishwaryabh: Added an integration test for this!

src/WebJobs.Script/Diagnostics/DiagnosticEventConstants.cs
aishwaryabh: It can happen either way! Fabio said we have a few weeks in between anyway when this change is merged in to update the documentation link, and as of right now it is pointing to documentation to switch to using identities (so it's still relevant)

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
aishwaryabh: Added!

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
```diff
@@ -0,0 +1,167 @@
...
+        {
+            try
+            {
+                if (_environment.IsCoreTools())
+                {
+                    return Task.CompletedTask;
+                }
+                if (_standbyOptionsMonitor.CurrentValue.InStandbyMode)
+                {
+                    _standbyOptionsMonitor.OnChange(standbyOptions =>
```
jviau: this returns an `IDisposable` that would be good to track and dispose as appropriate.

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
```diff
@@ -0,0 +1,167 @@
...
+                if (_analysisTask != null && !_analysisTask.IsCompleted)
+                {
+                    _logger.LogDebug("Token expiration service stopped before analysis completion. Waiting for cancellation.");
+
+                    return _analysisTask;
+                }
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex, "Error stopping Token expiration service. Handling error and continuing.");
```
jviau: I recommend removing this catch statement. Let the caller decide what to do with a failure.

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
```diff
@@ -0,0 +1,167 @@
...
+        {
+            try
+            {
+                _cancellationTokenSource?.Cancel();
+
+                if (_analysisTask != null && !_analysisTask.IsCompleted)
+                {
+                    _logger.LogDebug("Token expiration service stopped before analysis completion. Waiting for cancellation.");
+
+                    return _analysisTask;
```
kshyju: Question: Did you mean to return the `_analysisTask` here?  I wonder that might cause some sort of race condition. Do we know what the caller of this StopAsync method (which is the framework code) does? Is it awaiting the task returned by this? Could we run into an issue where that task never reaches a Completed status for some reason, which might prevent the proper app shutodown?

Why not just return `Task.CompletedTask;`?

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
```diff
@@ -0,0 +1,167 @@
...
+            _cancellationTokenSource = new CancellationTokenSource();
+
+            _analysisTask = Task.Delay(TimeSpan.FromSeconds(5), _cancellationTokenSource.Token)
+               .ContinueWith(t => AnalyzeSasTokenInUri());
+        }
+
+        private void AnalyzeSasTokenInUri()
+        {
+            // If AzureWebJobsStorage__accountName is set, we are using identities and don't need to check for SAS token expiration
+            if (!string.IsNullOrEmpty(_environment.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebJobsSecretStorage + "__" + "accountName")))
```
kshyju: switch to string interpolation?

jviau (commented):
I wonder if instead of `IHostedService`, if there is some options object we can attach a validator to?

Also, is there any consideration for a function app which is very long lived without any changes, and the SAS becomes expired during its life span?

kshyju (approved):

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
jviau: I don't believe there is any thread-safety requirement to `IHostedService`. `.Dispose()` and `.StopAsync()` shouldn't be called together. It will be Stop -> Dispose.

With that said, there is a different issue: Dispose -> Stop will cause issues, as `CancellationTokenSource.Cancel` will throw. I typically always cancel my CTS's during dispose. But not a big concern here.

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
jviau: Switching to `BackgroundService` will solve this concern: it includes code to wait for your `ExecuteAsync` task, but then short-circuit that wait if the host force shutdown begins.

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
aishwaryabh: Disposing it at end of for loop now

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
aishwaryabh: Removed the StopAsync since `TokenExpirationService` now implements `BackgroundService`

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
aishwaryabh: I removed the `StopAsync` method now since now we are implementing BackgroundService rather than `IHostedService`. BackgroundService now handles the proper shutdown and awaiting of the executing task to complete to prevent any race condition

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
aishwaryabh: done

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
```diff
@@ -0,0 +1,148 @@
...
+
+        public TokenExpirationService(IEnvironment environment, ILogger<TokenExpirationService> logger, IOptionsMonitor<StandbyOptions> standbyOptionsMonitor)
+        {
+            _environment = environment;
+            _logger = logger;
+            if (standbyOptionsMonitor.CurrentValue.InStandbyMode)
+            {
+                _listener = standbyOptionsMonitor.OnChange(standbyOptions =>
+                {
+                    if (!standbyOptions.InStandbyMode && !_analysisScheduled)
```
jviau: `analysisScheduled` should not be necessary anymore: the `_standby` TCS will only ever be set once regardless of that value.

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
```diff
@@ -0,0 +1,148 @@
...
+            await Task.Yield(); // force a yield, so our synchronous code does not slowdown startup
+            await _standby.Task.WaitAsync(cancellationToken);
+
+            if (_environment.IsCoreTools())
+            {
+                return;
+            }
+
+            // Execute first during startup
+            ScheduleTokenExpirationCheck();
```
jviau: We are already on a background thread here - we can perform the token expiration check directly. No need to use the `_analysisTask` or `Task.Delay` continuations.

``` CSharp
while (true)
{
    AnalyzeSasTokenInUri(); // pass in cancellationToken if you want to sprinkle some .ThrowIfCancellationRequested() calls.
    await Task.Delay(TimeSpan.FromHours(1), cancellationToken);
}
```

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
```diff
@@ -0,0 +1,148 @@
...
+            while (!cancellationToken.IsCancellationRequested)
+            {
+                // check at every hour
+                await Task.Delay(TimeSpan.FromHours(1), cancellationToken)
+               .ContinueWith(t => ScheduleTokenExpirationCheck());
+            }
+
+            // Dispose of the listener for the standby options monitor
+            _listener?.Dispose();
```
jviau: Move this to the `Dispose` implementation. 

`BackgroundService` already implements `IDisposable`, so make sure your implementation contains a `base.Dispose();` call.

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
```diff
@@ -0,0 +1,148 @@
...
+            }
+            else
+            {
+                _standby.TrySetResult(null);
+            }
+        }
+
+        protected override async Task ExecuteAsync(CancellationToken cancellationToken)
+        {
+            await Task.Yield(); // force a yield, so our synchronous code does not slowdown startup
```
jviau: If you want to retain your delay of 1 second you have right now (I may be missing context on why you have it), this can become:

```suggestion
            await Task.Delay(TimeSpan.FromSeconds(1), cancellationToken);
```

Both will achieve the same result of forcing a yield

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
```diff
@@ -0,0 +1,148 @@
...
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.TokenExpiration
+{
+    internal class TokenExpirationService : BackgroundService, IDisposable
+    {
+        private readonly TaskCompletionSource<object> _standby = new();
+        private readonly IEnvironment _environment;
+        private readonly ILogger<TokenExpirationService> _logger;
+        private IDisposable _listener;
+        private CancellationTokenSource _cancellationTokenSource;
```
jviau: This is not longer needed with `BackgroundService`

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
```diff
@@ -0,0 +1,148 @@
...
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Script.Diagnostics;
+using Microsoft.Azure.WebJobs.Script.Properties;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.TokenExpiration
+{
+    internal class TokenExpirationService : BackgroundService, IDisposable
```
jviau: `BackgroundService` already implements `IDisposable`.

```suggestion
    internal class TokenExpirationService : BackgroundService
```

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
jviau, line 1: Consider putting this in a folder like `Health` instead of `TokenExpiration`. As this is in essence a health check.

jviau (changes requested):

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
aishwaryabh: I don't think we actually need the delay of one second. That was old code from copying the `AssemblyAnalysisService`

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
aishwaryabh: Should we have the while loop condition be `while(true)` or `while(!cancellationToken.IsCancellationRequested)`?

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
aishwaryabh: done!

jviau (approved):

src/WebJobs.Script.WebHost/TokenExpiration/TokenExpirationService.cs
jviau: Either is fine. We will exit the loop on cancellation one way or another: either the loop test is false, or the `Task.Delay` throws `OperationCancelledException`.

src/WebJobs.Script.WebHost/Health/TokenExpirationService.cs
```diff
@@ -0,0 +1,136 @@
...
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Script.Diagnostics;
+using Microsoft.Azure.WebJobs.Script.Properties;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Health
+{
+    internal class TokenExpirationService : BackgroundService
```
jviau, line 16: One last thing: can you add a comment here saying this is a good health check candidate if we ever start using `Microsoft.Extensions.Diagnostics.HealthChecks`.

Francisco-Gamino (approved):

src/WebJobs.Script.WebHost/Health/TokenExpirationService.cs
```diff
@@ -0,0 +1,137 @@
...
+            var azureWebJobsStorage = _environment.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebJobsSecretStorage);
+
+            // Check if the value starts with "BlobEndpoint=" and contains "SharedAccessSignature="
+            if (!string.IsNullOrEmpty(azureWebJobsStorage) && azureWebJobsStorage.StartsWith("BlobEndpoint=", StringComparison.OrdinalIgnoreCase)
+                && azureWebJobsStorage.Contains("SharedAccessSignature=", StringComparison.OrdinalIgnoreCase))
+            {
+                appSettings.Add(EnvironmentSettingNames.AzureWebJobsSecretStorage, azureWebJobsStorage);
+            }
+
+            EmitLogging(appSettings);
```
mdarefull, line 88: Are you always invoking this method regardless of the values in AzureWebsiteRunFromPackage and AzureFilesConnectionString?

Reading through the EmitLogging method it seems you are assuming the value on these settings is always an URI, but  AzureWebsiteRunFromPackage is 1, at least on my APIs

PR #9402, opened 2023-07-19
Adding RestoreRawRequestPath middlware to restore the raw request path value to current request

### LINKED ISSUE #9290: Routing fails if parameters contains slash
Hi,
I see this was reported a while back and resolved, but i cannot find the resolution.

I need to be able to pass slashes in the final segment of my route:

i.e for a department called 'A/P'for  i want to be able 

https://somefunction/api/departments/A%2FP

I set up my route like this:
Route = "departments/{dept}"

it works fine if there are no slashes (encoded as %2F) in the department, but if there is a slash it returns a 404.

How can I fix this?

kshyju:
Fixes #9290

The current behavior is caused by the App service front end decoding the encoded request path value before forwarding the request to the application (in this case functions host instance). When doing this, the front end stores the raw request path value it received and add it as a special request header.

Below are some relevant comments from other thread where this issue was discussed before.

> [crosenblatt](https://github.com/crosenblatt) commented [on Oct 14, 2022](https://github.com/dotnet/aspnetcore/issues/44461#issuecomment-1279296476)
Hi @ccromer - This behavior is present in App Service for backwards compatibility reasons. In order to get the raw bytes of the url exactly as they are received on the wire, you can use the header "X-Waws-Unencoded-Url".

and 

> [bradygaster](https://github.com/bradygaster) commented [on Mar 30, 2022](https://github.com/dotnet/aspnetcore/issues/40532#issuecomment-1083562919)
After corresponding with the service team, I learned that this is By Design: IIS FE's URL parse the request URL during proxying, so the URI will be modified (canonicalized and normalized).
>
>The work around is for the app to use the HTTP_X_WAWS_UNENCODED_URL server variable (a.k.a. X-WAWS-Unencoded-URL: header). This header contains the exact bytes as they were received from the wire.

The fix here is to use the `X-Waws-Unencoded-Url` header value which has the raw request path (un decoded) and update the request path with that so that the routing middleware can use that and route to the correct endpoint. I added a new middleware to do this. This is an opt-in feature customers needs to explicitly opt-in by setting the `WEBSITE_RESTORE_RAW_REQUEST_PATH` app setting.

[This ](https://github.com/dotnet/aspnetcore/issues/49454)issue is a minimal repro of the original issue.

### Pull request checklist

    * [x] #9403

### DIFF
--- a/src/WebJobs.Script.WebHost/Middleware/RestoreRawRequestPathMiddleware.cs
+++ b/src/WebJobs.Script.WebHost/Middleware/RestoreRawRequestPathMiddleware.cs
@@ -1,0 +1,40 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Middleware
+{
+    /// <summary>
+    /// Middleware to restore the raw request URL path received, in the current request URL.
+    /// App service front end decodes encoded strings in the request URL path.
+    /// This causes routing to fail if the route param value has %2F in it.
+    /// Refer below issues for more context:
+    /// https://github.com/dotnet/aspnetcore/issues/40532#issuecomment-1083562919
+    /// https://github.com/Azure/azure-functions-host/issues/9290.
+    /// </summary>
+    internal class RestoreRawRequestPathMiddleware
+    {
+        private readonly RequestDelegate _next;
+        internal const string UnEncodedUrlPathHeaderName = "X-Waws-Unencoded-Url";
+
+        public RestoreRawRequestPathMiddleware(RequestDelegate next)
+        {
+            _next = next ?? throw new ArgumentNullException(nameof(next));
+        }
+
+        public async Task Invoke(HttpContext context)
+        {
+            if (context.Request.Headers.TryGetValue(UnEncodedUrlPathHeaderName, out var unencodedUrlValue) &&
+                unencodedUrlValue.Count > 0)
+            {
+                context.Request.Path = new PathString(unencodedUrlValue.First());
+            }
+
+            await _next(context);
+        }
+    }
+}
--- a/src/WebJobs.Script.WebHost/WebJobsApplicationBuilderExtension.cs
+++ b/src/WebJobs.Script.WebHost/WebJobsApplicationBuilderExtension.cs
@@ -35,6 +35,20 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
 
             // Ensure the ClrOptimizationMiddleware is registered before all middleware
             builder.UseMiddleware<ClrOptimizationMiddleware>();
+
+            // Update the request URL path(which was altered by App service FE) to raw request path
+            // only if customer has opted in using the app setting.
+            builder.UseWhen(
+                _ => string.Equals(
+                    environment.GetEnvironmentVariable(EnvironmentSettingNames.RestoreRawRequestPathEnabled), "1",
+                    StringComparison.Ordinal), config =>
+            {
+                config.UseMiddleware<RestoreRawRequestPathMiddleware>();
+                // We need to re-add routing middleware after any updates to request path.
+                config.UseRouting();
+                config.UseEndpoints(endpoints => endpoints.MapControllers());
+            });
+
             builder.UseMiddleware<HttpRequestBodySizeMiddleware>();
             builder.UseMiddleware<SystemTraceMiddleware>();
             builder.UseMiddleware<HostnameFixupMiddleware>();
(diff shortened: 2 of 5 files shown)


kshyju:
Tested the build from this branch via private site extension. Created 2 sites, both has same code (changes from this PR). One has the feature enabled (`WEBSITE_RESTORE_RAW_REQUEST_PATH` app setting is present).

##### Site without the feature: 
https://shkr-9290-a-202307181948.azurewebsites.net/api/departments/cloud%2Fdevdiv - 404 - Routing failed.

##### Site with the feature enabled:
https://shkr-9290-b-202307182000.azurewebsites.net/api/departments/cloud%2Fdevdiv - 200 OK - Routing worked.

src/WebJobs.Script.WebHost/Middleware/RestoreRawRequestPathMiddleware.cs
```diff
@@ -0,0 +1,40 @@
...
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Middleware
+{
+    /// <summary>
+    /// Middleware to restore the raw request URL path received, in the current request URL.
+    /// App service front end decodes encoded strings in the request URL path.
+    /// This causes routing to fail if the route param value has %2F in it.
+    /// Refer below issues for more context:
+    /// https://github.com/dotnet/aspnetcore/issues/40532#issuecomment-1083562919
```
surgupta-msft, line 16: Is it important to mention these github issues here? I think the comment you added above explains the purpose and we can also add these links to the PR description. Though it's fine if there is specific reason for them to be here.

src/WebJobs.Script.WebHost/Middleware/RestoreRawRequestPathMiddleware.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        public RestoreRawRequestPathMiddleware(RequestDelegate next)
+        {
+            _next = next ?? throw new ArgumentNullException(nameof(next));
+        }
+
+        public async Task Invoke(HttpContext context)
+        {
+            if (context.Request.Headers.TryGetValue(UnEncodedUrlPathHeaderName, out var unencodedUrlValue) &&
+                unencodedUrlValue.Count > 0)
```
surgupta-msft, line 32: Do we need a null check for `unencodedUrlValue` here or can it never be null as this point?

src/WebJobs.Script.WebHost/WebJobsApplicationBuilderExtension.cs
```diff
@@ -35,6 +35,20 @@ public static IApplicationBuilder UseWebJobsScriptHost(this IApplicationBuilder
 
             // Ensure the ClrOptimizationMiddleware is registered before all middleware
             builder.UseMiddleware<ClrOptimizationMiddleware>();
+
+            // Update the request URL path(which was altered by App service FE) to raw request path
```
satvu: nit:
```suggestion
            // Update the request URL path (which was altered by App service FE) to raw request path
```

satvu (approved):

src/WebJobs.Script.WebHost/Middleware/RestoreRawRequestPathMiddleware.cs
liliankasem: I don't think it hurts to have additional context available via the GH issue link, might get lost otherwise

liliankasem (approved):

kshyju:
@mattchenderson Let me know how do we want to proceed forward with this change? Did we hear any thing further from the platform folks?

MitchBodmer:
Any updates here? It looks code complete but I gather you're waiting on some input from another team?

kshyju:
Closing this function level fix as we got confirmation from Antares team that this can be fixed on the platform level.

https://msazure.visualstudio.com/Antares/_workitems/edit/25263505 is the (internal) work item tracking the fix.

johan-lindqvist:
Any way for external interested parties to track the platform level fix?

kwasilka:
I'm also eager for a fix for this.  I'm experiencing the issue with API's hosted from App Services.  I need a fix for this asap

MitchBodmer:
Hey @kshyju, I just talked to the team that would be fixing this for the platform and they said they wouldn't be able to get to this until mid/late 2024 _at the earliest_. Can we revisit this PR as a stopgap and remove it whenever they get around to fixing it? I think this has a lot of value.

MitchBodmer:
Also unfortunately they have no clear way of tracking the change publicly.

kshyju:
@mattchenderson Can you advice?

mcrio:
Hi. Any way to have a workaround for this problem when hosting a DotNet web app with Azure App Services? It's a release blocker and we would never expect the azure internal proxy to _**magically**_ alter parts of the original URL.

ccromer:
Double url encode your route values as you build your url.Sent from my iPhoneOn Aug 28, 2024, at 06:44, Nik ***@***.***> wrote:﻿
Hi. Any way to have a workaround for this problem when hosting a DotNet web app with Azure App Services? It's a release blocker and we would never expect the azure internal proxy to magically alter parts of the original URL.

JJong-nl:
One issue.. 
The `X-Waws-Unencoded-Url` contains the `PathAndQuery`.
The `Request.Path` must be set with only the AbsolutePath.

```
if (context.Request.Headers.TryGetValue(UnencodedUrlHeaderName, out var unencodedUrlValue) && unencodedUrlValue.Any())
{
    var uri = new Uri($"https://fake.url{unencodedUrlValue.First()}");
    context.Request.Path = new PathString(uri.AbsolutePath);
}
```

Kusnierz-Maciej:
Whast is the status of internal fix for this?

AdiuTankT72:
I have tried to use this WEBSITE_RESTORE_RAW_REQUEST_PATH on our linux web app and it is not working :/

PR #9429, opened 2023-07-28
Improving error handling in SecretManager

### LINKED ISSUE #8731: Improve error handling in SecretManager
For example, the `BlobStorageSecretsRepository` could potentially throw an exception during a call to `WriteAsync()` -- but the `SecretManager` doesn't handle that well.

The one specific scenario is this:
- multiple instances start up for the first time simultaneously
- no secrets are found, so each instance races to create them
- only one wins, causing exceptions in the others

https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs#L95-L101

This code could see the exception and know that it should retry the Load to see if it lost the race, then continue happily.

aishwaryabh:
This PR improves error handling in the Secret manager so that if there's a race condition (ie multiple secret repositories are created, each try to create their own version of a secret, and whichever one actually creates it first succeeds and the rest error out). If an error happens with `WriteAsync`, we simply call `LoadSecretsAsync` first before failing and throwing an exception to make sure the issue is not a race condition

resolves #8731

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
+++ b/src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
@@ -189,12 +189,12 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
                         }
                         catch (Exception ex)
                         {
-                            _logger.LogDebug("Potential conflict, attempting to load function secrets for the second time.");
+                            _logger.LogDebug(ex, "Exception while generating function secrets. This can happen if another instance is also generating secrets. Attempting to read secrets again.");
                             secrets = await LoadFunctionSecretsAsync(functionName);
 
                             if (secrets == null)
                             {
-                                _logger.LogError(ex, "Failed to load function secrets on second attempt");
+                                _logger.LogError("Function secrets are still null on second attempt.");
                                 throw;
                             }
                         }
--- a/src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
+++ b/src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
@@ -11,10 +11,13 @@ using System.Security.Cryptography;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
+using DryIoc;
+using Microsoft.Azure.Web.DataProtection;
 using Microsoft.Azure.WebJobs.Extensions.Http;
 using Microsoft.Azure.WebJobs.Script.Diagnostics;
 using Microsoft.Azure.WebJobs.Script.WebHost.Properties;
 using Microsoft.Azure.WebJobs.Script.WebHost.Security;
+using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
 using DataProtectionConstants = Microsoft.Azure.Web.DataProtection.Constants;
 
@@ -101,12 +104,12 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
                         }
                         catch (Exception ex)
                         {
-                            _logger.LogDebug("Potential conflict, attempting to load host secrets for the second time.");
+                            _logger.LogDebug(ex, "Exception while generating host secrets. This can happen if another instance is also generating secrets. Attempting to read secrets again.");
                             hostSecrets = await LoadSecretsAsync<HostSecrets>();
 
                             if (hostSecrets == null)
                             {
-                                _logger.LogError(ex, "Failed to load host secrets on second attempt");
+                                _logger.LogError("Host secrets are still null on second attempt.");
                                 throw;
                             }
                         }
--- a/src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
+++ b/src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
@@ -102,7 +102,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
                         }
                         catch (Exception ex)
                         {
-                            _logger.LogDebug("Trying again to ensure potential race condition is handled correctly");
+                            _logger.LogDebug("Potential conflict, attempting to load host secrets for the second time.");
                             hostSecrets = await LoadSecretsAsync<HostSecrets>();
 
                             if (hostSecrets == null)
--- a/src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
+++ b/src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
@@ -16,7 +16,6 @@ using Microsoft.Azure.WebJobs.Script.Diagnostics;
 using Microsoft.Azure.WebJobs.Script.WebHost.Properties;
 using Microsoft.Azure.WebJobs.Script.WebHost.Security;
 using Microsoft.Extensions.Logging;
-
 using DataProtectionConstants = Microsoft.Azure.Web.DataProtection.Constants;
 
 namespace Microsoft.Azure.WebJobs.Script.WebHost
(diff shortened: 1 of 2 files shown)


src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
```diff
@@ -90,12 +90,26 @@ public async virtual Task<HostSecretsInfo> GetHostSecretsAsync()
...
+                                await PersistSecretsAsync(hostSecrets);
+                            }
+                        }
+                        catch (Exception ex)
                         {
-                            // host secrets do not yet exist so generate them
-                            _logger.LogDebug(Resources.TraceHostSecretGeneration);
-                            hostSecrets = GenerateHostSecrets();
-                            await PersistSecretsAsync(hostSecrets);
+                            _logger.LogDebug("Trying again to ensure potential race condition is handled correctly");
```
liliankasem:
```suggestion
                            _logger.LogDebug("Potential conflict, attempting to load host secrets for the second time.");
```

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
```diff
@@ -90,12 +90,26 @@ public async virtual Task<HostSecretsInfo> GetHostSecretsAsync()
...
+                            hostSecrets = await LoadSecretsAsync<HostSecrets>();
+
+                            if (hostSecrets == null)
+                            {
+                                _logger.LogError(ex, "Failed to load host secrets on second attempt");
```
jviau: Normally I would put this in a loop and perform the entire load -> persist -> catch/ try again over like 10 attempts. But if you believe this won't be necessary here that is fine. I do see how you can assume if an exception is thrown _something else_ persisted the secrets.

Can you think of any other scenarios here that would throw but the secrets remain un-persisted?

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
```diff
@@ -90,12 +90,26 @@ public async virtual Task<HostSecretsInfo> GetHostSecretsAsync()
...
+                        {
+                            if (hostSecrets == null)
+                            {
+                                // host secrets do not yet exist so generate them
+                                _logger.LogDebug(Resources.TraceHostSecretGeneration);
+                                hostSecrets = GenerateHostSecrets();
+                                await PersistSecretsAsync(hostSecrets);
+                            }
+                        }
+                        catch (Exception ex)
```
jviau, line 105: Is there a more specific exception you can catch?

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
```diff
@@ -90,12 +90,26 @@ public async virtual Task<HostSecretsInfo> GetHostSecretsAsync()
                         _logger.LogDebug("Loading host secrets");
 
                         hostSecrets = await LoadSecretsAsync<HostSecrets>();
-                        if (hostSecrets == null)
+                        try
+                        {
+                            if (hostSecrets == null)
```
mathewc, line 97: We also have the same load/generate/persist flow for Function secrets as well [here](https://github.com/Azure/azure-functions-host/blob/1715c505237c0ccc8de8b16fd80b9d16bcda37c3/src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs#L168). Do the error logs show that we're also getting races there as well? If so we probably want to make similar changes in both codepaths.

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
brettsam: I was wondering this as well... but because there's a handful of different `ISecretsRepository` implementations (Blob, Kubernetes, KeyVault, FileSystem, etc), we can't really know every possibility of exception.

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
brettsam: There could be brief connection issues as well (as an example) but I do think that most of the SDKs we use for persistence already retry for those kinds of things internally.

@aishwaryabh and I chatted about this when she started this PR and I explained this as a "let's give this one more try" kind of error handling b/c we've seen evidence of these races in the past, and rather than giving up, let's see if maybe things actually worked.

Since this code has been around for so long, I'd suggest we do this "one more try" approach, then we can monitor for how often we fail to load on the second attempt (and what kind of errors we see).

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
aishwaryabh: I can leave this as is right now and once we have more data to validate the common exceptions thrown we can change it

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
aishwaryabh: Made the changes!

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
aishwaryabh: Left it as is for the one more try approach

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
```diff
@@ -90,12 +89,26 @@ public async virtual Task<HostSecretsInfo> GetHostSecretsAsync()
...
+                            {
+                                // host secrets do not yet exist so generate them
+                                _logger.LogDebug(Resources.TraceHostSecretGeneration);
+                                hostSecrets = GenerateHostSecrets();
+                                await PersistSecretsAsync(hostSecrets);
+                            }
+                        }
+                        catch (Exception ex)
+                        {
+                            _logger.LogDebug("Potential conflict, attempting to load host secrets for the second time.");
```
brettsam: Couple comments on this:
- Let's pass the exception in here so we have it captured. `_logger.LogDebug(ex, "....")`
- Reword a little so it's clear what happened: `Exception while generating host secrets. This can happen if another instance is also generating secrets. Attempting to read secrets again.`

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
```diff
@@ -90,12 +89,26 @@ public async virtual Task<HostSecretsInfo> GetHostSecretsAsync()
...
+                            hostSecrets = await LoadSecretsAsync<HostSecrets>();
+
+                            if (hostSecrets == null)
+                            {
+                                _logger.LogError(ex, "Failed to load host secrets on second attempt");
```
brettsam: Here... do *not* include the exception b/c it's not the one that actually caused the problem... and instead of `Failed to load...` maybe say `Host secrets are still null on second attempt.` We may have actually successfully loaded them, but there just weren't any...

brettsam (commented):
Just a couple comments on the error messages...

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
aishwaryabh: Changed!

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
```diff
@@ -158,14 +174,29 @@ public async virtual Task<HostSecretsInfo> GetHostSecretsAsync()
...
+                                string message = string.Format(Resources.TraceFunctionSecretGeneration, functionName);
+                                _logger.LogDebug(message);
+                                secrets = GenerateFunctionSecrets();
+
+                                await PersistSecretsAsync(secrets, functionName);
+                            }
+                        }
+                        catch (Exception ex)
+                        {
+                            _logger.LogDebug("Potential conflict, attempting to load function secrets for the second time.");
```
brettsam: Can you apply the same logging changes from above down here?

brettsam (approved):

satvu (approved):

PR #9433, opened 2023-08-02
Extend http proxying timeout

### LINKED ISSUE #1745: ASP.NET Core integration HttpRequest timeout
During testing of a long-running (~120s) .NET 7 Isolated HttpTrigger function, we found that by using the ASP.NET Core types provided by [Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore](https://www.nuget.org/packages/Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore) _1.0.0-preview2_ (set up as indicated [here](https://learn.microsoft.com/en-us/azure/azure-functions/dotnet-isolated-process-guide#aspnet-core-integration-preview)), the HttpRequest object would request cancellation by itself, allowing the function to finish properly but returning `HTTP 504 Gateway Timeout` anyway.

Initially we suspected the Azure Load Balancer from the HttpTrigger as stated [here](https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-http-webhook-trigger#limits):  
> If a function that uses the HTTP trigger doesn't complete within 230 seconds, the [Azure Load Balancer](https://learn.microsoft.com/en-us/azure/app-service/faq-availability-performance-application-issues#why-does-my-request-time-out-after-230-seconds-) will time out and return an HTTP 502 error. The function will continue running but will be unable to return an HTTP response. For long-running functions, we recommend that you follow async patterns and return a location where you can ping the status of the request.

Yet, since the http code and amount of time didn't really match, we took a closer look at the HttpRequest, where we found that while the **FunctionContext.CancellationToken** was still valid, the **HttpContext.RequestAborted** one was already on a Cancelled state before exiting the function.  
This behaviour is very similar to the one described on the [upcoming Request Timeouts middleware](https://learn.microsoft.com/en-us/aspnet/core/performance/timeouts), even though we are using .NET 7 and have not set up any middleware:
> When a timeout limit is hit, a [CancellationToken](https://learn.microsoft.com/en-us/dotnet/api/system.threading.cancellationtoken) in [HttpContext.RequestAborted](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.httpcontext.requestaborted#microsoft-aspnetcore-http-httpcontext-requestaborted) has [IsCancellationRequested](https://learn.microsoft.com/en-us/dotnet/api/system.threading.cancellationtoken.iscancellationrequested#system-threading-cancellationtoken-iscancellationrequested) set to true. [Abort()](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.httpcontext.abort#microsoft-aspnetcore-http-httpcontext-abort) isn't automatically called on the request, so the application may still produce a success or failure response. The default behavior if the app doesn't handle the exception and produce a response is to return status code 504.

By reverting back to the "native" HttpTrigger objects (ie. HttpRequestData) and Host builder defaults, the issue was solved -- with the function returning as intended. Fortunately, removing the NuGet package was not needed.  
This can be easily tested by running an empty HttpTrigger function with a Task.Delay over 2 minutes 😄 

Thank you for your help!

satvu:

resolves [#1745 in dotnet worker](https://github.com/Azure/azure-functions-dotnet-worker/issues/1745).

Default yarp timeout is 100 seconds ([docs](https://microsoft.github.io/reverse-proxy/articles/http-client-config.html#httprequest)) which is shorter than the Antares default for HTTP requests (230s) so we need to set our own default value.

### Pull request checklist

### Additional information

Filed an issue for http timeout test suite [here](https://github.com/Azure/azure-functions-host/issues/9438).

### DIFF
--- a/src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
+++ b/src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
@@ -31,7 +31,10 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
             };
 
             _messageInvoker = new HttpMessageInvoker(_handler);
-            _forwarderRequestConfig = new ForwarderRequestConfig();
+            _forwarderRequestConfig = new ForwarderRequestConfig
+            {
+                ActivityTimeout = TimeSpan.FromSeconds(230)
+            };
         }
 
         public void Dispose()
--- a/src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
+++ b/src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
@@ -33,7 +33,7 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
             _messageInvoker = new HttpMessageInvoker(_handler);
             _forwarderRequestConfig = new ForwarderRequestConfig
             {
-                ActivityTimeout = TimeSpan.FromSeconds(230)
+                ActivityTimeout = TimeSpan.FromSeconds(240)
             };
         }
 
--- a/release_notes.md
+++ b/release_notes.md
@@ -3,20 +3,4 @@
 <!-- Please add your release notes in the following format:
 - My change description (#PR)
 -->
-- Update Java Worker Version to [2.12.1](https://github.com/Azure/azure-functions-java-worker/releases/tag/2.12.1)
-- Update Python Worker Version to [4.15.0](https://github.com/Azure/azure-functions-python-worker/releases/tag/4.15.0)
-
-- Remove feature flag for http proxying (https://github.com/Azure/azure-functions-host/pull/9341)
-- Add error handling for http proxying failure scenarios (https://github.com/Azure/azure-functions-host/pull/9342)
-- Update PowerShell Worker 7.0 to 4.0.2850 [Release Note](https://github.com/Azure/azure-functions-powershell-worker/releases/tag/v4.0.2850)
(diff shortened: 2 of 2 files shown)


src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
```diff
@@ -31,7 +31,10 @@ public DefaultHttpProxyService(IHttpForwarder httpForwarder)
             };
 
             _messageInvoker = new HttpMessageInvoker(_handler);
-            _forwarderRequestConfig = new ForwarderRequestConfig();
+            _forwarderRequestConfig = new ForwarderRequestConfig
+            {
+                ActivityTimeout = TimeSpan.FromSeconds(230)
```
satvu: Load balancer is 230s, but final decision is still under discussion for this default timeout (use the function timeout instead? other options?)

brettsam: Want to loop in @mathewc on this too. He's probably looked at more http-related things than any of us in the past.

We likely want to bump this up to 240 (I see the Antares front end does this) to prevent any races with the Azure Load Balancer cutoff of 230.

But we should also do a test to see what happens at 230 seconds today with an in-proc app in prod:
- Do we get the Function's CancellationToken canceled?
- What is HttpContext.IsRequestAborted?

Then... what happens when out-of-proc?

brettsam (approved):

kshyju (approved):

PR #9453, opened 2023-08-10
Improve Sanitizer

TsuyoshiUshio:

resolves internal issue. 

### Pull request checklist

### Additional information

Fixing two issues. 
1. Sanitize when Exception is not happened.
2. Add known secret.

### DIFF
--- a/src/WebJobs.Script.WebHost/Diagnostics/AzureMonitorDiagnosticLogger.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/AzureMonitorDiagnosticLogger.cs
@@ -73,10 +73,8 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
             }
 
             (string exceptionType, string exceptionMessage, string exceptionDetails) = exception.GetExceptionDetails();
-            if (exception != null)
-            {
-                formattedMessage = Sanitizer.Sanitize(formattedMessage);
-            }
+
+            formattedMessage = Sanitizer.Sanitize(formattedMessage);
 
             // enumerate all the state values once, capturing the values we'll use below
             // last one wins
--- a/src/WebJobs.Script/Sanitizer.cs
+++ b/src/WebJobs.Script/Sanitizer.cs
@@ -16,7 +16,7 @@ namespace Microsoft.Azure.WebJobs.Logging
 
         // List of keywords that should not be replaced with [Hidden Credential]
         private static readonly string[] AllowedTokens = new string[] { "PublicKeyToken=" };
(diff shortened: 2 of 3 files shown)


src/WebJobs.Script/Sanitizer.cs
```diff
@@ -78,6 +78,6 @@ internal static string Sanitize(string input)
         /// Checks if a string even *possibly* contains one of our <see cref="CredentialTokens"/>.
         /// Useful for short-circuiting more expensive checks and replacements if it's known we wouldn't do anything.
         /// </summary>
-        internal static bool MayContainCredentials(string input) => input.Contains("=");
```
liliankasem, line 81: Please also update the comment above `if (!MayContainCredentials(input))` (line 33)

TsuyoshiUshio: Good catch. Done.

liliankasem (approved):

src/WebJobs.Script.WebHost/Diagnostics/AzureMonitorDiagnosticLogger.cs
```diff
@@ -73,10 +73,8 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
             }
 
             (string exceptionType, string exceptionMessage, string exceptionDetails) = exception.GetExceptionDetails();
-            if (exception != null)
-            {
-                formattedMessage = Sanitizer.Sanitize(formattedMessage);
-            }
+
+            formattedMessage = Sanitizer.Sanitize(formattedMessage);
```
jviau, line 77: I fear this is going to be a significant perf regression. Logging is expensive enough as is. But adding in scanning of every log for secrets, and then string modifications and allocations if there is? This will be a non-negligible perf impact.

Why are these secrets even being logged in the first place?

TsuyoshiUshio: HI @jviau 
According to the logging, the log was NOT sanitized at all. That means, the sanitizing was skipped. 
I don't have full context of this class; however, if you don't like this solution, how about when we don't have exception, we don't do any logging. (return  if exception == null) or any other good idea?

mathewc: Yes, this would be regressing work I did in https://github.com/Azure/azure-functions-host/pull/5897 for throughput improvements that greatly increased our throughput by removing Sanitize from the hotpath. Sanitization should be done upstream as needed upstream by the component logging.

TsuyoshiUshio: I did the research of the logging pipeline. This pipeline is for Azure Monitor. 
So that customer logging should not be sanitized by us.
In this case, we should not sanitize in here. We are keeping on discussing on the internal e-mail thread. 
I'll abandoned this PR.

PR #9464, opened 2023-08-14
Reducing noise for metadata logging

### LINKED ISSUE #9420: Host and worker metadata providers can happen at the same time
In reproducing #9396 or something like it, you might see the following:

> [2023-07-25T23:22:43.599Z] 0 functions found
> [2023-07-25T23:22:43.607Z] Reading functions metadata
> [2023-07-25T23:22:43.626Z] 1 functions found
> [2023-07-25T23:22:43.628Z] 0 functions loaded

This seems to be the result of both host and worker metadata being checked for in parallel. @fabiocav and I traced through this, but I can't recall the exact changes we identified to remediate this. Regardless, we need to remove the invalid resolution attempt and the confusing noise it adds to logs.

aishwaryabh:
The host and worker metadata providers can be checked in parallel due to `GetFunctionMetadata` being called in the `FunctionMetadataManager`. After stepping through the code, I saw that the function gets called twice, once when we set `includeCustomProviders` to false and the second time when `includeCustomProviders` is set to true. We don't want to log how many functions have been loaded after we log how many have been found.

Here's how the logging looked like before:

> [2023-07-25T23:22:43.607Z]  Loading functions metadata
> [2023-07-25T23:22:43.607Z] Reading functions metadata
> [2023-07-25T23:22:43.626Z] 1 functions found
> [2023-07-25T23:22:43.628Z] 0 functions loaded

And here is what the logging looks like now:

> [2023-07-25T23:22:43.607Z]  Loading functions metadata
> [2023-07-25T23:22:43.607Z] Reading functions metadata (Custom)
> [2023-07-25T23:22:43.626Z] Reading functions metadata (Custom)
> [2023-07-25T23:22:43.628Z] 0 functions loaded

resolves #9420 

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
+++ b/src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
@@ -123,7 +123,7 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics.Extensions
 
         private static readonly Action<ILogger, string, Exception> _readingFunctionMetadataFromProvider =
             LoggerMessage.Define<string>(LogLevel.Information,
-                new EventId(326, nameof(FunctionMetadataManagerLoadingFunctionsMetadata)),
+                new EventId(326, nameof(ReadingFunctionMetadataFromProvider)),
                 "Reading functions metadata ({provider})");
 
         private static readonly Action<ILogger, int, string, Exception> _functionsReturnedByProvider =
@@ -280,14 +280,14 @@ Lock file hash: {currentLockFileHash}";
             _functionMetadataManagerLoadingFunctionsMetadata(logger, null);
         }
 
-        public static void FunctionMetadataManagerFunctionsLoaded(this ILogger logger, int count)
-        {
-            _functionMetadataManagerFunctionsLoaded(logger, count, null);
-        }
-
         public static void ReadingFunctionMetadataFromProvider(this ILogger logger, string provider)
         {
             _readingFunctionMetadataFromProvider(logger, provider, null);
+        }
+
+        public static void FunctionMetadataManagerFunctionsLoaded(this ILogger logger, int count)
+        {
+            _functionMetadataManagerFunctionsLoaded(logger, count, null);
         }
 
         public static void FunctionsReturnedByProvider(this ILogger logger, int count, string provider)
--- a/src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
+++ b/src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
@@ -84,12 +84,12 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics.Extensions
         private static readonly Action<ILogger, Exception> _functionMetadataManagerLoadingFunctionsMetadata =
             LoggerMessage.Define(LogLevel.Information,
                 new EventId(314, nameof(FunctionMetadataManagerLoadingFunctionsMetadata)),
-                "Loading functions metadata");
+                "Loading functions metadata From FunctionMetadataManager");
 
         private static readonly Action<ILogger, int, Exception> _functionMetadataManagerFunctionsLoaded =
             LoggerMessage.Define<int>(LogLevel.Information,
                 new EventId(315, nameof(FunctionMetadataManagerFunctionsLoaded)),
-                "{count} functions loaded");
+                "{count} functions loaded from FunctionMetadataManager");
 
         private static readonly Action<ILogger, string, string, Exception> _autoRecoveringFileSystemWatcherFailureDetected =
             LoggerMessage.Define<string, string>(LogLevel.Warning,
@@ -121,15 +121,15 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics.Extensions
                 new EventId(325, nameof(ScriptStartUpLoadedExtension)),
                 "Loaded extension '{startupExtensionName}' ({startupExtensionVersion})");
 
-        private static readonly Action<ILogger, Exception> _functionMetadataProviderReadingMetadata =
-            LoggerMessage.Define(LogLevel.Information,
+        private static readonly Action<ILogger, string, Exception> _functionMetadataProviderReadingMetadata =
+            LoggerMessage.Define<string>(LogLevel.Information,
                 new EventId(326, nameof(FunctionMetadataManagerLoadingFunctionsMetadata)),
-                "Reading functions metadata");
+                "Reading functions metadata from {source}");
 
-        private static readonly Action<ILogger, int, Exception> _functionMetadataProviderFunctionsFound =
-            LoggerMessage.Define<int>(LogLevel.Information,
+        private static readonly Action<ILogger, int, string, Exception> _functionMetadataProviderFunctionsFound =
+            LoggerMessage.Define<int, string>(LogLevel.Information,
                 new EventId(327, nameof(FunctionMetadataManagerFunctionsLoaded)),
-                "{count} functions found");
+                "{count} functions found in {source}");
 
         private static readonly Action<ILogger, string, Guid, Exception> _customHandlerForwardingHttpTriggerInvocation =
             LoggerMessage.Define<string, Guid>(LogLevel.Debug,
@@ -182,6 +182,11 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics.Extensions
             LogLevel.Error,
             new EventId(335, nameof(MinimumBundleVersionNotSatisfied)),
             "Referenced bundle {bundleId} of version {bundleVersion} does not meet the required minimum version of {minimumVersion}. Update your extension bundle reference in host.json to reference {minimumVersion2} or later.");
+
+        private static readonly Action<ILogger, int, Exception> _functionMetadataProviderFunctionsFoundCustomProvider =
+            LoggerMessage.Define<int>(LogLevel.Information,
+            new EventId(336, nameof(FunctionMetadataManagerFunctionsLoaded)),
+            "{count} functions found in FunctionMetadataManager from custom provider");
 
         public static void ExtensionsManagerRestoring(this ILogger logger)
         {
@@ -285,14 +290,19 @@ Lock file hash: {currentLockFileHash}";
             _functionMetadataManagerFunctionsLoaded(logger, count, null);
         }
 
-        public static void FunctionMetadataProviderParsingFunctions(this ILogger logger)
+        public static void FunctionMetadataProviderParsingFunctions(this ILogger logger, string source)
         {
-            _functionMetadataProviderReadingMetadata(logger, null);
+            _functionMetadataProviderReadingMetadata(logger, source, null);
         }
 
-        public static void FunctionMetadataProviderFunctionFound(this ILogger logger, int count)
+        public static void FunctionMetadataProviderFunctionFound(this ILogger logger, int count, string source)
         {
-            _functionMetadataProviderFunctionsFound(logger, count, null);
+            _functionMetadataProviderFunctionsFound(logger, count, source, null);
+        }
+
+        public static void FunctionMetadataProviderFunctionFoundCustomProvider(this ILogger logger, int count)
+        {
+            _functionMetadataProviderFunctionsFoundCustomProvider(logger, count, null);
         }
 
         public static void AutoRecoveringFileSystemWatcherFailureDetected(this ILogger logger, string errorMessage, string path)
--- a/src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
+++ b/src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
@@ -86,9 +86,9 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics.Extensions
                 new EventId(314, nameof(FunctionMetadataManagerLoadingFunctionsMetadata)),
                 "Loading functions metadata");
 
-        private static readonly Action<ILogger, int, Exception> _functionMetadataProviderFunctionsLoaded =
+        private static readonly Action<ILogger, int, Exception> _functionMetadataManagerFunctionsLoaded =
             LoggerMessage.Define<int>(LogLevel.Information,
-                new EventId(315, nameof(FunctionsLoadedByProvider)),
+                new EventId(315, nameof(FunctionMetadataManagerFunctionsLoaded)),
                 "{count} functions loaded");
 
         private static readonly Action<ILogger, string, string, Exception> _autoRecoveringFileSystemWatcherFailureDetected =
@@ -280,12 +280,12 @@ Lock file hash: {currentLockFileHash}";
             _functionMetadataManagerLoadingFunctionsMetadata(logger, null);
         }
 
-        public static void FunctionsLoadedByProvider(this ILogger logger, int count)
+        public static void FunctionMetadataManagerFunctionsLoaded(this ILogger logger, int count)
         {
-            _functionMetadataProviderFunctionsLoaded(logger, count, null);
+            _functionMetadataManagerFunctionsLoaded(logger, count, null);
         }
 
-        public static void FunctionMetadataProviderParsingFunctions(this ILogger logger, string provider)
+        public static void ReadingFunctionMetadataFromProvider(this ILogger logger, string provider)
         {
             _readingFunctionMetadataFromProvider(logger, provider, null);
         }
--- a/src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
+++ b/src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
@@ -84,12 +84,12 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics.Extensions
         private static readonly Action<ILogger, Exception> _functionMetadataManagerLoadingFunctionsMetadata =
             LoggerMessage.Define(LogLevel.Information,
                 new EventId(314, nameof(FunctionMetadataManagerLoadingFunctionsMetadata)),
-                "Loading functions metadata From FunctionMetadataManager");
+                "Loading functions metadata");
 
         private static readonly Action<ILogger, int, Exception> _functionMetadataManagerFunctionsLoaded =
             LoggerMessage.Define<int>(LogLevel.Information,
                 new EventId(315, nameof(FunctionMetadataManagerFunctionsLoaded)),
-                "{count} functions loaded from FunctionMetadataManager");
+                "{count} functions loaded");
 
         private static readonly Action<ILogger, string, string, Exception> _autoRecoveringFileSystemWatcherFailureDetected =
             LoggerMessage.Define<string, string>(LogLevel.Warning,
--- a/src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
+++ b/src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
@@ -86,9 +86,9 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics.Extensions
                 new EventId(314, nameof(FunctionMetadataManagerLoadingFunctionsMetadata)),
                 "Loading functions metadata");
 
-        private static readonly Action<ILogger, int, Exception> _functionMetadataManagerFunctionsLoaded =
+        private static readonly Action<ILogger, int, Exception> _functionMetadataProviderFunctionsLoaded =
             LoggerMessage.Define<int>(LogLevel.Information,
-                new EventId(315, nameof(FunctionMetadataManagerFunctionsLoaded)),
+                new EventId(315, nameof(FunctionsLoadedByProvider)),
                 "{count} functions loaded");
 
         private static readonly Action<ILogger, string, string, Exception> _autoRecoveringFileSystemWatcherFailureDetected =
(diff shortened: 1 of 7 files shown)


liliankasem (approved):

liliankasem:
Don't forget to also update the change log!

mathewc:
I've been seeing these extraneous logs and have wondered where they were coming from. Good to clean this up since it's confusing during log investigations

src/WebJobs.Script/Host/FunctionMetadataManager.cs
```diff
@@ -87,7 +87,11 @@ public ImmutableArray<FunctionMetadata> GetFunctionMetadata(bool forceRefresh, b
             if (forceRefresh || _servicesReset || _functionMetadataArray.IsDefaultOrEmpty)
             {
                 _functionMetadataArray = LoadFunctionMetadata(forceRefresh, includeCustomProviders, workerConfigs: workerConfigs);
-                _logger.FunctionMetadataManagerFunctionsLoaded(ApplyAllowList(_functionMetadataArray).Count());
+                // The host and worker metadata are checked in parallel, so we don't want to emit this extra logging reading the metadata
+                if (!includeCustomProviders)
```
mathewc: Currently the only place this is called with a false value is [here](https://github.com/Azure/azure-functions-host/blob/2f0a85e7ec8d7654dc2ba9254d97c57777cdcb8c/src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs#L92) right?

I don't understand why you say we don't want to log how many functions we actually load. The first log entry we write is how many function directories we've discovered (or how many worker indexing return us). The second log is how many functions actually will be active/loaded based on filtering, right? Don't we care about both?

liliankasem: I was under (a potentially incorrect) assumption that we log "x functions loaded" again somewhere else? But I might have mixed that up with the double log "x functions found". I that case I think we should keep a "loaded" log and suppress the "found" log that happens twice

mathewc: That's what I remember seeing in the past - multiple "x functions loaded" lines where they disagree on the number. We should check prod logs to see

aishwaryabh: @mathewc so the found logs should be suppressed and not the loaded logs?

These are the prod logs:

> [2023-08-14T22:29:42.083Z] Loading functions metadata
[2023-08-14T22:29:42.085Z] Reading functions metadata
[2023-08-14T22:29:42.087Z] 0 functions found
[2023-08-14T22:29:42.097Z] Reading functions metadata
[2023-08-14T22:29:42.116Z] 1 functions found
[2023-08-14T22:29:42.118Z] 0 functions loaded

[This](https://vscode.dev/github/Azure/azure-functions-host/blob/aibhandari/metadata-logging-noise/src/WebJobs.Script/Host/HostFunctionMetadataProvider.cs#L52) is where the first `0 functions found` log happens.
[This](https://vscode.dev/github/Azure/azure-functions-host/blob/aibhandari/metadata-logging-noise/src/WebJobs.Script/Host/FunctionMetadataManager.cs#L235) is where the second `1 function found` log is emitted.

If I'm understanding your comment correctly, we would not need the second `1 function found` log? This happens since `includeCustomProviders` is set to true

liliankasem: Confirmed with prod it's that "found" is what is listed twice.

Does the second log always have the correct value or is it sometimes reset? i.e. 

1 functions found 
Reading functions metadata
0 functions found

I think if there is a race in some way here, it might be difficult to decide on which one to keep or get rid of?

Or do we even need to log what we found and can remove both, trusting that "loaded" is the most accurate value? But even with your example log above, thats also incorrect:

1 functions found 
0 functions loaded

- very confuse

aishwaryabh: @liliankasem I believe the second log always has the correct value, at least when I've tested it out. I think the second value differs because of `includeCustomProviders` being true [here](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Host/FunctionMetadataManager.cs#L161). Maybe it makes sense to only include the second found log then and not include the first found one?

mathewc: The confusion came when we added worker indexing and this new secondary codepath, but chose to reuse the same existing FunctionMetadataProviderFunctionFound message template. Looks like this error message is emitted in 3 different places. So definitely if we see logs from all 3 of those in one host startup trace it'll be very confusing.

I think we want to retain all the information the logs are providing, just make it clearer. E.g. it looks like we have multiple providers (HostFunctionMetadataProvider and WorkerFunctionMetadataProvider) both being consulted and logging their own intermediate results, with the outer orchestrating FunctionMetadataProvider calling these and logging itself.

Perhaps we just need to update the error message so it includes details on the provider that is doing the loading so the logs can be differentiated.

aishwaryabh: Updated the logging so that it specifies where that log came from (either HostFuncdtionMatadataProvider, WorkerFunctionMetadataProvider, or FunctionMetadataProvider)

src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
```diff
@@ -121,15 +121,15 @@ internal static class LoggerExtension
                 new EventId(325, nameof(ScriptStartUpLoadedExtension)),
                 "Loaded extension '{startupExtensionName}' ({startupExtensionVersion})");
 
-        private static readonly Action<ILogger, Exception> _functionMetadataProviderReadingMetadata =
```
mathewc, line 124: It's hard without tests to see what the results of these changes are. I see there's a FunctionMetadataProviderTests suite that is exercising worker/host indexing scenarios, but perhaps those tests aren't validating these logs.

I think we should update those tests so they're validating these logs, and or also add a test in there that repros the issue described in https://github.com/Azure/azure-functions-host/issues/9420 and validates/demonstrates the fix.

With the test validating log details, it'll be easy to see the results, and we'll have a regression test as well.

aishwaryabh: Updated some existing tests to also validate loggings

src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
```diff
@@ -121,15 +121,15 @@ internal static class LoggerExtension
...
-            LoggerMessage.Define(LogLevel.Information,
+        private static readonly Action<ILogger, string, Exception> _functionMetadataProviderReadingMetadata =
```
mathewc: Initially the naming for this field and it's corresponding method FunctionMetadataProviderParsingFunctions made sense before worker indexing was added and we started supporting multiple sources. Now it's being reused in 3 different places: Host file provider, Worker index provider, and Custom providers.

So I think we should rename the property/method to _readingFunctionMetadataFromProvider and ReadingFunctionMetadataFromProvider respectively, then change the message to "Reading functions metadata ({provider)}"

src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
```diff
@@ -121,15 +121,15 @@ internal static class LoggerExtension
...
+            LoggerMessage.Define<string>(LogLevel.Information,
                 new EventId(326, nameof(FunctionMetadataManagerLoadingFunctionsMetadata)),
-                "Reading functions metadata");
+                "Reading functions metadata from {source}");
 
-        private static readonly Action<ILogger, int, Exception> _functionMetadataProviderFunctionsFound =
-            LoggerMessage.Define<int>(LogLevel.Information,
+        private static readonly Action<ILogger, int, string, Exception> _functionMetadataProviderFunctionsFound =
```
mathewc: Similarly I think we should rename this to _functionsReturnedByProvider and FunctionsReturnedByProvider respectively, and message "{count} functions found ({provider}}"

src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
```diff
@@ -183,6 +183,11 @@ internal static class LoggerExtension
             new EventId(335, nameof(MinimumBundleVersionNotSatisfied)),
             "Referenced bundle {bundleId} of version {bundleVersion} does not meet the required minimum version of {minimumVersion}. Update your extension bundle reference in host.json to reference {minimumVersion2} or later.");
 
+        private static readonly Action<ILogger, int, Exception> _functionMetadataProviderFunctionsFoundCustomProvider =
```
mathewc: With the naming above, we can reuse the Provider parameterized messages as is - we don't need a separate one for custom.

For the Provider IDs passed in, rather than using the various class names, how about we just define a constant for each so we have: Host, Worker and Custom for function metadata providers.

src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
```diff
@@ -121,15 +121,15 @@ internal static class LoggerExtension
...
+            LoggerMessage.Define<string>(LogLevel.Information,
                 new EventId(326, nameof(FunctionMetadataManagerLoadingFunctionsMetadata)),
-                "Reading functions metadata");
+                "Reading functions metadata from {source}");
 
-        private static readonly Action<ILogger, int, Exception> _functionMetadataProviderFunctionsFound =
-            LoggerMessage.Define<int>(LogLevel.Information,
+        private static readonly Action<ILogger, int, string, Exception> _functionMetadataProviderFunctionsFound =
+            LoggerMessage.Define<int, string>(LogLevel.Information,
                 new EventId(327, nameof(FunctionMetadataManagerFunctionsLoaded)),
```
mathewc: In reviewing this code, I noticed what appears to be a copy paste error here (before your changes). As you can see `nameof(FunctionMetadataManagerFunctionsLoaded)` is used incorrectly here. It should be `nameof(FunctionMetadataProviderFunctionFound)`, then pattern here being the name of the event maps to the log method.

You might check the others for errors as well.

test/WebJobs.Script.Tests/FunctionMetadataManagerTests.cs
```diff
@@ -153,12 +159,18 @@ public void FunctionMetadataManager_GetsMetadata_FromFunctionProviders()
...
 
             Assert.Equal(0, testFunctionMetadataManager.Errors.Count);
             Assert.Equal(1, testFunctionMetadataManager.GetFunctionMetadata(true).Length);
+            var traces = testLoggerProvider.GetAllLogMessages();
             Assert.Equal("anotherFunction", testFunctionMetadataManager.GetFunctionMetadata(true).FirstOrDefault()?.Name);
+
+            // Assert logging traces print out as expected
+            Assert.Equal(7, traces.Count);
+            Assert.Equal(2, traces.Count(t => t.FormattedMessage.Contains("Reading functions metadata from FunctionMetadataManager")));
+            Assert.Equal(2, traces.Count(t => t.FormattedMessage.Contains("1 functions found in FunctionMetadataManager from custom provider")));
```
mathewc: Is the "0 functions loaded" also logged? Can you include asserts for this as well in these tests? I'm still confused why the updated logs you show in the PR description shows a function was found, but 0 loaded.

aishwaryabh: So the custom provider finds that one function is there which is why 1 function is found [here](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Host/FunctionMetadataManager.cs#L231). However, since the metadata array is empty it doesn't add it to the `functionMetadataList` which is what the loaded statement prints out at the end

aishwaryabh: So in the scenario listed in the PR description, it finds a function through the custom provider [here.](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Host/FunctionMetadataManager.cs#L231) For the specific function I was testing, there was no metadata associated with it,  I suppose which is why it says that no functions are loaded and it doesn't add it to the `functionMetadataList`.

For the mock test, it actually does load the function metadata since we explicitly add it so it actually does load it. I will add that in the test.

src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
aishwaryabh: Changed!

src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
```diff
@@ -86,9 +86,9 @@ internal static class LoggerExtension
                 new EventId(314, nameof(FunctionMetadataManagerLoadingFunctionsMetadata)),
                 "Loading functions metadata");
 
-        private static readonly Action<ILogger, int, Exception> _functionMetadataManagerFunctionsLoaded =
+        private static readonly Action<ILogger, int, Exception> _functionMetadataProviderFunctionsLoaded =
```
mathewc: I think the changes to this field/method were unintentional? It's only used by the FunctionMetadataManager class in one place, and you're not passing in a provider or anything as in the other log messages. I think you should revert the changes to this one.

src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
```diff
@@ -280,19 +280,19 @@ public static void FunctionMetadataManagerLoadingFunctionsMetadata(this ILogger
...
 
-        public static void FunctionMetadataManagerFunctionsLoaded(this ILogger logger, int count)
+        public static void FunctionsLoadedByProvider(this ILogger logger, int count)
         {
-            _functionMetadataManagerFunctionsLoaded(logger, count, null);
+            _functionMetadataProviderFunctionsLoaded(logger, count, null);
         }
 
-        public static void FunctionMetadataProviderParsingFunctions(this ILogger logger)
+        public static void FunctionMetadataProviderParsingFunctions(this ILogger logger, string provider)
```
mathewc: As per the comment, this log method should also be renamed to ReadingFunctionMetadataFromProvider. Field and method names in sync.

src/WebJobs.Script/Host/FunctionMetadataManager.cs
```diff
@@ -215,7 +216,7 @@ private void LoadCustomProviderFunctions(List<FunctionMetadata> functionMetadata
 
         private void AddMetadataFromCustomProviders(IEnumerable<IFunctionProvider> functionProviders, List<FunctionMetadata> functionMetadataList)
         {
-            _logger.FunctionMetadataProviderParsingFunctions();
+            _logger.FunctionMetadataProviderParsingFunctions(nameof(FunctionMetadataManager));
```
mathewc: I think you meant to use _customProviderLog here as well?

mathewc: Also, for all these Host/Worker/Custom const fields, let's name them all _metadataProviderName in each of their respective classes

test/WebJobs.Script.Tests/HostFunctionMetadataProviderTests.cs
```diff
@@ -31,14 +33,24 @@ public HostFunctionMetadataProviderTests()
...
             var workerConfigs = TestHelpers.GetTestWorkerConfigs();
 
             Assert.Equal(18, metadataProvider.GetFunctionMetadataAsync(workerConfigs, SystemEnvironment.Instance, false).Result.Length);
+            var traces = testLoggerProvider.GetAllLogMessages();
             Assert.True(AreRequiredMetricsEmitted(_testMetricsLogger));
+
+            // Assert that the logs contain the expected messages
+            Assert.Equal(2, traces.Count);
+            Assert.Equal("Reading functions metadata (Host)", traces[0].FormattedMessage);
+            Assert.Equal("18 functions found (Host)", traces[1].FormattedMessage);
```
mathewc, line 53: Why don't we get the "x functions loaded" log here?

test/WebJobs.Script.Tests/FunctionMetadataManagerTests.cs
```diff
@@ -153,12 +159,20 @@ public void FunctionMetadataManager_GetsMetadata_FromFunctionProviders()
...
             testFunctionMetadataManager.LoadFunctionMetadata();
 
             Assert.Equal(0, testFunctionMetadataManager.Errors.Count);
             Assert.Equal(1, testFunctionMetadataManager.GetFunctionMetadata(true).Length);
+            var traces = testLoggerProvider.GetAllLogMessages();
             Assert.Equal("anotherFunction", testFunctionMetadataManager.GetFunctionMetadata(true).FirstOrDefault()?.Name);
+
+            // Assert logging traces print out as expected
+            Assert.Equal(7, traces.Count);
+            Assert.Equal(2, traces.Count(t => t.FormattedMessage.Contains("Reading functions metadata (FunctionMetadataManager)")));
```
mathewc: This test discrepancy here FunctionMetadataManager/Custom points to your code issue :)

test/WebJobs.Script.Tests/FunctionMetadataManagerTests.cs
```diff
@@ -12,9 +12,12 @@
 using Microsoft.Azure.WebJobs.Script.Workers.Http;
 using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
 using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
+using Microsoft.WebJobs.Script.Tests;
 using Moq;
 using Xunit;
+using static Microsoft.Azure.AppService.Proxy.Common.Constants.WellKnownHttpHeaders;
```
mathewc, line 20: I see there's also a FunctionMetadataProviderTests test suite that exercises WorkerIndexing. I think you should add/update some tests there as well so we have test coverage for all 3 provider flows you're touching here.

src/WebJobs.Script/Host/FunctionMetadataManager.cs
aishwaryabh: Changed!

test/WebJobs.Script.Tests/FunctionMetadataManagerTests.cs
aishwaryabh: Changed!

test/WebJobs.Script.Tests/HostFunctionMetadataProviderTests.cs
aishwaryabh: The "x functions loaded" only happens in the `FunctionMetadataManager` [here](https://github.com/Azure/azure-functions-host/pull/9464/files#diff-08e726fbbb6cad2322ba3a945fc8002af0346b11566a5b94042e00f33ec9a592R91), so the host doesn't log it

src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
```diff
@@ -121,15 +121,15 @@ internal static class LoggerExtension
                 new EventId(325, nameof(ScriptStartUpLoadedExtension)),
                 "Loaded extension '{startupExtensionName}' ({startupExtensionVersion})");
 
-        private static readonly Action<ILogger, Exception> _functionMetadataProviderReadingMetadata =
-            LoggerMessage.Define(LogLevel.Information,
+        private static readonly Action<ILogger, string, Exception> _readingFunctionMetadataFromProvider =
+            LoggerMessage.Define<string>(LogLevel.Information,
                 new EventId(326, nameof(FunctionMetadataManagerLoadingFunctionsMetadata)),
```
mathewc: Looks like a naming mismatch here in the EventId. The patter for all of these is to use the method name that uses the field. In this case should be "ReadingFunctionMetadataFromProvider"

src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
```diff
@@ -19,6 +19,7 @@ namespace Microsoft.Azure.WebJobs.Script
 {
     internal class WorkerFunctionMetadataProvider : IWorkerFunctionMetadataProvider
     {
+        private const string _customProviderLog = "Worker";
```
mathewc: Should this be renamed to _metadataProviderName to match the others?

test/WebJobs.Script.Tests/WorkerFunctionMetadataProviderTests.cs
```diff
@@ -140,5 +143,32 @@ public void ValidateBindings_OutputNameWithoutDirection_Throws()
...
+            });
+
+            var workerFunctionMetadataProvider = new WorkerFunctionMetadataProvider(optionsMonitor, logger, SystemEnvironment.Instance, mockWebHostRpcWorkerChannelManager.Object);
+            await workerFunctionMetadataProvider.GetFunctionMetadataAsync(workerConfigs, false);
+
+            var traces = logger.GetLogMessages();
+
+            // Assert that the logs contain the expected messages
+            Assert.Equal(2, traces.Count);
+            Assert.Equal("Reading functions metadata (Worker)", traces[1].FormattedMessage);
```
mathewc, line 175: Is there also a "x functions found (Custom)" assert you can add here for completeness?

src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
aishwaryabh: Fixed!

test/WebJobs.Script.Tests/WorkerFunctionMetadataProviderTests.cs
aishwaryabh: So "x functions found (Custom)" doesn't appear in the `WorkerFunctionMetadataProvider` but I added the other log statement that does appear.

mathewc (approved):

PR #9486, opened 2023-08-24
fetching IsToolingConsoleJsonLogEntry earlier to break the debugging deadlock

### LINKED ISSUE #1818: Local debugging does not work with FunctionsEnableWorkerIndexing
Visual studio F5 debugging does not work when .csproj has `<FunctionsEnableWorkerIndexing>True</FunctionsEnableWorkerIndexing>`.

The console window will open. but it does not print anything (like the discovered functions in the app etc..)

soninaren:
Debugging isolated dotnet worker does not work with worker indexing enabled. This is because we have a deadlock.

- VS waits for json file to be written by core tools to attach a debugger
- core tools is waiting for worker console service to write logs
- Worker is waiting for debugger to attach in order to move on
- Webhost waits for response from the init / metadata response from the worker
- Worker service never starts as the metadata response has not been received yet.

This PR breaks the deadlock by logging message outside the context of worker service.

resolves https://github.com/Azure/azure-functions-dotnet-worker/issues/1818

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs
+++ b/src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs
@@ -26,13 +26,14 @@ namespace Microsoft.Azure.WebJobs.Script.Workers
         private readonly int processExitTimeoutInMilliseconds = 1000;
         private readonly IServiceProvider _serviceProvider;
         private readonly IDisposable _eventSubscription;
+        private readonly Lazy<ILogger> _toolingConsoleJsonLoggerLazy;
 
         private bool _useStdErrorStreamForErrorsOnly;
         private Queue<string> _processStdErrDataQueue = new Queue<string>(3);
         private IHostProcessMonitor _processMonitor;
         private object _syncLock = new object();
 
-        internal WorkerProcess(IScriptEventManager eventManager, IProcessRegistry processRegistry, ILogger workerProcessLogger, IWorkerConsoleLogSource consoleLogSource, IMetricsLogger metricsLogger, IServiceProvider serviceProvider, bool useStdErrStreamForErrorsOnly = false)
+        internal WorkerProcess(IScriptEventManager eventManager, IProcessRegistry processRegistry, ILogger workerProcessLogger, IWorkerConsoleLogSource consoleLogSource, IMetricsLogger metricsLogger, IServiceProvider serviceProvider, ILoggerFactory loggerFactory, bool useStdErrStreamForErrorsOnly = false)
         {
             _processRegistry = processRegistry;
             _workerProcessLogger = workerProcessLogger;
@@ -41,6 +42,7 @@ namespace Microsoft.Azure.WebJobs.Script.Workers
             _metricsLogger = metricsLogger;
             _useStdErrorStreamForErrorsOnly = useStdErrStreamForErrorsOnly;
             _serviceProvider = serviceProvider;
+            _toolingConsoleJsonLoggerLazy = new Lazy<ILogger>(() => loggerFactory.CreateLogger(WorkerConstants.ToolingConsoleLogCategoryName), isThreadSafe: true);
 
             // We subscribe to host start events so we can handle the restart that occurs
             // on host specialization.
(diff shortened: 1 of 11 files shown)


src/WebJobs.Script/Workers/ProcessManagement/WorkerProcessUtilities.cs
```diff
@@ -43,5 +43,10 @@ public static bool IsToolingConsoleJsonLogEntry(ConsoleLog consoleLog)
         {
             return consoleLog.Message.StartsWith(WorkerConstants.ToolingConsoleLogPrefix, StringComparison.OrdinalIgnoreCase);
```
brettsam: can you make this call your new method?

src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs
```diff
@@ -201,6 +201,12 @@ internal void BuildAndLogConsoleLog(string msg, LogLevel level)
             {
                 _consoleLogSource?.Log(consoleLog);
             }
+
+            if (WorkerProcessUtilities.IsToolingConsoleJsonLogEntry(msg))
+            {
+                // log with the message prefix as coretools expects it.
+                _workerProcessLogger?.Log(level, msg);
+            }
```
brettsam: Do the same thing that the other usage of this did -- I'm not sure if that Log category is required, but let's not change it:

```
_toolingConsoleJsonLoggerLazy = new Lazy<ILogger>(() => loggerFactory.CreateLogger(WorkerConstants.ToolingConsoleLogCategoryName), isThreadSafe: true);
```

src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs
```diff
@@ -201,6 +201,12 @@ internal void BuildAndLogConsoleLog(string msg, LogLevel level)
             {
                 _consoleLogSource?.Log(consoleLog);
             }
+
+            if (WorkerProcessUtilities.IsToolingConsoleJsonLogEntry(msg))
```
brettsam: move this up to an `else if` in the chain above? we shouldn't send this log to `_consoleLogSource` anymore...

src/WebJobs.Script/Workers/ProcessManagement/WorkerProcessUtilities.cs
fabiocav: Can we remove this method instead? I think the only code calling this is being removed as part of this PR.

soninaren: Done

src/WebJobs.Script/Workers/WorkerConsoleLogService.cs
```diff
@@ -62,16 +59,7 @@ internal async Task ProcessLogs()
                 while (await source.OutputAvailableAsync(_cts.Token))
                 {
                     var consoleLog = await source.ReceiveAsync();
-
-                    if (WorkerProcessUtilities.IsToolingConsoleJsonLogEntry(consoleLog))
```
kshyju, line 66: I think we added this case in #8435 Would that use case break now?

kshyju: https://github.com/Azure/azure-functions-dotnet-worker/issues/900

brettsam: should be the same behavior above (same logger category, too)... we're just moving it up a level so that it happens at the WebHost rather than JobHost

brettsam: (but @soninaren -- you should double-check that it still works)

soninaren: I tested it end to end, it works as expected

soninaren:
@fabiocav, @kshyju @brettsam Can i get PR approval? I think i have addressed all PR comments.

kshyju (approved):

brettsam (approved):

PR #9506, opened 2023-08-31
Flex Consumption Metrics Publishing

mathewc:
Implements the metrics publisher for Flex Consumption. The metrics publisher by default will publish metrics every 5 seconds. Metrics files will only be written when the right environment variable is set by Legion for file path (FUNCTIONS_METRICS_PUBLISH_PATH). Regardless of whether files are written, the publisher will log metrics details each interval.

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.WebHost/Configuration/FlexConsumptionMetricsPublisherOptionsSetup.cs
+++ b/src/WebJobs.Script.WebHost/Configuration/FlexConsumptionMetricsPublisherOptionsSetup.cs
@@ -1,0 +1,22 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Configuration
+{
+    public class FlexConsumptionMetricsPublisherOptionsSetup : IConfigureOptions<FlexConsumptionMetricsPublisherOptions>
+    {
+        private IEnvironment _environment;
+
+        public FlexConsumptionMetricsPublisherOptionsSetup(IEnvironment environment)
+        {
+            _environment = environment;
+        }
+
+        public void Configure(FlexConsumptionMetricsPublisherOptions options)
+        {
+            options.MetricsFilePath = _environment.GetEnvironmentVariable(EnvironmentSettingNames.FunctionsMetricsPublishPath);
+        }
+    }
+}
--- a/src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
+++ b/src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
@@ -1,0 +1,301 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using System.IO.Abstractions;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Script.Diagnostics;
+using Microsoft.Azure.WebJobs.Script.Diagnostics.Extensions;
+using Microsoft.Azure.WebJobs.Script.WebHost.Configuration;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Metrics
+{
+    public class FlexConsumptionMetricsPublisher : IMetricsPublisher, IDisposable
+    {
+        private readonly IOptionsMonitor<StandbyOptions> _standbyOptions;
+        private readonly FlexConsumptionMetricsPublisherOptions _options;
+        private readonly IEnvironment _environment;
+        private readonly ILogger<FlexConsumptionMetricsPublisher> _logger;
+        private readonly object _lock = new object();
+        private readonly IFileSystem _fileSystem;
+
+        private Timer _metricsPublisherTimer;
+        private bool _started = false;
+        private ValueStopwatch _instanceActivityStopwatch;
+        private ValueStopwatch _intervalStopwatch;
+        private IDisposable _standbyOptionsOnChangeSubscription;
+        private TimeSpan _metricPublishInterval;
+        private TimeSpan _initialPublishDelay;
+
+        public FlexConsumptionMetricsPublisher(IEnvironment environment, IOptionsMonitor<StandbyOptions> standbyOptions, IOptions<FlexConsumptionMetricsPublisherOptions> options, ILogger<FlexConsumptionMetricsPublisher> logger, IFileSystem fileSystem)
+        {
+            _standbyOptions = standbyOptions ?? throw new ArgumentNullException(nameof(standbyOptions));
+            _environment = environment ?? throw new ArgumentNullException(nameof(environment));
+            _options = options?.Value ?? throw new ArgumentNullException(nameof(options));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            _fileSystem = fileSystem ?? new FileSystem();
+
+            if (_standbyOptions.CurrentValue.InStandbyMode)
+            {
+                _logger.LogDebug("Registering StandbyOptions change subscription.");
+                _standbyOptionsOnChangeSubscription = _standbyOptions.OnChange(o => OnStandbyOptionsChange());
+            }
+            else
+            {
+                Start();
+            }
+        }
+
+        // internal properties for testing
+        internal long FunctionExecutionCount { get; set; }
+
+        internal long FunctionExecutionTimeMS { get; set; }
+
+        internal long ActiveFunctionCount { get; set; }
+
+        internal bool IsAlwaysReady { get; set; }
+
+        internal string MetricsFilePath { get; set; }
+
+        public void Start()
+        {
+            Initialize();
+
+            _logger.LogInformation($"Starting metrics publisher (AlwaysReady={IsAlwaysReady}, MetricsPath='{MetricsFilePath}').");
+
+            _metricsPublisherTimer = new Timer(OnFunctionMetricsPublishTimer, null, _initialPublishDelay, _metricPublishInterval);
+            _started = true;
+        }
+
+        /// <summary>
+        /// Initialize any environmentally derived state after specialization, prior to starting the publisher.
+        /// </summary>
+        internal void Initialize()
+        {
+            _metricPublishInterval = TimeSpan.FromMilliseconds(_options.MetricsPublishIntervalMS);
+            _initialPublishDelay = TimeSpan.FromMilliseconds(_options.InitialPublishDelayMS);
+            _intervalStopwatch = ValueStopwatch.StartNew();
+            MetricsFilePath = _options.MetricsFilePath;
+
+            IsAlwaysReady = _environment.GetEnvironmentVariable(EnvironmentSettingNames.FunctionsAlwaysReadyInstance) == "1";
+        }
+
+        internal async Task OnPublishMetrics()
+        {
+            try
+            {
+                if (FunctionExecutionCount == 0 && FunctionExecutionTimeMS == 0 && !IsAlwaysReady)
+                {
+                    // no activity to report
+                    return;
+                }
+
+                // we've been accumulating function activity for the entire period
+                // publish this activity and reset
+                Metrics metrics = null;
+                lock (_lock)
+                {
+                    metrics = new Metrics
+                    {
+                        TotalTimeMS = (long)_intervalStopwatch.GetElapsedTime().TotalMilliseconds,
+                        ExecutionCount = FunctionExecutionCount,
+                        ExecutionTimeMS = FunctionExecutionTimeMS,
+                        IsAlwaysReady = IsAlwaysReady
+                    };
+
+                    FunctionExecutionTimeMS = FunctionExecutionCount = 0;
+                }
+
+                await PublishMetricsAsync(metrics);
+            }
+            finally
+            {
+                _intervalStopwatch = ValueStopwatch.StartNew();
+            }
+        }
+
+        private async void OnFunctionMetricsPublishTimer(object state)
+        {
+            await OnPublishMetrics();
+        }
+
+        private async Task PublishMetricsAsync(Metrics metrics)
+        {
+            string fileName = string.Empty;
+
+            try
+            {
+                bool metricsPublishEnabled = !string.IsNullOrEmpty(MetricsFilePath);
+                if (metricsPublishEnabled && !PrepareDirectoryForFile())
+                {
+                    return;
+                }
+
+                string metricsContent = JsonConvert.SerializeObject(metrics);
+                _logger.PublishingMetrics(metricsContent);
+
+                if (metricsPublishEnabled)
+                {
+                    fileName = $"{Guid.NewGuid().ToString().ToLower()}.json";
+                    string filePath = Path.Combine(MetricsFilePath, fileName);
+
+                    using (var streamWriter = _fileSystem.File.CreateText(filePath))
+                    {
+                        await streamWriter.WriteAsync(metricsContent);
+                    }
+                }
+            }
+            catch (Exception ex) when (!ex.IsFatal())
+            {
+                // TODO: consider using a retry strategy here
+                _logger.LogError(ex, $"Error writing metrics file '{fileName}'.");
+            }
+        }
+
+        private bool PrepareDirectoryForFile()
+        {
+            if (string.IsNullOrEmpty(MetricsFilePath))
+            {
+                return false;
+            }
+
+            // ensure the directory exists
+            _fileSystem.Directory.CreateDirectory(MetricsFilePath);
+
+            var metricsDirectoryInfo = _fileSystem.DirectoryInfo.FromDirectoryName(MetricsFilePath);
+            var files = metricsDirectoryInfo.GetFiles().OrderBy(p => p.CreationTime).ToList();
+
+            // ensure we're under the max file count
+            if (files.Count < _options.MaxFileCount)
+            {
+                return true;
+            }
+
+            // we're at or over limit
+            // delete enough files that we have space to write a new one
+            int numToDelete = files.Count - _options.MaxFileCount + 1;
+            var filesToDelete = files.Take(numToDelete).ToArray();
+
+            _logger.LogDebug($"Deleting {filesToDelete.Length} metrics file(s).");
+
+            foreach (var file in filesToDelete)
+            {
+                try
+                {
+                    file.Delete();
+                }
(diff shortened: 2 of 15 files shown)


src/WebJobs.Script.WebHost/Configuration/FlexConsumptionMetricsPublisherOptionsSetup.cs
```diff
@@ -0,0 +1,22 @@
...
+        private IEnvironment _environment;
+
+        public FlexConsumptionMetricsPublisherOptionsSetup(IEnvironment environment)
+        {
+            _environment = environment;
+        }
+
+        public void Configure(FlexConsumptionMetricsPublisherOptions options)
+        {
+            options.MetricsFilePath = _environment.GetEnvironmentVariable(EnvironmentSettingNames.FunctionsMetricsPublishPath);
```
mathewc, line 19: The metrics file path will come as part of container start context, so this setup needs to run after specialization which I do.

src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
```diff
@@ -174,15 +174,23 @@ internal static class LoggerExtension
         private static readonly Action<ILogger, string, string, Version, string, string, Exception> _minimumExtensionVersionNotSatisfied =
             LoggerMessage.Define<string, string, Version, string, string>(
             LogLevel.Error,
-            new EventId(334, nameof(MinimumExtensionVersionNotSatisfied)),
```
mathewc, line 177: Some previous mistakes made in ordering here

src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
```diff
@@ -72,6 +72,8 @@ public static class EnvironmentSettingNames
         public const string FunctionsPlatformConfigFilePath = "FUNCTIONS_PLATFORM_CONFIG_FILE_PATH";
         public const string TargetBaseScalingEnabled = "TARGET_BASED_SCALING_ENABLED";
         public const string WebsiteNodeDefaultVersion = "WEBSITE_NODE_DEFAULT_VERSION";
+        public const string FunctionsMetricsPublishPath = "FUNCTIONS_METRICS_PUBLISH_PATH";
+        public const string FunctionsAlwaysReadyInstance = "FUNCTIONS_ALWAYSREADY_INSTANCE";
```
mathewc, line 76: @balag0 This FUNCTIONS_ALWAYSREADY_INSTANCE env var will need to be in the container start context for Always Ready instances as discussed. #Resolved

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
```diff
@@ -0,0 +1,290 @@
...
+
+                    using (var streamWriter = _fileSystem.File.CreateText(filePath))
+                    {
+                        await streamWriter.WriteAsync(metricsContent);
+                    }
+                }
+            }
+            catch (Exception ex) when (!ex.IsFatal())
+            {
+                _logger.LogError(ex, "Error writing metrics file.");
```
mathewc: Considering whether I need to implement a retry strategy here

src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
```diff
@@ -174,15 +174,23 @@ internal static class LoggerExtension
...
 
         private static readonly Action<ILogger, string, string, string, string, Exception> _minimumBundleVersionNotSatisfied =
             LoggerMessage.Define<string, string, string, string>(
             LogLevel.Error,
-            new EventId(335, nameof(MinimumBundleVersionNotSatisfied)),
+            new EventId(337, nameof(MinimumBundleVersionNotSatisfied)),
             "Referenced bundle {bundleId} of version {bundleVersion} does not meet the required minimum version of {minimumVersion}. Update your extension bundle reference in host.json to reference {minimumVersion2} or later.");
 
+        private static readonly Action<ILogger, string, Exception> _publishingMetrics =
+            LoggerMessage.Define<string>(LogLevel.Debug, new EventId(3338, nameof(PublishingMetrics)), "{metrics}");
```
anandagopal6: Is this supposed to be 338 or 3338?

mathewc: Haha - fixed the others then made my own mistake :)

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
```diff
@@ -0,0 +1,290 @@
...
+
+        private bool PrepareDirectoryForFile(IFileSystem fileSystem)
+        {
+            if (string.IsNullOrEmpty(MetricsFilePath))
+            {
+                return false;
+            }
+
+            // ensure the directory exists
+            _fileSystem.Directory.CreateDirectory(MetricsFilePath);
```
kshyju, line 169: Use the parameter passed to this method (`fileSystem`) ?. Same for line 161

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
kshyju: The `CreateDirectory` API could throw for a [few reasons](https://learn.microsoft.com/en-us/dotnet/api/system.io.directory.createdirectory?view=net-6.0#system-io-directory-createdirectory(system-string)). Should we also log the parameter passed to it as well (`MetricsFilePath`) ?

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
```diff
@@ -0,0 +1,290 @@
...
+            {
+                try
+                {
+                    file.Delete();
+                }
+                catch (Exception ex) when (!ex.IsFatal())
+                {
+                    // best effort
+                    _logger.LogError(ex, "Error deleting metrics file.");
+                    success = false;
```
kshyju: Would there be a use case where we needed to delete 4 files and first 3 were successful and last one fails. So we end up returning `false` even though we deleted some files ( so we have space for new ones). Could that cause metrics failing to publish?

Instead of returning success in 192, should we consider executing the same code from 162 -168 ?

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
```diff
@@ -0,0 +1,290 @@
...
+        {
+            try
+            {
+                bool metricsPublishEnabled = !string.IsNullOrEmpty(MetricsFilePath);
+                if (metricsPublishEnabled && !PrepareDirectoryForFile(_fileSystem))
+                {
+                    return;
+                }
+
+                string metricsContent = JsonConvert.SerializeObject(metrics);
```
kshyju, line 140: Should we consider switching to the [ source gen version of Serialize](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/source-generation) from STJ ?

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
mathewc: Added a TODO for the future

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
mathewc: Removed the parameter

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
mathewc: Fixed

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
mathewc: Might consider that in the future as an optimization if necessary. I think this standard serialization is fine for now, because this isn't a hotpath - we're logging every 5 seconds or so.

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
```diff
@@ -0,0 +1,301 @@
...
+                    ActiveFunctionCount--;
+                }
+
+                if (ActiveFunctionCount == 0)
+                {
+                    // we're transitioning from active to inactive accumulate the elapsed time,
+                    // applying the minimum interval
+                    var elapsedMS = _instanceActivityStopwatch.GetElapsedTime().TotalMilliseconds;
+                    var duration = Math.Max(elapsedMS, _optionsMonitor.CurrentValue.MinimumActivityIntervalMS);
+                    FunctionExecutionTimeMS += (long)duration;
```
kshyju, line 253: Question for my learning:
What does the `FunctionExecutionTimeMS` prop represents? Based on the logic above, it looks like a super set of execution time for a bunch of executions.

Let's say we have 3 function executions.

- A started at 10:00:00 and finished at 10:00:40
- B started at 10:00:00 and finished at 10:00:20
- C started at 10:00:15 and finished at 10:00:45

So the logic above will generate 45 seconds as the value of `elapsedMS`. What does that value represent? Where & how the value of `FunctionExecutionTimeMS` eventually used (after it gets published)?

src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
balag0: We might also end up making this more generic like
InstanceType = "AlwaysReady" or "HTTP" or ..

but this is good for now. we can readjust later if needed

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
kshyju: Also what happens one execution is still in progress while `OnFunctionMetricsPublishTimer` is triggered?

src/WebJobs.Script.WebHost/Configuration/FlexConsumptionMetricsPublisherOptionsSetup.cs
balag0: I was expecting this to be part of the initial environment in placeholder mode since we already know where this pod is hosted. 
this also works but curious why we delay the injection of this config (in the future if we ever do metrics for non placeholder based containers this would have to change so wanted to understand this better)

also it would be good if we set this up as part of initial environment before customer code and add logic to prevent customer env, from overriding this later accidentally (control plane doesnt prevent customers from setting FUNCTIONS_METRICS_PUBLISH_PATH)

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
```diff
@@ -0,0 +1,301 @@
...
+                if (metricsPublishEnabled)
+                {
+                    fileName = $"{Guid.NewGuid().ToString().ToLower()}.json";
+                    string filePath = Path.Combine(MetricsFilePath, fileName);
+
+                    using (var streamWriter = _fileSystem.File.CreateText(filePath))
+                    {
+                        await streamWriter.WriteAsync(metricsContent);
+                    }
+                }
```
balag0, line 152: >} [](http://example.com/codeflow?start=16&length=1)

since we know this publisher is only enabled if required and if metricsPublishEnabled is not true then it should be a clear signal of an error 

may be we can add an else block and an error log here.
I think we should also incorporate this failed state in the health check endpoint (but that discussion can be part of health checks). #Resolved

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
balag0, line 253: >FunctionExecutionTimeMS [](http://example.com/codeflow?start=20&length=23)

we use += here since within the same 5 second window we might have multiple transitions from active to inactive to active to inactive?
 #Resolved

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
```diff
@@ -0,0 +1,301 @@
...
+
+                // we've been accumulating function activity for the entire period
+                // publish this activity and reset
+                Metrics metrics = null;
+                lock (_lock)
+                {
+                    metrics = new Metrics
+                    {
+                        TotalTimeMS = (long)_intervalStopwatch.GetElapsedTime().TotalMilliseconds,
```
balag0, line 106: curious why we track this interval, is it check for any gaps in receiver side?

basically we just need to know in every 5 second interval
-how long was atleast 1 function active
right? #Resolved

balag0: oh i see. is it for always ready?

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
```diff
@@ -0,0 +1,301 @@
...
+            // ensure we're under the max file count
+            if (files.Count < _optionsMonitor.CurrentValue.MaxFileCount)
+            {
+                return true;
+            }
+
+            // we're at or over limit
+            // delete enough files that we have space to write a new one
+            int numToDelete = files.Count - _optionsMonitor.CurrentValue.MaxFileCount + 1;
+            var filesToDelete = files.Take(numToDelete).ToArray();
```
balag0, line 183: did we also consider the reader deleting the file once it has processed it? #Resolved

src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
mathewc: To generalize to a setting like FUNCTIONS_INSTANCE_TYPE I'd have to be confident that we actually will have more than two different distinct types. For the example you gave, it might be fine for there to just be a corresponding FUNCTION_HTTP_INSTANCE setting to capture that scenario. Or maybe no setting is needed at all for "http only" and that's transferred via function include list.

src/WebJobs.Script.WebHost/Configuration/FlexConsumptionMetricsPublisherOptionsSetup.cs
mathewc: I wasn't explicitly trying to delay injection of this. Definitely if this can be established in placeholder mode, makes sense to have it then. Likely it'd behave the same way as CONTAINER_NAME in Linux Consumption, which is also established in the placeholder.

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
mathewc: Yes, it's for AlwaysReady :) And generally it's just being precise about the duration of the metering interval. While I could have hard coded it to be the timer interval, I wanted it to be more precise in the face of timer slowdowns, etc.

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
mathewc: I chose to let the publisher to run in this mode to allow us to phase this into the runtime w/o breaking changes or starting to generate a bunch of errors for Flex Apps before the rest of the pipeline is hooked up.

Note that during startup above I do already log MetricsFilePath, so we can write queries to look for problems. We'll also have alerting for production apps.

If that's not sufficient, perhaps we add the error later after everything else is done?

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
mathewc: Yes, the reader will be deleting files when they're "consumed". This code is about making space to write a single new file. Do you see an issue here?

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
mathewc: Yes, as you say FunctionExecutionTimeMS is just the total time during a metering window where the host was running one or more functions. I tried to explain that in the [Metrics.ExecutionTimeMS](https://github.com/Azure/azure-functions-host/blob/08ab9671fb783d3d553b27a16253939d6d287f28/src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs#L286) property doc.

For your second question, OnPublishMetrics resets the activity timer each interval, so tracking starts anew.

mathewc: Correct. Consider a series of non-overlapped invocations. We'll be transitioning multiple times active/inactive.

src/WebJobs.Script.WebHost/Configuration/FlexConsumptionMetricsPublisherOptionsSetup.cs
mathewc: I've removed the post specialization re-initialization of this value. Therefore, because the publisher (like other metrics publishers) is a WebHost level level service created placeholder mode and the options are cached on construction, we'll be using the initial placeholder mode state.

kshyju (approved):

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
balag0: as long as we have something we can use to query/alert, we should be good.

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
balag0: so both writer and reader are deleting files?
its the same set of files ?

mathewc: Yes same set of files. Note that cleanup on this producer side only kicks in by default if there's 180 unprocessed files (15 minutes worth of files). That'd happen only if the consumer was stalled for a long period of time. When it comes back online you're right that the possibility a few conflicts might exist but they'd resolve quickly. E.g. consumer might list the directory and see 180 files to process, and while processing them it might find one or two were deleted. That's fine I think. The bulk of the files would be processed and we'd be back to steady state.

balag0: got it. thanks.

balag0 (approved):
:shipit:

PR #9510, opened 2023-09-05
[pack][bug] Do not restart worker channels during requests to get or update function metadata

### LINKED ISSUE #9501: In-portal editing leads to "Did not find any initialized language workers"
After editing a consumption function in the portal, when the function is invoked, we see a 500 error: `Unhandled host exception: Did not find any initialized language workers`

Investigate why this is happening and fix the bug.

**Repro:**

1. Create a python v1 app in the portal (this issue also happens with the node worker too)
    - Linux; Consumption
2. Create a HttpTrigger in the portal
3. Invoke the function - no issues
4. Edit the function in the portal and save
5. Invoke the function - should get 500 error

> Related CRI: 415987820

liliankasem:

resolves #9501

### Pull request checklist

### Additional information

This bug occurs when we make any request to the functions API to get or update function metadata. In this scenario, with worker indexing enabled (using the WorkerFunctionMetadataProvider), we shutdown the worker channel and create a new one that we do not setup the buffers for ([code](https://github.com/Azure/azure-functions-host/blob/88b610c61a6dc39e053f5fe7f902204e06aa1b41/src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs#L70-L80)).

After creating and initializing a new channel, we are not calling SetupFunctionInvocationBuffers ([code](https://github.com/Azure/azure-functions-host/blob/50788a70d720400dd0cbbd34aba0a06f60c941e6/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs#L474)) which adds "InvocationBuffersInitialized" to the gRPC channel _state. Because of this, during an invocation when we are checking for "initializedWorkers" ([code](https://github.com/Azure/azure-functions-host/blob/50788a70d720400dd0cbbd34aba0a06f60c941e6/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs#L435)), we get null because even though the channel has been initialized, we have not setup the invocation buffers. So the number of workers is 0 leading to "did not find any initialized language workers" ([code](https://github.com/Azure/azure-functions-host/blob/50788a70d720400dd0cbbd34aba0a06f60c941e6/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcherLoadBalancer.cs#L27)).

The gRPC channel's state has to have "Initialized" and "InvocationBuffersInitialized" for it to be deemed ready for invocations ([code](https://github.com/Azure/azure-functions-host/blob/50788a70d720400dd0cbbd34aba0a06f60c941e6/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs#L271)).

When worker indexing is disabled (using the HostFunctionMetadataProvider), we just parse the metadata and return it - we do not shut down the channel or create a new one.

There are two bug fixes this PR is addressing:

1. Do not shut down the worker channel when an API request is made to update or get the function metadata
    - The host will now just return the function metadata
    - If a file is changed, the FileMonitor will trigger a host restart
2. Fix an issue where host creates a new worker channel that does not get properly initialized
    - The host will now restart if it's in a running state and there are no channels available

### Testing

I've added an E2E test for this but have also tested:

1. prod: PrivateSiteExtension deployment with dotnet and dotnet-isolated
2. prod: Custom linux deployment with python
3. Specialization


release_notes.md
```diff
@@ -9,3 +9,8 @@
 - Update Node.js Worker Version to [3.8.1](https://github.com/Azure/azure-functions-nodejs-worker/releases/tag/v3.8.1)
 - Update PowerShell 7.4 Worker Version to [4.0.2930](https://github.com/Azure/azure-functions-powershell-worker/releases/tag/v4.0.2930)
 - Update WebJobsScriptHostService to remove hardcoded sleep during application shut down (#9520)
+- Bug fix: Do not restart a worker channel when an API request is made to update or get the function metadata
```
soninaren: Technically this is only true when there is no change in config.

src/WebJobs.Script.WebHost/Controllers/FunctionsController.cs
```diff
@@ -49,16 +50,25 @@ public FunctionsController(IWebFunctionsManager functionsManager, IWebJobsRouter
...
-            return Ok(result);
+            try
+            {
+                var result = await _functionsManager.GetFunctionsMetadataAsync(includeProxies, forceRefresh: false);
+                return Ok(result);
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex.Message);
+                Console.WriteLine(ex);
```
soninaren: Not sure why are doing `Console.WriteLine` here. I think logging the error is all we need.

src/WebJobs.Script.WebHost/Controllers/FunctionsController.cs
```diff
@@ -49,16 +50,25 @@ public FunctionsController(IWebFunctionsManager functionsManager, IWebJobsRouter
...
+            }
         }
 
         [HttpGet]
         [Route("admin/functions/{name}")]
         [Authorize(Policy = PolicyNames.AdminAuthLevel)]
         public async Task<IActionResult> Get(string name)
         {
-            (var success, var function) = await _functionsManager.TryGetFunction(name, Request);
+            (bool success, FunctionMetadataResponse function) = await _functionsManager.TryGetFunction(name, Request, forceRefresh: false);
```
soninaren: Should this call be in try catch block as well?

soninaren (changes requested):
Have added some comments. Looks like some additional commits were added to your PR by mistake. Can you take a look.

liliankasem:
Just FYI there's some WIP stuff whilst I fix the pipeline; it's not failing locally so I'm trying to debug the pipeline which is not fun ☠️

Edit: this is magically fixed for unknown reasons that I spent a whole day trying to understand

src/WebJobs.Script.WebHost/Controllers/FunctionsController.cs
liliankasem: trying to debug the pipeline, will be undone!

src/WebJobs.Script.WebHost/Controllers/FunctionsController.cs
liliankasem: Yeah I think a lot of these methods are missing proper error handling; I worry about breaking changes if we change the behaviour of these APIs  i.e if people are depending on a 500 but we start sending a 404 or other HTTP status'. I can still add the try catch and still return 500, but at least we will have more info about failure

src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
```diff
@@ -74,12 +77,25 @@ public async Task<FunctionMetadataResult> GetFunctionMetadataAsync(IEnumerable<R
...
+                // Start up GRPC channels if they are not already running.
                 if (channels?.Any() != true)
                 {
-                    await _channelManager.InitializeChannelAsync(workerConfigs, _workerRuntime);
+                    if (_scriptHostManager.State is ScriptHostState.Default
+                        || _scriptHostManager.State is ScriptHostState.Starting
+                        || _scriptHostManager.State is ScriptHostState.Initialized)
+                    {
+                        // We don't need to restart if the host hasn't even been created yet.
+                        await _channelManager.InitializeChannelAsync(workerConfigs, _workerRuntime);
```
brettsam, line 89: This is definitely better than before, but it still seems like there's a chance for races... can we add some logging so it's clear what we're doing here and below (with the restart)? Just in case we need to debug in the future, it'd be good to know the path we took through this code...

brettsam (approved):

soninaren (approved):

PR #9511, opened 2023-09-06
Improve error logging for scenario: Managed identity for storage + zipdeploy + Linux consumption

### LINKED ISSUE #9397: No-error-message: using Managed identity for storage + zipdeploy + Linux consumption
Currently it is not supported to use standard approach for managed identity on functions storage account in Linux Consumption with zip deploy.

This issue tracks the lack of error message when the user/deployment is in this state, because all you see is `0 Functions started` in the application insights logs.

#### Repro steps

To repro:
1) clone and run this hello world python v2 sample
[functions/starters/python at main · paulyuk/functions · GitHub](https://github.com/paulyuk/functions/tree/main/starters/python)

2) `azd up` (provisions the RG, does a zip package and zipdeploy)

Note the http function works now at route /api/http
 

3) Verify Identity - System Managed Identity on the Functionapp (already done ) is enabled

4) Add IAM roles to storage account, in particular the Storage Blob Data Owner and grant to the system managed identity of the function

Note the function *still works fine, but will break in the next step:

5) Change the connection string to use managed identity instead of connectionstring with key/secret

From

  {

    "name": "AzureWebJobsStorage",

    "value": "..shh its a secret..",

    "slotSetting": false

  }

 

To

  {

    "name": "AzureWebJobsStorage__accountname",

    "value": "stgvcf47zjvz27e",

    "slotSetting": false

  },

6) restart the function and test /api/http

 

Result:
Function is down / 404
App Insights lights show "1 functions found; 0 functions loaded".

#### Expected behavior

There is an error message describing this unsupported configuration of Linux Consumption with zipdeploy and managed identity, and see <url> for more info and workarounds (namely server build). 

-->

#### Known workarounds

Use server build (non tools based) methods described here

https://learn.microsoft.com/en-us/azure/azure-functions/functions-deployment-technologies?tabs=windows

surgupta-msft:

Issue #9397 

Currently it is not supported to use the standard approach for managed identity on functions storage account in Linux Consumption with zip deploy. This PR adds error logs for such scenario as currently the only error we get in applicationInsights is `0 Functions loaded`.

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
+++ b/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
@@ -234,7 +234,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
                 _environment.IsLinuxConsumptionOnAtlas() &&
                 !_environment.IsManagedAppEnvironment())
             {
-                _logger.LogInformation($"App has no functions deployed and remote build is not available since the app was not provisioned with valid {AzureWebJobsSecretStorage} connection string.");
+                _logger.LogError($"App has no functions deployed since the app was not provisioned with valid {AzureWebJobsSecretStorage} connection string.");
             }
         }
 
--- a/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
+++ b/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
@@ -224,17 +224,6 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
             }
         }
 
-        private bool IsAppUsingManagedIdentity()
-        {
-            // If AzureWebJobsStorage__accountName is set, we are using identities.
-            if (!string.IsNullOrEmpty(_environment.GetEnvironmentVariable(AzureWebJobsStorageAccountName)))
-            {
-                return true;
-            }
-
-            return false;
-        }
-
         private void ValidateLinuxSKUConfiguration()
         {
             var websiteRunFromPackageValue = _environment.GetEnvironmentVariable(AzureWebsiteRunFromPackage);
@@ -243,16 +232,9 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
             if (string.IsNullOrEmpty(websiteRunFromPackageValue) &&
                 string.IsNullOrEmpty(scmRunFromPackageValue) &&
                 _environment.IsLinuxConsumptionOnAtlas() &&
-                !_environment.IsManagedAppEnvironment() &&
-                IsAppUsingManagedIdentity())
+                !_environment.IsManagedAppEnvironment())
             {
-                _logger.LogWarning($"App has no functions deployed and remote build is not available since the app was not provisioned with valid {AzureWebJobsSecretStorage} connection string.");
-
-                DiagnosticEventLoggerExtensions.LogDiagnosticEventInformation(
-                            _logger,
-                            DiagnosticEventConstants.UnsupportedZipDeploymentOnLinuxConsumptionwithMSIErrorCode,
-                            $"App has no functions deployed and remote build is not available since the app was not provisioned with valid {AzureWebJobsSecretStorage} connection string.",
-                            DiagnosticEventConstants.UnsupportedZipDeploymentOnLinuxConsumptionwithMSIErrorCodeLink);
+                _logger.LogInformation($"App has no functions deployed and remote build is not available since the app was not provisioned with valid {AzureWebJobsSecretStorage} connection string.");
             }
         }
 
--- a/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
+++ b/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
@@ -234,7 +234,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
                 _environment.IsLinuxConsumptionOnAtlas() &&
                 !_environment.IsManagedAppEnvironment())
             {
-                _logger.LogError($"App has no functions deployed since the app was not provisioned with valid {AzureWebJobsSecretStorage} connection string.");
+                _logger.LogError($"Unable to load the functions payload since the app was not provisioned with valid {AzureWebJobsSecretStorage} connection string.");
             }
         }
 
--- a/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
+++ b/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
@@ -237,19 +237,21 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
 
         private void ValidateLinuxSKUConfiguration()
         {
-            var url = _environment.GetEnvironmentVariable(ScmRunFromPackage);
+            var websiteRunFromPackageValue = _environment.GetEnvironmentVariable(AzureWebsiteRunFromPackage);
+            var scmRunFromPackageValue = _environment.GetEnvironmentVariable(ScmRunFromPackage);
 
-            if (string.IsNullOrEmpty(url) &&
+            if (string.IsNullOrEmpty(websiteRunFromPackageValue) &&
+                string.IsNullOrEmpty(scmRunFromPackageValue) &&
                 _environment.IsLinuxConsumptionOnAtlas() &&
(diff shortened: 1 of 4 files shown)


src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -223,6 +224,22 @@ private void CheckFileSystem()
             }
         }
 
+        private void ValidateLinuxSKUConfiguration()
+        {
```
surgupta-msft, line 228: The logs from repro app shows that the indicator of error is below line - https://github.com/Azure/azure-functions-host/blob/88b610c61a6dc39e053f5fe7f902204e06aa1b41/src/WebJobs.Script.WebHost/Configuration/ScriptApplicationHostOptionsSetup.cs#L97

Added similar check here to identify if the app has - zip deploy + MSI + Linux Consumption to log the error and raise the diagnostic event.

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -223,6 +224,22 @@ private void CheckFileSystem()
...
+            if (string.IsNullOrEmpty(url) && _environment.IsLinuxConsumptionOnAtlas() && _environment.IsManagedAppEnvironment())
+            {
+                _logger.LogWarning($"Zip deployment is not supported on a Linux Consumption app configured to use Managed Identity. Functions may not be indexed correctly.");
+
+                DiagnosticEventLoggerExtensions.LogDiagnosticEventInformation(
+                            _logger,
+                            DiagnosticEventConstants.UnsupportedZipDeploymentOnLinuxConsumptionwithMSIErrorCode,
+                            "Zip deployment is not supported on a Linux Consumption app configured to use Managed Identity.",
+                            DiagnosticEventConstants.UnsupportedZipDeploymentOnLinuxConsumptionwithMSIErrorCodeLink);
+            }
```
surgupta-msft, line 238: @brettsam Do we want to stop the application here as we are doing in `CheckFileSystem()` (link below)? or is it sufficient to provide a warning via diagnostic event to the user? Wanted to double check so that we don't stop the application where it is not needed.

https://github.com/Azure/azure-functions-host/blob/88b610c61a6dc39e053f5fe7f902204e06aa1b41/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs#L221

src/WebJobs.Script/Diagnostics/DiagnosticEventConstants.cs
```diff
@@ -13,5 +13,8 @@ internal static class DiagnosticEventConstants
 
         public const string SasTokenExpiringErrorCode = "AZFD0006";
         public const string SasTokenExpiringErrorHelpLink = "https://go.microsoft.com/fwlink/?linkid=2244092";
+
+        public const string UnsupportedZipDeploymentOnLinuxConsumptionwithMSIErrorCode = "AZFD0007";
+        public const string UnsupportedZipDeploymentOnLinuxConsumptionwithMSIErrorCodeLink = "<ToDo>";
```
surgupta-msft: ToDo - Add document in [azure-docs repo](https://github.com/MicrosoftDocs/azure-docs/blob/main/articles/azure-functions/errors-diagnostics/diagnostic-events/azfd0006.md) and link it here before merging.

Sample changes for docs here - https://github.com/MicrosoftDocs/azure-docs/pull/114381/files

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
fabiocav: You'd need to check that this is not on Managed App Environment (ACA), so the check would need to be `!_environment.IsManagedAppEnvironment()`

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -223,6 +224,22 @@ private void CheckFileSystem()
             }
         }
 
+        private void ValidateLinuxSKUConfiguration()
+        {
+            var url = _environment.GetEnvironmentVariable(ScmRunFromPackage);
+
+            if (string.IsNullOrEmpty(url) && _environment.IsLinuxConsumptionOnAtlas() && _environment.IsManagedAppEnvironment())
+            {
+                _logger.LogWarning($"Zip deployment is not supported on a Linux Consumption app configured to use Managed Identity. Functions may not be indexed correctly.");
```
fabiocav: We need to perform the check to validate MI is being used.

fabiocav:
@pragnagopa / @balag0 can you please provide some feedback as @surgupta-msft iterates on this?

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
surgupta-msft: Added a similar check as used here - https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.WebHost/Health/TokenExpirationService.cs#L62-#L65. Is this sufficient for MI check?

```csharp
private void AnalyzeSasTokenInUri()
        {
            // If AzureWebJobsStorage__accountName is set, we are using identities and don't need to check for SAS token expiration
            if (!string.IsNullOrEmpty(_environment.GetEnvironmentVariable($"{EnvironmentSettingNames.AzureWebJobsSecretStorage}__accountName")))
            {
```

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -223,6 +224,36 @@ private void CheckFileSystem()
             }
         }
 
+        private bool IsAppUsingManagedIdentity()
+        {
+            // If AzureWebJobsStorage__accountName is set, we are using identities.
+            if (!string.IsNullOrEmpty(_environment.GetEnvironmentVariable(AzureWebJobsStorageAccountName)))
+            {
+                return true;
```
pragnagopa: just curious - does this check work across SKUs? Please verify on windows dedicated SKU / EP SKU via private site extension - https://github.com/Azure/azure-functions-host/wiki/Deploying-the-Functions-runtime-as-a-private-site-extension

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -223,6 +224,36 @@ private void CheckFileSystem()
...
+        private void ValidateLinuxSKUConfiguration()
+        {
+            var url = _environment.GetEnvironmentVariable(ScmRunFromPackage);
+
+            if (string.IsNullOrEmpty(url) &&
+                _environment.IsLinuxConsumptionOnAtlas() &&
+                !_environment.IsManagedAppEnvironment() &&
+                IsAppUsingManagedIdentity())
+            {
+                _logger.LogWarning($"Zip deployment is not supported on a Linux Consumption app configured to use Managed Identity. Functions may not be indexed correctly.");
```
pragnagopa: Also check that app setting `WEBSITE_RUN_FROM_PACKAGE` does not exist

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
surgupta-msft: Yes I used these steps - [link1](https://learn.microsoft.com/en-us/azure/azure-functions/functions-reference?tabs=blob#connecting-to-host-storage-with-an-identity), [link2](https://techcommunity.microsoft.com/t5/apps-on-azure-blog/use-managed-identity-instead-of-azurewebjobsstorage-to-connect-a/ba-p/3657606) with a windows dedicated .Net App to configure identity and the logic to check appsetting `AzureWebJobsStorage__accountName` for Managed identity is working. Let me know if you have concerns on this or there is another recommended way to detect MI case reliably.

There are other techniques mentioned [here](https://learn.microsoft.com/en-us/azure/azure-functions/functions-reference?tabs=blob#common-properties-for-identity-based-connections) for different cases like userAssigned, SystemAssigned. Should we include all of them in this check?

Still discussing with Bala offline to decide if we need this MI check here or remove it completely. Then I will circle back on how to reliably detect this case.

pragnagopa (approved):

liliankasem (approved):

PR #9514, opened 2023-09-07
Sending command line args with functions- prefix to prevent conflicts

### LINKED ISSUE #9504: Send command line args with non-generic names to language worker
The command line arguments we currently send to language worker process has generic names (HOST, PORT etc). This would cause conflict with other environment variables in certain environments. For example, dotnet isolated in Linux + VNET is one of those cases where the PORT gets overwritten when the customer calls "ConfigureAppConfiguration" in the main method (as shown in https://github.com/Azure/azure-functions-dotnet-worker/issues/1547)

We want to send these args with FUNCTIONS_ prefix. Initially we will send both (ex: HOST and FUNCTIONS_HOST) and once all language workers adopts to the new arguments & we are confident that no existing apps are using the old approach, we shall consider retiring the old ones.

Language workers are advised to use the new command line arguments if present, and fallback to existing one.

We also decided to combine the host and port part to a single argument "functions-uri".

For example,  for dotnet-isolated worker, we will do something like below

```csharp
grpcWorkerStartupOption.RequestId= config["functions-requestid"] ?? config["requestid"];
```
where `config` is an instance of `IConfiguration` which contains all command line arguments. We will first check `FUNCTIONS_HOST` entry is present and use it's value if present, else will use `host`.

#### Issues tracking fixes on language workers side to support the new arguments (to not throw while parsing)

- [x] https://github.com/Azure/azure-functions-powershell-worker/issues/992
- [x] https://github.com/Azure/azure-functions-dotnet-worker/issues/1900
- [x] https://github.com/Azure/azure-functions-java-worker/issues/724
- [x] https://github.com/Azure/azure-functions-python-worker/issues/1314

kshyju:
Fixes #9504

During initial testing, we identified 3 language workers were not able to handle the extra/new command line arguments sent to them. Fixed those on the language worker sides and host was updated to consume the new version of language worker nuget package with the fix.

1. PowerShell  - https://github.com/Azure/azure-functions-host/pull/9528
2. Java  - https://github.com/Azure/azure-functions-host/pull/9544
3. Python - https://github.com/Azure/azure-functions-host/pull/9556

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script/Workers/Rpc/RpcWorkerContext.cs
+++ b/src/WebJobs.Script/Workers/Rpc/RpcWorkerContext.cs
@@ -44,7 +44,11 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
 
         public override string GetFormattedArguments()
         {
-            return $" --host {ServerUri.Host} --port {ServerUri.Port} --workerId {WorkerId} --requestId {RequestId} --grpcMaxMessageLength {MaxMessageLength}";
+            // We are adding a second copy of the commandline arguments with the FUNCTIONS_ prefix.
+            // Once all workers starts using the new parameters, we can retire the old ones.
+            // Tracking item: https://github.com/Azure/azure-functions-host/issues/9504
+
+            return $" --host {ServerUri.Host} --port {ServerUri.Port} --workerId {WorkerId} --requestId {RequestId} --grpcMaxMessageLength {MaxMessageLength} --FUNCTIONS_HOST {ServerUri.Host} --FUNCTIONS_PORT {ServerUri.Port} --FUNCTIONS_WORKERID {WorkerId} --requestId {RequestId} --FUNCTIONS_GRPCMAXMESSAGELENGTH {MaxMessageLength}";
         }
     }
 }
--- a/src/WebJobs.Script/Workers/Rpc/RpcWorkerContext.cs
+++ b/src/WebJobs.Script/Workers/Rpc/RpcWorkerContext.cs
@@ -47,7 +47,7 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
             // Adding a second copy of the commandline arguments with the "functions-" prefix to prevent any conflicts caused by the existing generic names.
             // Language workers are advised to use the "functions-" prefix ones and if not present fallback to existing ones.
 
-            return $" --host {ServerUri.Host} --port {ServerUri.Port} --workerId {WorkerId} --requestId {RequestId} --grpcMaxMessageLength {MaxMessageLength} --functions-uri {ServerUri.AbsoluteUri} --functions-workerid {WorkerId} --functions-requestid {RequestId} --functions-grpcmaxmessagelength {MaxMessageLength}";
+            return $" --host {ServerUri.Host} --port {ServerUri.Port} --workerId {WorkerId} --requestId {RequestId} --grpcMaxMessageLength {MaxMessageLength} --functions-uri {ServerUri.AbsoluteUri} --functions-worker-id {WorkerId} --functions-request-id {RequestId} --functions-grpc-max-message-length {MaxMessageLength}";
         }
     }
 }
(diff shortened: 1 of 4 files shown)


liliankasem (approved):

src/WebJobs.Script/Workers/Rpc/RpcWorkerContext.cs
```diff
@@ -44,7 +44,10 @@ internal class RpcWorkerContext : WorkerContext
 
         public override string GetFormattedArguments()
         {
-            return $" --host {ServerUri.Host} --port {ServerUri.Port} --workerId {WorkerId} --requestId {RequestId} --grpcMaxMessageLength {MaxMessageLength}";
+            // Adding a second copy of the commandline arguments with the "functions-" prefix to prevent any conflicts caused by the existing generic names.
+            // Language workers are advised to use the "functions-" prefix ones and if not present fallback to existing ones.
+
+            return $" --host {ServerUri.Host} --port {ServerUri.Port} --workerId {WorkerId} --requestId {RequestId} --grpcMaxMessageLength {MaxMessageLength} --functions-uri {ServerUri.AbsoluteUri} --functions-worker-id {WorkerId} --functions-request-id {RequestId} --functions-grpc-max-message-length {MaxMessageLength}";
```
surgupta-msft, line 50: Who will use the arguments without "functions-" prefix? only Host or any other components?

surgupta-msft (approved):

src/WebJobs.Script/Workers/Rpc/RpcWorkerContext.cs
kshyju: The older versions of language workers. Ex: dotnet isolated function apps deployed using older versions of worker package ( which does not have the new logic to read from the new arguments).

mathewc: It's [this](https://github.com/Azure/azure-functions-host/blob/8cd901eb9b9879b0c2ecce07663b68aa3f238686/src/WebJobs.Script/Workers/ProcessManagement/DefaultWorkerProcessFactory.cs#L55) code that is the issue right? It's taking environment variables and overwriting args. Can't we just fix this there by not overwriting these special command line args (host/port/workerId/etc.)? That way we're not bloating this command line with duplicates that in the future may also be subject to the same issue.

kshyju: Synced offline with @mathewc . We still need this fix to address the issue dotnet language worker was running into because dotnet language worker uses IConfiguration to get the command line args and customer can add environment variables as a configuration source (after commandline arguments as source) which will cause env variable value to be returned with trying to read `configuration["port"] because IConfiguration returns the value from the last provider if exist.

Francisco-Gamino (approved):
@kshyju -- The code change LGTM.

Regarding the old command arguments, should we create a work item to remove support for these (`host`, `port`, `workerId`, `requestId`, and `grpcMaxMessageLength`) in the future? IMHO, the language workers should only use the new command line arguments going forward.

kshyju:
> @kshyju -- The code change LGTM.
> 
> Regarding the old command arguments, should we create a work item to remove support for these (`host`, `port`, `workerId`, `requestId`, and `grpcMaxMessageLength`) in the future? IMHO, the language workers should only use the new command line arguments going forward.

Yes, If the language worker is deployed with a host version, then please consider updating the language worker to remove reading the old params).

Francisco-Gamino:
> > @kshyju -- The code change LGTM.
> > Regarding the old command arguments, should we create a work item to remove support for these (`host`, `port`, `workerId`, `requestId`, and `grpcMaxMessageLength`) in the future? IMHO, the language workers should only use the new command line arguments going forward.
> 
> Yes, If the language worker is deployed with a host version, then please consider updating the language worker to remove reading the old params).

@kshyju -- I was also asking about removing support for the old command line arguments from the Functions Host. Thoughts?

kshyju:
> Functions

Dotnet isolated apps which were deployed with older version of .net worker nuget packages (which does not have the code to read the new args) will fail. So It is not safe for us to delete the old args at this point.

PR #9520, opened 2023-09-08
Update WebJobsScriptHostService to remove hardcoded sleep during shut down

### LINKED ISSUE #8681: Can this workaround be removed?
Hi, the code referenced below says ` // Workaround until https://github.com/Azure/azure-functions-host/issues/7188 is addressed.`. It seems that #7188 has since been fixed, can this sleep now be removed? It would be very helpful for us to not keep our pods running longer than they have to.

https://github.com/Azure/azure-functions-host/blob/b3d90c43e7a510ab0930202a3ed924249bb7e66c/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs#L799

liliankasem:

resolves #8681

### Pull request checklist

### Additional information

Antares does not rely on the flag `FUNCTIONS_ENABLE_DRAIN_ON_APP_STOPPING` – so we can safely assume that this code path is only used in LIMA. Speaking to the LIMA team we learnt that the time for sleep is inconsequential for the LIMA scenario and therefore should be safe to remove.

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -8,3 +8,4 @@
 - Increased maximum HTTP request content size to 210000000 Bytes (~200MB)
 - Update Node.js Worker Version to [3.8.1](https://github.com/Azure/azure-functions-nodejs-worker/releases/tag/v3.8.1)
 - Update PowerShell 7.4 Worker Version to [4.0.2930](https://github.com/Azure/azure-functions-powershell-worker/releases/tag/v4.0.2930)
+- Update WebJobsScriptHostService to remove hardcoded sleep during restart (#9520)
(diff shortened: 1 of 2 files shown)


fabiocav (approved):

chandrod (approved):

mathewc (approved):

mathewc:
Are we backporting this? I see you said no above. https://github.com/Azure/azure-functions-host/blob/3866b07f5779e68192a695db819aa7021d5afae4/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs#L796

fabiocav:
@mathewc we're no longer backporting any changes. Do you think this change should be considered?

mathewc:
I agree that we don't need to do a special v3 release for this, but if we end up needing to do one in the future for other reasons, it'd be good to have this change in there as well I think.

liliankasem:
> I agree that we don't need to do a special v3 release for this, but if we end up needing to do one in the future for other reasons, it'd be good to have this change in there as well I think.

That's fair, there's no harm in backporting any who

liliankasem:
Backport here: https://github.com/Azure/azure-functions-host/pull/9521

PR #9523, opened 2023-09-08
Implement host configuration property for handling pre-cancelled invocation requests

### LINKED ISSUE #9387: Regular Microsoft.Azure.WebJobs.Script.Description.FunctionInvocationCanceledException "The invocation request with id '<id>' was canceled before the request was sent to the worker."
Function app regularly throws FunctionInvocationCanceledException when being called by other service. 

From the other service we can see that the request takes a long time and is then cancelled after 10+ seconds. In the most recent occurrences of this we get requests less than a second before that we receive in the function app code and respond to as expected, but the second request that times out never reaches the function code for our function app. As far as we can tell this is not because of cold starts because we have the first request that completes as expected less than a second before the request that fails and throws the error.

Please provide the following:

- Timestamp: 
- Function App version: 4.23.0.20886
- Function App name: 
- Function name(s) (as appropriate):
- Invocation ID: 8070e284-acf7-404a-afde-f8247eae8b6f
- Region: North Europe

#### Related information 

* Programming language used: C#

liliankasem:

resolves #9387

### Pull request checklist

    * [x] Otherwise: Documentation issue linked to PR (in progress)

### Additional information

Recently there has been an uptick of GitHub issues and CRIs related to the `FunctionInvocationCanceled` exception introduced in [this PR](https://github.com/Azure/azure-functions-host/pull/9158). This is because the PR made changes that now expose existing issues that were hidden prior and thus has lead to an increased volume of invocations failures with this exception. Of the cases investigated, approximately 90% of them are due to client disconnects.

After some brainstorming, the team has decided that we should:

a) Change the log level of the "Cancellation has been requested" message to be information. Add to provide a link to documentation to explain to customers what could be causing a cancellation before an invocation is sent to the worker.

b) To make this behaviour configurable. By default, if cancellation has been requested before we send an invocation to the worker, we will fail the invocation with a `FunctionInvocationCanceled` exception (this is what happens today). If a worker handles the InvocationCancel capability, we will not cancel and will send the invocation to the worker. We are also adding a host configuration property to override that behaviour where if the worker handles InvocationCancel AND the property is set to false, we will cancel the result source and not send the cancelled invocation to the worker.


jviau:
I am not sure I like the idea of option (b). If we look at our host as a reverse-proxy in this case (which we essentially are, especially in the AspNetCore efforts), I don't think it is standard to forward a canceled request to the next node (the worker in this case). To me, adding this capability seems like extra effort on our end for something that is arguably a bad pattern for a customer to even use.

Do we have customer demand for this capability? If so, what are they wanting to do with that disconnected request?

cjaliaga:
> Do we have customer demand for this capability? If so, what are they wanting to do with that disconnected request?

We've seen some customers implementing a (not properly) fire-and-forget pattern with Http Triggers and on those scenarios the response is not important for them. I do agree this is a bad pattern and we have other options on the platform, such Durable, to implement the same in a better a safer way.

I would let the new behavior as default, not sending pre-cancelled invocations to the worker.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -746,12 +748,19 @@ internal async Task SendInvocationRequest(ScriptInvocationContext context)
...
                 }
 
-                // do not send an invocation request if cancellation has been requested
                 if (context.CancellationToken.IsCancellationRequested)
                 {
-                    _workerChannelLogger.LogWarning("Cancellation has been requested. The invocation request with id '{invocationId}' is canceled and will not be sent to the worker.", invocationId);
-                    context.ResultSource.TrySetCanceled();
-                    return;
+                    _workerChannelLogger.LogInformation("Cancellation was requested prior to the invocation request ('{invocationId}') being sent to the worker. " +
+                                                        "For more information see https://aka.ms/AAmnbwj", invocationId);
```
brettsam: is this the right aka link? can we make it more readable? (azure-functions-cancellations, or something similar?)

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -746,12 +748,19 @@ internal async Task SendInvocationRequest(ScriptInvocationContext context)
...
+
+                    // If a worker is not configured to handle already cancelled invocations, do not
+                    // send an invocation request to the worker.
+                    if (!_isHandlesCanceledInvocationRequestsCapabilityEnabled)
```
brettsam: I agree with @jviau on this one -- it feels wrong to send a known-cancelled function invocation to the worker. It was cancelled for a reason (host is shutting down, client disconnected, etc)... and seems more confusing to let a worker start processing that invocation when it may never complete.

jviau:
> We've seen some customers implementing a (not properly) fire-and-forget pattern with Http Triggers and on those scenarios the response is not important for them.

I think this is a pattern we explicitly do **not** want to support. Customer code is meant to run within the context of a trigger and not leak outside it. It is not just a bad pattern in your example, but an explicitly unsupported pattern.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
liliankasem: Can give it a proper display name; currently pointing to just the cancellation docs but will update the link in aka.ms to point to the new docs once released: 

Will update this to: https://aka.ms/azure-functions-cancellations

kshyju: FWIW, @mattchenderson created few starts with "https://aka.ms/azure-functions/" . So might consider using the same format:
> https://aka.ms/azure-functions/cancellations ?

liliankasem:
@jviau @brettsam - in the last design meeting we discussed setting the execution log message to "Canceled", this change goes into the WebJobs is SDK and roughly looks like this (and potentially more changes depending on the behaviour we want to expose):

- https://github.com/Azure/azure-webjobs-sdk/pull/3022

With this change, the logs look like this:

```LOG
warn: Function.Func[1]
      Executing 'Functions.Func' (Reason='This function was programmatically called via the host APIs.', Id=5050cfe6-9514-43a1-8ffe-caf3da59c9a1)
info: Function.Func[4]  // or we can log this as Warning
      Executed 'Functions.Func' (Canceled, Id=5050cfe6-9514-43a1-8ffe-caf3da59c9a1, Duration=9338ms)
fail: Host.Results[0]
      Microsoft.Azure.WebJobs.Host.FunctionInvocationException: Exception while executing function: Functions.Func
       ---> Microsoft.Azure.WebJobs.Host.FunctionInvocationCanceledException: The invocation request with id '5050cfe6-9514-43a1-8ffe-caf3da59c9a1' was canceled before the request was sent to the worker.
```

I think this is a breaking change potentially for WebJobs and for any services/detectors/portal etc. that rely on this log so I don't think is actually viable (at least not in v4).

Edit: Also I was looking into the ASP.NET Core behaviour for this and my understanding is that the client that disconnects gets a TaskCanceledException but the request still executes - might be worth a sanity check?

cc: @fabiocav fyi

jviau:
> Also I was looking into the ASP.NET Core behaviour for this and my understanding is that the client that disconnects gets a TaskCanceledException but the request still executes - might be worth a sanity check?

This depends entirely on individual apps. ASP.NET Core uses cooperating cancellation, so `HttpContext.RequestAborted` will be in a cancelled state. So, it fails on the first thing that decides to observe and throw that cancellation. This means it may throw in middleware, in MVC, in the customer controller - entirely depends on when it is canceled and what the request pipeline looks like for a given app.

src/WebJobs.Script/Config/ScriptApplicationHostOptions.cs
```diff
@@ -35,5 +35,12 @@ public class ScriptApplicationHostOptions
         public bool IsFileSystemReadOnly { get; set; }
 
         public bool IsScmRunFromPackage { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether the host sends cancelled invocations to the worker.
```
surgupta-msft: nit: typo - `cancelled` v/s `canceled` for consistency in code.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -746,12 +747,22 @@ internal async Task SendInvocationRequest(ScriptInvocationContext context)
...
-                // do not send an invocation request if cancellation has been requested
                 if (context.CancellationToken.IsCancellationRequested)
                 {
-                    _workerChannelLogger.LogWarning("Cancellation has been requested. The invocation request with id '{invocationId}' is canceled and will not be sent to the worker.", invocationId);
-                    context.ResultSource.TrySetCanceled();
-                    return;
+                    _workerChannelLogger.LogDebug("Cancellation was requested prior to the invocation request ('{invocationId}') being sent to the worker.", invocationId);
+
+                    // If the worker does not support handling InvocationCancel grpc messages, or if cancellation is supported and the customer opts-out
+                    // of sending cancelled invocations to the worker, we will cancel the result source and not send the invocation to the worker.
```
surgupta-msft, line 857: In general, what is expected from the worker if we send an invocation that is already requested for cancellation? Are workers supposed to detect this and cancel or something else?

release_notes.md
```diff
@@ -9,3 +9,7 @@
 - Update Node.js Worker Version to [3.8.1](https://github.com/Azure/azure-functions-nodejs-worker/releases/tag/v3.8.1)
 - Update PowerShell 7.4 Worker Version to [4.0.2930](https://github.com/Azure/azure-functions-powershell-worker/releases/tag/v4.0.2930)
 - Update WebJobsScriptHostService to remove hardcoded sleep during application shut down (#9520)
+- Implement host configuration property for handling pre-cancelled invocation requests (#9523)
+  - If a worker supports CancellationTokens, cancelled invocations will now be sent to the worker by default
+    - Customers can opt-out of this behavior by setting `SendCanceledInvocationsToTheWorker` to `false` in host.json
```
surgupta-msft: Can you share sample scenarios where customers would want cancelled invocations to be sent to the worker versus when they would like to opt-out? Feel free to point me to any docs link if this is already documented.

liliankasem: We're working on documentation for this as part of this issue: https://github.com/Azure/azure-functions-host/issues/9530

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
liliankasem: If a worker supports the cancellation (via capability) they should get the InvocationCanceled grpc message right after and handle that at they would any other cancelled invocation

src/WebJobs.Script/Config/ScriptJobHostOptionsSetup.cs
```diff
@@ -51,6 +51,12 @@ public void Configure(ScriptJobHostOptions options)
                 {
                     options.FileLoggingMode = fileLoggingMode.Value;
                 }
+
+                var sendCanceledInvocationsToTheWorker = jobHostSection.GetSection("SendCanceledInvocationsToTheWorker");
+                _applicationHostOptions.Value.SendCanceledInvocationsToTheWorker = sendCanceledInvocationsToTheWorker.Value is null
```
jviau: Cross configuration of different options objects from another configuration is not recommended. This is not deterministic and introduces race conditions. Options are lazily constructed on first import, so this only works if both `ScriptJobHostOptions` and `ScriptApplicationHostOptions` are imported. This is not intuitive and can easily break in the future.

_Reading_ an imported options object during configuration of another is fine. _Writing_ to the imported options object is the issue.

We will need to figure out a deterministic way to get this setting where we need it.

liliankasem: Yup agree, planning to sync with @fabiocav on this soon to see where is the appropriate place for this config to be

liliankasem: I had a sync with Fabio and looks like we can use this the host manager to get the jobhost options and avoid using the webhost options completly:

`_serviceProvider.GetService(typeof(IOptions<ScriptJobHostOptions>));`

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -149,6 +153,8 @@ internal partial class GrpcWorkerChannel : IRpcWorkerChannel, IDisposable
 
         public RpcWorkerConfig WorkerConfig => _workerConfig;
 
+        public IOptions<ScriptJobHostOptions> JobHostOptions => (IOptions<ScriptJobHostOptions>)_serviceProvider.GetService(typeof(IOptions<ScriptJobHostOptions>));
```
jviau: Can we fetch this only once and store the value?

liliankasem: We can cache but need to take in to account specialization where this value will/could change

liliankasem: @jviau I've refactored this to use ActiveHostChanged event for that scenario, thoughts?

https://github.com/Azure/azure-functions-host/pull/9523/commits/263a5b6e920e865235b0cc7a4b643387e0a49dc2

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -1304,6 +1356,7 @@ protected virtual void Dispose(bool disposing)
                     }
 
                     (_rpcWorkerProcess as IDisposable)?.Dispose();
+                    // (_scriptHostManager as IDisposable)?.Dispose();
```
liliankasem: Q: should I also dispose this or nah?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -149,6 +158,37 @@ internal partial class GrpcWorkerChannel : IRpcWorkerChannel, IDisposable
...
+            {
+                return;
+            }
+
+            LoadScriptJobHostOptions();
+        }
+
+        public void LoadScriptJobHostOptions()
+        {
+            if (_scriptHostManager?.State == ScriptHostState.Initialized
```
liliankasem: Found a bug with this logic so working to fix that now

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
fabiocav: No. The lifecycle of the manager is not controlled by the channel. This is passed in as a dependency here and ultimately comes from DI.

src/WebJobs.Script/Config/HostJsonFileConfigurationSource.cs
```diff
@@ -51,7 +51,7 @@ public class HostJsonFileConfigurationProvider : ConfigurationProvider
             {
                 "version", "functionTimeout", "retry", "functions", "http", "watchDirectories", "watchFiles", "queues", "serviceBus",
                 "eventHub", "singleton", "logging", "aggregator", "healthMonitor", "extensionBundle", "managedDependencies",
-                "customHandler", "httpWorker", "extensions", "concurrency"
+                "customHandler", "httpWorker", "extensions", "concurrency", "sendCanceledInvocationsToTheWorker"
```
fabiocav: nit,  naming: `sendCanceledInvocationsToTheWorker` -> `sendCanceledInvocationsToWorker`

src/WebJobs.Script/Config/ScriptJobHostOptions.cs
```diff
@@ -122,5 +122,12 @@ public string RootScriptPath
         /// Gets or sets a value indicating whether the ScriptHost is in standby mode.
         /// </summary>
         public bool IsStandbyConfiguration { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether the host sends cancelled invocations to the worker.
+        /// This defaults to true, meaning if cancellation is signalled we will still send the pre-cancelled
+        /// invocation to the worker.
+        /// </summary>
+        public bool SendCanceledInvocationsToTheWorker { get; set; }
```
fabiocav: Same comment on naming.

src/WebJobs.Script/Config/ScriptJobHostOptionsSetup.cs
```diff
@@ -51,6 +51,14 @@ public void Configure(ScriptJobHostOptions options)
                 {
                     options.FileLoggingMode = fileLoggingMode.Value;
                 }
+
+                // Default SendCanceledInvocationsToTheWorker to true if not provided
+                var sendCanceledInvocationsToTheWorker = jobHostSection.GetSection("SendCanceledInvocationsToTheWorker");
```
fabiocav: Didn't see more info on how you expect this to be structured in `host.json`, but I'd expect this lookup to use `GetValue` instead of treated as a section.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -761,12 +791,22 @@ internal async Task SendInvocationRequest(ScriptInvocationContext context)
...
                 {
-                    _workerChannelLogger.LogWarning("Cancellation has been requested. The invocation request with id '{invocationId}' is canceled and will not be sent to the worker.", invocationId);
-                    context.ResultSource.TrySetCanceled();
-                    return;
+                    _workerChannelLogger.LogDebug("Cancellation was requested prior to the invocation request ('{invocationId}') being sent to the worker.", invocationId);
+
+                    // If the worker does not support handling InvocationCancel grpc messages, or if cancellation is supported and the customer opts-out
+                    // of sending cancelled invocations to the worker, we will cancel the result source and not send the invocation to the worker.
+                    if (!_isHandlesInvocationCancelMessageCapabilityEnabled ||
+                        (_isHandlesInvocationCancelMessageCapabilityEnabled && !JobHostOptions.Value.SendCanceledInvocationsToTheWorker))
```
fabiocav: Simplify the logic here to:

``` csharp
if (!_isHandlesInvocationCancelMessageCapabilityEnabled || !JobHostOptions.Value.SendCanceledInvocationsToTheWorker))
```

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -779,9 +819,9 @@ internal async Task SendInvocationRequest(ScriptInvocationContext context)
                     InvocationRequest = invocationRequest
                 });
 
-                if (_cancelCapabilityEnabled != null && _cancelCapabilityEnabled.Value)
+                if (_isHandlesInvocationCancelMessageCapabilityEnabled)
                 {
-                    context.CancellationToken.Register(() => SendInvocationCancel(invocationRequest.InvocationId));
+                    _cancellationCtr = context.CancellationToken.Register(() => SendInvocationCancel(invocationRequest.InvocationId));
```
fabiocav: This is variable is scoped to this type, but being overridden for every request. Are you trying to keep track of this registration so you can remove it? If so, you may need to do something like storing this is the context and retrieving it when the invocation completes.

src/WebJobs.Script/Config/ScriptJobHostOptionsSetup.cs
liliankasem: Should not be in a section:

```json
{
    "version": "2.0",
    "sendCanceledInvocationsToWorker": false,
}
```

Will switch to GetValue

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
liliankasem: Yeah I think there was feedback somewhere from Jacob about needing to dispose of the registeration - will store this in context similar to how we handle this in dni

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -152,6 +161,26 @@ internal partial class GrpcWorkerChannel : IRpcWorkerChannel, IDisposable
...
+            }
+
+            LoadScriptJobHostOptions(e.NewHost.Services);
+        }
+
+        public void LoadScriptJobHostOptions(IServiceProvider provider)
+        {
+            if (provider?.GetService(typeof(IOptions<ScriptJobHostOptions>)) is IOptions<ScriptJobHostOptions> scriptHostOptions)
+            {
+                _scriptHostOptions = scriptHostOptions;
```
fabiocav, line 180: nit: should never happen (until it does :) ), but should we log here if we're unable to resolve the options?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -834,9 +872,10 @@ internal async Task SendInvocationRequest(ScriptInvocationContext context)
                     InvocationRequest = invocationRequest
                 });
 
-                if (_cancelCapabilityEnabled != null && _cancelCapabilityEnabled.Value)
+                if (_isHandlesInvocationCancelMessageCapabilityEnabled)
                 {
-                    context.CancellationToken.Register(() => SendInvocationCancel(invocationRequest.InvocationId));
+                    var cancellationCtr = context.CancellationToken.Register(() => SendInvocationCancel(invocationRequest.InvocationId));
+                    context.Properties.Add("CancellationTokenRegistration", cancellationCtr);
```
fabiocav: nit: make `CancellationTokenRegistration` a constant

src/WebJobs.Script/Config/ConfigurationPropertyNames.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Script.Configuration
+{
+    public static class ConfigurationPropertyNames
```
fabiocav: Would adding this to the existing `ConfigurationSectionNames` class be appropriate? Can we avoid the new file?

src/WebJobs.Script/Config/ScriptJobHostOptionsSetup.cs
```diff
@@ -45,12 +45,19 @@ public void Configure(ScriptJobHostOptions options)
...
-                    ?.GetValue<FileLoggingMode?>("fileLoggingMode");
+                var fileLoggingMode = jobHostSection.GetSection(ConfigurationSectionNames.Logging)?.GetValue<FileLoggingMode?>("fileLoggingMode");
                 if (fileLoggingMode != null)
                 {
                     options.FileLoggingMode = fileLoggingMode.Value;
                 }
+
+                // Default SendCanceledInvocationsToWorker to true if not provided
+                var sendCanceledInvocationsToWorker = jobHostSection.GetValue<bool?>(ConfigurationPropertyNames.SendCanceledInvocationsToWorker);
+                if (sendCanceledInvocationsToWorker is null)
```
fabiocav: Do we have tests covering this setup? Doesn't seem like we're assigning the value of the resolved property if that is set.

fabiocav (commented):
Mostly nits. 

Would flagged as approved, but I think we still need to review the setup. Once that is addressed, we should be able to get the change in.

src/WebJobs.Script/Config/ScriptJobHostOptionsSetup.cs
liliankasem: Added a test, code is working as expected `jobHostSection.Bind(options)` is populating the `sendCanceledInvocationsToWorker)` property and if it's still null and set the default.

fabiocav: yes, bind will handle this. We should also be able to remove this code altogether if you change the default.

fabiocav (approved):

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -152,6 +161,30 @@ internal partial class GrpcWorkerChannel : IRpcWorkerChannel, IDisposable
...
+        {
+            if (e.NewHost is null)
+            {
+                return;
+            }
+
+            LoadScriptJobHostOptions(e.NewHost.Services);
+        }
+
+        public void LoadScriptJobHostOptions(IServiceProvider provider)
```
kshyju: Does this needs to be `public` ?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -1038,6 +1081,12 @@ internal async Task InvokeResponse(InvocationResponse invokeResponse)
             if (_executingInvocations.TryRemove(invokeResponse.InvocationId, out var invocation))
             {
                 var context = invocation.Context;
+
+                if (context.Properties.TryGetValue(ScriptConstants.CancellationTokenRegistration, out CancellationTokenRegistration ctr))
+                {
+                    ctr.Dispose();
```
kshyju: `CancellationTokenRegistration` implements `IAsyncDisposable` as well. Should we call `DisposeAsync` ?

kshyju: Should we also do `context.Properties.Remove(ScriptConstants.CancellationTokenRegistration)` ?

release_notes.md
```diff
@@ -8,3 +8,7 @@
 - Add function grouping information (https://github.com/Azure/azure-functions-host/pull/9735)
 - Bump `Microsoft.IdentityModel.Tokens`, `Microsoft.IdentityModel.Protocols.OpenIdConnect`, and
   `System.IdentityModel.Tokens.Jwt` from 6.32.0 to 6.35.0 (#9793)
+- Implement host configuration property for handling pre-cancelled invocation requests (#9523)
+  - If a worker supports CancellationTokens, cancelled invocations will now be sent to the worker by default
+    - Customers can opt-out of this behavior by setting `SendCanceledInvocationsToTheWorker` to `false` in host.json
```
kshyju: Do we need to update this page to reflect this new property? https://learn.microsoft.com/en-us/azure/azure-functions/functions-host-json.

Not sure we support a json schema for this file. If so, that also needs to be updated?

liliankasem: I'm planning to make a PR to the schema and I think that auto updates this page

kshyju (approved):

PR #9551, opened 2023-09-19
Update FilePathToVfsUri so it works when filePath is not in "home" path

liliankasem:

`filePath .Substring(home.Length).Trim('\\', '/').Replace("\\", "/");` throws an exception if `filePath` doesn't start with the as `home` path. This can sometimes happen with the TestData is not in the same place as the app e.g.:

a) home (script root) is: `"C:\\Users\\likasem\\apps\\dni-app\\bin\\Debug\\net7.0";`
b) test data path is: `"C:\\Users\\likasem\\AppData\\Local\\Temp\\FunctionsData\\HttpTrigger.dat"`

This fix makes it so that in all scenarios we are able to get the last part of the filePath so that we can extract the file or directory name.

### Pull request checklist

* [x] I have added all required tests (Unit tests, E2E tests) (should be covered by existing tests)

### DIFF
--- a/src/WebJobs.Script.WebHost/Management/VirtualFileSystem.cs
+++ b/src/WebJobs.Script.WebHost/Management/VirtualFileSystem.cs
@@ -204,14 +204,13 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
 
         public static Uri FilePathToVfsUri(string filePath, string baseUrl, ScriptJobHostOptions config, bool isDirectory = false)
         {
-            var home = RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
-                ? ScriptSettingsManager.Instance.GetSetting(EnvironmentSettingNames.AzureWebsiteHomePath) ?? config.RootScriptPath
-                : Path.DirectorySeparatorChar.ToString();
+            int lastSeparatorIndex = filePath.LastIndexOfAny(new char[] { '\\', '/' });
 
-            filePath = filePath
-                .Substring(home.Length)
-                .Trim('\\', '/')
-                .Replace("\\", "/");
+            if (lastSeparatorIndex != -1)
+            {
+                // Gets the last component of 'filePath' which should be the file or directory name.
+                filePath = filePath.Substring(lastSeparatorIndex + 1);
+            }
 
             return new Uri($"{baseUrl}/admin/vfs/{filePath}{(isDirectory ? "/" : string.Empty)}");
         }


src/WebJobs.Script.WebHost/Management/VirtualFileSystem.cs
```diff
@@ -204,14 +204,13 @@ public virtual Task<HttpResponseMessage> DeleteItem(HttpRequest request, bool re
 
         public static Uri FilePathToVfsUri(string filePath, string baseUrl, ScriptJobHostOptions config, bool isDirectory = false)
         {
-            var home = RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
```
mathewc, line 207: Note that this same code lives in VfsUriToFilePath method below. It's still not clear to me why this code isn't working for you. It hasn't changed in a very long time AFAIK, and I believe we have E2E tests that are configuring these directories and not running into issues.

Any changes we make here we have to be sure aren't breaking our ARM APIs.

liliankasem: It's breaking for me when I run the host locally to debug some of the `admin/functions/` endpoints (on Windows, not having this issue on my mac), the TestData path isn't the same as the Function path and leads to issues, I'm not sure why that is as I didn't make any changes that would cause the path to be different (The script root was the /bin/Debug/net7.0 directory and the TestData path was a temp directory)

Also, when I was working on this PR: "https://github.com/Azure/azure-functions-host/pull/9510", this [test](https://github.com/Azure/azure-functions-host/blob/0adf5a6077c7b2e565a66e9f52ab965795b7520d/test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/CodelessEndToEndTests.cs#L168) randomly started breaking for the same reason - it seems to automatically resolve at some point when I was cleaning up the branch (I think there might have been a bad merge or rebase).

Anywho, this led me to feel like this is something that may need to be addressed? Or is it safe to assume that the TestData path is always going to be in the same place as "home" (script root?)?

> Any changes we make here we have to be sure aren't breaking our ARM APIs.

Do you have any guidance for testing this?

mathewc: The logic used to determine the paths is [here](https://github.com/Azure/azure-functions-host/blob/0adf5a6077c7b2e565a66e9f52ab965795b7520d/src/WebJobs.Script.WebHost/Configuration/WebScriptHostConfigurationSource.cs#L46). So generally when I'm testing locally, I'll configure my local environment such that IsAppServiceEnvironment returns true. You can also configure FUNCTIONS_TEST_DATA_PATH if you need to override the defaulting of it to the TEMP dir.

liliankasem: Gotcha, thanks for sharing :) I'll close this for now since there is no production impact

PR #9553, opened 2023-09-20
Limit dotnet-isolated specialization to 64 bit host process

### LINKED ISSUE #9548: Skip specialization using native placeholder if platform setting is not 64 bit
For dotnet-isolated placeholder support, our [documentation recommends](https://learn.microsoft.com/en-us/azure/azure-functions/dotnet-isolated-process-guide#placeholders-preview) customers to use "64 bit" platform setting. But as of today, if they don't make that change, specialization will fail because our native placeholder component would not be able to load 32 bit assemblies.  We should update host code to skip specialization code executing using the native component and fallback to loading the customer assembly directly.

kshyju:
Fixes #9548 

With this change, specialization for dotnet isolated using native placeholder will happen only if the host process is 64 bit.
Our Linux distros are 64 bit and customers cannot change that via portal.

https://learn.microsoft.com/en-us/azure/azure-functions/dotnet-isolated-process-guide#placeholders-preview

### Pull request checklist

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -16,3 +16,4 @@
   - This fixes a bug with worker indexing where we are shutting down worker channels and creating a new channel that never
     gets properly initialized as the invocation buffers are not created - this leads to a "Did not find initialized workers" error.
 - Check if a blob container or table exists before trying to create it (#9555)
+- Limit dotnet-isolated specialization to 64 bit host process (#9548)
--- a/src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
+++ b/src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
@@ -184,7 +184,7 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
                 // We support specialization of dotnet-isolated only on 64bit host process.
                 if (!_environment.Is64BitProcess)
                 {
-                    _logger.LogDebug("This app is configured as 32-bit and therefore does not leverage all performance optimizations. See https://aka.ms/azure-functions/dotnet/placeholders for more information.");
+                    _logger.LogInformation("This app is configured as 32-bit and therefore does not leverage all performance optimizations. See https://aka.ms/azure-functions/dotnet/placeholders for more information.");
                     return false;
                 }
 
--- a/src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
+++ b/src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
@@ -184,7 +184,8 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
                 // We support specialization of dotnet-isolated only on 64bit host process.
                 if (!_environment.Is64BitProcess)
                 {
-                    _logger.LogInformation("This app is configured as 32-bit and therefore does not leverage all performance optimizations. See https://aka.ms/azure-functions/dotnet/placeholders for more information.");
+                    _logger.LogInformation(new EventId(421, "PlaceholderMissDueToBitness"), 
+                        "This app is configured as 32-bit and therefore does not leverage all performance optimizations. See https://aka.ms/azure-functions/dotnet/placeholders for more information.");
                     return false;
                 }
 
--- a/src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
+++ b/src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
@@ -181,6 +181,15 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
                     return false;
                 }
 
+                // We support specialization of dotnet-isolated only on 64bit host process.
+                var is64Bit = _environment.Is64BitProcess;
+                _logger.LogDebug("Is64BitProcess: {is64Bit}", is64Bit.ToString());
+
+                if (!is64Bit)
+                {
+                    return false;
+                }
+
                 // Do not specialize if the placeholder is 6.0 but the site is 7.0 (for example).
                 var currentWorkerRuntimeVersion = _environment.GetEnvironmentVariable(RpcWorkerConstants.FunctionWorkerRuntimeVersionSettingName);
                 channel.WorkerProcess.Process.StartInfo.Environment.TryGetValue(RpcWorkerConstants.FunctionWorkerRuntimeVersionSettingName, out string placeholderWorkerRuntimeVersion);
--- a/src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
+++ b/src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
@@ -184,8 +184,9 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
                 // We support specialization of dotnet-isolated only on 64bit host process.
                 if (!_environment.Is64BitProcess)
                 {
-                    _logger.LogInformation(new EventId(421, "PlaceholderMissDueToBitness"), 
+                    _logger.LogInformation(new EventId(421, ScriptConstants.PlaceholderMissDueToBitnessEventName),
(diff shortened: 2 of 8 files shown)


liliankasem (approved):

src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
```diff
@@ -181,6 +181,15 @@ private bool UsePlaceholderChannel(IRpcWorkerChannel channel)
                     return false;
                 }
 
+                // We support specialization of dotnet-isolated only on 64bit host process.
+                var is64Bit = _environment.Is64BitProcess;
+                _logger.LogDebug("Is64BitProcess: {is64Bit}", is64Bit.ToString());
```
safihamid: let's add a more detailed log here. we want to inform customers in Risk Alerts if they can't use placeholders because their process isn't 64bit based on this log. this log is a bit too generic and might not be specific enough for this purpose.

src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
```diff
@@ -181,6 +181,15 @@ private bool UsePlaceholderChannel(IRpcWorkerChannel channel)
...
+
+                if (!is64Bit)
+                {
```
safihamid, line 186: Or rather add the log here when we skip placeholders based on this logic?

release_notes.md
```diff
@@ -8,4 +8,5 @@
 - Update WebJobsScriptHostService to remove hardcoded sleep during application shut down (#9520)
 - Update PowerShell 7.2 Worker Version to [4.0.2974](https://github.com/Azure/azure-functions-powershell-worker/releases/tag/v4.0.2974)
 - Update PowerShell 7.0 Worker Version to [4.0.2973](https://github.com/Azure/azure-functions-powershell-worker/releases/tag/v4.0.2973)
-- Add support for standalone executable (ie: `dotnet build --standalone`) for out-of-proc workers in Linux Consumption.
+- Limit dotnet-isolated specialization to 64 bit host process (#9548).
```
safihamid: nit: Limit dotnet-isolated use of new native placeholders to 64bit host processes only.

Specialization happens regardless if they use native placeholders or base placeholders.

safihamid:
>                 _logger.LogDebug("UsePlaceholderDotNetIsolated: {placeholderEnabled}", placeholderEnabled);

As discussed, while doing this could you also add the process name for placeholder to make sure we are also getting an exact match from Host as well? something like mawsfn......

---
Refers to: src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs:177 in 4b480db. [](commit_id = 4b480db3f4c9c58e5207dc2c65bdcb03bf9fa601, deletion_comment = False)

safihamid:
>                 if (UsePlaceholderChannel(rpcWorkerChannel))

if 'if statement' returns true, maybe better to add  a log here that verifies we are using native placeholder and log the PID and process name like mawsfn.... this is to cross check with what DWAS reports as exact match.

---
Refers to: src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs:126 in 4b480db. [](commit_id = 4b480db3f4c9c58e5207dc2c65bdcb03bf9fa601, deletion_comment = False)

src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
kshyju: Changed the log statement and location.

kshyju:
> > ```
> >             if (UsePlaceholderChannel(rpcWorkerChannel))
> > ```
> 
> if 'if statement' returns true, maybe better to add a log here that verifies we are using native placeholder and log the PID and process name like mawsfn.... this is to cross check with what DWAS reports as exact match.
> 
> Refers to: src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs:126 in [4b480db](https://github.com/Azure/azure-functions-host/commit/4b480db3f4c9c58e5207dc2c65bdcb03bf9fa601). [](commit_id = [4b480db](https://github.com/Azure/azure-functions-host/commit/4b480db3f4c9c58e5207dc2c65bdcb03bf9fa601), deletion_comment = False)

We already log the process id inside the method called inside `if (UsePlaceholderChannel(rpcWorkerChannel))` 
https://github.com/Azure/azure-functions-host/blob/v4.27.1/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs#L554

release_notes.md
kshyju: Updated.

src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
```diff
@@ -181,6 +181,13 @@ private bool UsePlaceholderChannel(IRpcWorkerChannel channel)
                     return false;
                 }
 
+                // We support specialization of dotnet-isolated only on 64bit host process.
+                if (!_environment.Is64BitProcess)
+                {
+                    _logger.LogInformation("This app is configured as 32-bit and therefore does not leverage all performance optimizations. See https://aka.ms/azure-functions/dotnet/placeholders for more information.");
```
safihamid: @finbarr I know you needed to disable/adjust some of the queries in Risk Alerts as they were using large amount of Kusto CPU cycles. Can you confirm if this is acceptable to check in Risk Alerts or we need to add a new field, ... 

We definitely need this log to be added and checked in Risk Alerts.

safihamid:
Can we add process name as well?

---
In reply to: [1735813698](https://github.com/Azure/azure-functions-host/pull/9553#issuecomment-1735813698) [](http://example.com/codeflow?ancestors=1735813698)

kshyju:
> if 'if statement' returns true, maybe better to add a log here that verifies we are using native placeholder and log the PID and process name like mawsfn.... this is to cross check with what DWAS reports as exact match.

Discussed offline. The process name here is "FunctionsNetHost". the data which starts with "mawsfn" is the runtime sitename which we log in the "RuntimeSiteName" for every log entry.

src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
FinVamp1: @safihamid  Hello, can you share an app that has this change deployed and I can test? I think the most performant way might be to check the bitness of the app and if it is dotnet-isolated then we do the check for the log.  If this is in Summary or Details then scanning this column can be expensive.

safihamid: @kshyju do you this log in private stamp already? if not, is there a way to check for bitness first and if it is dotnet-isolated by what we log in FunctionsLog? if that's the case I think we should be good here.

brettsam: Is it better to add an `EventName` to this log message? That would be easy.

brettsam: Can you see if customers get this log message? With it being at the WebHost level, I'm suspecting they won't see it...

Try adding App Insights to an application and testing...

kshyju: Will test in private stamp with latest iteration which includes logging with EventId.

kshyju: Synced offline. Updated log statement with an EventId

kshyju: Synced offline. This won't go to AI. Logging to FunctionsLogs will allow us to setup Risk alerts as suggested by @safihamid

safihamid (approved):
:shipit:

FinVamp1 (approved):
Approved, thanks for adding the Event Name.

src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
kshyju: Verified in private stamp.

image

PR #9555, opened 2023-09-20
Check if a container or table exists before trying to create it

liliankasem:

Resolves multiple GH issues where customers are seeing an excessive amount of "409 The specified container already exists" errors in their logs:

```LOG
warn: Azure.Core[8]
      Error response [d35f480a-744d-4871-ac64-20693baf387a] 409 The specified container already exists. (00.0s)
      Server:Windows-Azure-Blob/1.0 Microsoft-HTTPAPI/2.0
      x-ms-request-id:6ee968fc-c01e-00a4-5f07-ec787b000000
      x-ms-client-request-id:d35f480a-744d-4871-ac64-20693baf387a
      x-ms-version:2021-08-06
      x-ms-error-code:ContainerAlreadyExists
      Date:Wed, 20 Sep 2023 21:15:31 GMT
      Content-Length:230
      Content-Type:application/xml
```

`CreateIfNotExistsAsync` throws an exception if the container exists. An [issue](https://github.com/Azure/azure-sdk-for-net/issues/109) was opened for this in the sdk repo but was closed unresolved. This was an intention change and not a bug ([more details](https://github.com/Azure/azure-storage-net/issues/481)). However, it still does lead to excessive logs.

 The change may only need to happen for the blob container but figured better safe and do it for all usage of `CreateIfNotExistsAsync`.

Relevant issues:

- https://github.com/Azure/Azure-Functions/issues/2166
- https://github.com/Azure/Azure-Functions/issues/2370
- https://stackoverflow.com/questions/71202908/azure-function-works-but-reports-409-the-specified-container-already-exists

### Pull request checklist

### Additional information

Additional PR information


mathewc:
Is the issue you referenced for 409s talking about conflicts coming from storage operations performed by the Host (i.e. code in this repo), or by code in lower levels? I ask because most of the storage operations performed by the Host are done in the WebJobs SDK, and looking through all the CreateIfNotExistsAsync calls (e.g. [here](https://github.com/Azure/azure-webjobs-sdk/blob/b3a8e94461d6fb92aa3416f87dc22f8d1e6627ca/src/Microsoft.Azure.WebJobs.Host.Storage/Singleton/BlobLeaseDistributedLockManager.cs#L289)) they all seem to be using the existing pattern of just relying on the storage SDK method.

So I'm wondering whether the changes made in this PR will actually address the customer issue? I.e. we're using CreateIfNotExistsAsync all over the place, down through transitive dependencies.

src/WebJobs.Script.WebHost/Security/KeyManagement/BlobStorageSecretsRepository.cs
```diff
@@ -66,7 +66,11 @@ protected virtual BlobContainerClient CreateBlobContainerClient(string connectio
             if (_azureBlobStorageProvider.TryCreateBlobServiceClientFromConnection(connection, out BlobServiceClient blobServiceClient))
             {
                 var blobContainerClient = blobServiceClient.GetBlobContainerClient(_secretsContainerName);
-                blobContainerClient.CreateIfNotExists();
+
+                if (!blobContainerClient.Exists())
```
mathewc, line 70: Documentation of CreateIfNotExists says it does an existence check, so adding this additional existence check means we're checking twice, then doing the create operation when necessary?

Also, in https://github.com/Azure/azure-sdk-for-net/issues/109#issuecomment-1288786252 it's mentioned that this Exists check itself is also generates a warning "404 The specified container does not exist.". Is that true?

liliankasem:
> Is the issue you referenced for 409s talking about conflicts coming from storage operations performed by the Host (i.e. code in this repo), or by code in lower levels? I ask because most of the storage operations performed by the Host are done in the WebJobs SDK, and looking through all the CreateIfNotExistsAsync calls (e.g. [here](https://github.com/Azure/azure-webjobs-sdk/blob/b3a8e94461d6fb92aa3416f87dc22f8d1e6627ca/src/Microsoft.Azure.WebJobs.Host.Storage/Singleton/BlobLeaseDistributedLockManager.cs#L289)) they all seem to be using the existing pattern of just relying on the storage SDK method.
> 
> So I'm wondering whether the changes made in this PR will actually address the customer issue? I.e. we're using CreateIfNotExistsAsync all over the place, down through transitive dependencies.

That's a good point, a couple years ago we made a similar change but for the [lease acquisition](https://github.com/Azure/azure-webjobs-sdk/blob/b3a8e94461d6fb92aa3416f87dc22f8d1e6627ca/src/Microsoft.Azure.WebJobs.Host.Storage/Singleton/BlobLeaseDistributedLockManager.cs#L170-L186) for a similar reason, and I think we missed the container creation flow. I think we should change this in the WebJobs SDK too if we can. I did some testing locally and this did reduce some of the 409 noise.

src/WebJobs.Script.WebHost/Security/KeyManagement/BlobStorageSecretsRepository.cs
liliankasem: Yeah I don't love it but I'm not sure how else to address this given that that CreateIfNotExists with throw (catching won't help because App Insights will still log the operation).

As for the 404, I'm not sure I'll need to test it out myself

brettsam: I believe that the `CreateIfNotExists()` method does an attempt to create -- and silently handles the 409 conflict internally. This means you almost always get a 409 from this once the table is created. App Insights can see this and reports it as an error.

Switching to the `Exists()` check -- you'll get a 404 once, and then, generally, you'll never see it again as the table will always exist and you'll get a 200 from then on.

This has been an issue elsewhere and we fix them as we get complaints. I added this here a while back b/c we got the same complaints: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/storage/Microsoft.Azure.WebJobs.Extensions.Storage.Blobs/src/Config/BlobsExtensionConfigProvider.cs#L515-L520

brettsam (approved):

mathewc (approved):

PR #9570, opened 2023-09-26
logging improvements in GrpcWorkerChannel

brettsam:
Splitting out some minor log improvements I was making in another PR so they don't pollute my other one. 
- Using `LoggerMessage.Define` for some of our hot-path logs. These are happening (potentially) every invocation and this will save with allocations.
- No longer log any `RpcLog` messages that we receive from the worker. These were rarely used.
- Log if we "drop" an `RpcLog` message because it is not happening during an invocation. We have a bug around this today (https://github.com/Azure/azure-functions-host/issues/9238) and we weren't able to track it because we had no indication. My next PR will address this bug.
- Remove a StringBuilder we rarely use.

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.Log.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.Log.cs
@@ -1,0 +1,37 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.Logging;
+using static Microsoft.Azure.WebJobs.Script.Grpc.Messages.StreamingMessage;
+
+namespace Microsoft.Azure.WebJobs.Script.Grpc
+{
+    internal partial class GrpcWorkerChannel
+    {
+        // EventId range is 800-899
+        private static class Logger
+        {
+            private static readonly Action<ILogger, string, ContentOneofCase, Exception> _channelReceivedMessage = LoggerMessage.Define<string, ContentOneofCase>(
+                LogLevel.Debug,
+                new EventId(820, nameof(ChannelReceivedMessage)),
+                "[channel] received {workerId}: {msgType}");
+
+            private static readonly Action<ILogger, string, Exception> _invocationResponseReceived = LoggerMessage.Define<string>(
+                LogLevel.Debug,
+                new EventId(821, nameof(InvocationResponseReceived)),
+                "InvocationResponse received for invocation: '{invocationId}'");
+
+            private static readonly Action<ILogger, string, Exception> _ignoringRpcLog = LoggerMessage.Define<string>(
+                LogLevel.Debug,
+                new EventId(822, nameof(IgnoringRpcLog)),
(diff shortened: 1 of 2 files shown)


src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.Log.cs
```diff
@@ -0,0 +1,37 @@
...
+                new EventId(820, nameof(ChannelReceivedMessage)),
+                "[channel] received {workerId}: {msgType}");
+
+            private static readonly Action<ILogger, string, Exception> _invocationResponseReceived = LoggerMessage.Define<string>(
+                LogLevel.Debug,
+                new EventId(821, nameof(InvocationResponseReceived)),
+                "InvocationResponse received for invocation: '{invocationId}'");
+
+            private static readonly Action<ILogger, string, Exception> _ignoringRpcLog = LoggerMessage.Define<string>(
+                LogLevel.Debug,
```
brettsam, line 26: Leaving this as `Debug` for now as it may be noisy. My next PR which starts to address the out-of-order messages should reduce this significantly, but at the moment we have no idea how much this is happening.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -247,9 +247,9 @@ private async Task ProcessInbound()
                 {
                     while (_inbound.TryRead(out var msg))
                     {
-                        if (debug)
+                        if (debug && msg.MessageType != MsgType.RpcLog)
```
brettsam, line 250: Remove `RpcLog` from this... I know @mathewc has requested this in the past. It's not used much and can be very noisy.

liliankasem: YAY!

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -986,25 +986,24 @@ internal async Task InvokeResponse(InvocationResponse invokeResponse)
...
                                 default:
                                     break;
                             }
                         }
 
-                        if (usedSharedMemory)
+                        if (sharedMemoryLogBuilder != null)
                         {
-                            _workerChannelLogger.LogDebug("Shared memory usage for response of invocation '{invocationId}' is {SharedMemoryUsage}", invokeResponse.InvocationId, logBuilder.ToString());
+                            _workerChannelLogger.LogDebug("Shared memory usage for response of invocation '{invocationId}' is {SharedMemoryUsage}", invokeResponse.InvocationId, sharedMemoryLogBuilder.ToString());
```
jviau: nit: do not call `.ToString()`. Let the logger internals call `.ToString()`, this way if the log statement is filtered out we won't pay the price for that `ToString()` call at all.

brettsam: Good call. I didn't add that part with this PR, but I'll clean it up while I'm in here.

liliankasem (approved):

kshyju (approved):

PR #9573, opened 2023-09-27
Setting max supported bundle version via hosting config

### LINKED ISSUE #9537: Control max bundle version by hosting configuration
Now that we have the ability to use hosting configuration, we should look into adding the ability to control the max bundle version used by a runtime version via hosting configuration. This would be really handy in case we have a hotfix or if we are looking to permanently cap the bundle version to specific version.

soninaren:
Settting the max bundle version via hosting config if one is set

resolves #9537

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
+++ b/src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
@@ -271,18 +271,21 @@ namespace Microsoft.Azure.WebJobs.Script.ExtensionBundle
             }
 
             // Check to see if there is a max bundle version set via hosting configuration, if yes then use that instead of the one
-            // available on VM or local machine
-            if (!string.IsNullOrEmpty(configOption.MaximumBundleV3Version))
+            // available on VM or local machine. Only use MaximumBundleV3Version or MaximumBundleV4Version if the version configured
+            // by the customer resolved to version higher than the version set via hosting config.
+            if (!string.IsNullOrEmpty(configOption.MaximumBundleV3Version)
+                && matchingVersion?.Major == ScriptConstants.ExtensionBundleV3MajorVersion)
             {
                 var maximumBundleV3Version = NuGetVersion.Parse(configOption.MaximumBundleV3Version);
-                matchingVersion = matchingVersion?.Major == ScriptConstants.ExtensionBundleV3MajorVersion && matchingVersion > maximumBundleV3Version
-                                ? maximumBundleV3Version
-                                : matchingVersion;
+                matchingVersion = matchingVersion > maximumBundleV3Version ? maximumBundleV3Version : matchingVersion;
+                return matchingVersion?.ToString();
             }
-            if (!string.IsNullOrEmpty(configOption.MaximumBundleV4Version))
+
+            if (!string.IsNullOrEmpty(configOption.MaximumBundleV4Version)
+                && matchingVersion?.Major == ScriptConstants.ExtensionBundleV4MajorVersion)
             {
                 var maximumBundleV4Version = NuGetVersion.Parse(configOption.MaximumBundleV4Version);
-                matchingVersion = matchingVersion?.Major == ScriptConstants.ExtensionBundleV4MajorVersion && matchingVersion > maximumBundleV4Version
+                matchingVersion = matchingVersion > maximumBundleV4Version
                                 ? maximumBundleV4Version
                                 : matchingVersion;
             }
--- a/src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
+++ b/src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
@@ -275,14 +275,14 @@ namespace Microsoft.Azure.WebJobs.Script.ExtensionBundle
             if (!string.IsNullOrEmpty(configOption.MaximumBundleV3Version))
             {
                 var maximumBundleV3Version = NuGetVersion.Parse(configOption.MaximumBundleV3Version);
-                matchingVersion = matchingVersion?.Major == 3 && matchingVersion > maximumBundleV3Version
+                matchingVersion = matchingVersion?.Major == ScriptConstants.ExtensionBundleV3MajorVersion && matchingVersion > maximumBundleV3Version
                                 ? maximumBundleV3Version
                                 : matchingVersion;
             }
             if (!string.IsNullOrEmpty(configOption.MaximumBundleV4Version))
             {
                 var maximumBundleV4Version = NuGetVersion.Parse(configOption.MaximumBundleV4Version);
-                matchingVersion = matchingVersion?.Major == 4 && matchingVersion > maximumBundleV4Version
+                matchingVersion = matchingVersion?.Major == ScriptConstants.ExtensionBundleV4MajorVersion && matchingVersion > maximumBundleV4Version
                                 ? maximumBundleV4Version
                                 : matchingVersion;
             }
--- a/src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
+++ b/src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
@@ -10,6 +10,7 @@ using System.Linq;
 using System.Net.Http;
 using System.Resources;
 using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Script.Config;
 using Microsoft.Azure.WebJobs.Script.Configuration;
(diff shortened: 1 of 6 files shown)


src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -259,7 +263,31 @@ private static string FindBestVersionMatch(VersionRange versionRange, IEnumerabl
...
+            {
+                return matchingVersion?.ToString();
+            }
+
+            // Check to see if there is a max bundle version set via hosting configuration, if yes then use that instead of the one
+            // available on VM or local machine
+            if (!string.IsNullOrEmpty(configOption.MaximumBundleV3Version))
+            {
+                var maximumBundleV3Version = NuGetVersion.Parse(configOption.MaximumBundleV3Version);
+                matchingVersion = matchingVersion?.Major == 3 && matchingVersion > maximumBundleV3Version
```
surgupta-msft: declare constant for 3 and 4

test/WebJobs.Script.Tests/ExtensionBundle/ExtensionBundleManagerTests.cs
```diff
@@ -384,10 +385,27 @@ public async Task GetExtensionBundle_CannotReachZipEndpoint_ReturnsFalseAsync()
...
+        {
+            var range = VersionRange.Parse(versionRange);
+            var resolvedVersion = ExtensionBundleManager.FindBestVersionMatch(range, new List<string>()
+            { "3.7.0", "3.10.0", "3.11.0", "3.15.0", "3.14.0", "2.16.0", "3.13.0", "3.12.0", "3.9.1", "2.12.1", "2.18.0", "3.16.0", "2.19.0", "3.17.0", "4.0.2", "2.20.0", "3.18.0", "4.1.0", "4.2.0", "2.21.0", "3.19.0", "3.19.2", "4.3.0", "3.20.0" },
+            ScriptConstants.DefaultExtensionBundleId, new FunctionsHostingConfigOptions());
+
+            if (string.IsNullOrEmpty(version))
+            {
+                Assert.Equal("4.3.0", resolvedVersion);
+            }
```
surgupta-msft: Should we add an Assert for non-empty version also? And one more inline data for v3 - `[InlineData("[3.*, 4.0.0)", null)]`?

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -259,7 +263,31 @@ private static string FindBestVersionMatch(VersionRange versionRange, IEnumerabl
                 return version;
             }).Where(v => v != null);
 
-            return bundleVersions.OrderByDescending(version => version.Version).FirstOrDefault()?.ToString();
+            var matchingVersion = bundleVersions.OrderByDescending(version => version.Version).FirstOrDefault();
+
+            if (bundleId != ScriptConstants.DefaultExtensionBundleId)
+            {
```
surgupta-msft, line 269: Trying to understand which scenario will execute this `if-block`? Can you share examples of possible values of `bundleId` here?

surgupta-msft:
The [github issue](https://github.com/Azure/azure-functions-host/issues/9537) says that we also want to control the _minimum bundle version_ using Hosting configuration. Is the plan to handle that in a separate PR?

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
soninaren: We only want to match and limit the bundle version for the official bundle version
`Microsoft.Azure.Functions.ExtensionBundle`

Other possible values are.
`Microsoft.Azure.Functions.ExtensionBundle.Preview`
`Microsoft.Azure.Functions.ExtensionBundle.Workflows`

soninaren:
> The [github issue](https://github.com/Azure/azure-functions-host/issues/9537) says that we also want to control the _minimum bundle version_ using Hosting configuration. Is the plan to handle that in a separate PR?

This is the only change we would be making right now, bumping the minimum version would have some serious impact on our CDN. I will update the issue to reflect this.

liliankasem (approved):

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -259,7 +263,31 @@ private static string FindBestVersionMatch(VersionRange versionRange, IEnumerabl
...
+                return matchingVersion?.ToString();
+            }
+
+            // Check to see if there is a max bundle version set via hosting configuration, if yes then use that instead of the one
+            // available on VM or local machine
+            if (!string.IsNullOrEmpty(configOption.MaximumBundleV3Version))
+            {
+                var maximumBundleV3Version = NuGetVersion.Parse(configOption.MaximumBundleV3Version);
+                matchingVersion = matchingVersion?.Major == ScriptConstants.ExtensionBundleV3MajorVersion && matchingVersion > maximumBundleV3Version
+                                ? maximumBundleV3Version
```
kshyju: Just to confirm my understanding. I see we return `maximumBundleV3Version` when `matchingVersion > maximumBundleV3Version`. Did you mean to return that? Or did you wanted to return matchingVersion?

Should we add comments about the logic? (Why we do this)?

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -259,7 +263,31 @@ private static string FindBestVersionMatch(VersionRange versionRange, IEnumerabl
...
+                                : matchingVersion;
+            }
+            if (!string.IsNullOrEmpty(configOption.MaximumBundleV4Version))
```
kshyju: Is it possible that the HostingConfig has value for both V3 and V4? Do we need to execute both blocks here (v3 and v4)? Wondering it makes sense to return at like 280 if we got a new matchingVersion.

test/WebJobs.Script.Tests/ExtensionBundle/ExtensionBundleManagerTests.cs
```diff
@@ -384,10 +385,36 @@ public async Task GetExtensionBundle_CannotReachZipEndpoint_ReturnsFalseAsync()
             Assert.Null(await manager.GetExtensionBundlePath(httpClient));
         }
 
+        [Theory]
+        [InlineData("[3.*, 4.0.0)", "3.19.0")]
+        [InlineData("[4.*, 5.0.0)", "4.2.0")]
+        [InlineData("[4.*, 5.0.0)", null)]
+        [InlineData("[3.*, 4.0.0)", null)]
+        public void LimitMaxVersion(string versionRange, string version)
```
kshyju: In this test, where are we setting the hosting config value for MaxBundle version? Are we are not exercising the new logic you added in the if blocks?

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
soninaren: Yes your understanding is correct. We would use maxBundleVersion in case one is set via hosting configuration. I have some comments added explaining this. However there are a couple more things i can add.

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
soninaren: Yes, We can return on line 280. We do match major versions before use the max configured version. Going to make that change.

test/WebJobs.Script.Tests/ExtensionBundle/ExtensionBundleManagerTests.cs
soninaren: We are exercising the new logic in the if blocks. That logic only applies if the hosting config for v3 and v4 versions are set. In this test case we are checking scenarios, when hosting config is set and when it is not. When hosting config is set the bundle version should be limited to max version mainly it can use a lower version but not higher. When hosting config is not set app can use any version within the version range.

kshyju: Okay, but I am still confused about where we are setting the hostingConfig value for the test? If we do not do that, the `configOption.MaximumBundleV3Version` will return `null` and the code inside if block won't be tested via this UT.

soninaren: Oh man, i missed that completely. That was a good catch. I have fixed the test now.

satvu (approved):

PR #9576, opened 2023-09-27
feat: Add support for W3C trace context propagation in custom handlers using HTTP request forwarding

notheotherben:
### Issue
Azure Functions provides support for arbitrary host runtimes through the "custom handler" model, and in particular it provides a means for one to run arbitrary web services through the `enableForwardingHttpRequest` feature. In general the contract here is that making an HTTP request to the Azure Function will result in an equivalent request being made to your application, and that the corresponding response from your application will be returned to the caller of the Azure Function (i.e. that there's bi-directional pass-through of the request payload).

In general this is the case, with path, (most) headers, and request body being passed through - however one notable exception to this is the W3C `traceparent` and `tracestate` headers, both of which are intercepted by the Azure Functions platform and exposed through the [`TraceContext`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.functions.worker.tracecontext.traceparent?view=azure-dotnet) construct in integrated runtimes. Unfortunately, try as I might, I've yet to find a way to access this in custom handler functions with `enableForwardingHttpRequest` enabled.

The intent of this PR is to inject the requisite W3C Trace Context headers into the request message sent to the custom handler, enabling this information to both be:
 - Correlated with the Azure Functions logs as described [here](https://learn.microsoft.com/en-us/azure/azure-monitor/app/distributed-tracing-telemetry-correlation#correlation-headers-using-w3c-tracecontext).
 - Provided by callers to the Azure Function which intend to propagate trace context across this boundary.

### Pull request checklist


jviau:
@notheotherben I would think we are following client-server propagation pattern here, where instead of copying the trace headers outright, we emit a `Server` (or `Request` for AppInsights users) span in the host, and then we make the call to the downstream service with a `Client` (`Dependency` in AppInsights) span. 

Is this not the experience you see today? Do you not see _any_ `TraceParent` headers when we call your service?

notheotherben:
That may well be the case for the managed runtimes, and may even be the case for custom handlers when using the `enableForwardingHttpRequest: false` option, however when using a custom handler with `enableForwardingHttpRequest: true` I do not receive __any__ W3C trace propagation headers (or any other headers/payload information containing trace context).

In addition to this, even if I explicitly call my service with the `traceparent` header specified, this header is not propagated to my service (i.e. the runtime swallows the header, presumably for the reasons you've stated above). This makes it all but impossible to implement tracing using this hosting model (short of adopting a non-standard header to work around the limitation).

This PR introduces those trace headers based on the invocation context (i.e. it should propagate the correct `Client` span information if that's what's exposed through the invocation context).

**NOTE** I've confirmed that this behaviour differs to the behaviour present in the `func start` command, where these headers are correctly propagated and my service does receive them correctly. I'm expecting this is because `func start` doesn't intercept the trace headers.

If you need a repro to showcase this behaviour, I can certainly put one together.

jviau:
@notheotherben a repro for this would be very helpful. I want to make sure this behavior is correct for all cases. While it looks good from your perspective/expectations, I do worry this could be considered a breaking behavior change to others.

notheotherben:
@jviau here's the repro: https://github.com/notheotherben/azfn-customhandler-otel - you can see this behaviour by hitting [this endpoint](https://azfncustomhandlerotel.azurewebsites.net/api/hello).

While putting it together, I noticed that there's more at play here than just the runtime - it seems that when AppInsights is enabled we _do_ get a `traceparent` header, while when AppInsights is disabled (for example, if you're using your own OpenTelemetry sink) then this header is not set and is explicitly stripped from the request.

With that in mind, I suspect this PR is going to need to change somewhat; my expected behaviour here would be something along the lines of:

- **If AppInsights is enabled**: The incoming `traceparent` header should be used as the parent for the runtime's Span and the corresponding `Client` Span's trace context should be propagated to the application in the `traceparent` header (i.e. the current behaviour).
- **If AppInsights is disabled**: The incoming `traceparent` header should be transparently passed through to the application without manipulation (i.e. it should operate like any other header).

jviau:
@notheotherben I took your sample and debugged through it and the behavior I noticed is as follows (**without** any of your changes):

| AppInsights <sup>1</sup> | `TraceParent` header present on request | `TraceParent` header present in custom handler |
| ---- | ----- | --- |
| disabled | no | no |
| disabled | yes | yes, copied directly |
| enabled | no | yes, new one generated |
| enabled | yes | yes, copied directly <sup>2</sup> |

<sup>1</sup> App insights being enabled/disabled is not the real cause. Any listener to the HTTP diagnostic source will work here, as that controls logic for parent/child client spans on outgoing HTTP requests
<sup>2</sup> This is the only work I see: we should offer configurability here. Default will be to copy directly, but give an option to _not_ copy and let a child client span be generated.

This call:
https://github.com/Azure/azure-functions-host/blob/6ae170c32c97864d33424e1f8cc1783acdbc907a/src/WebJobs.Script/Workers/Http/DefaultHttpWorkerService.cs#L96

Copies all headers, including `traceparent` and `tracestate`.

https://github.com/Azure/azure-functions-host/blob/6ae170c32c97864d33424e1f8cc1783acdbc907a/src/WebJobs.Script/Extensions/HttpRequestExtensions.cs#L155-L158

I think what we need to do is instead of _copying_ the values from `scriptInvocationContext`, we should perform item <sup>2</sup>. By giving an option to not set those headers on the outgoing request, then we will let the HttpDiagnosticHandler do its thing and generate an outgoing `client` span, which is a child of the `scriptInvocationContext` span. This will give you a parent/child causal relationship in telemetry.

This must be an option though (false by default), otherwise it is a behavioral breaking change (as customers may already be relying on the fact we copy the incoming `TraceParent` directly).

EDIT: revamped comment, understand what you need now

notheotherben:
Thank you @jviau - I believe that addressing <sup>2</sup> should close up a gap for us and agree that making this configurable is the way to go (for folks using AppInsights they likely expect the framework spans to be parented under the incoming `traceparent`, with the same server/client pattern being employed internally. The scenarios where AppInsights are disabled seem good to me.

Separately, your findings pushed be to tear apart our client stack and I think I may have found an emission issue there which contributed to a portion of the issue - as for why that was only impacting our Azure Function services and not the on-prem deployments, I'm not yet sure.

Given that this PR is certainly not the way forward, I'm going to close it out to avoid any further confusion here - thank you for the assistance looking into this.

PR #9579, opened 2023-09-28
Changing the logic to identify dotnet isolated apps

soninaren:
The original logic for identifying the dotnet isolated apps was using worker config. However, we were getting a false positive on the check if the customer did not have `FUNCTIONS_WORKER_RUNTIME` set. The new check would take into account function metadata and worker runtime instead of worker config. 

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
+++ b/src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
@@ -81,6 +81,7 @@ namespace Microsoft.Azure.WebJobs.Script.DependencyInjection
             var bundleConfigured = _extensionBundleManager.IsExtensionBundleConfigured();
             bool isLegacyExtensionBundle = _extensionBundleManager.IsLegacyExtensionBundle();
             bool isPrecompiledFunctionApp = false;
+            bool isDotnetIsolatedApp = false;
 
             // dotnet app precompiled -> Do not use bundles
             var workerConfigs = _languageWorkerOptions.CurrentValue.WorkerConfigs;
@@ -101,9 +102,10 @@ namespace Microsoft.Azure.WebJobs.Script.DependencyInjection
                     }
                     isPrecompiledFunctionApp = isPrecompiledFunctionApp || functionMetadata.Language == DotNetScriptTypes.DotNetAssembly;
                 }
+
+                isDotnetIsolatedApp = IsDotnetIsolatedApp(functionMetadataCollection, SystemEnvironment.Instance);
             }
 
-            bool isDotnetIsolatedApp = IsDotnetIsolatedApp(workerConfigs);
             bool isDotnetApp = isPrecompiledFunctionApp || isDotnetIsolatedApp;
             var isLogicApp = SystemEnvironment.Instance.IsLogicApp();
 
@@ -335,14 +337,10 @@ namespace Microsoft.Azure.WebJobs.Script.DependencyInjection
             }
         }
 
-        private bool IsDotnetIsolatedApp(IList<RpcWorkerConfig> workerConfigs)
+        private bool IsDotnetIsolatedApp(IEnumerable<FunctionMetadata> functions, IEnvironment environment)
         {
-            if (workerConfigs != null)
-            {
-                return workerConfigs.Where(config => config.Description is not null && config.Description.Language == RpcWorkerConstants.DotNetIsolatedLanguageWorkerName).Any();
-            }
-
-            return false;
+            string workerRuntime = Utility.GetWorkerRuntime(functions, environment);
+            return workerRuntime.Equals(RpcWorkerConstants.DotNetIsolatedLanguageWorkerName, StringComparison.OrdinalIgnoreCase);
         }
 
         private class TypeNameEqualityComparer : IEqualityComparer<Type>
--- a/src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
+++ b/src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
@@ -340,7 +340,7 @@ namespace Microsoft.Azure.WebJobs.Script.DependencyInjection
         private bool IsDotnetIsolatedApp(IEnumerable<FunctionMetadata> functions, IEnvironment environment)
         {
             string workerRuntime = Utility.GetWorkerRuntime(functions, environment);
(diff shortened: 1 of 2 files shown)


jviau (commented):
> [x] I have added all required tests (Unit tests, E2E tests)

Is existing test coverage sufficient? I would feel a bit better if we add a unit test which fails before this change and succeeds afterwards.

soninaren:
> > [x] I have added all required tests (Unit tests, E2E tests)
> 
> Is existing test coverage sufficient? I would feel a bit better if we add a unit test which fails before this change and succeeds afterwards.

To be honest the existing unit tests does not cover the missing worker runtime scenario. I am working towards adding the unit / e2e test. The test on this one is going to be complicated. I needed eyes on the change before it makes it to hotfix.

jviau:
> > > [x] I have added all required tests (Unit tests, E2E tests)
> > 
> > 
> > Is existing test coverage sufficient? I would feel a bit better if we add a unit test which fails before this change and succeeds afterwards.
> 
> To be honest the existing unit tests does not cover the missing worker runtime scenario. I am working towards adding the unit / e2e test. The test on this one is going to be complicated. I needed eyes on the change before it makes it to hotfix.

Is there a test plan for this then before we include it in the hotfix?

liliankasem:
> > > > [x] I have added all required tests (Unit tests, E2E tests)
> > > 
> > > 
> > > Is existing test coverage sufficient? I would feel a bit better if we add a unit test which fails before this change and succeeds afterwards.
> > 
> > 
> > To be honest the existing unit tests does not cover the missing worker runtime scenario. I am working towards adding the unit / e2e test. The test on this one is going to be complicated. I needed eyes on the change before it makes it to hotfix.
> 
> Is there a test plan for this then before we include it in the hotfix?

Yeah I'm concerned that the hotfix branch was just merged without reviews from our team

soninaren:
> > > > > [x] I have added all required tests (Unit tests, E2E tests)
> > > > 
> > > > 
> > > > Is existing test coverage sufficient? I would feel a bit better if we add a unit test which fails before this change and succeeds afterwards.
> > > 
> > > 
> > > To be honest the existing unit tests does not cover the missing worker runtime scenario. I am working towards adding the unit / e2e test. The test on this one is going to be complicated. I needed eyes on the change before it makes it to hotfix.
> > 
> > 
> > Is there a test plan for this then before we include it in the hotfix?
> 
> Yeah I'm concerned that the hotfix branch was just merged without reviews from our team

Ok, I have added unit tests. We have sufficient coverage here now. 
@liliankasem , @jviau Can you take a look again?

jviau (approved):

liliankasem (approved):

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
```diff
@@ -101,9 +102,10 @@ public async Task<IEnumerable<Type>> GetExtensionsStartupTypesAsync()
                     }
                     isPrecompiledFunctionApp = isPrecompiledFunctionApp || functionMetadata.Language == DotNetScriptTypes.DotNetAssembly;
                 }
+
+                isDotnetIsolatedApp = IsDotnetIsolatedApp(functionMetadataCollection, SystemEnvironment.Instance);
```
brettsam, line 106: Can we use the `IEnvironment` abstraction here? Rather than hardcoding `SystemEnvironment`?

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
```diff
@@ -335,14 +337,10 @@ void CollectError(Type extensionType, Version minimumVersion, ExtensionStartupTy
             }
         }
 
-        private bool IsDotnetIsolatedApp(IList<RpcWorkerConfig> workerConfigs)
+        private bool IsDotnetIsolatedApp(IEnumerable<FunctionMetadata> functions, IEnvironment environment)
         {
-            if (workerConfigs != null)
-            {
-                return workerConfigs.Where(config => config.Description is not null && config.Description.Language == RpcWorkerConstants.DotNetIsolatedLanguageWorkerName).Any();
```
brettsam, line 342: So were we previously checking whether there was any dotnet-isolated worker config? and if so, assuming it was dotnet-isolated? 

It seems like this would have broken once we introduced the placeholder worker, which meant this config was now always present. 

Any idea why this started showing up as an issue now?

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
brettsam: Also, why did this check move inside the `bundleConfigured` check?

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
soninaren: This would have been an issue since it was introduced in 4.26.0. I think we just detected this late.

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
soninaren: I needed the `functionMetadataCollection` which is only available inside the if block . `isDotnetIsolatedApp` is only used if bundle is configured same as `isPrecompiledFunctionApp` so moving it inside would not break the logic.

soninaren: I don't think IEnvironment abstraction is available to `ScriptStartupTypeLocator`

brettsam: This reads a little confusingly b/c if `!bundlesConfigured`... then `isDotnetIsolatedApp` will always be `false` since we never evaluate it, which isn't really the case.

I think it's the variable naming that's throwing me off. 

Can you change that name to something like `workerSupportsBundles`? That may make things read better... and it'd basically evaluate with this...

` workerSupportsBundles = !IsDotnetIsolatedApp(functionMetadataCollection, SystemEnvironment.Instance);`

soninaren: sure

soninaren: Discussed this with @brettsam offline. Going to merge this PR and follow with the above suggested changes in another PR.

PR #9584, opened 2023-10-02
Rebase on dev-20231002 + removed failing build steps such as tests & SBAM production.

kshyju:


build/common.props
```diff
@@ -4,7 +4,7 @@
     <TargetFramework>netstandard2.1</TargetFramework>
     <LangVersion>latest</LangVersion>
     <MajorVersion>4</MajorVersion>
-    <MinorVersion>28</MinorVersion>
+    <MinorVersion>99</MinorVersion>
```
kshyju, line 7: Bumped version so that we can can be certain we are using the right build when checking the logs.

fabiocav (approved):

Issue #9588, opened 2023-10-05
Remove requirement for WEBSITE_USE_PLACEHOLDER_DOTNETISOLATED flag


mattchenderson:
As part of GA for the isolated worker model placeholders, the opt-in flag should no longer be required. The flag should be honored for the explicit disablement state, though.

|Setting value | Current behavior | Expected behavior at GA/default |
|-|-|-|
|(not provided) | ❌ Start without isolated worker model placeholder | ✅ Attempt to resolve isolated worker model placeholder |
| `0` | ❌ Start without isolated worker model placeholder | ❌ Start without isolated worker model placeholder|
|`1` | ✅ Attempt to resolve isolated worker model placeholder | ✅ Attempt to resolve isolated worker model placeholder |

All other requirements for resolution still apply as before.

mattchenderson:
This becomes an update to the host to consume configuration injected by the generated worker metadata library. There is a corresponding change needed for the metadata SDK to provide that value: https://github.com/Azure/azure-functions-dotnet-worker/issues/1951

georgeOsdDev:
Hi @mattchenderson 
Is this issue still open?
What is the actual behavior now? Do I still need to add WEBSITE_USE_PLACEHOLDER_DOTNETISOLATED=1 to app settings?
![image](https://github.com/user-attachments/assets/bf52e2e0-bdd1-4c49-840f-cea05827f7e3)

PR #9594, opened 2023-10-07
resolved Issue #9588

FACELESS-GOD:

Here the runtime will not try to attempt to resolve the place holder in an event that the " AzureWebsiteUsePlaceholderDotNetIsolated " flag is null

So Just added a separate condition to check for null

resolves #issue_for_this_pr


kshyju:
@FACELESS-GOD Thank you for your contribution. Upon investigating the impact of the originally planned change, we determined this change is not something we should be doing in the host code for now. The app setting will be enabled in a higher level component. So this change is not needed anymore.

We may reconsider this change in the future, but that requires some other work in the platform components as well. 

We appreciate and encourage external contributions. I strongly suggest you discuss the relevant implementation details with the team via the GH issue before raising a PR. Thank you!

PR #9597, opened 2023-10-10
Add diagnostic events for scenarios that prevent a function app from starting

Francisco-Gamino:

This PR resolve the following issues:

1) https://github.com/Azure/azure-functions-host/issues/9214

2) https://github.com/Azure/azure-functions-host/issues/9430. In this scenario, the Functions Host generates a diagnostic event when the app is unable to start due to the presence of 10 non-decryptable secrets backups. The event documentation will provide guidance to help the user delete the backups, as well as instructions on how to prevent the function app from reaching this state. The screenshot below shows the diagnostic event in the Portal. 

image

3) https://github.com/Azure/azure-functions-host/issues/9431. In this scenario, the Functions Host generates a diagnostic event when attempting to read secrets from blob storage. If the operation fails with a `RequestFailedException`, `Status` `409`, and `Error code` `BlobArchived`, the event is triggered.

image

4) https://github.com/Azure/azure-functions-host/issues/9595. In this scenario, the Functions Host generates a diagnostic event when the function app is unable to start due to parsing errors in the host configuration file. The screenshot below shows the diagnostic event in the Portal. 

image

/cc @eamonoreilly @mattchenderson @fabiocav 

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/Management/LinuxSpecialization/RunFromPackageHandler.cs
+++ b/src/WebJobs.Script.WebHost/Management/LinuxSpecialization/RunFromPackageHandler.cs
@@ -70,6 +70,10 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management.LinuxSpecialization
             }
             catch (Exception e)
             {
+                var message = "Failed to initialize Functions Host due to a problem with the application package.";
+                DiagnosticEventLoggerExtensions.LogDiagnosticEventError(
+                    _logger, DiagnosticEventConstants.RunFromPackageFailedErrorCode, message, DiagnosticEventConstants.RunFromPackageFailedErrorCodeHelpLink, e);
+
                 _logger.LogDebug(e, nameof(ApplyRunFromPackageContext));
                 if (throwOnFailure)
                 {
--- a/src/WebJobs.Script.WebHost/Management/LinuxSpecialization/RunFromPackageHandler.cs
+++ b/src/WebJobs.Script.WebHost/Management/LinuxSpecialization/RunFromPackageHandler.cs
@@ -70,10 +70,6 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management.LinuxSpecialization
             }
             catch (Exception e)
             {
-                var message = "Failed to initialize Functions Host due to a problem with the application package.";
-                DiagnosticEventLoggerExtensions.LogDiagnosticEventError(
-                    _logger, DiagnosticEventConstants.RunFromPackageFailedErrorCode, message, DiagnosticEventConstants.RunFromPackageFailedErrorCodeHelpLink, e);
-
                 _logger.LogDebug(e, nameof(ApplyRunFromPackageContext));
                 if (throwOnFailure)
                 {
--- a/src/WebJobs.Script.WebHost/Security/KeyManagement/BlobStorageSecretsRepository.cs
+++ b/src/WebJobs.Script.WebHost/Security/KeyManagement/BlobStorageSecretsRepository.cs
@@ -9,7 +9,9 @@ using Azure;
 using Azure.Storage.Blobs;
 using Azure.Storage.Blobs.Models;
 using Microsoft.Azure.WebJobs.Host.Storage;
+using Microsoft.Azure.WebJobs.Script.Diagnostics;
 using Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics.Extensions;
+using Microsoft.Azure.WebJobs.Script.WebHost.Properties;
 using Microsoft.Extensions.Logging;
 
 namespace Microsoft.Azure.WebJobs.Script.WebHost
@@ -19,6 +21,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
     /// </summary>
     public class BlobStorageSecretsRepository : BaseSecretsRepository
     {
+        private const string BlobArchivedName = "BlobArchived";
         private readonly string _secretsBlobPath;
         private readonly string _hostSecretsBlobPath;
         private readonly string _secretsContainerName = "azure-webjobs-secrets";
@@ -96,7 +99,21 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
             }
             catch (Exception ex)
             {
-                LogErrorMessage("read", ex);
+                // If the read operation failed because the blob access tier is set to archived, log a diagnostic event.
+                var rfex = ex as RequestFailedException;
+                var operation = "read";
+
+                if (rfex != null && rfex.Status == 409 && rfex.ErrorCode.Equals(BlobArchivedName, StringComparison.OrdinalIgnoreCase))
+                {
+                    DiagnosticEventLoggerExtensions.LogDiagnosticEventError(
+                        Logger,
+                        DiagnosticEventConstants.FailedToReadBlobStorageRepositoryErrorCode,
+                        Resources.FailedToReadBlobSecretRepositoryTierSetToArchive,
+                        DiagnosticEventConstants.FailedToReadBlobStorageRepositoryHelpLink,
+                        rfex);
+                }
+
+                LogErrorMessage(operation, ex);
                 throw;
             }
 
--- a/src/WebJobs.Script.WebHost/Security/KeyManagement/BlobStorageSecretsRepository.cs
+++ b/src/WebJobs.Script.WebHost/Security/KeyManagement/BlobStorageSecretsRepository.cs
@@ -85,7 +85,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
         {
             string secretsContent = null;
             string blobPath = GetSecretsBlobPath(type, functionName);
-            var operation = "read";
+            const string Operation = "read";
             try
             {
                 BlobClient secretBlobClient = Container.GetBlobClient(blobPath);
@@ -110,12 +110,12 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
                         rfex);
                 }
 
-                LogErrorMessage(operation, rfex);
+                LogErrorMessage(Operation, rfex);
                 throw;
             }
             catch (Exception ex)
             {
-                LogErrorMessage(operation, ex);
+                LogErrorMessage(Operation, ex);
                 throw;
             }
 
--- a/src/WebJobs.Script.WebHost/Security/KeyManagement/BlobStorageSecretsRepository.cs
+++ b/src/WebJobs.Script.WebHost/Security/KeyManagement/BlobStorageSecretsRepository.cs
@@ -21,7 +21,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
     /// </summary>
     public class BlobStorageSecretsRepository : BaseSecretsRepository
     {
-        private const string BlobArchivedName = "BlobArchived";
+        private readonly string _blobArchivedName = "BlobArchived";
         private readonly string _secretsBlobPath;
         private readonly string _hostSecretsBlobPath;
         private readonly string _secretsContainerName = "azure-webjobs-secrets";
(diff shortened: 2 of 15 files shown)


src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
```diff
@@ -590,7 +590,12 @@
                 {
                     string message = string.Format(Resources.ErrorTooManySecretBackups, ScriptConstants.MaximumSecretBackupCount, string.IsNullOrEmpty(keyScope) ? "host" : keyScope, await AnalyzeSnapshots(secretBackups));
                     _logger?.LogDebug(message);
-                    throw new InvalidOperationException(message);
+
+                    var exception = new InvalidOperationException(message);
+
+                    DiagnosticEventLoggerExtensions.LogDiagnosticEventError(_logger, DiagnosticEventConstants.MaximumSecretBackupCountErrorCode, message, DiagnosticEventConstants.MaximumSecretBackupCountHelpLink, exception);
```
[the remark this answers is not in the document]
Francisco-Gamino: The `message` string is populated from this resource string: https://github.com/Azure/azure-functions-host/blob/266f0fb78575bfe3b8182c21cf20fd01a64f794b/src/WebJobs.Script.WebHost/Properties/Resources.resx#L123-L124

src/WebJobs.Script.WebHost/Management/LinuxSpecialization/RunFromPackageHandler.cs
```diff
@@ -70,6 +70,10 @@ public async Task<bool> ApplyRunFromPackageContext(RunFromPackageContext pkgCont
             }
             catch (Exception e)
             {
+                var message = "Failed to initialize Functions Host due to a problem with the application package.";
```
balag0: 1) nit: Since this can fail due to other reasons unrelated to the package directly (Ex: ManagedIdentity issues) 

Failed to initialize Functions Host due to a problem **downloading** the application package.

2) Also not sure what's the best way to convey "failed to initialize". If the package download fails, then the host will still start up successfully but functions won't run.

If this is the message we are using for such semi broken state, then this wording is fine too.

3)If customers configure 2 types of deployments (ex: RunFromPackage and AzureFiles) we prefer RFP but fallback to AzureFiles. In this case if RFP fails but AF works, then this diagnostic message could be confusing since the app will still work fine.
I dont think this is a common scenario, so current implementation seems okay for now.

fabiocav: @Francisco-Gamino were there specific exceptions that would enable you to more accurately identify what happened (e.g., what is the exception you get if attempting to unzip an invalid file? What download exceptions can you get here?)

If possible, let's catch those more specific exceptions.

About @balag0 's comments above: 

> 3)If customers configure 2 types of deployments (ex: RunFromPackage and AzureFiles) we prefer RFP but fallback to AzureFiles. In this case if RFP fails but AF works, then this diagnostic message could be confusing since the app will still work fine.
I dont think this is a common scenario, so current implementation seems okay for now.

Even if not a common scenario, this would lead to a lot of confusion, both, for the customer and anyone investigating the issue. If the host is responsible for the flow, we should identify the condition to avoid logging a false positive.

src/WebJobs.Script.WebHost/Security/KeyManagement/BlobStorageSecretsRepository.cs
```diff
@@ -96,7 +99,21 @@ public override async Task<ScriptSecrets> ReadAsync(ScriptSecretsType type, stri
             }
             catch (Exception ex)
             {
-                LogErrorMessage("read", ex);
+                // If the read operation failed because the blob access tier is set to archived, log a diagnostic event.
+                var rfex = ex as RequestFailedException;
```
fabiocav: Avoid this cast by catching the exception directly.

src/WebJobs.Script/Config/HostJsonFileConfigurationSource.cs
```diff
@@ -205,7 +205,12 @@ internal JObject LoadHostConfig(string configFilePath)
                     }
                     catch (JsonException ex)
                     {
-                        throw new FormatException($"Unable to parse host configuration file '{configFilePath}'.", ex);
+                        var message = $"Unable to parse host configuration file '{configFilePath}'.";
+
+                        DiagnosticEventLoggerExtensions.LogDiagnosticEventError(
```
fabiocav: This will only log the inner exception (`ex`). For consistency, it would be good to create the `FormatException`, with the inner exception set as it was before. Log, then throw.

src/WebJobs.Script.WebHost/Security/KeyManagement/BlobStorageSecretsRepository.cs
Francisco-Gamino: Fixed.

src/WebJobs.Script/Config/HostJsonFileConfigurationSource.cs
Francisco-Gamino: Good point. Fixed.

src/WebJobs.Script.WebHost/Management/LinuxSpecialization/RunFromPackageHandler.cs
Francisco-Gamino: > @Francisco-Gamino were there specific exceptions that would enable you to more accurately identify what happened (e.g., what is the exception you get if attempting to unzip an invalid file? What download exceptions can you get here?)
> 
> If possible, let's catch those more specific exceptions.
> 
> About @balag0 's comments above:
> 
> > 3)If customers configure 2 types of deployments (ex: RunFromPackage and AzureFiles) we prefer RFP but fallback to AzureFiles. In this case if RFP fails but AF works, then this diagnostic message could be confusing since the app will still work fine.
> > I dont think this is a common scenario, so current implementation seems okay for now.
> 
> Even if not a common scenario, this would lead to a lot of confusion, both, for the customer and anyone investigating the issue. If the host is responsible for the flow, we should identify the condition to avoid logging a false positive.

Thank you @balag0 and @fabiocav for the PR review. I will rework the logic for diagnostic event https://github.com/Azure/azure-functions-host/issues/9401. For now, I have removed this event from this PR, so we can get the other 3 diagnostic events merged. Please let me know if you have any comments or concerns about the rest of the events. Thank you.

src/WebJobs.Script.WebHost/Security/KeyManagement/BlobStorageSecretsRepository.cs
```diff
@@ -94,9 +98,25 @@ public override async Task<ScriptSecrets> ReadAsync(ScriptSecretsType type, stri
                     }
                 }
             }
+            catch (RequestFailedException rfex) when (rfex.Status == 409)
+            {
+                // If the read operation failed because the blob access tier is set to archived, log a diagnostic event.
+                if (rfex.ErrorCode.Equals(BlobArchivedName, StringComparison.OrdinalIgnoreCase))
+                {
+                    DiagnosticEventLoggerExtensions.LogDiagnosticEventError(
```
soninaren: you can do `logger.LogDiagnosticEventError` here instead.

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
```diff
@@ -590,7 +590,12 @@ private IList<Key> RegenerateKeys(IList<Key> list, ulong seed)
...
```
soninaren: same here

soninaren (commented):
Minor comments, looks good otherwise.

src/WebJobs.Script.WebHost/Security/KeyManagement/BlobStorageSecretsRepository.cs
```diff
@@ -19,6 +21,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
     /// </summary>
     public class BlobStorageSecretsRepository : BaseSecretsRepository
     {
+        private const string BlobArchivedName = "BlobArchived";
```
eliaslopezgt:
```suggestion
        private readonly string _blobArchivedName = "BlobArchived";
```
Consider following the conventions in the file.

src/WebJobs.Script.WebHost/Security/KeyManagement/BlobStorageSecretsRepository.cs
eliaslopezgt: 👍 consider adding a `private readonly ILogger` in the constructor to support this and use _logger instead.

Francisco-Gamino: Fixed. I will use the existing `ILogger` from the base class. 

https://github.com/Azure/azure-functions-host/blob/6ae170c32c97864d33424e1f8cc1783acdbc907a/src/WebJobs.Script.WebHost/Security/KeyManagement/BaseSecretsRepository.cs#L30

which is defined here:
https://github.com/Azure/azure-functions-host/blob/6ae170c32c97864d33424e1f8cc1783acdbc907a/src/WebJobs.Script.WebHost/Security/KeyManagement/BaseSecretsRepository.cs#L43

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
Francisco-Gamino: Fixed.

src/WebJobs.Script.WebHost/Security/KeyManagement/BlobStorageSecretsRepository.cs
Francisco-Gamino: Fixed.

soninaren (approved):

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
```diff
@@ -590,7 +590,11 @@
                 {
                     string message = string.Format(Resources.ErrorTooManySecretBackups, ScriptConstants.MaximumSecretBackupCount, string.IsNullOrEmpty(keyScope) ? "host" : keyScope, await AnalyzeSnapshots(secretBackups));
                     _logger?.LogDebug(message);
-                    throw new InvalidOperationException(message);
+
+                    var exception = new InvalidOperationException(message);
+                    _logger?.LogDiagnosticEventError(DiagnosticEventConstants.MaximumSecretBackupCountErrorCode, message, DiagnosticEventConstants.MaximumSecretBackupCountHelpLink, exception);
```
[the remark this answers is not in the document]
Francisco-Gamino: We use `Resources.ErrorTooManySecretBackups` string resource to create the exception, so this value is not provided by the user.

src/WebJobs.Script.WebHost/Security/KeyManagement/BlobStorageSecretsRepository.cs
```diff
@@ -82,6 +85,7 @@ public override async Task<ScriptSecrets> ReadAsync(ScriptSecretsType type, stri
         {
             string secretsContent = null;
             string blobPath = GetSecretsBlobPath(type, functionName);
+            var operation = "read";
```
aishwaryabh: should this be a const?

aishwaryabh (approved):
LGTM! Just left a small comment :)

src/WebJobs.Script.WebHost/Security/KeyManagement/BlobStorageSecretsRepository.cs
Francisco-Gamino: Good catch. Fixed.

PR #9602, opened 2023-10-10
Handling env reload response from native placeholder for failure case.

### LINKED ISSUE #9606: Prevent placeholder attempts with older versions of the SDK

kshyju:
If customer payload for dotnet isolated app is using an older version of worker package (which does not have the support to handle environment reload request via native placeholder, currently the request will crash. [We recently updated the NativePlaceholder to handle this case](https://github.com/Azure/azure-functions-dotnet-worker/pull/1955) and return a response for the env reload request. That version of Native placeholder is included in this PR. We also updated the code to handle the new response. If the failure is caused  due to customer payload having an older version of package, then we will execute our fallback logic, which is to shutdown the placeholder channel and start a new channel where we will load the customer assembly directly using dotnet.exe.

resolves #9606 

### Pull request checklist

### Additional information

Additional PR information


test/DotNetIsolated60/DotNetIsolated60.csproj
```diff
@@ -6,13 +6,15 @@
...
+    <FunctionsEnableExecutorSourceGen>True</FunctionsEnableExecutorSourceGen>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.Functions.Worker" Version="1.18.0" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker" Version="1.19.0" />
     <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Http" Version="3.0.13" />
     <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore" Version="1.0.0-preview2" />
     <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Storage" Version="6.0.0" />
-    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.12.0" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.15.1" />
```
kshyju, line 17: Updated the existing test project to use the new worker.sdk project so that the "Placeholder hit" tests passes.

test/DotNetIsolated60/Program.cs
```diff
@@ -11,14 +11,12 @@
 if (useProxy)
 {
     hostBuilder
-        .ConfigureFunctionsWebApplication()
-        .ConfigureGeneratedFunctionMetadataProvider();
```
kshyju, line 15: Removed this explicit extension method call in favor of the msbuild property (which does the auto registration of this methods).

test/DotNetIsolated60/DotNetIsolated60.csproj
```diff
@@ -6,13 +6,15 @@
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
     <FunctionsEnableWorkerIndexing>True</FunctionsEnableWorkerIndexing>
+    <FunctionsAutoRegisterGeneratedMetadataProvider>True</FunctionsAutoRegisterGeneratedMetadataProvider>
+    <FunctionsEnableExecutorSourceGen>True</FunctionsEnableExecutorSourceGen>
```
kshyju, line 10: Added these msbuild properties for auto registration and removed the explicit extension method calls from Program.cs

test/DotNetIsolatedUnsupportedWorker/DotNetIsolatedUnsupportedWorker.csproj
```diff
@@ -0,0 +1,29 @@
...
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <FunctionsEnableWorkerIndexing>True</FunctionsEnableWorkerIndexing>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Azure.Functions.Worker" Version="1.19.0" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Http" Version="3.0.13" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore" Version="1.0.0" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Storage" Version="6.0.0" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.14.0" />
```
kshyju, line 15: This is a new test project I added which references an older version of Worker.Sdk package which does not support placeholder. The `DotNetIsolated_PlaceholderMiss_UnsupportedWorkerPackage` test uses the publish output of this project as the function app payload.

test/EmptyScriptRoot/host.json
```diff
@@ -0,0 +1,7 @@
+{
```
kshyju, line 1: Could not check in an empty folder. Our host code automatically generates this file if not exist. So I decided to check in this so that the empty folder structure exist and can be used for `DotNetIsolated_PlaceholderMiss_EmptyScriptRoot` test.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -376,13 +376,55 @@ internal void FunctionEnvironmentReloadResponse(FunctionEnvironmentReloadRespons
...
+            // Currently we support non fatal environment reload errors for dotnet isolated worker only.
+            if (workerRuntime != RpcWorkerConstants.DotNetIsolatedLanguageWorkerName)
+            {
+                return true;
+            }
+
+            var rpcException = response.Result.Exception;
+
+            // Currently the below 2 errors are considered as non fatal.
+            if (rpcException.Message.StartsWith("Microsoft.Azure.Functions.Worker.EnvironmentReloadNotSupportedException"))
```
kshyju: The native host will include this specific exception if the app payload uses an older version of worker package which does not support placeholders. https://github.com/Azure/azure-functions-dotnet-worker/blob/a70ead9e497c990d6c29b920f7ed1c3044f6c2f7/host/src/FunctionsNetHost/Grpc/IncomingGrpcMessageHandler.cs#L64-L67

fabiocav: We should have this exception message be something that isn't specific to .NET. 

Ideally, this would be a flow any worker (any stack) can use.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -376,13 +376,55 @@ internal void FunctionEnvironmentReloadResponse(FunctionEnvironmentReloadRespons
...
+            {
+                _workerChannelLogger.LogDebug(new EventId(422, ScriptConstants.PlaceholderMissDueToBitnessEventName),
```
fabiocav: We shouldn't log this here for the same reason. 

If we want to log details, we should pass that information from the worker and have the host log what came from it.

fabiocav (approved):

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
kshyju: Working on validating an updated version...

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -376,13 +376,55 @@ internal void FunctionEnvironmentReloadResponse(FunctionEnvironmentReloadRespons
 
             if (res.Result.IsFailure(out Exception reloadEnvironmentVariablesException))
             {
-                _workerChannelLogger.LogError(reloadEnvironmentVariablesException, "Failed to reload environment variables");
-                _reloadTask.SetException(reloadEnvironmentVariablesException);
+                if (IsEnvironmentReloadFailureFatalError(res))
```
brettsam: What are the cases where it's not fatal? I assumed that if the worker says it failed, we assume it's fatal. 

No one else even uses this result (b/c it never worked), so we can define the scenario...

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -566,7 +581,9 @@ public Task SendFunctionEnvironmentReloadRequest()
                 FunctionEnvironmentReloadRequest = request
             });
 
-            return _reloadTask.Task;
+            var environmentReloadRequestResult = _reloadTask.Task;
+
+            return environmentReloadRequestResult;
```
brettsam: is this change needed?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -374,12 +374,18 @@ internal void FunctionEnvironmentReloadResponse(FunctionEnvironmentReloadRespons
 
             ApplyCapabilities(res.Capabilities, res.CapabilitiesUpdateStrategy.ToGrpcCapabilitiesUpdateStrategy());
 
-            if (res.Result.IsFailure(out Exception reloadEnvironmentVariablesException))
+            if (res.Result.IsFailure(IsUserCodeExceptionCapabilityEnabled(), out var reloadEnvironmentVariablesException))
             {
-                _workerChannelLogger.LogError(reloadEnvironmentVariablesException, "Failed to reload environment variables");
-                _reloadTask.SetException(reloadEnvironmentVariablesException);
+                if (res.Result.Exception is not null && reloadEnvironmentVariablesException is not null)
```
kshyju, line 379: Checking `res.Result.Exception` for null here because I found when `res.Result.Exception` is null, we still create an instance of the RpcException with empty property values. Did not want to update that in this PR. Perhaps we can fix that later.

null-exception

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
kshyju: Pushed an update which logs the exception sent from the worker as it is. We set the "IsUserCodeException" to true so that message goes to AI logs.

brettsam (approved):

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -414,6 +420,15 @@ internal void WorkerInitResponse(GrpcEvent initEvent)
             _workerInitTask.TrySetResult(true);
         }
 
+        private bool IsUserCodeExceptionCapabilityEnabled()
+        {
+            var enableUserCodeExceptionCapability = string.Equals(
```
kshyju, line 425: The native placeholder worker will send this capability.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
kshyju: not needed. Reverted this in the latest push,.

fabiocav (approved):

soninaren (approved):

safihamid:
>         private bool UsePlaceholderChannel(IRpcWorkerChannel channel)

@kshyju remind me again, do we need to shutdown the channel immediately or can wait like 5 sec? trying to see if this will have negative cold start impact on existing apps once we switch to opt out only

---
Refers to: src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs:160 in 42e4df3. [](commit_id = 42e4df33fd85f582d3375000a0466fdc2bfadade, deletion_comment = False)

kshyju:
> > ```
> >     private bool UsePlaceholderChannel(IRpcWorkerChannel channel)
> > ```
> 
> @kshyju remind me again, do we need to shutdown the channel immediately or can wait like 5 sec? trying to see if this will have negative cold start impact on existing apps once we switch to opt out only
> 
> Refers to: src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs:160 in [42e4df3](https://github.com/Azure/azure-functions-host/commit/42e4df33fd85f582d3375000a0466fdc2bfadade). [](commit_id = [42e4df3](https://github.com/Azure/azure-functions-host/commit/42e4df33fd85f582d3375000a0466fdc2bfadade), deletion_comment = False)

Opened https://github.com/Azure/azure-functions-host/issues/9615 to investigate that.

PR #9603, opened 2023-10-10
Enabling unit test, Updating IsDotnetIsolatedApp to be more accurately captured

soninaren:

Making a change as a response to a comment (https://github.com/Azure/azure-functions-host/pull/9579#discussion_r1341677850) on a pervious PR. This would more accurately capture whether an app is dotnet isolated or not. Currently the log statement would sometimes say that isDotnetIsolated is false, even though in some cases it may be true. This does not affect the bundle selection logic, but the log statement would be in accurate under certain configurations.

The PR updates it so that we make use of all available information available at the time.

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
+++ b/src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
@@ -81,16 +81,16 @@ namespace Microsoft.Azure.WebJobs.Script.DependencyInjection
             var bundleConfigured = _extensionBundleManager.IsExtensionBundleConfigured();
             bool isLegacyExtensionBundle = _extensionBundleManager.IsLegacyExtensionBundle();
             bool isPrecompiledFunctionApp = false;
-            bool isDotnetIsolatedApp = false;
 
             // dotnet app precompiled -> Do not use bundles
             var workerConfigs = _languageWorkerOptions.CurrentValue.WorkerConfigs;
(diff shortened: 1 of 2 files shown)


src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
```diff
@@ -102,10 +102,9 @@ public async Task<IEnumerable<Type>> GetExtensionsStartupTypesAsync()
                     }
                     isPrecompiledFunctionApp = isPrecompiledFunctionApp || functionMetadata.Language == DotNetScriptTypes.DotNetAssembly;
                 }
-
-                isDotnetIsolatedApp = IsDotnetIsolatedApp(functionMetadataCollection, SystemEnvironment.Instance);
             }
 
+            bool isDotnetIsolatedApp = IsDotnetIsolatedApp(functionMetadataCollection, SystemEnvironment.Instance);
```
mathewc, line 107: The below test was failing on [this](https://github.com/Azure/azure-functions-host/blob/266f0fb78575bfe3b8182c21cf20fd01a64f794b/test/WebJobs.Script.Tests/ScriptStartupTypeDiscovererTests.cs#L512) line periodically. This was because IsDotnetIsolatedApp wasn't correct sometimes, causing changes to the bundles path resolution and loading?

Are you sure this is only a logging issue? The code below seems to resolve the path differently based on this value. Could apps be using the wrong path?

PR #9634, opened 2023-10-23
Using source generated version of Regex in cold start code path

kshyju:
Using source generated version of Regex (instead of Regex constructor) in `WorkerFunctionDescriptorProvider` constructor which gets executed during specialization.

https://github.com/Azure/azure-functions-host/blob/6760b11e582c7dc2fb3e33ab9b77611238d4f837/src/WebJobs.Script/Description/Workers/WorkerFunctionDescriptorProvider.cs#L39

[Benchmark ](https://github.com/kshyju/Benchmarks/blob/6e5bd90b541274073de58e2effc89b77b0e26e62/src/Benchmarks.ConsoleApp/RegexBenchmarks.cs)results

```

BenchmarkDotNet v0.13.9+228a464e8be6c580ad9408e98f18813f6407fb5a, Windows 11 (10.0.22621.2428/22H2/2022Update/SunValley2)
11th Gen Intel Core i7-1185G7 3.00GHz, 1 CPU, 8 logical and 4 physical cores
.NET SDK 8.0.100-rc.2.23502.2
  [Host]     : .NET 8.0.0 (8.0.23.47906), X64 RyuJIT AVX2
  DefaultJob : .NET 8.0.0 (8.0.23.47906), X64 RyuJIT AVX2


```
| Method           | Mean         | Ratio |
|----------------- |-------------:|------:|
| RegexConstructor | 9,442.424 ns | 1.000 |
| SourceGenRegex   |     5.675 ns | 0.001 |

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script/Host/ProxyFunctionProvider.cs
+++ b/src/WebJobs.Script/Host/ProxyFunctionProvider.cs
@@ -94,7 +94,7 @@ namespace Microsoft.Azure.WebJobs.Script
             var functionErrors = new Dictionary<string, ICollection<string>>();
             Collection<ProxyFunctionMetadata> proxies = ReadProxyMetadata(_scriptOptions.Value.RootScriptPath, _logger, functionErrors);
 
-            ImmutableArray<FunctionMetadata> metadata;
+            ImmutableArray<FunctionMetadata> metadata = ImmutableArray<FunctionMetadata>.Empty;
             if (proxies != null && proxies.Any())
             {
                 metadata = proxies.ToImmutableArray<FunctionMetadata>();
--- a/src/WebJobs.Script/WebJobs.Script.csproj
+++ b/src/WebJobs.Script/WebJobs.Script.csproj
@@ -1,6 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="..\..\build\common.props" />
   <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
     <PackageId>Microsoft.Azure.WebJobs.Script</PackageId>
     <AssemblyName>Microsoft.Azure.WebJobs.Script</AssemblyName>
     <RootNamespace>Microsoft.Azure.WebJobs.Script</RootNamespace>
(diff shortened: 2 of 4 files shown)


src/WebJobs.Script/Host/ProxyFunctionProvider.cs
```diff
@@ -94,7 +94,7 @@ private ImmutableArray<FunctionMetadata> LoadFunctionMetadata()
             var functionErrors = new Dictionary<string, ICollection<string>>();
             Collection<ProxyFunctionMetadata> proxies = ReadProxyMetadata(_scriptOptions.Value.RootScriptPath, _logger, functionErrors);
 
-            ImmutableArray<FunctionMetadata> metadata;
+            ImmutableArray<FunctionMetadata> metadata = ImmutableArray<FunctionMetadata>.Empty;
```
kshyju, line 97: Unrelated to the regex change. after upgrading the project to net8, this line was throwing
 _CS0165 Use of unassigned local variable 'metadata'_

src/WebJobs.Script/WebJobs.Script.csproj
```diff
@@ -1,6 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="..\..\build\common.props" />
   <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
```
mathewc, line 4: In common props, the default target framework is netstandard2.1. The WebHost and other Script projects target net6.0. Should we strive to make the core projects consistent?

fabiocav: The new APIs being used are only available in .NET 7+.

@kshyju I am thinking this needs to be better planned (this is not the only place we'll need to update to use this).

It's fine if want to measure with the changes, but I want to avoid trickling this in.

kshyju: This PR is targeting the feature branch which has net8 TFM.
We could update the code with `#if NET7_0_OR_GREATER` to conditionally use this so that this change can be merged to dev branch. 

@fabiocav there are other places where regex can be replaced with src gen version, but this is the only one I spotted (via cpu profiling using diagnostic tools ) in specialization flow.

kshyju:
Closing as we will add the changes later.

PR #9641, opened 2023-10-25
Fixing null reference exception when AzureWebJobsStorage is not set.

### LINKED ISSUE #9640: Log diagnostic event throws null Reference exception when AzureWebJobsStorage is not set
Static web apps do not have AzureWebJobStorage connection string set for functions. This causes a null reference exception when trying to log diagnostic events.

soninaren:

closes #9640

Static web apps do not have AzureWebJobStorage connection string set for functions. This causes a null reference exception when trying to log diagnostic events. This PR is addressing the null reference exception and fixing the relevant test.

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventNullRepository.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventNullRepository.cs
@@ -1,0 +1,23 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Concurrent;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Azure.Cosmos.Table;
+using Microsoft.Azure.WebJobs.Host.Executors;
+using Microsoft.Azure.WebJobs.Script.WebHost.Helpers;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
+{
+    public class DiagnosticEventNullRepository : IDiagnosticEventRepository
+    {
+        public void WriteDiagnosticEvent(DateTime timestamp, string errorCode, LogLevel level, string message, string helpLink, Exception exception)
+        {
+            return;
+        }
+    }
+}
--- a/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventNullRepository.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventNullRepository.cs
@@ -15,9 +15,6 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 {
     public class DiagnosticEventNullRepository : IDiagnosticEventRepository
     {
-        public void WriteDiagnosticEvent(DateTime timestamp, string errorCode, LogLevel level, string message, string helpLink, Exception exception)
-        {
-            return;
-        }
+        public void WriteDiagnosticEvent(DateTime timestamp, string errorCode, LogLevel level, string message, string helpLink, Exception exception) { }
     }
 }
--- a/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
@@ -53,15 +53,15 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
                 if (!_environment.IsPlaceholderModeEnabled() && _tableClient == null)
                 {
                     string storageConnectionString = _configuration.GetWebJobsConnectionString(ConnectionStringNames.Storage);
-                    if (string.IsNullOrEmpty(storageConnectionString))
-                    {
-                        _logger.LogError("Azure Storage connection string is empty or invalid. Unable to write diagnostic events.");
-                    }
-
-                    if (CloudStorageAccount.TryParse(storageConnectionString, out CloudStorageAccount account))
+                    if (!string.IsNullOrEmpty(storageConnectionString)
+                        && CloudStorageAccount.TryParse(storageConnectionString, out CloudStorageAccount account))
                     {
                         var tableClientConfig = new TableClientConfiguration();
                         _tableClient = new CloudTableClient(account.TableStorageUri, account.Credentials, tableClientConfig);
+                    }
+                    else
+                    {
+                        _logger.LogError("Azure Storage connection string is empty or invalid. Unable to write diagnostic events.");
                     }
                 }
 
--- a/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
@@ -124,10 +124,17 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
                 return;
             }
 
-            table = table ?? GetDiagnosticEventsTable();
-
             try
             {
+                table = table ?? GetDiagnosticEventsTable();
+
+                if (table == null)
+                {
+                    _logger.LogError("Unable to get table reference. Aborting write operation");
+                    StopTimer();
+                    return;
+                }
+
                 bool tableCreated = await TableStorageHelpers.CreateIfNotExistsAsync(table, _tableCreationRetries);
                 if (tableCreated)
                 {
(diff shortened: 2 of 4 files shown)


src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -124,10 +124,16 @@ internal virtual async Task FlushLogs(CloudTable table = null)
                 return;
             }
 
-            table = table ?? GetDiagnosticEventsTable();
-
             try
             {
+                table = table ?? GetDiagnosticEventsTable();
```
soninaren, line 129: `table` would be null when AzureWebJobsStorage is not set.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -124,10 +124,16 @@ internal virtual async Task FlushLogs(CloudTable table = null)
...
+
+                if (table == null)
+                {
+                    _logger.LogError("Unable to get table reference. Aborting write operation");
+                    return;
+                }
+
                 bool tableCreated = await TableStorageHelpers.CreateIfNotExistsAsync(table, _tableCreationRetries);
```
soninaren, line 138: Null value of `table` would cause a NullRef exception here.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -136,11 +142,13 @@ internal virtual async Task FlushLogs(CloudTable table = null)
             }
             catch (Exception ex)
```
soninaren, line 144: NullRef exception would be caught here.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -136,11 +142,13 @@ internal virtual async Task FlushLogs(CloudTable table = null)
...
             {
-                _logger.LogError(ex, $"Unable to create table '{table.Name}' after {_tableCreationRetries} retries. Aborting write operation {ex}");
```
soninaren, line 139: There would be a new NullRef exception here as table is null

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
pragnagopa: Possible to add check for AzureWebJobsStorage app setting - disable the feature - perhaps here - https://github.com/Azure/azure-functions-host/blob/de24ce6714a2092f05f2f63ff34f64beba880ce4/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventLogger.cs#L37 or  here https://github.com/Azure/azure-functions-host/blob/de24ce6714a2092f05f2f63ff34f64beba880ce4/src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs#L148 ?

fabiocav: That pushes knowledge of the repository implementation into components that shouldn't have that responsibility. We should avoid sprinkling logic like this throughout the codebase.

Having this check in the actual repository implementation (the component that depends on that app setting) or the factory (allowing it to return a fallback implementation that does not have that dependency) would make sense.

mathewc: The pattern for this repository is based on TableStorageScaleMetricsRepository, and checking that code for usages of [GetTable](https://github.com/Azure/azure-functions-host/blob/de24ce6714a2092f05f2f63ff34f64beba880ce4/src/WebJobs.Script.WebHost/Scale/TableStorageScaleMetricsRepository.cs#L81) most do proper null checks. However I do see a couple that might not be. Please fix those as well.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -136,11 +142,13 @@ internal virtual async Task FlushLogs(CloudTable table = null)
...
-
+                _logger.LogError(ex, $"Unable to get table reference or create table. Aborting write operation {ex}");
```
mathewc: I don't think this error format is correct - don't need the "ex" param. LogError already logs the exception details.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -124,10 +124,16 @@ internal virtual async Task FlushLogs(CloudTable table = null)
...
 
-            table = table ?? GetDiagnosticEventsTable();
-
             try
             {
+                table = table ?? GetDiagnosticEventsTable();
+
+                if (table == null)
+                {
+                    _logger.LogError("Unable to get table reference. Aborting write operation");
```
mathewc, line 133: I don't think we want to see a persistent error on each time the flush timer fires. Instead, I don't think we should be running the flush timer at all if we don't have storage, and to Pragna's point, we can probably even disable the diagnostic logger even earlier so it isn't even operating?

soninaren: That makes sense. disabling the diagnostic event logger makes sense, other loggers will catch and log the statements anyway.

fabiocav: Instead of disabling the feature and pushing knowledge about the repository implementation to other components, I would recommend the approach I’ve suggested as that enables us to provide an efficient implementation with good separation of concerns.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
soninaren: @mathewc I see the code you are referencing in TableStorageScaleMetricsRepository. Would like to send a separate PR for that change we can keep the changes isolated.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -53,16 +53,16 @@ internal CloudTableClient TableClient
...
-                    {
-                        _logger.LogError("Azure Storage connection string is empty or invalid. Unable to write diagnostic events.");
-                    }
-
-                    if (CloudStorageAccount.TryParse(storageConnectionString, out CloudStorageAccount account))
+                    if (!string.IsNullOrEmpty(storageConnectionString)
+                        && CloudStorageAccount.TryParse(storageConnectionString, out CloudStorageAccount account))
                     {
                         var tableClientConfig = new TableClientConfiguration();
                         _tableClient = new CloudTableClient(account.TableStorageUri, account.Credentials, tableClientConfig);
```
kshyju, line 60: If more than one thread accesses "TableClient" internal property, wouldn't that cause this code to create multiple instances of `CloudTableClient`? and when it's time to do the flush, we may do that for the last one only. Should we make sure this code executes only for one thread ?

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventNullRepository.cs
```diff
@@ -0,0 +1,23 @@
...
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
+{
+    public class DiagnosticEventNullRepository : IDiagnosticEventRepository
+    {
+        public void WriteDiagnosticEvent(DateTime timestamp, string errorCode, LogLevel level, string message, string helpLink, Exception exception)
+        {
+            return;
```
kshyju: nit: Can remove this return.

kshyju (commented):
One question about thread safety of the table client creation code path.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
soninaren: In case there are 2 threads trying to create a tableClient at same time, They would end up creating a tableClient pointing to the same table. As a result, the logs would end up in the same table. As of right now only flush logs is using tableClient, which runs once every 10 minutes.

kshyju (approved):

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
kshyju: `_tableClient = new CloudTableClient`  - everytime we executes this , we are creating a new `CloudTableClient` intance. So if 2 threads ends up executing (line 60), one after the other, it is possible for the `TableClient` property to return 2 different objects.

soninaren: right, but both objects would end up pointing to the same table in azure storage. As the name is based on the date. So even if they end up using different objects the logs would end up in the same table.

pragnagopa (approved):
LGTM

PR #9643, opened 2023-10-25
logging error code as part of eventId for diagnostic events

### LINKED ISSUE #9347: Surface diagnostic events in a detector
We need a detector that will provide an aggregated list of diagnostic events logged within the specified time window (in App Lens), showing the error code and hit count.

Reference IPD Task 640

soninaren:

Updating diagnostic event log to record error code in kusto

resolves #9347

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script/Diagnostics/DiagnosticEventLoggerExtensions.cs
+++ b/src/WebJobs.Script/Diagnostics/DiagnosticEventLoggerExtensions.cs
@@ -25,14 +25,12 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics
 
         public static void LogDiagnosticEventInformation(this ILogger logger, string errorCode, string message, string helpLink)
         {
-            int eventId = 0; // Dummy value - diagnostic events does not log an eventId.
-            logger.LogDiagnosticEvent(LogLevel.Information, eventId, errorCode, message, helpLink, null);
+            logger.LogDiagnosticEvent(LogLevel.Information, 0, errorCode, message, helpLink, null);
         }
 
         public static void LogDiagnosticEventError(this ILogger logger, string errorCode, string message, string helpLink, Exception exception)
         {
-            int eventId = 0; // Dummy value - diagnostic events does not log an eventId.
-            logger.LogDiagnosticEvent(LogLevel.Error, eventId, errorCode, message, helpLink, exception);
+            logger.LogDiagnosticEvent(LogLevel.Error, 0, errorCode, message, helpLink, exception);
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/WebJobs.Script/Diagnostics/DiagnosticEventLoggerExtensions.cs
```diff
@@ -20,7 +20,7 @@ public static void LogDiagnosticEvent(this ILogger logger, LogLevel level, int e
                 { ScriptConstants.ErrorCodeKey, errorCode }
             };
 
-            logger.Log(level, eventId, stateDict, exception, (state, ex) => message);
+            logger.Log(level, new EventId(0, $"DiagnosticEvent-{errorCode}"), stateDict, exception, (state, ex) => message);
```
jviau: Should the event ID be passed in?

```suggestion
            logger.Log(level, new EventId(eventId, $"DiagnosticEvent-{errorCode}"), stateDict, exception, (state, ex) => message);
```

src/WebJobs.Script/Diagnostics/DiagnosticEventLoggerExtensions.cs
```diff
@@ -20,7 +20,7 @@ public static void LogDiagnosticEvent(this ILogger logger, LogLevel level, int e
                 { ScriptConstants.ErrorCodeKey, errorCode }
             };
 
-            logger.Log(level, eventId, stateDict, exception, (state, ex) => message);
```
brettsam: `eventId` isn't used anymore... can we remove it from the method completely and just hardcode it as 0?

brettsam (approved):

src/WebJobs.Script/Diagnostics/DiagnosticEventLoggerExtensions.cs
soninaren: event id is hard coded to 0, Not really used anywhere. The documentation for diagnostic event also suggests hardcoding it to 0. We use error code as the key identifier. Would want to keep it as is to avoid confusion.

jviau: I am not sure this is a good use of `EventId`.

1. This looks like a log breaking change. We are previously logging with a non-zero event ID, and now it will be 0.
2. These are unique diagnostic events after all, they should have their own event ID. 
3. We will have the same eventId with different event names - I think they are meant to always be 1 to 1.
4. Why do they need to be the same?

jviau: Looking at usage, I don't see anyone using a non-zero eventId. In that case, I recommend we remove `eventId` from the API entirely.

But I am also curious why we need error code in the `EventId` at all. It is already an attribute `ScriptCnstants.ErrorCodeKey` - can the kusto query not consume that attribute?

jviau (approved):

PR #9644, opened 2023-10-25
KeyVaultSecretsRepository comparisons are now case insensitive 

### LINKED ISSUE #9620: KeyVaultSecretsRepository comparisons need to be case insensitive
A recent CRI showed us that if you manually create keys as fully lowercase -- everything kinda goes off the rails. We need all comparisons in the KeyVaultSecretsRepository to be case insensitive.

One example:
1. Manually create a `host--masterkey--master` (all lowercase) secret in a KeyVault.
2. The host will do a case-sensitive comparison and effectively think we do not have any keys. (https://github.com/Azure/azure-functions-host/blob/97613c60f14a838933f0c0ef565e5f460d059126/src/WebJobs.Script.WebHost/Security/KeyManagement/KeyVaultSecretsRepository.cs#L128C11-L128C11)
3. It will then go to regenerate and persist the new keys.
4. As part of this, it deletes any keys it does not think we need anymore. And because this one is not a match, it deletes it. In addition to all the comparisons needing to be case-insensitive, this dictionary lookup also needs to be: https://github.com/Azure/azure-functions-host/blob/97613c60f14a838933f0c0ef565e5f460d059126/src/WebJobs.Script.WebHost/Security/KeyManagement/KeyVaultSecretsRepository.cs#L81C58-L81C58

This behavior is bad, but especially bad if the KeyVault has soft-delete enabled. This means it will throw an exception and we'll not be able to recreate the keys.

cjaliaga:
Updating comparisons on `KeyVaultSecretsRepository` to ignore casing as KeyVault secrets are case insensitive. 
resolves #9620

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.WebHost/Security/KeyManagement/KeyVaultSecretsRepository.cs
+++ b/src/WebJobs.Script.WebHost/Security/KeyManagement/KeyVaultSecretsRepository.cs
@@ -50,6 +50,12 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
             _environment = environment ?? throw new ArgumentNullException(nameof(environment));
         }
 
+        // For testing
+        internal KeyVaultSecretsRepository(SecretClient secretClient, string secretsSentinelFilePath, ILogger logger, IEnvironment environment) : base(secretsSentinelFilePath, logger, environment)
+        {
+            _secretClient = new Lazy<SecretClient>(() => secretClient);
+        }
+
         public override bool IsEncryptionSupported
         {
             get
--- a/src/WebJobs.Script.WebHost/Security/KeyManagement/KeyVaultSecretsRepository.cs
+++ b/src/WebJobs.Script.WebHost/Security/KeyManagement/KeyVaultSecretsRepository.cs
@@ -9,6 +9,7 @@ using Azure;
 using Azure.Core;
 using Azure.Identity;
 using Azure.Security.KeyVault.Secrets;
+using Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics.Extensions;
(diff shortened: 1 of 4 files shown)


src/WebJobs.Script.WebHost/Security/KeyManagement/KeyVaultSecretsRepository.cs
```diff
@@ -75,7 +81,7 @@ public override async Task WriteAsync(ScriptSecretsType type, string functionNam
             List<Task> deleteTasks = new List<Task>();
             string prefix = (type == ScriptSecretsType.Host) ? HostPrefix : FunctionPrefix + Normalize(functionName);
 
-            foreach (SecretProperties item in await FindSecrets(secretsPages, x => x.Name.StartsWith(prefix)))
+            foreach (SecretProperties item in await FindSecrets(secretsPages, x => x.Name.StartsWith(prefix, StringComparison.OrdinalIgnoreCase)))
             {
                 // Delete only keys which no longer exist in passed-in secrets
                 if (!dictionary.Keys.Contains(item.Name))
```
brettsam, line 88: Can you add a log message right below here telling us that we're deleting a key, and which one it is? That would help in the future identify exactly when we are deleting customer's keys.

brettsam: And same goes for when we Set the keys below -- it'd be a huge help with debugging.

brettsam (approved):

kshyju (approved):
Update the [release notes](https://github.com/Azure/azure-functions-host/blob/dev/release_notes.md) to include this change ?

Issue #9648, opened 2023-10-28
Flex Consumption SKU for Placeholder 


Tonewall:
#### What problem would the feature you're requesting solve? Please describe.
We will have three different scenarios for Linux Consumption hosts:
1. Consumption V1 running on Atlas
2. Consumption V1 running on Legion (new)
3. Flex Consumption running on Legion

We can differentiate between scenarios **1 vs 2** and **1 vs 3** based on the platform it's running on (IsLinuxConsumptionOnAtlas() and IsLinuxConsumptionOnLegion())
We will need a way to differentiate between scenarios **2 vs 3** for when we migrate cv1 onto Legion. 
The **_WEBSITE_SKU_** can only be referenced for specialized hosts, and the environment will be the same for consumption v1 and flex consumption because both will be running on Legion.

#### Describe the solution you'd like
FunctionsPodSerivce injects **_WEBSITE_SKU_NAME_** for Flex Consumption pods which will allow us to determine the sku of the host even when in placeholder mode ([code](https://msazure.visualstudio.com/One/_git/AAPT-Antares-Websites?path=/src/Hosting/Functions/FunctionsPodService/PodBuilder/FunctionsWorkerPodBuilderStrategy.cs&version=GBdev&line=60&lineEnd=61&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents)). 
For the 1st and 2nd scenarios, **_WEBSITE_SKU_NAME_** will not be set (no new changes)
For the 3rd scenario, **_WEBSITE_SKU_NAME_** will be set to _FlexConsumption_ [PR](https://msazure.visualstudio.com/One/_git/AAPT-Antares-Websites/pullrequest/8991851)

#### Describe alternatives you've considered
Thought about just using the existing **_WEBSITE_SKU_** but this could result in unexpected changes

PR #9649, opened 2023-10-28
U/tonychoi/cv1 migration runtime environment

Tonewall:
Issue #9648 

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
@@ -417,7 +417,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
                 { "triggers", triggersArray }
             };
 
-            if (_environment.IsFlexConsumptionSku())
+            if (_environment.IsFlexConsumptionSku() || _environment.IsConsumptionV1RunningOnLegion())
             {
                 // Currently we're only sending the HostId for Flex Consumption. Eventually we'll do this for all SKUs.
                 // When the HostId is sent, ScaleController will use it directly rather than compute it itself.
--- a/src/WebJobs.Script/Environment/EnvironmentExtensions.cs
+++ b/src/WebJobs.Script/Environment/EnvironmentExtensions.cs
@@ -231,6 +231,26 @@ namespace Microsoft.Azure.WebJobs.Script
         }
 
         /// <summary>
+        /// Gets a value indicating whether the application is migrated Consumption V1 running in Legion environment.
+        /// </summary>
+        /// <param name="environment">The environment to verify.</param>
+        /// <returns><see cref="true"/> if migrated Consumption V1 running in Legion, false otherwise.</returns>
+        public static bool IsConsumptionV1RunningOnLegion(this IEnvironment environment)
+        {
+            // Check if host is running in Legion and if host is in placeholder mode
+            // There shouldn't be any difference for Legion placeholders running in Flex vs migrated Consumption V1
+            if (!environment.IsLinuxConsumptionOnLegion() || environment.IsPlaceholderModeEnabled())
+            {
+                return false;
+            }
+
+            // Check if the SKU is set to FlexConsumptionSku
+            string sku = environment.GetEnvironmentVariable(AzureWebsiteSku);
+            bool isFlex = string.Equals(sku, ScriptConstants.FlexConsumptionSku, StringComparison.OrdinalIgnoreCase);
+            return !isFlex;
+        }
+
+        /// <summary>
         /// Returns true if the app is running on Virtual Machine Scale Sets (VMSS).
         /// </summary>
         public static bool IsVMSS(this IEnvironment environment)
--- a/src/WebJobs.Script/Environment/EnvironmentExtensions.cs
+++ b/src/WebJobs.Script/Environment/EnvironmentExtensions.cs
@@ -209,7 +209,7 @@ namespace Microsoft.Azure.WebJobs.Script
         /// <returns><see cref="true"/> if running in one of our Consumption SKUs, false otherwise.</returns>
         public static bool IsConsumptionSku(this IEnvironment environment)
         {
-            return IsWindowsConsumption(environment) || IsAnyLinuxConsumption(environment) || IsFlexConsumptionSku(environment);
+            return IsWindowsConsumption(environment) || IsAnyLinuxConsumption(environment) || IsFlexConsumptionSku(environment) || IsConsumptionV1RunningOnLegion(environment);
         }
 
         /// <summary>
@@ -227,7 +227,7 @@ namespace Microsoft.Azure.WebJobs.Script
 
             // when in placeholder mode, site settings like SKU are not available
             // to enable this check to run in both modes, we check additional settings
-            return environment.IsLinuxConsumptionOnLegion();
+            return environment.IsPlaceholderModeEnabled() && environment.IsLinuxConsumptionOnLegion();
         }
 
         /// <summary>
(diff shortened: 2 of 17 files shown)


liliankasem:
@Tonewall please make sure you go through the PR checklist and update the title of the PR to be descriptive of the change

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -219,15 +219,31 @@ public static bool IsConsumptionSku(this IEnvironment environment)
         /// <returns><see cref="true"/> if running in the FlexConsumption Sku, false otherwise.</returns>
         public static bool IsFlexConsumptionSku(this IEnvironment environment)
         {
-            string value = environment.GetEnvironmentVariable(AzureWebsiteSku);
-            if (string.Equals(value, ScriptConstants.FlexConsumptionSku, StringComparison.OrdinalIgnoreCase))
+            string sku = environment.GetEnvironmentVariable(WebsiteSkuName);
+
+            // Currently set the sku to Dynamic in FPS
+            // TODO: Remove this once FPS sets the value to FlexConsumption
+            if (string.Equals(sku, ScriptConstants.DynamicSku, StringComparison.OrdinalIgnoreCase))
```
mathewc, line 226: This doesn't look right to me - if the sku is set to Dynamic we don't want to return true here. We only want to return true when it's "FlexConsumption". I get that you're saying that there's currently some buggy/incorrect FPS behavior that is setting this to Dynamic for Flex, but we shouldn't code to that here. When FPS is using the right value, we can add an additional check for WEBSITE_SKU_NAME=FlexConsumption if we needed to, but I don't see why we need to.

Also, having both WEBSITE_SKU and WEBSITE_SKU_NAME is confusing.

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -219,15 +219,31 @@ public static bool IsConsumptionSku(this IEnvironment environment)
         /// <returns><see cref="true"/> if running in the FlexConsumption Sku, false otherwise.</returns>
         public static bool IsFlexConsumptionSku(this IEnvironment environment)
         {
-            string value = environment.GetEnvironmentVariable(AzureWebsiteSku);
-            if (string.Equals(value, ScriptConstants.FlexConsumptionSku, StringComparison.OrdinalIgnoreCase))
+            string sku = environment.GetEnvironmentVariable(WebsiteSkuName);
```
mathewc, line 222: I think we should leave this as it was, based on WEBSITE_SKU, and using IsLinuxConsumptionOnLegion as a backup check. Why isn't that sufficient?

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -209,7 +209,7 @@ public static bool IsWindowsConsumption(this IEnvironment environment)
         /// <returns><see cref="true"/> if running in one of our Consumption SKUs, false otherwise.</returns>
         public static bool IsConsumptionSku(this IEnvironment environment)
         {
-            return IsWindowsConsumption(environment) || IsAnyLinuxConsumption(environment) || IsFlexConsumptionSku(environment);
+            return IsWindowsConsumption(environment) || IsAnyLinuxConsumption(environment) || IsFlexConsumptionSku(environment) || IsConsumptionV1RunningOnLegion(environment);
```
mathewc, line 212: I would think that IsAnyLinuxConsumption would return true for cv1 on Legion without needing to add a new check here?

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -219,15 +219,31 @@ public static bool IsConsumptionSku(this IEnvironment environment)
...
+            // Currently set the sku to Dynamic in FPS
+            // TODO: Remove this once FPS sets the value to FlexConsumption
+            if (string.Equals(sku, ScriptConstants.DynamicSku, StringComparison.OrdinalIgnoreCase))
             {
                 return true;
             }
+            return string.Equals(sku, ScriptConstants.FlexConsumption, StringComparison.OrdinalIgnoreCase);
+        }
+
+        public static bool IsConsumptionV1(this IEnvironment environment)
```
mathewc, line 233: I'd rename this to IsLinuxConsumptionV1 if that's what you're trying to capture?

src/WebJobs.Script/Host/ScriptHostIdProvider.cs
```diff
@@ -50,7 +50,7 @@ internal static HostIdResult GetDefaultHostId(IEnvironment environment, ScriptAp
         {
             HostIdResult result = new HostIdResult();
 
-            if (environment.IsFlexConsumptionSku())
+            if (environment.IsFlexConsumptionSku() || environment.IsConsumptionV1RunningOnLegion())
```
mathewc, line 53: Why are you making this change? We don't intend to change host behaviors for Cv1 when running on Legion. It's just a migration.

src/WebJobs.Script/ScriptConstants.cs
```diff
@@ -112,6 +112,7 @@ public static class ScriptConstants
         public const string FlexConsumptionSku = "FlexConsumption";
         public const string ElasticPremiumSku = "ElasticPremium";
         public const string DefaultProductionSlotName = "production";
+        public const string FlexConsumption = "FlexConsumption";
```
mathewc, line 115: There's already a constant defined above for this

mathewc:
The referenced issue still doesn't make it clear exactly what you're trying to do here. Is the goal simply to have a single new environment extension helper that can identify Linux Consumption V1 running on Legion? Seems like we can add that in a more simple/direct way.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -417,7 +417,7 @@ private JObject GetMinimalPayload(string hostId, JArray triggersArray)
                 { "triggers", triggersArray }
             };
 
-            if (_environment.IsFlexConsumptionSku())
+            if (_environment.IsFlexConsumptionSku() || _environment.IsConsumptionV1RunningOnLegion())
```
mathewc, line 420: Why are you making this change? We don't intend to change host behaviors for Cv1 when running on Legion. It's just a migration.

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
Tonewall: Yes, it is a little confusing. I could just replace that environment variable entirely with a new environment variable (e.g. IS_FLEX or CONSUMPTION_VERSION).

And you're saying that we should hold off the change until the Antares change is already released with the correct environment variable, right?

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
Tonewall: The reason for the change is because our current IsFlexConsumptionSku() doesn't actually check the sku of the pods correctly. It will return true for both flex consumption and migrated cv1 pods as long as it's running on legion. And if we want any variance between the migrated cv1 and flex consumption, we will need to have a definitive way to distinguish between the two.

So far, there doesn't seem to be any changes needed for migrated cv1 apps for the host runtime for EUAP, but there could be a few. For example, we currently emit FES for flex consumption, but the mesh process doesn't do anything with the data, so we could get rid of it for flex consumption containers. However, I believe our plan is to maintain FES for migrated cv1 pods. Another change we could have for migrated cv1 and flex would be billing. I believe we want to keep the billing the same for migrated cv1 as the existing cv1.

@balag0 What do you think?

Tonewall:
Closing PR to wait for changes on FPS first

PR #9655, opened 2023-11-01
Performance improvements for file logging

mathewc:
Making some performance improvements to file logging based on a recent CRI investigation. In that investigation, we were seeing thread exhaustion in the host process with many threads blocked on file IO. When the issue occurred, the app had FileLoggingMode enabled (see doc for [fileLoggingMode](https://learn.microsoft.com/en-us/azure/azure-functions/functions-host-json#logging)) because they had the portal open on the app. This causes the log stream to also go to files. At the time the app was under high load, so there was a lot of file logging happening. Compounding things was the fact that the customer was also logging errors, triggering our "force flush on error" logic.

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script/Diagnostics/FileLogger.cs
+++ b/src/WebJobs.Script/Diagnostics/FileLogger.cs
@@ -92,14 +92,8 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics
             try
             {
                 _fileWriter.AppendLine(formattedMessage);
-
-                // flush errors immediately
-                if (logLevel == LogLevel.Error || exception != null)
-                {
-                    _fileWriter.Flush();
-                }
             }
-            catch (Exception)
+            catch (Exception ex) when (!ex.IsFatal())
             {
                 // Make sure the Logger doesn't throw if there are Exceptions (disk full, etc).
             }
--- a/src/WebJobs.Script/Diagnostics/FileWriter.cs
+++ b/src/WebJobs.Script/Diagnostics/FileWriter.cs
@@ -30,7 +30,7 @@ namespace Microsoft.Azure.WebJobs.Script
         private Timer _flushTimer;
         private ConcurrentQueue<string> _logBuffer = new ConcurrentQueue<string>();
 
-        public FileWriter(string logFilePath)
+        public FileWriter(string logFilePath, int flushIntervalMs = LogFlushIntervalMs)
         {
             _logFilePath = logFilePath;
             _instanceId = GetInstanceId();
@@ -39,8 +39,8 @@ namespace Microsoft.Azure.WebJobs.Script
             // start a timer to flush accumulated logs in batches
             _flushTimer = new Timer
             {
-                AutoReset = true,
-                Interval = LogFlushIntervalMs
+                AutoReset = false,
+                Interval = flushIntervalMs
             };
             _flushTimer.Elapsed += OnFlushLogs;
             _flushTimer.Start();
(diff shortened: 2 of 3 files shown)


src/WebJobs.Script/Diagnostics/FileLogger.cs
```diff
@@ -27,6 +28,9 @@ public FileLogger(string categoryName, IFileWriter fileWriter, Func<bool> isFile
             _categoryName = categoryName;
             _logType = logType;
             _scopeProvider = scopeProvider ?? throw new ArgumentNullException(nameof(scopeProvider));
+
+            _flushLogs = () => _fileWriter.Flush();
+            _flushLogs = _flushLogs.Debounce();
```
mathewc: Default debounce interval is 300ms which I think is fine here. We can adjust if needed based on load testing which we're doing.

src/WebJobs.Script/Diagnostics/FileLogger.cs
```diff
@@ -93,10 +97,10 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
             {
                 _fileWriter.AppendLine(formattedMessage);
 
-                // flush errors immediately
+                // we issue an immediate flush to ensure errors are logged as quickly as possible
                 if (logLevel == LogLevel.Error || exception != null)
                 {
-                    _fileWriter.Flush();
+                    _flushLogs();
```
mathewc: Previously every single error logged would result in a separate file write. These are now debounced to a max of 1 write every 300ms.

src/WebJobs.Script/Diagnostics/FileWriter.cs
```diff
@@ -39,8 +39,8 @@ public FileWriter(string logFilePath)
             // start a timer to flush accumulated logs in batches
             _flushTimer = new Timer
             {
-                AutoReset = true,
-                Interval = LogFlushIntervalMs
+                AutoReset = false,
```
mathewc, line 42: Analysis of the dump file also showed many outstanding *concurrent* flush operations coming from timer invocations. I.e. when under load, file contention might result in many timer invocations backing up waiting of file operations, resulting in many invocations outstanding concurrently.

Instead we want there to be at most 1 file flush per interval. We get this by manually restarting the timer after every invocation.

src/WebJobs.Script/Diagnostics/FileLogger.cs
fabiocav: Worth mentioning that with the debouncing, this is no longer going to be synchronous, which was another issue causing problems in the RPC logging path. 

Have you validated the behavior when an exception is thrown as a result, though? With your debounce change, that exception will now be thrown in an unobserved `Task`, and no longer caught by your exception handler below.

We need to make sure your debounced action is handling any exceptions previously handled in this block.

fabiocav (changes requested):

fabiocav (approved):

fabiocav:
Approved the changes, but some of the tests seem to be failing here, likely due to the timing difference introducing races.

brettsam (approved):

src/WebJobs.Script/Diagnostics/FileLogger.cs
```diff
@@ -92,14 +92,8 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
             try
             {
                 _fileWriter.AppendLine(formattedMessage);
-
-                // flush errors immediately
```
mathewc, line 96: Updated the PR to just remove this flush on error behavior. The underlying file is flushed on a 1s interval already, so we really won't save much time attempting to force a flush. Also, implementing the logic to force flush with a limited frequency would require synchronization and I don't want to put that on this logging hotpath.

PR #9675, opened 2023-11-09
FlexConsumption Non-Encrypted Assign Endpoint for Legion

### LINKED ISSUE #9674: Non-Encrypted Assign Endpoint for Legion Pods
Adding a new /admin/instance/assign endpoint to accept non-encrypted content for Legion hosts. Encrypting host context is not needed for legion hosts because the AIP's are not publicly accessible outside of the nested vm's

Getting rid of encryption for the host context

Tonewall:

resolves #9674 

Adding a new /admin/instance/assign endpoint to accept non-encrypted content for Legion hosts. Encrypting host context is not needed for legion hosts because the AIP's are not publicly accessible outside of the nested vm's

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
+++ b/src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
@@ -67,7 +67,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Controllers
         [HttpPost]
         [Route("admin/instance/legion-assign")]
         [Authorize(Policy = PolicyNames.AdminAuthLevel)]
-        public async Task<IActionResult> LegionAssign([FromBody] HostAssignmentContext hostAssignmentContext)
+        public IActionResult LegionAssign([FromBody] HostAssignmentContext hostAssignmentContext)
         {
             _logger.LogDebug($"Starting legion container assignment for host : {Request?.Host}");
 
--- a/src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
+++ b/src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
@@ -64,6 +64,27 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Controllers
                 : StatusCode(StatusCodes.Status409Conflict, "Instance already assigned");
         }
 
+        [HttpPost]
+        [Route("admin/instance/legion-assign")]
+        [Authorize(Policy = PolicyNames.AdminAuthLevel)]
+        public async Task<IActionResult> LegionAssign([FromBody] HostAssignmentContext hostAssignmentContext)
+        {
+            _logger.LogDebug($"Starting legion container assignment for host : {Request?.Host}");
+
+            if (!_environment.IsFlexConsumptionSku())
+            {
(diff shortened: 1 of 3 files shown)


src/WebJobs.Script.WebHost/StartupContextProvider.cs
```diff
@@ -155,6 +155,24 @@ public virtual HostAssignmentContext SetContext(EncryptedHostAssignmentContext e
             return hostAssignmentContext;
         }
 
+        /// <summary>
+        /// Decrypt and deserialize the specified context, and apply values from it to the
+        /// startup cache context.
+        /// </summary>
+        /// <param name="hostAssignmentContext">The encrypted assignment context.</param>
```
CooperLink, line 162: Is this still encrypted?

CooperLink: Update comment

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
```diff
@@ -64,6 +64,27 @@ public async Task<IActionResult> Assign([FromBody] EncryptedHostAssignmentContex
                 : StatusCode(StatusCodes.Status409Conflict, "Instance already assigned");
         }
 
+        [HttpPost]
+        [Route("admin/instance/legion-assign")]
```
mathewc, line 68: I don't think we need to add a new endpoint for this. Just use the existing admin/instance/assign endpoint. It can handle both scenarios.

Tonewall: Are you saying make changes to the existing admin/instance/assign endpoint to be able to accept both encrypted and non-encrypted host assignment context? Currently, the endpoint specifically takes EncryptedHostAssignmentContext  and decrypts it via SetContext() and we have the code changes in websites that will send either the encrypted or the non-encrypted context based on the assign endpoint [code](https://msazure.visualstudio.com/One/_git/AAPT-Antares-Websites?path=/src/Hosting/DataService/Microsoft.Web.Hosting.DataService/Runtime/LinuxConsumption/Specialization/Legion/PayloadHelper.cs&version=GBdev&line=141&lineEnd=155&lineStartColumn=13&lineEndColumn=85&lineStyle=plain&_a=contents)

mathewc: Yes, I'm saying that we don't want a new admin/instance/legion-assign endpoint on the host. The existing endpoint suffices and can just vary its behavior based on the incoming payload. Also, "legion" is really just an internal codename, so shouldn't be making its way into our public API contract. @fabiocav might also have thoughts on this.

fabiocav: Indeed. Thank you for flagging this, @mathewc 

@Tonewall beyond just the name, can we also please make sure we don't have any routes with references specific SKUs or SKU specific logic? We can discuss the design if needed.

Let's also please strive to avoid a lot of SKU specific logic sprinkled throughout the code. As much as possible, let's either consider the existing extensibility points to add support for what you need, or discuss potentially new ones to introduce the logic you need.

fabiocav (changes requested):

PR #9730, opened 2023-12-08
[pack] Implementation of host metrics meter and provider

### LINKED ISSUE #9763: Implement host metrics and integrate with the metrics publisher
Implement .NET metric meters to capture host metrics and integrate with the FlexConsumptionMetricsPublisher to write these metrics to a file.

This work does not need to capture every metric but lays out the groundwork to additional metrics in future.

P0 metrics to capture:

- AppFailureCount
- ActiveFunctionCount

liliankasem:

resolves #9763

### Pull request checklist

    * [ ] Otherwise: Documentation issue linked to PR (in progress)

### Additional information

Introduces the use of .NET Metrics to capture function host metrics, following OTel schematic standards. This PR also integrates with `FlexConsumptionMetricsPublisher` to capture these new host metrics into a file.

### DIFF
--- a/src/WebJobs.Script.WebHost/Diagnostics/FunctionInstanceLogCollectorProvider.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/FunctionInstanceLogCollectorProvider.cs
@@ -5,6 +5,7 @@ using System;
 using Microsoft.Azure.WebJobs.Host.Executors;
 using Microsoft.Azure.WebJobs.Host.Loggers;
 using Microsoft.Azure.WebJobs.Script.Diagnostics;
+using Microsoft.Azure.WebJobs.Script.Metrics;
 using Microsoft.Azure.WebJobs.Script.WebHost.Storage;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
@@ -15,20 +16,22 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
     {
         private readonly IFunctionMetadataManager _metadataManager;
         private readonly IMetricsLogger _metrics;
+        private readonly IHostMetrics _hostMetrics;
         private readonly IConfiguration _configuration;
         private readonly ILoggerFactory _loggerFactory;
 
-        public FunctionInstanceLogCollectorProvider(IFunctionMetadataManager metadataManager, IMetricsLogger metrics, IConfiguration configuration, ILoggerFactory loggerFactory)
+        public FunctionInstanceLogCollectorProvider(IFunctionMetadataManager metadataManager, IMetricsLogger metrics, IHostMetrics hostMetrics, IConfiguration configuration, ILoggerFactory loggerFactory)
         {
             _metadataManager = metadataManager ?? throw new ArgumentNullException(nameof(metadataManager));
             _metrics = metrics ?? throw new ArgumentNullException(nameof(metrics));
+            _hostMetrics = hostMetrics ?? throw new ArgumentNullException(nameof(hostMetrics));
             _configuration = configuration ?? throw new ArgumentNullException(nameof(configuration));
             _loggerFactory = loggerFactory ?? throw new ArgumentNullException(nameof(loggerFactory));
         }
 
         public IAsyncCollector<FunctionInstanceLogEntry> Create()
         {
-            return new FunctionInstanceLogger(_metadataManager, _metrics, _configuration, _loggerFactory);
+            return new FunctionInstanceLogger(_metadataManager, _metrics, _hostMetrics, _configuration, _loggerFactory);
         }
     }
 }
--- a/src/WebJobs.Script.WebHost/Diagnostics/FunctionInstanceLogger.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/FunctionInstanceLogger.cs
@@ -10,6 +10,7 @@ using Microsoft.Azure.WebJobs.Host.Loggers;
 using Microsoft.Azure.WebJobs.Logging;
 using Microsoft.Azure.WebJobs.Script.Description;
 using Microsoft.Azure.WebJobs.Script.Diagnostics;
+using Microsoft.Azure.WebJobs.Script.Metrics;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
 
@@ -21,15 +22,17 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 
         private readonly IMetricsLogger _metrics;
         private readonly IFunctionMetadataManager _metadataManager;
+        private readonly IHostMetrics _hostMetrics;
         private ConcurrentDictionary<(string, string, bool, bool), string> _eventDataCache = new ConcurrentDictionary<(string, string, bool, bool), string>();
         private ConcurrentDictionary<BindingMetadata, string> _bindingMetricEventNames = new ConcurrentDictionary<BindingMetadata, string>();
 
         public FunctionInstanceLogger(
             IFunctionMetadataManager metadataManager,
             IMetricsLogger metrics,
+            IHostMetrics hostMetrics,
             IConfiguration configuration,
             ILoggerFactory loggerFactory)
-            : this(metadataManager, metrics)
+            : this(metadataManager, metrics, hostMetrics)
         {
             ArgumentNullException.ThrowIfNull(configuration);
             ArgumentNullException.ThrowIfNull(loggerFactory);
@@ -41,9 +44,10 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
             }
         }
 
-        internal FunctionInstanceLogger(IFunctionMetadataManager metadataManager, IMetricsLogger metrics)
+        internal FunctionInstanceLogger(IFunctionMetadataManager metadataManager, IMetricsLogger metrics, IHostMetrics hostMetrics)
         {
             _metrics = metrics ?? throw new ArgumentNullException(nameof(metrics));
+            _hostMetrics = hostMetrics ?? throw new ArgumentNullException(nameof(hostMetrics));
             _metadataManager = metadataManager ?? throw new ArgumentNullException(nameof(metadataManager));
         }
 
@@ -63,6 +67,8 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 
         private void StartFunction(FunctionInstanceLogEntry item)
         {
+            _hostMetrics.IncrementStartedInvocationCount();
+
             if (_metadataManager.TryGetFunctionMetadata(item.LogName, out FunctionMetadata function))
             {
                 var startedEvent = new FunctionStartedEvent(item.FunctionInstanceId, function);
--- a/src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
+++ b/src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
@@ -9,10 +9,12 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Script.Diagnostics;
 using Microsoft.Azure.WebJobs.Script.Diagnostics.Extensions;
+using Microsoft.Azure.WebJobs.Script.Metrics;
 using Microsoft.Azure.WebJobs.Script.WebHost.Configuration;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 
 namespace Microsoft.Azure.WebJobs.Script.WebHost.Metrics
 {
@@ -22,6 +24,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Metrics
         private readonly FlexConsumptionMetricsPublisherOptions _options;
         private readonly IEnvironment _environment;
         private readonly ILogger<FlexConsumptionMetricsPublisher> _logger;
+        private readonly IHostMetricsProvider _metricsProvider;
         private readonly object _lock = new object();
         private readonly IFileSystem _fileSystem;
 
@@ -33,13 +36,15 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Metrics
         private TimeSpan _metricPublishInterval;
         private TimeSpan _initialPublishDelay;
 
-        public FlexConsumptionMetricsPublisher(IEnvironment environment, IOptionsMonitor<StandbyOptions> standbyOptions, IOptions<FlexConsumptionMetricsPublisherOptions> options, ILogger<FlexConsumptionMetricsPublisher> logger, IFileSystem fileSystem)
+        public FlexConsumptionMetricsPublisher(IEnvironment environment, IOptionsMonitor<StandbyOptions> standbyOptions, IOptions<FlexConsumptionMetricsPublisherOptions> options,
+            ILogger<FlexConsumptionMetricsPublisher> logger, IFileSystem fileSystem, IHostMetricsProvider metricsProvider)
         {
             _standbyOptions = standbyOptions ?? throw new ArgumentNullException(nameof(standbyOptions));
             _environment = environment ?? throw new ArgumentNullException(nameof(environment));
             _options = options?.Value ?? throw new ArgumentNullException(nameof(options));
             _logger = logger ?? throw new ArgumentNullException(nameof(logger));
             _fileSystem = fileSystem ?? new FileSystem();
+            _metricsProvider = metricsProvider ?? throw new ArgumentNullException(nameof(metricsProvider));
 
             if (_standbyOptions.CurrentValue.InStandbyMode)
             {
@@ -90,7 +95,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Metrics
         {
             try
             {
-                if (FunctionExecutionCount == 0 && FunctionExecutionTimeMS == 0 && !IsAlwaysReady)
+                if (FunctionExecutionCount == 0 && FunctionExecutionTimeMS == 0 && !IsAlwaysReady && !_metricsProvider.HasMetrics())
                 {
                     // no activity to report
                     return;
@@ -108,6 +113,16 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Metrics
                         ExecutionTimeMS = FunctionExecutionTimeMS,
                         IsAlwaysReady = IsAlwaysReady
                     };
+
+                    var scaleMetrics = _metricsProvider.GetHostMetricsOrNull();
+                    if (scaleMetrics is not null)
+                    {
+                        metrics.InstanceId = _metricsProvider.InstanceId;
+                        metrics.FunctionGroup = _metricsProvider.FunctionGroup;
+                        metrics.AppFailureCount = scaleMetrics.TryGetValue(HostMetrics.AppFailureCount, out long appFailureCount) ? appFailureCount : 0;
+                        metrics.StartedInvocationCount = scaleMetrics.TryGetValue(HostMetrics.StartedInvocationCount, out long startedInvocationCount) ? startedInvocationCount : 0;
+                        metrics.ActiveInvocationCount = scaleMetrics.TryGetValue(HostMetrics.ActiveInvocationCount, out long activeInvocationCount) ? activeInvocationCount : 0;
+                    }
 
                     FunctionExecutionTimeMS = FunctionExecutionCount = 0;
                 }
@@ -296,6 +311,32 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Metrics
             /// AlwaysReady.
             /// </summary>
             public bool IsAlwaysReady { get; set; }
+
+            /// <summary>
+            /// Gets or sets the instance Id.
+            /// </summary>
+            public string InstanceId { get; set; }
+
+            /// <summary>
+            /// Gets or sets the function group name. This can be either http, durable or
+            /// the name of a function.
+            /// </summary>
+            public string FunctionGroup { get; set; }
+
+            /// <summary>
+            /// Gets or sets the total number of permanent host failures.
+            /// </summary>
+            public long AppFailureCount { get; set; }
+
+            /// <summary>
+            /// Gets or sets the total number of in-progress function invocations.
+            /// </summary>
+            public long ActiveInvocationCount { get; set; }
+
+            /// <summary>
+            /// Gets or sets the total number of function invocations that have started.
+            /// </summary>
+            public long StartedInvocationCount { get; set; }
         }
     }
 }
--- a/src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
+++ b/src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
@@ -111,14 +111,14 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Metrics
                         TotalTimeMS = (long)_intervalStopwatch.GetElapsedTime().TotalMilliseconds,
                         ExecutionCount = FunctionExecutionCount,
                         ExecutionTimeMS = FunctionExecutionTimeMS,
-                        IsAlwaysReady = IsAlwaysReady
+                        IsAlwaysReady = IsAlwaysReady,
+                        InstanceId = _metricsProvider.InstanceId,
+                        FunctionGroup = _metricsProvider.FunctionGroup
                     };
 
                     var scaleMetrics = _metricsProvider.GetHostMetricsOrNull();
                     if (scaleMetrics is not null)
                     {
-                        metrics.InstanceId = _metricsProvider.InstanceId;
-                        metrics.FunctionGroup = _metricsProvider.FunctionGroup;
                         metrics.AppFailureCount = scaleMetrics.TryGetValue(HostMetrics.AppFailureCount, out long appFailureCount) ? appFailureCount : 0;
                         metrics.StartedInvocationCount = scaleMetrics.TryGetValue(HostMetrics.StartedInvocationCount, out long startedInvocationCount) ? startedInvocationCount : 0;
                         metrics.ActiveInvocationCount = scaleMetrics.TryGetValue(HostMetrics.ActiveInvocationCount, out long activeInvocationCount) ? activeInvocationCount : 0;
--- a/src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
+++ b/src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
@@ -11,6 +11,7 @@ using Microsoft.Azure.WebJobs.Script.Config;
 using Microsoft.Azure.WebJobs.Script.Configuration;
 using Microsoft.Azure.WebJobs.Script.Diagnostics;
 using Microsoft.Azure.WebJobs.Script.Grpc;
+using Microsoft.Azure.WebJobs.Script.Metrics;
 using Microsoft.Azure.WebJobs.Script.Middleware;
 using Microsoft.Azure.WebJobs.Script.WebHost.Configuration;
 using Microsoft.Azure.WebJobs.Script.WebHost.ContainerManagement;
@@ -90,6 +91,10 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
             services.TryAddSingleton<IStandbyManager, StandbyManager>();
             services.TryAddSingleton<IServiceCollection>(services);
             services.TryAddSingleton<IScriptHostBuilder, DefaultScriptHostBuilder>();
+
+            // Metrics
+            services.AddSingleton<IHostMetricsProvider, HostMetricsProvider>();
+            services.AddSingleton<IHostMetrics, HostMetrics>();
 
             // Linux container services
             services.AddLinuxContainerServices();
@@ -271,7 +276,8 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
                     var options = s.GetService<IOptions<FlexConsumptionMetricsPublisherOptions>>();
                     var standbyOptions = s.GetService<IOptionsMonitor<StandbyOptions>>();
                     var logger = s.GetService<ILogger<FlexConsumptionMetricsPublisher>>();
-                    return new FlexConsumptionMetricsPublisher(environment, standbyOptions, options, logger, new FileSystem());
+                    var metricsProvider = s.GetService<IHostMetricsProvider>();
+                    return new FlexConsumptionMetricsPublisher(environment, standbyOptions, options, logger, new FileSystem(), metricsProvider);
                 }
                 else if (environment.IsLinuxMetricsPublishingEnabled())
                 {
--- a/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
+++ b/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
@@ -17,6 +17,7 @@ using Microsoft.Azure.WebJobs.Logging;
 using Microsoft.Azure.WebJobs.Script.Configuration;
 using Microsoft.Azure.WebJobs.Script.Diagnostics;
 using Microsoft.Azure.WebJobs.Script.Eventing;
+using Microsoft.Azure.WebJobs.Script.Metrics;
 using Microsoft.Azure.WebJobs.Script.Scale;
 using Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics.Extensions;
 using Microsoft.Azure.WebJobs.Script.Workers;
@@ -47,6 +48,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
         private readonly SemaphoreSlim _hostStartSemaphore = new SemaphoreSlim(1, 1);
         private readonly TaskCompletionSource<bool> _hostStartedSource = new TaskCompletionSource<bool>();
         private readonly Task _hostStarted;
+        private readonly IHostMetrics _hostMetrics;
         private IScriptEventManager _eventManager;
 
         private IHost _host;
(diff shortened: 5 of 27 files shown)


src/WebJobs.Script/WebJobs.Script.csproj
```diff
@@ -76,6 +76,7 @@
     <PackageReference Include="StyleCop.Analyzers" Version="1.2.0-beta.435">
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
+    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="8.0.0" />
```
jviau, line 79: Are we okay with upgrading diagnostic source? This has been a contentious point for a while as there is concern it will cause issues in customer apps.

src/WebJobs.Script/Metrics/HostMetricsCollector.cs
```diff
@@ -0,0 +1,113 @@
...
+
+            foreach (var tag in tags)
+            {
+                if (tag.Key == MetricConstants.FunctionExecutionResult)
+                {
+                    var executionResult = tag.Value?.ToString();
+
+                    if (executionResult == "Failed")
+                    {
+                        _metricsCache.AddOrUpdate(MetricConstants.FunctionFailureCount, 1, (key, oldValue) => (long)oldValue + 1);
```
jviau: If possible, I recommend using MDM here. Single metric "FunctionExecutions" and have a tag "success" with  true/false

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
```diff
@@ -150,6 +151,21 @@ public static void AddWebJobsScriptHost(this IServiceCollection services, IConfi
...
+                var options = s.GetService<IOptions<FlexConsumptionMetricsPublisherOptions>>();
+                var standbyOptions = s.GetService<IOptionsMonitor<StandbyOptions>>();
+                var logger = s.GetService<ILogger<FlexConsumptionMetricsPublisher>>();
+                var metricsCollector = s.GetService<IHostMetricsCollector>();
+                return new FlexConsumptionMetricsPublisher(environment, standbyOptions, options, logger, new FileSystem(), metricsCollector);
+            });
+
+            // Metrics
+            services.AddSingleton<IHostMetricsCollector, HostMetricsCollector>();
+            services.AddSingleton<HostMetricsMeter>();
```
jviau: To use `IMeterFactory` in DI, I think you need to call `.AddMetrics()`

https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.metricsserviceextensions.addmetrics?view=dotnet-plat-ext-8.0

src/WebJobs.Script/Metrics/MetricConstants.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Microsoft.Azure.WebJobs.Script.Metrics
+{
+    public class MetricConstants
+    {
+        // Meters
+        public const string Host = "Host";
+
+        // Metrics
+        public const string AppFailureCount = "host.app-failure-count";
```
jviau: Have we considered using OTel FaaS metric semantic conventions here? https://opentelemetry.io/docs/specs/semconv/faas/faas-metrics/

jviau: OTel advice is to use snake case for within a section -> `host.app_failure_count`. But they also recommend pluralization and leaving the `count` off of counter values, so `host.app_failures`.

But then they also advise to have the namespace considered universally, so all your metrics should start with someone distinct to your scenario. In this case `host` is a bit generic. We would need something like `azure.functions.host.app_failures` or `azfunc.host.app_failures`.

But _then_ there is already the `faas` semantic conventions. Does this just fit with that? (See other comment). If not, do we add our own namespace to `faas`? So `faas.azfunc.host.app_failures` (or something like that).

src/WebJobs.Script.WebHost/Diagnostics/FunctionInstanceLogger.cs
```diff
@@ -63,6 +68,8 @@ public Task AddAsync(FunctionInstanceLogEntry item, CancellationToken cancellati
 
         private void StartFunction(FunctionInstanceLogEntry item)
         {
+            _hostMetrics.IncrementActiveFunctionCount();
```
mathewc: There's an existing IFunctionActivityStatusProvider service implemented in the WebJobs SDK that tracks this information already. You should use that. It also accounts for retries - i.e. even if no invocations are active, we might be processing retries for an invocation.

mathewc: See here for how we use this https://github.com/Azure/azure-functions-host/blob/92d51f0fdf842af8dec3f1ea3275785ec89d3819/src/WebJobs.Script.WebHost/Controllers/HostController.cs#L220

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
```diff
@@ -90,11 +95,13 @@ internal async Task OnPublishMetrics()
         {
             try
             {
-                if (FunctionExecutionCount == 0 && FunctionExecutionTimeMS == 0 && !IsAlwaysReady)
-                {
-                    // no activity to report
-                    return;
-                }
+                // This is publishing metrics only when there are function executions, this would not work
```
mathewc: Consider having a method `_metricsCollector.HasMetrics()` that could be included in this check to avoid writing metrics when not necessary

src/WebJobs.Script/Host/ScriptHost.cs
```diff
@@ -433,6 +437,7 @@ private void PreInitialize()
 
                 if (string.IsNullOrEmpty(extensionVersion))
                 {
+                    _hostMetrics.AppFailure();
```
mathewc: Seems like we should handle failures centrally in the host init exception handler. Looks like you're doing that below so this line isn't needed?

mathewc: WebJobsScriptHostService.UnsynchronizedStartHostAsync is the main host loop responsible for handling startup failures and retires. See the logic there - seems like you should be doing your metering there https://github.com/Azure/azure-functions-host/blob/92d51f0fdf842af8dec3f1ea3275785ec89d3819/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs#L401.

src/WebJobs.Script/Metrics/HostMetricsCollector.cs
```diff
@@ -0,0 +1,92 @@
...
+            long measurement,
+            ReadOnlySpan<KeyValuePair<string, object?>> tags,
+            object? state)
+        {
+            if (instrument == null)
+            {
+                throw new ArgumentNullException(nameof(instrument));
+            }
+
+            _logger.LogDebug("{name} recorded measurement {measurement}", instrument.Name, measurement);
```
mathewc: Do we need this log? Seems like this will produce unnecessary log noise.

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
```diff
@@ -203,6 +211,16 @@ private bool PrepareDirectoryForFile()
             return files.Count < _options.MaxFileCount;
         }
 
+        private HostMetrics PrepareHostMetrics()
```
mathewc: Might be better for MetricsCollector.GetHostMetrics to do all this logic internally and just return a HostMetrics instance. Might be GetHostMetricsOrNull and return null if no metrics to report.

src/WebJobs.Script/Metrics/HostMetricsCollector.cs
liliankasem: Nope, was just using it for testing. Will remove

src/WebJobs.Script/WebJobs.Script.csproj
liliankasem: Will try to remember why I did this, I think there were some failures without it. Will check and write back

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
liliankasem: Hmm I have been able to use the factory without .AddMetrics - I think it's added by default in .net 8?

src/WebJobs.Script/Host/ScriptHost.cs
liliankasem: Just to clarify, the suggestion is to remove all records of "AppFailure();" currently sprinkled around the host, and to just have one occurance of this in UnsynchronizedStartHostAsync

src/WebJobs.Script/Metrics/IHostMetricsCollector.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Azure.WebJobs.Script.Metrics
+{
+    public interface IHostMetricsCollector
```
mathewc: Usually a "collector" has Add methods that allow data to be pumped in. This interface seems more like a "provider"? Is this pattern you're using with IHostMetricsCollector and HostMetricsMeter an established pattern for Otel or something? I'm not familiar.

liliankasem: No real reason I picked collector, but thinking about this more, I agree provider makes more sense.

As for the HostMetricsMeter -> IHostMetricsCollector pattern: using a `Meter` with a `MeterListener` is the recommended .NET Metrics pattern, I didn't end up naming it HostMetricsListener since it was doing more than just listening for metrics, the class is collecting metrics over time and making them available to consumers. The pattern can be found here: https://learn.microsoft.com/en-us/dotnet/core/diagnostics/metrics-collection#create-a-custom-collection-tool-using-the-net-meterlistener-api.

src/WebJobs.Script.WebHost/Diagnostics/FunctionInstanceLogger.cs
liliankasem: @mathewc I tried this approach out and within the EndFunction() method, the function status returns `OutstandingInvocations=1` and at this stage I would expect it to be 0 when the function completes. Is there a better place for me to be incrementing and decrementing the `ActiveInvocationCount` than here?

src/WebJobs.Script/Metrics/HostMetricsCollector.cs
liliankasem: Removing for now as we focus on the P0 items, but will keep that in mind as we add new metircs

src/WebJobs.Script/WebJobs.Script.csproj
liliankasem: So without this we do not have access to `IMeterFactory` (.NET 8+) or `UpDownCounter` (.NET 7+)

https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.metrics.imeterfactory?view=net-8.0

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
liliankasem: > Starting in .NET 8, hosts will automatically register [IMeterFactory](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.metrics.imeterfactory) in the service container or you can manually register the type in any [IServiceCollection](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.iservicecollection) by calling [AddMetrics](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.metricsserviceextensions.addmetrics)

Given this branch moves us to .NET 8 I think we are okay to not do this manually.

src/WebJobs.Script.WebHost/Diagnostics/FunctionInstanceLogCollectorProvider.cs
```diff
@@ -13,22 +13,26 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
...
         private readonly IConfiguration _configuration;
         private readonly ILoggerFactory _loggerFactory;
 
-        public FunctionInstanceLogCollectorProvider(IFunctionMetadataManager metadataManager, IMetricsLogger metrics, IConfiguration configuration, ILoggerFactory loggerFactory)
+        public FunctionInstanceLogCollectorProvider(IScriptHostManager scriptHostManager, IFunctionMetadataManager metadataManager, IMetricsLogger metrics, HostMetricsMeter hostMetrics, IConfiguration configuration, ILoggerFactory loggerFactory)
         {
+            _scriptHostManager = scriptHostManager ?? throw new ArgumentNullException(nameof(scriptHostManager));
             _metadataManager = metadataManager ?? throw new ArgumentNullException(nameof(metadataManager));
             _metrics = metrics ?? throw new ArgumentNullException(nameof(metrics));
+            _hostMetrics = hostMetrics ?? throw new ArgumentNullException(nameof(_hostMetrics));
```
kshyju: Did you mean `hostMetrics` ?

src/WebJobs.Script.WebHost/Diagnostics/FunctionInstanceLogger.cs
```diff
@@ -125,6 +138,14 @@ private void EndFunction(FunctionInstanceLogEntry item)
...
 
             _metrics.EndEvent(invokeLatencyEvent);
+
+            // Currently this always returns 1 for outstanding functions even though the function is complete
+            // Utility.TryGetHostService(_scriptHostManager, out IFunctionActivityStatusProvider functionActivityStatusProvider);
+            // var functionActivityStatus = functionActivityStatusProvider.GetStatus();
+            // if (functionActivityStatus.OutstandingInvocations == 0 && functionActivityStatus.OutstandingRetries == 0)
+            // {
+            //     _hostMetrics.DecrementActiveInvocationCount();
+            // }
```
kshyju: Do we need to keep this commented out snippet?

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
```diff
@@ -150,6 +151,21 @@ public static void AddWebJobsScriptHost(this IServiceCollection services, IConfi
                 services.TryAddSingleton<IDiagnosticEventRepositoryFactory, DiagnosticEventRepositoryFactory>();
             }
 
+            // For testing: registering the publisher for all, not just linux
```
kshyju: Can we improve this approach? Is there a better way where we can register these to the container only for Linux and test scenarios? May be a factory can help?

src/WebJobs.Script/Metrics/HostMetricsMeter.cs
```diff
@@ -0,0 +1,66 @@
...
+            }
+
+            if (environment == null)
+            {
+                throw new ArgumentNullException(nameof(environment));
+            }
+
+            var region = environment.GetEnvironmentVariableOrDefault(EnvironmentSettingNames.RegionName, string.Empty);
+            var resourceId = string.Empty; // Is it possible to get this value?
+            var instanceId = environment.GetInstanceId(); // Is CONTAINER_NAME still available in flex?
```
kshyju: Can we get answers for this questions and update the comments as needed, before this gets checked in. This will help future readers.

src/WebJobs.Script/Metrics/HostMetricsMeter.cs
```diff
@@ -0,0 +1,66 @@
...
+                    { MetricConstants.CloudResourceId, resourceId },
+                    { MetricConstants.ServiceInstanceId, instanceId },
+                    { MetricConstants.ServiceName, appName },
+                    { MetricConstants.AzureFunctionsGroup, functionGroup }
+                }
+            };
+            var meter = meterFactory.Create(hostMeterOptions);
+
+            _appFailureCount = meter.CreateCounter<long>(MetricConstants.AppFailureCount, "integer", "Number of times the host has failed to start.");
+            _startedInvocationCount = meter.CreateCounter<long>(MetricConstants.StartedInvocationCount, "integer", "Number of function invocations that have started.");
```
kshyju: We pass `long` as T for CreateCounter. Should this unit param reflect that? Or should we use a more generic unit value like "numeric"?

src/WebJobs.Script/Metrics/HostMetricsProvider.cs
```diff
@@ -0,0 +1,121 @@
...
+using System.Diagnostics.Metrics;
+using System.Linq;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Script.Metrics
+{
+    public class HostMetricsProvider : IHostMetricsProvider, IDisposable
+    {
+        private readonly MeterListener _meterListener;
+        private readonly ILogger<HostMetricsProvider> _logger;
```
kshyju: `_logger` seems to be unused.

src/WebJobs.Script/Metrics/IHostMetricsProvider.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Azure.WebJobs.Script.Metrics
+{
+    public interface IHostMetricsProvider
```
kshyju, line 15: It would be nice to have summary on these public types/methods.

kshyju (commented):
@fabiocav Do we want to merge changes directly to oop-host branch?

src/WebJobs.Script.WebHost/Diagnostics/FunctionInstanceLogger.cs
liliankasem: Nope! I'm waiting for feedback from @mathewc on this and will make a change depending on that. There won't be any code commented out when I merge.

src/WebJobs.Script/Metrics/HostMetricsMeter.cs
liliankasem: Yup, not planning to merge with these comments. I'm trying to get answers from Pragna and Bala and will update the PR soon

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
liliankasem: Not a bad idea, I don't know that I want to handle that at the moment but if I have time I will make a change for for this

liliankasem: to be clear: I will be deleting this code before I merge - this is just what I am doing to test on my local machine right now.

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
```diff
@@ -106,7 +111,8 @@ internal async Task OnPublishMetrics()
                         TotalTimeMS = (long)_intervalStopwatch.GetElapsedTime().TotalMilliseconds,
                         ExecutionCount = FunctionExecutionCount,
                         ExecutionTimeMS = FunctionExecutionTimeMS,
-                        IsAlwaysReady = IsAlwaysReady
+                        IsAlwaysReady = IsAlwaysReady,
+                        HostMetrics = _metricsProvider.GetHostMetricsOrNull()
```
mathewc: I'm confused about the structure here. All the existing flat record metrics are already "host metrics" - so why do we need this new "host metrics" nesting?

src/WebJobs.Script/Metrics/HostMetrics.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        /// <summary>
+        /// Gets or sets the total number of in-progress function invocations.
+        /// </summary>
+        public long ActiveInvocationCount { get; set; }
+
+        /// <summary>
+        /// Gets or sets the total number of function invocations that have started.
+        /// </summary>
+        public long StartedInvocationCount { get; set; }
```
mathewc: The existing ExecutionCount metric we're already logging is a count of the completed invocations during an interval. Wondering if we need this separate count.

liliankasem: Bala would like to capture all executions even if they don't complete so I thought it better to create a new metric for it. I believe this is to be a running total as well

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
liliankasem: I thought it would be cleaner to separate the ones we need for scaling vs the general host metrics

src/WebJobs.Script/WebJobs.Script.csproj
liliankasem: @kshyju @fabiocav - any concerns on this depencency?

src/WebJobs.Script/Metrics/MetricConstants.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace Microsoft.Azure.WebJobs.Script.Metrics
+{
+    public class MetricConstants
+    {
+        // Cloud Information
+        public const string CloudPlatformName = "azure_functions";
+
+        // Meters
+        public const string AzureFunctions = "azure.functions";
```
jviau: Meter names are intended to be fully quality namespaces.

Something like `Microsoft.Azure.WebJobs.Script.Host.Internal` would be preferred. Note, I added `.Internal` here because I want to express customers/extensions are not to take a dependency on this meter, and we reserve the right to perform breaking changes on it with no notice.

We can have official customer-facing ones as part of the broader OTel work.

src/WebJobs.Script/WebJobs.Script.csproj
jviau: Meant to reply - there is no issue with this dependency. I missed that this is for out-of-proc worker only. So we are already on .net8 here

src/WebJobs.Script/Metrics/HostMetricsMeter.cs
```diff
@@ -0,0 +1,67 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Diagnostics.Metrics;
+
+namespace Microsoft.Azure.WebJobs.Script.Metrics
+{
+    public class HostMetricsMeter
```
jviau: nit:

1. extract interface (so it can be mocked for unit tests) _or_ make all methods `virtual`.
2. `Meter` sounds superfluous to me. What do you think of just `HostMetrics` (or `IHostMetrics` if you extract an interface)? Would need to rename `HostMetrics` payload to something else - maybe `HostMetricsPayload`? That type will go away eventually so not concerned about an ugly name.

src/WebJobs.Script/Metrics/MetricConstants.cs
jviau:
```suggestion
        public const string AzureFunctionsMeter = "azure.functions";
```

but see my other comment for moving this to `HostMetrics`.

src/WebJobs.Script/Metrics/MetricConstants.cs
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Script.Metrics
+{
+    public class MetricConstants
```
jviau: nit: these can be used for more than just metrics. Recommend the following:

1. Move attributes (tags) to another file `TelemetryAttributes`.
2. For metrics meter and instrument names, we can either have this file (`MetricConstants`), or we can have the metric implementation itself have these constants (so move them into `HostMetricsMeter`, which would be renamed `HostMetrics` by my other suggestion).
3. With them moved to `HostMetrics`, can rename `AzureFunctionsMeter` to just `MeterName`.

src/WebJobs.Script/Metrics/HostMetrics.cs
```diff
@@ -0,0 +1,41 @@
...
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs.Script.Metrics
+{
+    public class HostMetrics
+    {
+        /// <summary>
+        /// Gets or sets the instance Id.
+        /// </summary>
+        public string InstanceId { get; set; }
```
jviau: Will the platforms metric pipeline set this already?

liliankasem: I checked with Anjuna and apparently not, we will need to provide this value

src/WebJobs.Script/Metrics/HostMetricsMeter.cs
liliankasem: I was also thinking ScaleMetrics - also don't care too much as the plan will be to remove this

jviau (approved):
LGTM. Also I do like your idea of naming the payload `ScaleMetrics`. Up to you if you want to go with that.

src/WebJobs.Script.WebHost/Diagnostics/FunctionInstanceLogger.cs
mathewc: I think the difficulty is that you're attempting to maintain a separate incremented count for this. Instead, you'd just call IFuncitonActivityStatusProvider at the point in time when you need to get the current count and log it. Not each time a function ends. The platform is already maintaining the count [here](https://github.com/Azure/azure-webjobs-sdk/blob/1f5c638d08a15360e54fd620d34276b10f186da6/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs#L160).

src/WebJobs.Script.WebHost/Diagnostics/FunctionInstanceLogger.cs
mathewc: Will mark this unresolved then until it is resolved

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
mathewc: I don't see why - these are all just host level metrics, to be used by consumers for whatever purposes they may need. Logically they all seem to be peers to the metrics we're already logging, so I say we just inline them.

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -398,6 +401,7 @@ private async Task UnsynchronizedStartHostAsync(ScriptHostStartupOperation activ
                     logger.ErrorOccurredInactive(activeOperation.Id, exc);
                 }
 
+                _hostMetrics.AppFailure();
```
mathewc, line 404: When in a crash loop, attemptCount will keep increasing, but will be reset back to zero after the first successful startup. Won't your new metric just monotonically increase? Is the idea that it's up to the consumer of the metric to do diffs at different points in time to determine when the app is healthy? How long to wait between samples - e.g. we have an exponential backoff on retries, so they'll eventually slow down to once every 2 minutes.

src/WebJobs.Script/Host/ScriptHost.cs
mathewc: Is the goal of this new metric to capture any time a host level startup error occurs? I think those can happen in lots of places. However, I think the catch handler in UnsynchronizedStartHostAsync will end up catching all your cases so I'm not sure if you need to instrument all over (might lead to duplicate counting even). It's like a big try/catch around all these startup operations.

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
```diff
@@ -265,6 +269,8 @@ private void ValidateBundleRequirements(ExtensionBundleDetails bundleDetails)
                 if (bundleVersion < minimumVersion)
                 {
                     _logger.MinimumBundleVersionNotSatisfied(bundleDetails.Id, bundleDetails.Version, requirement.MinimumVersion);
+                    // Add tag? Or use health check instead of metric?
+                    _hostMetric.AppFailure();
```
mathewc: I'm thinking a failure here will already be caught and counted by the main startup loop you've already instrumented. Same for the other instances in this file.

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -437,7 +437,12 @@ public static string GetInstanceId(this IEnvironment environment)
         {
             if (environment.IsAnyLinuxConsumption())
             {
-                return environment.GetEnvironmentVariableOrDefault(ContainerName, string.Empty);
```
mathewc, line 440: Can you explain this - seems unrelated to your changes.

src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
```diff
@@ -76,6 +76,8 @@ public static class EnvironmentSettingNames
         public const string FunctionsAlwaysReadyInstance = "FUNCTIONS_ALWAYSREADY_INSTANCE";
         public const string FunctionsTimeZone = "TZ";
         public const string FunctionsWebsiteTimeZone = "WEBSITE_TIME_ZONE";
+        public const string FunctionsTargetGroup = "FUNCTIONS_TARGET_GROUP";
+        public const string WebsiteArmResourceId = "WEBSITE_ARM_RESOURCE_ID";
```
mathewc, line 80: Please explain this new ARM ID thing - who is setting it on the app?

src/WebJobs.Script/Metrics/HostMetrics.cs
```diff
@@ -0,0 +1,73 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Diagnostics.Metrics;
+
+namespace Microsoft.Azure.WebJobs.Script.Metrics
+{
+    public class HostMetrics : IHostMetrics
```
mathewc, line 11: please add xml comments for new types describing them

src/WebJobs.Script/Metrics/IHostMetrics.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    /// <summary>
+    /// Defines the host metric meter methods.
+    /// </summary>
+    public interface IHostMetrics
+    {
+        /// <summary>
+        /// Increments the app failure count.
+        /// </summary>
+        public void AppFailure();
```
mathewc, line 18: Wondering if the name of this isn't too general - are we trying to capture only host level fatal startup errors? E.g. a failed function invocation is a failure but wouldn't be counted here.

At a minimum, I think you should update this method to detail exactly what an "AppFailure" is. Or perhaps we need to rename it to something more specific like HostStartupFailureCount.

src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
liliankasem: Arjun is working on injecting this during specialization; this work is being tracked internally

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
liliankasem: In flex, we want to replace CONTAINER_NAME with WEBSITE_POD_NAME. Bala's suggestion for getting the instance ID is to try and get WEBSITE_POD_NAME first and if not available fall back to CONTAINER_NAME. Both should be available right from container creation time and doesn't need specialization.

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
liliankasem: Sure. In that case can we move the `internal class Metrics` class to Core? HostMetricsProvider.GetHostMetricsOrNull would need to change in how we provide these values instead of creating HostMetricsPayload like we do now

src/WebJobs.Script/Metrics/IHostMetrics.cs
liliankasem: At the moment this is for any permenant app failures where customer action is required to fix the issue i.e unable to download app content due to a 404, or the dotnet worker SDK not supported or anytime we encounter a `HostInitializationException` (this is all outlined in the design doc). A future evolution Jacob and I have disuccused is adding a dimension to this for `severity` where critical is for all permenant app failures and error for other types of issues, but that is being figured out by the larger host metrics picture.

If you feel the use cases I mentioned above can be encalsulated by `HostStartupFailureCount` then we can rename it to that, otherwise I can just add more info in the summary

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
liliankasem: Yup, I thought I removed these instances already - thanks for catching

src/WebJobs.Script.WebHost/Diagnostics/FunctionInstanceLogger.cs
liliankasem: Fixed by moving this to the host metric provider and not using .net metrics to capture this value.

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
liliankasem: moved host metrics inline

src/WebJobs.Script/Host/ScriptHost.cs
liliankasem: Reduced all use of AppFailure() to just UnsynchronizedStartHostAsync.

> Is the goal of this new metric to capture any time a host level startup error occurs?

I think I answer this question here: https://github.com/Azure/azure-functions-host/pull/9730#discussion_r1471959622

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
liliankasem: Hmm that's a good point. I think this would be okay? The consumer would be getting a time series of metrics every 30 seconds or so, and the metrics provider clears the cache every time we write to the metric file via the flex publisher. The consumer would be able to see if this goes back to 0 for a considerate amount of time. How often do we retry and do we ever stop retrying?

On the consumer side (this is the main use case), if this value is > 1 the instance/pod is signaled as unhealthy and there is action to recycle. If every instance in a site has this value > 1 - then we scale down to 1 instance until we see this value at 0 for a couple minutes. 

I'm open to suggestions here on how to best set this up to enable the use case I mention

src/WebJobs.Script/Metrics/IHostMetrics.cs
liliankasem: Discussed with Fabio and we're going to leave this as is for now. We need to finalize these changes to unblock other components. We can always refactor this once we have our initial iteration working.

mathewc (approved):

PR #9743, opened 2023-12-14
Improving worker shutdown test to prevent regressions

brettsam:
Test-only change.

In https://github.com/Azure/azure-functions-host/pull/9738, I fixed a race condition on worker shutdown. However, the test I wrote was very hard to use as a repro and thus was not a great regression test. I've since found a better way to time the multiple threads. 

~Currently submitting this back when the bug existed to ensure it would have caught the regression.~
Edit: This is now ready.

### Pull request checklist


brettsam:
For one of the steps here I reverted the fix and let the test run to ensure we would have caught it.

Proof that the test failed in CI before the fix was applied: https://azfunc.visualstudio.com/Azure%20Functions/_build/results?buildId=158816&view=ms.vss-test-web.build-test-results-tab&runId=1371853&resultId=101182&paneView=debug

liliankasem (approved):

soninaren (approved):

PR #9747, opened 2023-12-19
Logging Settings for host split

### LINKED ISSUE #9721: Host Split: explicitly log settings and root path
We'll be rolling out an in-proc/out-of-proc host split in the near future. As we test this on individual stamps during the rollout, we'll need a concise log emitted from the host to let us know if anything appears wrong. This log should contain:
- `FUNCTIONS_WORKER_RUNTIME` value
- `FUNCTIONS_EXTENSION_VERSION` value
- `ASPNETCORE_IIS_PHYSICAL_PATH`  value
- The contents of the `ScmHostingConfiguration.txt` file (which should already be available in `IConfiguration`.

This log should also have an explicit `EventName` so it's easy to query.

aishwaryabh:

resolves #9721 

This PR logs the properties that will be useful to have for the host split project. Here is an example logging:
image

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
@@ -182,6 +182,12 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics.Extensions
                 new EventId(529, nameof(HostStateChanged)),
                 "Host state changed from {previousState} to {newState}.");
 
+        private static readonly Action<ILogger, string, string, string, string, Exception> _logSettings =
+            LoggerMessage.Define<string, string, string, string>(
+                LogLevel.Information,
+                new EventId(530, nameof(LogSettings)),
+                Properties.Resources.LogSettings);
+
         public static void HostStateChanged(this ILogger logger, ScriptHostState previousHostState, ScriptHostState newHostState)
         {
             var newState = newHostState.ToString();
@@ -322,6 +328,11 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics.Extensions
         public static void ScriptHostServiceRestartCanceledByRuntime(this ILogger logger)
         {
             _scriptHostServiceRestartCanceledByRuntime(logger, null);
+        }
+
+        public static void LogSettings(this ILogger logger, string functionWorkerRuntime, string functionExtensionVersion, string currentDirectory, string inStandbyMode)
+        {
+            _logSettings(logger, functionWorkerRuntime, functionExtensionVersion, currentDirectory, inStandbyMode, null);
         }
     }
 }
--- a/src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
@@ -182,11 +182,11 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics.Extensions
                 new EventId(529, nameof(HostStateChanged)),
                 "Host state changed from {previousState} to {newState}.");
 
-        private static readonly Action<ILogger, string, string, string, string, Exception> _logSettings =
-            LoggerMessage.Define<string, string, string, string>(
-                LogLevel.Information,
-                new EventId(530, nameof(LogSettings)),
-                Properties.Resources.LogSettings);
+        private static readonly Action<ILogger, string, string, string, bool, Exception> _logHostInitializationSettings =
+            LoggerMessage.Define<string, string, string, bool>(
+                LogLevel.Debug,
+                new EventId(530, nameof(LogHostInitializationSettings)),
+                Properties.Resources.LogHostInitializationSettings);
 
         public static void HostStateChanged(this ILogger logger, ScriptHostState previousHostState, ScriptHostState newHostState)
         {
@@ -330,9 +330,9 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics.Extensions
             _scriptHostServiceRestartCanceledByRuntime(logger, null);
         }
 
-        public static void LogSettings(this ILogger logger, string functionWorkerRuntime, string functionExtensionVersion, string currentDirectory, string inStandbyMode)
+        public static void LogHostInitializationSettings(this ILogger logger, string functionWorkerRuntime, string functionExtensionVersion, string currentDirectory, bool inStandbyMode)
         {
-            _logSettings(logger, functionWorkerRuntime, functionExtensionVersion, currentDirectory, inStandbyMode, null);
+            _logHostInitializationSettings(logger, functionWorkerRuntime, functionExtensionVersion, currentDirectory, inStandbyMode, null);
         }
     }
 }
--- a/src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
@@ -184,7 +184,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics.Extensions
 
         private static readonly Action<ILogger, string, Exception> _logHostInitializationSettings =
             LoggerMessage.Define<string>(
-                LogLevel.Debug,
+                LogLevel.Information,
                 new EventId(530, nameof(LogHostInitializationSettings)),
                 "{hostInitializationSettings}");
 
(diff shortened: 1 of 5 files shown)


src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
```diff
@@ -182,6 +182,12 @@ public static class ScriptHostServiceLoggerExtension
                 new EventId(529, nameof(HostStateChanged)),
                 "Host state changed from {previousState} to {newState}.");
 
+        private static readonly Action<ILogger, string, string, string, string, Exception> _logSettings =
+            LoggerMessage.Define<string, string, string, string>(
+                LogLevel.Information,
```
brettsam: Let's log this as Debug

src/WebJobs.Script.WebHost/Properties/Resources.resx
```diff
@@ -283,6 +283,14 @@
   <data name="HostSpecializationTrace" xml:space="preserve">
     <value>Starting host specialization</value>
   </data>
+  <data name="LogSettings" xml:space="preserve">
+    <value>Settings: {{
```
brettsam: Let's call these "Host initialization settings", maybe? Rather than just "Settings"...

src/WebJobs.Script.WebHost/Properties/Resources.resx
```diff
@@ -283,6 +283,14 @@
...
+  "functionsWorkerRuntime": "{functionWorkerRuntime}",
+  "functionsExtensionVersion": "{functionsExtensionVersion}",
+  "currentDirectory": "{currentDirectory}",
+  "inStandbyMode": "{inStandbyMode}"
```
brettsam: since `inStandbyMode` is a bool... no need for quotes around it

src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
```diff
@@ -323,5 +329,10 @@ public static void ScriptHostServiceRestartCanceledByRuntime(this ILogger logger
         {
             _scriptHostServiceRestartCanceledByRuntime(logger, null);
         }
+
+        public static void LogSettings(this ILogger logger, string functionWorkerRuntime, string functionExtensionVersion, string currentDirectory, string inStandbyMode)
```
brettsam: make `inStandbyMode` a `bool` here.

brettsam: also rename to `LogHostInitializationSettings` (assuming you like that name) -- and that'll update the event name as well.

src/WebJobs.Script.WebHost/Properties/Resources.resx
aishwaryabh: done!

src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
aishwaryabh: changed!

src/WebJobs.Script.WebHost/Properties/Resources.resx
```diff
@@ -283,6 +283,14 @@
   <data name="HostSpecializationTrace" xml:space="preserve">
     <value>Starting host specialization</value>
   </data>
+  <data name="LogHostInitializationSettings" xml:space="preserve">
+    <value>Settings: {{
+  "functionsWorkerRuntime": "{functionWorkerRuntime}",
```
kshyju: Does it make a difference if we were to not define the schema here? What if we just log a string (The string can be a JSON produced from the caller code).

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -685,6 +685,14 @@ private void LogInitialization(IHost host, bool isOffline, int attemptCount, int
                 // Reading the string from resources to make sure resource loading code path is warmed up during placeholder as well.
                 logger.InStandByMode(operationId);
             }
+
+            // Log settings
+            var functionWorkerRuntime = _environment.GetEnvironmentVariable(FunctionWorkerRuntime);
+            var functionExtensionVersion = _environment.GetEnvironmentVariable(FunctionsExtensionVersion);
+            var currentDirectory = Directory.GetCurrentDirectory();
```
kshyju, line 692: Why do we want to log the current directory? Is it to get a sense of what site extension directory was used to start host? 

I observed that the env variable `ASPNETCORE_IIS_PHYSICAL_PATH` the path to the extension folder (example value: `C:\Program Files (x86)\SiteExtensions\Functions\4.28.3\64bit\`). But this is not available in Linux.

src/WebJobs.Script.WebHost/Properties/Resources.resx
aishwaryabh: I believe @brettsam said it was more efficient to just plug in JSON values like so instead of creating a JSON value ourselves

brettsam: Yeah, in my mind, this would be more efficient during specialization... although honestly we're not talking about much and it's only logged a single time. I'll sync with @aishwaryabh and maybe we can simplify this to touch less files...

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
brettsam: You kinda answered the question :-)

Yeah -- we want to know what SiteExtension is running the current process. We surprisingly don't have that information anywhere today unless you join across backend tables.

The goal with this change is to put all the details we need to generate a "host split report" in kusto to see whether we're missing anything when we start to roll this out. Having this all in a single log message makes that a ton easier.

src/WebJobs.Script.WebHost/Properties/Resources.resx
aishwaryabh: Just changed into a string!

src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
```diff
@@ -332,7 +332,13 @@ public static void ScriptHostServiceRestartCanceledByRuntime(this ILogger logger
 
         public static void LogHostInitializationSettings(this ILogger logger, string functionWorkerRuntime, string functionExtensionVersion, string currentDirectory, bool inStandbyMode)
         {
-            _logHostInitializationSettings(logger, functionWorkerRuntime, functionExtensionVersion, currentDirectory, inStandbyMode, null);
+            var hostInitializationSettings = $@"Settings: {{
+    ""functionsWorkerRuntime"": ""{functionWorkerRuntime}"",
```
kshyju: Question: I see we are not logging a JSON string in other places, Ex: https://github.com/Azure/azure-functions-host/blob/aae3d338434532ea4998049e4e9ee69a6fe039b7/src/WebJobs.Script/Diagnostics/Extensions/ScriptHostLoggerExtension.cs#L142C59-L142C59

Did we want to log this as a string which contains a JSON string?

src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
```diff
@@ -332,7 +332,13 @@ public static void ScriptHostServiceRestartCanceledByRuntime(this ILogger logger
 
         public static void LogHostInitializationSettings(this ILogger logger, string functionWorkerRuntime, string functionExtensionVersion, string currentDirectory, bool inStandbyMode)
         {
-            _logHostInitializationSettings(logger, functionWorkerRuntime, functionExtensionVersion, currentDirectory, inStandbyMode, null);
+            var hostInitializationSettings = $@"Settings: {{
```
kshyju: Instead of "Settings", should we use a more specific term here?

src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
aishwaryabh: Are we able to parse that in kusto? I think we decided to log the changes as a JSON string so that it would be easy to parse

aishwaryabh: Talked offline and we decided to format the string to be a proper JSON string:
{ "HostInitializationSettings" : { ..} }

src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
aishwaryabh: Updated!

kshyju (approved):
PR description can be updated to reflect the new keyname (for currentDirectory)

brettsam (approved):

PR #9785, opened 2024-01-10
[Flex][Scaling] Allow for running only a specified group of functions

### LINKED ISSUE #9734: [Flex][Scaling] Selectively run functions if a target group is set
This issue tracks the work to update the host to:

- Detect if a function group target has been configured before/part of specialization
- Run only functions that are part of the group
- All other functions are to be indexed, but will not be triggered automatically

jviau:

resolves #9734

### Pull request checklist

### Additional information

This PR adds support for specifying and running a select function group in flex consumption only. This is accomplished by checking for the env variable `FUNCTIONS_TARGET_GROUP`. If it is available, we will enable _only_ functions that match that group. This is accomplished by doing the inverse: functions _not_ in the group have their listener suppressed by inserting a no-op listener ahead of it. This allows for the function to appear present in all other systems: not disabled, part of sync triggers, can be manually triggered. But its listener will never be started and thus the function is never automatically triggered.

### DIFF
--- a/src/WebJobs.Script/Description/FunctionGroupListenerDecorator.cs
+++ b/src/WebJobs.Script/Description/FunctionGroupListenerDecorator.cs
@@ -1,0 +1,78 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host.Listeners;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Script.Description
+{
+    internal class FunctionGroupListenerDecorator : IListenerDecorator
+    {
+        private static readonly NoOpListener _noOpListener = new();
+        private readonly IFunctionMetadataManager _metadata;
+        private readonly IEnvironment _environment; // TODO: replace options pattern
+        private readonly ILogger _logger;
+
+        public FunctionGroupListenerDecorator(
+            IFunctionMetadataManager metadata,
+            IEnvironment environment,
+            ILogger<FunctionGroupListenerDecorator> logger)
+        {
+            _metadata = metadata ?? throw new ArgumentNullException(nameof(metadata));
+            _environment = environment ?? throw new ArgumentNullException(nameof(environment));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        }
+
+        public IListener Decorate(ListenerDecoratorContext context)
+        {
+            if (context is null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            if (!_environment.TryGetFunctionsTargetGroup(out string targetGroup))
+            {
+                // If no group configured, short-circuit.
+                return context.Listener;
+            }
+
+            _logger.LogInformation("Function group target is {targetGroup}", targetGroup);
+            string functionName = context.FunctionDefinition.Descriptor.ShortName;
+            if (!_metadata.TryGetFunctionMetadata(functionName, out FunctionMetadata m))
+            {
+                _logger.LogWarning("Unable to find function metadata for function {functionName}", functionName);
+                return context.Listener;
+            }
+
+            string group = m.GetFunctionGroup() ?? functionName;
+            if (string.Equals(targetGroup, group, StringComparison.OrdinalIgnoreCase))
+            {
+                _logger.LogDebug("Enabling function {functionName}", functionName);
+                return context.Listener;
+            }
+
+            // A target function group is configured and this function is not part of it.
+            // By giving a no-op listener, we will prevent it from triggering without 'disabling' it.
+            _logger.LogDebug("Disabling function {functionName}", functionName);
+            return _noOpListener;
+        }
+
+        private class NoOpListener : IListener
+        {
+            public void Cancel()
+            {
+            }
+
+            public void Dispose()
+            {
+            }
+
+            public Task StartAsync(CancellationToken cancellationToken) => Task.CompletedTask;
+
+            public Task StopAsync(CancellationToken cancellationToken) => Task.CompletedTask;
+        }
+    }
+}
--- a/src/WebJobs.Script/Description/FunctionGroupListenerDecorator.cs
+++ b/src/WebJobs.Script/Description/FunctionGroupListenerDecorator.cs
@@ -41,13 +41,13 @@ namespace Microsoft.Azure.WebJobs.Script.Description
 
             _logger.LogInformation("Function group target is {targetGroup}", targetGroup);
             string functionName = context.FunctionDefinition.Descriptor.ShortName;
-            if (!_metadata.TryGetFunctionMetadata(functionName, out FunctionMetadata m))
+            if (!_metadata.TryGetFunctionMetadata(functionName, out FunctionMetadata functionMetadata))
             {
                 _logger.LogWarning("Unable to find function metadata for function {functionName}", functionName);
                 return context.Listener;
             }
 
-            string group = m.GetFunctionGroup() ?? functionName;
+            string group = functionMetadata.GetFunctionGroup() ?? functionName;
             if (string.Equals(targetGroup, group, StringComparison.OrdinalIgnoreCase))
             {
                 _logger.LogDebug("Enabling function {functionName}", functionName);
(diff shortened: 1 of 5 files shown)


liliankasem:
I thought flex changes were going into the oop-host feature branch? or is this a generic change that will be available for everything and if so I think we need to bring these changes into the oop feature branch for flex

jviau:
Oh was not aware of that. This is innocent and can go in dev, but I'll check if this should be that branch you mentioned.

jviau:
~@liliankasem I have updated this PR to target feature/oop-host~

Talked to Fabio, I think we actually want to continue with targeting `dev` for now and merging `dev` to the feature branch. There are no PR checks for this feature branch yet.

src/WebJobs.Script/Description/FunctionGroupListenerDecorator.cs
```diff
@@ -0,0 +1,79 @@
...
+            string group = m.GetFunctionGroup() ?? functionName;
+            if (string.Equals(targetGroup, group, StringComparison.OrdinalIgnoreCase))
+            {
+                _logger.LogDebug("Enabling function {functionName}", functionName);
+                return context.Listener;
+            }
+
+            // A target function group is configured and this function is not part of it.
+            // By giving a no-op listener, we will prevent it from triggering without 'disabling' it.
+            _logger.LogDebug("Disabling function {functionName}", functionName);
```
liliankasem: A little worried about the saying "disabled" then it's not technically disabled in the functions sense?

src/WebJobs.Script/Description/FunctionGroupListenerDecorator.cs
```diff
@@ -0,0 +1,79 @@
...
+
+            if (!_environment.TryGetFunctionsTargetGroup(out string targetGroup))
+            {
+                // If no group configured, short-circuit.
+                return context.Listener;
+            }
+
+            _logger.LogInformation("Function group target is {targetGroup}", targetGroup);
+            string functionName = context.FunctionDefinition.Descriptor.ShortName;
+            if (!_metadata.TryGetFunctionMetadata(functionName, out FunctionMetadata m))
```
liliankasem:
```suggestion
            if (!_metadata.TryGetFunctionMetadata(functionName, out FunctionMetadata functionMetadata))
```

src/WebJobs.Script/Description/FunctionGroupListenerDecorator.cs
```diff
@@ -0,0 +1,79 @@
...
+            {
+                _logger.LogWarning("Unable to find function metadata for function {functionName}", functionName);
+                return context.Listener;
+            }
+
+            string group = m.GetFunctionGroup() ?? functionName;
+            if (string.Equals(targetGroup, group, StringComparison.OrdinalIgnoreCase))
```
liliankasem, line 52: Just for my understanding, this is saying that if a given function is part of a function group (specifically the function group of this host instance), it is enabled for this instance of the host, otherwise disabled. How do we determine which host instance is assigned which functions group? Is that where the functions scaling and DataRole comes into play?

liliankasem (approved):

liliankasem:
> ~@liliankasem I have updated this PR to target feature/oop-host~
> 
> Talked to Fabio, I think we actually want to continue with targeting `dev` for now and merging `dev` to the feature branch. There are no PR checks for this feature branch yet.

Sounds good. RE: pr checks, yeah... I need those to get my PR merged so will have to put some effort this week into reenabling these

src/WebJobs.Script/Description/FunctionGroupListenerDecorator.cs
jviau: Good point, will improve the wording here.

src/WebJobs.Script/Description/FunctionGroupListenerDecorator.cs
jviau: Yes, datarole should be setting the `FUNCTIONS_TARGET_GROUP` env variable when specializing this instance.

src/WebJobs.Script/Description/FunctionGroupListenerDecorator.cs
```diff
@@ -0,0 +1,79 @@
...
+            {
+                _logger.LogDebug("Enabling function {functionName}", functionName);
+                return context.Listener;
+            }
+
+            // A target function group is configured and this function is not part of it.
+            // By giving a no-op listener, we will prevent it from triggering without 'disabling' it.
+            _logger.LogDebug("Function {functionName} is not part of group {functionGroup}. Listener will not be enabled.", functionName, targetGroup);
+            context.Listener?.Dispose(); // this will not be used, lets dispose it now.
```
mathewc, line 60: Not sure you should dispose this here - the underlying host owns the listeners and will attempt to dispose them itself on shutdown here https://github.com/Azure/azure-webjobs-sdk/blob/1e731b26b33e4cb302e5a30447f2655e6e291b0f/src/Microsoft.Azure.WebJobs.Host/Listeners/CompositeListener.cs#L72

src/WebJobs.Script/Description/FunctionGroupListenerDecorator.cs
```diff
@@ -0,0 +1,79 @@
...
+using Microsoft.Azure.WebJobs.Host.Listeners;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Script.Description
+{
+    internal class FunctionGroupListenerDecorator : IListenerDecorator
+    {
+        private static readonly NoOpListener _noOpListener = new();
+        private readonly IFunctionMetadataManager _metadata;
+        private readonly IEnvironment _environment; // TODO: replace options pattern
```
mathewc, line 15: Address this TODO?

src/WebJobs.Script/Description/FunctionGroupListenerDecorator.cs
```diff
@@ -0,0 +1,79 @@
...
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host.Listeners;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Script.Description
+{
+    internal class FunctionGroupListenerDecorator : IListenerDecorator
+    {
+        private static readonly NoOpListener _noOpListener = new();
+        private readonly IFunctionMetadataManager _metadata;
```
mathewc: nit: _metadataManager

src/WebJobs.Script/Description/FunctionGroupListenerDecorator.cs
mathewc: Need to think about this - the platform [here](https://github.com/Azure/azure-webjobs-sdk/blob/1e731b26b33e4cb302e5a30447f2655e6e291b0f/src/Microsoft.Azure.WebJobs.Host/Listeners/HostListenerFactory.cs#L66) is actually holding onto a reference to this listener for scale monitoring purposes. I suspect that if you test this code and call the host scale status APIs you'll get a null ref

src/WebJobs.Script/Description/FunctionGroupListenerDecorator.cs
```diff
@@ -0,0 +1,79 @@
...
+            IFunctionMetadataManager metadata,
+            IEnvironment environment,
+            ILogger<FunctionGroupListenerDecorator> logger)
+        {
+            _metadata = metadata ?? throw new ArgumentNullException(nameof(metadata));
+            _environment = environment ?? throw new ArgumentNullException(nameof(environment));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        }
+
+        public IListener Decorate(ListenerDecoratorContext context)
```
mathewc, line 28: Should add some unit tests to test this logic

src/WebJobs.Script/Description/FunctionGroupListenerDecorator.cs
jviau: I want to approach that in another story. We have the larger question of separating user and system properties, preventing them from overriding one another. I'd like for us to come up with a holistic answer to that before solving that here.

src/WebJobs.Script/Description/FunctionGroupListenerDecorator.cs
jviau: Ah I see. Do we even want that code to run in this scenario?

jviau: Updated so that we don't dispose until later.

jviau: With the dispose change the scale endpoint still works (even when the listener was never started). However, will need to discuss if this endpoint is even used in per-func scaling. It does not appear to be designed for it, as there is only a single current worker count provided: _which group_ does that worker count represent?

mathewc: My main point is that the host believes it owns the listener instance and will dispose it. So I don't think you should dispose it.

jviau: When looking at other decorator listeners, they hold a reference of the decorated instance and dispose it on their own disposal. I switched to that pattern to be in line with them.

mathewc (approved):

PR #9797, opened 2024-01-12
Update IdentityDependencyVersion 

liliankasem:
Cherry-pick of #9793


fabiocav:
Let's not cherry-pick commits, but instead, bring this branch current. 

Will work with folks on this today

PR #9799, opened 2024-01-12
[pack] adding warning if FUNCTIONS_WORKER_RUNTIME is missing

### LINKED ISSUE #9801: New Diagnostic Event for cases where FUNCTIONS_WORKER_RUNTIME is not set.
While there exists some fallback logic today, this is not the most performant and prevents some features from being leveraged. In some cases, it can prevent the app from starting. We should flag this earlier than runtime fault through a diagnostic event.

Arguably, this should also be entered into deployment logs. It might make sense to split that off as a separate issue.

brettsam:
resolves #9801 

Begin warning if `FUNCTIONS_WORKER_RUNTIME` is null or empty. Changes coming in the future for both performance and reliability require that we have this set so we can guarantee we run the correct host version. Even though it's been in templates for quite a while, we still have some sites without it. There is a plan to gradually get the number of null `FUNCTIONS_WORKER_RUNTIME` values down to zero, and then enforcing that it is set. This is the first step towards that.

Documentation PR is here: https://github.com/MicrosoftDocs/azure-docs-pr/pull/264142

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script/Diagnostics/DiagnosticEventConstants.cs
+++ b/src/WebJobs.Script/Diagnostics/DiagnosticEventConstants.cs
@@ -25,5 +25,8 @@ namespace Microsoft.Azure.WebJobs.Script
 
         public const string LinuxConsumptionTimeZoneErrorCode = "AZFD0010";
         public const string LinuxConsumptionTimeZoneErrorHelpLink = "https://go.microsoft.com/fwlink/?linkid=2250165";
+
+        public const string MissingFunctionsWorkerRuntimeErrorCode = "AZFD0011";
+        public const string MissingFunctionsWorkerRuntimeHelpLink = "https://go.microsoft.com/fwlink/?linkid=2257963";
     }
 }
--- a/src/WebJobs.Script/Host/ScriptHost.cs
+++ b/src/WebJobs.Script/Host/ScriptHost.cs
@@ -446,20 +446,7 @@ namespace Microsoft.Azure.WebJobs.Script
                 }
             }
 
-            if (string.IsNullOrEmpty(_environment.GetFunctionsWorkerRuntime()))
-            {
-                string baseMessage = $"The '{EnvironmentSettingNames.FunctionWorkerRuntime}' setting is required. Please specify a valid value. See {DiagnosticEventConstants.MissingFunctionsWorkerRuntimeHelpLink} for more information.";
-
-                if (_hostingConfigOptions.Value.ThrowOnMissingFunctionsWorkerRuntime)
-                {
-                    _logger.LogDiagnosticEventError(DiagnosticEventConstants.MissingFunctionsWorkerRuntimeErrorCode, baseMessage, DiagnosticEventConstants.MissingFunctionsWorkerRuntimeHelpLink, null);
-                    throw new HostInitializationException(baseMessage);
-                }
-
-                string warningMessage = baseMessage + " The application will continue to run, but may throw an exception in a future release.";
-                _logger.LogDiagnosticEventWarning(DiagnosticEventConstants.MissingFunctionsWorkerRuntimeErrorCode, warningMessage, DiagnosticEventConstants.MissingFunctionsWorkerRuntimeHelpLink, null);
-                _logger.MissingFunctionsWorkerRuntime(warningMessage);
-            }
+            ValidateFunctionsWorkerRuntime(_environment, _hostingConfigOptions, _logger);
 
             // Log whether App Insights is enabled
             if (!string.IsNullOrEmpty(_settingsManager.ApplicationInsightsInstrumentationKey) || !string.IsNullOrEmpty(_settingsManager.ApplicationInsightsConnectionString))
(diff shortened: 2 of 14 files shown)


src/WebJobs.Script/Host/ScriptHost.cs
```diff
@@ -443,6 +446,20 @@ private void PreInitialize()
...
+            {
+                string baseMessage = $"The '{EnvironmentSettingNames.FunctionWorkerRuntime}' App Setting is required. Please specify a valid value. See {DiagnosticEventConstants.MissingFunctionsWorkerRuntimeHelpLink} for more information.";
+                if (_hostingConfigOptions.Value.ThrowOnMissingFunctionsWorkerRuntime)
+                {
+                    _logger.LogDiagnosticEventError(DiagnosticEventConstants.MissingFunctionsWorkerRuntimeErrorCode, baseMessage, DiagnosticEventConstants.MissingFunctionsWorkerRuntimeHelpLink, null);
+                    throw new HostInitializationException(baseMessage);
+                }
+
+                string warningMessage = baseMessage + " The app will continue to run, but may throw an exception in a future release.";
+                _logger.LogDiagnosticEventError(DiagnosticEventConstants.MissingFunctionsWorkerRuntimeErrorCode, warningMessage + " Please .", DiagnosticEventConstants.MissingFunctionsWorkerRuntimeHelpLink, null);
```
kshyju: Should we (create and) use `LogDiagnosticEventError` here instead of error ?

kshyju: Do we need this part `+ " Please .", DiagnosticEventConstants.MissingFunctionsWorkerRuntimeHelpLink` ? The base message already include the link.

src/WebJobs.Script/Diagnostics/DiagnosticEventConstants.cs
```diff
@@ -25,5 +25,8 @@ internal static class DiagnosticEventConstants
 
         public const string LinuxConsumptionTimeZoneErrorCode = "AZFD0010";
         public const string LinuxConsumptionTimeZoneErrorHelpLink = "https://go.microsoft.com/fwlink/?linkid=2250165";
+
+        public const string MissingFunctionsWorkerRuntimeErrorCode = "AZFD0011";
+        public const string MissingFunctionsWorkerRuntimeHelpLink = "https://go.microsoft.com/fwlink/?linkid=2257963";
```
fabiocav, line 30: Do we have anything tracking the configuration of the forward link?

src/WebJobs.Script/Host/ScriptHost.cs
fabiocav: Agree, the concatenation here is a bit confusing. Was this intentional?

brettsam: I've updated this to use `LogDiagnosticEventWarning`.

brettsam: Nope -- that was a "Friday afternoon mistake" -- I've cleaned it up.

src/WebJobs.Script/Diagnostics/DiagnosticEventConstants.cs
brettsam: I'm going to jump over and add text for this right now (that PR may be out before this goes in). Will link them together.

src/WebJobs.Script/Host/ScriptHost.cs
```diff
@@ -443,6 +446,21 @@ private void PreInitialize()
                 }
             }
 
+            if (string.IsNullOrEmpty(_environment.GetFunctionsWorkerRuntime()))
```
brettsam: What are everyone's thoughts on this firing in Core Tools? I can make sure the docs explain how to fix this by adding it to `local.settings.json`, but do we think this will cause an outcry? 

I think I'd rather take the hit and get this in front of everyone everywhere we can... but open to toughts.

liliankasem: It's just a warning right? So I think it's fine to fire in CT - this way folks can get ahead of it before they even deploy

liliankasem (approved):

src/WebJobs.Script/Host/ScriptHost.cs
brettsam: Yes, just a warning.

liliankasem: There is also no harm at all for them to add this app setting so I wouldn't expect too much outcry

kshyju (approved):
Update the release notes?

src/WebJobs.Script/Host/ScriptHost.cs
```diff
@@ -456,6 +461,23 @@ private void PreInitialize()
...
+                string baseMessage = $"The '{EnvironmentSettingNames.FunctionWorkerRuntime}' setting is required. Please specify a valid value. See {DiagnosticEventConstants.MissingFunctionsWorkerRuntimeHelpLink} for more information.";
+
+                if (hostingConfigOptions.Value.ThrowOnMissingFunctionsWorkerRuntime)
+                {
+                    logger.LogDiagnosticEventError(DiagnosticEventConstants.MissingFunctionsWorkerRuntimeErrorCode, baseMessage, DiagnosticEventConstants.MissingFunctionsWorkerRuntimeHelpLink, null);
+                    throw new HostInitializationException(baseMessage);
+                }
+
+                string warningMessage = baseMessage + " The application will continue to run, but may throw an exception in a future release.";
+                logger.LogDiagnosticEventWarning(DiagnosticEventConstants.MissingFunctionsWorkerRuntimeErrorCode, warningMessage, DiagnosticEventConstants.MissingFunctionsWorkerRuntimeHelpLink, null);
```
brettsam, line 477: I validated in prod that this log is good enough to cause the Warning to show in App Insights, Kusto, and the warning banner in the portal.

src/WebJobs.Script/Diagnostics/DiagnosticEventConstants.cs
brettsam: FYI -- PR for this is here -- https://github.com/MicrosoftDocs/azure-docs/pull/119164

PR #9817, opened 2024-01-18
Host Split Update host build pipeline to publish a new FunctionsInproc site extension

### LINKED ISSUE #9722: Host Split: Update host build pipeline to publish a new FunctionsInproc site extension
As part of the in-proc/out-of-proc host split, we'll be temporarily shipping two identical host releases as two separate site extensions (eventually they will diverge).

The first step of this is to update the `dev` branch to define a new build to publish a new site extension artifact, producing a new  `FunctionsInproc` site extension in addition to the `Functions` site extension in this new build.

aishwaryabh:

resolves #9722 

On a high level, the pipeline pushes the dev branch to `release/4.x` and `release/inproc6/4.x`. The first branch will generate an artifact called `Functions.4.1030.zip` and the second branch will generate an artifact called `FunctionsInProc.4.630.zip`.

We are calling in-proc `4.6xx` and OOP `4.10xx` to specify the host .NET version without having to make another query. So `4.6xx` means we're using .NET 6. We chose to add `10` to the minor prefix of the version since we want the version for isolated to be larger than the version for in-proc.

The following site extension is built when testing with `aishwarya/release/inproc6/4.99`:
image

And the following site extension is built when testing with `aishwarya/release/4.99`:
image

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/build/build-extensions.ps1
+++ b/build/build-extensions.ps1
@@ -1,7 +1,7 @@
 param (
   [string]$buildNumber = "0",  
   [string]$suffix = "",
-  [string]$minorVersionPrefix = "10",
+  [string]$minorVersionPrefix = "",
   [string]$hashesForHardlinksFile = "hashesForHardlinks.txt"
 )
 
@@ -245,7 +245,7 @@ function CreateSiteExtensions() {
     if ($minorVersionPrefix -eq "10") {
         ZipContent $siteExtensionPath "$zipOutput\Functions.$extensionVersion$runtimeSuffix.zip"
     } else {
-		ZipContent $siteExtensionPath "$zipOutput\FunctionsInProc.$extensionVersion$runtimeSuffix.zip"
+        ZipContent $siteExtensionPath "$zipOutput\FunctionsInProc.$extensionVersion$runtimeSuffix.zip"
 	}
 
     # Create directory for content even if there is no patch build. This makes artifact uploading easier.
--- a/build/build-extensions.ps1
+++ b/build/build-extensions.ps1
@@ -242,13 +242,11 @@ function CreateSiteExtensions() {
     
     $zipOutput = "$buildOutput\SiteExtension"
     New-Item -Itemtype directory -path $zipOutput -Force > $null
-    if ($minorVersionPrefix -eq "6") {
-        ZipContent $siteExtensionPath "$zipOutput\FunctionsInProc.$extensionVersion$runtimeSuffix.zip"
-    } else {
+    if ($minorVersionPrefix -eq "10") {
         ZipContent $siteExtensionPath "$zipOutput\Functions.$extensionVersion$runtimeSuffix.zip"
     } else {
-		ZipContent $siteExtensionPath "$zipOutput\FunctionsInProc.$extensionVersion$runtimeSuffix.zip"
-	}
+        ZipContent $siteExtensionPath "$zipOutput\FunctionsInProc.$extensionVersion$runtimeSuffix.zip"
+    }
 
     # Create directory for content even if there is no patch build. This makes artifact uploading easier.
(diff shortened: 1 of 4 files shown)


build/common.props
```diff
@@ -4,7 +4,8 @@
     <TargetFramework>netstandard2.1</TargetFramework>
     <LangVersion>latest</LangVersion>
     <MajorVersion>4</MajorVersion>
-    <MinorVersion>30</MinorVersion>
+	<MinorVersionPrefix Condition="'$(MinorVersionPrefix)' == ''">10</MinorVersionPrefix>
+    <MinorVersion>$(MinorVersionPrefix)30</MinorVersion>
```
satvu, line 8: nit: indentation

satvu (approved):

jviau (commented):
Can you elaborate on why we chose in-proc to have `4.6xx` versions and out of proc `4.10xx`?

aishwaryabh:
> Can you elaborate on why we chose in-proc to have `4.6xx` versions and out of proc `4.10xx`?

I added an additional description to the PR:

We are calling in-proc `4.6xx` and OOP `4.10xx` to specify the host .NET version without having to make another query. So `4.6xx` means we're using .NET 6. We chose to add `10` to the minor prefix of the version since we want the version for isolated to be larger than the version for in-proc.

build/common.props
aishwaryabh: wait I don't know why it shows up as indented here when it's not indented in the actual code

image

jviau:
Okay. Now I am wondering why `10` and not `8` 😆 , as it is .net8

brettsam:
@jviau -- there will be a `4.8xx` coming soon -- specifically for .NET 8 in-proc support (we'll have 3 flavors of v4 for some time...). 

We chose `10xx` just to get higher than that so we would avoid any future conflict. We don't expect many .NET 8 in-proc releases -- and certainly not 200 of them to catch up, but `10xx` just seemed nice and round.

And while the new out-of-proc is going to be .NET 8... it'll also move to .NET 9, 10, 11, etc, so the prefix wasn't super-critical for matching that version. It's going to be long-lived so we just wanted that prefix to be high enough to be "out of the way".

build/common.props
brettsam: tabs vs spaces?

build/build-extensions.ps1
```diff
@@ -241,7 +242,11 @@ function CreateSiteExtensions() {
     
     $zipOutput = "$buildOutput\SiteExtension"
     New-Item -Itemtype directory -path $zipOutput -Force > $null
-    ZipContent $siteExtensionPath "$zipOutput\Functions.$extensionVersion$runtimeSuffix.zip"
+    if ($minorVersionPrefix -eq "10") {
+        ZipContent $siteExtensionPath "$zipOutput\Functions.$extensionVersion$runtimeSuffix.zip"
+    } else {
+		ZipContent $siteExtensionPath "$zipOutput\FunctionsInProc.$extensionVersion$runtimeSuffix.zip"
```
brettsam: i think this one is using tabs instead of spaces too... probably renders fine in your editor, but not here. See if you can fix those up?

build/build-extensions.ps1
```diff
@@ -1,14 +1,15 @@
 param (
   [string]$buildNumber = "0",  
-  [string]$suffix = "",  
+  [string]$suffix = "",
+  [string]$minorVersionPrefix = "10",
```
brettsam: This is a big nit... but leave this one default to `""` and let the proj file apply the default. I foresee some confusion in the future if we change the proj file and then gotta figure out where the `10` is coming from.

build/common.props
aishwaryabh: fixed!

brettsam (approved):

PR #9831, opened 2024-01-24
Fix "Debuger" Host log spelling error #9509

### LINKED ISSUE #9509: Fix "Debuger" Host log spelling error
In reviewing logs, I noticed we're logging messages with an event type "DebugerManagerUnableToUpdateSentinelFile". This is because we have a spelling error [here](https://github.com/Azure/azure-functions-host/blob/88b610c61a6dc39e053f5fe7f902204e06aa1b41/src/WebJobs.Script/Diagnostics/Extensions/HostLoggerExtension.cs#L59) we need to fix.

ashpeterson:
Fixed simple spelling error.

resolves #9509

### DIFF
--- a/src/WebJobs.Script/Diagnostics/Extensions/HostLoggerExtension.cs
+++ b/src/WebJobs.Script/Diagnostics/Extensions/HostLoggerExtension.cs
@@ -53,7 +53,7 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics.Extensions
             new EventId(206, nameof(HostConfigFileSystemReadOnly)),
             "File system is read-only. Skipping {fileName} creation.");
 
-        private static readonly Action<ILogger, Exception> _debugerManagerUnableToUpdateSentinelFile =
+        private static readonly Action<ILogger, Exception> _debuggerManagerUnableToUpdateSentinelFile =
             LoggerMessage.Define(
             LogLevel.Error,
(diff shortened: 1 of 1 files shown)


src/WebJobs.Script/Diagnostics/Extensions/HostLoggerExtension.cs
```diff
@@ -56,7 +56,7 @@ public static class HostLoggerExtension
         private static readonly Action<ILogger, Exception> _debugerManagerUnableToUpdateSentinelFile =
             LoggerMessage.Define(
             LogLevel.Error,
-            new EventId(207, nameof(DebugerManagerUnableToUpdateSentinelFile)),
+            new EventId(207, nameof(DebuggerManagerUnableToUpdateSentinelFile)),
```
mathewc, line 59: Needs to be updated in the method name below, as well as at the call site calling that method. So it'll be 3 locations.

ashpeterson: Ah yes, have now updated in all locations.

mathewc: As I mentioned the call site of the method you renamed below also needs to be updated https://github.com/Azure/azure-functions-host/blob/711e83b65d0d80bbec7f2c6e640fe7daa8cb91b6/src/WebJobs.Script/Host/DebugManager.cs#L57

PR #9836, opened 2024-01-26
[pack] adding more values to the log initialization output

brettsam:
Expanding logging for telemetry during host-split and placeholder work. I've tested this on my private stamp and we can now see at-a-glance stuff like this (a dotnet-isolated 6.0 app was specialized onto a java 11 placeholder):

```
{
  "OriginalFunctionWorkerRuntime": "java",
  "FunctionsWorkerRuntime": "dotnet-isolated",
  "OriginalFunctionWorkerRuntimeVersion": "11",
  "FunctionsWorkerRuntimeVersion": "6.0",
  "FunctionsExtensionVesion": "~4",
  "SiteExtensionDirectory": "C:\\Program Files (x86)\\SiteExtensions\\Functions\\4.30.0\\64bit",
  "InStandbyMode": false,
  "HasBeenSpecialized": true,
  "UsePlaceholderDotNetIsolated": true,
  "WebSiteSku": "Dynamic",
  "FeatureFlags": null,
  "HostingConfig": {}
}
```

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
@@ -331,7 +331,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics.Extensions
         }
 
         public static void LogHostInitializationSettings(this ILogger logger, string originalFunctionWorkerRuntime, string functionWorkerRuntime,
-            string originalFunctionWorkerRuntimeVersion, string functionsWorkerRuntimeVersion, string functionExtensionVersion, string siteExtensionDirectory,
+            string originalFunctionWorkerRuntimeVersion, string functionsWorkerRuntimeVersion, string functionExtensionVersion, string currentDirectory,
             bool inStandbyMode, bool hasBeenSpecialized, bool usePlaceholderDotNetIsolated, string websiteSku, string featureFlags,
             IDictionary<string, string> hostingConfig)
         {
@@ -344,7 +344,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics.Extensions
                 OriginalFunctionWorkerRuntimeVersion = originalFunctionWorkerRuntimeVersion,
                 FunctionsWorkerRuntimeVersion = functionsWorkerRuntimeVersion,
                 FunctionsExtensionVesion = functionExtensionVersion,
-                SiteExtensionDirectory = siteExtensionDirectory,
+                HostDirectory = currentDirectory,
                 InStandbyMode = inStandbyMode,
                 HasBeenSpecialized = hasBeenSpecialized,
                 UsePlaceholderDotNetIsolated = usePlaceholderDotNetIsolated,
--- a/src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
@@ -331,7 +331,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics.Extensions
         }
 
         public static void LogHostInitializationSettings(this ILogger logger, string originalFunctionWorkerRuntime, string functionWorkerRuntime,
-            string originalFunctionWorkerRuntimeVersion, string functionsWorkerRuntimeVersion, string functionExtensionVersion, string currentDirectory,
+            string originalFunctionWorkerRuntimeVersion, string functionsWorkerRuntimeVersion, string functionExtensionVersion, string hostDirectory,
             bool inStandbyMode, bool hasBeenSpecialized, bool usePlaceholderDotNetIsolated, string websiteSku, string featureFlags,
             IDictionary<string, string> hostingConfig)
         {
@@ -344,7 +344,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics.Extensions
                 OriginalFunctionWorkerRuntimeVersion = originalFunctionWorkerRuntimeVersion,
                 FunctionsWorkerRuntimeVersion = functionsWorkerRuntimeVersion,
                 FunctionsExtensionVesion = functionExtensionVersion,
-                HostDirectory = currentDirectory,
+                HostDirectory = hostDirectory,
                 InStandbyMode = inStandbyMode,
                 HasBeenSpecialized = hasBeenSpecialized,
                 UsePlaceholderDotNetIsolated = usePlaceholderDotNetIsolated,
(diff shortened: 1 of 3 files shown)


src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
```diff
@@ -330,17 +330,33 @@ public static void ScriptHostServiceRestartCanceledByRuntime(this ILogger logger
...
+            // This is a dump of values for telemetry right now, but eventually we will refactor this
+            // into a proper Options object
+            var initializationSettings = new
+            {
+                OriginalFunctionWorkerRuntime = originalFunctionWorkerRuntime,
+                FunctionsWorkerRuntime = functionWorkerRuntime,
+                OriginalFunctionWorkerRuntimeVersion = originalFunctionWorkerRuntimeVersion,
+                FunctionsWorkerRuntimeVersion = functionsWorkerRuntimeVersion,
+                FunctionsExtensionVesion = functionExtensionVersion,
+                HostDirectory = currentDirectory,
```
brettsam: Note i changed the name of this from `SiteExtensionDirectory` to `HostDirectory` as I realized this wouldn't make sense anywhere but Windows App Service (as it's the only thing that uses a Site Extension).

src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
```diff
@@ -330,17 +330,33 @@ public static void ScriptHostServiceRestartCanceledByRuntime(this ILogger logger
             _scriptHostServiceRestartCanceledByRuntime(logger, null);
         }
 
-        public static void LogHostInitializationSettings(this ILogger logger, string functionWorkerRuntime, string functionExtensionVersion, string siteExtensionDirectory, bool inStandbyMode)
+        public static void LogHostInitializationSettings(this ILogger logger, string originalFunctionWorkerRuntime, string functionWorkerRuntime,
+            string originalFunctionWorkerRuntimeVersion, string functionsWorkerRuntimeVersion, string functionExtensionVersion, string currentDirectory,
```
kshyju: Should we change the parameter name to `hostDirectory`? It's up to the caller to use the current directory or build it some other way.

src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
```diff
@@ -330,17 +330,33 @@ public static void ScriptHostServiceRestartCanceledByRuntime(this ILogger logger
...
+                HostDirectory = currentDirectory,
+                InStandbyMode = inStandbyMode,
+                HasBeenSpecialized = hasBeenSpecialized,
+                UsePlaceholderDotNetIsolated = usePlaceholderDotNetIsolated,
+                WebSiteSku = websiteSku,
+                FeatureFlags = featureFlags,
+                HostingConfig = hostingConfig
+            };
+
+            var options = new JsonSerializerOptions { WriteIndented = true };
```
kshyju, line 356: Question: Do we really care about pretty printing this JSON when logging it? Anyone inspecting this log can copy and paste to a file and an IDE can format it for them.

kshyju (approved):

src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
```diff
@@ -330,17 +330,33 @@ public static void ScriptHostServiceRestartCanceledByRuntime(this ILogger logger
...
-        ""siteExtensionDirectory"": ""{siteExtensionDirectory}"",
-        ""inStandbyMode"": {inStandbyMode}
-    }}
-}}";
-            _logHostInitializationSettings(logger, hostInitializationSettings, null);
+            // This is a dump of values for telemetry right now, but eventually we will refactor this
+            // into a proper Options object
+            var initializationSettings = new
+            {
+                OriginalFunctionWorkerRuntime = originalFunctionWorkerRuntime,
```
kshyju, line 342: `Original` here means  `Placeholder`. Correct?

kshyju (approved):

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -47,6 +49,10 @@ public class WebJobsScriptHostService : IHostedService, IScriptHostManager, ISer
         private readonly SemaphoreSlim _hostStartSemaphore = new SemaphoreSlim(1, 1);
         private readonly TaskCompletionSource<bool> _hostStartedSource = new TaskCompletionSource<bool>();
         private readonly Task _hostStarted;
+        private readonly IOptions<FunctionsHostingConfigOptions> _hostingConfigOptions;
```
aishwaryabh, line 52: Where is FunctionsHostingConfig options coming from?

aishwaryabh (approved):

src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
brettsam: yeah, let me update that...

src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
brettsam: Technically, there's really no reason to write it prettily... but in reality, it's a lot nicer to parse when reading kusto results...

src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
brettsam: Correct. But not all hosts are specialized, and I didn't want to make this too crazily complicated and only write some values when specialize and leave ones off when not specialized, etc. So, I decided this would hopefully be clear enough for us to pick apart the values we wanted for reports. In some cases of pre-specialization and non-specialization, these values will always be the same. But I think that's okay.

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
brettsam: It's injected into the constructor via DI -- but ultimately it gets created by parsing a file that Antares writes for us here: https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Config/FunctionsHostingConfigSource.cs#L24. 

That file gets turned into a dictionary, which goes into `IConfiguration`... which then is used to create us this `Options` object via DI.

brettsam: This file has key/value pairs of "flags" that we can set on a stamp-by-stamp value. They're sometimes important to how we decide to specialize, etc, so wanted to make sure we could see all available flags in one place (we currently don't log them).

PR #9846, opened 2024-01-31
 Adding an app setting to allow CORS configuration

### LINKED ISSUE #8936: Support CORS configuration via environment variables when not running in Linux consumption

Currently, it's possible to configure CORS via environment variables, but the variables are only used when running in Linux consumption:

https://github.com/Azure/azure-functions-host/blob/172c57f3bc8d3093284f7282ab3ddd5598edde76/src/WebJobs.Script.WebHost/Configuration/CorsOptionsSetup.cs#L26

It can be useful to configure CORS via environment variables, such as when running in a container. Would be great to allow this.

Allow CORS configuration via environment variables when running outside Linux consumption.

Setting a variable named `CONTAINER_NAME` will cause the host to go down the same code path as Linux consumption, but on environments such as Container Apps or Kubernetes, setting that variable will provide the same container name to all running replicas/pods. This might have unintended consequences if other parts of the host uses the container name and expects it to be unique per host instance.

#### Additional context

Thread: https://hachyderm.io/@anthonychu/109400532768512364

soninaren:
resolves #8936

Host currently honors 'CORS_ALLOWED_ORIGINS' app setting only for linux consumption. Windows and linux app service go thorugh a seperate channel. However there are additional hosting scenarios that have no way of setting up CORS configuration. This change adds an app setting that allows customer to set up CORS configuration by setting `"FUNCTIONS_ENABLE_CORS_CONFIGURATION" = "true"`

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -13,3 +13,4 @@
     - Customers can opt-out of this behavior by setting `SendCanceledInvocationsToWorker` to `false` in host.json
   - If a worker does not support CancellationTokens, cancelled invocations will not be sent to the worker
 - Warn when `FUNCTIONS_WORKER_RUNTIME` is not set (#9799)
+- Add an app setting to allow CORS configuration (#9846)
--- a/src/WebJobs.Script.WebHost/Configuration/CorsOptionsSetup.cs
+++ b/src/WebJobs.Script.WebHost/Configuration/CorsOptionsSetup.cs
@@ -2,12 +2,8 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Generic;
 using System.Linq;
-using System.Runtime.InteropServices;
-using System.Threading.Tasks;
 using Microsoft.AspNetCore.Cors.Infrastructure;
-using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Options;
 
 namespace Microsoft.Azure.WebJobs.Script.WebHost.Configuration
@@ -25,7 +21,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Configuration
 
         public void Configure(CorsOptions options)
         {
-            if (_env.IsAnyLinuxConsumption() || _env.IsCorsConfigurationAllowed())
+            if (_env.IsAnyLinuxConsumption() || _env.IsCorsConfigurationEnabled())
             {
                 string[] allowedOrigins = _hostCorsOptions.Value.AllowedOrigins?.ToArray() ?? Array.Empty<string>();
                 var policyBuilder = new CorsPolicyBuilder(allowedOrigins);
--- a/src/WebJobs.Script.WebHost/Configuration/CorsOptionsSetup.cs
+++ b/src/WebJobs.Script.WebHost/Configuration/CorsOptionsSetup.cs
@@ -7,6 +7,7 @@ using System.Linq;
 using System.Runtime.InteropServices;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Cors.Infrastructure;
+using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Options;
 
 namespace Microsoft.Azure.WebJobs.Script.WebHost.Configuration
@@ -24,7 +25,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Configuration
 
         public void Configure(CorsOptions options)
         {
-            if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+            if (_env.IsAnyLinuxConsumption() || _env.IsCorsConfigurationAllowed())
             {
                 string[] allowedOrigins = _hostCorsOptions.Value.AllowedOrigins?.ToArray() ?? Array.Empty<string>();
                 var policyBuilder = new CorsPolicyBuilder(allowedOrigins);
--- a/src/WebJobs.Script.WebHost/Configuration/CorsOptionsSetup.cs
+++ b/src/WebJobs.Script.WebHost/Configuration/CorsOptionsSetup.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.InteropServices;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Cors.Infrastructure;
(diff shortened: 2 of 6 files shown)


src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
```diff
@@ -123,11 +124,15 @@ public static class WebScriptHostBuilderExtension
                     services.TryAddSingleton<IJobHostMiddlewarePipeline, DefaultMiddlewarePipeline>();
                     services.TryAddEnumerable(ServiceDescriptor.Singleton<IJobHostHttpMiddleware, CustomHttpHeadersMiddleware>());
                     services.TryAddEnumerable(ServiceDescriptor.Singleton<IJobHostHttpMiddleware, HstsConfigurationMiddleware>());
-                    if (environment.IsAnyLinuxConsumption())
+
+                    if (environment.IsAnyLinuxConsumption() || environment.IsCorsConfigurationAllowed())
```
kshyju: you could store the result of `environment.IsAnyLinuxConsumption()` to a variable and use in line 128 and 134, so that the code inside that method gets executed only once.

src/WebJobs.Script.WebHost/Configuration/CorsOptionsSetup.cs
```diff
@@ -4,8 +4,10 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.InteropServices;
 using System.Threading.Tasks;
```
kshyju: nit: Remove unused using directive statements.

src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
```diff
@@ -112,6 +112,7 @@ public static class EnvironmentSettingNames
         public const string LinuxAzureAppServiceStorage = "WEBSITES_ENABLE_APP_SERVICE_STORAGE";
         public const string CoreToolsEnvironment = "FUNCTIONS_CORETOOLS_ENVIRONMENT";
         public const string RunningInContainer = "DOTNET_RUNNING_IN_CONTAINER";
+        public const string AllowCorsConfiguration = "ALLOW_CORS_CONFIGURATION";
```
kshyju: Question: Should we use the "FUNCTIONS_" or "WEBSITE_" prefix to be more specific (to prevent conflicts). The current name is more generic.

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -165,6 +165,14 @@ public static bool IsContainer(this IEnvironment environment)
                 && runningInContainerValue;
         }
 
+        public static bool IsCorsConfigurationAllowed(this IEnvironment environment)
+        {
+            var allowCorsConfiguration = environment.GetEnvironmentVariable(AllowCorsConfiguration);
+            return !string.IsNullOrEmpty(allowCorsConfiguration)
+                && bool.TryParse(allowCorsConfiguration, out bool allowCorsConfigurationValue)
```
kshyju: In other places, we look for the value "1" for truthiness. Should we stick to that approach?

https://github.com/Azure/azure-functions-host/blob/41b15ce28d4a012a4118840332a2cdd2b88ced85/src/WebJobs.Script/Environment/EnvironmentExtensions.cs#L50-L57

release_notes.md
```diff
@@ -13,3 +13,4 @@
     - Customers can opt-out of this behavior by setting `SendCanceledInvocationsToWorker` to `false` in host.json
   - If a worker does not support CancellationTokens, cancelled invocations will not be sent to the worker
 - Warn when `FUNCTIONS_WORKER_RUNTIME` is not set (#9799)
+- Add an app setting to allow CORS configuration (#9846)
```
kshyju, line 16: Wonder we should make this more specific. something like "Expanding CORS support for non Linux scenarios" or something along that line.

kshyju (approved):
A few questions, otherwise LGTM.

src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
kshyju: We previously got bitten by using generic env variable names. See https://github.com/Azure/azure-functions-host/issues/9504

release_notes.md
soninaren: The app setting would apply to all OS though. So it would be little misleading to say that.

src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
soninaren: That is good point, let me update the variable name.

fabiocav: Prefixing is a good idea (`FUNCTIONS_` would be preferrable.

We should use "Enable" here instead of "Allow". Same applies to the extension method exposed.

release_notes.md
liliankasem: "Expanding" implies that this feature already exists for linux

liliankasem (approved):

Boryslav:
I've been trying to use combination of FUNCTIONS_ENABLE_CORS_CONFIGURATION and CORS_ALLOWED_ORIGINS in a function app hosted in Azure Kubernetes but could not get it to work, CORS requests fail.
Still only way it works for me is by using CONTAINER_NAME and CORS_ALLOWED_ORIGINS.
If anyone had success with using FUNCTIONS_ENABLE_CORS_CONFIGURATION, please let me know.

devodo:
> I've been trying to use combination of FUNCTIONS_ENABLE_CORS_CONFIGURATION and CORS_ALLOWED_ORIGINS in a function app hosted in Azure Kubernetes but could not get it to work, CORS requests fail. Still only way it works for me is by using CONTAINER_NAME and CORS_ALLOWED_ORIGINS. If anyone had success with using FUNCTIONS_ENABLE_CORS_CONFIGURATION, please let me know.

@Boryslav I got CORS working locally with the following:
- `CORS_ALLOWED_ORIGINS="[\"*\"]"`
- `FUNCTIONS_ENABLE_CORS_CONFIGURATION=1`

Note: `FUNCTIONS_ENABLE_CORS_CONFIGURATION="true"` as per the PR description did not work for me.

PR #8465, opened 2022-06-13
[pack] Create hardlink marker as part of build

yojagad:
### Issue describing the changes in this PR

Temporary workaround until Ant 99 goes out.

Test Build : https://azfunc.visualstudio.com/Azure%20Functions/_build/results?buildId=75328&view=results

### Pull request checklist

* [ ] My changes **do not** require documentation changes
    * [ ] Otherwise: Documentation issue linked to PR
* [ ] My changes **should not** be added to the release notes for the next release
    * [ ] Otherwise: I've added my notes to `release_notes.md`
* [ ] My changes **do not** need to be backported to a previous version
    * [ ] Otherwise: Backport tracked by issue/PR #issue_or_pr
* [ ] My changes **do not** require diagnostic events changes
    * Otherwise: I have added/updated all related diagnostic events and their documentation (Documentation issue linked to PR)
* [ ] I have added all required tests (Unit tests, E2E tests)

### Additional information

Additional PR information

PR #9862, opened 2024-02-07
re-enabling hard links

brettsam:
We use hard links to save disk space in production. #8465 disabled them due to a regression several years ago and it was never re-enabled. 

### Pull request checklist

### DIFF
--- a/build/build-extensions.ps1
+++ b/build/build-extensions.ps1
@@ -226,9 +226,6 @@ function CreateSiteExtensions() {
     # This goes in the root dir
     Copy-Item $rootDir\src\WebJobs.Script.WebHost\extension.xml $siteExtensionPath > $null
     
-    #This needs to be removed post Ant 99 as it's a temporary workaround
-    New-Item "$siteExtensionPath\$extensionVersion.hardlinksCreated" > $null
-    
     Write-Host "Done copying. Elapsed: $($stopwatch.Elapsed)"
     Write-Host "======================================"
     Write-Host ""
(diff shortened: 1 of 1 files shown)


kshyju (approved):

build/build-extensions.ps1
```diff
@@ -241,18 +238,25 @@ function CreateSiteExtensions() {
...
-        Write-Host "Removing workers."
+        Write-Host "Removing workers before zipping."
         Remove-Item -Recurse -Force "$siteExtensionPath\$extensionVersion$runtimeSuffix\workers" -ErrorAction Stop
         # The host requires that this folder exists, even if it's empty
-        New-Item -Itemtype directory -path $siteExtensionPath\$extensionVersion$runtimeSuffix\workers
+        New-Item -Itemtype directory -path $siteExtensionPath\$extensionVersion$runtimeSuffix\workers > $null 
         Write-Host
         ZipContent $siteExtensionPath "$zipOutput\FunctionsInProc8.$extensionVersion$runtimeSuffix.zip"
     } elseif ($minorVersionPrefix -eq "6") {
+        # Only the "Functions" site extension supports hard links
```
mattchenderson, line 257: Do we not need `hashesForHardLinks` temporarily still? "6" remains in "Functions" at the moment. Or I guess we're saying they only become available once "10" is firmly in place? Without the `hardlinksCreated` marker. I guess we don't need the hashes to be provided?

mattchenderson (commented):
I confused myself a bit looking over this. I might appreciate a quick chat about it, but I shouldn't be a blocker.

build/build-extensions.ps1
brettsam: `hashesforhardlinks` is the signal to Antares that this extension supports hard links and it should wait for the `hardlinkscreated` file to exist.

so we could either:
1. artificially create that `hardlinkscreated` file (this is what we'd been doing in Functions)
2. not include the `hashesforhardlinks` at all

number 2 is more correct so i'm opting for deleting that file from FunctionsInProc and FunctiosnInProc8 and only leaving it for Functions (which is all that's supported)

fabiocav (approved):

PR #9872, opened 2024-02-09
Adding HIS Enforcement feature flags and metrics logging

mathewc:
These changes add new capabilities to the host for both logging HIS (highly identifiable secret) status for loaded keys, as well as allowing the host to be put in strict/warn modes for HIS enforcement.

Related to the work we did a couple years ago in https://github.com/Azure/azure-functions-host/pull/8071. After those changes, all keys the platform generates are HIS. However, any keys generated before that remain non-HIS. The changes in this PR enable customers to identify legacy keys (and rotate them) via the Warn mode, as well as enforce that no non-HIS keys can be added/used in the future via the Strict mode.

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
+++ b/src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
@@ -11,14 +11,13 @@ using System.Security.Cryptography;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
-using DryIoc;
-using Microsoft.Azure.Web.DataProtection;
 using Microsoft.Azure.WebJobs.Extensions.Http;
+using Microsoft.Azure.WebJobs.Script.Config;
 using Microsoft.Azure.WebJobs.Script.Diagnostics;
 using Microsoft.Azure.WebJobs.Script.WebHost.Properties;
 using Microsoft.Azure.WebJobs.Script.WebHost.Security;
-using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
+using static Microsoft.Azure.WebJobs.Script.WebHost.Models.FunctionAppSecrets;
 using DataProtectionConstants = Microsoft.Azure.Web.DataProtection.Constants;
 
 namespace Microsoft.Azure.WebJobs.Script.WebHost
@@ -30,6 +29,9 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
         private readonly ISecretsRepository _repository;
         private readonly HostNameProvider _hostNameProvider;
         private readonly StartupContextProvider _startupContextProvider;
+        private readonly Lazy<bool> _strictHISFeatureEnabled = new Lazy<bool>(() => FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagStrictHISModeEnabled));
+        private readonly Lazy<bool> _strictHISWarnFeatureEnabled = new Lazy<bool>(() => FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagStrictHISModeWarn));
+        private readonly HashSet<string> _invalidNonHISKeys = new HashSet<string>();
         private ConcurrentDictionary<string, IDictionary<string, string>> _functionSecrets;
         private ConcurrentDictionary<string, (string, AuthorizationLevel)> _authorizationCache = new ConcurrentDictionary<string, (string, AuthorizationLevel)>(StringComparer.OrdinalIgnoreCase);
         private HostSecretsInfo _hostSecrets;
@@ -137,11 +139,17 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
                             await RefreshSecretsAsync(hostSecrets);
                         }
 
+                        // before caching  any secrets, validate them
+                        string masterKeyValue = hostSecrets.MasterKey.Value;
+                        var functionKeys = hostSecrets.FunctionKeys.ToDictionary(p => p.Name, p => p.Value);
+                        var systemKeys = hostSecrets.SystemKeys.ToDictionary(p => p.Name, p => p.Value);
+                        ValidateHostSecrets(masterKeyValue, functionKeys, systemKeys);
+
                         _hostSecrets = new HostSecretsInfo
                         {
-                            MasterKey = hostSecrets.MasterKey.Value,
-                            FunctionKeys = hostSecrets.FunctionKeys.ToDictionary(s => s.Name, s => s.Value),
-                            SystemKeys = hostSecrets.SystemKeys.ToDictionary(s => s.Name, s => s.Value)
+                            MasterKey = masterKeyValue,
+                            FunctionKeys = functionKeys,
+                            SystemKeys = systemKeys
                         };
                     }
                     finally
@@ -219,7 +227,10 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
                             await RefreshSecretsAsync(secrets, functionName);
                         }
 
+                        // before caching any secrets, validate them
                         var result = secrets.Keys.ToDictionary(s => s.Name, s => s.Value);
+                        ValidateSecrets(result, SecretGenerator.FunctionKeySeed, functionName);
+
                         functionSecrets = _functionSecrets.AddOrUpdate(functionName, result, (n, r) => result);
                     }
                     finally
@@ -290,6 +301,15 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
                 }
                 else
                 {
+                    if (_strictHISFeatureEnabled.Value)
+                    {
+                        // if an explicit value has been provided and strict HIS mode is enabled, validate the secret
+                        if (!SecretGenerator.TryValidateSecret(value, SecretGenerator.MasterKeySeed))
+                        {
+                            return new KeyOperationResult(value, OperationResult.BadRequest);
+                        }
+                    }
+
                     // Use the provided secret
                     masterKey = value;
                     result = OperationResult.Updated;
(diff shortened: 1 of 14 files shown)


src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
```diff
@@ -514,6 +558,52 @@ private static bool HasMatchingKey(IDictionary<string, string> secrets, string k
...
+            }
+            else
+            {
+                _metricsLogger.LogEvent(MetricEventNames.NonIdentifiableSecretLoaded, functionName);
+
+                if (_strictHISFeatureEnabled.Value)
+                {
+                    // when strict mode is enabled, we won't let the application continue to run with
+                    // non-compliant secrets
+                    throw new InvalidOperationException(string.Format(Resources.NonHISSecret, SecretGenerator.GetKeyType(seed), keyName, functionName));
```
mathewc: I've decided to make this a hard error - the host will simply refuse to load a non-HIS secret in this mode. Depending on the codepath that is loading secrets, the behavior may differ. On an http request path (user http or admin APIs), the request will fail. On some host startup paths (e.g. when startup context cache is enabled [here](https://github.com/Azure/azure-functions-host/blob/ad0843d26bfe7c2555bcee250da1dacccdc61adf/src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs#L842)) the host startup will fail.

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
```diff
@@ -514,6 +558,52 @@
...
+                }
+                else if (_strictHISWarnFeatureEnabled.Value)
+                {
+                    // in warn mode, we'll log a warning, but still allow the application to run
+                    string message = string.Format(Resources.NonHISSecret, SecretGenerator.GetKeyType(seed), keyName, functionName);
+                    _logger.LogDiagnosticEvent(LogLevel.Warning, 0, DiagnosticEventConstants.NonHISSecretLoaded, message, DiagnosticEventConstants.NonHISSecretLoadedHelpLink, exception: null);
```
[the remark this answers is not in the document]
mathewc: I think this is a false positive

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
```diff
@@ -329,11 +342,42 @@ public async Task<bool> DeleteSecretAsync(string secretName, string keyScope, Sc
...
+            }
+            else if (secretsType == ScriptSecretsType.Host)
+            {
+                if (string.Equals(keyScope, HostKeyScopes.FunctionKeys, StringComparison.OrdinalIgnoreCase))
+                {
+                    return SecretGenerator.FunctionKeySeed;
+                }
+                else if (string.Equals(keyScope, HostKeyScopes.FunctionKeys, StringComparison.OrdinalIgnoreCase))
+                {
+                    return SecretGenerator.FunctionKeySeed;
```
paulbatum: I am probably missing something simple, but it looks like both of these ifs check for the same thing and the branches both do the same thing

mathewc: Nice catch! My integration test didn't catch this because it only exercised Function keys, not system keys. I've added a unit test GetKeySeed_ReturnsExpectedResult.

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
```diff
@@ -30,6 +29,8 @@ public class SecretManager : IDisposable, ISecretManager
         private readonly ISecretsRepository _repository;
         private readonly HostNameProvider _hostNameProvider;
         private readonly StartupContextProvider _startupContextProvider;
+        private readonly Lazy<bool> _strictHISFeatureEnabled = new Lazy<bool>(() => FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagStrictHISModeEnabled));
+        private readonly Lazy<bool> _strictHISWarnFeatureEnabled = new Lazy<bool>(() => FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagStrictHISModeWarn));
```
brettsam, line 33: Is the plan that customers set these flags themselves? Are we going to eventually roll "Warn mode" out to everyone via ScmHostingConfig (or just make it the default at some point?)

mathewc: No, we will never be rolling out a change because that would be breaking. Customer need to opt into this, and the two different modes allow them to work towards strict, by first using warn, regenerating the specified keys, then switching to strict. In strict mode, attempts to use non-HIS keys will result in 401s.

We might eventually support an Azure Policy addressable ARM mode that can be enabled for this. That would allow organizations to enforce this for all their apps.

src/WebJobs.Script/Diagnostics/DiagnosticEventConstants.cs
```diff
@@ -28,5 +28,8 @@ internal static class DiagnosticEventConstants
 
         public const string MissingFunctionsWorkerRuntimeErrorCode = "AZFD0011";
         public const string MissingFunctionsWorkerRuntimeHelpLink = "https://go.microsoft.com/fwlink/?linkid=2257963";
+
+        public const string NonHISSecretLoaded = "AZFD0012";
+        public const string NonHISSecretLoadedHelpLink = "https://aka.ms/functions-non-his-secrets";
```
mathewc, line 33: This is link that I generated that points to the internal page we have as part of the Project Standard Secrets work. Initially we'll be using this feature for first party internal customers. However, given this is a feature that the public could also use if they wish, need to think about what this should point to.

brettsam (approved):

src/WebJobs.Script.WebHost/Security/SecretGenerator.cs
```diff
@@ -45,5 +60,22 @@ internal static string GenerateIdentifiableSecret(ulong seed)
             // equal signs do not require encoding.
             return IdentifiableSecrets.GenerateUrlSafeBase64Key(seed, 40, AzureFunctionsSignature, elidePadding: false);
         }
+
+        internal static bool ValidateSecret(string key, ulong seed)
+        {
+            return IdentifiableSecrets.ValidateBase64Key(key, seed, SecretGenerator.AzureFunctionsSignature, encodeForUrl: true);
```
fabiocav: nit: `SecretGenerator.AzureFunctionsSignature` -> `AzureFunctionsSignature`

src/WebJobs.Script/Diagnostics/DiagnosticEventConstants.cs
fabiocav: We need public docs for this so we can have a link we can use in a diagnostic event.

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
```diff
@@ -329,11 +349,42 @@ public async Task<bool> DeleteSecretAsync(string secretName, string keyScope, Sc
             }
         }
 
+        internal static ulong GetKeySeed(ScriptSecretsType secretsType, string keyScope)
```
fabiocav, line 352: `SecretGenerator` has been modified to return a key type from a seed, would it make sense to move this method there as well so we have type -> seed and seed -> type handed in the same place?

src/WebJobs.Script.WebHost/Security/SecretGenerator.cs
```diff
@@ -45,5 +60,22 @@ internal static string GenerateIdentifiableSecret(ulong seed)
...
+        }
+
+        internal static bool TryValidateSecret(string key, ulong seed)
```
fabiocav, line 54: Could we update this API to take a secret type instead of a seed? That would simplify usage as it wouldn't require the resolution from secret type to a seed first (or having knowledge of what seed to use).

Once that is in place, you might be able to completely remove `GetKeyType`

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
```diff
@@ -419,6 +469,22 @@ private async Task<ScriptSecrets> LoadSecretsAsync(ScriptSecretsType type, strin
 
         public async Task<(string KeyName, AuthorizationLevel Level)> GetAuthorizationLevelOrNullAsync(string keyValue, string functionName = null)
         {
+            // local helper function to get auth level and also enforce HIS validation
+            var getAuthorizationLevelAndValidateAsync = async (string keyValue, string functionName) =>
```
fabiocav: Use a local function here instead

src/WebJobs.Script/Diagnostics/DiagnosticEventConstants.cs
mathewc: What I'll do is write up an initial wiki help page on https://github.com/Azure/azure-functions and update the aka link to point to that. I don't think finalizing this needs to block the PR since that link can be updated independently.

fabiocav: We have a pattern for diagnostic events and they're all setup here: https://learn.microsoft.com/en-us/azure/azure-functions/errors-diagnostics/diagnostic-events/azfd0001

Let me know if you need any help with this. The links also feed into reports used to make documentation improvements.

mathewc: I'll create a separate PR adding a new page here as you suggest. I'll do this separate from this PR, and update the aka link when done.
image

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
mathewc: Done

src/WebJobs.Script.WebHost/Security/SecretGenerator.cs
mathewc: I looked at this and making that change will result in a bunch of cascading changes that I don't want to make right now. Can consider as a future refactoring.

src/WebJobs.Script/Diagnostics/DiagnosticEventConstants.cs
mathewc: Doc PR is out here: https://github.com/MicrosoftDocs/azure-docs/pull/120308

paulbatum (approved):

fabiocav (approved):

PR #9891, opened 2024-02-22
removing call to Configure() for FunctionsHostingConfigOptions

brettsam:
Calling `Configure()` ultimately calls `IConfiguration.Bind(options)` on the specified `Options` object, This was problematic for this `FunctionsHostingConfigOptions` because several of us attempted to use ints (`myoption=1`) in the ScmHostingConfiguration file... which ended up throwing exceptions because ints are not parse-able into bools.

The longer-term approach here is that everything is driven by the `Features` dictionary, which is already populated by the `FunctionsHostingConfigOptionsSetup` class. There was no need to call `Configure()` so I'm removing it for safety going forward. As an extra precaution, I've move all properties to `internal` just in case this is re-introduced at any point.

Some other interesting notes:
- a bunch of properties do not have an exact match between what they use in ScmHostingConfigurations.txt and the property. This is why many don't see that issue (`WorkerIndexingEnabled`, for example).
- only when several of us added properties where the name exactly matches that in ScmHostingConfigurations.txt did we run into issues, which actually can cause outages.
- a few features have already rolled out using "True" or "False" as their string. So I've added a way to easily support "True/False/1/0". Ideally everyone going forward just uses "1/0".

I've also added a test to ensure that future additions here add new validation tests so we're all clear what is/isn't supported.

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
+++ b/src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
@@ -208,13 +208,25 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
             services.ConfigureOptionsWithChangeTokenSource<HttpBodyControlOptions, HttpBodyControlOptionsSetup, SpecializationChangeTokenSource<HttpBodyControlOptions>>();
             services.ConfigureOptions<FlexConsumptionMetricsPublisherOptionsSetup>();
             services.ConfigureOptions<ConsoleLoggingOptionsSetup>();
-            services.ConfigureOptions<FunctionsHostingConfigOptionsSetup>();
+            services.AddHostingConfigOptions(configuration);
 
             services.TryAddSingleton<IDependencyValidator, DependencyValidator>();
             services.TryAddSingleton<IJobHostMiddlewarePipeline>(s => DefaultMiddlewarePipeline.Empty);
 
             // Add AzureBlobStorageProvider to WebHost (also needed for ScriptHost)
             services.AddAzureBlobStorageProvider();
+        }
+
+        internal static void AddHostingConfigOptions(this IServiceCollection services, IConfiguration configuration)
+        {
+            services.ConfigureOptions<FunctionsHostingConfigOptionsSetup>();
+
+            if (configuration != null)
+            {
+                // Refresh IOptionsMonitor<FunctionsHostingConfigOptions> when the config section changes
+                var configSection = configuration.GetSection(ScriptConstants.FunctionsHostingConfigSectionName);
+                services.AddSingleton<IOptionsChangeTokenSource<FunctionsHostingConfigOptions>>(new ConfigurationChangeTokenSource<FunctionsHostingConfigOptions>(configSection));
+            }
         }
 
         private static void AddStandbyServices(this IServiceCollection services)
--- a/src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
+++ b/src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
@@ -209,10 +209,6 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
             services.ConfigureOptions<FlexConsumptionMetricsPublisherOptionsSetup>();
             services.ConfigureOptions<ConsoleLoggingOptionsSetup>();
             services.ConfigureOptions<FunctionsHostingConfigOptionsSetup>();
-            if (configuration != null)
-            {
-                services.Configure<FunctionsHostingConfigOptions>(configuration.GetSection(ScriptConstants.FunctionsHostingConfigSectionName));
-            }
 
             services.TryAddSingleton<IDependencyValidator, DependencyValidator>();
             services.TryAddSingleton<IJobHostMiddlewarePipeline>(s => DefaultMiddlewarePipeline.Empty);
--- a/src/WebJobs.Script/Config/FunctionsHostingConfigOptions.cs
+++ b/src/WebJobs.Script/Config/FunctionsHostingConfigOptions.cs
@@ -50,7 +50,7 @@ namespace Microsoft.Azure.WebJobs.Script.Config
         {
             get
             {
-                return GetFeatureAsBooleanOrDefault(RpcWorkerConstants.ShutdownWebhostWorkerChannelsOnHostShutdown, false);
+                return GetFeatureAsBooleanOrDefault(RpcWorkerConstants.ShutdownWebhostWorkerChannelsOnHostShutdown, true);
             }
 
             set
--- a/src/WebJobs.Script/Config/FunctionsHostingConfigOptions.cs
+++ b/src/WebJobs.Script/Config/FunctionsHostingConfigOptions.cs
@@ -24,7 +24,7 @@ namespace Microsoft.Azure.WebJobs.Script.Config
         /// <summary>
         /// Gets a value indicating whether worker concurrency feature is enabled in the hosting config.
         /// </summary>
-        public bool FunctionsWorkerDynamicConcurrencyEnabled
+        internal bool FunctionsWorkerDynamicConcurrencyEnabled
         {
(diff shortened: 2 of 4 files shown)


src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
```diff
@@ -221,6 +217,18 @@ public static void AddWebJobsScriptHost(this IServiceCollection services, IConfi
...
 
+        internal static void AddHostingConfigOptions(this IServiceCollection services, IConfiguration configuration)
+        {
+            services.ConfigureOptions<FunctionsHostingConfigOptionsSetup>();
+
+            if (configuration != null)
+            {
+                // Refresh IOptionsMonitor<FunctionsHostingConfigOptions> when the config section changes
+                var configSection = configuration.GetSection(ScriptConstants.FunctionsHostingConfigSectionName);
+                services.AddSingleton<IOptionsChangeTokenSource<FunctionsHostingConfigOptions>>(new ConfigurationChangeTokenSource<FunctionsHostingConfigOptions>(configSection));
```
brettsam, line 228: note -- the call to `Configure()` did this internally already (and we have tests ensuring that `IOptionsMonitor<>` works) but it also follws up with a call to `Bind()`, which we don't want. Separating this out avoids the call to `Bind()`.

src/WebJobs.Script/Config/FunctionsHostingConfigOptions.cs
```diff
@@ -24,7 +24,7 @@ public FunctionsHostingConfigOptions()
         /// <summary>
         /// Gets a value indicating whether worker concurrency feature is enabled in the hosting config.
         /// </summary>
-        public bool FunctionsWorkerDynamicConcurrencyEnabled
+        internal bool FunctionsWorkerDynamicConcurrencyEnabled
```
fabiocav, line 27: Are you changing visibility to avoid having this being evaluated/set by `Bind()` calls in case that ever happens again? With the current changes, this would no longer be a requirement, right?

fabiocav (approved):

fabiocav:
Approved, but @mathewc will be interested in this change as he was recently hit by a type conversion issue as a result of a call to `Bind()`

Updating:

> a few features have already rolled out using "True" or "False" as their string. So I've added a way to easily support "True/False/1/0". Ideally everyone going forward just uses "1/0".

This was the exact issue he ran into

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
mathewc: If we no longer Bind, does that mean we'll lose the ability to configure hosting config via app settings? For example say we've rolled out a particular config value, but it breaks a customer resulting in a CRI. We could currently tell that customer to revert the config via app settings. That's a nice capability. Though I realize the options type wasn't really designed for this - most props don't have setters, and for many the setting name doesn't match the property name.

Note that there are currently config provider ordering issues preventing app settings override like this, but to fix that we just need to swap the order https://github.com/Azure/azure-functions-host/blob/937267a5d78b006456446be62ecb120d10c0208d/src/WebJobs.Script.WebHost/Program.cs#L78 so it comes BEFORE environment config, allowing environment to override.

brettsam: That will work once the ordering is fixed (I'd rather do that separately). Ultimately, the settings are all fulfilled via `IConfiguration` today, just via our custom `Setup` class here: https://github.com/Azure/azure-functions-host/blob/937267a5d78b006456446be62ecb120d10c0208d/src/WebJobs.Script/Config/FunctionsHostingConfigOptionsSetup.cs#L23C1-L36C10.

That populates `Features` with `IConfiguration`. So, if a customer adds an app setting that overrides a specific value being looked for here, it'll work. You won't need setters, either -- the getter can evaluate what's in the `Features` dict and decide true/false from there.

The `Bind()` wasn't even doing anything unless the Property name matched the ScmHostingConfig key name -- which, until you added yours, had never happened. Then I followed up with the same thing (matching names) and fell into the same issue.

src/WebJobs.Script/Config/FunctionsHostingConfigOptions.cs
brettsam: Correct. But it's a double-protection. Because the `Bind()` wasn't obvious previously, I was worried someone could inadvertently add some code that calls `Bind()` again. There's no reason for these to be public, and it makes it impossible for them to be set by any future `Bind()`.

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SamplesEndToEndTests_Node_MultipleProcessesNoBundle.cs
```diff
@@ -23,35 +19,31 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.EndToEnd
...
-            // Verify node process is different after host restart
-            var result = nodeProcessesAfter.Where(pId1 => !nodeProcessesBeforeHostRestart.Any(pId2 => pId2 == pId1) && !_fixture.NodeProcessesBeforeTestStarted.Any(pId3 => pId3 == pId1));
-            Assert.Equal(3, result.Count());
+            await TestHelpers.Await(() =>
+                {
+                    IEnumerable<int> nodeProcessesAfter = Process.GetProcessesByName("node").Select(p => p.Id);
+                    // Verify node process is different after host restart
+                    var result = nodeProcessesAfter.Where(pId1 => !nodeProcessesBeforeHostRestart.Any(pId2 => pId2 == pId1) && !_fixture.NodeProcessesBeforeTestStarted.Any(pId3 => pId3 == pId1));
+                    return result.Count() == 3;
+                });
```
brettsam, line 46: Was debugging a test failure and didn't like the permanent delay here. This should save ~50 seconds off the runtime of this test.

mathewc (approved):

mathewc:
Approved - thanks for taking care of this Brett :)

brettsam:
Wanted to add a couple extra post-merge notes as they came up during reviews:

### Removing call to `Configure<TOptions>()`:
We were previously calling [`Configure<TOptions>(IConfiguration)`](https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Options.ConfigurationExtensions/src/OptionsConfigurationServiceCollectionExtensions.cs#L72-L74), which ultimately does 3 things:
  1. [`AddOptions()`](https://github.com/dotnet/runtime/blob/438dabb3e98440ceab44dda4faa4ce018165e1c0/src/libraries/Microsoft.Extensions.Options/src/OptionsServiceCollectionExtensions.cs#L22-L32) which registers a bunch of the default options services if they're not already registered.
  2. Registers an `ConfigurationChangeTokenSource` so that `IOptionsMonitor<>` will be updated if the config file changes. This allows anyone using `IOptionsMonitor<FunctionsHostingConfigOptions>` to get the latest values immediately if they need them.
  3. Calls [`config.Bind()`](https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Options.ConfigurationExtensions/src/NamedConfigureFromConfigurationOptions.cs#L38) via the `NamedConfigureFromConfigurationOptions` class.

We ultimately didn't want the call to `Bind()`, but did want the other two. So I decomposed this in my change above. The only difference now is we do not bind (the `ChangeTokenSource` was always reigistered and we had tests that failed before I re-added it).

### Making properties internal vs removing setters
I made the properties on `FunctionsHostingConfigOptions` internal rather than just removing the setters because even without setters, we were still getting a `FormatException`. This is because internally the call to `Bind()` tries to perform the conversion *before* it checks whether the property is read-only. It's possible this has been fixed in .NET 7.0+ (see here -- https://github.com/dotnet/runtime/pull/66723) but I have not validated this. Making the properties internal makes this safer for our scenario, so I'm going with that.

PR #9898, opened 2024-02-24
Updating SyncTriggers to include host config

mathewc:
Needed for Flex Consumption scaling. When there's an active host running during a sync triggers operation, we want to include host config info so it's persisted by the platform and accessible to other platform components like ScaleController.

Because the sync triggers payload will be different based on whether there's an active host or not, this means that the payload can oscillate from call to call.  Changes are being made in the FrontEnd to handle this and those changes must be complete before these host changes are active. The FE will correctly handle the presence/absence of the hostConfig portion of the payload by doing a merge operation with any existing sync value it has stored. 

Because the payload can oscillate based on presence of hostConfig, I was also thinking initially that for the background sync triggers hash check, we'd need to track 2 different hash values - one for an "active" sync where the hostConfig is present, and one for an "inactive" sync where hostConfig is not present. However, I don't think we need to make any changes there. A background sync is only ever done when there's an active host running (code [here](https://github.com/Azure/azure-functions-host/blob/937267a5d78b006456446be62ecb120d10c0208d/src/WebJobs.Script.WebHost/FunctionsSyncService.cs#L41)), so it should never (or very rarely in race condition cases) happen that we might make an extraneous call due to oscillation based on active/inactive host.

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.WebHost/Controllers/HostController.cs
+++ b/src/WebJobs.Script.WebHost/Controllers/HostController.cs
@@ -16,8 +16,10 @@ using Microsoft.AspNetCore.Mvc.WebApiCompatShim;
 using Microsoft.Azure.WebJobs.Host;
 using Microsoft.Azure.WebJobs.Host.Executors;
 using Microsoft.Azure.WebJobs.Host.Scale;
+using Microsoft.Azure.WebJobs.Script.Diagnostics;
 using Microsoft.Azure.WebJobs.Script.ExtensionBundle;
 using Microsoft.Azure.WebJobs.Script.Scale;
+using Microsoft.Azure.WebJobs.Script.WebHost.Extensions;
 using Microsoft.Azure.WebJobs.Script.WebHost.Filters;
 using Microsoft.Azure.WebJobs.Script.WebHost.Management;
 using Microsoft.Azure.WebJobs.Script.WebHost.Models;
@@ -44,6 +46,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Controllers
         private readonly IScriptHostManager _scriptHostManager;
         private readonly IFunctionsSyncManager _functionsSyncManager;
         private readonly HostPerformanceManager _performanceManager;
+        private readonly IMetricsLogger _metricsLogger;
         private static readonly SemaphoreSlim _drainSemaphore = new SemaphoreSlim(1, 1);
         private static readonly SemaphoreSlim _resumeSemaphore = new SemaphoreSlim(1, 1);
 
@@ -52,7 +55,8 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Controllers
             IEnvironment environment,
             IScriptHostManager scriptHostManager,
             IFunctionsSyncManager functionsSyncManager,
-            HostPerformanceManager performanceManager)
+            HostPerformanceManager performanceManager,
+            IMetricsLogger metricsLogger)
         {
             _applicationHostOptions = applicationHostOptions;
             _logger = loggerFactory.CreateLogger(ScriptConstants.LogCategoryHostController);
@@ -60,6 +64,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Controllers
             _scriptHostManager = scriptHostManager;
             _functionsSyncManager = functionsSyncManager;
             _performanceManager = performanceManager;
+            _metricsLogger = metricsLogger;
         }
 
         [HttpGet]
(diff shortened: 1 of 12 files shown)


src/WebJobs.Script/Config/HostJsonFileConfigurationSource.cs
```diff
@@ -283,73 +284,6 @@ private JObject TryAddBundleConfiguration(JObject content, string bundleId, stri
                 }
                 return content;
             }
-
-            internal static string SanitizeHostJson(JObject hostJsonObject)
```
mathewc, line 287: I needed a JObject recursive sanitizer, so I generalized the below code since only a single portion of it was specific to host.json (the list of properties to sanitize).

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -388,6 +392,21 @@ public IActionResult LaunchDebugger()
         [Authorize(Policy = PolicyNames.AdminAuthLevelOrInternal)]
         public async Task<IActionResult> SyncTriggers()
         {
+            _metricsLogger.LogEvent(MetricEventNames.SyncTriggersInvoked);
```
mathewc, line 396: These two new metrics will allow us to calculate the percentage of externally invoked sync operations that are made when the host is in a running state.

src/WebJobs.Script.WebHost/Extensions/ScriptHostManagerExtensions.cs
```diff
@@ -0,0 +1,67 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc.Filters;
+using Microsoft.Azure.WebJobs.Script.WebHost;
+using Microsoft.Azure.WebJobs.Script.WebHost.Extensions;
+
+namespace Microsoft.Azure.WebJobs.Script
+{
+    public static class ScriptHostManagerExtensions
+    {
+        public static async Task WaitForRunningHostAsync(this IScriptHostManager hostManager, ScriptApplicationHostOptions applicationHostOptions, HttpContext httpContext, int timeoutSeconds = ScriptConstants.HostTimeoutSeconds, int pollingIntervalMilliseconds = ScriptConstants.HostPollingIntervalMilliseconds, ActionExecutionDelegate next = null)
```
mathewc: I factored this "wait for running host" logic out of the RequiresRunningHost attribute so it could be reused. To do so, I had to move the ScriptHostManagerExtensions helper to WebHost level since it operates on WebHost only types.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -313,9 +317,22 @@ public async Task<SyncTriggersPayload> GetSyncTriggersPayload()
             var triggersArray = new JArray(triggers);
             int count = triggersArray.Count;
 
+            JObject hostConfig = null;
+            if (_environment.IsFlexConsumptionSku())
```
mathewc, line 320: To prevent sync payload vacillation, we can only include the config section when we know there'll ALWAYS be a running host for syncs performed on the app.

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -388,6 +392,21 @@ public IActionResult LaunchDebugger()
...
+
+            // TEMP: Collecting temporary metrics on the host state during SyncTrigger requests
+            if (!_scriptHostManager.HostIsInitialized())
+            {
+                _metricsLogger.LogEvent(MetricEventNames.SyncTriggersHostNotInitialized);
+            }
+
+            // TODO: We plan on making SyncTriggers RequiresRunningHost across the board,
+            // but for now only for Flex.
+            if (_environment.IsFlexConsumptionSku())
```
fabiocav, line 407: Can we create an issue to track this?

src/WebJobs.Script.WebHost/Extensions/ScriptHostManagerExtensions.cs
fabiocav: The code here was previously written to work in the HTTP pipeline, so it throws an `HttpException` expecting that to be handled by the middleware. Just want to make sure we're thinking about how other code could be calling this and what that experience will look like (even if we still have a dependency on `HttpContext` here)

I'm also not sure copying the code over from the action filter attribute works when exposing the logic as an API. There, it was all in the attribute logic, so doing something like setting the status code and response information was clear. When moved to an API named `WaitForRunningHostAsync`, that behavior might be surprising to callers. Would be good to consider potential refactoring options to improve clarity and avoid misuse of the API. That would also improve reusability, as other code would be able to use it in different contexts.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -63,10 +65,11 @@ public class FunctionsSyncManager : IFunctionsSyncManager, IDisposable
...
 
         private BlobClient _hashBlobClient;
 
-        public FunctionsSyncManager(IHostIdProvider hostIdProvider, IOptionsMonitor<ScriptApplicationHostOptions> applicationHostOptions, ILogger<FunctionsSyncManager> logger, IHttpClientFactory httpClientFactory, ISecretManagerProvider secretManagerProvider, IScriptWebHostEnvironment webHostEnvironment, IEnvironment environment, HostNameProvider hostNameProvider, IFunctionMetadataManager functionMetadataManager, IAzureBlobStorageProvider azureBlobStorageProvider, IOptions<FunctionsHostingConfigOptions> functionsHostingConfigOptions)
+        public FunctionsSyncManager(IHostIdProvider hostIdProvider, IOptionsMonitor<ScriptApplicationHostOptions> applicationHostOptions, ILogger<FunctionsSyncManager> logger, IHttpClientFactory httpClientFactory, ISecretManagerProvider secretManagerProvider, IScriptWebHostEnvironment webHostEnvironment, IEnvironment environment, HostNameProvider hostNameProvider, IFunctionMetadataManager functionMetadataManager, IAzureBlobStorageProvider azureBlobStorageProvider, IOptions<FunctionsHostingConfigOptions> functionsHostingConfigOptions, IServiceProvider serviceProvider)
```
fabiocav: Can we take a dependency on `IScriptHostManager` here instead? This avoids confusion about which service provider is used by the type (App or Job Host levels) and makes the service dependency clearer in tests.

brettsam (approved):

src/WebJobs.Script.WebHost/Controllers/HostController.cs
fabiocav: This will be really helpful!

src/WebJobs.Script.WebHost/Controllers/HostController.cs
mathewc: Done

fabiocav (approved):

PR #9912, opened 2024-03-07
Logging function name from FunctionException

### LINKED ISSUE #5171: Listener errors should have function name in system logs
Had a customer case with the following error:

`The listener for function 'Functions.foo' was unable to start.`

The function name column was not populated. Our detector tooling was unable to show these function specific errors because the column is not populated.

Observed this behavior on host version `2.0.12808.0`.

cjaliaga:
This PR covers the scenario where as part of a listener start failure, FunctionListenerException is thrown. The function name is extracted from the exception, where WebJobs is setting the ShortName: https://github.com/Azure/azure-webjobs-sdk/blob/f1d6728283ce5a93f6f81858a9e1c81145742ad6/src/Microsoft.Azure.WebJobs.Host/Listeners/FunctionListener.cs#L73

resolves #5171

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.WebHost/Diagnostics/SystemLogger.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/SystemLogger.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Azure.WebJobs.Host;
 using Microsoft.Azure.WebJobs.Host.Indexers;
+using Microsoft.Azure.WebJobs.Host.Listeners;
 using Microsoft.Azure.WebJobs.Logging;
 using Microsoft.Azure.WebJobs.Script.Configuration;
 using Microsoft.Azure.WebJobs.Script.Eventing;
@@ -175,6 +176,10 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
                 if (string.IsNullOrEmpty(functionName) && exception is FunctionInvocationException fex)
                 {
                     functionName = string.IsNullOrEmpty(fex.MethodName) ? string.Empty : fex.MethodName.Replace("Host.Functions.", string.Empty);
+                }
+                else if (string.IsNullOrEmpty(functionName) && exception is FunctionListenerException flex)
+                {
+                    functionName = string.IsNullOrEmpty(flex.MethodName) ? string.Empty : Utility.GetFunctionShortName(flex.MethodName);
                 }
 
                 (innerExceptionType, innerExceptionMessage, details) = exception.GetExceptionDetails();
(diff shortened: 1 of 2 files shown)


src/WebJobs.Script.WebHost/Diagnostics/SystemLogger.cs
```diff
@@ -176,6 +177,10 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
                 {
                     functionName = string.IsNullOrEmpty(fex.MethodName) ? string.Empty : fex.MethodName.Replace("Host.Functions.", string.Empty);
                 }
+                else if (string.IsNullOrEmpty(functionName) && exception is FunctionListenerException flex)
```
mathewc: Let's generalize this to the base exception FunctionException. Currently both FunctionListenerException and FunctionIndexingException are subclasses, so we'll get your improvements for more cases :)

test/WebJobs.Script.Tests/Eventing/SystemLoggerTests.cs
```diff
@@ -149,6 +150,27 @@ public void Log_Error_EmitsExpectedEvent()
...
+        [InlineData("TestFunction")]
+        public void Log_ListenerError_EmitsExpectedEvent(string functionName)
+        {
+            string eventName = string.Empty;
+            string message = $"The listener for function '{functionName}' was unable to start.";
+            string functionInvocationId = string.Empty;
+            string activityId = string.Empty;
+
+            Exception innerException = new Exception("Kaboom");
+            Exception ex = new FunctionListenerException(functionName, innerException);
```
mathewc: Can then add a FunctionIndexingException as well, or test with your own subclass, etc.

src/WebJobs.Script.WebHost/Diagnostics/SystemLogger.cs
cjaliaga: That was the initial idea but then I've found that `MethodName` in the `FunctionListenerException` could come as `Functions.{functionName}` even if in the WebJobs Host we are using the `FunctionDescriptor.ShortName`. Then in `FunctionInvocationException` we're removing _Host.Functions_ but not using the short name. Since I didn't find a common normalization for both, I went this way.

Please let me know if you have a better idea on this, if you prefer to go with the base exception or let the _Functions._ in the FunctionName for these scenarios...

Thanks!

mathewc: So there's actually 3 subclasses of this exception type then. I think we should just handle the base exception type and use whatever information we have - some information is better than nothing, right? If we really need to normalize the name at the Functions layer we could do that. Is your concern how detectors would deal with the different formats?

cjaliaga: I'm aware of FunctionListenerException and FunctionInvocationException extending FunctionExtension. I was looking to maintain consistency, as in other places we log the short name without _Functions._. I'm good with handling FunctionException here and avoid checking for the individual type.

mathewc (approved):

mathewc:
@brettsam We just need another review on this. Look good to you?

kshyju (approved):

PR #9930, opened 2024-03-13
Introduce new ApplicationHostOptions model for serialization

liliankasem:

The restart API currently returns the entire [ScriptApplicationHostOptions](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Config/ScriptApplicationHostOptions.cs):

https://github.com/Azure/azure-functions-host/blob/485d3831a0878247d5119bd20970615e74632c05/src/WebJobs.Script.WebHost/Controllers/HostController.cs#L399-L406

This includes a service provider property:

`public IServiceProvider RootServiceProvider { get; set; }`

When serialized this causes an exception which leads to an unhandled host exception:

```LOG
Newtonsoft.Json.JsonSerializationException : Self referencing loop detected for property 'serviceProvider' with type 'Microsoft.Extensions.DependencyInjection.ServiceLookup.ServiceProviderEngineScope'. Path 'rootServiceProvider'.    at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.CheckForCircularReference(JsonWriter writer,Object value,JsonProperty property,JsonContract contract,JsonContainerContract containerContract,JsonProperty containerProperty)
```

To fix, we are introducing a model so we have a strict contract on what exactly is returned when the `admin/host/restart` API is called - this also ensures we avoid future breaking changes as additional properties will have to be **intentionally** added to the new model that we use, and the restart API updated to send the new property over.

To our knowledge, the platform is not using the restart API response at all, so we are considering in future to remove the response all together, but we cannot be sure that this is not a breaking change for someone somewhere depending on this API, therefore we will respond with every options property except for the `IServiceProvider`.

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script/Config/ApplicationHostOptions.cs
+++ b/src/WebJobs.Script/Config/ApplicationHostOptions.cs
@@ -3,11 +3,14 @@
 
 using System;
 
-// TODO: documentation. This is a copy of the ScriptApplicationHostOptions class from the Microsoft.Azure.WebJobs.Script namespace.
-// This is returned when admin/host/restart is called
+/// <summary>
+/// This is a copy of the <see cref="ScriptApplicationHostOptions"/> excluding the <see cref="IServiceProvider"/> property.
+/// The purpose of this class is to provide a subset of the <see cref="ScriptApplicationHostOptions"/> that can be serialized
+/// when used in API responses.
+/// </summary>
 namespace Microsoft.Azure.WebJobs.Script
 {
-    public class ApplicationHostOptions // TODO: name this more specifically to what the restart API needs?
+    public class ApplicationHostOptions
     {
         public bool IsSelfHost { get; set; }
 
@@ -21,7 +24,7 @@ namespace Microsoft.Azure.WebJobs.Script
 
         public bool HasParentScope { get; set; }
 
-        public bool IsStandbyConfiguration { get; internal set; }
+        public bool IsStandbyConfiguration { get; set; }
 
         public bool IsFileSystemReadOnly { get; set; }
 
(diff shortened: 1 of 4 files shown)


src/WebJobs.Script/Config/ApplicationHostOptions.cs
```diff
@@ -0,0 +1,33 @@
...
+using System;
+
+/// <summary>
+/// This is a copy of the <see cref="ScriptApplicationHostOptions"/> excluding the <see cref="IServiceProvider"/> property.
+/// The purpose of this class is to provide a subset of the <see cref="ScriptApplicationHostOptions"/> that can be serialized
+/// when used in API responses.
+/// </summary>
+namespace Microsoft.Azure.WebJobs.Script
+{
+    public class ApplicationHostOptions
```
liliankasem: Open to better naming suggestions here: @fabiocav @kshyju

fabiocav: It would be helpful to avoid the `Options` suffix here to avoid confusing this with the other types used with the Options pattern.

Even something very specific like `HostRestartResponse`, `ApplicationHostModel`, etc. Anything without the `Options` suffix works

liliankasem: I'll go with `ApplicationHostModel` - I don't want to use HostRestartResponse as this could apply to any API that wants to respond with this information

fabiocav: > I don't want to use HostRestartResponse as this could apply to any API that wants to respond with this information.

Do we have use cases? I'm good with the name you've picked, but you can also make the argument that this response could include more than the information contained in `ApplicationHostOptions`.

Approving as this is not a blocker.

fabiocav (approved):

kshyju (approved):

src/WebJobs.Script/Config/ApplicationHostOptions.cs
liliankasem: That's a fair point; no use cases I am aware of - just future proofing.

PR #9934, opened 2024-03-16
Start metrics listener after specialization

liliankasem:

A bug was found where it appears that metrics are being captured pre-specialization and being persisted, leading to an inflated started invocation count e.g:

`'AppFailureCount':0,'ActiveInvocationCount':0,'StartedInvocationCount':123` 

With this change, we are waiting for specialization to happen before we collect metrics. I am deploying these changes to test them out on a legion stamp.

### Pull request checklist

### Additional information

Additional PR information


src/WebJobs.Script/Metrics/HostMetricsProvider.cs
```diff
@@ -93,33 +104,29 @@ private void AddOrUpdateMetricsCache(string key, long value)
...
             }
-            finally
+
+            IReadOnlyDictionary<string, long> metrics;
+
+            lock (_lock)
             {
-                PurgeMetricsCache();
+                metrics = new Dictionary<string, long>(_metricsCache);
+                _metricsCache.Clear();
```
liliankasem, line 119: @fabiocav I did look into Interlocked.Exchange but it would have meant having to return a concurrent dictionary instead of an `IReadOnlyDictionary` and I would like to avoid the extra allocation of converting the empty concurrent dict to a read only one. Therefore I went for this approach instead. Thoughts?

PR #9949, opened 2024-03-26
Fixing issues with HTTP request forwarding and adding retries

### LINKED ISSUE #9954: HTTP proxying fails at initialization if worker endpoint is not yet ready
At initialization (specialization or worker launch), when using HTTP proxying, if the worker HTTP endpoint is not fully initialized, the request will never be sent to the worker and there are no logs showing an issue.

The proxying flow should be improved to deal with potential transient failures and ensure any issues are properly logged.

fabiocav:
Resolves #9954 

This PR fixes a few issues identified while investigating support cases and internal reports.

- Addresses a potential race with worker initialization, primarily handled by a retry
- Ensures the forwarding task is evaluated in failed execution scenarios
- Ensures any exceptions captured by the forwarder are thrown/logged, so we have information about underlying failures
- Refactored code in the invocation handling to simplify the method
- A few performance optimizations in that hot path

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
@@ -29,10 +29,8 @@ using Microsoft.Azure.WebJobs.Script.ManagedDependencies;
 using Microsoft.Azure.WebJobs.Script.Workers;
 using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
 using Microsoft.Azure.WebJobs.Script.Workers.SharedMemoryDataTransfer;
-using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
-using Yarp.ReverseProxy.Forwarder;
 using static Microsoft.Azure.WebJobs.Script.Grpc.Messages.RpcLog.Types;
 using FunctionMetadata = Microsoft.Azure.WebJobs.Script.Description.FunctionMetadata;
 using MsgType = Microsoft.Azure.WebJobs.Script.Grpc.Messages.StreamingMessage.ContentOneofCase;
@@ -59,11 +57,13 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
         private readonly ChannelWriter<OutboundGrpcEvent> _outbound;
         private readonly ChannelReader<InboundGrpcEvent> _inbound;
         private readonly IOptions<FunctionsHostingConfigOptions> _hostingConfigOptions;
+        private readonly string _workerInvocationSucccededMetric;
+        private readonly string _workerInvocationFailedMetric;
+        private readonly string _workerId;
         private IDisposable _functionLoadRequestResponseEvent;
         private bool _disposed;
         private bool _disposing;
         private WorkerInitResponse _initMessage;
-        private string _workerId;
         private RpcWorkerChannelState _state;
         private IDictionary<string, Exception> _functionLoadErrors = new Dictionary<string, Exception>();
         private IDictionary<string, Exception> _metadataRequestErrors = new Dictionary<string, Exception>();
@@ -148,6 +148,10 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
 
             _scriptHostManager.ActiveHostChanged += HandleActiveHostChange;
 
+            // Reused metrics strings
+            _workerInvocationSucccededMetric = string.Format(MetricEventNames.WorkerInvokeSucceeded, Id);
+            _workerInvocationFailedMetric = string.Format(MetricEventNames.WorkerInvokeFailed, Id);
+
             LoadScriptJobHostOptions(_scriptHostManager as IServiceProvider);
         }
 
@@ -162,6 +166,8 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
         public RpcWorkerConfig WorkerConfig => _workerConfig;
 
         public IOptions<ScriptJobHostOptions> JobHostOptions => _scriptHostOptions;
+
+        internal bool IsSharedMemoryDataTransferEnabled => _isSharedMemoryDataTransferEnabled;
 
         private void HandleActiveHostChange(object sender, ActiveHostChangedEventArgs e)
         {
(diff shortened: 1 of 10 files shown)


src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -148,6 +148,10 @@ internal partial class GrpcWorkerChannel : IRpcWorkerChannel, IDisposable
 
             _scriptHostManager.ActiveHostChanged += HandleActiveHostChange;
 
+            // Reused metrics strings
+            _workerInvocationSucccededMetric = string.Format(MetricEventNames.WorkerInvokeSucceeded, Id);
```
fabiocav, line 152: Caching these strings as they were recreated with every invocation.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -1088,83 +1090,114 @@ internal async Task InvokeResponse(InvocationResponse invokeResponse)
...
+            catch (Exception exc)
+            {
+                _workerChannelLogger.LogWarning(exc, "Failed to cleanup shared memory resources");
+            }
+        }
+
+        private void LogSharedMemoryUsage(InvocationResponse invokeResponse)
+        {
+            // Return early if debug logging is not enabled.
+            if (!_workerChannelLogger.IsEnabled(LogLevel.Debug)
```
fabiocav: Performing checks now to make sure this doesn't run when not needed.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -1088,83 +1090,114 @@ internal async Task InvokeResponse(InvocationResponse invokeResponse)
...
+            };
+
+            return result;
+        }
+
+        private void TryCloseSharedMemoryResources(InvocationResponse invokeResponse)
+        {
+            try
+            {
+                if (!IsSharedMemoryDataTransferEnabled())
```
fabiocav: Performing checks now to make sure this doesn't run when not needed.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -1088,83 +1090,114 @@ internal async Task InvokeResponse(InvocationResponse invokeResponse)
...
+                }
+                catch (Exception exc)
+                {
+                    context.ResultSource.TrySetException(exc);
+                }
+                finally
+                {
+                    TryCloseSharedMemoryResources(invokeResponse);
 
                     invocation.Dispose();
```
fabiocav, line 1127: Fixing this issue, moving to the finally block here to ensure this is executed regardless of the result.

satvu (approved):

src/WebJobs.Script.Grpc/Server/RetryProxyHandler.cs
```diff
@@ -0,0 +1,50 @@
...
+
+                    _logger.LogWarning("Failed to proxy request to the worker. Retrying in {delay}ms. Attempt {attemptCount} of {maxRetries}.",
+                        currentDelay, attemptCount, _maxRetries);
+
+                    await Task.Delay(currentDelay * attemptCount, cancellationToken);
+                }
+            }
+
+            // This should never be reached.
+            throw null;
```
kshyju, line 47: If all attempts fails, 47 will be executed. In that case, should we throw a more appropriate exception?

src/WebJobs.Script.Grpc/Server/IHttpProxyService.cs
```diff
@@ -10,6 +10,8 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
 {
     public interface IHttpProxyService
     {
-        ValueTask<ForwarderError> ForwardAsync(ScriptInvocationContext context, Uri httpUri);
+        Task EnsureSucessfulForwardingAsync(ScriptInvocationContext context);
```
kshyju: Typo.

src/WebJobs.Script.Grpc/Server/RetryProxyHandler.cs
```diff
@@ -0,0 +1,50 @@
...
+                    return await base.SendAsync(request, cancellationToken);
+                }
+                catch (HttpRequestException) when (attemptCount < _maxRetries)
+                {
+                    currentDelay *= attemptCount;
+
+                    _logger.LogWarning("Failed to proxy request to the worker. Retrying in {delay}ms. Attempt {attemptCount} of {maxRetries}.",
+                        currentDelay, attemptCount, _maxRetries);
+
+                    await Task.Delay(currentDelay * attemptCount, cancellationToken);
```
kshyju: Did you mean `await Task.Delay(currentDelay , cancellationToken);` ?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
kshyju: Do we need to execute this method for every invocation response? Looks like it does some try parse + logging inside it. Could we use `_isSharedMemoryDataTransferEnabled` variable?

src/WebJobs.Script.Grpc/Server/RetryProxyHandler.cs
fabiocav: If the last attempt fails, the `catch` block condition will prevent the exception from being caught, and the exception that triggers that failure will bubble up, so this code should never be reached. 

One of two things will happen:
1. The result will be returned if successful within the retry window
2. The exception that occurs in the last retry it will happen and bubble up from this method

src/WebJobs.Script.Grpc/Server/RetryProxyHandler.cs
fabiocav: Yes! Good catch. Missed this after a last-minute refactoring to add the log.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
fabiocav: Addressed this. Checking for the variable now, but also updated the method to make that API a bit clearer and avoid having it used the same way again. Also removed state dependency there.

src/WebJobs.Script.Grpc/Server/RetryProxyHandler.cs
fabiocav: Flagging as resolved, but please let me know if you have any questions/concerns

kshyju: Yea, that makes sense. We need to keep the throw statement to keep the compiler happy.

kshyju (approved):
release notes update missing.

PR #9957, opened 2024-03-28
Update host metrics to get function target group after specialization

liliankasem:

Bug found where the FunctionGroup being emitted by metrics is empty. This is because the FUNCTIONS_TARGET_GROUP environment variable is only set after specialization. The changes in this PR are two parts:

1. Within HostMetrics, instead of making FunctionGroup a global tag, it is now a tag added to each metric. The reason for this is that we cannot update a tag list once the meter is created. The only other alternative here is to have HostMetrics.cs listen for specialization which would require moving this class to the WebHost project amongst other changes - this doesn't feel right to me but I am open to feedback on it.

2. Within the HostMetricsProducer, instead of getting the FunctionGroup value from the meter tag list (which is created pre-specialization), we now just get this value directly from the environment. The producer is initialized after specialization so this ensures we have the updated post-specialization value for the function group. Eventually we will remove the producer in favour of using an OTel exporter, so I think this is the best change for now.

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script/Metrics/HostMetrics.cs
+++ b/src/WebJobs.Script/Metrics/HostMetrics.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.Diagnostics.Metrics;
 
--- a/src/WebJobs.Script/Metrics/HostMetrics.cs
+++ b/src/WebJobs.Script/Metrics/HostMetrics.cs
@@ -9,6 +9,8 @@ namespace Microsoft.Azure.WebJobs.Script.Metrics
 {
     public class HostMetrics : IHostMetrics
     {
+        private readonly IEnvironment _environment;
+
         public const string MeterName = "Microsoft.Azure.WebJobs.Script.Host.Internal";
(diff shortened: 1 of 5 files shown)


jviau (approved):

src/WebJobs.Script/Metrics/HostMetrics.cs
```diff
@@ -57,8 +57,34 @@ public HostMetrics(IMeterFactory meterFactory, IEnvironment environment)
...
+            {
+                if (_cachedFunctionGroupTag != null)
+                {
+                    return _cachedFunctionGroupTag.Value;
+                }
+
+                var functionGroup = _environment.GetEnvironmentVariableOrDefault(EnvironmentSettingNames.FunctionsTargetGroup, string.Empty);
+                var functionGroupTag = new KeyValuePair<string, object>(TelemetryAttributes.AzureFunctionsGroup, functionGroup);
+
+                if (!string.IsNullOrEmpty(functionGroup))
```
fabiocav, line 75: Should we log when we're unable to resolve a function group? Having telemetry here would really help if this is missing when expected to be present.

liliankasem: Done!

fabiocav (approved):

PR #9959, opened 2024-03-29
Perf optimizations with function ID resolution

fabiocav:
Perf optimizations when resolving function IDs.

For the method optimization alone, these are the high level results:

Case	ResultsGraph	Mean	Min	Max	Range	    AllocatedBytes	Operations	Phase
TestNew		14.90 ns	14.58 ns	15.31 ns	5%	1,544	503,316,480	Complete
TestOld		29.29 ns	28.85 ns	30.14 ns	4%	1,652	234,881,024	Complete

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
@@ -585,10 +585,11 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
             _functions = functions;
             foreach (FunctionMetadata metadata in functions)
             {
-                _workerChannelLogger.LogDebug("Setting up FunctionInvocationBuffer for function: '{functionName}' with functionId: '{functionId}'", metadata.Name, metadata.GetFunctionId());
-                _functionInputBuffers[metadata.GetFunctionId()] = new BufferBlock<ScriptInvocationContext>();
+                string functionId = metadata.GetFunctionId();
+                _workerChannelLogger.LogDebug("Setting up FunctionInvocationBuffer for function: '{functionName}' with functionId: '{functionId}'", metadata.Name, functionId);
+                _functionInputBuffers[functionId] = new BufferBlock<ScriptInvocationContext>();
             }
-            _state = _state | RpcWorkerChannelState.InvocationBuffersInitialized;
+            _state |= RpcWorkerChannelState.InvocationBuffersInitialized;
         }
 
         public void SendFunctionLoadRequests(ManagedDependencyOptions managedDependencyOptions, TimeSpan? functionTimeout)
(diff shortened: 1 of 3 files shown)


src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -741,7 +742,11 @@ internal FunctionEnvironmentReloadRequest GetFunctionEnvironmentReloadRequest(ID
         internal void SendFunctionLoadRequest(FunctionMetadata metadata, ManagedDependencyOptions managedDependencyOptions)
         {
             _functionLoadRequestResponseEvent = _metricsLogger.LatencyEvent(MetricEventNames.FunctionLoadRequestResponse);
-            _workerChannelLogger.LogDebug("Sending FunctionLoadRequest for function: '{functionName}' with functionId: '{functionId}'", metadata.Name, metadata.GetFunctionId());
+
+            if (_workerChannelLogger.IsEnabled(LogLevel.Debug))
```
kshyju, line 746: Is it a valid case where we will disable debug level log?
We have other `LogDebug` calls in the cold start path. Should all of them be fixed?

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -406,10 +406,11 @@ public async Task InvokeAsync(ScriptInvocationContext invocationContext)
             // This could throw if no initialized workers are found. Shut down instance and retry.
             IEnumerable<IRpcWorkerChannel> workerChannels = await GetInitializedWorkerChannelsAsync(invocationContext.FunctionMetadata.Language ?? _workerRuntime);
             var rpcWorkerChannel = _functionDispatcherLoadBalancer.GetLanguageWorkerChannel(workerChannels);
-            if (rpcWorkerChannel.FunctionInputBuffers.TryGetValue(invocationContext.FunctionMetadata.GetFunctionId(), out BufferBlock<ScriptInvocationContext> bufferBlock))
+            string functionId = invocationContext.FunctionMetadata.GetFunctionId();
```
kshyju, line 409: Can you update line 203 to 
```csharp
&& webhostLanguageWorkerChannel.Any(a => a.Value.Task.Status == TaskStatus.RanToCompletion))
```

Replacing Where + Any to just Any.

kshyju: Also 547 to 
```csharp
if (!(await GetInitializedWorkerChannelsAsync()).Any())
```

kshyju (approved):

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
fabiocav: Yes, it is valid. Even enabled by default in hosted environments, running the host in different contexts (e.g., self-hosted, custom containers) should support this.

We should do the same for other calls, but I want to use a better pattern that is both, more performant and cleaner. Will have that as a separate change, but wanted to get this one in.

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
fabiocav: Made those changes in addition to a couple of other improvements I was tracking.

liliankasem (approved):

PR #9970, opened 2024-04-03
Flex placeholder fix - always treat flex as zip deployment

paulbatum:
Recent changes in the flex deployment codepath have removed the use of the SCM_RUN_FROM_PACKAGE setting. This causes the host to go down a codepath where it thinks the app content is mutable and avoid using the language worker placeholder. This impact cold start because it has to start a new language worker from scratch.

The fix takes advantage of the fact that app content is always readonly in the context of flex. 

I included a small logging improvement that should help us understand when we are specifically failing to use the placeholder channel for this reason.

### DIFF
--- a/src/WebJobs.Script.WebHost/Configuration/ScriptApplicationHostOptionsSetup.cs
+++ b/src/WebJobs.Script.WebHost/Configuration/ScriptApplicationHostOptionsSetup.cs
@@ -71,6 +71,14 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Configuration
 
         private bool IsZipDeployment(out bool isScmRunFromPackage)
         {
+            // Flex consumption does not support mutable filesystems, and therefore always counts as zip deployment. It does not use scmRunFromPackage.
+            // TODO: revisit whether this flex consumption path should return false when in placeholder mode? Thats how it appears to work for other SKUs.
+            if (_environment.IsFlexConsumptionSku())
+            {
+                isScmRunFromPackage = false;
+                return true;
+            }
+
             // Check app settings for run from package.
             bool runFromPkgConfigured = Utility.IsValidZipSetting(_environment.GetEnvironmentVariable(AzureWebsiteZipDeployment)) ||
                 Utility.IsValidZipSetting(_environment.GetEnvironmentVariable(AzureWebsiteAltZipDeployment)) ||
--- a/src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
+++ b/src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
@@ -206,7 +206,7 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
                 return versionMatches;
             }
 
-            // Special case: node and PowerShell apps must be read-only to use the placeholder mode channel
+            // Special case: node, python and PowerShell apps must be read-only to use the placeholder mode channel
             // Also cannot use placeholder worker that is targeting ~3 but has backwards compatibility with V2 enabled
             // TODO: Remove special casing when resolving https://github.com/Azure/azure-functions-host/issues/4534
             if (string.Equals(workerRuntime, RpcWorkerConstants.NodeLanguageWorkerName, StringComparison.OrdinalIgnoreCase)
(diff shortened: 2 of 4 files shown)


eliaslopezgt (approved):

src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
```diff
@@ -206,15 +206,22 @@ private bool UsePlaceholderChannel(IRpcWorkerChannel channel)
...
                 || string.Equals(workerRuntime, RpcWorkerConstants.PowerShellLanguageWorkerName, StringComparison.OrdinalIgnoreCase)
                 || string.Equals(workerRuntime, RpcWorkerConstants.PythonLanguageWorkerName, StringComparison.OrdinalIgnoreCase))
             {
                 // Use if readonly and not v2 compatible on ~3 extension
-                return _applicationHostOptions.CurrentValue.IsFileSystemReadOnly && !_environment.IsV2CompatibleOnV3Extension();
+                var isReadOnlyAndNoCompat = _applicationHostOptions.CurrentValue.IsFileSystemReadOnly && !_environment.IsV2CompatibleOnV3Extension();
+
+                if (!isReadOnlyAndNoCompat)
+                {
+                    _logger.LogInformation("App will not use placeholder channel - ReadOnly: {isReadOnly}. NoCompat: {noCompat}.", _applicationHostOptions.CurrentValue.IsFileSystemReadOnly, !_environment.IsV2CompatibleOnV3Extension());
```
pragnagopa: nit: change to LogDebug

src/WebJobs.Script.WebHost/Configuration/ScriptApplicationHostOptionsSetup.cs
```diff
@@ -71,6 +71,14 @@ public void Configure(string name, ScriptApplicationHostOptions options)
 
         private bool IsZipDeployment(out bool isScmRunFromPackage)
         {
+            // Flex consumption does not support mutable filesystems, and therefore always counts as zip deployment. It does not use scmRunFromPackage.
+            // TODO: revisit whether this flex consumption path should return false when in placeholder mode? Thats how it appears to work for other SKUs.
```
fabiocav, line 75: Can we have an issue tracking this item?

src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
fabiocav: Thank you! :)

fabiocav (approved):

src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
paulbatum: Any reason why? Given that this is impacting behavior (not using channel), I think information is correct.

src/WebJobs.Script.WebHost/Configuration/ScriptApplicationHostOptionsSetup.cs
paulbatum: https://github.com/Azure/azure-functions-host/issues/9971

src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
pragnagopa: For Kusto debug is default level; LogInformation, is default level for app insights is information; My recommendation anything "System Logs" that does not have useful information for Cx, should be debug

pragnagopa:
please add unit tests here -https://github.com/Azure/azure-functions-host/blob/dev/test/WebJobs.Script.Tests/Configuration/ScriptApplicationHostOptionsSetupTests.cs if possible

kshyju:
@paulbatum I pushed an update to bump the minor version number to match our tag

paulbatum:
> please add unit tests here -https://github.com/Azure/azure-functions-host/blob/dev/test/WebJobs.Script.Tests/Configuration/ScriptApplicationHostOptionsSetupTests.cs if possible

Thanks for the pointer! I failed to find this because I went looking for code that calls IsZipDeployment and didn't find any. I see now the tests call it indirectly. I will add.

src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
paulbatum: Done.

paulbatum:
Added some basic test coverage in `ScriptApplicationHostOptionsSetupTests`. 

I also looked at covering this case in `WebHostRpcWorkerChannelManagerTests` that call `UsePlaceholderChannel` but those tests have an explicit options monitor and don't end up calling IsZipDeployment, so adding a case there doesnt seem to help.

pragnagopa (approved):

PR #9976, opened 2024-04-04
Adjust Metrics Granularity for On-Demand

michaelpeng36:

Based on further discussion, we want to adjust the granularity of ExecutionTimeMS for OnDemand. **This PR does not adjust granularity for the TotalTimeMS metric for AlwaysReady.**

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
+++ b/src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
@@ -288,7 +288,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Metrics
                     {
                         // Accumulate the duration for this interval, applying the minimum
                         var duration = Math.Max(elapsedMS, _options.MinimumActivityIntervalMS);
-                        FunctionExecutionTimeMS += (long)duration;
+                        FunctionExecutionTimeMS += RoundMS((long)duration, _options.MetricsGranularityMS);
 
                         // Move the high watermark timestamp forward
                         _activityIntervalHighWatermark = adjustedActivityIntervalStart.AddMilliseconds(duration);
@@ -303,6 +303,13 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Metrics
         public void AddFunctionExecutionActivity(string functionName, string invocationId, int concurrency, string executionStage, bool success, long executionTimeSpan, string executionId, DateTime eventTimeStamp, DateTime functionStartTime)
         {
             // nothing to do here - we only care about Started/Completed events.
+        }
+
+        // Rounds the given metric to the specified granularity in the options. This function should only be used for metrics expressed in ms.
+        private long RoundMS(long metricMS, int granularity)
+        {
+            var excess = metricMS % granularity;
+            return metricMS % granularity == 0 ? metricMS : metricMS - excess + granularity;
         }
 
         public void Dispose()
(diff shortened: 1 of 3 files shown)


src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
```diff
@@ -305,6 +305,13 @@ public void AddFunctionExecutionActivity(string functionName, string invocationI
             // nothing to do here - we only care about Started/Completed events.
         }
 
+        // Rounds the given metric to the specified granularity in the options. This function should only be used for metrics expressed in ms.
+        private long RoundMS(long metricMS, int granularity)
+        {
+            var excess = metricMS % granularity;
```
paulbatum: I found this calculation involving excess and the ternary operator to be slightly unintuitive.

Do you think this version is any clearer?

`return Math.Ceiling(metricMS / granularity) * granularity;` 

Also maybe update the method name to be clearer that the intent is to always round up?

paulbatum (approved):
The updated examples look correct to me. Left a small suggestion but otherwise good.

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
michaelpeng36: I think we'll have to cast `metricMS` to be a double before applying `Math.Ceiling`, but I think your version is clearer in intent. I'll update and adjust the method name as well

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
```diff
@@ -305,6 +305,12 @@ public void AddFunctionExecutionActivity(string functionName, string invocationI
             // nothing to do here - we only care about Started/Completed events.
         }
 
+        // Rounds up the given metric to a specified granularity. For example, RoundUp(1320.00, 100) = 1400, but RoundUp(1300.00, 100) = 1300.
+        private long RoundUp(double metric, int granularity)
```
fabiocav: nit: make this static.

fabiocav (commented):
Can you please look at the failing tests?

fabiocav (approved):
Approving. As long as the tests are happy, good to merge.

michaelpeng36:
@fabiocav, I think the tests that failed are actually inherently flaky since in the unit test, we are including the time it takes between consecutive `OnPublishMetrics` invocations in the `metrics.TotalTimeMS`. Essentially, we are assuming the below code
```
FileInfo[] files = GetMetricsFilesSafe(_metricsFilePath);

FlexConsumptionMetricsPublisher.Metrics metrics = null;
FileInfo metricsFile = null;
if (!isAlwaysReadyInstance)
{
    // don't expect a file to be written when no activity
    Assert.Equal(0, files.Length);
}
else
{
    Assert.Equal(1, files.Length);

    metricsFile = files[0];
    metrics = await ReadMetricsAsync(metricsFile.FullName);
    Assert.True(metrics.IsAlwaysReady);
    ValidateTotalTime(metrics.TotalTimeMS, delay);
    Assert.Equal(0, metrics.ExecutionCount);
    Assert.Equal(0, metrics.ExecutionTimeMS);

    metricsFile.Delete();
}

int executionDurationMS = 5700;
publisher.FunctionExecutionCount = 123;
publisher.FunctionExecutionTimeMS = executionDurationMS;

delay = (int)executionDurationMS + 100;
```
takes under 150 ms, so I think CI just had a really bad run.

PR #9981, opened 2024-04-05
Avoid using HttpRequest from log scope (#9979) (#9980)

jviau:
Cherry-pick of https://github.com/Azure/azure-functions-host/pull/9980


VpOfEngineering (changes requested):
can we increase the patch version in `build/common.props` too? Thanks 🫡

VpOfEngineering (approved):

PR #9985, opened 2024-04-10
OpenTelemetry support

### LINKED ISSUE #9273: Azure Functions – Observability
Azure Function has robust integration with Application Insights to monitor Functions execution. Along with Application Insights, Function also supports integration with other APM systems through Azure Monitor integration. Logs and metrics are also sent to Geneva/kusto for debugging/diagnostic and billing purposes. These integrations allow developers to gain insights into the performance and behavior of their functions, as well as diagnose issues and troubleshoot problems.

Logs, Traces and Metrics are primarily consumed by:
- Microsoft - supportability and billing purpose (only platform logs and metrics).
- Customers 

image

In Azure Functions, the host process is responsible for capturing and processing all logs, including logs generated by the worker process. The host process then sends these logs to all configured sinks, which can include Application Insights, Azure Monitor, File, Kusto and Log Streaming.
The host process in Azure Functions captures logs from several sources, including:
- Platform logs (Host): These are logs generated by the Azure Functions host process itself, which provide information about the startup and shutdown of the host process, as well as any errors or warnings encountered during execution.
- Platform logs (worker): These are logs generated by the worker process.
- Function logs: These are logs generated by the customer code that runs in the worker process. These logs can be captured using the logging framework provided by the language runtime, such as the ILogger interface in .NET.
- Dependency logs: These are logs generated by dependencies used by the function code, such as database drivers, HTTP clients, or messaging libraries.

## Challenges and Gaps
- While Azure Functions offers built-in integration with Application Insights for monitoring and logging, this can make it difficult to use other third-party Application Performance Management (APM) providers. This lack of flexibility can be challenging for organizations that have already invested in other APM solutions. 
- The host process in Azure Functions is responsible for handling all logs, including logs from the worker process, and forwarding them to the appropriate sinks. This can limit the flexibility and control that developers have over logging and tracing in their applications.
- Some of the logs/traces are emitted by the Host e.g.  Trigger and binding related traces, linked activities.
- Azure Functions does not currently offer native support for OpenTelemetry, a popular observability framework that enables developers to collect, analyze, and export telemetry data from their applications. Some of the logs and traces in Azure Functions are emitted by the host process, such as metrics, platform logs and trigger/binding related traces. There is no way for a customer to export these using OpenTelemetry.

## Goals 
- Adopt OpenTelemetry: Azure Functions plans to adopt OpenTelemetry to enable developers to collect, analyze, and export telemetry data from their applications.
- Follow FaaS specification: All the logs, traces, and metrics emitted by Azure Functions will follow the Function-as-a-Service (FaaS) specification, enabling developers to use a variety of observability tools and services.
- Empower customers: Azure Functions will empower customers to take control of their observability by allowing them to define their own configuration and exporter.
- Default option: Azure Functions will provide a default option for telemetry data for customers who are not interested in managing it themselves. The default option will send all the telemetry data to Application Insights. 

## Proposal
 
image

Customers have complete control over the telemetry data from the worker process and can configure the OpenTelemetry Protocol (OTLP) exporter to send this data to any supported endpoint. 
Customer may be interested in some of the telemetry data emitted by the host. They can also choose to export host telemetry data to an OTLP-supported endpoint.
As part of the default option, Azure Functions will send both worker and host telemetry data to Application Insights.

# Step Back

## File system logging 
The initial implementation of file system logging in Azure Functions was done because there was no other option to log/stream data at that time. However, the plan is to stop supporting file system logging because it does not add much value, has significant overhead, and can potentially cause other issues.
Going forward, Application Insights will be the only option to stream logs on the portal and other client tooling.

## Azure monitor Integration
Azure Monitor has both platform logs (generated by the Azure Functions platform) and customer logs (generated by the customer's functions code). This was very helpful as there was no other supported way to export telemetry data to an external ecosystem. 
This requires Microsoft to process and store customer data in a secure manner, which can be a potential security risk. The plan is to stop processing customer logs through Azure monitor integration.
By separating the customer logs from platform logs, the customer will have complete control over their data, including where it is stored and who has access to it. Customers will have the option to configure their Application Insights resource to forward logs to a Log Analytics Workspace. This approach will provide customers with richer experience and will also prevent Microsoft from processing sensitive data.

## Telemetry Initializers and Processors
Azure Functions InProc model does provide out of the box support to enrich/override telemetry with additional information. As both, the Host process and the customer code, are running in the same process any initializers and processers if implemented, applies to the logs and traces emitted in both host and customer code.
In the proposed solution, as the host and worker will have an independent integration with the ApplicationInsights, the host OTLP extension won’t support initializers and processors. This is also true for the [Isolated Application Insights Package](https://github.com/Azure/azure-functions-dotnet-worker/pull/944#issue-1282987627).

RohitRanjanMS:
Introducing OpenTelemetry support for the Host enables users to adopt a method supported by OpenTelemetry for exporting telemetry data. This addition accommodates both existing AI SDK and new OpenTelemetry instrumentations, offering users the option to activate OpenTelemetry by adjusting the telemetry mode.

**Telemetry Mode Configuration:**

By default, the Host process is integrated with Application Insights through the AI SDK, emitting telemetry for requests, metrics, and logs. While logs are produced using Illoger, traces and metrics do not utilize OpenTelemetry's recommended APIs, such as ActivitySource and Meter.

To shift to a mode where the Host outputs telemetry in alignment with OpenTelemetry's standards and APIs, users can modify the telemetry mode within the host.json file or through an environmental variable override as follows:

```
{
  "telemetryMode": "openTelemetry"
}
```
**Export Options:**
Azure Monitor Exporter: Users can export telemetry to Application Insights by adding the APPLICATIONINSIGHTS_CONNECTION_STRING, enabling the Azure Monitor exporters and LiveMetrics.

OTLP Exporter: Facilitates telemetry export to any OTLP-compatible endpoint. Configuration is possible by setting the OTEL_EXPORTER_OTLP_ENDPOINT and OTEL_EXPORTER_OTLP_HEADERS for OTLP endpoints.

This implementation is compatible with all standard OpenTelemetry environment variables, allowing for the configuration of log and trace providers. As the OpenTelemetry [Configuration ](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/file-configuration.md)specification becomes stable, there are plans to support more detailed configuration options, enabling users to fine-tune their telemetry behavior.

resolves #9273

**Dependencies changes**
______________________________________________________________________________________________________
 Microsoft.Azure.WebJobs.Script.Tests.DependencyTests.Verify_DepsJsonChanges
   Source: DependencyTests.cs line 53
   Duration: 605 ms

  Message: 

Previous file: C:\Users\Rohit Ranjan\Source\Repos\azure-functions-host\test\WebJobs.Script.Tests\bin\Debug\net8.0\Microsoft.Azure.WebJobs.Script.WebHost.deps.json
New file:      C:\Users\Rohit Ranjan\Source\Repos\azure-functions-host\src\WebJobs.Script.WebHost\bin\debug\net8.0\Microsoft.Azure.WebJobs.Script.WebHost.deps.json

  Changed:
    - Azure.Core.dll: 1.35.0.0/1.3500.23.45706 -> 1.38.0.0/1.3800.24.12602

  Removed:

  Added:
    - Azure.Monitor.OpenTelemetry.AspNetCore.dll: 1.2.0.0/1.200.24.16203
    - Azure.Monitor.OpenTelemetry.Exporter.dll: 1.3.0.0/1.300.24.10803
    - Azure.Monitor.OpenTelemetry.LiveMetrics.dll: 1.0.0.0/1.0.24.15803
    - OpenTelemetry.Api.dll: 1.0.0.0/1.8.0.1229
    - OpenTelemetry.Api.ProviderBuilderExtensions.dll: 1.0.0.0/1.8.0.1229
    - OpenTelemetry.dll: 1.0.0.0/1.8.0.1229
    - OpenTelemetry.Exporter.Console.dll: 1.0.0.0/1.6.0.1006
    - OpenTelemetry.Exporter.OpenTelemetryProtocol.dll: 1.0.0.0/1.8.0.1229
    - OpenTelemetry.Extensions.Hosting.dll: 1.0.0.0/1.8.0.1229
    - OpenTelemetry.Instrumentation.AspNetCore.dll: 1.0.0.0/1.7.0.1114
    - OpenTelemetry.Instrumentation.Http.dll: 1.0.0.0/1.7.0.1114
    - OpenTelemetry.PersistentStorage.Abstractions.dll: 1.0.0.0/1.0.0.0
    - OpenTelemetry.PersistentStorage.FileSystem.dll: 1.0.0.0/1.0.0.0
    - System.ClientModel.dll: 1.0.0.0/1.0.24.5302

Expected: True
Actual:   False

  Stack Trace: 
DependencyTests.Verify_DepsJsonChanges() line 125
RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)

I will update the documentation later.
### Pull request checklist


src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -1670,29 +1677,56 @@ private void AddSample<T>(List<T> samples, T sample)
...
         private void AddAdditionalTraceContext(MapField<string, string> attributes, ScriptInvocationContext context)
         {
-            // This is only applicable for AI agents running along side worker
+            bool isOtelEnabled = false;
+            if (_scriptHostOptions?.Value.TelemetryMode == TelemetryMode.OpenTelemetry)
+            {
+                isOtelEnabled = true;
+            }
+
+            bool isAIEnabled = false;
```
jviau:
```suggestion
            bool isAIEnabled = _environment.IsApplicationInsightsAgentEnabled();
```

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -1670,29 +1677,56 @@ private void AddSample<T>(List<T> samples, T sample)
 
         private void AddAdditionalTraceContext(MapField<string, string> attributes, ScriptInvocationContext context)
         {
-            // This is only applicable for AI agents running along side worker
+            bool isOtelEnabled = false;
```
jviau:
```suggestion
            bool isOtelEnabled = _scriptHostOptions?.Value.TelemetryMode == TelemetryMode.OpenTelemetry;
```

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -958,6 +962,28 @@ public object GetService(Type serviceType)
...
+            foreach (var tracerProvider in host.Services.GetServices<TracerProvider>())
+            {
+                logger.LogDebug(@"Flushing {providerName} ...", tracerProvider);
+                tracerProvider.ForceFlush();
+            }
+
+            foreach (var logProvider in host.Services.GetServices<ILoggerProvider>().Where(i => i is OpenTelemetryLoggerProvider))
+            {
+                logger.LogDebug(@"Disposing {providerName} ...", logProvider);
+                logProvider.Dispose();
```
jviau, line 983: Is this type castable to the OTel `LoggerProvider` base class? Looks like they have a `ForceFlush()` extension method for that.

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -875,6 +878,7 @@ private async Task Orphan(IHost instance, CancellationToken cancellationToken =
                     instance.Dispose();
                     _logger.LogDebug("ScriptHost disposed");
                 }
+                FlushOpenTelemetry(instance);
```
jviau: If we are disposing the host here, I wonder if all the services will already be disposed and flushed as a part of that?

src/WebJobs.Script/Diagnostics/MetricEventNames.cs
```diff
@@ -9,6 +9,8 @@ public static class MetricEventNames
         public const string ApplicationStartLatency = "host.application.start";
         public const string ApplicationInsightsEnabled = "host.applicationinsights.enabled";
         public const string ApplicationInsightsDisabled = "host.applicationinsights.disabled";
+        public const string OpenTelemetryAzMonEnabled = "host.opentelemetry.azmon.enabled";
```
jviau: What do you think of using `otel`? I am good either way, just want to make sure we discuss our options. Especially considering you are already abbreviating `azmon` and `otlp`.

```suggestion
        public const string OpenTelemetryAzMonEnabled = "host.otel.azmon.enabled";
```

src/WebJobs.Script/Diagnostics/OpenTelemetry/ActivitySanitizingProcessor.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Collections.Immutable;
+using System.Diagnostics;
+using Microsoft.Azure.WebJobs.Logging;
+using OpenTelemetry;
+
+namespace Microsoft.Azure.WebJobs.Script.Diagnostics.OpenTelemetry
+{
+    internal class ActivitySanitizingProcessor : BaseProcessor<Activity>
+    {
+        private static readonly ImmutableArray<string> TagsToSanitize = ImmutableArray.Create("url.query", "url.full");
```
jviau: nit: should `url.query` and `url.full` be from a constants file?

src/WebJobs.Script/Diagnostics/OpenTelemetry/FunctionsResourceDetector.cs
```diff
@@ -0,0 +1,76 @@
...
+        {
+            { ResourceAttributeConstants.CloudRegion, ResourceAttributeConstants.RegionNameEnvVar },
+        };
+
+        public Resource Detect()
+        {
+            List<KeyValuePair<string, object>> attributeList = new(5);
+            try
+            {
+                var siteName = Environment.GetEnvironmentVariable(ResourceAttributeConstants.SiteNameEnvVar);
```
jviau: Technically many of these attributes should only be set when this app is actually running in a functions-hosted serverless environment.

If they are self-hosting or running locally, these values should be different.

src/WebJobs.Script/Diagnostics/OpenTelemetry/FunctionsResourceDetector.cs
```diff
@@ -0,0 +1,76 @@
...
+                attributeList.Add(new KeyValuePair<string, object>(ResourceAttributeConstants.CloudProvider, ResourceAttributeConstants.AzureCloudProviderValue));
+                attributeList.Add(new KeyValuePair<string, object>(ResourceAttributeConstants.CloudPlatform, ResourceAttributeConstants.AzurePlatformValue));
+
+                var version = Assembly.GetEntryAssembly()?.GetName().Version.ToString() ?? "0:0:0";
```
jviau: Should we even emit this attribute if we cannot get the version? And if we do, the formatting is `.` separated, not `:`. Also I think assemblies typically have 4 segments. `0.0.0.0`

src/WebJobs.Script/Diagnostics/OpenTelemetry/FunctionsResourceDetector.cs
```diff
@@ -0,0 +1,76 @@
...
+            return new Resource(attributeList);
+        }
+
+        private static string GetAzureResourceURI(string websiteSiteName)
+        {
+            string websiteResourceGroup = Environment.GetEnvironmentVariable(ResourceAttributeConstants.ResourceGroupEnvVar);
+            string websiteOwnerName = Environment.GetEnvironmentVariable(ResourceAttributeConstants.OwnerNameEnvVar) ?? string.Empty;
+
+#if NET6_0_OR_GREATER
+            int idx = websiteOwnerName.IndexOf('+', StringComparison.Ordinal);
```
jviau, line 60: Why the if-def? These lines appear identical to me.

src/WebJobs.Script/Diagnostics/OpenTelemetry/OpenTelemetryConfigurationExtensions.cs
```diff
@@ -0,0 +1,136 @@
...
+                    if (!string.IsNullOrEmpty(azMonConnectionString))
+                    {
+                        o.AddAzureMonitorLogExporter(options => options.ConnectionString = azMonConnectionString);
+                    }
+                    o.IncludeFormattedMessage = true;
+                    o.IncludeScopes = false;
+                })
+                // These are messages piped back to the host from the worker - we don't handle these anymore if the worker has OpenTelemetry enabled.
+                // Instead, we expect the user's own code to be logging these where they want them to go.
+                .AddFilter<OpenTelemetryLoggerProvider>("Host.*", _ => !ScriptHost.WorkerOpenTelemetryEnabled)
```
jviau: We should verify this, I don't think it is blanket true statement. There may be code in the host which still starts with these category names. `Azure.*` for sure, as the host will use Azure SDKs. `Microsoft.Azure.WebJobs.*` also - those assemblies are all in the host, not the worker.

`Function.*` and `Host.*` we will need to see.

Also, if we are already sending a capability from the worker back to the host, can the worker just opt to _not_ send telemetry back to the host when it emits that capability? We already do this for app insights in dotnet isolated.

src/WebJobs.Script/Diagnostics/OpenTelemetry/OpenTelemetryConfigurationExtensions.cs
```diff
@@ -0,0 +1,136 @@
...
+                    b.AddProcessor(ActivitySanitizingProcessor.Instance);
+                    b.AddProcessor(TraceFilterProcessor.Instance);
+                });
+
+            string eventLogLevel = GetConfigurationValue(EnvironmentSettingNames.OpenTelemetryEventListenerLogLevel, context.Configuration);
+            if (!string.IsNullOrEmpty(eventLogLevel))
+            {
+                if (Enum.TryParse(eventLogLevel, ignoreCase: true, out EventLevel level))
+                {
+                    loggingBuilder.Services.AddSingleton<IHostedService>(serviceProvider => new OpenTelemetryEventListenerService(level));
```
jviau: If you don't need DI, you can pass the instance directly. Also, use `AddHostedService`

```suggestion
                    loggingBuilder.Services.AddHostedService(new OpenTelemetryEventListenerService(level));
```

src/WebJobs.Script/Diagnostics/OpenTelemetry/OpenTelemetryConfigurationExtensions.cs
```diff
@@ -0,0 +1,136 @@
...
+                }
+                else
+                {
+                    throw new InvalidEnumArgumentException($"Invalid `{EnvironmentSettingNames.OpenTelemetryEventListenerLogLevel}`.");
```
jviau: nit:
1. Use single quotes `'` not backtick.
2. Include the value that was present. Maybe even include the possible values.

```suggestion
                    throw new InvalidEnumArgumentException($"Invalid '{EnvironmentSettingNames.OpenTelemetryEventListenerLogLevel}' of '{eventLogLevel}'.");
```

src/WebJobs.Script/Diagnostics/OpenTelemetry/OpenTelemetryConfigurationExtensions.cs
```diff
@@ -0,0 +1,136 @@
...
+                }
+            }
+            else
+            {
+                // Log all warnings and above by default.
+                loggingBuilder.Services.AddSingleton<IHostedService>(serviceProvider => new OpenTelemetryEventListenerService(EventLevel.Warning));
```
jviau:
```suggestion
                loggingBuilder.Services.AddHostedService(new OpenTelemetryEventListenerService(EventLevel.Warning));
```

src/WebJobs.Script/Diagnostics/OpenTelemetry/OpenTelemetryConfigurationExtensions.cs
```diff
@@ -0,0 +1,136 @@
...
+            }
+
+            static ResourceBuilder ConfigureResource(ResourceBuilder r)
+            {
+                string serviceName = Environment.GetEnvironmentVariable(ResourceAttributeConstants.SiteNameEnvVar) ?? "azureFunctions";
```
jviau: I am not sure I like the fallback of `azureFunctions`. Is there something better?

src/WebJobs.Script/Diagnostics/OpenTelemetry/ResourceAttributeConstants.cs
```diff
@@ -0,0 +1,47 @@
...
+        internal const string AttributeName = "faas.name";
+        internal const string AttributeVersion = "faas.version";
+        internal const string AttributeInstance = "faas.instance";
+
+        internal const string AzureCloudProviderValue = "azure";
+        internal const string AzurePlatformValue = "azure_functions";
+        internal const string AttributeSDKPrefix = "ai.sdk.prefix";
+        internal const string AttributeProcessId = "process.pid";
+
+        internal const string SDKPrefix = "azurefunctions";
```
jviau: should these go into a separate file? I'd prefer we keep the telemetry attributes themselves in their own file.

src/WebJobs.Script/Diagnostics/OpenTelemetry/TraceFilterProcessor.cs
```diff
@@ -0,0 +1,82 @@
...
+{
+    internal class TraceFilterProcessor : BaseProcessor<Activity>
+    {
+        private TraceFilterProcessor() { }
+
+        public static TraceFilterProcessor Instance { get; } = new TraceFilterProcessor();
+
+        public override void OnEnd(Activity data)
+        {
+            var dataTags = data.Tags.ToImmutableDictionary();
```
jviau: Can we avoid this call? It looks very expensive to be in a telemetry processor. Avoid allocations as much as possible. I recommend looking into OTel's code, they have some high-performance way to interacting with tags I believe.

src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
```diff
@@ -29,6 +29,9 @@ public static class EnvironmentSettingNames
         public const string AppInsightsAuthenticationString = "APPLICATIONINSIGHTS_AUTHENTICATION_STRING";
         public const string AppInsightsEventListenerLogLevel = "APPLICATIONINSIGHTS_EVENT_LISTENER_LOGLEVEL";
         public const string AppInsightsQuickPulseAuthApiKey = "APPINSIGHTS_QUICKPULSEAUTHAPIKEY";
+        public const string OtlpEndpoint = "OTEL_EXPORTER_OTLP_ENDPOINT";
+        public const string OpenTelemetryEventListenerLogLevel = "OTEL_EVENT_LISTENER_LOGLEVEL";
+        public const string TelemetryMode = "AzureFunctionsJobHost:telemetryMode";
```
jviau: We shouldn't use `:` in env settings, not all platforms allow that. Also, lets choose an existing prefix / naming convention. There are several in this file so we will need to choose one. We also do not use `AzureFunctionsJobHost` anywhere for env variables.

Options:
`FUNCTIONS_HOST_TELEMETRY_MODE`
`AzureWebJobsHostTelemetryMode`

src/WebJobs.Script/Host/ScriptHost.cs
```diff
@@ -458,6 +461,18 @@ private void PreInitialize()
                 _metricsLogger.LogEvent(MetricEventNames.ApplicationInsightsDisabled);
             }
 
+            if (!string.IsNullOrEmpty(_settingsManager.TelemetryMode) && _settingsManager.TelemetryMode.Equals(OpenTelemetryConstants.OpenTelemetry, StringComparison.OrdinalIgnoreCase))
```
jviau: You have an enum for `TelemetryMode` already, can settings manager use that as the type?

src/WebJobs.Script/Diagnostics/OpenTelemetry/TelemetryMode.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Script.Diagnostics.OpenTelemetry
+{
+    internal enum TelemetryMode
+    {
+        ApplicationInsights = 0b0000,
```
jviau: nit, include a `None` or `Default`. option for enums like this (so you can differentiate between enum not configured and app insights explicitly set.

```suggestion
        None = 0, // or Default
        ApplicationInsights = 1,
```

test/WebJobs.Script.Tests/Configuration/ApplicationInsightsConfigurationTests.cs
```diff
@@ -24,7 +24,7 @@ public void ServicesDisabled_InPlaceholderMode()
                 host = new HostBuilder()
                     .ConfigureLogging((context, builder) =>
                     {
-                        ScriptHostBuilderExtensions.ConfigureApplicationInsights(context, builder);
+                        ScriptHostBuilderExtensions.ConfigureApplicationInsights(builder, context);
```
jviau: This is an extension method right?

```suggestion
                        builder.ConfigureApplicationInsights(context);
```

jviau (changes requested):
Customer config approach for preview looks good to me. Primary areas of concern are:

1. Some telemetry semantics.
2. Conditioning the resource detector / azure-specific stuff to only when actually hosted by functions serverless model.
3. Performance impact of the telemetry processors in this PR. Lets do our best to avoid allocations as much as possible in these as they are hot-paths.
  - OTel code may have examples

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -1677,13 +1677,13 @@ private void AddSample<T>(List<T> samples, T sample)
 
         private void AddAdditionalTraceContext(MapField<string, string> attributes, ScriptInvocationContext context)
         {
-            bool isOtelEnabled = false;
+            bool isOtelEnabled = _scriptHostOptions?.Value.TelemetryMode == TelemetryMode.OpenTelemetry;
             if (_scriptHostOptions?.Value.TelemetryMode == TelemetryMode.OpenTelemetry)
```
jviau: Ah sorry, the suggestion is also to remove the lines after this.

jviau (changes requested):

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
RohitRanjanMS: yes, I have removed it in my local changes.

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
RohitRanjanMS: Moved this before dispose.

src/WebJobs.Script/Diagnostics/OpenTelemetry/FunctionsResourceDetector.cs
RohitRanjanMS: removed unnecessary code.

src/WebJobs.Script/Diagnostics/OpenTelemetry/OpenTelemetryConfigurationExtensions.cs
RohitRanjanMS: Removed `Host.*` and `Microsoft.Azure.WebJobs.*`. There's lot of noise coming from `Azure.*`. I will evaluate this again.

We still have to support FIleLogging and DiagnosticsSettings, we need worker logs in the host for this.

src/WebJobs.Script/Diagnostics/OpenTelemetry/OpenTelemetryConfigurationExtensions.cs
RohitRanjanMS: No longer adding serviceName if the value is null.

src/WebJobs.Script/Diagnostics/OpenTelemetry/TraceFilterProcessor.cs
RohitRanjanMS: Removed ToImmutableDictionary.

src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
RohitRanjanMS: No longer need this env variable.

src/WebJobs.Script/Diagnostics/OpenTelemetry/ActivitySanitizingProcessor.cs
```diff
@@ -10,7 +10,7 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics.OpenTelemetry
 {
     internal class ActivitySanitizingProcessor : BaseProcessor<Activity>
     {
-        private static readonly ImmutableArray<string> TagsToSanitize = ImmutableArray.Create("url.query", "url.full");
+        private static readonly ImmutableArray<string> TagsToSanitize = ImmutableArray.Create(OpenTelemetryConstants.QueryUrl, OpenTelemetryConstants.FullUrl);
```
jviau: not a big deal, but immutable collections are overkill for a lot of the scenarios we use them in. A simpler lower cost `IReadOnlyCollection` would be more efficient in this case. But again, not a big deal as this is a 1 time cost as it is static. You could even use just a list or array - it is private after all, this class can trust itself to not modify the collection.

Immutables collections are primarily useful when you need to "modify" the collection, but not have that modification mutate the original collection. Without that modify part of the usage, you are better off with readonly.

src/WebJobs.Script/Diagnostics/OpenTelemetry/FunctionsResourceDetector.cs
```diff
@@ -3,44 +3,44 @@
...
+
+                // Add these attributes only if running in Azure.
+                if (!string.IsNullOrEmpty(serviceName))
                 {
-                    var azureResourceUri = GetAzureResourceURI(siteName);
-                    if (azureResourceUri != null)
+                    attributeList.Add(new KeyValuePair<string, object>(ResourceSemanticConventions.ServiceName, serviceName));
+                    attributeList.Add(new KeyValuePair<string, object>(ResourceSemanticConventions.CloudProvider, OpenTelemetryConstants.AzureCloudProviderValue));
+                    attributeList.Add(new KeyValuePair<string, object>(ResourceSemanticConventions.CloudPlatform, OpenTelemetryConstants.AzurePlatformValue));
+                    attributeList.Add(new KeyValuePair<string, object>(ResourceSemanticConventions.FaaSVersion, version));
```
jviau: FaaS version can be added always, as it is independent of running on Azure or not.

src/WebJobs.Script/Diagnostics/OpenTelemetry/FunctionsResourceDetector.cs
```diff
@@ -3,44 +3,44 @@
...
+                attributeList.Add(new KeyValuePair<string, object>(ResourceSemanticConventions.ServiceVersion, version));
+                attributeList.Add(new KeyValuePair<string, object>(OpenTelemetryConstants.AISDKPrefix, $@"{OpenTelemetryConstants.SDKPrefix}:{version}"));
+                attributeList.Add(new KeyValuePair<string, object>(ResourceSemanticConventions.ProcessId, Process.GetCurrentProcess().Id));
+
+                // Add these attributes only if running in Azure.
+                if (!string.IsNullOrEmpty(serviceName))
                 {
-                    var azureResourceUri = GetAzureResourceURI(siteName);
-                    if (azureResourceUri != null)
+                    attributeList.Add(new KeyValuePair<string, object>(ResourceSemanticConventions.ServiceName, serviceName));
```
jviau, line 30: note for later (not this PR): should we add attributes based on SKU/platform? IE: windows or linux, and consumption or EP or dedicated?

src/WebJobs.Script/Diagnostics/OpenTelemetry/OpenTelemetryConstants.cs
```diff
@@ -5,6 +5,28 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics.OpenTelemetry
...
+
+        internal const string AzureCloudProviderValue = "azure";
+        internal const string AzurePlatformValue = "azure_functions";
+        internal const string AISDKPrefix = "ai.sdk.prefix";
+        internal const string SDKPrefix = "azurefunctions";
+        internal const string SiteNameEnvVar = "WEBSITE_SITE_NAME";
+        internal const string RegionNameEnvVar = "REGION_NAME";
+        internal const string ResourceGroupEnvVar = "WEBSITE_RESOURCE_GROUP";
+        internal const string OwnerNameEnvVar = "WEBSITE_OWNER_NAME";
+        internal const string AzureFunctionsGroup = "azure.functions.group";
```
jviau, line 15: Should this one, `AISDKPrefix`, `QueryUrl`, and `Fullurl` go into the semantic conventions file? Even if they are not from OTel spec, they are from ours.

src/WebJobs.Script/Diagnostics/OpenTelemetry/ResourceSemanticConventions.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Script.Diagnostics.OpenTelemetry
+{
+    internal class ResourceSemanticConventions
```
jviau: `static`?

src/WebJobs.Script/Config/ConfigurationSectionNames.cs
```diff
@@ -24,5 +24,6 @@ public static class ConfigurationSectionNames
         public const string Retry = "retry";
         public const string SequentialJobHostRestart = JobHost + ":sequentialRestart";
         public const string SendCanceledInvocationsToWorker = "sendCanceledInvocationsToWorker";
+        public const string TelemetryMode = "telemetrymode";
```
jviau: tiny nit:
```suggestion
        public const string TelemetryMode = "telemetryMode";
```

src/WebJobs.Script/Config/ScriptJobHostOptionsSetup.cs
```diff
@@ -73,6 +74,16 @@ public void Configure(ScriptJobHostOptions options)
             options.TestDataPath = webHostOptions.TestDataPath;
             options.IsFileSystemReadOnly = webHostOptions.IsFileSystemReadOnly;
             options.IsStandbyConfiguration = webHostOptions.IsStandbyConfiguration;
+
+            var telemetryModeSection = jobHostSection.GetSection(ConfigurationSectionNames.TelemetryMode);
```
jviau, line 78: This might work?
```suggestion
            options.TelemetryMode = jobHostSection.GetValue(ConfigurationSectionNames.TelemetryMode, TelemetryMode.ApplicationInsights);
```

Can then remove the `if`/`else` statement

src/WebJobs.Script/Diagnostics/MetricEventNames.cs
```diff
@@ -9,6 +9,8 @@ public static class MetricEventNames
         public const string ApplicationStartLatency = "host.application.start";
         public const string ApplicationInsightsEnabled = "host.applicationinsights.enabled";
         public const string ApplicationInsightsDisabled = "host.applicationinsights.disabled";
+        public const string OpenTelemetryAzMonEnabled = "host.otel.azmon.enabled";
+        public const string OpenTelemetryOtlpEnabled = "host.opentelemetry.otlp.enabled";
```
jviau:
```suggestion
        public const string OpenTelemetryOtlpEnabled = "host.otel.otlp.enabled";
```

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -407,7 +408,34 @@ public static IHostBuilder SetAzureFunctionsConfigurationRoot(this IHostBuilder
             return builder;
         }
 
-        internal static void ConfigureApplicationInsights(HostBuilderContext context, ILoggingBuilder builder)
+        internal static void ConfigureTelemetry(this ILoggingBuilder loggingBuilder, HostBuilderContext context)
+        {
+            TelemetryMode mode;
+            var telemetryModeSection = context.Configuration.GetSection(ConfigurationPath.Combine(ConfigurationSectionNames.JobHost, ConfigurationSectionNames.TelemetryMode));
```
jviau, line 414: This looks like the same logic as the other place you retrieve `TelemetryMode` from the config - should you extract to a common helper?

jviau (changes requested):
Are we not providing an environment variable way of enabling OTel? Or relying on default env variable config mapping?

src/WebJobs.Script/Diagnostics/OpenTelemetry/ActivitySanitizingProcessor.cs
RohitRanjanMS: Done.

src/WebJobs.Script/Diagnostics/OpenTelemetry/FunctionsResourceDetector.cs
RohitRanjanMS: yes, we will have to improve this.

src/WebJobs.Script/Config/ScriptJobHostOptionsSetup.cs
RohitRanjanMS: I will have to use try catch block with `GetValue` as it throws argument exception. Leaving this as-is for now.

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
RohitRanjanMS: I will have a look at this later.

src/WebJobs.Script/Config/ScriptJobHostOptionsSetup.cs
jviau: https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.configurationbinder.getvalue?view=dotnet-plat-ext-8.0#microsoft-extensions-configuration-configurationbinder-getvalue-1(microsoft-extensions-configuration-iconfiguration-system-string-0)

What argument exception will it throw?

But fine to address this later.

jviau (approved):

src/WebJobs.Script/Config/ScriptJobHostOptionsSetup.cs
RohitRanjanMS: yes, `jobHostSection.GetValue(ConfigurationSectionNames.TelemetryMode, TelemetryMode.ApplicationInsights);` is throwing an exception if string is not a valid Enum.

src/WebJobs.Script/Diagnostics/MetricEventNames.cs
```diff
@@ -9,6 +9,8 @@ public static class MetricEventNames
         public const string ApplicationStartLatency = "host.application.start";
         public const string ApplicationInsightsEnabled = "host.applicationinsights.enabled";
         public const string ApplicationInsightsDisabled = "host.applicationinsights.disabled";
+        public const string OpenTelemetryAzMonEnabled = "host.otel.azmon.enabled";
+        public const string OpenTelemetryOtlpEnabled = "host.otel.otlp.enabled";
```
liliankasem, line 13: HostMetrics also defines some metric event names:

``` 
public const string AppFailureCount = "azure.functions.app_failures";
public const string ActiveInvocationCount = "azure.functions.active_invocations";
public const string StartedInvocationCount = "azure.functions.started_invocations";
```

Raising this for awareness; just want to make sure everything is aligned. @jviau - any concerns on naming? I know we decided on `azure.functions` for the new metrics we added, but do we also need to follow this naming pattern for existing metrics like the ones in this doc?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -1670,29 +1677,47 @@ private void AddSample<T>(List<T> samples, T sample)
 
         private void AddAdditionalTraceContext(MapField<string, string> attributes, ScriptInvocationContext context)
         {
-            // This is only applicable for AI agents running along side worker
-            if (_environment.IsApplicationInsightsAgentEnabled())
+            bool isOtelEnabled = _scriptHostOptions?.Value.TelemetryMode == TelemetryMode.OpenTelemetry;
+            bool isAIEnabled = _environment.IsApplicationInsightsAgentEnabled();
```
brettsam, line 1681: Isn't this environment check only for Java? dotnet-isolated (and hopefully future languages) all communicate that they are supporting App Insights worker-side with a capability. That needs to be a part of this check as well. Those checks are combined here:

https://github.com/Azure/azure-functions-host/blob/0dc8c60731dc965b2d82b018db563feb01611349/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs#L561-L566

brettsam (approved):

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
RohitRanjanMS: I will fix this when we enable OTel support for Java.

src/WebJobs.Script/Diagnostics/MetricEventNames.cs
RohitRanjanMS: I will follow-up on this.

src/WebJobs.Script/Metrics/HostMetrics.cs
```diff
@@ -5,6 +5,7 @@
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Diagnostics.Metrics;
```
liliankasem, line 7: On line 19 we set CloudPlatformName, should this be updated to use 

OpenTelemetryConstants.AzureCloudProviderValue

RohitRanjanMS: I will change this in next PR.

julealgon:
Where is this documented exactly? Is this PR the only place where there are setup instructions? They feel incredibly barebones to me.

Additionally, does this _only_ support environment variable-based configuration? All of the OpenTelemetry libraries work with the `IConfiguration` model and that's how we setup up _all_ of our services, by having settings such as `OTEL_EXPORTER_OTLP_ENDPOINT` in our shared Azure AppConfiguration instance that propagates it to dozens of separate projects.

The description on the PR seems to indicate our setup would not work, which is why I'd like for someone here to link to a more thorough documentation of OpenTelemetry + Azure Functions.

EDIT:
Found what seems to be the main documentation here:
- [Use OpenTelemetry with Azure Functions](https://learn.microsoft.com/en-us/azure/azure-functions/opentelemetry-howto?tabs=app-insights&pivots=programming-language-csharp)

PR #9997, opened 2024-04-11
Include route params in RpcHttp instance when http proxying is enabled.

### LINKED ISSUE #9840: Provide route params to worker when using http proxy
We'd like the host to pass the worker route param info when using http proxies. We have the url and could potentially parse it ourselves, but the user registers routes using ASP.NET route syntax since the host handles all the actual routing logic. It would be pretty complicated to parse ASP.NET route syntax (e.g. `example/{name}`) since it's different than typical Node.js route syntax (e.g. `example/:name`).

Theoretically this should be easy for the host to do for us since all the route info is in the url, not the request body itself. The host could either send the route param over rpc (we still get an rpc request per invocation) or tacked on to the request sent through the proxy as a header or something. Either way would work for us.

[Discussed offline here](https://teams.microsoft.com/l/message/19:aca957d82eba48f8a6dadf5ba1117d7d@thread.tacv2/1706565951224?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=201c2bb2-812d-4986-ac92-9bfad0d6cc59&parentMessageId=1706565951224&teamName=Azure%20Functions%20Team&channelName=Host&createdTime=1706565951224) (Internal-only thread)

kshyju:
Fixes #9840 
Adding support for workers to request route params data populated in the RpcHttp instance when Http proxying is enabled.
Worker needs to include "**RequiresRouteParameters**" capability in the capability list they send to host during init/env reload request for this to happen.

### Pull request checklist

* [] Backporting to the `in-proc` branch is not required
    * Otherwise: Link to backporting PR - backport PR follows shortly.

### DIFF
--- a/src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs
+++ b/src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs
@@ -101,13 +101,13 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
 
         internal static async Task<TypedData> ToRpcHttp(this HttpRequest request, ILogger logger, GrpcCapabilities capabilities)
         {
-            var handlesHttpRouteParamsWhenProxying = !string.IsNullOrEmpty(capabilities.GetCapabilityState(RpcWorkerConstants.HandlesRouteParamsWhenHttpProxying));
-            var skipHttpInputs = !string.IsNullOrEmpty(capabilities.GetCapabilityState(RpcWorkerConstants.HttpUri));
-            var shouldUseNullableValueDictionary = ShouldUseNullableValueDictionary(capabilities);
+            bool requiresRouteParams = !string.IsNullOrEmpty(capabilities.GetCapabilityState(RpcWorkerConstants.RequiresRouteParameters));
+            bool isHttpProxying = !string.IsNullOrEmpty(capabilities.GetCapabilityState(RpcWorkerConstants.HttpUri));
+            bool shouldUseNullableValueDictionary = ShouldUseNullableValueDictionary(capabilities);
 
             //  If proxying the http request to the worker and
-            //  worker can handle route params, send an empty rpc http object with only params populated.
-            if (skipHttpInputs && handlesHttpRouteParamsWhenProxying)
+            //  worker requesting route params, send an empty rpc http object with only params populated.
+            if (isHttpProxying && requiresRouteParams)
             {
                 var typedDataWithRouteParams = new TypedData
                 {
@@ -120,7 +120,7 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
             }
 
             // If proxying the http request to the worker, keep the grpc message minimal
-            if (skipHttpInputs)
+            if (isHttpProxying)
             {
                 return EmptyRpcHttp;
             }
@@ -223,17 +223,12 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
 
         private static void PopulateHttpRouteDataAsParams(HttpRequest request, RpcHttp http, bool shouldUseNullableValueDictionary)
         {
-            if (!request.HttpContext.Items.TryGetValue(HttpExtensionConstants.AzureWebJobsHttpRouteDataKey, out var routeData))
+            if (!request.HttpContext.Items.TryGetValue(HttpExtensionConstants.AzureWebJobsHttpRouteDataKey, out var routeData)
+                || routeData is not Dictionary<string, object> parameters)
             {
                 return;
             }
 
-            if (routeData == null)
-            {
-                return;
-            }
-
-            var parameters = (Dictionary<string, object>)routeData;
             foreach (var pair in parameters)
             {
                 if (pair.Value != null)
--- a/src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs
+++ b/src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs
@@ -101,8 +101,25 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
 
         internal static async Task<TypedData> ToRpcHttp(this HttpRequest request, ILogger logger, GrpcCapabilities capabilities)
         {
+            var handlesHttpRouteParamsWhenProxying = !string.IsNullOrEmpty(capabilities.GetCapabilityState(RpcWorkerConstants.HandlesRouteParamsWhenHttpProxying));
+            var skipHttpInputs = !string.IsNullOrEmpty(capabilities.GetCapabilityState(RpcWorkerConstants.HttpUri));
+            var shouldUseNullableValueDictionary = ShouldUseNullableValueDictionary(capabilities);
+
+            //  If proxying the http request to the worker and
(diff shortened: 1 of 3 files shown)


ejizba (approved):
Tested on a Node.js app and worked just fine!

YunchuWang:
Validated the change works in python http v2 app, can grab the route params from rpchttp and set into fastapi req
image
image
image

YunchuWang (approved):

src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs
```diff
@@ -101,8 +101,25 @@ internal static TypedData ToModelBindingDataArray(this ParameterBindingData[] da
 
         internal static async Task<TypedData> ToRpcHttp(this HttpRequest request, ILogger logger, GrpcCapabilities capabilities)
         {
+            var handlesHttpRouteParamsWhenProxying = !string.IsNullOrEmpty(capabilities.GetCapabilityState(RpcWorkerConstants.HandlesRouteParamsWhenHttpProxying));
+            var skipHttpInputs = !string.IsNullOrEmpty(capabilities.GetCapabilityState(RpcWorkerConstants.HttpUri));
```
satvu: Would these capabilities have to be rechecked for each HTTP request? Wondering if it's worth trying to optimize this if it may be a hot path.

fabiocav: +1 here. We can cache these at initialization.

src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs
```diff
@@ -101,8 +101,25 @@ internal static TypedData ToModelBindingDataArray(this ParameterBindingData[] da
...
+            var shouldUseNullableValueDictionary = ShouldUseNullableValueDictionary(capabilities);
```
fabiocav: nit: for cases where assigning a method result, `var` -> type

src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs
```diff
@@ -101,8 +101,25 @@ internal static TypedData ToModelBindingDataArray(this ParameterBindingData[] da
...
+
+            //  If proxying the http request to the worker and
+            //  worker can handle route params, send an empty rpc http object with only params populated.
```
fabiocav: I think the `can handle` comment here might be misleading. Those are workers requesting route params.

src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs
fabiocav: I realize this is pre-existing, but can we name `skipHttpInputs` something a little more descriptive?

src/WebJobs.Script/Workers/Rpc/RpcWorkerConstants.cs
```diff
@@ -57,6 +57,11 @@ public static class RpcWorkerConstants
         public const string HttpUri = "HttpUri";
         public const string WorkerOpenTelemetryEnabled = nameof(WorkerOpenTelemetryEnabled);
 
+        /// <summary>
+        /// Indicates whether the RpcHttp request should include the route parameters when the request is being proxied to an HTTP worker.
+        /// </summary>
+        public const string HandlesRouteParamsWhenHttpProxying = "HandlesRouteParamsWhenHttpProxying";
```
fabiocav: naming: Perhaps something like `RequiresRouteParameters`, trying to avoid `Handles` here. Not sure we need the HTTP proxying qualification.

src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs
```diff
@@ -222,6 +221,35 @@ internal static async Task<TypedData> ToRpcHttp(this HttpRequest request, ILogge
...
         }
 
+        private static void PopulateHttpRouteDataAsParams(HttpRequest request, RpcHttp http, bool shouldUseNullableValueDictionary)
+        {
+            if (!request.HttpContext.Items.TryGetValue(HttpExtensionConstants.AzureWebJobsHttpRouteDataKey, out var routeData))
+            {
+                return;
+            }
+
+            if (routeData == null)
```
fabiocav: move this to the condition above ( `|| routeData == null`)

src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs
kshyju: One possible optimization is to have a new type with boolean properties for each of these capabilities. This type should be initialized/updated during worker init/env reload response. Using that will bring a minor improvement as the dictionary look up can be removed. but that changes goes beyond the scope of this PR. We can consider making that change and update a bunch of places to use that.

satvu (approved):

src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs
```diff
@@ -222,6 +221,35 @@ internal static async Task<TypedData> ToRpcHttp(this HttpRequest request, ILogge
...
+            {
+                return;
+            }
+
+            var parameters = (Dictionary<string, object>)routeData;
```
fabiocav: Can we change this to perform a safe cast and log if that fails instead? You can combine this into the condition above and that would eliminate the need for the `null` check, simplifying this code.

src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs
kshyju: Switched to explicit types.

src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs
kshyju: Updated comment to "_worker requesting route params.._"

src/WebJobs.Script/Workers/Rpc/RpcWorkerConstants.cs
kshyju: Renamed capability name and constant to `RequiresRouteParameters`

src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs
kshyju: Done.

src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs
kshyju: > I realize this is pre-existing, but can we name `skipHttpInputs` something a little more descriptive?

Renamed to `isHttpProxying`

fabiocav (approved):

PR #10011, opened 2024-04-13
Introduce versioning for DiagnosticEvent

liliankasem:
### Pull request checklist

    * [x] Otherwise: Link to backporting PR- TODO

### Additional information

The changes implement a versioning mechanism for DiagnosticEvent entries stored in Azure Table Storage. This ensures that only the latest version of each event is retained, preventing the accumulation of outdated or redundant data.​

#### Key Changes

- Event Versioning: Introduced a Version property to the DiagnosticEvent class to track the version of each event.​
- Purge Mechanism: Implemented a method to purge older versions of events, ensuring that only the most recent version is stored.​
- Deletion Delay Handling: Added a 30-second delay when a table is deleted to handle eventual consistency in Azure Table Storage.​
- Host State Provider Retrieval: Modified the retrieval of IPrimaryHostStateProvider to obtain it via the script host, ensuring accurate host state information.​
- Caching Behavior: Adjusted the caching behavior to avoid caching the IPrimaryHostStateProvider, ensuring that the latest state is always retrieved.

### DIFF
--- a/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEvent.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEvent.cs
@@ -11,6 +11,8 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 {
     public class DiagnosticEvent : TableEntity
     {
+        internal const string CurrentEventVersion = "2024-05-01";
+
         public DiagnosticEvent() { }
 
         public DiagnosticEvent(string hostId, DateTime timestamp)
@@ -18,7 +20,10 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
             RowKey = TableStorageHelpers.GetRowKey(timestamp);
             PartitionKey = $"{hostId}-{timestamp:yyyyMMdd}";
             Timestamp = timestamp;
+            EventVersion = CurrentEventVersion;
         }
+
+        public string EventVersion { get; set; }
 
         public int HitCount { get; set; }
 
--- a/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
@@ -3,10 +3,13 @@
 
 using System;
 using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.Cosmos.Table;
 using Microsoft.Azure.WebJobs.Host.Executors;
+using Microsoft.Azure.WebJobs.Hosting;
 using Microsoft.Azure.WebJobs.Logging;
 using Microsoft.Azure.WebJobs.Script.WebHost.Helpers;
 using Microsoft.Extensions.Configuration;
@@ -25,6 +28,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
         private readonly IHostIdProvider _hostIdProvider;
         private readonly IEnvironment _environment;
         private readonly ILogger<DiagnosticEventTableStorageRepository> _logger;
+        private readonly IPrimaryHostStateProvider _primaryHostStateProvider;
         private readonly object _syncLock = new object();
 
         private ConcurrentDictionary<string, DiagnosticEvent> _events = new ConcurrentDictionary<string, DiagnosticEvent>();
@@ -32,19 +36,23 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
         private CloudTable _diagnosticEventsTable;
         private string _hostId;
         private bool _disposed = false;
+        private bool _purged = false;
         private string _tableName;
 
-        internal DiagnosticEventTableStorageRepository(IConfiguration configuration, IHostIdProvider hostIdProvider, IEnvironment environment, ILogger<DiagnosticEventTableStorageRepository> logger, int logFlushInterval)
+        internal DiagnosticEventTableStorageRepository(IConfiguration configuration, IHostIdProvider hostIdProvider, IEnvironment environment, IPrimaryHostStateProvider primaryHostStateProvider,
+            ILogger<DiagnosticEventTableStorageRepository> logger, int logFlushInterval)
         {
             _configuration = configuration;
             _hostIdProvider = hostIdProvider;
             _environment = environment;
+            _primaryHostStateProvider = primaryHostStateProvider;
             _logger = logger;
             _flushLogsTimer = new Timer(OnFlushLogs, null, logFlushInterval, logFlushInterval);
         }
 
-        public DiagnosticEventTableStorageRepository(IConfiguration configuration, IHostIdProvider hostIdProvider, IEnvironment environment, ILogger<DiagnosticEventTableStorageRepository> logger)
-            : this(configuration, hostIdProvider, environment, logger, LogFlushInterval) { }
+        public DiagnosticEventTableStorageRepository(IConfiguration configuration, IHostIdProvider hostIdProvider, IEnvironment environment, IPrimaryHostStateProvider primaryHostStateProvider,
+            ILogger<DiagnosticEventTableStorageRepository> logger)
+            : this(configuration, hostIdProvider, environment, primaryHostStateProvider, logger, LogFlushInterval) { }
 
         internal CloudTableClient TableClient
         {
@@ -88,7 +96,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
             if (TableClient != null)
             {
                 now = now ?? DateTime.UtcNow;
-                string currentTableName = GetCurrentTableName(now.Value);
+                string currentTableName = GetTableName(now.Value);
 
                 // update the table reference when date rolls over to a new month
                 if (_diagnosticEventsTable == null || currentTableName != _tableName)
@@ -101,14 +109,61 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
             return _diagnosticEventsTable;
         }
 
-        private static string GetCurrentTableName(DateTime now)
+        private static string GetTableName(DateTime date)
         {
-            return $"{TableNamePrefix}{now:yyyyMM}";
+            return $"{TableNamePrefix}{date:yyyyMM}";
         }
 
         protected internal virtual async void OnFlushLogs(object state)
         {
             await FlushLogs();
+        }
+
+        private async Task PurgePreviousEventVersions()
+        {
+            _logger.LogTrace("Purging diagnostic events with versions older than '{currentEventVersion}'.", DiagnosticEvent.CurrentEventVersion);
+
+            await Utility.InvokeWithRetriesAsync(async () =>
+            {
+                try
+                {
+                    var tables = (await TableStorageHelpers.ListTablesAsync(TableClient, TableNamePrefix)).ToList();
+
+                    foreach (var table in tables)
+                    {
+                        var tableRecords = await table.ExecuteQuerySegmentedAsync(new TableQuery<DiagnosticEvent>(), null);
+
+                        // Skip tables that have 0 records
+                        if (tableRecords.Results.Count == 0)
+                        {
+                            continue;
+                        }
+
+                        // Delete table if it doesn't have records with EventVersion
+                        var eventVersionDoesNotExists = tableRecords.Results.Any(record => string.IsNullOrEmpty(record.EventVersion) == true);
+                        if (eventVersionDoesNotExists)
+                        {
+                            _logger.LogTrace("Deleting table '{tableName}' as it contains records without an EventVersion.", table.Name);
+                            await table.DeleteIfExistsAsync();
+                            continue;
+                        }
+
+                        // If the table does have EventVersion, query if it is an outdated version
+                        var eventVersionOutdated = tableRecords.Results.Any(record => string.Compare(DiagnosticEvent.CurrentEventVersion, record.EventVersion, StringComparison.Ordinal) > 0);
+                        if (eventVersionOutdated)
+                        {
+                            _logger.LogTrace("Deleting table '{tableName}' as it contains records with an outdated EventVersion.", table.Name);
+                            await table.DeleteIfExistsAsync();
+                        }
+                    }
+
+                    _purged = true;
+                }
+                catch (Exception ex)
+                {
+                    _logger.LogError(ex, "Error occurred when attempting to purge previous diagnostic event versions.");
+                }
+            }, maxRetries: 5, retryInterval: TimeSpan.FromSeconds(5));
         }
 
         internal virtual async Task FlushLogs(CloudTable table = null)
@@ -118,13 +173,18 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
                 return;
             }
 
+            if (_primaryHostStateProvider.IsPrimary && !_purged)
+            {
+                _ = PurgePreviousEventVersions();
+            }
+
             try
             {
                 table = table ?? GetDiagnosticEventsTable();
 
                 if (table == null)
                 {
-                    _logger.LogError("Unable to get table reference. Aborting write operation");
+                    _logger.LogError("Unable to get table reference. Aborting write operation.");
                     StopTimer();
                     return;
                 }
@@ -132,12 +192,13 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
                 bool tableCreated = await TableStorageHelpers.CreateIfNotExistsAsync(table, TableCreationMaxRetryCount);
                 if (tableCreated)
                 {
+                    _logger.LogTrace("Queueing background table purge.");
                     TableStorageHelpers.QueueBackgroundTablePurge(table, TableClient, TableNamePrefix, _logger);
                 }
             }
             catch (Exception ex)
             {
-                _logger.LogError(ex, $"Unable to get table reference or create table. Aborting write operation.");
+                _logger.LogError(ex, "Unable to get table reference or create table. Aborting write operation.");
                 // Clearing the memory cache to avoid memory build up.
                 _events.Clear();
                 return;
@@ -169,9 +230,9 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
                 await table.ExecuteBatchAsync(batch);
                 events.Clear();
             }
-            catch (Exception e)
+            catch (Exception ex)
             {
-                _logger.LogError(e, $"Unable to write diagnostic events to table storage:{e}");
+                _logger.LogError(ex, "Unable to write diagnostic events to table storage.");
             }
         }
 
@@ -204,7 +265,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 
         private void StopTimer()
         {
-            _logger.LogInformation("Stopping the flush logs timer");
+            _logger.LogInformation("Stopping the flush logs timer.");
             _flushLogsTimer?.Change(Timeout.Infinite, Timeout.Infinite);
         }
 
--- a/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
@@ -13,6 +13,7 @@ using Microsoft.Azure.WebJobs.Hosting;
 using Microsoft.Azure.WebJobs.Logging;
 using Microsoft.Azure.WebJobs.Script.WebHost.Helpers;
 using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 
 namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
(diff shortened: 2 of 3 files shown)


src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -17,6 +17,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
     public class DiagnosticEventTableStorageRepository : IDiagnosticEventRepository, IDisposable
     {
         internal const string TableNamePrefix = "AzureFunctionsDiagnosticEvents";
+        private const string CurrentEventVersion = "2024-05-01";
```
liliankasem: TODO:// avoid hardcoding this twice

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -101,16 +102,55 @@ internal CloudTable GetDiagnosticEventsTable(DateTime? now = null)
...
-            return $"{TableNamePrefix}{now:yyyyMM}";
+            return $"{TableNamePrefix}{date:yyyyMM}";
         }
 
         protected internal virtual async void OnFlushLogs(object state)
         {
             await FlushLogs();
         }
 
+        // TODO:// Call this ~1 minute after host startup to avoid blocking host start & impacting cold start.
```
liliankasem: TODO:// need to invoke this function somewhere

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -101,16 +102,55 @@ internal CloudTable GetDiagnosticEventsTable(DateTime? now = null)
...
+                        await table.DeleteIfExistsAsync();
+                        continue;
+                    }
+
+                    // If the table does have event version, query if it is an outdated version
+                    var eventVersionOutdated = await table.ExecuteQuerySegmentedAsync(eventVersionOutdatedQuery, null);
+                    if (eventVersionOutdated.Results.Count > 0)
+                    {
+                        // TODO:// We also have the option here to delete the records instead of the entire table.
+                        await table.DeleteIfExistsAsync();
```
brettsam: I do think that deleting the whole table is a better approach, however I recall us running into lots of issues upon recreation in the past -- see this -- https://learn.microsoft.com/en-us/rest/api/storageservices/Delete-Table#remarks.

It may take 40 seconds for that table to be allowed to be re-created. So I think we'll need to do something in the create flow (if it's not already there) to backoff for a minute and retry a couple times if we're unable to recreate when we go to flush the next time.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
brettsam: On second thought -- let's just make this an int? it'll be easier to parse and determine less-than, etc.

And you can make an internal static like `DiagnosticEvent.CurrentEventVersion` to prevent double-hardcoding.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -101,16 +102,55 @@ internal CloudTable GetDiagnosticEventsTable(DateTime? now = null)
...
+            return $"{TableNamePrefix}{date:yyyyMM}";
         }
 
         protected internal virtual async void OnFlushLogs(object state)
         {
             await FlushLogs();
         }
 
+        // TODO:// Call this ~1 minute after host startup to avoid blocking host start & impacting cold start.
+        private async void PurgePreviousEventVersions()
```
brettsam: i think we should have some logs in here for when we delete tables... probably same for when we create/delete in the other Purge method

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
liliankasem: The Table query is able to compare these just fine and it's working as expected (I tested with dates before and after May 1st and it was dropping the correct tables); would you prefer this to be something like 1.0?

liliankasem: ~I just tried out the int version instead in the console app and it's not deleting any table with the value less than the current; given the string one is working i would like to stick with that~

I've switch approach to linq so doing an int comparison should be fine if we want to still switch to a number instead.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
liliankasem: resolving; CreateIfNotExistsAsync helper takes care of these issues

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
liliankasem: I've added a generic log at the start of this method, but would like to not add logs to the foreach loop - that would be chaos :D

fabiocav: Can we add logic for the table operations, though? Not every loop, but any time the evaluation results in a table deletion, it would be good to log.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -101,16 +101,57 @@ internal CloudTable GetDiagnosticEventsTable(DateTime? now = null)
...
+
+                TableQuery<DiagnosticEvent> eventVersionOutdatedQuery = new TableQuery<DiagnosticEvent>()
+                    .Where(TableQuery.GenerateFilterCondition(nameof(DiagnosticEvent.EventVersion), QueryComparisons.LessThan, DiagnosticEvent.CurrentEventVersion));
+
+                foreach (var table in tables)
+                {
+                    var eventVersionExists = await table.ExecuteQuerySegmentedAsync(eventVersionExistsQuery, null);
+
+                    // Delete table if it doesn't have records with event version
+                    if (eventVersionExists.Results.Count == 0)
```
fabiocav: Could this check lead to scenarios where different instances will be deleting a table that was just created to persist new entries (with the new version)? If the table is empty, it could be that it was about to be used by a different instance to persist data.

liliankasem: This is a great point, yes it would delete those too I believe. There is no way to check for if a "column" exists though :(

liliankasem: We could just delete the records here instead of the entire table? But that might be excessive. Or just ignore tables with no records

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -101,43 +109,92 @@ internal CloudTable GetDiagnosticEventsTable(DateTime? now = null)
...
         {
             if (_environment.IsPlaceholderModeEnabled())
             {
                 return;
             }
 
+            if (_primaryHostStateProvider.IsPrimary && !_purged)
+            {
+                // await PurgePreviousEventVersions();
+                var tIgnore = Task.Run(async () => { await PurgePreviousEventVersions(); });
```
liliankasem: To ignore or not to ignore, that is the question

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
liliankasem: Okay I've refactored this to ignore tables that have 0 records - thoughts?

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
liliankasem: I can add this, but my concern is still the same; if there are 10s of tables, we could be writing a log for each time we delete which is excessive? Maybe this is not a real concern as we purge every month anyway so we are not really dealing with any more than a table or 2?

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
fabiocav: If ignoring, you can use a [discard](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/discards) here. 

You should also be able to call `PurgePreviousEventVersions` without awaiting and directly assigning to a discard.

fabiocav (approved):

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
liliankasem: @brettsam updated this to purge in background (`_ = PurgePreviousEventVersions();` - if you feel strongly about otherwise feel free to change.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -101,43 +109,93 @@ internal CloudTable GetDiagnosticEventsTable(DateTime? now = null)
...
             await FlushLogs();
         }
 
+        private async Task PurgePreviousEventVersions()
+        {
+            _logger.LogTrace("Purging previous diagnostic event versions.");
+
+            try
+            {
+                var tables = await TableStorageHelpers.ListTablesAsync(TableClient, TableNamePrefix);
```
kshyju: Should we materialize this IEnumerable? (calling `ToArray` or `ToList) ? Wondering If another add happens while this gets executed, that ends up throwing.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -101,43 +109,93 @@ internal CloudTable GetDiagnosticEventsTable(DateTime? now = null)
...
+                        _logger.LogTrace("Deleting table '{tableName}' as it contains records with an outdated EventVersion.", table.Name);
+                        await table.DeleteIfExistsAsync();
+                    }
+                }
+
+                _purged = true;
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex, "Error occurred when attempting to purge previous diagnostic event versions.");
```
kshyju: Should we also log, how many tables were there and how many were successfully purged? Wondering that would be helpful in the case of partial purge and and an exception occurred.

We can do that logging in a finally block.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEvent.cs
```diff
@@ -11,15 +11,20 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 {
     public class DiagnosticEvent : TableEntity
     {
+        internal const string CurrentEventVersion = "2024-05-01";
```
kshyju, line 14: "20240501" anyone ?

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -101,43 +109,93 @@ internal CloudTable GetDiagnosticEventsTable(DateTime? now = null)
...
         }
 
         protected internal virtual async void OnFlushLogs(object state)
         {
             await FlushLogs();
         }
 
+        private async Task PurgePreviousEventVersions()
+        {
+            _logger.LogTrace("Purging previous diagnostic event versions.");
```
kshyju: Should we also include what the current Event version value is at the time of this operation? Would it add any value?

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -101,43 +109,93 @@ internal CloudTable GetDiagnosticEventsTable(DateTime? now = null)
...
+                    var eventVersionDoesNotExists = tableRecords.Results.Any(record => string.IsNullOrEmpty(record.EventVersion) == true);
+                    if (eventVersionDoesNotExists)
+                    {
+                        _logger.LogTrace("Deleting table '{tableName}' as it contains records without an EventVersion.", table.Name);
+                        await table.DeleteIfExistsAsync();
+                        continue;
+                    }
+
+                    // If the table does have EventVersion, query if it is an outdated version
+                    var eventVersionOutdated = tableRecords.Results.Any(record => record.EventVersion.CompareTo(DiagnosticEvent.CurrentEventVersion) < 0);
```
kshyju: `String.Compare` to is culture sensitive. Would it be a problem?

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -101,43 +109,93 @@ internal CloudTable GetDiagnosticEventsTable(DateTime? now = null)
...
+                    if (eventVersionOutdated)
+                    {
+                        _logger.LogTrace("Deleting table '{tableName}' as it contains records with an outdated EventVersion.", table.Name);
+                        await table.DeleteIfExistsAsync();
+                    }
+                }
+
+                _purged = true;
+            }
+            catch (Exception ex)
```
kshyju: Should we also wrap this code inside a retry snippet so that transient issues (connection dropped to table storage) has minimal impact?

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -101,43 +109,93 @@ internal CloudTable GetDiagnosticEventsTable(DateTime? now = null)
...
+                        _logger.LogTrace("Deleting table '{tableName}' as it contains records without an EventVersion.", table.Name);
+                        await table.DeleteIfExistsAsync();
+                        continue;
+                    }
+
+                    // If the table does have EventVersion, query if it is an outdated version
+                    var eventVersionOutdated = tableRecords.Results.Any(record => record.EventVersion.CompareTo(DiagnosticEvent.CurrentEventVersion) < 0);
+                    if (eventVersionOutdated)
+                    {
+                        _logger.LogTrace("Deleting table '{tableName}' as it contains records with an outdated EventVersion.", table.Name);
```
kshyju: What is a possible `EventVersion` value which is outdated (as of today)? We are deleting the old records (without `EventVersion`) in 145. Is that not enough?

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
fabiocav: This is a good callout. We should swap to using `string.Compare` and be explicit about the comparison (ordinal / case insensitive here)

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEvent.cs
liliankasem: I like the azure api version model

https://learn.microsoft.com/en-us/rest/api/searchmanagement/management-api-versions#stable-versions

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
liliankasem: Good idea

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
liliankasem: As of today, nothing. But we want to future proof this by deleting tables that have an old API version in future

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
liliankasem: I don't think it would throw? But if doing ToList is safer, we can do that!

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
liliankasem: Not every table we find might need to be deleted so we would have to track something like `identifiedForDeletionCount` and `deletedTableCount` and report those at the end.

I don't think we need this log since we already have logs before each table gets deleted saying "Deleting table {blah}" And I think from those logs it's enough to identify if a table that was supposed to be deleted was or wasn't. I think that is sufficient, what do you think?

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -101,43 +109,96 @@ internal CloudTable GetDiagnosticEventsTable(DateTime? now = null)
...
+                        if (eventVersionOutdated)
+                        {
+                            _logger.LogTrace("Deleting table '{tableName}' as it contains records with an outdated EventVersion.", table.Name);
+                            await table.DeleteIfExistsAsync();
+                        }
+                    }
+
+                    _purged = true;
+                }
+                catch (Exception ex)
```
kshyju, line 167: Currently if any code inside the try block throws (ex: `DeleteIfExistsAsync` method call throws), the catch block here in line 162 will catch it and no retries will happen. 

We should change it to something like this.
```csharp
try
{
    await Utility.InvokeWithRetriesAsync(async () =>
    {
        // try to delete.
        _purged = true;
    }, maxRetries: 5, retryInterval: TimeSpan.FromSeconds(5));
}
catch (Exception ex)
{
    // Even retries failed. So log error.
    _logger.LogError(ex, "Error occurred when attemp....");
}
```

liliankasem: Good catch

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
brettsam: There should only be one table in most cases, I think? We should be purging old tables every time we create a new one, so it shouldn't be too chaotic. It'd be good to know... or even write a single log at the end with "Purged the following tables: ...."

kshyju (approved):

src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
```diff
@@ -161,6 +161,12 @@ public static class WebScriptHostBuilderExtension
                     // Logging and diagnostics
                     services.AddSingleton<IMetricsLogger>(a => new NonDisposableMetricsLogger(metricsLogger));
                     services.AddSingleton<IEventCollectorProvider, FunctionInstanceLogCollectorProvider>();
+                    if (!FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagDisableDiagnosticEventLogging))
```
brettsam: Why did this move?

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -101,43 +109,96 @@ internal CloudTable GetDiagnosticEventsTable(DateTime? now = null)
...
         internal virtual async Task FlushLogs(CloudTable table = null)
         {
             if (_environment.IsPlaceholderModeEnabled())
             {
                 return;
             }
 
+            if (_primaryHostStateProvider.IsPrimary && !_purged)
+            {
+                _ = PurgePreviousEventVersions();
```
brettsam: If we kick this off... we probably want to at least sleep here for 30 seconds or so? Otherwise, I can imagine a race where the code below thinks the table exists and starts writing to it... only for it to be deleted out from under it. Waiting 30 seconds or so if we *know* we're deleting a table would calm that down.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -101,43 +109,96 @@ internal CloudTable GetDiagnosticEventsTable(DateTime? now = null)
...
-                    _logger.LogError("Unable to get table reference. Aborting write operation");
+                    _logger.LogError("Unable to get table reference. Aborting write operation.");
                     StopTimer();
                     return;
                 }
 
                 bool tableCreated = await TableStorageHelpers.CreateIfNotExistsAsync(table, TableCreationMaxRetryCount);
                 if (tableCreated)
                 {
+                    _logger.LogTrace("Queueing background table purge.");
```
brettsam: All these logs will have to be `Debug` if we want to see them in kusto...

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
liliankasem: We don't delete empty tables so this should be fine?

Hmm I guess that does not take into account tables with mixed records that would get deleted. Will look into adding a delay. Or should we not make this a background task and wait on the delete to complete?

src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
liliankasem: Host would not start due to this error:

> A host error has occurred during startup operation '06dce65b-84c5-4978-96ad-09bec9ee7edc'. System.InvalidOperationException: Unable to resolve service for type 'Microsoft.Azure.WebJobs.Hosting.IPrimaryHostStateProvider' while attempting to activate 'Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics.DiagnosticEventTableStorageRepository'.

I believe it's to do with the
DiagnosticEventTableStorageRepository
being registered at Web vs Script host and the primary host provider not being available at that time; I moved it and now this issue is fixed. Is there something to be concerned about or another way to address the error?

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
liliankasem: Done

src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
fabiocav: One side effect you'll see is that we'll no longer have the logger provider injected at the Web Application level. This could impact any flows that currently raise a diagnostic event outside of the ScriptHost scope (I haven't reviewed to validate whether we have any).

This will also have a (minor) perf impact at cold start.

A safer option here might be to take a dependency on the `WebJobsScriptHostService` (which will be available at the Web Application level) and get the `IPrimaryHostStateProvider` service from the active host (or directly from `WebJobsScriptHostService`). 

That would enable you to keep the scopes/levels unchanged.

liliankasem: Reverted this change and switching to getting the `IPrimaryHostStateProvider` via `_serviceProvider = scriptHostManager as IServiceProvider;`

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
liliankasem: Added 30s delay if a table is deleted, also changed this to no longer be a background task and is now awaited

src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
brettsam: Coming in late to this -- but Fabio is correct that we added DiagnosticEvents at the WebHost level at one point, which means those would stop firing without these services registered there.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
brettsam: Yeah, I was thinking of this mixed-version case. Over time this wouldn't really be an issue (until we change versions again). 

And I think you're right that awaiting is a good approach. There's no rush to flush these (we've already waited 10 minutes) so being more deterministic works for us in this case.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -83,12 +93,14 @@ internal string HostId
 
         internal ConcurrentDictionary<string, DiagnosticEvent> Events => _events;
 
+        internal IPrimaryHostStateProvider HostStateProvider => _primaryHostStateProvider ??= _serviceProvider?.GetService<IPrimaryHostStateProvider>();
```
brettsam: I don't think we can cache this... because of the scopes potentially being different with WebHost vs JobHost, you could end up holding onto an `IPrimaryHostStateProvider` for a host that has been disposed.

I think that every time we need to run through this and take whatever `IPrimaryHostStateProvder` is returned... or skip the purge if null, which you're already doing below. Perf isn't really a concern here so it shouldn't be a big deal to not cache this.

Would recommend making this property into something like `GetPrimaryHostStateProvider()` that tries to resolve the service every time.

liliankasem: At yup good catch, was not intending to cache this. No coding after 10pm for me :)

PR #10037, opened 2024-04-19
Update Azure.Identity to 1.11.2

jviau:
### Issue describing the changes in this PR

resolves #issue_for_this_pr

### Pull request checklist

**IMPORTANT**: Currently, changes must be backported to the `in-proc` branch to be included in Core Tools and non-Flex deployments.

* [ ] Backporting to the `in-proc` branch is not required
    * Otherwise: Link to backporting PR
* [x] My changes **do not** require documentation changes
    * [ ] Otherwise: Documentation issue linked to PR
* [x] My changes **should not** be added to the release notes for the next release
    * [ ] Otherwise: I've added my notes to `release_notes.md`
* [x] My changes **do not** need to be backported to a previous version
    * [ ] Otherwise: Backport tracked by issue/PR #issue_or_pr
* [x] My changes **do not** require diagnostic events changes
    * Otherwise: I have added/updated all related diagnostic events and their documentation (Documentation issue linked to PR)
* [x] I have added all required tests (Unit tests, E2E tests)

### Additional information

Addresses https://github.com/advisories/GHSA-x674-v45j-fwxw

**ALTERNATIVE PR:** #10038

PR #10038, opened 2024-04-19
Remove explicit vulnerabilities check

jviau:

resolves #issue_for_this_pr

### Pull request checklist

### Additional information

This is an alternative to #10037

Removes the explicit vulnerabilities check. Now that we use the .NET8 SDK this is no longer needed as there is a built in nuget audit as part of the restore phase.

**IMPORTANT**: There is a behavior difference (which is the goal here), we only fail on moderate and above now. In this case we have [CVE GHSA-x674-v45j-fwxw](https://github.com/advisories/GHSA-x674-v45j-fwxw) which does not affect us, yet our current approach blocks the build. I could work on a way to integrate suppressions into the existing vuln check script, but when moving to 1ES we will be covered by component governance.

### DIFF
--- a/src/WebJobs.Script/WebJobs.Script.csproj
+++ b/src/WebJobs.Script/WebJobs.Script.csproj
@@ -86,9 +86,12 @@
     <PackageReference Include="System.IO.Abstractions" Version="2.1.0.227">
       <NoWarn>NU1701</NoWarn>
     </PackageReference>
+    <PackageReference Include="System.Net.Http" Version="4.3.4" />
     <PackageReference Include="System.Reactive.Linq" Version="5.0.0" />
     <PackageReference Include="System.Reactive.Core" Version="5.0.0" />
     <PackageReference Include="System.Runtime.Loader" Version="4.3.0" />
+    <PackageReference Include="System.Text.RegularExpressions" Version="4.3.1" />
+
 <!--
     Explicitly referencing version 6.0.9 of System.Text.Json to get two bug fixes for the Source Generator. Without this, the 6.0.0 version 
     of the Source Generator is used, which causes issues. This reference can be removed in future framework versions.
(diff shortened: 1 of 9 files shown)


liliankasem (approved):

FinVamp1:
Adding @fabiocav  as well.  We added this to break the build on checking packages. We should discuss how Component Governance can assist here.

fabiocav:
Will this create a temporary gap? This check has already helped us identify issues prior to releases, and if we're removing, we want to make sure we don't end up with CVE impacted transitive dependencies as those will be flagged by other tools post-deployment/release.

jviau:
@FinVamp1, component governance is the official way for Microsoft repos to manage security audits when using open-source software. It will cover transitive dependencies, set due dates, allow extensions and dismissing of not applicable incidents.

jviau:
@fabiocav , @FinVamp1 - I updated the built in dotnet sdk nuget audit: scans transitive dependencies, only fails on a moderate and above.

src/WebJobs.Script/WebJobs.Script.csproj
```diff
@@ -78,17 +65,16 @@
...
     <PackageReference Include="System.Drawing.Common" Version="8.0.0" />
-    <PackageReference Include="System.IO.Abstractions" Version="2.1.0.227">
-      <NoWarn>NU1701</NoWarn>
-    </PackageReference>
+    <PackageReference Include="System.IO.Abstractions" Version="2.1.0.227" NoWarn="NU1701" />
+    <PackageReference Include="System.Net.Http" Version="4.3.4" />
     <PackageReference Include="System.Reactive.Linq" Version="5.0.0" />
     <PackageReference Include="System.Reactive.Core" Version="5.0.0" />
     <PackageReference Include="System.Runtime.Loader" Version="4.3.0" />
+    <PackageReference Include="System.Text.RegularExpressions" Version="4.3.1" />
```
brettsam, line 76: there's a few new refs here -- was that intentional?

jviau: Component governance flags these as vulnerable. Although it is not an issue at runtime since the WebHost has thse versions. Just adding them here to satisfy CG.

jviau:
Will be replaced with 1ES work

PR #9858, opened 2024-02-02
[pack] Use dotnet common artifacts output

jviau:
### Issue describing the changes in this PR

Part of #9824

### Pull request checklist

* [x] My changes **do not** require documentation changes
    * [ ] Otherwise: Documentation issue linked to PR
* [ ] My changes **should not** be added to the release notes for the next release
    * [ ] Otherwise: I've added my notes to `release_notes.md`
* [x] My changes **do not** need to be backported to a previous version
    * [ ] Otherwise: Backport tracked by issue/PR #issue_or_pr
* [x] My changes **do not** require diagnostic events changes
    * Otherwise: I have added/updated all related diagnostic events and their documentation (Documentation issue linked to PR)
* [x] I have added all required tests (Unit tests, E2E tests)

### Additional information

This PR onboards this repo to .NET8 artifacts output: https://learn.microsoft.com/en-us/dotnet/core/sdk/artifacts-output. Using this system, all build artifacts from all projects will go into a common directory in the repo root: `out/`. The layout is:

- `out/bin/{ProjectName}/{Pivot}`: build output
- `out/obj/{ProjectName}/{Pivot}`: intermediate output
- `out/pub/{ProjectName}/{Pivot}`: publish output
- `out/pkg/{ProjectName}/{Pivot}`: pack output

`{Pivot}` being `{configuration}(_{targetFramework})?(_{runtime})?`

CI scripts have been updated to handle the new output locations.

Test updates are still in progress

PR #10079, opened 2024-04-29
[in-proc backport] Use dotnet common artifacts output (#9858)

jviau:
Backport of #9858

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 bin
 obj
+out
 csx
 .vs
 .vscode
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,0 +1,9 @@
+<Project>
+
+  <PropertyGroup>
+    <ArtifactsPath>$(MSBuildThisFileDirectory)out</ArtifactsPath>
+    <ArtifactsPublishOutputName>pub</ArtifactsPublishOutputName>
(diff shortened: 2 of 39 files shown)


liliankasem:
@jviau  Can you remove the `Update to .NET8 SDK` commit?

jviau:
> @jviau Can you remove the `Update to .NET8 SDK` commit?

This functionality is only available in .NET8 SDK. We will also need this when mult-targeting net8 & net6 anyways.

I won't be checking this in until we resolve the docker build issue.

liliankasem:
> > @jviau Can you remove the `Update to .NET8 SDK` commit?
> 
> This functionality is only available in .NET8 SDK. We will also need this when mult-targeting net8 & net6 anyways.
> 
> I won't be checking this in until we resolve the docker build issue.

Then I think this should be a separate PR, I think the backport PR should be purely backport only without any extra changes. We also want to put all .NET 8 changes into a feature branch first before bringing it all into in-proc. If it makes more sense, maybe these two PRs can point at the in-proc .net 8 feature branch

Edit: created a feature branch: https://github.com/Azure/azure-functions-host/tree/feature/in-proc8

fabiocav:
If this is blocked by the docker issues, we need to consider bringing this change to the feature branch used for multi-targeting.

jviau:
This was ported in another PR

PR #10083, opened 2024-04-30
Add policy for issues with no activity in over 3 years

liliankasem:
Add 'Needs: Author Feedback' label to issues that have not been touched in 3 years. This policy runs every Monday at 10am.


jviau (approved):

fabiocav (approved):

mattchenderson (changes requested):
I seriously dislike requesting action from someone without describing the action for them to take. But the action here is separate from the use of this label anyway. This catches actionable items that have been triaged into the backlog. And it will create re-entrants if authors do come back. I think we need to discuss and rethink.

fabiocav:
Closing this until we discuss the approach we want to take.

PR #10138, opened 2024-05-09
[dev backport] Add AssistantSkillTrigger to WebHook triggers (#10120)

jviau:
Backport of #10120 from flex-consumption to dev


liliankasem:
How did we get into a situation where we are backporting flex to dev? I thought our plan is always check in to dev and cherry pick to flex when needed?

liliankasem (approved):

jviau:
> How did we get into a situation where we are backporting flex to dev? I thought our plan is always check in to dev and cherry pick to flex when needed?

that was my mistake, did it in the reverse order

fabiocav (approved):

PR #10144, opened 2024-05-09
Reference CSharpPrecompiledTestProjects directly

liliankasem:
Updating integration test project to reference the CSharpPrecompiledTestProjects directly; a backport from what we are now doing in the in-proc branch:

https://github.com/Azure/azure-functions-host/pull/10117


jviau (approved):
Can we delete the unused precompiled solution file? Or is that for another PR?

liliankasem:
> Can we delete the unused precompiled solution file? Or is that for another PR?

I can delete it, but I didn't in the in-proc branch so we'll need another PR; maybe we can lump that in with something else

kshyju (approved):

Issue #10135, opened 2024-05-08
Address flaky unit tests


jviau:
Using this issue to track flaky unit tests as they are discovered.

## Flaky tests

| Fix PR | Test | Failed Run |
| - | - | - |
| #10128 | https://github.com/Azure/azure-functions-host/blob/1bf081347aa0ff3fa02e846802ae330611d192ff/test/WebJobs.Script.Tests/FileMonitoringServiceTests.cs#L286 | N/A | 
| TBD | https://github.com/Azure/azure-functions-host/blob/1bf081347aa0ff3fa02e846802ae330611d192ff/test/WebJobs.Script.Tests/Security/SecretManagerTests.cs#L938 | [Failed run](https://azfunc.visualstudio.com/Azure%20Functions/_build/results?buildId=165675&view=logs&j=d090b295-8b65-5b33-df45-4b0e4ec519f8&t=c74105ee-1189-504a-7f57-81aa2e7f3871&s=96ac2280-8cb4-5df5-99de-dd2da759617d) |
| #10145 | https://github.com/Azure/azure-functions-host/blob/1bf081347aa0ff3fa02e846802ae330611d192ff/test/WebJobs.Script.Tests/Workers/Http/HttpFunctionInvocationDispatcherTests.cs#L23 | [Failed run](https://azfunc.visualstudio.com/Azure%20Functions/_build/results?buildId=165720&view=logs&jobId=d090b295-8b65-5b33-df45-4b0e4ec519f8&j=d090b295-8b65-5b33-df45-4b0e4ec519f8&t=c74105ee-1189-504a-7f57-81aa2e7f3871) |
| #10146 | https://github.com/Azure/azure-functions-host/blob/1bf081347aa0ff3fa02e846802ae330611d192ff/test/WebJobs.Script.Tests/FunctionsSyncServiceTests.cs#L18 |
| TBD | https://github.com/Azure/azure-functions-host/blob/1bf081347aa0ff3fa02e846802ae330611d192ff/test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/WebJobsStartupEndToEndTests.cs |
[...]

PR #10146, opened 2024-05-09
Flaky test: FunctionsSyncServiceTests: Increase test due time & add tsc callback

liliankasem:

Part of #10135

### Pull request checklist

### Additional information

Increasing the test due time to something more reasonable


jviau:
Is there a way we can use mock callbacks to signal the expected method was called? This way we can increase the timeout significantly, but also end delaying once it is called. See https://github.com/Azure/azure-functions-host/pull/10128 for an example

liliankasem:
> Is there a way we can use mock callbacks to signal the expected method was called? This way we can increase the timeout significantly, but also end delaying once it is called. See #10128 for an example

Will take a stab at this approach

jviau (approved):

brettsam (approved):

PR #10148, opened 2024-05-10
[.NET 8 in-proc] Clean up to remove older release branches(inproc6,inproc8) from build YAML 

kshyju:
Removing old TFM specific branch names.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -11,8 +11,7 @@ pr:
     - dev
     - in-proc
     - release/4.*
-    - release/inproc6/4.*
-    - release/inproc8/4.*
+    - release/in-proc
 
 trigger:
   branches:
@@ -20,8 +19,7 @@ trigger:
     - dev
     - in-proc
     - release/4.*
-    - release/inproc6/4.*
-    - release/inproc8/4.*
+    - release/in-proc
 
 jobs:
 - job: InitializePipeline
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -11,17 +11,15 @@ pr:
...
-    - release/inproc6/4.*
-    - release/inproc8/4.*
+    - release/in-proc
 
 trigger:
   branches:
     include:
     - dev
     - in-proc
     - release/4.*
```
brettsam, line 21: we don't need `dev` or `release/4.*` here either, do we?

kshyju: We don't. Removed.

jviau: I recommend keeping trigger section consistent between all branches. So leaving `dev` and `release/4.*` is probably a good idea.

azure-pipelines.yml
```diff
@@ -8,20 +8,14 @@ variables:
 pr:
   branches:
     include:
```
jviau: contrary to the other suggestion, please add `dev` and `release/4.*` back to pr and trigger section. I want to keep these identical between all branches.

jviau (changes requested):

azure-pipelines.yml
kshyju: Okie, Added back "dev" and "release/4.*".

jviau (approved):

liliankasem (approved):

PR #10154, opened 2024-05-13
adding support for net8 assembly list in emg

### LINKED ISSUE #10112: Update EMG to use new runtime assemblies manifests

brettsam:
resolves #10112

Adding support for separate assembly cleanup lists for net6 and net8 in ExtensionsMetadataGenerator.

### Pull request checklist

### DIFF
--- a/tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator/BuildTasks/RemoveRuntimeDependencies.cs
+++ b/tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator/BuildTasks/RemoveRuntimeDependencies.cs
@@ -18,9 +18,21 @@ namespace Microsoft.Azure.WebJobs.Script.ExtensionsMetadataGenerator.BuildTasks
         private const string Net6File = ".runtimeAssemblies-net6.txt";
         private const string Net8File = ".runtimeAssemblies-net8.txt";
 
+        // everything up until net6 maps to the net6 assembly list
+        // net8 only maps to the net8 assembly list
+        // anything else (such as future versions) are unsupported
         private static readonly IDictionary<string, string> _runtimeToAssemblyFileMap = new Dictionary<string, string>()
         {
+            { "netstandard2.0", Net6File },
+            { "netstandard2.1", Net6File },
+            { "netcoreapp2.1", Net6File },
+            { "netcoreapp2.2", Net6File },
+            { "netcoreapp3.0", Net6File },
+            { "netcoreapp2.0", Net6File },
+            { "netcoreapp3.1", Net6File },
+            { "net5.0", Net6File },
             { "net6.0", Net6File },
+
             { "net8.0", Net8File }
         };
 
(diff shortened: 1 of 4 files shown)


tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator/BuildTasks/RemoveRuntimeDependencies.cs
```diff
@@ -32,8 +38,14 @@ public override bool Execute()
                 ignoreFilesSet.Add(item.ItemSpec);
             }
 
+            if (!_runtimeToAssemblyFileMap.TryGetValue(TargetFramework, out var assemblyListFileName))
+            {
+                Log.LogError($"The TargetFramework '{TargetFramework}' is not supported in this project.");
```
fabiocav: Wondering if we could run into problems here. We might still have users targeting different TFMs other than net6.0 (e.g. .NET standard). We could still enforce this here, but wondering if we should improve the message to list the supported targets. Thoughts?

tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator/ExtensionsMetadataGenerator.csproj
```diff
@@ -37,11 +37,11 @@
 
   <Target Name="PackReferenceAssemblies">
     <ItemGroup>
-      <Content Include="$(OutputPath)\netstandard2.0\generator\*">
-        <Pack>true</Pack>
+      <Content Include="$(ArtifactsPath)\bin\ExtensionsMetadataGenerator\release_netstandard2.0\generator\*">
+        <Pack>true</Pack>
```
fabiocav: nit: if making other changes; space.

liliankasem:
I separated the EMG tests from the rest of the unit tests because of multi-targeting issues; can you group them back together and see if things work properly now?

https://github.com/Azure/azure-functions-host/blob/cea988b1753e2de4ec64b35cb55f31203d453b8e/azure-pipelines.yml#L322-L329

jviau:
I am curious - how much work would it be to rewrite this as an .netstandard2.0 msbuild task?

brettsam:
> I am curious - how much work would it be to rewrite this as an .netstandard2.0 msbuild task?

@jviau -- Which part? The ExtensionsMetadataGenerator assembly (the one with the Tasks) currently multi-targets netstandard2.0 and net46. I think we had to do that because of how VS loads the tasks and we needed the AppDomainIsolatedTask for that... @fabiocav may remember the exact details.

brettsam:
> I separated the EMG tests from the rest of the unit tests because of multi-targeting issues; can you group them back together and see if things work properly now?

@liliankasem -- I reverted this, added them back, saw the build continue to fail. Then I realized... oh wait, splitting this out is actually the right approach. We'll always build this solution as net6 and it's unrelated to the host framework. So, I'm keeping this split out.

liliankasem:
> > I separated the EMG tests from the rest of the unit tests because of multi-targeting issues; can you group them back together and see if things work properly now?
> 
> @liliankasem -- I reverted this, added them back, saw the build continue to fail. Then I realized... oh wait, splitting this out is actually the right approach. We'll always build this solution as net6 and it's unrelated to the host framework. So, I'm keeping this split out.

Thank you for checking!

fabiocav (approved):

liliankasem (approved):

PR #10167, opened 2024-05-15
[1ES] Migrate Diagnostic Events to Azure.Data.Tables

### LINKED ISSUE #10165: Migrate Diagnostic Events to Azure.Data.Tables

cjaliaga:
This PR migrates the Diagnostic Events to use Azure.Data.Tables from Microsoft.Azure.Cosmos.Table

resolves #10165 

### Pull request checklist

    * Otherwise: Link to backporting PR - #10218 


src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -55,18 +55,17 @@ public class DiagnosticEventTableStorageRepository : IDiagnosticEventRepository,
...
+        internal TableServiceClient TableClient
         {
             get
             {
                 if (!_environment.IsPlaceholderModeEnabled() && _tableClient == null)
                 {
                     string storageConnectionString = _configuration.GetWebJobsConnectionString(ConnectionStringNames.Storage);
                     if (!string.IsNullOrEmpty(storageConnectionString)
-                        && CloudStorageAccount.TryParse(storageConnectionString, out CloudStorageAccount account))
+                        && Microsoft.Azure.Storage.CloudStorageAccount.TryParse(storageConnectionString, out _))
```
jviau: I think you can switch to a `try/catch`, remove `Microsoft.Azure.Storage.CloudStorageAccount.TryParse`, and add the log statement in the `catch` block

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -134,30 +133,36 @@ private async Task PurgePreviousEventVersions()
 
                     foreach (var table in tables)
                     {
-                        var tableRecords = await table.ExecuteQuerySegmentedAsync(new TableQuery<DiagnosticEvent>(), null);
+                        var tableQuery = table.QueryAsync<DiagnosticEvent>(cancellationToken: default);
+                        var tableRecords = new List<DiagnosticEvent>();
+
+                        await foreach (var entity in tableQuery)
```
jviau: Can we refactor this method to avoid adding to the list? Is there a way to perform all of this in the `await foreach`?

src/WebJobs.Script.WebHost/Helpers/TableStorageHelpers.cs
```diff
@@ -74,38 +73,49 @@ internal static void QueueBackgroundTablePurge(CloudTable currentTable, CloudTab
...
-        internal static async Task DeleteOldTablesAsync(CloudTable currentTable, CloudTableClient tableClient, string tableNamePrefix, ILogger logger)
+        internal static async Task DeleteOldTablesAsync(TableClient currentTable, TableServiceClient tableClient, string tableNamePrefix, ILogger logger)
         {
             var tablesToDelete = await ListOldTablesAsync(currentTable, tableClient, tableNamePrefix);
             logger.LogDebug($"Deleting {tablesToDelete.Count()} old tables.");
             foreach (var table in tablesToDelete)
             {
                 logger.LogDebug($"Deleting table '{table.Name}'");
-                await table.DeleteIfExistsAsync();
+                await tableClient.DeleteTableAsync(table.Name);
```
jviau, line 86: Will this throw if it does not exist? Do we need a catch statement or is there a no-throw overload?

src/WebJobs.Script.WebHost/Helpers/TableStorageHelpers.cs
```diff
@@ -74,38 +73,49 @@ internal static void QueueBackgroundTablePurge(CloudTable currentTable, CloudTab
...
+        internal static async Task<IEnumerable<TableClient>> ListTablesAsync(TableServiceClient tableClient, string tableNamePrefix)
         {
-            List<CloudTable> tables = new List<CloudTable>();
-            TableContinuationToken continuationToken = null;
+            // Azure.Data.Tables doesn't have a direct way to list tables with a prefix
+            AsyncPageable<TableItem> tablesQuery = tableClient.QueryAsync(p => p.Name.CompareTo(tableNamePrefix) >= 0);
+            var tables = new List<TableClient>();
 
-            do
+            await foreach (var page in tablesQuery.AsPages())
```
jviau: Can the non-page version be used? It still pages behind the scene, but lets you avoid an extra `foreach` level.

```suggestion
            await foreach (var table in tablesQuery)
```

src/WebJobs.Script.WebHost/Helpers/TableStorageHelpers.cs
```diff
@@ -74,38 +73,49 @@ internal static void QueueBackgroundTablePurge(CloudTable currentTable, CloudTab
...
+                {
+                    tables.Add(tableClient.GetTableClient(table.Name));
+                }
             }
-            while (continuationToken != null);
 
             return tables;
         }
+
+        internal static async Task<bool> CheckIfTableExistsAsync(TableServiceClient tableClient, string tableName)
```
jviau: This isn't used, but also checking for existence then operating on the table could be a race condition: what if the table is deleted between the two statements?

src/WebJobs.Script.WebHost/Helpers/TableStorageHelpers.cs
cjaliaga: This is the only overload that I've seen. On my tests If it doesn't exist it returns a 404 as part of the response, but it's not throwing.

src/WebJobs.Script.WebHost/Helpers/TableStorageHelpers.cs
cjaliaga: I agree, on those situations we would have a conflict/error when using it. I added it to migrate something that we do in a test but then didn't go this way and forgot to remove it.

src/WebJobs.Script.WebHost/Helpers/TableStorageHelpers.cs
jviau: So long as it does not throw.

src/WebJobs.Script.WebHost/Helpers/TableStorageHelpers.cs
```diff
@@ -18,27 +20,24 @@ internal static string GetRowKey(DateTime now)
...
             int attempt = 0;
             do
             {
                 try
                 {
-                    if (!table.Exists())
-                    {
-                        return await table.CreateIfNotExistsAsync();
-                    }
+                    return (await table.CreateIfNotExistsAsync()).Value is not null;
```
brettsam: You need to keep that check for `!table.Exists()` -- without it you can end up with tons of AppInsights logs with errors (and thus issues/CRIs). We added this check to prevent those. https://github.com/Azure/azure-functions-host/pull/9555

cjaliaga:
deps.json unit test:

```


Previous file: D:\a_work\1\s\out\bin\WebJobs.Script.Tests\debug\Microsoft.Azure.WebJobs.Script.WebHost.deps.json
New file:      D:\a_work\1\s\out\bin\WebJobs.Script.WebHost\debug\Microsoft.Azure.WebJobs.Script.WebHost.deps.json


Changed:


Removed:


Added:
- Azure.Data.Tables.dll: 12.8.3.0/12.800.324.10602
```

src/WebJobs.Script.WebHost/WebJobs.Script.WebHost.csproj
```diff
@@ -62,6 +62,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Azure.Data.Tables" Version="12.8.3" />
```
jviau, line 65: What work is remaining, if any, to remove `Microsoft.Azure.Cosmos.Table` package reference?

jviau (commented):
Looks good. Just @brettsam's comment regarding checking table exists before creation to minimize telemetry. Also wondering when we can remove the deprecated `Microsoft.Azure.Cosmos.Table` package.

src/WebJobs.Script.WebHost/WebJobs.Script.WebHost.csproj
cjaliaga: We also need to migrate the Scale Metrics https://github.com/Azure/azure-functions-host/issues/10166. I thought on splitting in 2 different PRs.

cjaliaga:
> Looks good. Just @brettsam's comment regarding checking table exists before creation to minimize telemetry.

The Azure.Data.Tables SDK after 12.7.0 seems to have that issue resolved. I'm double checking that's actually the case: https://github.com/Azure/azure-sdk-for-net/pull/30849

src/WebJobs.Script.WebHost/Helpers/TableStorageHelpers.cs
cjaliaga: In both cases, 12.8 and 12.6, the conflict that comes from the service and the dependency with a 409 is emitted.  The difference is that in 12.8 is emitted as _informational_ versus 12.6 where is emitted as _warning_ (failed scope):

**12.6:**
```
[Warning] Azure-Core: Error response [955c902a-a10e-44fd-8d53-7c1490ab561f] 409 Conflict (00.5s)
Cache-Control:no-cache
Transfer-Encoding:chunked
Server:Windows-Azure-Table/1.0,Microsoft-HTTPAPI/2.0
x-ms-request-id:783f9c24-e002-001a-0fb2-b1b240000000
x-ms-client-request-id:955c902a-a10e-44fd-8d53-7c1490ab561f
x-ms-version:REDACTED
X-Content-Type-Options:REDACTED
Date:Wed, 29 May 2024 10:28:38 GMT
Content-Type:application/json; odata=minimalmetadata; streaming=true; charset=utf-8
```
**12.8:**
```
[Informational] Azure-Core: Response [d0228e9f-9e96-481a-b2eb-513c9764210d] 409 Conflict (00.4s)
Cache-Control:no-cache
Transfer-Encoding:chunked
Server:Windows-Azure-Table/1.0 Microsoft-HTTPAPI/2.0
x-ms-request-id:b51963d4-3002-0031-2bb5-b1c6f8000000
x-ms-client-request-id:d0228e9f-9e96-481a-b2eb-513c9764210d
x-ms-version:REDACTED
X-Content-Type-Options:REDACTED
Preference-Applied:REDACTED
Date:Wed, 29 May 2024 10:43:36 GMT
Content-Type:application/json;odata=minimalmetadata;streaming=true;charset=utf-8
```
In App Insights translates to a success call (even if 409) in 12.8 versus a failed dependency call in 12.6.

**12.6:**

image

**12.8:**

image

I'm OK with both options: checking if exists (adding an extra successful dependency call to list the tables that should end in 200) or letting the 409 as a successful call. That said, with 12.8 it doesn't seem to appear as a warning so I think we should be fine. 

Let me know what you think @brettsam @jviau

jviau: @cjaliaga how often is this call made? And what is the telemetry item you have shared - dependency or trace? I wonder how this will look if customer (or we) eventually includes this in a metric. There is no separate success/failed attribute from status code, so there wouldn't be a way to represent "409, but expected". In this case it may be better to make the GET call.

cjaliaga: Every time we flush the logs so it might be fairly frequent. The telemetry item was a trace and the screenshot shows a dependency item as well. I agree with your reasoning, let's check if it already exists.

jviau (approved):

test/WebJobs.Script.Tests/Microsoft.Azure.WebJobs.Script.WebHost.deps.json
```diff
@@ -28,11 +29,9 @@
           "Microsoft.Azure.WebJobs.Script": "4.1033.6",
           "Microsoft.Azure.WebJobs.Script.Grpc": "4.1033.6",
           "Microsoft.Azure.WebSites.DataProtection": "2.1.91-alpha",
-          "Microsoft.Identity.Client": "4.60.3",
```
fabiocav: This is private, but is this related to your change? I wouldn't expect it to be the case.

Did it cause a test failure flagging the removal? If not, can you please revert this change (and the SourceLink one) so we can clean this in a separate PR?

cjaliaga: It's not related to my change and indeed doesn't cause the deps test to fail. Just reverted them and let the addition of Azure.Data.Tables.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -134,31 +133,27 @@ private async Task PurgePreviousEventVersions()
...
-                            await table.DeleteIfExistsAsync();
-                            tableDeleted = true;
-                            continue;
-                        }
+                        var tableQuery = table.QueryAsync<DiagnosticEvent>(cancellationToken: default);
 
-                        // If the table does have EventVersion, query if it is an outdated version
-                        var eventVersionOutdated = tableRecords.Results.Any(record => string.Compare(DiagnosticEvent.CurrentEventVersion, record.EventVersion, StringComparison.Ordinal) > 0);
-                        if (eventVersionOutdated)
+                        await foreach (var record in tableQuery)
```
liliankasem, line 138: This makes it so that we have to loop each record which might be overkill here - is there no way to better query the table for the two conditions we need here? 

There is no schema so not every record in the table will be the same in terms of which columns they have

cjaliaga: We could change the implementation trying to see if we can query the server with both conditions and check what returns to perform the clean up, that looks cleaner indeed.

The reasoning behind this implementation is that I've tried to follow the same logic that we currently have to the extent that the new SDK allowed: 

Previously we would dump the whole table and then run a couple of Any() in memory to check for the condition(s). Now we query the table, letting the SDK paginate internally thanks to the IAsyncEnumerable and we check for both conditions on the same iteration, so in the best case scenario we could even end up pulling less records from the table than before 😊.

jviau: @liliankasem - this is a net improvement over what exists today. I am not sure if today is a bug or not, but we are using `ExecuteQuerySegmentedAsync` and never continuing the query. We only check the first page for the condition.

The changes here will continue the query checking later pages.

liliankasem: Thanks for the info, this helps :)

liliankasem (approved):

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
jviau: But that isn't to discount that including in a query condition is an even further improvement. But I believe that goes beyond the scope of this PR (which is to just migrate libraries with functionality parity)

cjaliaga: Completely agree with both of you. Let's focus on migrating everything to Azure.Data.Tables as soon as possible and then I can revisit this on a future PR applying @liliankasem  great suggestion.

fabiocav (approved):

PR #10198, opened 2024-05-30
Address 1ES CodeQL issues

jviau:

resolves #issue_for_this_pr

### Pull request checklist

### Additional information

CodeQL in 1ES is detecting this `.java` file and expecting a build action for it (which it will then hook into and perform analysis). This never happens so CodeQL times out. Since this is a sample that is not production code, the fix is to change the extension so codeql no longer detects it.

Additional changes in this PR are:

1. Optimize CodeQL in our jobs
    - Move 'source' languages (js, ts, python) to SDL stage only
    - Skip compiled languages for initialize-pipeline job.
2. Enable CodeQL on public build.
    - Our official and public builds run against different repos (Azure Devops, GitHub respectively). So we need it to be reported for _both_.
    - Not enabled for PR builds (as `Build.SourceBranch` will never match)
3. Adds a scheduled build for once a week to ensure SDL is reported

### DIFF
--- a/sample/Java/HttpTrigger/Function.java.txt
+++ b/sample/Java/HttpTrigger/Function.java.txt
@@ -1,4 +1,5 @@
 // This is a .txt extension intentionally so CodeQL does not expect and wait for a java compilation step.
+// Real java function apps would have this as .java extension.
 
 package Microsoft.Azure.WebJobs.Script.Tests.EndToEnd;
 
--- a/sample/Java/HttpTrigger/Function.java.txt
+++ b/sample/Java/HttpTrigger/Function.java.txt
@@ -1,3 +1,5 @@
+// This is a .txt extension intentionally so CodeQL does not expect and wait for a java compilation step.
+
 package Microsoft.Azure.WebJobs.Script.Tests.EndToEnd;
 
 import java.util.*;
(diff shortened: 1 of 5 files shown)


liliankasem (approved):
Assuming I will need these changes for the dotnet worker?

jviau:
> Assuming I will need these changes for the dotnet worker?

Most likely. I will be adding guidance to the docs soon.

UPDATE: CodeQL section is added to 1ESTips in our internal docs.

sample/Java/HttpTrigger/Function.java.txt
```diff
@@ -1,3 +1,6 @@
+// This is a .txt extension intentionally so CodeQL does not expect and wait for a java compilation step.
+// Real java function apps would have this as .java extension.
+
```
brettsam, line 3: We have (in the past) used these as direct samples for customers to reference. It's unfortunate if this is the only way we can do this. Is there not a way to ignore files or folders?

jviau: Our options are:

1. Rename the file (done in this PR)
2. Configure CodeQL to ignore Java. But reporting will _still_ expect a Java run, and so we will also need to file for an exception.
3. Add a Java build task in a separate stage to satisfy CodeQL.

Given the simplicity and that we have samples in other locations, I went for option 1. If this is an issue we can explorer the other options, but maybe at a later date?

brettsam (approved):

PR #10210, opened 2024-06-05
Updating end of interval Flex metrics publishing

mathewc:
Currently, we only calculate and emit activity metrics during periods of function inactivity. This works well when relatively short lived invocations start stop often during metering windows. However, this doesn't work well in cases where functions are long running (e.g. longer than the default 5ms metering interval), or when invocations are very frequent such that they overlap and there's rarely a period of inactivity. In such cases, we can go many metering intervals without emitting metrics, continuing to accumulate in memory. Eventually the correct duration will be emitted, but it's all shoved into a single interval. In addition to making metrics reporting spiky, this also makes it more likely that metrics may be lost, e.g. if the process dies abruptly.

Instead, we want to ensure that during any interval, we calculate and emit metrics for any outstanding activity in that interval. In this way, even if there are no idle periods, metrics are progressively emitted.

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
+++ b/src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
@@ -91,10 +91,19 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Metrics
             IsAlwaysReady = _environment.GetEnvironmentVariable(EnvironmentSettingNames.FunctionsAlwaysReadyInstance) == "1";
         }
 
-        internal async Task OnPublishMetrics()
+        internal async Task OnPublishMetrics(DateTime now)
         {
             try
             {
+                lock (_lock)
+                {
+                    if (ActiveFunctionCount > 0)
+                    {
+                        // at the end of an interval, we'll meter any outstanding activity up to the end of the interval
+                        MeterCurrentActiveInterval(now);
+                    }
+                }
+
                 if (FunctionExecutionCount == 0 && FunctionExecutionTimeMS == 0 && !IsAlwaysReady && !_metricsProvider.HasMetrics())
                 {
                     // no activity to report
@@ -137,7 +146,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Metrics
 
         private async void OnFunctionMetricsPublishTimer(object state)
         {
-            await OnPublishMetrics();
+            await OnPublishMetrics(DateTime.UtcNow);
         }
 
         private async Task PublishMetricsAsync(Metrics metrics)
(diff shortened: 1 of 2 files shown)


src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
```diff
@@ -312,6 +302,31 @@ public void AddFunctionExecutionActivity(string functionName, string invocationI
...
+            double elapsedMS = (now - adjustedActivityIntervalStart).TotalMilliseconds;
+            if (elapsedMS > 0)
+            {
+                // Accumulate the duration for this interval, applying the minimum
+                var duration = Math.Max(elapsedMS, _options.MinimumActivityIntervalMS);
+                FunctionExecutionTimeMS += RoundUp(duration, _options.MetricsGranularityMS);
+
+                // Move the high watermark timestamp forward to the point
+                // up to which we've metered
+                _activityIntervalHighWatermark = adjustedActivityIntervalStart.AddMilliseconds(duration);
```
paulbatum, line 327: Not an issue that was introduced in this review, but is it correct that we update the high watermark based on duration, rather than FunctionExecutionTimeMS (the rounded up duration based on metric granularity)?

paulbatum (commented):
This batch of changes looks good, but I did find one thing I'm not sure about in the existing logic.

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
mathewc: Yes, when MetricsGranularityMS was added, it should have been applied to the duration local var, so it was carried to the high water mark. I'll fix in this review.

paulbatum (approved):

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
```diff
@@ -91,10 +91,19 @@ internal void Initialize()
             IsAlwaysReady = _environment.GetEnvironmentVariable(EnvironmentSettingNames.FunctionsAlwaysReadyInstance) == "1";
         }
 
-        internal async Task OnPublishMetrics()
+        internal async Task OnPublishMetrics(DateTime now)
```
jviau, line 94: tiny tiny nit: always prefer `DateTimeOffset` over `DateTime`.

jviau (approved):

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
mathewc: Gonna leave this as is, thanks.

PR #10249, opened 2024-06-25
Adding a timeout to `IFunctionProvider.GetFunctionMetadataAsync` 

### LINKED ISSUE #10219: Deadlock in GetFunctionMetadataAsync when IFunctionMetadataProvider.GetFunctionMetadataAsync does not return
The `AddMetadataFromCustomProviders` method gets stuck if any of the `IFunctionMetadataProvider.GetFunctionMetadataAsync` does not return a result.

If you look at the code, we have 2 logging statements, line 217 and 228. When the `IFunctionMetadataProvider.GetFunctionMetadataAsync` does not return a result, the second log statement is never executed.

https://github.com/Azure/azure-functions-host/blob/d1067c54b27a8bb491867d692c9f7154f50552b8/src/WebJobs.Script/Host/FunctionMetadataManager.cs#L217-L230

The issue can be reproduced by adding the below test to `FunctionMetadataManagerTests.cs` and run it. The test pass confirms that the log message "1 functions found (Custom)" is not produced.

```csharp
[Fact]
public void FunctionMetadataManager_GetsMetadata_FromFunctionProviders_Deadlock()
{
    var functionMetadataCollection = new Collection<FunctionMetadata>();
    var mockFunctionErrors = new Dictionary<string, ImmutableArray<string>>();
    var mockFunctionMetadataProvider = new Mock<IFunctionMetadataProvider>();
    var mockFunctionProvider = new Mock<IFunctionProvider>();
    var workerConfigs = TestHelpers.GetTestWorkerConfigs();
    var testLoggerProvider = new TestLoggerProvider();
    var loggerFactory = new LoggerFactory();
    loggerFactory.AddProvider(testLoggerProvider);

    mockFunctionMetadataProvider.Setup(m => m.GetFunctionMetadataAsync(workerConfigs, SystemEnvironment.Instance, false)).Returns(Task.FromResult(new Collection<FunctionMetadata>().ToImmutableArray()));
    mockFunctionMetadataProvider.Setup(m => m.FunctionErrors).Returns(new Dictionary<string, ICollection<string>>().ToImmutableDictionary(kvp => kvp.Key, kvp => kvp.Value.ToImmutableArray()));

    functionMetadataCollection.Add(GetTestFunctionMetadata("somefile.dll", name: "anotherFunction"));

    // Mock GetFunctionMetadataAsync to never return a result.
    var tcs = new TaskCompletionSource<ImmutableArray<FunctionMetadata>>();
    mockFunctionProvider.Setup(m => m.GetFunctionMetadataAsync()).Returns(tcs.Task);

    mockFunctionProvider.Setup(m => m.FunctionErrors).Returns(mockFunctionErrors.ToImmutableDictionary());

    FunctionMetadataManager testFunctionMetadataManager = TestFunctionMetadataManager.GetFunctionMetadataManager(new OptionsWrapper<ScriptJobHostOptions>(_scriptJobHostOptions),
        mockFunctionMetadataProvider.Object, new List<IFunctionProvider>() { mockFunctionProvider.Object }, new OptionsWrapper<HttpWorkerOptions>(_defaultHttpWorkerOptions), loggerFactory, new TestOptionsMonitor<LanguageWorkerOptions>(TestHelpers.GetTestLanguageWorkerOptions()));

    // Run LoadFunctionMetadata in a separate Task to avoid blocking the test thread
    var loadFunctionMetadataTask = Task.Run(() => testFunctionMetadataManager.LoadFunctionMetadata());

    // wait 5 seconds for the task to complete
    Assert.False(loadFunctionMetadataTask.Wait(TimeSpan.FromSeconds(5)));
    var traces = testLoggerProvider.GetAllLogMessages();
    Assert.Equal(1, traces.Count(t => t.FormattedMessage.Contains("Reading functions metadata (Custom)")));
    Assert.DoesNotContain(traces, t => t.FormattedMessage.Contains("1 functions found (Custom)"));
}
```

kshyju:
resolves #10219 

The fix is to wrap the `GetFunctionMetadataAsync` with a timeout task so that if the `GetFunctionMetadataAsync` does not return within a specific time, we throw a timeout exception which the caller will handle.

    * Otherwise: Link to backporting PR : https://github.com/Azure/azure-functions-host/pull/10273
    * [x] Otherwise: Backport tracked by issue/PR #issue_or_pr - https://github.com/Azure/azure-functions-host/pull/10273

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -17,4 +17,4 @@
 - Ordered invocations are now the default (#10201)
 - Skip worker description if none of the profile conditions are met (#9932)
 - Fixed incorrect function count in the log message.(#10220)
-- Adding a timeout to `GetFunctionMetadataAsync` to prevent deadlocks (#10219)
+- Adding a timeout when retrieving function metadata from metadata providers (#10219)
--- a/release_notes.md
+++ b/release_notes.md
@@ -17,3 +17,4 @@
 - Ordered invocations are now the default (#10201)
 - Skip worker description if none of the profile conditions are met (#9932)
 - Fixed incorrect function count in the log message.(#10220)
+- Adding a timeout to `GetFunctionMetadataAsync` to prevent deadlocks (#10219)
--- a/src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
+++ b/src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
@@ -34,6 +34,7 @@ namespace Microsoft.Azure.WebJobs.Script
         public const string AppInsightsAgent = "APPLICATIONINSIGHTS_ENABLE_AGENT";
         public const string FunctionsExtensionVersion = "FUNCTIONS_EXTENSION_VERSION";
         public const string FunctionWorkerRuntime = "FUNCTIONS_WORKER_RUNTIME";
+        public const string MetadataProviderTimeoutInSeconds = "METADATA_PROVIDER_TIMEOUT_IN_SECONDS";
         public const string ContainerName = "CONTAINER_NAME";
         public const string WebsitePodName = "WEBSITE_POD_NAME";
         public const string LegionServiceHost = "LEGION_SERVICE_HOST";
--- a/src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
+++ b/src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
@@ -34,7 +34,6 @@ namespace Microsoft.Azure.WebJobs.Script
         public const string AppInsightsAgent = "APPLICATIONINSIGHTS_ENABLE_AGENT";
         public const string FunctionsExtensionVersion = "FUNCTIONS_EXTENSION_VERSION";
         public const string FunctionWorkerRuntime = "FUNCTIONS_WORKER_RUNTIME";
-        public const string MetadataProviderTimeoutInSeconds = "METADATA_PROVIDER_TIMEOUT_IN_SECONDS";
         public const string ContainerName = "CONTAINER_NAME";
         public const string WebsitePodName = "WEBSITE_POD_NAME";
         public const string LegionServiceHost = "LEGION_SERVICE_HOST";
--- a/src/WebJobs.Script/Host/FunctionMetadataManager.cs
+++ b/src/WebJobs.Script/Host/FunctionMetadataManager.cs
@@ -21,8 +21,9 @@ namespace Microsoft.Azure.WebJobs.Script
 {
     public class FunctionMetadataManager : IFunctionMetadataManager
     {
-        private const string _functionConfigurationErrorMessage = "Unable to determine the primary function script.Make sure atleast one script file is present.Try renaming your entry point script to 'run' or alternatively you can specify the name of the entry point script explicitly by adding a 'scriptFile' property to your function metadata.";
-        private const string _metadataProviderName = "Custom";
+        private const string FunctionConfigurationErrorMessage = "Unable to determine the primary function script.Make sure atleast one script file is present.Try renaming your entry point script to 'run' or alternatively you can specify the name of the entry point script explicitly by adding a 'scriptFile' property to your function metadata.";
+        private const string MetadataProviderName = "Custom";
+        private const int DefaultMetadataProviderTimeoutInSeconds = 30;
         private readonly IServiceProvider _serviceProvider;
         private IFunctionMetadataProvider _functionMetadataProvider;
         private bool _isHttpWorker;
@@ -79,7 +80,7 @@ namespace Microsoft.Azure.WebJobs.Script
         /// </summary>
         /// <param name="forceRefresh">Forces reload from all providers.</param>
         /// <param name="applyAllowList">Apply functions allow list filter.</param>
-        /// <param name="includeCustomProviders">Include any metadata provided by IFunctionProvider when loading the metadata</param>
+        /// <param name="includeCustomProviders">Include any metadata provided by IFunctionProvider when loading the metadata.</param>
         /// <returns> An Immutable array of FunctionMetadata.</returns>
         public ImmutableArray<FunctionMetadata> GetFunctionMetadata(bool forceRefresh, bool applyAllowList = true, bool includeCustomProviders = true, IList<RpcWorkerConfig> workerConfigs = null)
         {
@@ -188,7 +189,7 @@ namespace Microsoft.Azure.WebJobs.Script
             {
                 if (string.IsNullOrEmpty(functionMetadata.ScriptFile) && !_isHttpWorker && !functionMetadata.IsProxy() && _servicesReset)
                 {
-                    throw new FunctionConfigurationException(_functionConfigurationErrorMessage);
+                    throw new FunctionConfigurationException(FunctionConfigurationErrorMessage);
                 }
             }
             catch (FunctionConfigurationException exc)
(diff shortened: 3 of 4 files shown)


src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
```diff
@@ -34,6 +34,7 @@ public static class EnvironmentSettingNames
         public const string AppInsightsAgent = "APPLICATIONINSIGHTS_ENABLE_AGENT";
         public const string FunctionsExtensionVersion = "FUNCTIONS_EXTENSION_VERSION";
         public const string FunctionWorkerRuntime = "FUNCTIONS_WORKER_RUNTIME";
+        public const string MetadataProviderTimeoutInSeconds = "METADATA_PROVIDER_TIMEOUT_IN_SECONDS";
```
kshyju: This is only used to set the timeout value from tests. For normal use case, it will use the default value (30 seconds). I am open to suggestion to handle this in a better way.

jviau: I'd prefer avoiding something as broad as an environment variable for testing purposes. Can you use a settable static internal property? You can then set it from the test.

kshyju: I switched to a property in the latest iteration pushed.

src/WebJobs.Script/Host/FunctionMetadataManager.cs
```diff
@@ -56,6 +57,8 @@ public class FunctionMetadataManager : IFunctionMetadataManager
             };
         }
 
+        internal int MetadataProviderTimeoutInSeconds { get; set; } = DefaultMetadataProviderTimeoutInSeconds;
```
jviau, line 61: consider adding a comment to explain settable for testing only.

jviau (approved):

release_notes.md
```diff
@@ -17,3 +17,4 @@
 - Ordered invocations are now the default (#10201)
 - Skip worker description if none of the profile conditions are met (#9932)
 - Fixed incorrect function count in the log message.(#10220)
+- Adding a timeout to `GetFunctionMetadataAsync` to prevent deadlocks (#10219)
```
fabiocav: Can we update the note here so it's more end user friendly (i.e., doesn't require understanding of the codebase). Something along the lines of `Adding a timeout when retrieving function metadata from metadata providers`

src/WebJobs.Script/Host/FunctionMetadataManager.cs
```diff
@@ -214,12 +218,36 @@ private void LoadCustomProviderFunctions(List<FunctionMetadata> functionMetadata
...
+                var delayTask = Task.Delay(TimeSpan.FromSeconds(MetadataProviderTimeoutInSeconds));
+
+                var completedTask = Task.WhenAny(getFunctionMetadataFromProviderTask, delayTask).ContinueWith(t =>
+                {
+                    if (t.Result == getFunctionMetadataFromProviderTask)
+                    {
+                        if (getFunctionMetadataFromProviderTask.IsFaulted)
+                        {
+                            // Task completed but with an error
+                            _logger.LogWarning($"Failure in retrieving metadata from '{functionProvider.GetType().FullName}': {getFunctionMetadataFromProviderTask.Exception?.Flatten().ToString()}");
```
fabiocav: Pass the values as arguments into your log call:

Example:
``` csharp
_logger.LogWarning("Failure in retrieving metadata from '{typeName}': {exception}", functionProvider.GetType().FullName, getFunctionMetadataFromProviderTask.Exception?.Flatten());
````

File: src/WebJobs.Script/Host/FunctionMetadataManager.cs [hunk 5]
```diff
@@ -214,12 +218,36 @@ private void LoadCustomProviderFunctions(List<FunctionMetadata> functionMetadata
...
+                            return ImmutableArray<FunctionMetadata>.Empty;
+                        }
+                        else if (getFunctionMetadataFromProviderTask.IsCompletedSuccessfully)
+                        {
+                            return getFunctionMetadataFromProviderTask.Result;
+                        }
+                    }
+                    // Timeout case. getFunctionMetadataFromProviderTask was not the one that completed
+                    _logger.LogWarning($"Timeout or failure in retrieving metadata from '{functionProvider.GetType().FullName}'.");
```
fabiocav: Same comment about logging arguments.

src/WebJobs.Script/Host/FunctionMetadataManager.cs
```diff
@@ -214,12 +218,36 @@ private void LoadCustomProviderFunctions(List<FunctionMetadata> functionMetadata
...
+
+                var completedTask = Task.WhenAny(getFunctionMetadataFromProviderTask, delayTask).ContinueWith(t =>
+                {
+                    if (t.Result == getFunctionMetadataFromProviderTask)
+                    {
+                        if (getFunctionMetadataFromProviderTask.IsFaulted)
+                        {
+                            // Task completed but with an error
+                            _logger.LogWarning($"Failure in retrieving metadata from '{functionProvider.GetType().FullName}': {getFunctionMetadataFromProviderTask.Exception?.Flatten().ToString()}");
+                            return ImmutableArray<FunctionMetadata>.Empty;
```
fabiocav: Was the behavior difference considered here? Prior to this change, this would throw an exception that would bubble up in the metadata loading path (failing initialization). This will now log and move on, initializing with an empty list of functions.

src/WebJobs.Script/Host/FunctionMetadataManager.cs
fabiocav: Also, should this be an error instead?

src/WebJobs.Script/Host/FunctionMetadataManager.cs
fabiocav: Also, should this be an error?

src/WebJobs.Script/Host/FunctionMetadataManager.cs
```diff
@@ -214,12 +218,36 @@ private void LoadCustomProviderFunctions(List<FunctionMetadata> functionMetadata
...
 
             var functionProviderTasks = new List<Task<ImmutableArray<FunctionMetadata>>>();
+
             foreach (var functionProvider in functionProviders)
             {
-                functionProviderTasks.Add(functionProvider.GetFunctionMetadataAsync());
+                var getFunctionMetadataFromProviderTask = functionProvider.GetFunctionMetadataAsync();
+                var delayTask = Task.Delay(TimeSpan.FromSeconds(MetadataProviderTimeoutInSeconds));
+
+                var completedTask = Task.WhenAny(getFunctionMetadataFromProviderTask, delayTask).ContinueWith(t =>
```
fabiocav, line 230: It's a bit painful that this lives in `WebJobs.Script` (and that it targets .NET standard). `WaitAsync` would be a cleaner approach.

release_notes.md
kshyju: Updated as suggested.

src/WebJobs.Script/Host/FunctionMetadataManager.cs
kshyju: Removed the logging calls in the latest iteration (throwing an exception instead)

src/WebJobs.Script/Host/FunctionMetadataManager.cs
kshyju: Removed this behavior change. In the latest iteration, the timeout case will throw a `TimeOutException`

src/WebJobs.Script/Host/FunctionMetadataManager.cs
```diff
@@ -214,12 +218,27 @@ private void LoadCustomProviderFunctions(List<FunctionMetadata> functionMetadata
...
+                {
+                    if (t.Result == getFunctionMetadataFromProviderTask && getFunctionMetadataFromProviderTask.IsCompletedSuccessfully)
```
kshyju, line 232: Added tests for both timeout and exception cases.

fabiocav (approved):

PR #10255, opened 2024-06-27
Ensure extension RPC endpoints ready before processing gRPC messages

### LINKED ISSUE #10251: Service bus message actions throwing Grpc.Core.RpcException when sending multiple requests
### Description

This occurs with a C# isolated function, using .NET 8 and latest versions of all libraries.
 
For some reason, when using the message actions with batched service bus messages- some of them will fail with the following error: 
`Call failed with gRPC error status. Status code: 'Unimplemented', Message: 'Service is unimplemented.`

This looks similar to Azure/azure-functions-dotnet-worker#1974, however only occurs after multiple requests have been made within one function execution.

My best guess is some kind of limit, be it connections / pooling not being handled correctly.  

### Steps to reproduce

- Have a function app which receives a large batch of messages (experienced with ~100).
- Use message actions to deadletter, abandon or complete a large number of the messages.
- Have multiple function executions occurring in short time of each other.
   - Short cut could be to fill service bus with a decent number of messages, then call abandon on each message with a large retry set.
 
Some of the requests will then fail with the GRPC error.

My specific scenario was a function which:
- Receives a batch of messages.
- Tries to bulk insert messages into cosmos.
- Completes successful requests, and abandons failed requests
   - This is because comos can hit rate limits, and only allow a portion of the records to be created. Simply allowing the function to retry or complete would cause issues, as either requests would be duplicated or missed. 

Happy to provide more info / source code as needed.

jviau:

resolves #10251

### Pull request checklist

### Additional information

It is believed there is a race condition on startup between the first gRPC message coming in and extension RPC endpoints being registered. While `EndpointDataSource` has a change token system, leading us to believe that we can update endpoints post startup, there is one of two possibilities occurring:

1. Only a subset of messages fail due to a time gap between host/worker startup and the extension RPC endpoints being registered.
2. OR AspNetCore locks-in the available endpoints when `UseRouting` middleware is first encountered. In which case this host instance would not have extension RPC endpoints for its lifetime.

This fix should address both of those possible scenarios, as routing middleware is only initialized on first call. By ensuring we collect extension endpoints before the first routing occurs we avoid the race condition.

**RISK**: there is a small risk this could introduce a deadlock. If there is some dependency existing, or later introduced, which requires RPC communication between host and worker _before_ all extensions are loaded host-side, then this could cause a circular dependency and deadlock. I do not believe this is the case today, as testing has shown endpoints to be available immediately on startup before the first worker RPC message comes in.

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -17,3 +17,4 @@
 - Ordered invocations are now the default (#10201)
 - Skip worker description if none of the profile conditions are met (#9932)
 - Fixed incorrect function count in the log message.(#10220)
+- Fix race condition on startup with extension RPC endpoints not being available. (#10255)
--- a/src/WebJobs.Script.Grpc/Server/ExtensionsCompositeEndpointDataSource.cs
+++ b/src/WebJobs.Script.Grpc/Server/ExtensionsCompositeEndpointDataSource.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Threading;
+using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Routing;
 using Microsoft.Azure.WebJobs.Rpc.Core.Internal;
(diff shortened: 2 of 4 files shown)


fabiocav (approved):

brettsam (approved):

krishna-pandey-git:
Has this been rolled out yet, what needs to be done to use it ?

PR #10263, opened 2024-07-02
[in-proc] Updating FunctionsNetHost (dotnet isolated worker) to 1.0.9

kshyju:
in-proc port of https://github.com/Azure/azure-functions-host/pull/10262

### DIFF
--- a/test/WebJobs.Script.Tests/Workers/Rpc/RpcWorkerConfigFactoryTests.cs
+++ b/test/WebJobs.Script.Tests/Workers/Rpc/RpcWorkerConfigFactoryTests.cs
@@ -153,6 +153,7 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.Workers.Rpc
             var config = configBuilder.Build();
             var scriptSettingsManager = new ScriptSettingsManager(config);
             var testLogger = new TestLogger("test");
+            _testEnvironment.SetEnvironmentVariable(EnvironmentSettingNames.AzureWebsitePlaceholderMode, "1");
             using (var variables = new TestScopedSettings(scriptSettingsManager, testEnvVariables))
             {
                 var configFactory = new RpcWorkerConfigFactory(config, testLogger, _testSysRuntimeInfo, _testEnvironment, new TestMetricsLogger(), _testWorkerProfileManager);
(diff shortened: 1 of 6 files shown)


kshyju:
Many tests depend on the worker config provided by the referenced worker packages. The new [DNI worker config](https://github.com/Azure/azure-functions-dotnet-worker/pull/2552) requires a specific environment variable (`WEBSITE_PLACEHOLDER_MODE`) to be present. If this variable is absent, the worker will be skipped. Therefore, the tests need to include this environment variable. I am currently working on fixing them.

test/WebJobs.Script.Tests/Workers/Rpc/RpcWorkerConfigFactoryTests.cs
```diff
@@ -153,6 +153,7 @@ public void DefaultWorkerConfigs_Overrides_DefaultWorkerRuntimeVersion_AppSettin
             var config = configBuilder.Build();
             var scriptSettingsManager = new ScriptSettingsManager(config);
             var testLogger = new TestLogger("test");
+            _testEnvironment.SetEnvironmentVariable(EnvironmentSettingNames.AzureWebsitePlaceholderMode, "1");
```
fabiocav, line 156: Still want to make sure we've reviewed the test behavior here to ensure this won't invalidate the test.

fabiocav (approved):

test/WebJobs.Script.Tests/Workers/Rpc/RpcWorkerConfigFactoryTests.cs
kshyju: This test is primarily checking the worker configs are populated with right values, specifically the worker runtime verion. In that context, the behavior should be same for both placeholder mode and non-placeholder mode. By setting this env variable to the test, I effectively changed the test from "non-placeholder" mode test to "placeholder" mode test.

As discussed, offline, ideally, we should create 2 tests, one for placeholder and one for non-placeholder scenario. That is beyond the scope of this PR and can be addressed in future.

soninaren (approved):

brettsam (approved):

PR #10276, opened 2024-07-09
[1ES] Migrate Scale Metrics to Azure.Data.Tables

### LINKED ISSUE #10166: Migrate Scale Metrics to Azure.Data.Tables

### LINKED ISSUE #10359: Consolidate scale metrics table client with shared services
After https://github.com/Azure/azure-functions-host/issues/10129 is complete, we should move this to not be a separately initialized client. This should just use the same host shared services for storage.

cjaliaga:
This PR migrates the Scale Metrics table storage repository to use Azure.Data.Tables instead Microsoft.Azure.Cosmos.Table

resolves #10166 

### Pull request checklist

    * Otherwise: #10436 


cjaliaga:
Not ready yet, pending:

1) Check what's the best way to migrate the delegating handler usage as the new SDK by default will use a shared HttpClientTransport  (same HttpClient) and we might need to pass the HttpClient configured with our DelegatingHandler to the TableServiceClient.

2) Deploy the Private SiteExtension and test some scenarios

src/WebJobs.Script.WebHost/Scale/TableEntityConverter.cs
```diff
@@ -3,27 +3,27 @@
...
-using Microsoft.Azure.Cosmos.Table;
+using Azure;
+using Azure.Data.Tables;
 using Newtonsoft.Json.Linq;
 
 namespace Microsoft.Azure.WebJobs.Script.WebHost.Scale
 {
     /// <summary>
-    /// Class providing methods to convert between DynamicTableEntity and custom pocos
+    /// Class providing methods to convert between DynamicTableEntity and custom pocos.
```
mathewc: Still have a reference in comment to DynamicTableEntity

src/WebJobs.Script.WebHost/Scale/TableEntityConverter.cs
```diff
@@ -3,27 +3,27 @@
...
     /// </summary>
     internal static class TableEntityConverter
     {
-        public static DynamicTableEntity ToEntity(object o,
+        public static TableEntity ToEntity(object o,
```
mathewc, line 17: Just to confirm - the new SDK you're moving to doesn't have any built ins to handle POCO conversions? I added this converter class because there wasn't before, and we need a way to convert to/from POCOs that don't implement ITableEntity

src/WebJobs.Script.WebHost/Scale/TableEntityConverter.cs
```diff
@@ -33,108 +33,70 @@ internal static class TableEntityConverter
...
             return (TOutput)ToObject(typeof(TOutput), properties);
         }
 
-        public static object ToObject(Type type, IDictionary<string, EntityProperty> properties)
+        public static object ToObject(Type type, IDictionary<string, object> properties)
         {
             var jo = new JObject();
             foreach (var pair in properties)
             {
-                ApplyProperty(jo, pair.Key, pair.Value);
```
mathewc, line 66: How are you making sure there are no breaking changes in here? Wondering if data that has been written with the old code can be read without issues by this new code. We've had versioning issues in the past with this code.

src/WebJobs.Script.WebHost/Scale/TableStorageScaleMetricsRepository.cs
```diff
@@ -36,7 +38,7 @@ public class TableStorageScaleMetricsRepository : IScaleMetricsRepository
         private readonly ILogger _logger;
         private readonly int _tableCreationRetries;
         private readonly IDelegatingHandlerProvider _delegatingHandlerProvider;
-        private CloudTableClient _tableClient;
+        private TableServiceClient _tableClient;
```
mathewc: Rename member to _tableServiceClient

src/WebJobs.Script.WebHost/Scale/TableStorageScaleMetricsRepository.cs
```diff
@@ -54,40 +56,38 @@ public TableStorageScaleMetricsRepository(IConfiguration configuration, IHostIdP
             _delegatingHandlerProvider = delegatingHandlerProvider ?? throw new ArgumentNullException(nameof(delegatingHandlerProvider));
         }
 
-        internal CloudTableClient TableClient
+        internal TableServiceClient TableClient
```
mathewc: Rename the property to TableServiceClient

src/WebJobs.Script.WebHost/Scale/TableStorageScaleMetricsRepository.cs
```diff
@@ -54,40 +56,38 @@ public TableStorageScaleMetricsRepository(IConfiguration configuration, IHostIdP
...
             {
-                if (_tableClient == null)
+                if (_tableClient is null)
                 {
                     string storageConnectionString = _configuration.GetWebJobsConnectionString(ConnectionStringNames.Storage);
-                    if (!string.IsNullOrEmpty(storageConnectionString) &&
-                        CloudStorageAccount.TryParse(storageConnectionString, out CloudStorageAccount account))
-                    {
-                        var restConfig = new RestExecutorConfiguration { DelegatingHandler = _delegatingHandlerProvider.Create() };
-                        var tableClientConfig = new TableClientConfiguration { RestExecutorConfiguration = restConfig };
```
mathewc, line 68: You're removing the delegating handler and other logic (potentially leaving DefaultDelegatingHandlerProvider an unused type?). @brettsam added this logic - wondering what issues may arise if we remove this?

src/WebJobs.Script.WebHost/Scale/TableStorageScaleMetricsRepository.cs
```diff
@@ -293,85 +263,54 @@ internal static TableOperation CreateMetricsInsertOperation(ScaleMetrics metrics
...
 
-        internal async Task<IEnumerable<DynamicTableEntity>> ExecuteQuerySafeAsync(CloudTable metricsTable, TableQuery query)
+        internal async Task<IEnumerable<TableEntity>> ExecuteQuerySafeAsync(TableClient metricsTable, Expression<Func<TableEntity, bool>> query)
         {
             try
             {
-                return await ExecuteQueryWithContinuationAsync(metricsTable, query);
+                List<TableEntity> results = new List<TableEntity>();
+
+                await foreach (var result in metricsTable.QueryAsync(query))
```
mathewc: So the manual paging we were doing previously is no longer required?

src/WebJobs.Script.WebHost/Scale/TableStorageScaleMetricsRepository.cs
cjaliaga: This is still pending, I initially removed it to test everything else. I wanted to discuss what's the best way of incorporating the DelegatingHandler into the new SDK. By default it will use a shared HttpClientTransport (same HttpClient) for all the clients.

From my initial investigation we would need to provide an HttpClient that uses our DelegatingHandler when we instantiate the TableServiceClient via the TableClientOptions. 

I also think that we should check the way we instantiate these clients, same as for Diagnostic Events, specially if we need to enable Managed Identity support instead of using the connection string for these scenarios. That might need to be discussed separately on a different PR.

https://github.com/Azure/azure-functions-host/blob/49c49c9ea2153515a2d7a2de71a43f4dd80cb856/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs#L68

src/WebJobs.Script.WebHost/Scale/TableStorageScaleMetricsRepository.cs
cjaliaga: Based on conversations with @jviau the new SDK should paginate automatically. See https://github.com/Azure/azure-functions-host/pull/10167#discussion_r1604009472.

jviau: Yes - the track 2 SDKs have custom `IAsyncEnumerable` implementations that will handle pagination for you.

mathewc:
> Not ready yet, pending:
> 
> 1. Check what's the best way to migrate the delegating handler usage as the new SDK by default will use a shared HttpClientTransport  (same HttpClient) and we might need to pass the HttpClient configured with our DelegatingHandler to the TableServiceClient.
> 2. Deploy the Private SiteExtension and test some scenarios

I see. When a PR isn't ready, it's good to mark it as a draft so reviewers know. I converted it to draft just now.

src/WebJobs.Script.WebHost/Scale/TableEntityConverter.cs
cjaliaga: I've been running tests with 3 monitors (storage queue, eventhub and service bus topic) with runtime scaling:

1) Configured them on 4.34
2) Added messages to trigger and scale -> Scale samples aggregated on the Table Storage
3) Platform scaling
4) Stopped and deployed the private site extension to the app
5) Continued executing functions, read scale metrics and scaled back out to continue processing

@mathewc is there an specific extension that you believe I should test (or maybe all the supported)? Anything that failed in the past?

src/WebJobs.Script.WebHost/Scale/TableStorageScaleMetricsRepository.cs
cjaliaga: I ended up simplifying a bit the `DelegatingHandler` and passing it to the `TableServiceClient `via `ClientOptions` (just for consumption), at least for now. 

I think we should better discuss adding a new `TableStorageProvider` and replace the manual creation of the clients (here and for DiagnosticEvents) in a follow-up PR, for example things such: do we need to add a `IAzureTableStorageProvider` in the [Webjobs.Host.Storage package](https://github.com/Azure/azure-webjobs-sdk/tree/dev/src/Microsoft.Azure.WebJobs.Host.Storage)?

@fabiocav

src/WebJobs.Script.WebHost/Scale/TableEntityConverter.cs
jviau: I don't think it does. You need to use `TableEntity` or implement `ITableEntity` still.

src/WebJobs.Script.WebHost/Scale/TableStorageScaleMetricsRepository.cs
jviau: @cjaliaga using tables is an implementation detail here, we wouldn't want to expose a public contract just for providing the clients.

Instead, there is already a public contract for this: `IAzureClientFactory<TClient>` from `Microsoft.Extensions.Azure`. We can evaluate using this to extract client construction to DI in a later PR.

Looking at the existing delegating handler, the comment is that it is "because connections are limited". I think what they are referring to in connection pooling: re-use existing HttpClients instead of instantiating new ones repeatedly. As @cjaliaga mentioned, the new track2 clients have made great improvements here. I do wonder if it is even needed anymore?

src/WebJobs.Script.WebHost/Scale/TableStorageScaleMetricsRepository.cs
```diff
@@ -54,40 +58,49 @@ public TableStorageScaleMetricsRepository(IConfiguration configuration, IHostIdP
...
+                            options = new TableClientOptions
+                            {
+                                Transport = new HttpClientTransport(handler)
+                            };
+                        }
+
+                        _tableServiceClient = new TableServiceClient(storageConnectionString, options);
                     }
-                    else
+                    catch (Exception ex)
```
jviau: Whenever catching plain `Exception` type, recommend checking it is not fatal:

```suggestion
                    catch (Exception ex) when (!ex.IsFatal())
```

src/WebJobs.Script.WebHost/Scale/TableStorageScaleMetricsRepository.cs
```diff
@@ -195,28 +209,28 @@ internal Type GetMonitorScaleMetricsTypeOrNull(IScaleMonitor monitor)
             return null;
         }
 
-        internal void LogStorageException(StorageException ex)
+        internal void LogStorageException(RequestFailedException ex)
         {
-            _logger.LogError(ex, $"An unhandled storage exception occurred when reading/writing scale metrics: {ex.ToString()}");
+            _logger.LogError(ex, $"An unhandled storage exception occurred when reading/writing scale metrics: {ex}");
```
jviau: we shouldn't be including exceptions in the log message body. However, I am not sure if it is a breaking change to remove it.

If we do remove it, should format some properties from the `RequestFailedException` onto the message (using log templating, not string interpolation).

src/WebJobs.Script.WebHost/Scale/TableStorageScaleMetricsRepository.cs
```diff
@@ -293,85 +276,54 @@ internal static TableOperation CreateMetricsInsertOperation(ScaleMetrics metrics
...
-            do
-            {
-                var results = await TableClient.ListTablesSegmentedAsync(TableNamePrefix, continuationToken);
-                continuationToken = results.ContinuationToken;
-                tables.AddRange(results.Results);
-            }
-            while (continuationToken != null);
-
-            return tables;
+            return await ExecuteQuerySafeAsync(metricsTable, entity => entity.PartitionKey == hostId && string.Compare(entity.RowKey, ticks) < 0);
```
jviau: How often is this invoked? If its ran often enough, we should consider formatting a query by string to avoid the expression parsing.

src/WebJobs.Script.WebHost/Scale/TableStorageScaleMetricsRepository.cs
cjaliaga: @jviau for what I've seen by default it will use the a shared `HttpClient` with a default `SocketsHttpHandler`, where the `MaxConnectionsPerServer` is `int.MaxValue` unless there's another runtime setting that changes that default. If you run it on .NET Framework the SDKs will increase `MaxConnectionsPerServer` to 50 only in case they are using the default value of 2. 

I miss some context about the reason of adding this specific handler with the old SDK, hopefully @brettsam could add some insights about that: was it because we would use more than 50 concurrent connections? Or maybe the SDK was spawning new clients but the `DelegatingHandler` was shared between all of them so we could control the max number that way?

Hopefully we could remove it and let the default behavior for all the hosting plans.

src/WebJobs.Script.WebHost/Scale/TableStorageScaleMetricsRepository.cs
cjaliaga: It's fairly often. I replaced it with the string flavor using the `CreateQueryFilter` helper.

src/WebJobs.Script.WebHost/Scale/TableStorageScaleMetricsRepository.cs
cjaliaga: An example of the exception message (when the failure comes from the service) would be this:

```
The table specified does not exist.
RequestId:0e0308a0-b002-0025-4595-e3385b000000
Time:2024-07-31T22:05:25.9397718Z
Status: 404 (Not Found)
ErrorCode: TableNotFound

Content:
{"odata.error":{"code":"TableNotFound","message":{"lang":"en-US","value":"The table specified does not exist.\nRequestId:0e0308a0-b002-0025-4595-e3385b000000\nTime:2024-07-31T22:05:25.9397718Z"}}}

Headers:
Transfer-Encoding: chunked
Server: Windows-Azure-Table/1.0 Microsoft-HTTPAPI/2.0
x-ms-request-id: 0e0308a0-b002-0025-4595-e3385b000000
x-ms-client-request-id: 7d830039-2628-46b8-91f5-1795b250ab34
x-ms-version: REDACTED
X-Content-Type-Options: REDACTED
Date: Wed, 31 Jul 2024 22:05:25 GMT
Content-Type: application/json;odata=minimalmetadata;streaming=true;charset=utf-8
```

Some of these values can be retrieved as `RequestFailedException` properties, such `Status `or `ErrorCode`. Others like RequestId or the specific message are part of the response `RequestFailedException.GetRawResponse()`: as properties 
like `RequestId`,  and others require to read the JSON content and parse it. `GetRawResponse` might return null when the failure is part of a batch.

More than happy to go with structured logging, my only concern is that we would need to be cautious to ensure that as part of exception processing we don't add too much overhead and also that we don't miss details when the reason is not a service issue itself: the inner could be `SocketException`, to name one example.

src/WebJobs.Script.WebHost/Scale/TableStorageScaleMetricsRepository.cs
```diff
@@ -54,40 +57,49 @@ public TableStorageScaleMetricsRepository(IConfiguration configuration, IHostIdP
...
+                            };
+                        }
+
+                        _tableServiceClient = new TableServiceClient(storageConnectionString, options);
                     }
-                    else
+                    catch (Exception ex) when (!ex.IsFatal())
                     {
-                        _logger.LogError("Azure Storage connection string is empty or invalid. Unable to read/write scale metrics.");
+                        _logger.LogError(ex, "Azure Storage connection string is empty or invalid. Unable to read/write scale metrics.");
```
fabiocav: No need to address this as part of this PR as it is beyond the scope of what is being tracked, but we need to flag this for refactoring. Currently, whenever this property is called, you'll end up with this error log if there's an issue creating the client, as we'll try to create it every time since the condition only checks for `null` and that that state won't change if there's a problem.

src/WebJobs.Script.WebHost/Scale/TableStorageScaleMetricsRepository.cs
fabiocav: It would be good to have clarity on the requirement and need for the behavior. We can defer any changes to when we move to the client factory, but this will be happening soon.

fabiocav (approved):

src/WebJobs.Script.WebHost/Scale/TableStorageScaleMetricsRepository.cs
```diff
@@ -30,64 +29,51 @@ public class TableStorageScaleMetricsRepository : IScaleMetricsRepository
...
-            _delegatingHandlerProvider = delegatingHandlerProvider ?? throw new ArgumentNullException(nameof(delegatingHandlerProvider));
         }
 
-        internal CloudTableClient TableClient
+        internal TableServiceClient TableServiceClient
         {
             get
             {
-                if (_tableClient == null)
+                if (_tableServiceClient is null && !_azureTableStorageProvider.TryCreateHostingTableServiceClient(out _tableServiceClient))
```
cjaliaga, line 58: It would be really good here to do a _try create and log failure just once_, to avoid logging the error every time the property is accessed. 

`TableStorageScaleMetricsRepository` is registered as singleton, `ScriptHost `level if I'm not mistaken: https://github.com/Azure/azure-functions-host/blob/17bad4fd6ac4a00a0d703d8dbcdbf49b221dcfe5/src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs#L149

Does that mean that specialization wouldn't affect here and we're safe assuming that the `IConfiguration` used by `IAzureTableStorageProvider` has the right entries any time this property is accessed? Just trying to avoid having a repository instance not initialized due to checking and marking as invalid too soon.

cc: @fabiocav

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
```diff
@@ -220,7 +221,7 @@ public static void AddWebJobsScriptHost(this IServiceCollection services, IConfi
             services.TryAddSingleton<IJobHostMiddlewarePipeline>(s => DefaultMiddlewarePipeline.Empty);
 
             // Add AzureBlobStorageProvider to WebHost (also needed for ScriptHost)
```
satvu: nit: this comment is no longer accurate/relevant

src/WebJobs.Script/WebJobs.Script.csproj
```diff
@@ -40,6 +40,7 @@
   <ItemGroup>
     <!-- Dependencies needed for Storage Providers -->
     <PackageReference Include="Azure.Core" Version="1.38.0" />
+    <PackageReference Include="Azure.Data.Tables" Version="12.8.3" />
```
kshyju, line 43: Should we use the latest version of this package? https://www.nuget.org/packages/Azure.Data.Tables/12.9.0

src/WebJobs.Script/StorageProvider/TableServiceClientProvider.cs
```diff
@@ -0,0 +1,85 @@
...
+        public TableServiceClientProvider(AzureComponentFactory componentFactory, AzureEventSourceLogForwarder logForwarder) : base(componentFactory, logForwarder)
+        {
+        }
+
+        protected override TableServiceClient CreateClient(IConfiguration configuration, TokenCredential tokenCredential, TableClientOptions options)
+        {
+            // If connection string is present, it will be honored first; bypass creating a serviceUri
+            if (!IsConnectionStringPresent(configuration))
+            {
+                var serviceUri = configuration.Get<StorageServiceUriOptions>().GetTableServiceUri();
```
kshyju: Is it possible that `configuration.Get<StorageServiceUriOptions>()` could return `null`? Should we do a null check before calling the `GetTableServiceUri` method?

src/WebJobs.Script/StorageProvider/TableServiceClientProvider.cs
```diff
@@ -0,0 +1,85 @@
...
+using System;
+using System.Globalization;
+using Azure.Core;
+using Azure.Data.Tables;
+using Microsoft.Extensions.Azure;
+using Microsoft.Extensions.Configuration;
+
+namespace Microsoft.Azure.WebJobs.Script
+{
+    internal class TableServiceClientProvider : StorageClientProvider<TableServiceClient, TableClientOptions>
```
kshyju:
```suggestion
    internal sealed class TableServiceClientProvider : StorageClientProvider<TableServiceClient, TableClientOptions>
```

src/WebJobs.Script/StorageProvider/TableServiceClientProvider.cs
```diff
@@ -0,0 +1,85 @@
...
+        internal class StorageServiceUriOptions
+        {
+            private const string DefaultScheme = "https";
+            private const string DefaultEndpointSuffix = "core.windows.net";
+
+            /// <summary>
+            /// Gets or sets the resource URI for table storage. If this property is given explicitly, it will be
+            /// honored over the AccountName property.
+            /// </summary>
+            public string TableServiceUri { get; set; }
```
kshyju: Should we use `Uri` type here instead of string?

src/WebJobs.Script/StorageProvider/HostAzureTableStorageProvider.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        public HostAzureTableStorageProvider(IConfiguration configuration, ILogger<HostAzureTableStorageProvider> logger, AzureComponentFactory componentFactory, AzureEventSourceLogForwarder logForwarder, IDelegatingHandlerProvider delegatingHandlerProvider)
+        {
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            _delegatingHandlerProvider = delegatingHandlerProvider;
+            _tableServiceClientProvider = new TableServiceClientProvider(componentFactory, logForwarder);
+            Configuration = configuration ?? throw new ArgumentNullException(nameof(configuration));
+        }
+
+        public virtual IConfiguration Configuration { get; private set; }
```
kshyju: Should this need to be a public property? Looks like it's been used only inside the class. So we could keep it as a private variable. Remove it from the interface as well?

src/WebJobs.Script/StorageProvider/HostAzureTableStorageProvider.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        public virtual bool TryCreateTableServiceClientFromConnection(string connection, out TableServiceClient client)
```
kshyju: Do these methods needs to be `virtual` ? If not, I suggest removing them and marking the class as `sealed`. We can always unseal it if a need arises.

src/WebJobs.Script/StorageProvider/IAzureTableStorageProvider.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace Microsoft.Azure.WebJobs.Script
+{
+    public interface IAzureTableStorageProvider
+    {
+        IConfiguration Configuration { get; }
+
+        bool TryCreateHostingTableServiceClient(out TableServiceClient client);
+
+        bool TryCreateTableServiceClientFromConnection(string connection, out TableServiceClient client);
```
kshyju: Should we use the same name for both methods? T`ryCreateTableServiceClient`. The second one is just another overload which takes the connection string.

src/WebJobs.Script/StorageProvider/HostAzureTableStorageProvider.cs
```diff
@@ -0,0 +1,60 @@
...
+        {
+            var connectionToUse = connection ?? ConnectionStringNames.Storage;
```
kshyju: Since we have the other overload which uses the default ( which is `ConnectionStringNames.Storage`). Should we just throw if connection parameter is null here?

src/WebJobs.Script/StorageProvider/HostAzureTableStorageProvider.cs
```diff
@@ -0,0 +1,60 @@
...
+                        Transport = new HttpClientTransport(handler)
+                    };
+                }
+
+                client = _tableServiceClientProvider.Create(connectionToUse, Configuration, options);
+                return true;
+            }
+            catch (Exception e)
+            {
+                _logger.LogDebug($"Could not create TableServiceClient. Exception: {e}");
```
kshyju: Consider structured logging.

src/WebJobs.Script/WebJobs.Script.csproj
cjaliaga: WebHost is currently referencing 12.8.3 so I would like to avoid more changes to this PR whenever possible. Are we OK with bumping versions and normalize dependencies in a future PR?

kshyju: yea, Let's use the same existing version in that case for this PR.

src/WebJobs.Script/StorageProvider/HostAzureTableStorageProvider.cs
cjaliaga: Indeed not for our current usage. I based this provider in its equivalent for `BlobStorage`: `HostAzureTableStorageProvider` where the interface comes from `WebJobs.Host.Storage`. Let me apply your suggestions and we can always check in the future if there's a need to move this to the WebJobs layer and/or increase/modify the public surface.

src/WebJobs.Script/StorageProvider/IAzureTableStorageProvider.cs
cjaliaga: The reasoning behind that name was more semantic than technical. To be explicit about it would create a client used for our "Hosting" purposes a.k.a AzureWebJobsStorage client. I can clarify that in the summary of the method. What do you think?

kshyju: I am not sure the method name should indicate how the caller is going to use the TableServiceClient we return. So, I like using the `TryCreateTableServiceClient` for both.  This is not a blocker BTW.

kshyju (approved):

src/WebJobs.Script/StorageProvider/IDelegatingHandlerProvider.cs
```diff
@@ -3,7 +3,7 @@
 
 using System.Net.Http;
 
-namespace Microsoft.Azure.WebJobs.Script.WebHost.Storage
+namespace Microsoft.Azure.WebJobs.Script
```
fabiocav: Can we avoid the type change in this PR?

src/WebJobs.Script/StorageProvider/TableServiceClientProvider.cs
```diff
@@ -0,0 +1,86 @@
...
+        /// <param name="componentFactory">The Azure factory responsible for creating clients. <see cref="AzureComponentFactory"/>.</param>
+        /// <param name="logForwarder">Log forwarder that forwards events to ILogger. <see cref="AzureEventSourceLogForwarder"/>.</param>
+        public TableServiceClientProvider(AzureComponentFactory componentFactory, AzureEventSourceLogForwarder logForwarder) : base(componentFactory, logForwarder)
+        {
+        }
+
+        protected override TableServiceClient CreateClient(IConfiguration configuration, TokenCredential tokenCredential, TableClientOptions options)
+        {
+            // If connection string is present, it will be honored first; bypass creating a serviceUri
+            if (!IsConnectionStringPresent(configuration))
```
fabiocav, line 31: Can we make this comment clearer? (e.g., `Resolve the service URI if the connection string is not defined.`)

src/WebJobs.Script.WebHost/Scale/TableStorageScaleMetricsRepository.cs
```diff
@@ -225,50 +212,19 @@ internal async Task ExecuteBatchSafeAsync(TableBatchOperation batch, DateTime? n
...
-                    }
-                    throw;
-                }
-
-                return;
+            if (tableCreated && _scaleOptions.MetricsPurgeEnabled)
+            {
+                // when we roll over to a new table, it's a good time to
+                // do a background purge of any old tables
+                QueueBackgroundMetricsTablePurge();
```
fabiocav, line 223: Based on the implementation of `QueueBackgroundMetricsTablePurge`, I don't really see how the previous exception handling logic was ever engaged.

src/WebJobs.Script/StorageProvider/IDelegatingHandlerProvider.cs
cjaliaga: Are you referring just to the namespace change (set it back to `Microsoft.Azure.WebJobs.Script.WebHost.Storage`) or also the project movement: from `WebJobs.Script.WebHost` to `WebJobs.Script`?

The interface was moved to `WebJobs.Script` as `HostAzureTableStorageProvider` dependency.

src/WebJobs.Script.WebHost/Scale/TableStorageScaleMetricsRepository.cs
cjaliaga: This logic hasn't change with the migration, It'll log whatever error and stop processing. On the worst case-scenario, it might also happen that if the issue is not solved, it never purges old tables, filling the storage account. Shall we flag this for a future refactor or did you have something else in mind for this PR?

src/WebJobs.Script/StorageProvider/IDelegatingHandlerProvider.cs
fabiocav: I missed the move. This type was public before, but the only reasonable consumer would have been Core Tools (wouldn't expect usage there, but worth checking to avoid wasted cycles). Would be great if we could avoid this move/change, but this is something we can live with as this is a type I'm hoping we'll be able to remove down the road.

src/WebJobs.Script.WebHost/Scale/TableStorageScaleMetricsRepository.cs
fabiocav: No action required here.

src/WebJobs.Script/StorageProvider/HostAzureTableStorageProvider.cs
```diff
@@ -0,0 +1,62 @@
...
+namespace Microsoft.Azure.WebJobs.Script
+{
+    internal sealed class HostAzureTableStorageProvider : IAzureTableStorageProvider
+    {
+        private readonly ILogger<HostAzureTableStorageProvider> _logger;
+        private readonly IDelegatingHandlerProvider _delegatingHandlerProvider;
+        private readonly TableServiceClientProvider _tableServiceClientProvider;
+        private readonly IConfiguration _configuration;
+
+        public HostAzureTableStorageProvider(IConfiguration configuration, ILogger<HostAzureTableStorageProvider> logger, AzureComponentFactory componentFactory, AzureEventSourceLogForwarder logForwarder, IDelegatingHandlerProvider delegatingHandlerProvider)
```
mathewc: What is the FeatureFlagDisableMergedWebHostScriptHostConfiguration config merging logic that HostAzureBlobStorageProvider is doing (@fabiocav may know?) Wondering why it's only needed by one of these providers

src/WebJobs.Script/StorageProvider/HostAzureTableStorageProvider.cs
```diff
@@ -0,0 +1,62 @@
...
+        public bool TryCreateTableServiceClient(string connection, out TableServiceClient client)
+        {
+            if (string.IsNullOrEmpty(connection))
+            {
+                throw new ArgumentException($"'{nameof(connection)}' cannot be null or empty.", nameof(connection));
+            }
+
+            try
+            {
+                DelegatingHandler handler = _delegatingHandlerProvider?.Create();
```
mathewc: Can/should _delegatingHandlerProvider ever be null?l Looks like it's a DI injected ctor param

src/WebJobs.Script/StorageProvider/TableServiceClientProvider.cs
```diff
@@ -0,0 +1,86 @@
...
+using System;
+using System.Globalization;
+using Azure.Core;
+using Azure.Data.Tables;
+using Microsoft.Extensions.Azure;
+using Microsoft.Extensions.Configuration;
+
+namespace Microsoft.Azure.WebJobs.Script
+{
+    internal sealed class TableServiceClientProvider : StorageClientProvider<TableServiceClient, TableClientOptions>
```
mathewc, line 17: Consider adding a comment block here similar to BlobServiceClientProvider

src/WebJobs.Script/StorageProvider/HostAzureTableStorageProvider.cs
cjaliaga: My understanding is that allows to get a merged configuration sources of the WebHost  `IConfiguration` and the `IConfiguration` at `ScriptHost/JobHost` level, and it reacts when we instantiate a new ScriptHost and loads its configuration entries.

For the `TableStorage`, since it doesn't depend on `JobHostInternalStorageOptions` and for AzureWebJobsStorage, it should come as environment variable, available in the WebHost `IConfiguration`, I made the (questionable?) decision of not merging configurations for this provider and depend on the WebHost one.

I'll let @fabiocav confirm that.

src/WebJobs.Script/StorageProvider/StorageClientProvider.cs
```diff
@@ -37,7 +37,7 @@ public virtual TClient Create(string name, INameResolver resolver, IConfiguratio
             return this.Create(resolvedName, configuration);
         }
 
-        public virtual TClient Create(string name, IConfiguration configuration)
+        public virtual TClient Create(string name, IConfiguration configuration, TClientOptions options = null)
```
mathewc, line 41: Reviewing this code I realize that there's a Sanitization improvement we should make below as well, similar to what we did in https://github.com/Azure/azure-webjobs-sdk/pull/3082 in the StorageClientProvider in the WebJobs SDK.

src/WebJobs.Script/StorageProvider/TableServiceClientProvider.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        /// <summary>
+        /// An options class that constructs a storage service URI from a different properties.
+        /// These properties are specific to WebJobs, as there may be other relevant properties used downstream
+        /// to create storage clients.
+        /// <seealso cref="Microsoft.Extensions.Azure.ClientFactory" />
+        /// A storage service URI can be built using just the account name along with default
+        /// parameters for Scheme and Endpoint Suffix.
+        /// </summary>
+        internal class StorageServiceUriOptions
```
mathewc: I think we should promote the existing Type to a shared type that both Blob and Table Service Client Providers use, adding TableServiceUri/GetTableServiceUri members to the existing Type

test/WebJobs.Script.Tests.Integration/Storage/AzureTableStorageProviderTests.cs
```diff
@@ -0,0 +1,61 @@
...
+using Xunit;
+
+namespace Microsoft.Azure.WebJobs.Script.Tests.Integration.Storage
+{
+    public class AzureTableStorageProviderTests
+    {
+        private const string StorageConnection = "AzureWebJobsStorage";
+
+        [Fact]
+        public async Task TestAzureTableStorageProvider_ConnectionInWebHostConfiguration()
```
mathewc: nit: the usual pattern is for the test method name to start with the method being tested, e.g. here it'd be TryCreateHostingTableServiceClient_ConnectionInWebHostConfiguration.

Same comment for the other test methods below

test/WebJobs.Script.Tests.Integration/Storage/TableServiceProviderTests.cs
```diff
@@ -0,0 +1,92 @@
...
+                    services.AddAzureStorageCoreServices();
+                }).Build();
+
+            var componentFactory = tempHost.Services.GetRequiredService<AzureComponentFactory>();
+            var logForwarder = tempHost.Services.GetRequiredService<AzureEventSourceLogForwarder>();
+            _tableServiceClientProvider = new TableServiceClientProvider(componentFactory, logForwarder);
+        }
+
+        [Fact]
+        public async Task TestTableStorageProvider_ConnectionName()
```
mathewc: Same naming comment on test methods

src/WebJobs.Script/StorageProvider/HostAzureTableStorageProvider.cs
cjaliaga: With our current usage, it cannot be necessary since, as you mentioned, it's a DI injected constructor parameter. I didn't enforce the null check on the constructor because it's not essential for the provider to have a delegating handler provider—opting out could imply that one isn't needed. On second thought based on how we're using it, I'll enforce the null check on the ctor.

src/WebJobs.Script/StorageProvider/IDelegatingHandlerProvider.cs
cjaliaga: Checked Core Tools and there's no explicit reference to the interface nor the implementation type (which I kept in the original namespace/project). It's registered on DI [here](
https://github.com/Azure/azure-functions-core-tools/blob/ae9eff234b96a6120d9106a739252aa200476c31/src/Azure.Functions.Cli/Actions/HostActions/Startup.cs#L87).

All this said, I've been doing some digging and I think I have good news. When `Webjobs.Host.Storage` [was migrated to Track 2 SDKs](https://github.com/Azure/azure-webjobs-sdk/pull/2764), we stopped using the `DelegatingHandlerProviders` with these clients to limit max number of connections in places where we used them [with the previous SDK](https://github.com/Azure/azure-webjobs-sdk/blob/8c90eb9bf2f6562cdd82463371fa38f427206315/src/Microsoft.Azure.WebJobs.Host.Storage/StorageServiceCollectionExtensions.cs#L83). This would make sense because the new SDKs invested on connection handling and pooling. 

On the Host, the only usage of these `DelegatingHandlers` is with the `CloudTableClient` used in _Scale Metrics_ and that belongs to previous SDK.  I would say that the solution to all our concerns is, as part of this migration, to not use them at all. And even remove the types 😊.

src/WebJobs.Script/StorageProvider/StorageClientProvider.cs
fabiocav: Addressed below.

mathewc (approved):

PR #10309, opened 2024-07-19
Supporting WorkloadIdentityCredential in KeyVaultSecretsRespository

mattchenderson:
Existing behavior can't be leveraged in CI without secrets or in alternative hosting environments. This PR expands the last fallback option to include `WorkloadIdentityCredential`.

### Pull request checklist

    * Otherwise: https://github.com/Azure/azure-functions-host/pull/10310

### DIFF
--- a/src/WebJobs.Script.WebHost/Security/KeyManagement/KeyVaultSecretsRepository.cs
+++ b/src/WebJobs.Script.WebHost/Security/KeyManagement/KeyVaultSecretsRepository.cs
@@ -26,6 +26,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
         private const string FunctionKeyPrefix = "host--functionKey--";
         private const string SystemKeyPrefix = "host--systemKey--";
 
+        private readonly Lazy<TokenCredential> _tokenCredential;
         private readonly Lazy<SecretClient> _secretClient;
         private readonly IEnvironment _environment;
 
@@ -38,17 +39,35 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
 
             Uri keyVaultUri = string.IsNullOrEmpty(vaultUri) ? throw new ArgumentException(nameof(vaultUri)) : new Uri(vaultUri);
 
+            _tokenCredential = new Lazy<TokenCredential>(() =>
+            {
+                if (!TryCreateTokenCredential(clientId, clientSecret, tenantId, out TokenCredential credential))
+                {
+                    throw new InvalidOperationException("Failed to create token credential for KeyVaultSecretsRepository");
+                }
+
+                return credential;
+            });
+
             _secretClient = new Lazy<SecretClient>(() =>
             {
-                // If clientSecret and tenantId are provided, use ClientSecret credential; otherwise use managed identity
-                TokenCredential credential = !string.IsNullOrEmpty(clientSecret) && !string.IsNullOrEmpty(tenantId)
-                    ? new ClientSecretCredential(tenantId, clientId, clientSecret)
-                    : new ChainedTokenCredential(new ManagedIdentityCredential(clientId), new ManagedIdentityCredential());
-
-                return new SecretClient(keyVaultUri, credential);
+                return new SecretClient(keyVaultUri, _tokenCredential.Value);
             });
 
             _environment = environment ?? throw new ArgumentNullException(nameof(environment));
+        }
+
+        internal KeyVaultSecretsRepository(string secretsSentinelFilePath, string vaultUri, string clientId, string clientSecret, string tenantId, ILogger logger, IEnvironment environment, TokenCredential testEnvironmentTokenCredential) : this(secretsSentinelFilePath, vaultUri, clientId, clientSecret, tenantId, logger, environment)
+        {
+            _tokenCredential = new Lazy<TokenCredential>(() =>
+            {
+                if (!TryCreateTokenCredential(clientId, clientSecret, tenantId, out TokenCredential credential))
+                {
+                    throw new InvalidOperationException("Failed to create token credential for KeyVaultSecretsRepository");
+                }
+
+                return new ChainedTokenCredential(testEnvironmentTokenCredential, credential);
+            });
         }
 
         // For testing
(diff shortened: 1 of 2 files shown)


src/WebJobs.Script.WebHost/Security/KeyManagement/KeyVaultSecretsRepository.cs
```diff
@@ -72,6 +72,25 @@ public override async Task<ScriptSecrets> ReadAsync(ScriptSecretsType type, stri
...
         }
 
+        private static bool TryCreateTokenCredential(string clientId, string clientSecret, string tenantId, out TokenCredential credential)
+        {
+            if (!string.IsNullOrEmpty(clientSecret) && !string.IsNullOrEmpty(tenantId))
+            {
+                credential = new ClientSecretCredential(tenantId, clientId, clientSecret);
+                return true;
+            }
+            else if (!string.IsNullOrEmpty(clientId))
```
jviau, line 101: workload identity also has a client ID concept, as well as tenant ID. Do we need to support allowing customer to specify those values for workload?

mattchenderson: Here, I think we should just rely on folks/pipelines setting the `AZURE_TENANT_ID`, `AZURE_CLIENT_ID`, and `AZURE_FEDERATED_TOKEN_FILE` environment variables. `WorkloadIdentityCredential` will be use these in the absence of the options params. Documented [here](https://learn.microsoft.com/en-us/dotnet/api/azure.identity.workloadidentitycredential?view=azure-dotnet#constructors).

jviau: I'm not a big fan of relying on only environment variables. What if the customer has multiple identities in their application? How do they differentiate the environment variables between them?

mattchenderson: I very much agree with that in general. The primary motivation here is to enable integration test pipeline changes where we do not necessarily need to expand surface area through our own settings. Those standard env vars should be present. I was hoping to keep this more constrained if possible.

However, there is potentially greater risk if we rely on a contract from a dependency which could at some point change beneath us. That has happened before with this component. I could see an argument that, since the client ID and tenant ID are already captured, we'd really just need to get the federated token file bit handled. If that's a change we'd like to see, I can put together a new iteration. CI will need to be updated to remap to our new value.

(This actually also made me realize there's a missing client ID check in the first if block, so additional iteration required regardless)

jviau: Ah if this is for integration tests only, I am fine with a limited support for now and then we can revisit and expand upon this behavior in the future.

Is there a feature flag or environment variable we can hide workload identity support behind, so it is not unintentionally used outside of integration tests?

mattchenderson: Per chat with @fabiocav, I've shuffled this around so the mainline product surface is not impacted by this. Our test itself can inject the `WorkloadIdentityCredential`.

src/WebJobs.Script.WebHost/Security/KeyManagement/KeyVaultSecretsRepository.cs
```diff
@@ -38,19 +39,37 @@ public KeyVaultSecretsRepository(string secretsSentinelFilePath, string vaultUri
...
+        internal KeyVaultSecretsRepository(string secretsSentinelFilePath, string vaultUri, string clientId, string clientSecret, string tenantId, ILogger logger, IEnvironment environment, TokenCredential testEnvironmentTokenCredential) : this(secretsSentinelFilePath, vaultUri, clientId, clientSecret, tenantId, logger, environment)
+        {
+            _tokenCredential = new Lazy<TokenCredential>(() =>
+            {
+                TokenCredential credential;
+                if (!TryCreateTokenCredential(clientId, clientSecret, tenantId, out credential))
+                {
+                    throw new InvalidOperationException("Failed to create token credential for KeyVaultSecretsRepository");
+                }
+                return new ChainedTokenCredential(testEnvironmentTokenCredential, credential);
```
jviau, line 69: nit: add empty newline after closing `}` whitespace

src/WebJobs.Script.WebHost/Security/KeyManagement/KeyVaultSecretsRepository.cs
```diff
@@ -38,19 +39,37 @@ public KeyVaultSecretsRepository(string secretsSentinelFilePath, string vaultUri
...
 
             _environment = environment ?? throw new ArgumentNullException(nameof(environment));
         }
 
+        internal KeyVaultSecretsRepository(string secretsSentinelFilePath, string vaultUri, string clientId, string clientSecret, string tenantId, ILogger logger, IEnvironment environment, TokenCredential testEnvironmentTokenCredential) : this(secretsSentinelFilePath, vaultUri, clientId, clientSecret, tenantId, logger, environment)
+        {
+            _tokenCredential = new Lazy<TokenCredential>(() =>
+            {
+                TokenCredential credential;
+                if (!TryCreateTokenCredential(clientId, clientSecret, tenantId, out credential))
```
jviau: can inline the credential type:

```suggestion
                if (!TryCreateTokenCredential(clientId, clientSecret, tenantId, out TokenCredential credential))
```

src/WebJobs.Script.WebHost/Security/KeyManagement/KeyVaultSecretsRepository.cs
```diff
@@ -38,19 +39,37 @@ public KeyVaultSecretsRepository(string secretsSentinelFilePath, string vaultUri
...
 
-            _secretClient = new Lazy<SecretClient>(() =>
+            _tokenCredential = new Lazy<TokenCredential>(() =>
             {
-                // If clientSecret and tenantId are provided, use ClientSecret credential; otherwise use managed identity
-                TokenCredential credential = !string.IsNullOrEmpty(clientSecret) && !string.IsNullOrEmpty(tenantId)
-                    ? new ClientSecretCredential(tenantId, clientId, clientSecret)
-                    : new ChainedTokenCredential(new ManagedIdentityCredential(clientId), new ManagedIdentityCredential());
+                TokenCredential credential;
+                if (!TryCreateTokenCredential(clientId, clientSecret, tenantId, out credential))
```
jviau:
```suggestion
                if (!TryCreateTokenCredential(clientId, clientSecret, tenantId, out TokenCredential credential))
```

and remove the separate declaration above.

jviau (approved):

fabiocav (approved):

PR #10345, opened 2024-07-30
preventing placeholder hosts from draining worker console logs

### LINKED ISSUE #9360: "Worker" logs do not show up in app insights
Per [these docs](https://learn.microsoft.com/en-us/azure/azure-functions/configure-monitoring?tabs=v2#configure-categories), the "Worker" category logs are supposed to be showing up in the "traces" table for app insights, but they aren't. They show up in our internal kusto FunctionsLogs table and they are shown when running locally with core tools, but not in a user's app insights. These types of logs are crucial for the Node.js team to communicate with our customers and help them fix their own issues. I tested and these logs are actually missing for all languages, granted some workers log more information than others. Also, the app setting `logging__logLevel__Worker` has no effect in Azure, although it does help filter logs when running locally with core tools.

## Repro steps:

1. Clone this sample Node.js app and deploy: https://github.com/ejizba/workerlogsexample
2. Hit the HttpTrigger1 url
3. Check app insights after several minutes have passed

Here are a couple sample logs, as seen from our kusto FunctionsLogs that we should be seeing in app insights

> "Summary": Error: 'done' has already been called. Please check your script for extraneous calls to 'done'.,
> "Source": Worker.LanguageWorkerChannel.node.c70ab712-70a6-4eca-88f1-d6e30f691a98,
> "Level": 2,

> "Summary": Worker was unable to load entry point "index2.js": abc123,
> "Source": Worker.LanguageWorkerChannel.node.c70ab712-70a6-4eca-88f1-d6e30f691a98,
> "Level": 3,

## More information

This could arguably be a breaking change if these logs just started showing up all the sudden, especially since they can contain errors/warnings which the user may interpret as an urgent problem. I am open to ideas to fix this under a worker capability, feature flag, or whatever so that we can still get it fixed ASAP without worrying about breaking people.

### LINKED ISSUE #10222: Stdout/stderr logs during worker init not sent to app insights
## Repro steps

1. Clone this repo: https://github.com/ejizba/func-stdout-file-load
2. Run `npm install`
3. Deploy to a flex consumption Node.js app
4. Hit the http trigger
5. Run the following query in app insights:

    > traces
    > | where message startswith "stdout log" or message startswith "rpc log"

### Expected behavior

These are the logs I see on windows consumption:

windows consumption

### Actual behavior

These are the logs I see on flex:

flex

### Related information 

This is a big deal because the workers send app startup errors over stdout/stderr (the suggested workaround for https://github.com/Azure/azure-functions-host/issues/9360) and app startup errors are quite common/frustrating with the new programming models. I suspect this issue might be related to the host split work, so I feel it's important we figure this out before moving the host split to non-flex skus.

brettsam:

Resolves #10222 and #9360

During specialization, there can be a race to process logs being generated by the placeholder worker. If a worker writes console output during startup (say, during a file load like the issue describes), the placeholder can dequeue these and log them, which effectively means they go nowhere.

This change prevents placeholder hosts from dequeuing these messages at all, so they are ready to be consumed by the user's host starting up.

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs
+++ b/src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs
@@ -14,6 +14,7 @@ using Microsoft.Azure.WebJobs.Script.Diagnostics;
 using Microsoft.Azure.WebJobs.Script.Eventing;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
 using Mono.Unix;
 
 namespace Microsoft.Azure.WebJobs.Script.Workers
@@ -30,13 +31,15 @@ namespace Microsoft.Azure.WebJobs.Script.Workers
         private readonly IDisposable _eventSubscription;
         private readonly Lazy<ILogger> _toolingConsoleJsonLoggerLazy;
         private readonly IEnvironment _environment;
+        private readonly IOptionsMonitor<ScriptApplicationHostOptions> _scriptApplicationHostOptions;
 
         private bool _useStdErrorStreamForErrorsOnly;
         private Queue<string> _processStdErrDataQueue = new Queue<string>(3);
         private IHostProcessMonitor _processMonitor;
         private object _syncLock = new object();
 
-        internal WorkerProcess(IScriptEventManager eventManager, IProcessRegistry processRegistry, ILogger workerProcessLogger, IWorkerConsoleLogSource consoleLogSource, IMetricsLogger metricsLogger, IServiceProvider serviceProvider, ILoggerFactory loggerFactory, IEnvironment environment, bool useStdErrStreamForErrorsOnly = false)
+        internal WorkerProcess(IScriptEventManager eventManager, IProcessRegistry processRegistry, ILogger workerProcessLogger, IWorkerConsoleLogSource consoleLogSource, IMetricsLogger metricsLogger,
+            IServiceProvider serviceProvider, ILoggerFactory loggerFactory, IEnvironment environment, IOptionsMonitor<ScriptApplicationHostOptions> scriptApplicationHostOptions, bool useStdErrStreamForErrorsOnly = false)
         {
             _processRegistry = processRegistry;
             _workerProcessLogger = workerProcessLogger;
@@ -46,6 +49,7 @@ namespace Microsoft.Azure.WebJobs.Script.Workers
             _useStdErrorStreamForErrorsOnly = useStdErrStreamForErrorsOnly;
             _serviceProvider = serviceProvider;
             _environment = environment;
+            _scriptApplicationHostOptions = scriptApplicationHostOptions;
             _toolingConsoleJsonLoggerLazy = new Lazy<ILogger>(() => loggerFactory.CreateLogger(WorkerConstants.ToolingConsoleLogCategoryName), isThreadSafe: true);
 
             // We subscribe to host start events so we can handle the restart that occurs
@@ -218,7 +222,13 @@ namespace Microsoft.Azure.WebJobs.Script.Workers
             }
             else
             {
-                _consoleLogSource?.Log(consoleLog);
+                // These are 'user' console logs and should be forwarded to the ScriptHost, unless
+                // we're in placeholder mode. In that case, we ignore the logs so they are not captured
+                // by the specialized host when it starts.
+                if (!_scriptApplicationHostOptions.CurrentValue.IsStandbyConfiguration)
+                {
+                    _consoleLogSource?.Log(consoleLog);
+                }
             }
         }
 
--- a/src/WebJobs.Script/Workers/WorkerConsoleLogService.cs
+++ b/src/WebJobs.Script/Workers/WorkerConsoleLogService.cs
@@ -7,18 +7,21 @@ using System.Threading.Tasks;
 using System.Threading.Tasks.Dataflow;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
 
 namespace Microsoft.Azure.WebJobs.Script.Workers
 {
     internal class WorkerConsoleLogService : IHostedService, IDisposable
     {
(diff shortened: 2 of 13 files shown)


test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SpecializationE2ETests.cs
```diff
@@ -1034,6 +1039,46 @@ public async Task SpecializedSite_StopsHostBeforeWorker()
             Assert.Empty(completedLogs.Where(p => p.Level == LogLevel.Error));
         }
 
+        [Fact]
+        public async Task Specialization_Writes_WorkerStartupLogs()
```
brettsam, line 1044: This test failed before my change...

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SpecializationE2ETests.cs
```diff
@@ -1034,6 +1039,46 @@ public async Task SpecializedSite_StopsHostBeforeWorker()
...
+            response.EnsureSuccessStatusCode();
+
+            _environment.SetEnvironmentVariable("AzureWebJobsStorage", storageValue);
+            _environment.SetEnvironmentVariable(EnvironmentSettingNames.AzureWebsiteContainerReady, "1");
+            _environment.SetEnvironmentVariable(EnvironmentSettingNames.AzureWebsitePlaceholderMode, "0");
+
+            response = await client.GetAsync("api/HttpRequestDataFunction");
+            response.EnsureSuccessStatusCode();
+
+            var placeholderLogger = perScriptHostLoggers.Where(p => p.IsPlaceholderMode).Single().Logger;
```
kshyju: nit: Can simplify to just the `Single` call (for 1072 and 1073)

```csharp
perScriptHostLoggers.Single(p => p.IsPlaceholderMode).Logger
```

src/WebJobs.Script/Workers/WorkerConsoleLogService.cs
```diff
@@ -37,7 +41,12 @@ internal WorkerConsoleLogService(ILogger logger, IWorkerConsoleLogSource console
 
         public Task StartAsync(CancellationToken cancellationToken)
         {
-            _processingTask = ProcessLogs();
+            // Don't start processing logs if this is a standby configuration. This can lead
+            // to placeholder loggers capturing user logs during specialization.
+            if (!_options.IsStandbyConfiguration)
```
kshyju, line 46: Does this mean that, any logs emitted from language worker process (such as FunctionsNetHost) during placeholder mode will not be logged until specialization happens?

brettsam: to be clear it's *user* logs, of which there really shouldn't be any during placeholder mode -- @kshyju and i discussed this offline. He had a scenario where he used this for debugging locally but we came up with a better solution.

surgupta-msft: Does this mean that anything that the placeholder used to log before, will either not be logged now or will be processed along with user logs? Any chance of placeholder logs getting mixed with user logs in kusto or reaching to appInsights?

kshyju (approved):

kshyju:
Do we need to add an entry to release notes for this change?

src/WebJobs.Script/Workers/WorkerConsoleLogService.cs
brettsam: It's possible. And a good point. I'm going to lock this down a bit more so that we don't even emit logs to this path when in placeholder mode (they don't really go anywhere, anyway as there's no App Insights hooked up).

src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs
```diff
@@ -218,7 +222,13 @@ internal void BuildAndLogConsoleLog(string msg, LogLevel level)
             }
             else
             {
-                _consoleLogSource?.Log(consoleLog);
+                // These are 'user' console logs and should be forwarded to the ScriptHost, unless
```
liliankasem, line 225: Not blocking: What makes it obvious that these logs are this stage are just "user" logs? Are we just basing this on the "prefix" that the helpers in WorkerProcessUtilities are checking for? (In this case "LanguageWorkerConsoleLogPrefix" and "ToolingConsoleLogPrefix") ?

src/WebJobs.Script/Workers/WorkerConsoleLogService.cs
liliankasem: > he had a scenario where he used this for debugging locally but we came up with a better solution.

Can you share more info about this and what the solution was? IIRC I also use logs for local debugging and would love to know if this is similar and would have an impact here

surgupta-msft:
As mentioned in this issue - https://github.com/Azure/azure-functions-host/issues/9360, can this be a breaking change as users will start receiving logs which they were not getting before? Some logs will be errors/warnings and can have sensitive information as part of exception details (which can be a separate issue to fix but calling it out). Do we need to inform css about this as there could be more CRIs when users see additional errors/warnings in their appInsights?

surgupta-msft (approved):

brettsam:
@surgupta-msft -- good question. I wouldn't consider this a breaking change because the behavior isn't really changing today -- the problem is that during specialization, some logs disappeared.
 
But w/o specialization, it all worked (which is why it was tough to repro initially) -- so some customers would sometimes see these logs and other times not, depending on the specialization.
 
So i'd consider this a "fix" to make it consistently appear, even during specialization

src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs
brettsam: Yes. Anything from us should be using those special prefixes when emitting console logs, which we'll see and forward to kusto. If it doesn't have that prefix, we cannot trust it and we consider it a 'user' log and forward back to the current ScriptHost to handle with it's logger (via the `_consoleLogSource`)

PR #10363, opened 2024-08-02
Fixing thread safety issue in the GrpcWorkerChannel.LoadResponse method

### LINKED ISSUE #10352: GrpcWorkerChannel.LoadResponse is not threadsafe
There is a race in GrpcWorkerChannel that results in failed function executions on this codepath:

System.ArgumentOutOfRangeException : capacity was less than the current size. (Parameter 'value')
    at System.ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument argument,ExceptionResource resource)
    at System.Collections.Generic.List`1.set_Capacity(Int32 value)
    at System.Collections.Generic.List`1.AddWithResize(T item)
    at Microsoft.Azure.WebJobs.Script.Grpc.GrpcWorkerChannel.LoadResponse(FunctionLoadResponse loadResponse) at /src/azure-functions-host/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs : 835
    at Microsoft.Azure.WebJobs.Script.Grpc.GrpcWorkerChannel.PendingItem.SetResult(InboundGrpcEvent message) at /src/azure-functions-host/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs : 1777

The problem is here:
https://github.com/Azure/azure-functions-host/blob/94bd90c9ef10c6be61ff9232c189f5721a512152/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs#L835

There can be concurrent calls to LoadResponse and they race to call List.Add(..). I suggest using a threadsafe collection for _inputLinks instead of List.

FunctionsLogs
| where PreciseTimeStamp >= datetime(2024-07-30T22:30:00.0000000Z) and PreciseTimeStamp <= datetime(2024-07-30T23:00:00.0000000Z)
| where RoleInstance == "5986acec-de97-44ab-852a-2f23dd8cb503"

#### Repro steps

Hard to repro due to the race - in general it seems to require very large scale before you get unlucky with the timing and hit this.

kshyju:
- Fixed thread safety issue in the `GrpcWorkerChannel.LoadResponse` method. 
- Added a test to handle multiple individual LoadResponses, which was previously missing.

resolves #10352

### Pull request checklist

    * Otherwise: Link to backporting PR - #10400
    * [x] Otherwise: Backport tracked by issue/PR #10400

### DIFF
--- a/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
@@ -71,7 +71,7 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
         private ConcurrentDictionary<string, ExecutingInvocation> _executingInvocations = new();
         private IDictionary<string, BufferBlock<ScriptInvocationContext>> _functionInputBuffers = new ConcurrentDictionary<string, BufferBlock<ScriptInvocationContext>>();
         private ConcurrentDictionary<string, TaskCompletionSource<bool>> _workerStatusRequests = new ConcurrentDictionary<string, TaskCompletionSource<bool>>();
-        private List<IDisposable> _inputLinks = new List<IDisposable>();
+        private ConcurrentBag<IDisposable> _inputLinks = new ConcurrentBag<IDisposable>();
         private List<IDisposable> _eventSubscriptions = new List<IDisposable>();
         private IDisposable _startLatencyMetric;
         private IEnumerable<FunctionMetadata> _functions;
--- a/test/WebJobs.Script.Tests/Workers/Rpc/GrpcWorkerChannelTests.cs
+++ b/test/WebJobs.Script.Tests/Workers/Rpc/GrpcWorkerChannelTests.cs
@@ -966,8 +966,11 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.Workers.Rpc
                 Task.Run(() => _testFunctionRpcService.PublishFunctionLoadResponseEvent(function.Id)));
 
             await Task.WhenAll(publishFunctionLoadResponseTasks);
-            // Give some time for the logs to be written.
-            await Task.Delay(TimeSpan.FromMilliseconds(500));
+
+            await TestHelpers.Await(() =>
+            {
+                return _logger.GetLogMessages().Count(m => m.FormattedMessage.StartsWith("Received FunctionLoadResponse")) == allFunctionIdsAndNames.Count;
+            });
 
             var traces = _logger.GetLogMessages();
 
--- a/test/WebJobs.Script.Tests/Workers/Rpc/GrpcWorkerChannelTests.cs
+++ b/test/WebJobs.Script.Tests/Workers/Rpc/GrpcWorkerChannelTests.cs
@@ -940,6 +940,45 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.Workers.Rpc
         }
 
         [Fact]
+        public async Task Receives_Individual_FunctionLoadResponses_Parallel()
+        {
+            await CreateDefaultWorkerChannel();
+
+            var startStreamMessage = new StreamingMessage()
+            {
+                StartStream = new StartStream()
+                {
+                    WorkerId = _workerId
+                }
+            };
+
+            var rpcEvent = new GrpcEvent(_workerId, startStreamMessage);
+            _workerChannel.SendWorkerInitRequest(rpcEvent);
+
+            var functionMetadataList = GetTestFunctionsList("node", numberOfFunctions: 250);
+            _workerChannel.SetupFunctionInvocationBuffers(functionMetadataList);
+            _workerChannel.SendFunctionLoadRequests(managedDependencyOptions: null, TimeSpan.FromSeconds(1));
+
+            var allFunctionIdsAndNames = functionMetadataList.Select(f => new { Id = f.Properties["FunctionId"].ToString(), f.Name }).ToList();
+
+            // Send function load responses for each function, not necessarily in the order the load requests were sent.
+            var publishFunctionLoadResponseTasks = allFunctionIdsAndNames.Select(function =>
+                Task.Run(() => _testFunctionRpcService.PublishFunctionLoadResponseEvent(function.Id)));
+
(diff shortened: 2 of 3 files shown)


src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -71,7 +71,7 @@ internal partial class GrpcWorkerChannel : IRpcWorkerChannel, IDisposable
         private ConcurrentDictionary<string, ExecutingInvocation> _executingInvocations = new();
         private IDictionary<string, BufferBlock<ScriptInvocationContext>> _functionInputBuffers = new ConcurrentDictionary<string, BufferBlock<ScriptInvocationContext>>();
         private ConcurrentDictionary<string, TaskCompletionSource<bool>> _workerStatusRequests = new ConcurrentDictionary<string, TaskCompletionSource<bool>>();
-        private List<IDisposable> _inputLinks = new List<IDisposable>();
+        private ConcurrentBag<IDisposable> _inputLinks = new ConcurrentBag<IDisposable>();
```
kshyju, line 74: We do not rely on the order of items in the collection. (_List is ordered, but ConcurrentBag has no order_)

paulbatum (approved):

jviau (commented):
Do you know if disposal will ever run concurrently with `LoadResponse`? If so, we would have a race condition with not disposing a recently added link.

paulbatum:
> Do you know if disposal will ever run concurrently with `LoadResponse`? If so, we would have a race condition with not disposing a recently added link.

Just to confirm, this would be an existing issue if true right? i.e. this PR doesn't introduce this problem?

test/WebJobs.Script.Tests/Workers/Rpc/GrpcWorkerChannelTests.cs
```diff
@@ -939,6 +939,45 @@ public async Task ReceivesInboundEvent_FunctionLoadResponse()
...
+
+            var allFunctionIdsAndNames = functionMetadataList.Select(f => new { Id = f.Properties["FunctionId"].ToString(), f.Name }).ToList();
+
+            // Send function load responses for each function, not necessarily in the order the load requests were sent.
+            var publishFunctionLoadResponseTasks = allFunctionIdsAndNames.Select(function =>
+                Task.Run(() => _testFunctionRpcService.PublishFunctionLoadResponseEvent(function.Id)));
+
+            await Task.WhenAll(publishFunctionLoadResponseTasks);
+            // Give some time for the logs to be written.
+            await Task.Delay(TimeSpan.FromMilliseconds(500));
```
fabiocav: Can we rely on a better signal instead of a delay? It's fine to have a reasonable timeout, but tests based on timing like this are a bit unrealiable.

test/WebJobs.Script.Tests/Workers/Rpc/GrpcWorkerChannelTests.cs
```diff
@@ -939,6 +939,45 @@ public async Task ReceivesInboundEvent_FunctionLoadResponse()
             Assert.True(traces.Any(m => string.Equals(m.FormattedMessage, "Received FunctionLoadResponse for function: 'js1' with functionId: 'TestFunctionId1'.")), "FunctionLoadResponse TestFunctionId1");
         }
 
+        [Fact]
+        public async Task Receives_Individual_FunctionLoadResponses_Parallel()
+        {
```
fabiocav, line 944: Did you confirm this was failing before the change?

kshyju: I even tried 1000 functions, and multiple local test run could not repro this issue because it is such a rare race condition to repro).

test/WebJobs.Script.Tests/Workers/Rpc/GrpcWorkerChannelTests.cs
kshyju: All the test in this file follows this pattern. I agree we could do better. Not sure that should go in the scope of this fix.

kshyju:
> > Do you know if disposal will ever run concurrently with `LoadResponse`? If so, we would have a race condition with not disposing a recently added link.
> 
> Just to confirm, this would be an existing issue if true right? i.e. this PR doesn't introduce this problem?

Correct. That is an existing issue.

satvu (approved):

test/WebJobs.Script.Tests/Workers/Rpc/GrpcWorkerChannelTests.cs
fabiocav: It's a problematic pattern that is the source of a lot of test reliability issues. My hope is that we can avoid this expanding the use of this pattern. I'm confident we can come up with a more reliable flow to validate the behavior.

brettsam: Using `TestHelpers.Await(callback)`... or whatever it is called... to check for the condition is better here. It will be faster most of the time, and also won't fail if a test occasionally takes 501ms.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
brettsam: I bet this is related to some handling I added a while back in https://github.com/Azure/azure-functions-host/pull/9339 (I wanted more telemetry). I have not investigated, but we'd occasionally see nullrefs during disposal when looping through `_inputLinks`. I bet there has always been a race there.

brettsam (approved):

test/WebJobs.Script.Tests/Workers/Rpc/GrpcWorkerChannelTests.cs
kshyju: ~~I am going to try to switch to that method, not just for this test, but other tests as well, but as a separate PR.~~ Will fix this test to use that.

kshyju: Replaced `Task.Delay(500)` with `TestHelpers.Await` for the new test.

PR #10367, opened 2024-08-05
Worker termination path update

surgupta-msft:

resolves #[issue_for_this_pr](https://github.com/orgs/Azure/projects/473/views/9?sliceBy%5Bvalue%5D=surgupta-msft&pane=issue&itemId=72810177)

There will be a follow-up item here to simplify the class `Sanitizer.cs` for performance - https://github.com/Azure/azure-functions-host/issues/10383

### Pull request checklist

    * Otherwise: Link to backporting PR --> https://github.com/Azure/azure-functions-host/pull/10397

### Additional information

Additional PR information


src/WebJobs.Script/Sanitizer.cs
```diff
@@ -20,6 +21,9 @@ internal static class Sanitizer
         internal static readonly string[] CredentialTokens = new string[] { "Token=", "DefaultEndpointsProtocol=http", "AccountKey=", "Data Source=", "Server=", "Password=", "pwd=", "&amp;sig=", "&sig=", "?sig=", "SharedAccessKey=", "&amp;code=", "&code=", "?code=" };
         private static readonly string[] CredentialNameFragments = new[] { "password", "pwd", "key", "secret", "token", "sas" };
 
+        private static readonly string WebSocketPattern = "(?i)((amqp|ssh|(ht|f)tps?)://[^%:\\s\"'/][^:\\s\"'/\\$]+[^:\\s\"'/\\$%]:([^%\\s\"'/][^@\\s\"'/]{0,100}[^%\\s\"'/])@[\\$a-z0-9:\\._%\\?=/]+|[a-z0-9]{3,5}://[^%:\\s\"'/][^:\\s\"'/\\$]+[^:\\s\"'/\\$%]:([^%\\s\"'/][^@\\s\"'/]{0,100}[^%\\s\"'/])@[\\$a-z0-9:\\._%\\?=/\\-]+)";
```
surgupta-msft: Got this pattern from the incident

src/WebJobs.Script/SanitizerRegex.cs
```diff
@@ -0,0 +1,32 @@
...
+                                                ([^@/\s]+)               # Capture password
+                                                @                        # '@'
+                                                ([^:/\s]+)               # Capture address
+                                                :                        # ':'
+                                                ([0-9]+)\b               # Capture port number
+                                            ";
+
+#if NET8_0_OR_GREATER
+        [GeneratedRegex(Pattern, RegexOptions.IgnoreCase | RegexOptions.IgnorePatternWhitespace, "en-US")]
+        internal static partial Regex Regex();
```
surgupta-msft: Using .Net feature `GeneratedRegex` for performance - https://learn.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-source-generators?pivots=dotnet-8-0#source-generation

src/WebJobs.Script/Sanitizer.cs
```diff
@@ -32,6 +33,8 @@ internal static string Sanitize(string input)
                 return string.Empty;
             }
 
+            input = SanitizerRegex.Regex().Replace(input, SecretReplacement);
```
brettsam: Can we move this *below* the check for `MayContainCredentials()`? And maybe add a check whether ` || input.Contains(":")` to that method? That's meant to be a very quick check to see if we need to go any further and the way you have this set up now, we'd do a regex replacement always, no matter what...

surgupta-msft: Done

brettsam (approved):

src/WebJobs.Script/SanitizerRegex.cs
```diff
@@ -0,0 +1,32 @@
...
+                                                :                        # ':'
+                                                ([^@/\s]+)               # Capture password
+                                                @                        # '@'
+                                                ([^:/\s]+)               # Capture address
+                                                :                        # ':'
+                                                ([0-9]+)\b               # Capture port number
+                                            ";
+
+#if NET8_0_OR_GREATER
+        [GeneratedRegex(Pattern, RegexOptions.IgnoreCase | RegexOptions.IgnorePatternWhitespace, "en-US")]
```
kshyju: Is this code buildable ? Don't we need to add `<TargetFramework>net8.0</TargetFramework>` to the csproj for this to work?

kshyju: Should `Pattern` which is `readonly` works here? I thought it should be `const`

kshyju: The code could be buildable, but would it throw during runtime due to the other error I mentioned about using `readonly`

surgupta-msft: Yes this wont build as of now. When we add TFM to net8 and move all other regex in our repo to use this feature then this one will automatically work. Or may be I can remove using this feature for now. I want to keep scope of this PR only to sanitize the log.

benchmarks/WebJobs.Script.Benchmarks/SanitizerBenchmarks.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    public class SanitizerBenchmarks
+    {
+        [Benchmark]
+        public void Sanitize()
+        {
+            Sanitizer.Sanitize("testprotocol://name:password@address:1111");
+        }
+    }
+}
```
kshyju, line 17: Do we need to add this?

test/WebJobs.Script.Tests/Workers/Rpc/RpcWorkerProcessTests.cs
```diff
@@ -163,6 +164,18 @@ public void ErrorMessageQueue_Full_Enqueue_Success()
             Assert.Equal("Error2,Error3,Error4", exceptionMessage);
         }
 
+        [Fact]
+        public void VerifySanitizedErrorMessage_Success()
```
kshyju: Is this test doing same thing as the `SanitizeString` test in SanitizerTests.cs? I am wondering this joined string can be passed as an inline data to that test.

kshyju (approved):
Approving with 2 open questions.

benchmarks/WebJobs.Script.Benchmarks/SanitizerBenchmarks.cs
surgupta-msft: @brettsam recommended adding this. I am not familiar with it but it seems that benchmarks give us the speed and memory footprint (allocations) of the specific method, so we can come in and do tweaks and compare.

test/WebJobs.Script.Tests/Workers/Rpc/RpcWorkerProcessTests.cs
surgupta-msft: Yes its doing the same thing. I will remove this one. The second input data I added to sanitizeString is this joined string.

satvu (approved):

benchmarks/WebJobs.Script.Benchmarks/SanitizerBenchmarks.cs
kshyju: It can provide that info. But that is useful when we are comparing 2 different implementations. So I am not sure what value this will add by itself.

surgupta-msft: By 2 different implementations, do you mean to also call `Sanitizer.Sanitze("<input>")` with some different input? And then we will get comparison between 2 inputs?

I referred [this](https://github.com/Azure/azure-functions-host/blob/dev/benchmarks/WebJobs.Script.Benchmarks/ScriptLoggingBuilderExtensionsBenchmarks.cs) class for implementation here.

PR #10369, opened 2024-08-06
Forwarding the LanguageWorkerOptions from the host to the job host scope and ensuring we use the provided instances.

fabiocav:

Addressing an issue leading to double initialization of worker configuration and profile evaluation.

### Pull request checklist


src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -329,8 +328,17 @@ public static IHostBuilder AddScriptHostCore(this IHostBuilder builder, ScriptAp
 
                 services.AddSingleton<IFileLoggingStatusManager, FileLoggingStatusManager>();
 
-                if (!applicationHostOptions.HasParentScope)
+                if (applicationHostOptions.HasParentScope)
+                {
+                    // Forward th host LanguageWorkerOptions to the Job Host.
```
liliankasem: super nit: can we clarify this comment a little more in terms of where the options are being forwarded from i.e. 'Forward the LanguageWorkerOptions from the parent (WebHost?) to the ScriptHost.'

Also just a refresher for myself: "HasParentScope" is used to check if this instance is running inside the WebHost or not correct? If there is no parent, does this mean that this instance is the parent? Or not necessarily?

And does not having a parent mean that it is the WebHost or is that still a ScriptHost but without a parent? Does that make it rogue if so?

liliankasem:
Is it possible to add tests to validate this change?

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
fabiocav: Flagging the current PR as a draft as @kshyju is helping iterate on the changes.

Will update the comment for clarity, but to answer your question: `HasParentScope` has the behavior you mentioned. If this is `false`, it signals that we're not running withing a parent container, and services that would otherwise be registered at the parent scope level must be registered within the JobHost container, otherwise they'd be missing. This is only ever the case when running outside of the WebHost, and the only scenario we maintain where that is the case are tests.

liliankasem: Thank you!

kshyju:
@brettsam Ping on the review.

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
jviau:
```suggestion
                    // Forward the host LanguageWorkerOptions to the Job Host.
```

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -329,8 +328,17 @@ public static IHostBuilder AddScriptHostCore(this IHostBuilder builder, ScriptAp
...
+                    var languageWorkerOptions = applicationHostOptions.RootServiceProvider.GetService<IOptionsMonitor<LanguageWorkerOptions>>();
+                    services.AddSingleton(languageWorkerOptions);
+                    services.AddSingleton<IOptions<LanguageWorkerOptions>>((s) => new OptionsWrapper<LanguageWorkerOptions>(languageWorkerOptions.CurrentValue));
```
jviau: nit: don't wrap single param in `()`

```suggestion
                    services.AddSingleton<IOptions<LanguageWorkerOptions>>(s => new OptionsWrapper<LanguageWorkerOptions>(languageWorkerOptions.CurrentValue));
```

src/WebJobs.Script/Workers/Rpc/Configuration/LanguageWorkerOptionsSetup.cs
```diff
@@ -42,18 +47,53 @@ public void Configure(LanguageWorkerOptions options)
         {
             string workerRuntime = _environment.GetEnvironmentVariable(RpcWorkerConstants.FunctionWorkerRuntimeSettingName);
 
-            // Parsing worker.config.json should always be done in case of multi language worker
+            // Parsing worker.latestConfiguration.json should always be done in case of multi language worker
```
brettsam: Is `worker.latestConfiguration.json` a new thing?

src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
```diff
@@ -57,6 +58,16 @@ public WebHostRpcWorkerChannelManager(IScriptEventManager eventManager,
             _applicationHostOptions = applicationHostOptions;
             _hostingConfigOptions = hostingConfigOptions;
 
+            languageWorkerOptions.OnChange(async languageWorkerOptions =>
+            {
+                IRpcWorkerChannel rpcWorkerChannel = await GetChannelAsync(_workerRuntime);
+                if (rpcWorkerChannel != null && !UsePlaceholderChannel(rpcWorkerChannel))
+                {
+                    _logger.LogInformation("Language worker options changed, and the placeholder worker channel is invalid for other reasons. Shutting down the channel.");
+                    await ShutdownChannelIfExistsAsync(_workerRuntime, rpcWorkerChannel.Id);
```
brettsam, line 67: What's the scenario here, and why didn't we have this behavior before?

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -329,8 +328,17 @@ public static IHostBuilder AddScriptHostCore(this IHostBuilder builder, ScriptAp
 
                 services.AddSingleton<IFileLoggingStatusManager, FileLoggingStatusManager>();
 
-                if (!applicationHostOptions.HasParentScope)
+                if (applicationHostOptions.HasParentScope)
+                {
+                    // Forward th host LanguageWorkerOptions to the Job Host.
+                    var languageWorkerOptions = applicationHostOptions.RootServiceProvider.GetService<IOptionsMonitor<LanguageWorkerOptions>>();
```
jviau, line 343: I am slightly concerned about how this will work in practice. There are subtleties to the way options are registered and calculated. I would feel better if we did the following:

``` CSharp
services.AddSingleton(_ => applicationHostOptions.RootServiceProvider.GetService<IOptionsMonitor<LanguageWorkerOptions>>());
services.AddScoped(_ => applicationHostOptions.RootServiceProvider.GetService<IOptionsSnapshot<LanguageWorkerOptions>>());
services.AddSingleton(_ => applicationHostOptions.RootServiceProvider.GetService<IOptions<LanguageWorkerOptions>>());
```

Could even lift this to an extension method `IServiceCollection ForwardOptionsFrom<TOptions>(this IServiceCollection, IServiceProvider source);`

src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
fabiocav: Will need to review this to comment. That's part of a set of changes pushed by @kshyju .

Changes to the signal used to reload `LanguageWorkerOptions` were required. I'm concerned that the logic here is assuming this will always change as a result of specialization, which may change in the future.

At the very least, the message needs to be a bit more descriptive.

src/WebJobs.Script/Workers/Rpc/Configuration/LanguageWorkerOptionsSetup.cs
fabiocav: This is looking like an accidental change to me. I don't expect a different file name.

fabiocav:
@jviau / @brettsam / @liliankasem I'm reviewing the latest changes, and will take point on addressing any feedback.

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
fabiocav: Can you expand on the concerns? I like the idea of wrapping that into an extension to make this cleaner and reusable (this is applied to other options). But for the functionality, in this case, we are explicitly trying to avoid another cache miss when using `IOptions<T>`, for `LanguageWorkerOptions` specifically, given the cost of running its setup today and the state it tracks (those instances would be identical).

fabiocav: Also, one thing to note is that using the suggested pattern will defer execution of the setup to the time services in the child container request the options, if nothing at the host scope consumes one of those option types.

jviau: > Also, one thing to note is that using the suggested pattern will defer execution of the setup to the time services in the child container request the options, if nothing at the host scope consumes one of those option types.

That was intentional. I wanted to ensure `IOptions` in both parent and child container would be the same values. Otherwise, the two containers could potentially have different values if config changes between the two. Unlikely, but that would be one confusing bug if it ever happens.

> we are explicitly trying to avoid another cache miss when using IOptions<T>

Would this cache miss not already exist in the parent container?

fabiocav: With the pattern currently used in the code, the values would be the same, also refreshed when using `IOptionsMonitor`, one of the differences is that when using `IOptions<T>`, you'd end up with the same instance that is returned by `IOptionsMonitor<T>` (at the initial construction here), but that's intentional.

> Would this cache miss not already exist in the parent container?

Right now, there are two cache misses in the JobHost/ScriptHost scope with every single initialization:

1. Services taking a dependency on `IOptionsMonitor<LanguageWorkerOptions>`
2. Services taking a dependency on `IOptions<LanguageWorkerOptions>`

If no WebHost services are using `IOptions<LanguageWorkerOptions>` (or if that ever changes in the future as a result of changes to WebHost scoped services), we'd end up with a cache miss when initializing the JobHost, which would have an impact on cold start in some cases (specialization flows shouldn't be as impacted as the expectation is that placeholders would have forced this to happen).

The `IOptions<LanguageWorkerOptions>` is the second hit we're trying to avoid.

For the WebHost, in specialization cases, even if the host has services consuming `IOptions<LanguageWorkerOptions>`, those wouldn't impact customer cold start as they would happen in placeholders.

Ultimately, much of what we're dealing with here is the fact the this current setup implementation is doing way too much and that logic should be refactored into a separate, singleton, service that is reused, avoiding the duplicate execution of that expensive logic that is not expected to run more than once anyway.

jviau: The WebHost won't ever be realizing `IOptions<LanguageWorkerOptions>` when we have a customer payload? If that is the case, then I am fine with the approach as is.

src/WebJobs.Script/Workers/Rpc/Configuration/LanguageWorkerOptionsSetup.cs
```diff
@@ -42,18 +47,53 @@ public void Configure(LanguageWorkerOptions options)
...
+        public JobHostLanguageWorkerOptionsSetup(ILoggerFactory loggerFactory)
+        {
+            _loggerFactory = loggerFactory;
+        }
+
+        public void PostConfigure(string name, LanguageWorkerOptions options)
+        {
+            var message = $"Call to configure {nameof(LanguageWorkerOptions)} from the JobHost scope. " +
+                $"If using {nameof(IOptions<LanguageWorkerOptions>)}, please use {nameof(IOptionsMonitor<LanguageWorkerOptions>)} instead.";
+            Debug.Fail(message);
```
jviau, line 97: Can you explain the intention behind this? The comment says to use `IOptionsMonitor<LanguageWorkerOptions>`, but won't this class still run (and fail) for that type?

fabiocav: The message is a bit confusing, but with the forwarding, no setups should run for `IOptionsMonitor<LanguageWorkerOptions>` within the JobHost scope. The same applies to `IOptions<LanguageWorkerOptions`, though, so either option would take the host injected instances and not run any setup logic within that scope.

jviau: Ah, can we clarify that?

1. Add a comment explaining this is only registered for JobHost scope
2. Update debug message to something like "Unexpected configuration of LanguageWorkerOptions from the JobHost scope. LanguageWorkerOptions should be forwarded from the parent scope with no additional configuration."

kshyju:
Rebased on dev.

src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
kshyju: This is the scenario described in the test [Specialization_RestartWorkerWithWorkerArguments](https://github.com/Azure/azure-functions-host/blob/dev/test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SpecializationE2ETests.cs#L418). The test fails without this code snippet and the new changes to reuse the language worker options from parent scope.

A language worker can provide arguments(ex: `languageWorkers:node:arguments`) and in that case, we do not want to use the channel created using placeholder mode. The `UsePlaceholderChannel` has logic to trigger restart of current worker process(from placeholder mode) in that case.

https://github.com/Azure/azure-functions-host/blob/e57d37f1190cf05ff2a3159e45fea6991b55ba88/src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs#L173-L178

src/WebJobs.Script/Workers/Rpc/Configuration/LanguageWorkerOptionsSetup.cs
kshyju: Updated the comment with the explicit  message suggested above.

kshyju (approved):

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
```diff
@@ -207,16 +207,19 @@ public static void AddWebJobsScriptHost(this IServiceCollection services, IConfi
...
             services.ConfigureOptionsWithChangeTokenSource<HttpBodyControlOptions, HttpBodyControlOptionsSetup, SpecializationChangeTokenSource<HttpBodyControlOptions>>();
             services.ConfigureOptions<FlexConsumptionMetricsPublisherOptionsSetup>();
             services.ConfigureOptions<ConsoleLoggingOptionsSetup>();
             services.AddHostingConfigOptions(configuration);
             services.ConfigureOptions<ExtensionRequirementOptionsSetup>();
 
+            // Refresh LanguageWorkerOptions when HostBuiltChangeTokenSource is triggered.
+            services.AddSingleton<HostBuiltChangeTokenSource<LanguageWorkerOptions>>();
+            services.AddSingleton<IOptionsChangeTokenSource<LanguageWorkerOptions>>(s => s.GetRequiredService<HostBuiltChangeTokenSource<LanguageWorkerOptions>>());
+
```
fabiocav: Replace with: 
``` csharp
services.AddSingleton<IOptionsChangeTokenSource<LanguageWorkerOptions>, HostBuiltChangeTokenSource<LanguageWorkerOptions>>();
```

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -90,6 +92,7 @@ public WebJobsScriptHostService(IOptionsMonitor<ScriptApplicationHostOptions> ap
             RegisterApplicationLifetimeEvents();
 
             _metricsLogger = metricsLogger;
+            _hostBuiltChangeTokenSourceForLanguageWorkerOptions = hostBuiltChangeTokenSource ?? throw new ArgumentNullException(nameof(hostBuiltChangeTokenSource));
```
fabiocav: nit: smaller variable name here (e.g., `_languageWorkerOptionsChangeTokenSource`)

src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
fabiocav: Can we capture that in the message, with something like:
```
Language worker options changed, and the placeholder worker channel is invalid for other reasons (e.g., worker configuration changes). Shutting down the channel.
```

fabiocav (commented):
@kshyju I can't approve since I originally opened the PR, but I'm good with the changes. Left some minor comments.

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
brettsam: Does our helper `ConfigureOptionsWithChangeTokenSource()` handle the wire-up the way you need it? We wrote that a while back b/c we kept getting it wrong :-)

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -338,8 +337,17 @@ public static IHostBuilder AddScriptHostCore(this IHostBuilder builder, ScriptAp
...
                 services.AddSingleton<IFileLoggingStatusManager, FileLoggingStatusManager>();
 
-                if (!applicationHostOptions.HasParentScope)
+                if (applicationHostOptions.HasParentScope)
+                {
+                    // Forward the host LanguageWorkerOptions to the Job Host.
+                    var languageWorkerOptions = applicationHostOptions.RootServiceProvider.GetService<IOptionsMonitor<LanguageWorkerOptions>>();
+                    services.AddSingleton(languageWorkerOptions);
+                    services.AddSingleton<IOptions<LanguageWorkerOptions>>(s => new OptionsWrapper<LanguageWorkerOptions>(languageWorkerOptions.CurrentValue));
+                    services.ConfigureOptions<JobHostLanguageWorkerOptionsSetup>();
```
brettsam, line 346: Why is this here? This Setup should never be used in the JobHost I thought?

src/WebJobs.Script/Workers/Rpc/Configuration/LanguageWorkerOptionsSetup.cs
```diff
@@ -18,18 +20,21 @@ internal class LanguageWorkerOptionsSetup : IConfigureOptions<LanguageWorkerOpti
...
         private readonly IWorkerProfileManager _workerProfileManager;
+        private readonly IScriptHostManager _scriptHostManager;
 
         public LanguageWorkerOptionsSetup(IConfiguration configuration,
                                           ILoggerFactory loggerFactory,
                                           IEnvironment environment,
                                           IMetricsLogger metricsLogger,
-                                          IWorkerProfileManager workerProfileManager)
+                                          IWorkerProfileManager workerProfileManager,
+                                          IScriptHostManager scriptHostManager)
```
brettsam, line 30: can't this run at JobHost-only level too (without a WebHost)? Is there a ScriptHostManager in that scenario?

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
brettsam: Oh... I see below now that this does a debug check.

src/WebJobs.Script/Workers/Rpc/Configuration/LanguageWorkerOptionsSetup.cs
```diff
@@ -52,8 +57,47 @@ public void Configure(LanguageWorkerOptions options)
...
+        {
+            _loggerFactory = loggerFactory;
+        }
+
+        public void PostConfigure(string name, LanguageWorkerOptions options)
+        {
+            var message = "Unexpected configuration of LanguageWorkerOptions from the JobHost scope. LanguageWorkerOptions should be forwarded from the parent scope with no additional configuration.";
+            Debug.Fail(message);
+
+            var logger = _loggerFactory.CreateLogger("Host.LanguageWorkerConfig");
```
brettsam: Can we create a type-based logger here (`CreateLogger<JobHostLanguageWorkerOptionsSetup>()`)? We're trying to get away from these "Host..." ones so let's not try to add more.

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -361,6 +364,8 @@ private async Task UnsynchronizedStartHostAsync(ScriptHostStartupOperation activ
                     deferredLogProvider.ProcessBufferedLogs(selectedProviders);
                 }
 
+                _hostBuiltChangeTokenSourceForLanguageWorkerOptions.TriggerChange();
```
brettsam: This really only works for the `LanguageWorkerOptions`. i.e. If I wanted to have an `IOptionsMonitor<SomeOtherOptions>` that re-built whenever the host was built, this wouldn't trigger. We likely need the `IChangeToken` to be static here and wrapped in the `HostBuiltChangeTokenSource<TOptions>`. We do this today with `StandbyManager.ChangeToken` -- triggering that one static token fires the event for anyone wanting to know when we've specialized.

 It's probably okay for now... but a little refactoring could make it usable elsewhere too.

src/WebJobs.Script/Workers/Rpc/Configuration/LanguageWorkerOptionsSetup.cs
```diff
@@ -52,8 +57,47 @@ public void Configure(LanguageWorkerOptions options)
...
+            {
+                var latestConfiguration = scriptHostManagerServiceProvider.GetService<IConfiguration>();
+                if (latestConfiguration is not null)
+                {
+                    configuration = new ConfigurationBuilder()
+                        .AddConfiguration(_configuration)
+                        .AddConfiguration(latestConfiguration)
+                        .Build();
+                }
+            }
```
brettsam, line 73: Is it possible that this runs once *before* we have the JobHost built, meaning the _scriptHostManager doesn't have an ActiveHost yet, therefore this throws when calling `GetService<T>()`? I just ran into this in some tests, so it's fresh on my mind...

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
kshyju: > Replace with:
> 
> ```cs
> services.AddSingleton<IOptionsChangeTokenSource<LanguageWorkerOptions>, HostBuiltChangeTokenSource<LanguageWorkerOptions>>();
> ```

That would not work. It causes below error. 

>Unable to resolve service for type 'Microsoft.Azure.WebJobs.Script.WebHost.Configuration.HostBuiltChangeTokenSource`1[Microsoft.Azure.WebJobs.Script.Workers.Rpc.LanguageWorkerOptions]' while attempting to activate 'Microsoft.Azure.WebJobs.Script.WebHost.WebJobsScriptHostService'.

kshyju: > Does our helper `ConfigureOptionsWithChangeTokenSource()` handle the wire-up the way you need it? We wrote that a while back b/c we kept getting it wrong :-)

I do not think that method can be used here. First of it will not register the singleton `HostBuiltChangeTokenSource<LanguageWorkerOptions>` instance to DI, will cause the same  _Unable to resolve service for type 'Microsoft.Azure.WebJobs.Script.WebHost.Configuration.HostBuiltChangeTokenSource`1[Microsoft.Azure.WebJobs.Script.Workers.Rpc.LanguageWorkerOptions]' while attempting to activate 'Microsoft.Azure.WebJobs.Script.WebHost.WebJobsScriptHostService_ error.

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
kshyju: Yea, I see your point. Let me follow up offline,

src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
kshyju: Switched to this language.

src/WebJobs.Script/Workers/Rpc/Configuration/LanguageWorkerOptionsSetup.cs
kshyju: Yes, When running on WebHost level, there is a ScriptHostManager instance, but the ActiveHost property is null.

src/WebJobs.Script/Workers/Rpc/Configuration/LanguageWorkerOptionsSetup.cs
kshyju: In my testing,  When running on WebHost level, there is a ScriptHostManager instance. The `ActiveHost` property on that instance is `null`. The `scriptHostManagerServiceProvider.GetService<IConfiguration>()` returns `null` and our null check handles it.

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
fabiocav: >That would not work. It causes below error.

> Unable to resolve service for type 'Microsoft.Azure.WebJobs.Script.WebHost.Configuration.HostBuiltChangeTokenSource`1[Microsoft.Azure.WebJobs.Script.Workers.Rpc.LanguageWorkerOptions]' while attempting to activate 'Microsoft.Azure.WebJobs.Script.WebHost.WebJobsScriptHostService'.

This because 'Microsoft.Azure.WebJobs.Script.WebHost.WebJobsScriptHostService' is taking a dependency on the concrete type (which I didn't expect). We could change this if desired.

kshyju: Yea, My original thought was to pass that explicit change token source (HostBuilt) for clarity. Synced with @fabiocav  yesterday and he suggested we could pass the interface and then do a cast later to call the method from the concrete type. Switched to that approach. This now allows us to use that `ConfigureOptionsWithChangeTokenSource` method which @brettsam suggested.

src/WebJobs.Script/Workers/Rpc/Configuration/LanguageWorkerOptionsSetup.cs
brettsam: ping on this one...

brettsam (approved):

src/WebJobs.Script/Workers/Rpc/Configuration/LanguageWorkerOptionsSetup.cs
brettsam: nevermind, i'm dumb. wrong iteration :)

src/WebJobs.Script/Workers/Rpc/Configuration/LanguageWorkerOptionsSetup.cs
kshyju: accidental typo. Was fixed in the later iteration.

PR #10399, opened 2024-08-13
Disable all non host startup logs for v3.x

surgupta-msft:

resolves #[issue_for_this_pr](https://github.com/orgs/Azure/projects/473/views/9?sliceBy%5Bvalue%5D=surgupta-msft&pane=issue&itemId=72810459)

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
+++ b/src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
@@ -64,7 +64,6 @@ namespace Microsoft.Azure.WebJobs.Script
         public const string FunctionsRequestBodySizeLimit = "FUNCTIONS_REQUEST_BODY_SIZE_LIMIT";
         public const string FunctionsHostIdCheckLevel = "FUNCTIONS_HOSTID_CHECK_LEVEL";
         public const string FunctionsPlatformConfigFilePath = "FUNCTIONS_PLATFORM_CONFIG_FILE_PATH";
-        public const string EnableLogsInHostV3 = "EnableLogsInHostV3";
 
         //Function in Kubernetes
         public const string PodNamespace = "POD_NAMESPACE";
--- a/src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
+++ b/src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
@@ -64,6 +64,7 @@ namespace Microsoft.Azure.WebJobs.Script
         public const string FunctionsRequestBodySizeLimit = "FUNCTIONS_REQUEST_BODY_SIZE_LIMIT";
         public const string FunctionsHostIdCheckLevel = "FUNCTIONS_HOSTID_CHECK_LEVEL";
         public const string FunctionsPlatformConfigFilePath = "FUNCTIONS_PLATFORM_CONFIG_FILE_PATH";
+        public const string EnableLogsInHostV3 = "EnableLogsInHostV3";
 
         //Function in Kubernetes
         public const string PodNamespace = "POD_NAMESPACE";
--- a/src/WebJobs.Script/Extensions/ScriptLoggingBuilderExtensions.cs
+++ b/src/WebJobs.Script/Extensions/ScriptLoggingBuilderExtensions.cs
@@ -1,10 +1,13 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Concurrent;
+using System.Collections.Immutable;
 using System.Linq;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.Azure.WebJobs.Script;
+using Microsoft.Azure.WebJobs.Script.Config;
 using Microsoft.Azure.WebJobs.Script.Configuration;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Hosting;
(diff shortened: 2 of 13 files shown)


src/WebJobs.Script/Extensions/ScriptLoggingBuilderExtensions.cs
```diff
@@ -36,7 +38,11 @@ internal static bool Filter(string category, LogLevel actualLevel, LogLevel minL
 
         private static bool IsFiltered(string category)
         {
-            return _filteredCategoryCache.GetOrAdd(category, c => ScriptConstants.SystemLogCategoryPrefixes.Where(p => category.StartsWith(p)).Any());
+            ImmutableArray<string> systemLogCategoryPrefixes = FeatureFlags.IsEnabled(EnvironmentSettingNames.EnableLogsInHostV3)
+                                                                ? ScriptConstants.SystemLogCategoryPrefixes
+                                                                : ScriptConstants.SystemLogCategoryPrefixesForV3;
+
+            return _filteredCategoryCache.GetOrAdd(category, c => systemLogCategoryPrefixes.Where(p => category.StartsWith(p)).Any());
```
kshyju: Instead of `Where` + `Any`, use just `Any`

```csharp
systemLogCategoryPrefixes.Any(p => category.StartsWith(p))
```

kshyju (commented):
Are we disabling non startup logs for all categories or just Informational and below? For example, an app crashes and tries to log error log entry (Level==2), this change will prevent that from being logged. Is that the desired behavior?

src/WebJobs.Script/Extensions/ScriptLoggingBuilderExtensions.cs
surgupta-msft: Yes, the desired behavior is to disable all levels of non-startup logs. This is also required from security point of view.

src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
```diff
@@ -64,6 +64,7 @@ public static class EnvironmentSettingNames
         public const string FunctionsRequestBodySizeLimit = "FUNCTIONS_REQUEST_BODY_SIZE_LIMIT";
         public const string FunctionsHostIdCheckLevel = "FUNCTIONS_HOSTID_CHECK_LEVEL";
         public const string FunctionsPlatformConfigFilePath = "FUNCTIONS_PLATFORM_CONFIG_FILE_PATH";
+        public const string EnableLogsInHostV3 = "EnableLogsInHostV3";
```
brettsam: Nit -- can you put this with the other FeatureFlag constants in ScriptConstants.cs? Like next to `FeatureFlagEnableProxies`, for example?

src/WebJobs.Script/ScriptConstants.cs
```diff
@@ -214,6 +214,7 @@ public static class ScriptConstants
         public static readonly long DefaultMaxRequestBodySize = 104857600;
 
         public static readonly ImmutableArray<string> SystemLogCategoryPrefixes = ImmutableArray.Create("Microsoft.Azure.WebJobs.", "Function.", "Worker.", "Host.");
+        public static readonly ImmutableArray<string> SystemLogCategoryPrefixesForV3 = ImmutableArray.Create("Host.Startup");
```
brettsam: My only real concern here is that we'll lose backend telemetry that tells us real usage of our older versions. @mattchenderson I know I mentioned this off-hand, but do you know if we capture telemetry based on Metrics or if its based on FunctionStarted/Completed traces?

I'm nervous we'll see some drop and say that we've got almost no v3 usage, when in reality...

src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
surgupta-msft: done

kshyju (approved):

brettsam (approved):

src/WebJobs.Script/ScriptConstants.cs
surgupta-msft: Discussed offline

src/WebJobs.Script/ScriptConstants.cs
```diff
@@ -124,6 +124,7 @@ public static class ScriptConstants
         public const string FeatureFlagDisableWorkerIndexing = "DisableWorkerIndexing";
         public const string FeatureFlagEnableMultiLanguageWorker = "EnableMultiLanguageWorker";
         public const string FeatureFlagEnableLinuxEPExecutionCount = "EnableLinuxFEC";
+        public const string FeatureFlagEnableLogsInHostV3 = "EnableLogsInHostV3";
```
fabiocav: Let's simply call this `EnableHostLogs`

src/WebJobs.Script/ScriptConstants.cs
```diff
@@ -214,6 +215,7 @@ public static class ScriptConstants
         public static readonly long DefaultMaxRequestBodySize = 104857600;
 
         public static readonly ImmutableArray<string> SystemLogCategoryPrefixes = ImmutableArray.Create("Microsoft.Azure.WebJobs.", "Function.", "Worker.", "Host.");
+        public static readonly ImmutableArray<string> SystemLogCategoryPrefixesForV3 = ImmutableArray.Create("Host.Startup");
```
fabiocav: Suggest changing the variable name to something like `RestrictedSystemLogCategoryPrefixes`

src/WebJobs.Script/ScriptConstants.cs
surgupta-msft: Done. Naming looks much better now. Thanks

fabiocav (approved):

PR #10432, opened 2024-08-27
HostJson validation

### LINKED ISSUE #10021: Telemetry for improperly zipped apps
Follow up of [this issue](https://github.com/Azure/azure-functions-dotnet-worker/issues/2335). We've been seeing an increase in "No Job Functions Found" errors as a result of improperly zipped apps that resulted in an unexpected folder structure. 

While these apps have the correct files needed to start up (like host.json), the incorrect folder and file structure can lead to vague errors that prevent the apps from behaving as expected. Some considerations for tracking this:

- Track guaranteed failure cases (double-zipping an app, similar to issue linked above)
  - Runtime languages have different payloads, need to avoid throwing errors for valid apps
- Logging (minimal) file structure to help with servicing and debugging this error
- Design error/failure behavior for improperly zipped apps, troubleshooting guides

RohitRanjanMS:
Implementing a validation check for host.json to handle scenarios such as incorrectly zipped files. This check will log errors to both the console for local development and to App Insights, allowing users to troubleshoot these issues independently.

resolves #10021 

Scenario: Improperly zipped apps- host.json is not at the root but in one of the subfolders.
`Error`
```
No functions were found. A valid host.json file wasn't found in the package root. However, one was located at: {hostJsonFilesPath}. This state indicates that your deployment package was created incorrectly. For deployment package requirements, see https://aka.ms/deployment-zip-push.
```
Scenario: Failed deployment or when a new app is created
`Information`
```
No functions were found. This can occur before you deploy code to your function app or when the host.json file is missing from the most recent deployment. Make sure that your deployment package includes the host.json file in the root of the package. For deployment package requirements, see https://aka.ms/functions-deployment-technologies.
```

### Pull request checklist

    * Otherwise: Link to backporting


src/WebJobs.Script/Host/FunctionMetadataManager.cs
```diff
@@ -285,5 +293,55 @@ private void AddMetadataFromCustomProviders(IEnumerable<IFunctionProvider> funct
...
+                    _logger.ValidateHostJsonZipIssue(hostJsonFilesPath);
+                }
+                else
+                {
+                    _logger.ValidateHostJsonNoHostJson();
+                }
+            }
+            else
+            {
+                _logger.ValidateHostJson();
```
RohitRanjanMS: I'm uncertain whether logging is necessary in this scenario where the host.json file exists, but we still have zero functions.

satvu: I think it can be helpful by explicitly ruling out `host.json` as the reason for 0 functions found, while also logging that this _might_ be the result of a deployment issue (basing this off of the log message defined in the `LoggerExtension` file). 

There was an interesting incident recently where 0 functions were found, and it was due to a breaking change in GitHub Actions which customers were using to deploy (see [here](https://github.com/Azure/azure-functions-dotnet-worker/issues/2683#issuecomment-2329502236)). I can see having a message like this log leading customers to inspect their deployment method more quickly.

However, I don't know for certain if deployment issues make up a large enough majority of `0 functions found` to log this whenever 0 functions are found and `host.json` exists.

src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
```diff
@@ -183,6 +183,21 @@ internal static class LoggerExtension
...
 
+        private static readonly Action<ILogger, string, Exception> _validateHostJsonZipIssue =
+    LoggerMessage.Define<string>(LogLevel.Information,
+        new EventId(338, nameof(ValidateHostJsonZipIssue)),
+        "No functions were found. The host.json file was not located at the root, but it was found at the following path(s): {hostJsonFilesPath}. A default host.json was created at the root using a default configuration to start the function app. See https://aka.ms/deployment-zip-push for more information on deployment.");
+
+        private static readonly Action<ILogger, Exception> _validateHostJsonNoHostJson =
+            LoggerMessage.Define(LogLevel.Information,
+                new EventId(339, nameof(ValidateHostJsonNoHostJson)),
+                "No functions were found. The host.json file was not located at the root. A default host.json was created at the root using a default configuration to start the function app. See https://aka.ms/functions-deployment-technologies for more information on deployment.");
```
satvu: Is this the case where no `host.json` file is found at all? If so, can this be expanded to "host.json file was not found in the deployment payload" or something similar, instead of "not located at the root"?

src/WebJobs.Script/Host/FunctionMetadataManager.cs
```diff
@@ -285,5 +293,55 @@ private void AddMetadataFromCustomProviders(IEnumerable<IFunctionProvider> funct
                 }
             }
         }
+
+        private void ValidateHostJsonIfNoFunctionsFound()
```
surgupta-msft: Naming could be - `VerifyHostJsonPresence` or `ValidateHostJsonExists` or something along this line.

src/WebJobs.Script/Host/FunctionMetadataManager.cs
```diff
@@ -285,5 +293,55 @@ private void AddMetadataFromCustomProviders(IEnumerable<IFunctionProvider> funct
...
+        {
+            string hostFilePath = Path.Combine(_scriptOptions.Value.RootScriptPath, ScriptConstants.HostMetadataFileName);
+
+            // Search for the host.json file within nested directories to verify scenarios where it isn't located at the root. This situation often occurs when a function app has been improperly zipped.
+            IEnumerable<string> hostJsonFiles = null;
+            try
+            {
+                hostJsonFiles = Directory.GetFiles(_scriptOptions.Value.RootScriptPath, ScriptConstants.HostMetadataFileName, SearchOption.AllDirectories)
+                    .Where(file => !file.Equals(hostFilePath, StringComparison.OrdinalIgnoreCase));
```
surgupta-msft: What will happen if there are no functions in the customer payload but there are 2 host.json files (created by Cx) - one at the root and another in a subdirectory? I could be wrong, but I think that the variable `hostJsonFiles` will have the path of subdirectory one and it will incorrectly print the message of `_validateHostJsonZipIssue`. Worth validating this scenario.

src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
```diff
@@ -183,6 +183,21 @@ internal static class LoggerExtension
             new EventId(337, nameof(MinimumBundleVersionNotSatisfied)),
             "Referenced bundle {bundleId} of version {bundleVersion} does not meet the required minimum version of {minimumVersion}. Update your extension bundle reference in host.json to reference {minimumVersion2} or later.");
 
+        private static readonly Action<ILogger, string, Exception> _validateHostJsonZipIssue =
+    LoggerMessage.Define<string>(LogLevel.Information,
+        new EventId(338, nameof(ValidateHostJsonZipIssue)),
+        "No functions were found. The host.json file was not located at the root, but it was found at the following path(s): {hostJsonFilesPath}. A default host.json was created at the root using a default configuration to start the function app. See https://aka.ms/deployment-zip-push for more information on deployment.");
+
+        private static readonly Action<ILogger, Exception> _validateHostJsonNoHostJson =
```
surgupta-msft: Nit: How about the name `_validateNoHostJson`?

src/WebJobs.Script/Host/FunctionMetadataManager.cs
surgupta-msft: Yup I am also not sure if we should add here. We already have another log for count of functions found - [FunctionsReturnedByProvider](https://github.com/Azure/azure-functions-host/blob/6bccacfed716d1ec778c94fc3513cc351f497f14/src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs#L129). We are logging in [FunctionMetadataManager](https://github.com/Azure/azure-functions-host/blob/6bccacfed716d1ec778c94fc3513cc351f497f14/src/WebJobs.Script/Host/FunctionMetadataManager.cs#L250) just before calling method `ValidateHostJsonIfNoFunctionsFound()`. We need to combine these 2 logs to avoid redundancy.

src/WebJobs.Script/Host/FunctionMetadataManager.cs
```diff
@@ -285,5 +293,55 @@ private void AddMetadataFromCustomProviders(IEnumerable<IFunctionProvider> funct
...
+
+            if (configuration == null)
+            {
+                return false;
+            }
+
+            // The host creates a default host.json if it is not found at the root. A newly created function app from the portal would not have a host.json file.
+            // Determine if the host is using a default host configuration.
+            IConfigurationSection jobHostSection = configuration.GetSection(ConfigurationSectionNames.JobHost);
+            string isDefaultHostConfigValue = jobHostSection.GetSection(ConfigurationSectionNames.IsDefaultHostConfig).Value;
```
surgupta-msft: I think we need null check for section `IsDefaultHostConfig`? The host.json created by Cx will not have this, right?

src/WebJobs.Script/Host/FunctionMetadataManager.cs
```diff
@@ -285,5 +293,55 @@ private void AddMetadataFromCustomProviders(IEnumerable<IFunctionProvider> funct
...
+            {
+                hostJsonFiles = Directory.GetFiles(_scriptOptions.Value.RootScriptPath, ScriptConstants.HostMetadataFileName, SearchOption.AllDirectories)
+                    .Where(file => !file.Equals(hostFilePath, StringComparison.OrdinalIgnoreCase));
+            }
+            catch
+            {
+                // Ignore any exceptions while looking for host.json files in nested directories.
+            }
+
+            if (IsDefaultHostConfig())
```
surgupta-msft: My understanding is that the field `defaultHostConfig` is being used to differentiate between the host.json that comes with Customer payload versus the Host.json created by Host if it does not find the file in Cx payload. Is this correct?

If Customer has an app with no functions and their host.json also has this field, we may end up printing incorrect log?

test/WebJobs.Script.Tests/FunctionMetadataManagerTests.cs
```diff
@@ -438,6 +438,43 @@ public void FunctionMetadataManager_ResetProviders_OnRefresh()
             Assert.Equal("newFunction", testFunctionMetadataManager.GetFunctionMetadata(true).FirstOrDefault()?.Name);
         }
 
+        [Fact]
+        public void FunctionMetadataManager_GetsMetadata_ValidateHostJson()
+        {
```
surgupta-msft, line 479: Is it possible to add test cases for other scenarios also - where Host.json is present at root or in a subdirectory or at both the places? If not, we can test these scenarios using private site extension.

src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
```diff
@@ -367,5 +382,20 @@ public static void MinimumBundleVersionNotSatisfied(this ILogger logger, string
         {
             _minimumBundleVersionNotSatisfied(logger, bundleId, bundleVersion, minimumVersion, minimumVersion, null);
         }
+
+        public static void ValidateHostJsonZipIssue(this ILogger logger, string path)
```
surgupta-msft: Does any log written here go to both AppInsights and Console?

RohitRanjanMS: yes, to both, console and AppInsights.

src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
```diff
@@ -183,6 +183,21 @@ internal static class LoggerExtension
...
+    LoggerMessage.Define<string>(LogLevel.Information,
+        new EventId(338, nameof(ValidateHostJsonZipIssue)),
+        "No functions were found. The host.json file was not located at the root, but it was found at the following path(s): {hostJsonFilesPath}. A default host.json was created at the root using a default configuration to start the function app. See https://aka.ms/deployment-zip-push for more information on deployment.");
+
+        private static readonly Action<ILogger, Exception> _validateHostJsonNoHostJson =
+            LoggerMessage.Define(LogLevel.Information,
+                new EventId(339, nameof(ValidateHostJsonNoHostJson)),
+                "No functions were found. The host.json file was not located at the root. A default host.json was created at the root using a default configuration to start the function app. See https://aka.ms/functions-deployment-technologies for more information on deployment.");
+
+        private static readonly Action<ILogger, Exception> _validateHostJson =
```
kshyju: Could we improve the readability of this variable name? From the name, I assumed it logs something like "Validating host json", but the message we are logging here is about a deployment issue.

src/WebJobs.Script/Host/FunctionMetadataManager.cs
```diff
@@ -285,5 +293,55 @@ private void AddMetadataFromCustomProviders(IEnumerable<IFunctionProvider> funct
...
+            {
+                if (hostJsonFiles != null && hostJsonFiles.Any())
```
kshyju: What is the behavior in placeholder scenario?. We will not have an app payload at that point.

src/WebJobs.Script/Host/FunctionMetadataManager.cs
```diff
@@ -249,6 +252,11 @@ private void AddMetadataFromCustomProviders(IEnumerable<IFunctionProvider> funct
 
             _logger.FunctionsReturnedByProvider(totalFunctionsCount, MetadataProviderName);
 
+            if (totalFunctionsCount == 0)
```
kshyju: Also, should we do this inside `LoadFunctionMetadata` instead of `AddMetadataFromCustomProviders` method?

src/WebJobs.Script/Config/HostJsonFileConfigurationSource.cs
```diff
@@ -245,7 +245,8 @@ internal JObject LoadHostConfig(string configFilePath)
 
             private JObject GetDefaultHostConfigObject()
             {
-                var hostJsonJObj = JObject.Parse("{'version': '2.0'}");
+                // isDefaultHostConfig is used to determine if the host.json file was created by the system
```
surgupta-msft, line 249: I feel this comment should be in `ConfigurationSectionNames.cs` where we are defining this variable.

src/WebJobs.Script/Config/HostJsonFileConfigurationSource.cs
```diff
@@ -245,7 +245,8 @@ internal JObject LoadHostConfig(string configFilePath)
...
+                var hostJsonJObj = JObject.Parse("{'version': '2.0', 'isDefaultHostConfig': true}");
```
surgupta-msft, line 250: Are we okay if Customers start seeing this variable - `isDefaultHostConfig` and start using it in their host.json? My concern is on introducing a new Cx facing variable only for this validation. Can this create confusion among customers?

How about we remove this variable from host.json after our validation is done? But I also don't want to perform extra file read operation if it's unnecessary.

src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
```diff
@@ -183,6 +183,21 @@ internal static class LoggerExtension
             new EventId(337, nameof(MinimumBundleVersionNotSatisfied)),
             "Referenced bundle {bundleId} of version {bundleVersion} does not meet the required minimum version of {minimumVersion}. Update your extension bundle reference in host.json to reference {minimumVersion2} or later.");
 
+        private static readonly Action<ILogger, string, Exception> _hostJsonZipDeploymentIssue =
+    LoggerMessage.Define<string>(LogLevel.Warning,
+        new EventId(338, nameof(HostJsonZipDeploymentIssue)),
+        "No functions were found. The host.json file was not located at the root, but it was found at the following path(s): {hostJsonFilesPath}. A default host.json was created at the root using a default configuration to start the function app. See https://aka.ms/deployment-zip-push for more information on deployment.");
```
surgupta-msft: Nit - _"A default host.json file with **standard** configuration was created at the root to **initialize** the function app"_ - to keep it consistent with the log `NoHostJsonFile`

src/WebJobs.Script/Host/FunctionMetadataManager.cs
```diff
@@ -285,5 +294,53 @@ private void AddMetadataFromCustomProviders(IEnumerable<IFunctionProvider> funct
                 }
             }
         }
+
+        private void HostJsonFileValidator()
```
surgupta-msft: I think since this is a method, the name would be `ValidateHostJsonFile()`. We can also keep this code in a separate class `HostJsonFileValidator` for single responsibility.

src/WebJobs.Script/Host/FunctionMetadataManager.cs
```diff
@@ -285,5 +294,53 @@ private void AddMetadataFromCustomProviders(IEnumerable<IFunctionProvider> funct
...
+                        _logger.HostJsonZipDeploymentIssue(hostJsonFilesPath);
+                    }
+                    else
+                    {
+                        _logger.NoHostJsonFile();
+                    }
+                }
+                else
+                {
+                    _logger.MissingFunctionsDeploymentIssue();
```
surgupta-msft: If cx app has host.json file at the root folder (and Cx has added field `isDefaultHostConfig` to true) but the app does not have any functions. We will not be printing "no functions found" message for that case.

To avoid customer using this variable incorrectly, we need to have its correct usage in public documentation.

src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
```diff
@@ -183,6 +183,21 @@ internal static class LoggerExtension
...
+
+        private static readonly Action<ILogger, Exception> _noHostJsonFile =
+    LoggerMessage.Define(LogLevel.Information,
+        new EventId(339, nameof(NoHostJsonFile)),
+        "No functions were found. A default host.json file with standard configuration was created at the root to initialize the function app. If the app was deployed, the host.json file was missing from the deployment payload. For more information on deployment, see https://aka.ms/functions-deployment-technologies.");
+
+        private static readonly Action<ILogger, Exception> _missingFunctionsDeploymentIssue =
+    LoggerMessage.Define(LogLevel.Warning,
+        new EventId(340, nameof(MissingFunctionsDeploymentIssue)),
+        "We located the host.json file in the root directory, but no functions were found. This could be due to a deployment issue. For more details on deployment, please visit https://aka.ms/functions-deployment-technologies.");
```
surgupta-msft: We can also have detectors on top of these logs so the Applens can show reason for "no functions found".

src/WebJobs.Script/Host/FunctionMetadataManager.cs
```diff
@@ -285,5 +294,53 @@ private void AddMetadataFromCustomProviders(IEnumerable<IFunctionProvider> funct
...
+            }
+            catch
+            {
+                // Ignore any exceptions.
+            }
+        }
+
+        private bool IsDefaultHostConfig()
+        {
+            IConfiguration configuration = _serviceProvider?.GetRequiredService<IConfiguration>();
```
soninaren: Can we try to read this value from `ScriptApplicationHostOptions`? Would be good to avoid manually reading from IConfiguration .

soninaren (commented):
Left one comment. Looks good otherwise.

src/WebJobs.Script/Host/FunctionMetadataManager.cs
surgupta-msft: we should add tests to make sure it prints "no functions found" for the valid cases. Meaning when the app does not have any functions whether they have `isDefaultHostConfig` or not

surgupta-msft: @RohitRanjanMS is this the log we discussed about removing?

RohitRanjanMS: @surgupta-msft, yes this is the log. I have updated the PR. Removed log for this scenario.

RohitRanjanMS: We have 2 scenarios now - 
1. host.json doesn't exist at the root (or the customer added a host.json with the `isDefaultHostConfig` flag), and there is a host.json in one of the subfolders of the zip package, resulting in 0 functions found.
```
Error - No functions were found. A valid host.json file wasn't found in the package root. However, one was located at: {hostJsonFilesPath}. This state indicates that your deployment package was created incorrectly. For deployment package requirements, see https://aka.ms/deployment-zip-push.
```
2. host.json doesn't exist at the root (or the customer added a host.json with the `isDefaultHostConfig` flag) and 0 functions found. 
```
Information - No functions were found. This can occur before you deploy code to your function app or when the host.json file is missing from the most recent deployment. Make sure that your deployment package includes the host.json file in the root of the package. For deployment package requirements, see https://aka.ms/functions-deployment-technologies.
```

We discussed the misuse of isDefaultHostConfig and how it might influence this logic, but chose not to document it to avoid confusion. `isDefaultHostConfig` in host.json won't have any impact unless we encounter the 0 functions found issue.

src/WebJobs.Script/Config/ConfigurationSectionNames.cs
```diff
@@ -25,5 +25,10 @@ public static class ConfigurationSectionNames
         public const string SequentialJobHostRestart = JobHost + ":sequentialRestart";
         public const string SendCanceledInvocationsToWorker = "sendCanceledInvocationsToWorker";
         public const string TelemetryMode = "telemetryMode";
+
+        /// <summary>
+        /// Determine if the host.json file was created by the system.
+        /// </summary>
+        public const string IsDefaultHostConfig = "isDefaultHostConfig";
```
liliankasem: Doesn't this property need to be defined in ScriptApplicationHostOptions? That is where the summary comment should be instead of here. E.g. https://github.com/Azure/azure-functions-host/blob/5bac7614d792ff897daff3579623874b614ab8a0/src/WebJobs.Script/Config/ScriptJobHostOptions.cs#L122-L125

src/WebJobs.Script/Host/FunctionMetadataManager.cs
liliankasem: +1

surgupta-msft (approved):

soninaren (approved):

src/WebJobs.Script/Config/ConfigurationSectionNames.cs
RohitRanjanMS: Added it in ScriptJobHostOptions.

src/WebJobs.Script/Host/FunctionMetadataManager.cs
RohitRanjanMS: Done.

PR #10436, opened 2024-08-27
[in-proc] Migrate Scale Metrics to Azure.Data.Tables with support for Identity-based connections

cjaliaga:
in-proc backport of #10276


mattchenderson (approved):
Approved as backport.

satvu (approved):

PR #10438, opened 2024-08-28
Add support for identity-based connections to Diagnostic Events

### LINKED ISSUE #9569: Diagnostic Events API should work with a Managed Identity
Currently, the Diagnostic Events API relies on a connection string to read the diagnostic events from table storage. However, if a user has configured their app to use Managed Identity, they will not be able to get these events to help troubleshoot any issues with their function app.

/cc @soninaren @eamonoreilly  @FinVamp1 @fabiocav @mattchenderson

cjaliaga:

Adds support for identity-based connections to Diagnostic Events via creating the `TableServiceClient` with `IAzureTableStorageProvider`.

resolves #9569 

### Pull request checklist

    * Otherwise: #10442 

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -3,9 +3,12 @@
 <!-- Please add your release notes in the following format:
 - My change description (#PR)
 -->
-- Update Python Worker Version to [4.33.0](https://github.com/Azure/azure-functions-python-worker/releases/tag/4.33.0)
+- Update Python Worker Version to [4.34.0](https://github.com/Azure/azure-functions-python-worker/releases/tag/4.34.0)
 - Sanitize exception logs (#10443)
 - Improving console log handling during specialization (#10345)
 - Update Node.js Worker Version to [3.10.1](https://github.com/Azure/azure-functions-nodejs-worker/releases/tag/v3.10.1)
 - Remove packages `Microsoft.Azure.Cosmos.Table` and `Microsoft.Azure.DocumentDB.Core` (#10503)
+- Implement host configuration property to allow configuration of the metadata provider timeout period (#10526)
+  - The value can be set via `metadataProviderTimeout` in host.json and defaults to "00:00:30" (30 seconds).
+  - For logic apps, unless configured via the host.json, the timeout is disabled by default.
 - Added support for identity-based connections to Diagnostic Events (#10438)
--- a/release_notes.md
+++ b/release_notes.md
@@ -8,4 +8,4 @@
 - Improving console log handling during specialization (#10345)
 - Update Node.js Worker Version to [3.10.1](https://github.com/Azure/azure-functions-nodejs-worker/releases/tag/v3.10.1)
 - Remove packages `Microsoft.Azure.Cosmos.Table` and `Microsoft.Azure.DocumentDB.Core` (#10503)
-- Added support for Identity-based connections to Diagnostic Events (#10438)
+- Added support for identity-based connections to Diagnostic Events (#10438)
--- a/release_notes.md
+++ b/release_notes.md
@@ -8,3 +8,4 @@
 - Improving console log handling during specialization (#10345)
 - Update Node.js Worker Version to [3.10.1](https://github.com/Azure/azure-functions-nodejs-worker/releases/tag/v3.10.1)
 - Remove packages `Microsoft.Azure.Cosmos.Table` and `Microsoft.Azure.DocumentDB.Core` (#10503)
+- Added support for Identity-based connections to Diagnostic Events (#10438)
--- a/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventRepositoryFactory.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventRepositoryFactory.cs
@@ -4,7 +4,6 @@
 using System;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.Logging;
 
 namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 {
@@ -21,12 +20,12 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 
         public IDiagnosticEventRepository Create()
         {
-            // Using this to break circular dependency for ILoggers. Typically you cannot log errors within the logging pipeline because it creates infinte loop.
+            // Using this to break circular dependency for ILoggers. Typically you cannot log errors within the logging pipeline because it creates infinite loop.
             // However in this case that loop is broken because of the filtering in the DiagnosticEventLogger
 
-            IConfigurationSection connectionSection = _configuration.GetWebJobsConnectionSection(ConnectionStringNames.Storage);
+            IConfigurationSection storageConnectionSection = _configuration.GetWebJobsConnectionSection(ConnectionStringNames.Storage);
 
-            if (connectionSection == null || !connectionSection.Exists())
+            if (storageConnectionSection is null || !storageConnectionSection.Exists())
             {
                 return new DiagnosticEventNullRepository();
             }
--- a/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventRepositoryFactory.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventRepositoryFactory.cs
@@ -21,11 +21,12 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 
         public IDiagnosticEventRepository Create()
         {
-            // Using this to break ciruclar dependency for ILoggers. Typically you cannot log errors within the logging pipeline because it creates infinte loop.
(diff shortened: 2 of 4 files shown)


src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -41,43 +41,30 @@ public class DiagnosticEventTableStorageRepository : IDiagnosticEventRepository,
...
                 {
-                    string storageConnectionString = _configuration.GetWebJobsConnectionString(ConnectionStringNames.Storage);
-
-                    try
-                    {
-                        _tableClient = new TableServiceClient(storageConnectionString);
-
-                        // The TableServiceClient only verifies the format of the connection string.
-                        // To ensure the storage account exists and supports Table storage, validate the connection string by retrieving the properties of the table service.
-                        _ = _tableClient.GetProperties();
```
cjaliaga, line 73: Removed this check since it could disable Diagnostic events for some scenarios where we might not want to:
- The permissions assigned to the Identity don't allow to get the properties but allow to write diagnostic events
- Transient network issue on `TableServiceClient` creation that could be auto-resolved so we could write events on that instance in the future
- Permissions: missing permissions on `TableServiceClient` creation that are either solved by the customer later on or take time to be propagated, which would fix writing diagnostic events without a restart.

cc @RohitRanjanMS

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -41,43 +40,30 @@ public class DiagnosticEventTableStorageRepository : IDiagnosticEventRepository,
...
-                    }
-                    catch (Exception ex)
-                    {
-                        _logger.LogError(ex, "The Azure Storage connection string is either empty or invalid. Unable to record diagnostic events, so the diagnostic logging service is being stopped.");
-                        _isEnabled = false;
-                        StopTimer();
-                    }
+                    _logger.LogError("The Azure Storage connection string is either empty or invalid. Unable to record diagnostic events, so the diagnostic logging service is being stopped.");
```
kshyju: Is this error message still valid? Looks like `TryCreateHostingTableServiceClient` method logs the connection string related errors in it (or the child method it calls). Wondering there could be other reasons where the `TryCreateHostingTableServiceClient` returns `false`.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
brettsam: Does this mean we could see errors somewhere else later? I assume this check was added for that reason at some point?

brettsam (commented):
Wasn't there additional work that needed to be done to really support this in the portal? Like we could write, but the portal wouldn't be able to read, so it wouldn't work anyway? I think @soninaren mentioned it...

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
cjaliaga: We added that check for the current release in #10395, to disable the repository if the storage account connection string was not present. The main issue was due to the fact that when you were using Identity based connections, we would have tons of logs about not having the connection string to be used. That situation is still resolved because with this change we support identity based connections and also disabled the repo if we fail to create the client.

The `GetProperties()` call was also added because that way we could disable the repository on a situation that we knew there was issues connecting to Table Storage, right on client initialization. However while adding support for MI, I've found and tested some cases where would be good to let it fail as they would be recoverable errors. Not all the cases ofc, for example an invalid account name would require a config change and a restart, but that one would also cause hard-failures host-wise....

After weighing it up, I decided to remove that check.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
cjaliaga: On this one we explicitly say that we're disabling the repository. I agree with you that the specific details are logged as part of the call to `TryCreateHostingTableServiceClient`. It could return `false` for example if it tries to create the `TokenCredential` and there's something incorrect in some of the configuration entries... Anything around an invalid configuration for the connection configuration section.

We could remove specifics around the storage connection string as there will be another log with that exception and write something around `An error occurred initializing the Table Storage Client. We are unable to record diagnostic events, so the diagnostic logging service is being stopped.`

What do you think?

@mattchenderson

cjaliaga:
> Wasn't there additional work that needed to be done to really support this in the portal? Like we could write, but the portal wouldn't be able to read, so it wouldn't work anyway? I think @soninaren mentioned it...

~~I believe the changes in Geo were already deployed if I remember correctly when we discussed this. @soninaren could you please confirm this?~~

Control plane changes are not yet completed.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
jviau: I like the new message you suggest.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
jviau: Should we check for a `403` response and disable the repository when we try and write?

jviau (approved):

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
cjaliaga: We could indeed, but we might still have some scenarios where the situation could be resolved (such role/permissions propagation) without requiring a host restart and on those scenarios we would be disabling prematurely. There are some SKUs where the same host can be running without being restarted for longer periods so there it could impact more.

release_notes.md
```diff
@@ -8,3 +8,4 @@
 - Improving console log handling during specialization (#10345)
 - Update Node.js Worker Version to [3.10.1](https://github.com/Azure/azure-functions-nodejs-worker/releases/tag/v3.10.1)
 - Remove packages `Microsoft.Azure.Cosmos.Table` and `Microsoft.Azure.DocumentDB.Core` (#10503)
+- Added support for Identity-based connections to Diagnostic Events (#10438)
```
liliankasem:
```suggestion
- Added support for identity-based connections to Diagnostic Events (#10438)
```

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventRepositoryFactory.cs
```diff
@@ -21,11 +21,12 @@ public DiagnosticEventRepositoryFactory(IServiceProvider serviceProvider, IConfi
 
         public IDiagnosticEventRepository Create()
         {
-            // Using this to break ciruclar dependency for ILoggers. Typically you cannot log errors within the logging pipeline because it creates infinte loop.
+            // Using this to break circular dependency for ILoggers. Typically you cannot log errors within the logging pipeline because it creates infinte loop.
             // However in this case that loop is broken because of the filtering in the DiagnosticEventLogger
 
-            string storageConnectionString = _configuration.GetWebJobsConnectionString(ConnectionStringNames.Storage);
-            if (string.IsNullOrEmpty(storageConnectionString))
+            IConfigurationSection connectionSection = _configuration.GetWebJobsConnectionSection(ConnectionStringNames.Storage);
```
liliankasem: nit: storageConnectionSection

liliankasem (approved):

release_notes.md
```diff
@@ -3,9 +3,12 @@
 <!-- Please add your release notes in the following format:
 - My change description (#PR)
 -->
-- Update Python Worker Version to [4.33.0](https://github.com/Azure/azure-functions-python-worker/releases/tag/4.33.0)
+- Update Python Worker Version to [4.34.0](https://github.com/Azure/azure-functions-python-worker/releases/tag/4.34.0)
 - Sanitize exception logs (#10443)
 - Improving console log handling during specialization (#10345)
 - Update Node.js Worker Version to [3.10.1](https://github.com/Azure/azure-functions-nodejs-worker/releases/tag/v3.10.1)
 - Remove packages `Microsoft.Azure.Cosmos.Table` and `Microsoft.Azure.DocumentDB.Core` (#10503)
+- Implement host configuration property to allow configuration of the metadata provider timeout period (#10526)
```
liliankasem, line 11: Rebase?

kbalakumar1:
@cjaliaga We are still seeing a lot of exceptions on this issue in our logs with the "Unable to get table reference or create table. Aborting write operation" error. Do we need to upgrade any library (or do any other steps) in order to get the latest changes mentioned in this thread to fix it?

PR #10476, opened 2024-09-17
Pass ARM ID header for Linux consumption metrics requests

v-shenoy:
This PR includes the Azure Resource Id of the function app as a header when publishing metrics for Linux Consumption. This change is required per the changes in internal dependencies.

    * Otherwise link to back port: #10477 

### DIFF
--- a/src/WebJobs.Script.WebHost/Metrics/LinuxContainerMetricsPublisher.cs
+++ b/src/WebJobs.Script.WebHost/Metrics/LinuxContainerMetricsPublisher.cs
@@ -27,6 +27,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Metrics
         public const string ContainerNameHeader = "x-ms-functions-container-name";
         public const string HostNameHeader = "x-ms-functions-host-name";
         public const string StampNameHeader = "x-ms-site-home-stamp";
+        public const string ResourceIdHeader = "x-ms-site-arm-id";
 
         private const string _requestUriFormat = "https://{0}:31002/api/metrics";
         private const int _maxErrorLimit = 5;
@@ -63,6 +64,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Metrics
         private Timer _metricsPublisherTimer;
         private int _errorCount = 0;
         private string _stampName;
+        private string _resourceId;
         private bool _initialized = false;
 
         public LinuxContainerMetricsPublisher(IEnvironment environment, IOptionsMonitor<StandbyOptions> standbyOptions, ILogger<LinuxContainerMetricsPublisher> logger, HostNameProvider hostNameProvider, IOptions<FunctionsHostingConfigOptions> functionsHostingConfigOptions, HttpClient httpClient = null)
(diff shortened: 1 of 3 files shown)


krishnajaju (approved):

src/WebJobs.Script.WebHost/Metrics/LinuxContainerMetricsPublisher.cs
```diff
@@ -27,6 +27,7 @@ public class LinuxContainerMetricsPublisher : IMetricsPublisher
         public const string ContainerNameHeader = "x-ms-functions-container-name";
         public const string HostNameHeader = "x-ms-functions-host-name";
         public const string StampNameHeader = "x-ms-site-home-stamp";
+        public const string ResourceIdHeader = "x-ms-site-arm-id";
```
mathewc, line 30: Is the name of this header being dictated by the downstream service, or are we establishing a new header name ourselves here?

src/WebJobs.Script.WebHost/Metrics/LinuxContainerMetricsPublisher.cs
```diff
@@ -247,6 +249,7 @@ internal async Task SendRequest<T>(ConcurrentQueue<T> activitiesToPublish, strin
         public void Initialize()
         {
             _stampName = _environment.GetEnvironmentVariable(EnvironmentSettingNames.WebSiteHomeStampName);
+            _resourceId = _environment.GetEnvironmentVariable(EnvironmentSettingNames.WebsiteArmResourceId);
```
mathewc, line 252: Looks like this WEBSITE_ARM_RESOURCE_ID env var is never set by the platform. So likely we can't rely on this currently.

v-shenoy: Yes, as discussed, I will raise a PR for the platform in order to populate this variable.

src/WebJobs.Script.WebHost/Metrics/LinuxContainerMetricsPublisher.cs
v-shenoy: We are the ones introducing this change in the host as well as the downstream, so this name is up to us.

src/WebJobs.Script.WebHost/Metrics/LinuxContainerMetricsPublisher.cs
```diff
@@ -291,6 +294,7 @@ private HttpRequestMessage BuildRequest<TContent>(HttpMethod method, string path
             request.Headers.Add(ContainerNameHeader, _containerName);
             request.Headers.Add(HostNameHeader, _hostNameProvider.Value);
             request.Headers.Add(StampNameHeader, _stampName);
+            request.Headers.Add(ResourceIdHeader, _resourceId);
```
mathewc, line 297: As discussed on Teams - I don't think we need the Functions Host to send this. It can be handled instead on the agent side - agent can receive the Resource ID mapping to the site as part of token auth.

PR #10477, opened 2024-09-17
[in-proc] Pass ARM ID header for Linux consumption metrics requests

v-shenoy:
Backport of #10476


krishnajaju (approved):

src/WebJobs.Script.WebHost/Metrics/LinuxContainerMetricsPublisher.cs
```diff
@@ -291,6 +294,7 @@ private HttpRequestMessage BuildRequest<TContent>(HttpMethod method, string path
             request.Headers.Add(ContainerNameHeader, _containerName);
             request.Headers.Add(HostNameHeader, _hostNameProvider.Value);
             request.Headers.Add(StampNameHeader, _stampName);
+            request.Headers.Add(ResourceIdHeader, _resourceId);
```
mathewc, line 297: As discussed on Teams - I don't think we need the Functions Host to send this. It can be handled instead on the agent side - agent can receive the Resource ID mapping to the site as part of token auth.

PR #10497, opened 2024-09-25
Include all release/in-proc branches for PR checks

castrodd:
This PR modifies our PR checks to run on all branches that begin with `release/in-proc`. Previously, we were only running checks on the `release/in-proc` branch, but we now have a `release/in-proc-hotfix` branch (as an example).


fabiocav:
Discussed this with @Francisco-Gamino and we're closing this for now as this shouldn't be needed. We can re-evaluate if we feel this is still required in the future.

PR #10501, opened 2024-09-26
In proc host release should publish package named `Microsoft.Azure.WebJobs.Script.WebHost.InProc`

### LINKED ISSUE #10458: In proc host release must publish package named `Microsoft.Azure.WebJobs.Script.WebHost.InProc`
We need an equivalent version of `Microsoft.Azure.WebJobs.Script.WebHost` for the inproc host to be consumed by the version of core tools that uses inproc. Package name will be `Microsoft.Azure.WebJobs.Script.WebHost.InProc`

aishwaryabh:

Resolves [#10458](https://github.com/Azure/azure-functions-host/issues/10458)

Tested in host.official pipeline that artifact is generated:
image

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -12,6 +12,7 @@ pr:
     - in-proc
     - release/4.*
     - release/in-proc
+    - aibhandari/*
 
 trigger:
   branches:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -12,7 +12,6 @@ pr:
     - in-proc
     - release/4.*
     - release/in-proc
-    - aibhandari/*
 
 trigger:
   branches:
--- a/eng/ci/code-mirror.yml
+++ b/eng/ci/code-mirror.yml
@@ -5,6 +5,7 @@ trigger:
     - in-proc
     - release/4.*
     - release/in-proc
+    - aibhandari/*
 
 resources:
   repositories:
--- a/eng/ci/code-mirror.yml
+++ b/eng/ci/code-mirror.yml
@@ -5,7 +5,6 @@ trigger:
     - in-proc
     - release/4.*
     - release/in-proc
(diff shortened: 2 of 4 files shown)


jviau:
Do all the other packages need the same change?

fabiocav (approved):

aishwaryabh:
> Do all the other packages need the same change?

@jviau I don't think so, since we want to use the `Microsoft.Azure.Webjobs.Script.WebHost` package in the in-proc core tools and OOP core tools. We just wanted a way to separate them to cause less confusion. Does that make sense?

jviau:
> > Do all the other packages need the same change?
> 
> @jviau I don't think so, since we want to use the `Microsoft.Azure.Webjobs.Script.WebHost` package in the in-proc core tools and OOP core tools. We just wanted a way to separate them to cause less confusion. Does that make sense?

Are transitive dependencies to the other packages involved though? If so, how will those impact this (if at all)?

aishwaryabh:
> > > Do all the other packages need the same change?
> > 
> > 
> > @jviau I don't think so, since we want to use the `Microsoft.Azure.Webjobs.Script.WebHost` package in the in-proc core tools and OOP core tools. We just wanted a way to separate them to cause less confusion. Does that make sense?
> 
> Are transitive dependencies to the other packages involved though? If so, how will those impact this (if at all)?

Synced with Fabio and there aren't any transitive dependencies involved to the other packages since this is specifically for the inproc version of the package and the only other dependencies for the other packages would be WebJobs and other runtime dependencies.

surgupta-msft:
This may already be considered as part of the process but I believe we also need to inform the release team that the package name is changing for "inproc"? The above changes are not only for coretools but also for the host releases for in-proc.

surgupta-msft (approved):

PR #10522, opened 2024-10-10
Do not unregister extension endpoints on shutdown

### LINKED ISSUE #10251: Service bus message actions throwing Grpc.Core.RpcException when sending multiple requests

jviau:

resolves #10251

### Pull request checklist

### Additional information

Addresses an issue where on host shutdown gRPC extension endpoints would be unregistered before functions had finished draining. This could lead to gRPC "unimplemented" errors if those functions attempted to call a gRPC extension.

The fix is to _not_ unregister extension endpoints when active host changes to null. Instead, only non-null active host changes are processed.

#### Concerns

I am unsure if this is ever a possible scenario, but there is undefined behavior if we are transitioning from one non-null active host to another. We will still switch over to the new host's endpoints (if any) before old hosts' functions complete.

**UPDATE:** Synced with @fabiocav and there are definitely scenarios this fix will not cover (one JobHost draining while another spins up). However, this is still a net-improvement (solving the common host shutdown issue). Addressing the last issue is a **much** larger work item. To address this, I am going to explore moving gRPC extension server into the JobHost itself, which will completely solve the lifetime issue among others. I will explore this at a later time, proceeding with this improvement for now.

### DIFF
--- a/src/WebJobs.Script.Grpc/Server/ExtensionsCompositeEndpointDataSource.cs
+++ b/src/WebJobs.Script.Grpc/Server/ExtensionsCompositeEndpointDataSource.cs
@@ -191,6 +191,8 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
             lock (_lock)
             {
                 // Do not clear out endpoints when host changes to null, as functions may still be running.
+                // TODO: there are still edge cases where we will switch active hosts, leading to potential
+                // issues with draining invocations.
                 if (args?.NewHost?.Services is { } services)
                 {
                     _extensionServices = services;
--- a/src/WebJobs.Script.Grpc/Server/ExtensionsCompositeEndpointDataSource.cs
+++ b/src/WebJobs.Script.Grpc/Server/ExtensionsCompositeEndpointDataSource.cs
@@ -26,10 +26,11 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
     internal sealed class ExtensionsCompositeEndpointDataSource : EndpointDataSource, IDisposable
     {
         private readonly object _lock = new();
-        private readonly List<EndpointDataSource> _dataSources = new();
         private readonly IScriptHostManager _scriptHostManager;
         private readonly TaskCompletionSource _initialized = new();
+        private readonly ILogger _logger;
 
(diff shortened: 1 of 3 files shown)


liliankasem:
> Instead, only non-null active host changes are processed.

This makes sense to me, but could we still unregister extension endpoints _after_ we finish draining invocations?

> I am unsure if this is ever a possible scenario, but there is undefined behavior if we are transitioning from one non-null active host to another. We will still switch over to the new host's endpoints (if any) before old hosts' functions complete.

This is a good point; I don't know if this is possible either but if there's anything we can do to protect against it we should. Or create a follow up design item to look into this?

jviau:
> This makes sense to me, but could we still unregister extension endpoints after we finish draining invocations?

I considered this but ultimately decided not to. The reason being if invocations are drained, who would even be calling the extension endpoints? I feel that the added complexity of that would expose this to more bugs without much of a benefit.

liliankasem (approved):

fabiocav (approved):

PR #10524, opened 2024-10-14
Bump dependency version

liliankasem:


test/WebJobs.Script.Tests/Microsoft.Azure.WebJobs.Script.WebHost.deps.json
```diff
@@ -290,7 +289,7 @@
           "Microsoft.AspNetCore.Http": "2.2.2",
           "Microsoft.Extensions.Configuration.Json": "3.1.0",
           "Microsoft.Extensions.Logging.ApplicationInsights": "2.22.0",
-          "System.Text.Encodings.Web": "8.0.0"
+          "System.Text.Encodings.Web": "4.7.2"
```
jviau: What happened here? Why did this go from 8.0.0 to 4.7.2? STJ 8.0.5 is supposed to require >= 8.0.0

liliankasem: Yeah this is really weird; I don't know why this would happen. I'll try to regenerate in case something was off the first time round

liliankasem: Yeah no change :/ @fabiocav any thoughts?

liliankasem: The only solution I have for this is to add the pkg ref for it directly:

`<PackageReference Include="System.Text.Encodings.Web" Version="8.0.0" />`

Let me know if there are other ways/ideas to address this.

fabiocav: This is being caused because `System.Text.Encodings.Web` was being pushed to `8.0.0` by `System.Text.Json` `8.0.4` (https://www.nuget.org/packages/System.Text.Json/8.0.4#dependencies-body-tab).

In version `8.0.5`, they dropped that dependency for `net8.0` (https://www.nuget.org/packages/System.Text.Json/8.0.5#dependencies-body-tab), so nothing referenced was pushing that dependency up anymore, causing it to fall back to `4.7.2`.

To be safe, you want to add an explicit reference to `System.Text.Encodings.Web` `8.0.0` and that should keep your deps as it was before (except for the STJ update of course)

jviau (approved):

surgupta-msft (approved):

fabiocav (approved):

PR #10526, opened 2024-10-15
Add metadata provider timeout host config

### LINKED ISSUE #10529: Make metadata provider timeout configurable
Make metadata provider timeout configurable via host.json

liliankasem:

resolves #10529

### Pull request checklist

    * Otherwise: Link to backporting PR: #10545
    * [x] Otherwise: Documentation issue linked to PR: **TODO:// Update host.json spec**
    * [x] Otherwise: Backport tracked by issue/PR #10545

### Additional information

Adding a metadata provider timeout setting to the host.config to make this value configurable. For logic apps, we will not have a timeout limit.

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -8,3 +8,5 @@
 - Improving console log handling during specialization (#10345)
 - Update Node.js Worker Version to [3.10.1](https://github.com/Azure/azure-functions-nodejs-worker/releases/tag/v3.10.1)
 - Remove packages `Microsoft.Azure.Cosmos.Table` and `Microsoft.Azure.DocumentDB.Core` (#10503)
+- Implement host configuration property to all configuration of the metadata provider timeout (#10526)
+  - The value can be set via `metadataProviderTimeout` in host.json and defaults to "00:00:30" (30 seconds)
--- a/release_notes.md
+++ b/release_notes.md
@@ -8,5 +8,6 @@
 - Improving console log handling during specialization (#10345)
 - Update Node.js Worker Version to [3.10.1](https://github.com/Azure/azure-functions-nodejs-worker/releases/tag/v3.10.1)
 - Remove packages `Microsoft.Azure.Cosmos.Table` and `Microsoft.Azure.DocumentDB.Core` (#10503)
-- Implement host configuration property to all configuration of the metadata provider timeout (#10526)
-  - The value can be set via `metadataProviderTimeout` in host.json and defaults to "00:00:30" (30 seconds)
+- Implement host configuration property to allow configuration of the metadata provider timeout period (#10526)
+  - The value can be set via `metadataProviderTimeout` in host.json and defaults to "00:00:30" (30 seconds).
+  - For logic apps, unless configured via the host.json, the timeout is disabled by default.
--- a/src/WebJobs.Script/Config/ScriptJobHostOptions.cs
+++ b/src/WebJobs.Script/Config/ScriptJobHostOptions.cs
@@ -135,5 +135,11 @@ namespace Microsoft.Azure.WebJobs.Script
         /// Gets or sets the telemetry mode.
         /// </summary>
         internal TelemetryMode TelemetryMode { get; set; } = TelemetryMode.ApplicationInsights;
+
+        /// <summary>
+        /// Gets or sets a value indicating the timeout duration for the function metadata provider.
+        /// Defaults to 30 seconds.
+        /// </summary>
+        public TimeSpan? MetadataProviderTimeout { get; set; }
     }
 }
--- a/src/WebJobs.Script/Config/ScriptJobHostOptions.cs
+++ b/src/WebJobs.Script/Config/ScriptJobHostOptions.cs
@@ -138,8 +138,7 @@ namespace Microsoft.Azure.WebJobs.Script
 
         /// <summary>
         /// Gets or sets a value indicating the timeout duration for the function metadata provider.
-        /// Defaults to 30 seconds.
         /// </summary>
-        public TimeSpan MetadataProviderTimeout { get; set; } = TimeSpan.FromSeconds(30);
+        public TimeSpan MetadataProviderTimeout { get; set; } = TimeSpan.Zero;
     }
 }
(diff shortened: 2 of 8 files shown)


src/WebJobs.Script/Config/ScriptJobHostOptions.cs
```diff
@@ -135,5 +135,11 @@ public string RootScriptPath
         /// Gets or sets the telemetry mode.
         /// </summary>
         internal TelemetryMode TelemetryMode { get; set; } = TelemetryMode.ApplicationInsights;
+
+        /// <summary>
+        /// Gets or sets a value indicating the timeout duration for the function metadata provider.
+        /// Defaults to 30 seconds.
```
fabiocav: Since the default is not controlled/set here, let's remove that mention (or switch to indicate a default of `null`)

fabiocav (approved):

fabiocav:
Approved... the tests seem to be unhappy with the string change :). Good change, but no good deed goes unpunished.

src/WebJobs.Script/Config/ScriptJobHostOptionsSetup.cs
```diff
@@ -59,6 +60,12 @@ public void Configure(ScriptJobHostOptions options)
             // FunctionTimeout
             ConfigureFunctionTimeout(options);
 
+            if (_environment.IsLogicApp())
```
fabiocav: Wondering if we should default to infinite, but honor if explicitly configured (i.e., keep the value if set in config)

release_notes.md
```diff
@@ -8,3 +8,5 @@
 - Improving console log handling during specialization (#10345)
 - Update Node.js Worker Version to [3.10.1](https://github.com/Azure/azure-functions-nodejs-worker/releases/tag/v3.10.1)
 - Remove packages `Microsoft.Azure.Cosmos.Table` and `Microsoft.Azure.DocumentDB.Core` (#10503)
+- Implement host configuration property to all configuration of the metadata provider timeout (#10526)
```
cjaliaga:
```suggestion
- Implement host configuration property to allow configuration of the metadata provider timeout (#10526)
```

src/WebJobs.Script/Host/FunctionMetadataManager.cs
```diff
@@ -21,9 +22,8 @@ namespace Microsoft.Azure.WebJobs.Script
 {
     public class FunctionMetadataManager : IFunctionMetadataManager
     {
-        private const string FunctionConfigurationErrorMessage = "Unable to determine the primary function script.Make sure atleast one script file is present.Try renaming your entry point script to 'run' or alternatively you can specify the name of the entry point script explicitly by adding a 'scriptFile' property to your function metadata.";
+        private const string FunctionConfigurationErrorMessage = "Unable to determine the primary function script. Make sure atleast one script file is present. Try renaming your entry point script to 'run' or alternatively you can specify the name of the entry point script explicitly by adding a 'scriptFile' property to your function metadata.";
```
mathewc: Fix "atleast"

src/WebJobs.Script/Config/ScriptJobHostOptionsSetup.cs
```diff
@@ -59,6 +60,12 @@ public void Configure(ScriptJobHostOptions options)
...
+            {
+                // Logic Apps requires no timeout
+                options.MetadataProviderTimeout = Timeout.InfiniteTimeSpan;
```
mathewc: I get that you're defaulting this to Infinite to effectively remove this completely for LAs. However, this timeout was added presumably for a reason - to catch truly runaway cases. So I'm wondering if we shouldn't just configure a longer but still reasonable timeout here? E.g. anything on the order of minutes would likely be considered a broken LA from a customer point of view, and LA needs to optimize/fix their extension.

How long are we seeing LA take in practice - what range?

src/WebJobs.Script/Config/ScriptJobHostOptionsSetup.cs
liliankasem: Don't we still want to have a timeout set for general customers?

src/WebJobs.Script/Config/ScriptJobHostOptionsSetup.cs
liliankasem: We had discussed a longer timeout period with the LA team, but they requested no timeout until they figure this out. An action item post this fix is to resolve that:

From Laveesh: `Long term, functions team does want to enforce a timeout. To ensure we honor the timeout properly, functions and logic apps PG teams will engage to figure out what the proper fixes should be on both our sides and work on them. This will be done independent of our response to this incident.`

src/WebJobs.Script/Config/ScriptJobHostOptionsSetup.cs
liliankasem: Sounds good to me!

src/WebJobs.Script/Config/ScriptJobHostOptionsSetup.cs
liliankasem: I've updated the PR too so that despite defaulting to infinite for logic apps, the configured value will be respected and can be overridden for logic apps as well.

After design conversations with LA later, we can still change the LA default to a reasonable value instead of disabling it entirely (via infinite).

src/WebJobs.Script/Host/FunctionMetadataManager.cs
```diff
@@ -221,11 +218,13 @@ private void AddMetadataFromCustomProviders(IEnumerable<IFunctionProvider> funct
             _logger.ReadingFunctionMetadataFromProvider(MetadataProviderName);
 
             var functionProviderTasks = new List<Task<ImmutableArray<FunctionMetadata>>>();
+            var metadataProviderTimeout = _scriptOptions.Value.MetadataProviderTimeout;
```
liliankasem, line 221: As long as we go through options setup, this should never be null. But do we want to protect extra just in case it somehow is? I am trying to avoid having to set a default value here as well / avoid duplication of logic

mathewc (approved):

PR #10542, opened 2024-10-18
Update runtime site handling for token audiences

mathewc:
The previous fix I made in this area in https://github.com/Azure/azure-functions-host/pull/10183 works for most SKUs. However, it turns out that for Linux EP LWAS doesn't set WEBSITE_DEPLOYMENT_ID so we don't end up adding the runtime site name to audiences on that SKU. This came up in recent CRI https://portal.microsofticm.com/imp/v5/incidents/details/534693023. Example query showing that on the latest host version with the previous fix, we still get audience validation failures:

```
All("FunctionsLogs")
| where PreciseTimeStamp > ago(1h)
| where SourceNamespace == "LWAWS"
| where Source == "Host.Authentication"
| where Summary contains "Token audience validation failed for audience"
| where HostVersion == "4.1036.2.2"
| project PreciseTimeStamp, SourceNamespace, AppName, Summary
| take 10
```

This PR addresses this by performing a normalization on incoming audience values, removing any slot component and checking for an audience match again.

### Pull request checklist

    * https://github.com/Azure/azure-functions-host/pull/10544

### DIFF
--- a/src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
+++ b/src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
@@ -25,6 +25,7 @@ namespace Microsoft.Extensions.DependencyInjection
 {
     public static class ScriptJwtBearerExtensions
     {
+        private const string DeploymentSlotSeparator = "__";
         private static double _specialized = 0;
 
         public static AuthenticationBuilder AddScriptJwtBearer(this AuthenticationBuilder builder)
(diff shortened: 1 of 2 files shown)


src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
```diff
@@ -110,15 +111,26 @@ private static IEnumerable<string> GetValidAudiences()
             string runtimeSiteName = ScriptSettingsManager.Instance.GetSetting(AzureWebsiteRuntimeSiteName);
             var audiences = new List<string>
             {
-                string.Format(SiteAzureFunctionsUriFormat, siteName),
-                string.Format(SiteUriFormat, siteName)
+                string.Format(SiteUriFormat, siteName),
+                string.Format(SiteAzureFunctionsUriFormat, siteName)
```
mathewc, line 115: The change here is just an optimization, since the SiteUriFormat is the 99% case so should be ordered first.

src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
```diff
@@ -25,6 +25,7 @@ namespace Microsoft.Extensions.DependencyInjection
 {
     public static class ScriptJwtBearerExtensions
     {
+        private const string SlotSeparator = "__";
```
liliankasem: nit: a little extra information on what "slot" is

```suggestion
        private const string DeploymentSlotSeparator = "__";
```

liliankasem (approved):

src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
```diff
@@ -158,14 +170,63 @@ private static string IssuerValidator(string issuer, SecurityToken securityToken
...
+            }
+
+            return false;
+        }
+
+        private static bool TryGetNormalizedAudience(string audience, out string normalizedAudience)
+        {
+            normalizedAudience = null;
+
+            if (!string.IsNullOrEmpty(audience))
```
fabiocav, line 200: Could we instead use the runtime site name provided by the environment? Does that environment variable provide what we need to get the proper app name? If so, that would be a more reliable way to get this information without needing to parse it from the original audience URI.

mathewc: Work was done in the previous set of changes https://github.com/Azure/azure-functions-host/pull/10183 to do just that. The issue is that it turns out that not all SKUs (LWAS) are setting WEBSITE_DEPLOYMENT_ID it turns out (other SKUs do).

We can independently update LWAS and others to ensure we're setting this, but as you know those changes would take a long time to roll out. It's much easier/faster to update the Functions host to handle this, and that also makes us less dependent on upstream.

fabiocav: Can we at least, to be safer, look for the last index of the separator?

mathewc:
After further investigation, I'm going to abandon this change for now, in favor of a platform fix tracked by https://msazure.visualstudio.com/Antares/_workitems/edit/29942043/.

PR #10548, opened 2024-10-21
Ensure we don't try to purge diagnostic events when the repository is disabled

### LINKED ISSUE #10539: DiagnosticEventTableRepository shouldn't try to purge events when the repository is disabled
We disable the `DiagnosticEventTableRepository` (and stop the timer to avoid trying to flush the logs) in case there's not a valid connection string, however this happens the first time the property `TableClient `is accessed (after the host is specialized so it's not running in placeholder mode):

https://github.com/Azure/azure-functions-host/blob/65e1bcf5d8cfa88c9887ec7704a035b81e504637/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs#L59-L85

It can happen that the first time that we access the property is when we're trying to flush the logs and try to purge the previous events. This will currently cause to throw a `NullReferenceException`  inside while listing the tables here:

https://github.com/Azure/azure-functions-host/blob/65e1bcf5d8cfa88c9887ec7704a035b81e504637/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs#L139

Inside the TableStorageHelper, since tableClient will be a null value:

https://github.com/Azure/azure-functions-host/blob/65e1bcf5d8cfa88c9887ec7704a035b81e504637/src/WebJobs.Script.WebHost/Helpers/TableStorageHelpers.cs#L101

And this sequence of events, trying to stop the timer several times:

| **Message**                                                                                                      |
|------------------------------------------------------------------------------------------------------------------|
| Purging diagnostic events with versions older than '2024-05-01'.                                                 |
| The Azure Storage connection string is either empty or invalid. Unable to record diagnostic events, so the diagnostic logging service is being stopped. |
| Stopping the flush logs timer.                                                                                   |
| Error occurred when attempting to purge previous diagnostic event versions.                                       |
| The Azure Storage connection string is either empty or invalid. Unable to record diagnostic events, so the diagnostic logging service is being stopped. |
| Stopping the flush logs timer.                                                                                   |
| Unable to get table reference. Aborting write operation.                                                         |

cjaliaga:

resolves #10539 

### Pull request checklist

    * Otherwise: #10549 

### DIFF
--- a/test/WebJobs.Script.Tests.Integration/Diagnostics/DiagnosticEventTableStorageRepositoryTests.cs
+++ b/test/WebJobs.Script.Tests.Integration/Diagnostics/DiagnosticEventTableStorageRepositoryTests.cs
@@ -255,7 +255,7 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.Integration.Diagnostics
         }
 
         [Fact]
-        public async Task FlushLogs_LogsErrorAndClearsEvents_WhenTableCreatingFails()
+        public async Task FlushLogs_OnPrimaryHost_DoesNotTryToPurgeEvents_WhenTableClientNotInitialized()
         {
             // Arrange
             IEnvironment testEnvironment = new TestEnvironment();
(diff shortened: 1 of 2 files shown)


test/WebJobs.Script.Tests.Integration/Diagnostics/DiagnosticEventTableStorageRepositoryTests.cs
```diff
@@ -255,7 +255,7 @@ await TestHelpers.Await(async () =>
         }
 
         [Fact]
-        public async Task FlushLogs_LogsErrorAndClearsEvents_WhenTableCreatingFails()
```
cjaliaga, line 258: Removed this test because it was no longer valid. Events wouldn't be added to the list of pending to be flushed because we're checking if the `TableClient` is valid on `WriteDiagnosticEvent`.

We're not reaching the part were they would actually be removed because it would return early here: https://github.com/Azure/azure-functions-host/blob/4295b50cf19c4e1be3770321229365f19675473a/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs#L188

That scenario is valid, for example when it fails to create the table but the connection string/identity is initially valid. There we would store events and then we would try to flush them.

jviau (approved):

fabiocav (approved):

liliankasem (approved):

PR #10550, opened 2024-10-21
Address LanguageWorkerOptions null in FunctionMetadataManager

### LINKED ISSUE #9859: NullRef in GetFallbackWorkerConfig
I was just investigating a CRI and I saw the following null ref in the logs. Looking more widely I see it's happening across many apps and versions (v4) on different codepaths. Here's the stacktrace:

```
System.NullReferenceException : Object reference not set to an instance of an object.
   at Microsoft.Azure.WebJobs.Script.FunctionMetadataManager.GetFallbackWorkerConfig() at /_/src/WebJobs.Script/Host/FunctionMetadataManager.cs : 129
   at Microsoft.Azure.WebJobs.Script.FunctionMetadataManager.LoadFunctionMetadata(Boolean forceRefresh,Boolean includeCustomProviders,IFunctionInvocationDispatcher dispatcher,IList`1 workerConfigs) at /_/src/WebJobs.Script/Host/FunctionMetadataManager.cs : 137
   at Microsoft.Azure.WebJobs.Script.FunctionMetadataManager.GetFunctionMetadata(Boolean forceRefresh,Boolean applyAllowList,Boolean includeCustomProviders,IList`1 workerConfigs) at /_/src/WebJobs.Script/Host/FunctionMetadataManager.cs : 90
   at Microsoft.Azure.WebJobs.Script.WebHost.Management.WebFunctionsManager.GetFunctionsMetadata(Boolean includeProxies,Boolean forceRefresh) at /_/src/WebJobs.Script.WebHost/Management/WebFunctionsManager.cs : 74
   at async Microsoft.Azure.WebJobs.Script.WebHost.Management.WebFunctionsManager.GetFunctionsMetadata(??)
```

jviau:

resolves #9859 

### Pull request checklist

### Additional information

This PR reverts a portion of #9264 and uses a different approach to solve stale function metadata. This alternative approach also addresses the null-ref from #9859.

#### Background

#9264 set out to solve function metadata being calculated and cached with a stale set of worker configs. To address this, the PR allowed for worker configs to be explicitly supplied to the `GetFunctionMetadata` method. When configs were not explicitly supplied, it would get 'fallback' configs from the current service provider. However, that provider is the active host. If this was called too early in the hosts lifetime, the active host would be `null` (leading to the null-ref).

Examining the issue it is a cache invalidation problem. `GetFunctionMetadata` will get the current value of `LanguageWorkerOptions`, then build **and cache** the function metadata results. That caching is the problem, as language worker options changes over the lifetime of the functions host.

#### Fix

To address the null ref and after taking a closer look at the root issue, a new approach has been taken. Instead of explicitly supplying the RPC configs (which all 2 instances of that were from the current value of `LanguageWorkerOptions`), we go back to the original approach of getting the current value of those options. With one major difference: we now subscribe to the `OnChange` of those options and will invalidate the cache whenever they change. The next call to `GetFunctionMetadata` will then re-build and cache the function metadata set.

#### Additional Information

#9264 would get the `LanguageWorkerOptions` from the job host. This PR now gets them from the script host (parent of job host). These options are **identical** between the two hosts. PR #10369 even sets out to make the job hosts copy of these options a direct forwarding of the script hosts copy.


liliankasem (dismissed):

jviau:
Was testing this further and I am seeing an actual discrepancy between the worker configs between WebHost and ScriptHost, will need to investigate that further.

fabiocav:
Blocked on #10369

kshyju:
> Blocked on #10369
#10369 is merged now.

jviau:
@fabiocav @kshyju I have updated this PR now that the blocker is checked in. It is ready for review. Thanks!

src/WebJobs.Script/Host/FunctionMetadataManager.cs
```diff
@@ -49,6 +53,9 @@ public FunctionMetadataManager(IOptions<ScriptJobHostOptions> scriptOptions, IFu
             _isHttpWorker = httpWorkerOptions?.Value?.Description != null;
             _environment = environment;
 
+            _languageOptions = languageOptions;
+            _onChangeSubscription = languageOptions.OnChange(_ => _servicesReset = true);
```
kshyju, line 57: Should updating this class level local variable be inside a lock or `Interlocked.Exchange`? I am wondering there could be a race condition if the `OncChange` and `GetFunctionMetdata` public method gets invoked at the same time which might cause any sort of unintentional behavior.

jviau: Read and write of boolean is atomic, so no need to wrap in `Interlocked`.

src/WebJobs.Script/Host/FunctionMetadataManager.cs
```diff
@@ -7,40 +7,44 @@
...
+    public sealed class FunctionMetadataManager : IFunctionMetadataManager, IDisposable
     {
         private const string FunctionConfigurationErrorMessage = "Unable to determine the primary function script. Make sure at least one script file is present. Try renaming your entry point script to 'run' or alternatively you can specify the name of the entry point script explicitly by adding a 'scriptFile' property to your function metadata.";
         private const string MetadataProviderName = "Custom";
         private readonly IServiceProvider _serviceProvider;
-        private IFunctionMetadataProvider _functionMetadataProvider;
+        private readonly IFunctionMetadataProvider _functionMetadataProvider;
+        private readonly IEnvironment _environment;
+
+        private IOptionsMonitor<LanguageWorkerOptions> _languageOptions;
```
kshyju: Should these be also marked `readonly`?

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
```diff
@@ -84,18 +84,17 @@ public async Task<IEnumerable<Type>> GetExtensionsStartupTypesAsync()
             bool isPrecompiledFunctionApp = false;
 
             // dotnet app precompiled -> Do not use bundles
-            var workerConfigs = _languageWorkerOptions.CurrentValue.WorkerConfigs;
```
kshyju, line 87: The `IOptionsMonitor<LanguageWorkerOptions>` dependency is nolonger used in this class. Should we consider removing it from being injected?

kshyju (dismissed):

kshyju (dismissed):

kshyju (approved):

surgupta-msft (approved):

PR #10555, opened 2024-10-22
Fixing formattedMessage bug

RohitRanjanMS:
Fixing formattedMessage sanitize logic.

### Pull request checklist

### Additional information

Additional PR information


cjaliaga (approved):
We need to ensure that `formattedMessage` has either a value or empty string. `Sanitizer.Sanitize` checks that and we will return early if `formattedMessage` is one of those so LGTM!

src/WebJobs.Script.WebHost/Diagnostics/SystemLogger.cs
```diff
@@ -181,8 +180,7 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
                 formattedMessage = Sanitizer.Sanitize(formattedMessage);
                 innerExceptionMessage = innerExceptionMessage ?? string.Empty;
             }
-
-            _eventGenerator.LogFunctionTraceEvent(logLevel, subscriptionId, appName, functionName, eventName, source, details, summary, innerExceptionType, innerExceptionMessage, invocationId, _hostInstanceId, activityId, runtimeSiteName, slotName, DateTime.UtcNow);
+            _eventGenerator.LogFunctionTraceEvent(logLevel, subscriptionId, appName, functionName, eventName, source, details, formattedMessage, innerExceptionType, innerExceptionMessage, invocationId, _hostInstanceId, activityId, runtimeSiteName, slotName, DateTime.UtcNow);
```
jviau, line 181: Null coalesce `formattedMessage` to empty string.

jviau:
> We need to ensure that `formattedMessage` has either a value or empty string. `Sanitizer.Sanitize` checks that and we will return early if `formattedMessage` is one of those so LGTM!

The sanitize only happens if there is an exception. We need to ensure it is non-null in other scenarios.

jviau (dismissed):

src/WebJobs.Script.WebHost/Diagnostics/SystemLogger.cs
RohitRanjanMS: There is already a check for [`IsNullEmpty`](https://github.com/Azure/azure-functions-host/blob/351083550777eadf9fe793823ea03d7be1c33e00/src/WebJobs.Script.WebHost/Diagnostics/SystemLogger.cs#L122)`

cjaliaga:
> > We need to ensure that `formattedMessage` has either a value or empty string. `Sanitizer.Sanitize` checks that and we will return early if `formattedMessage` is one of those so LGTM!
> 
> The sanitize only happens if there is an exception. We need to ensure it is non-null in other scenarios.

Reading it now, it doesn't say what I wanted to say 🥲. I mean that `Sanitizer.Sanitize` checks that **and** that we also return early here, so we should be covered:

https://github.com/Azure/azure-functions-host/blob/351083550777eadf9fe793823ea03d7be1c33e00/src/WebJobs.Script.WebHost/Diagnostics/SystemLogger.cs#L122-L125

liliankasem (approved):

jviau (approved):

liliankasem (approved):

PR #10565, opened 2024-10-23
Cancel drain/resume calls on client disconnect

jviau:

resolves #issue_for_this_pr

### Pull request checklist

### Additional information

This PR mitigates a bug with our drain and resume flow. We have seen instances where a previous drain call can still be holding the semaphore when a new call comes in. In this scenario the client will eventually disconnect but we will still process the drain call when the previous one releases the semaphore. This leads to a drain being activated well after the intended time. This has been observed as causing out of sync issues between data role and host instance as the drain activates at a time the data role is not expecting it.

To mitigate this, we will abort the drain call if the client disconnects. This will reduce the chance we process a drain well after client disconnect.

**NOTE:** the cancellation token is intentionally not passed into `EnableDrainModeAsync` as it is a fire & forget background task. Passing in the token could cause unknown behavior.

### DIFF
--- a/src/WebJobs.Script.WebHost/Controllers/HostController.cs
+++ b/src/WebJobs.Script.WebHost/Controllers/HostController.cs
@@ -202,6 +202,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Controllers
                 await _drainSemaphore.WaitAsync(cancellation);
 
                 // Stop call to some listeners gets stuck, not waiting for the stop call to complete
+                // We do not pass in this APIs cancellation token as we do not want drain to be aborted once we have the semaphore.
                 _ = drainModeManager.EnableDrainModeAsync(CancellationToken.None)
                                     .ContinueWith(
                                         antecedent =>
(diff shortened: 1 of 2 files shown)


src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -197,21 +197,29 @@ public async Task<IActionResult> Drain([FromServices] IScriptHostManager scriptH
                 return StatusCode(StatusCodes.Status503ServiceUnavailable);
             }
 
-            await _drainSemaphore.WaitAsync();
+            try
+            {
+                await _drainSemaphore.WaitAsync(cancellation);
```
mathewc, line 202: I think we'd want the same handling for the resume API as well, right? We have the same issue on that API.

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -197,21 +197,29 @@ public async Task<IActionResult> Drain([FromServices] IScriptHostManager scriptH
...
 
-            // Stop call to some listeners gets stuck, not waiting for the stop call to complete
-            _ = drainModeManager.EnableDrainModeAsync(CancellationToken.None)
-                                .ContinueWith(
-                                    antecedent =>
-                                    {
-                                        if (antecedent.Status == TaskStatus.Faulted)
+                // Stop call to some listeners gets stuck, not waiting for the stop call to complete
+                _ = drainModeManager.EnableDrainModeAsync(CancellationToken.None)
```
mathewc, line 206: Should we also be using the cancellation token here as well?

jviau: No - I have a comment in the PR description about that. I should add it to the code directly though. This is a fire & forget task, I am uncertain what the behavior would be if it is kicked off and then AspNetCore signals cancellation. So long as we make it past the semaphore this change will address the main issue it set out to

mathewc: Missed that, thanks.

src/WebJobs.Script.WebHost/Controllers/HostController.cs
jviau: hmm yeah. I wasn't going to originally, but I think thats a good idea to include it. We could have a resume call start at some random time when we thought it was draining.

brettsam (approved):

mathewc (approved):

PR #10568, opened 2024-10-24
Updating gitignore

mathewc:
Ensuring that test files don't get checked in. See https://github.com/Azure/azure-functions-host/commit/5c05252b13b534e257acfd987483e1bd0217aa8b. Also recently we've seen cases where local host.json files used for testing were also showing up in PRs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,6 +65,7 @@ BenchmarkDotNet.Artifacts/
 project.lock.json
 project.fragment.lock.json
 artifacts/
+appsettings.Development.json
 
 # Tye
 .tye/
@@ -498,3 +499,4 @@ local.settings.json
 /tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator/runtimeassemblies.txt
 /tools/ExtensionsMetadataGenerator/test/ExtensionsMetadataGeneratorTests/runtimeAssemblies.txt
 BenchmarkDotNet.Artifacts/
+/src/WebJobs.Script.WebHost/host.json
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -65,6 +65,7 @@ BenchmarkDotNet.Artifacts/
 project.lock.json
 project.fragment.lock.json
 artifacts/
+appsettings.Development.json
```
jviau, line 68: This is meant to be checked in. These are common settings shared for local development. If you want user-specific settings, user-secrets should be used

jviau (changes requested):

.gitignore
mathewc: We just deleted the file in https://github.com/Azure/azure-functions-host/commit/5c05252b13b534e257acfd987483e1bd0217aa8b. Checking the file in leads to potential issues as you can see. With the file excluded, devs can have whatever local file they wish with their custom settings.

jviau: The problem is not that the file was checked in, but that the dev put that secret in the wrong location. What if a secret is added to `appsettings.json`? Will we exclude that also?

jviau: There is a solution to this: https://learn.microsoft.com/en-us/aspnet/core/security/app-secrets?view=aspnetcore-8.0&tabs=windows

This allows for storing of app-settings outside of the repository, so they will not be included in git.

mathewc: Yeah, we don't want either file checked in really, but when we looked to remove appsettings.json we realized that the host IS actually loading that one, so for now we didn't want to remove that. At least in one case, we've had devs check secrets in via these files, and also it's a problem that our site extension package process is including these files as part of its build. The idea was to just remove the issue at its source.

@fabiocav and I had discussed and he also agreed that the file shouldn't exist in the repo (hence the deletion).

jviau: We could address the packaging process issue by excluding the file from publish (`CopyToPublishDirectory="false"` in msbuild).

I just see ignoring this file as addressing a symptom, not the issue. The issue is someone is not aware of user secrets (right click project in VS -> User Secrets).

Issue #10504, opened 2024-09-30
Remove in-proc support from OOP host - Part 2


fabiocav:
Item tracking full removal of in-proc functionality from the OOP host (part 2)

PR #10573, opened 2024-10-25
Initial block for .NET in-proc indexing in the OOP host.

### LINKED ISSUE #10481: Remove in-proc support from OOP host - Part 1
Item tracking blocking of in-proc indexing from the OOP host (part 1).

Part 2 will remove the capability.

fabiocav:
Immediate fix to ensure proper handling of in-proc payloads in environments where it isn't supported. This defaults to Flex consumption today, with support for hosting configuration.

This currently ignores restrictions for placeholders to avoid potential impact to cold start. The downside is that it misses the opportunity to reduce memory consumption in those cases, but that enhancement will came in #10504 

resolves #10481 

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script/Environment/EnvironmentExtensions.cs
+++ b/src/WebJobs.Script/Environment/EnvironmentExtensions.cs
@@ -115,6 +115,12 @@ namespace Microsoft.Azure.WebJobs.Script
                    !string.IsNullOrEmpty(environment.GetEnvironmentVariable(AzureFilesContentShare));
         }
 
+        public static bool IsDotNetInProcSupported(this IEnvironment environment)
+        {
+            return !IsFlexConsumptionSku(environment)
+                   && !string.Equals(environment.GetEnvironmentVariable(EnvironmentSettingNames.FunctionsDisableInProc), "1", StringComparison.OrdinalIgnoreCase);
+        }
+
         public static string GetAzureWebsiteHomePath(this IEnvironment environment)
         {
             return environment.GetEnvironmentVariable(AzureWebsiteHomePath);
--- a/src/WebJobs.Script/Host/ScriptHost.cs
+++ b/src/WebJobs.Script/Host/ScriptHost.cs
@@ -737,7 +737,7 @@ namespace Microsoft.Azure.WebJobs.Script
                         scriptFile);
 
                     // Adding a function error will cause this function to get ignored
-                    Utility.AddFunctionError(this.FunctionErrors, metadata.Name, msg);
+                    Utility.AddFunctionError(FunctionErrors, metadata.Name, msg);
 
                     _logger.ConfigurationError(msg);
                 }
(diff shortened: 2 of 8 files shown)


src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -115,6 +115,12 @@ public static bool AzureFilesAppSettingsExist(this IEnvironment environment)
                    !string.IsNullOrEmpty(environment.GetEnvironmentVariable(AzureFilesContentShare));
         }
 
+        public static bool IsDotNetInProcSupported(this IEnvironment environment)
+        {
+            return !IsFlexConsumptionSku(environment)
```
pragnagopa, line 120: add unit tests?

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -115,6 +115,12 @@ public static bool AzureFilesAppSettingsExist(this IEnvironment environment)
...
+                   && !string.Equals(environment.GetEnvironmentVariable(EnvironmentSettingNames.FunctionsDisableInProc), "1", StringComparison.OrdinalIgnoreCase);
+        }
```
pragnagopa, line 122: Does `EnvironmentSettingNames.FunctionsDisableInProc` default to true for FlexConsumption or does it have to be set to true in dockerfile for FlexConsumption image?

fabiocav: No. This gives you the ability to set this outside of Flex or an environment where a hosting config is applied. Flex will be automatically picked up by the SKU check.

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
fabiocav: Validation was performed on this. I don't expect we'd have this logic here for longer than a release, but I can follow up with some additional tests.

src/WebJobs.Script/Host/ScriptHost.cs
```diff
@@ -823,15 +823,33 @@ internal async Task<Collection<FunctionDescriptor>> GetFunctionDescriptorsAsync(
...
+                var inProcIndexingSupported = _environment.IsPlaceholderModeEnabled()
+                                              || (_environment.IsDotNetInProcSupported()
+                                                  && !_hostingConfigOptions.Value.IsDotNetInProcDisabled);
+
                 foreach (FunctionMetadata metadata in functions)
                 {
                     try
                     {
-                        bool created = false;
+                        // If this is metadata represents a function that supports direct type indexing (in-proc), log a warning and skip it.
```
mathewc: If I'm reading the below code correctly, it looks like we won't skip the function like you say with a warning, but you throw an uncaught error?

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
mathewc: Yeah, would be good to add tests for these new helpers.

src/WebJobs.Script/Host/ScriptHost.cs
fabiocav: That's correct. The comment is stale from an initial version that was going to emit a warning. Will update.

pragnagopa (approved):

mathewc (approved):

pragnagopa (approved):

fabiocav:

PR #10589, opened 2024-11-01
Set the MetadataProvider timeout to InfiniteTimeSpan for LA

liliankasem:
Update the timeout period for the MetadataProvider so that it is set to InfiniteTimeSpan when the environment is a Logic App.


fabiocav (approved):

amamounelsayed:
Please @liliankasem could you update the release notes? thank you so much

PR #10597, opened 2024-11-04
Return metadata provider task whether it is successful or not

liliankasem:
With the current behaviour, even if the `getFunctionMetadataFromProviderTask` completes before the timeout period, we get a timeout exception if it is not successful. This is a bug and ignores potential failures from the task.


fabiocav (approved):
Approving. A test here with a failed task would be good to cover the scenario (not blocking)

liliankasem:
> Approving. A test here with a failed task would be good to cover the scenario (not blocking)

I had modified a test locally to validate this behaviour but never check it in. I will do so in the dev/in-proc PRs to save time here on the hotfix.

satvu (approved):

PR #10168, opened 2024-05-15
Add SiteExtension.proj to replace build-extension.ps1

jviau:
### Issue describing the changes in this PR

resolves #issue_for_this_pr

### Pull request checklist

**IMPORTANT**: Currently, changes must be backported to the `in-proc` branch to be included in Core Tools and non-Flex deployments.

* [ ] Backporting to the `in-proc` branch is not required
    * Otherwise: Link to backporting PR -- TODO
* [x] My changes **do not** require documentation changes
    * [ ] Otherwise: Documentation issue linked to PR
* [x] My changes **should not** be added to the release notes for the next release
    * [ ] Otherwise: I've added my notes to `release_notes.md`
* [x] My changes **do not** need to be backported to a previous version
    * [ ] Otherwise: Backport tracked by issue/PR #issue_or_pr
* [x] My changes **do not** require diagnostic events changes
    * Otherwise: I have added/updated all related diagnostic events and their documentation (Documentation issue linked to PR)
* [x] I have added all required tests (Unit tests, E2E tests)

### Additional information

This PR introduces a new project `WebHost.Script.SiteExtension`, which is a replacement for `build-extension.ps1`. This project can be used via `dotnet` CLI to restore, build, and publish all the artifacts needed for the Antares site extension. Note that this project has _support_ for multi-targeting, but only targets a single TFM. This is in preparation for porting to `in-proc` branch to keep the diff minimal between the two.

### Other Changes

- Removes `update-hostreferences.ps1`, this is no longer used anywhere
- Groups all worker versions into `workers.props` file. Left python out due to code owners, but ideally python will be moved there as well.
- zip-after-publish is extracted into a set of re-usable targets `ZipPublish.targets`.
- `RuntimeIdentifiers` added to `WebHost.csproj`. Having this set of identifiers listed improves the restore task (otherwise each time you want to use a different runtime, you have to re-restore).

PR #10649, opened 2024-11-21
[in-proc backport] Replace build-extension.ps1 with SiteExtension.csproj

jviau:

This is the in-proc backport for #10168 

### Pull request checklist

### Additional information

Replaces `build-extension.ps1` with an msbuild project to build our site extension for net6.0 and net8.0 simultaneously.


.gitignore
```diff
@@ -499,6 +499,4 @@ local.settings.json
 /tools/ExtensionsMetadataGenerator/test/ExtensionsMetadataGeneratorTests/runtimeAssemblies.txt
 
 local.settings.json
-msbuild.binlog
```
jviau, line 502: these are already in gitignore

liliankasem (approved):

jviau:
Current blocker for this PR:

Part of this change was to leverage msbuild project dependencies to build in parallel and avoid double-building when we publish the separate net6 and net8 site extensions. A result of this is changing how the version property resolves: using TFM instead of a supplied parameter. This has led to an issue: `WebJobs.Script` is **not** multi-targeted like our other assemblies, meaning it is `4.37.0` when others are `4.{6|8}37.0`. And this has real impact as we use `WebJobs.Script`'s assembly version as the reported version everywhere, meaning we will report `4.37.0` with this change.

Fix options:

1. Multi-target `WebJobs.Script` to `net6/net8`
    - I think we should do this eventually as my testing shows it brings in extra analyzers, which are catching potentially real issues with that assembly.
2. Explicitly build each TFM and use old versioning logic.


I will go with option 2 for now, then follow up on option 1 in another PR.

satvu (approved):

build/python.props
```diff
@@ -1,5 +1,6 @@
 <Project>
-  <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.Functions.PythonWorker" Version="4.34.0" />
-  </ItemGroup>
+  <!-- See /eng/build/Workers.Python.props -->
+  <Target Name="_DoNotImport" BeforeTargets="Build">
+    <Error Text="This file should not be imported. Import '$(TargetsRoot)Workers.props'." />
```
hallvictoria, line 4: Should this be `Import '$(WorkersProps)'.`? Based off of what's currently in the dev branch

hallvictoria (approved):

build/python.props
jviau: That change will come in a follow up PR to port from dev.

PR #10664, opened 2024-12-03
Refactor msbuild props and targets

jviau:
### Pull request checklist

### Additional information

This PR refactors our msbuild targets & props.

- MSBuild files restructured, removing msbuild files in `build/` and spreading that logic around in `eng/build`.
- Versioning scheme overhauled.
  - Versions are now controlled for directory cones by adding a `Directory.Version.props` file.
  - Versioning scheme for non-release builds now calculates a suffix based on build reason and a number calculated from the current date.
  - `dev` for local, `pr` for PR builds, `ci` for all other non-release CI builds. eg: `4.38.0-dev.24306.0` for a local build.
- Adds logic for incorporating release notes and README into nuget packages
   - not particularly needed for this repo, but I am using this as a staging ground for a set of reusable targets that will eventually be packaged up and consumed in all our repos.
- `initialize-pipeline.yml` stage removed as it is no longer needed (this work is not done directly in msbuild)
- Placeholder configurations removed and replaced with a props file approach that is uncommented only when needed
   - Has a safeguard to fail CI if it is accidentally included in a PR uncommented
- `DotnetIsolated*.csproj` projects now built as a direct dependency of integration test _project_ (and not in the test itself).
   - Encountered a bug (https://github.com/Azure/azure-functions-dotnet-worker/issues/2888) with our aspnetcore worker analyzers. Disabled analyzers to unblock the build.


liliankasem (approved):

eng/build/Engineering.props
```diff
@@ -0,0 +1,37 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <AdditionalFiles Include="$(RepoRoot)stylecop.json" Link="stylecop.json" />
+  </ItemGroup>
+
+  <Import Project="$(MSBuildThisFileDirectory)Release.props" />
+
+  <!-- Uncomment below to run with placeholder simulation enabled. -->
+  <!-- <Import Project="$(MSBuildThisFileDirectory)PlaceholderSimulation.props" /> -->
```
jviau: @kshyju want to highlight this change for you. The additional configurations were causing pain points with this refactor so I went with an alternative approach: uncomment this line when you want to test placeholder simulation. The file has a safeguard to fail the build in CI if it is included uncommented in a PR.

If you want a different approach let me know. Another approach is to set an environment variable before build / before opening VS.

liliankasem: I would much prefer the environment variable approach over comment/uncommenting code. Seems like something easy to forget, then you have to wait for the CI to start before you realize you forgot to uncomment it etc. etc.

eng/build/Release.targets
```diff
@@ -0,0 +1,33 @@
...
+    Release.targets: targets relating to the release integrity of our product.
+    - Nuget packaging targets
+    - Versioning targets
+  -->
+
+  <PropertyGroup>
+    <BeforePack>$(BeforePack);GetReleaseNotes;AddTagToReleaseNotes</BeforePack>
+  </PropertyGroup>
+
+  <!-- This target lets us define a release notes file per package. -->
```
liliankasem, line 13: Do we want to consider a similar approach to the Azure SDK repo with one long running change log file?

https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/storage/Microsoft.Azure.WebJobs.Extensions.Storage.Blobs/CHANGELOG.md

eng/res/icon.png
liliankasem, line 1: ?

eng/build/Release.targets
jviau: I view `CHANGELOG.md` as a parallel solution to `RELEASE_NOTES.md`.

Changelog is long running and release_notes is for an individual release. It may be valuable for us to include both, but that would be separate work.

eng/res/icon.png
jviau: Moved directories - not a new file.

liliankasem (approved):
lgtm, minor comments above ^

eng/build/Engineering.props
kshyju: I’m not in favor of the uncomment/comment approach. Using configurations made it straightforward to toggle between placeholder and non-placeholder scenarios during Visual Studio F5 debugging with a simple dropdown selection. Additionally, it allowed us to generate artifacts using `dotnet publish -c ReleasePlaceholder`, which could be used for local benchmarking, measurement, and end-to-end validations, as well as potentially utilized in a CI environment. I’d prefer to retain that level of flexibility if possible.

jviau: @kshyju the property/env variable approach sounds more suitable then. 

1. For publish it will be done via:  `dotnet publish -c release -p:PlaceholderSimulation=true`
2. For local dev, you can set `$env:PlaceholderSimulation = true` before launching VS or VSCode.

Is that acceptable?

kshyju: That is certainly better than comment/uncomment approach. I am curious to learn about the issues you ran into while keeping the additional configurations.

jviau: The issues are with the config mapping in the solution file. VS kept complaining about invalid config mapping. This wasn't the first time that config mapping got in the way of MSBuild refactors, so I opted to remove the config entirely.

eng/ci/templates/official/jobs/build-artifacts-windows.yml
```diff
@@ -51,7 +46,7 @@ jobs:
     inputs:
       command: custom
       custom: pack
-      arguments: -p:BuildNumber=$(buildNumber) -c release $(packSuffixSwitch) -o $(nuget_package_path)
```
kshyju, line 54: Don't we need to pass `BuildNumber` argument anymore?

test/DotNetIsolatedUnsupportedWorker/DotNetIsolatedUnsupportedWorker.csproj
```diff
@@ -6,6 +6,7 @@
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
     <FunctionsEnableWorkerIndexing>True</FunctionsEnableWorkerIndexing>
+    <RunAnalyzers>false</RunAnalyzers>
```
kshyju, line 9: Should we add a "TO DO" to revert this change once https://github.com/Azure/azure-functions-dotnet-worker/issues/2888 is fixed?

kshyju (approved):
Approving so we can take if forward and iterate as needed.

test/DotNetIsolatedUnsupportedWorker/DotNetIsolatedUnsupportedWorker.csproj
jviau: Not really necessary, we have no need for analyzers in this project as it is only for a test scenario (which does not care about analyzers)

eng/ci/templates/official/jobs/build-artifacts-windows.yml
jviau: Nope! The versioning refactor has this all calculated in msbuild.

1. `BuildNumber` is now a "short date" calculation (taken from dotnet's versioning scheme). `SHORT_DATE := yy * 1000 + MM * 50 + dd `. The need for this calculation is to ensure it is less than 65,535, which is a windows limitation for build numbers.
2. I also have incorporation for a "build counter" to differentiate similar short dates, but that may need more testing/work (not super important). It is supposed to be an incrementing counter from 0 for the same day. so `{ShortDate}.{Counter}`

I moved it all to msbuild so that you get the same versioning scheme between local and CI builds without any special setup.

gavin-aguiar (approved):

PR #10678, opened 2024-12-06
Using hostMetadataProvider for CreateOrUpdate call in FunctionController

### LINKED ISSUE #10631: Azure Functions Host 1036 results in a 500 Error on a PUT to /admin/functions
For issues in production, please check for a solution to common issues in the Azure portal before opening a bug. In the Azure portal, navigate to your function app, select `Diagnose and solve problems` from the left, and view relevant dashboards before opening your issue.

Please provide the following:

- Timestamp: 2024-11-13 22:13:56.4826033 to 2024-11-13 22:19:42.4812748
- Function App version: 4.1036.3.23284
- Function App name: testfinnodeFGG
- Function name(s) (as appropriate):HttpTrigger1
- Invocation ID:
- Region: East US

#### Repro steps

1) Create a new Function App targeting Node 18 or 20 to allow Portal Create.
2) Try and add a Function via the Portal and see that it fails with 500 Server error. 
3) The Function does get created but the Script Host restarts. 
4) Pin back to 4.36.0 and try step 2 again. There are no errors. 

#### Expected behavior

1) The Function create should not result in a 500 Server error.

#### Actual behavior

Provide a description of the actual behavior observed.

1) The Portal Function Function creation calls /admin/functions and this call returrns a 500 error but the files do get created.
2) The Host detects a file change notification from the Put and Initiates a restart.
3) There is an exception during the restart with this call stack.

System.ObjectDisposedException : Cannot access a disposed object. 
Object name: 'IServiceProvider'.
at Microsoft.Extensions.DependencyInjection.ServiceLookup.ThrowHelper.ThrowObjectDisposedException()
at Microsoft.Extensions.DependencyInjection.ServiceProvider.GetService(ServiceIdentifier serviceIdentifier,ServiceProviderEngineScope serviceProviderEngineScope)
at Microsoft.Extensions.DependencyInjection.ServiceLookup.ServiceProviderEngineScope.GetService(Type serviceType)
at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService(IServiceProvider provider,Type serviceType)
at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService[T](IServiceProvider provider)
at Microsoft.AspNetCore.Mvc.ObjectResult.ExecuteResultAsync(ActionContext context)
at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.InvokeResultAsync(IActionResult result)
at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.ResultNext[TFilter,TFilterAsync](State& next,Scope& scope,Object& state,Boolean& isCompleted)

The ARM Template is from DataDog and an example deployment creates the Files in the ARM Template e.g [datadog-serverless-functions/azure/deploy-to-azure/function_template.json at master · DataDog/datadog-serverless-functions](https://github.com/DataDog/datadog-serverless-functions/blob/master/azure/deploy-to-azure/function_template.json)

#### Known workarounds

Provide a description of any known workarounds.

Pin back to 4.36

#### Related information 

* Programming language used: Node
* Links to source : https://github.com/DataDog/datadog-serverless-functions/tree/master/azure/activity_logs_monitoring
* Bindings used : HTTP

soninaren:

We encountered 500s from the CreateUpdate flow from FunctionController. This was caused because of multiple host restarts that used to happen during this flow. @fabiocav fixed host restarts in this [PR](https://github.com/Azure/azure-functions-host/pull/10638). However we still had one remaining when forceRefresh was set to true. This caused the function host to restart and resulted in 500s. The forceRefresh caused host restart while fetching the metadata from function workers. However in this flow we do not need to fetch the metadata from the workers. So to resolve this we are directly using HostMetadataProvider. This prevents the restart and still gives us fresh metadata as we reading the function files again.

Resolves  #10631 

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/Management/WebFunctionsManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/WebFunctionsManager.cs
@@ -11,6 +11,8 @@ using Microsoft.AspNetCore.Http;
 using Microsoft.Azure.WebJobs.Script.Description;
 using Microsoft.Azure.WebJobs.Script.Management.Models;
 using Microsoft.Azure.WebJobs.Script.WebHost.Extensions;
+using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Newtonsoft.Json;
@@ -27,8 +29,11 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
         private readonly IFunctionsSyncManager _functionsSyncManager;
         private readonly HostNameProvider _hostNameProvider;
         private readonly IFunctionMetadataManager _functionMetadataManager;
+        private readonly IHostFunctionMetadataProvider _hostFunctionMetadataProvider;
+        private readonly IOptionsMonitor<LanguageWorkerOptions> _languageWorkerOptions;
 
-        public WebFunctionsManager(IOptionsMonitor<ScriptApplicationHostOptions> applicationHostOptions, ILoggerFactory loggerFactory, IHttpClientFactory httpClientFactory, ISecretManagerProvider secretManagerProvider, IFunctionsSyncManager functionsSyncManager, HostNameProvider hostNameProvider, IFunctionMetadataManager functionMetadataManager)
+        public WebFunctionsManager(IOptionsMonitor<ScriptApplicationHostOptions> applicationHostOptions, ILoggerFactory loggerFactory, IHttpClientFactory httpClientFactory, ISecretManagerProvider secretManagerProvider, IFunctionsSyncManager functionsSyncManager, HostNameProvider hostNameProvider, IFunctionMetadataManager functionMetadataManager, IHostFunctionMetadataProvider hostFunctionMetadataProvider,
+            IOptionsMonitor<LanguageWorkerOptions> languageWorkerOptions)
         {
             _applicationHostOptions = applicationHostOptions;
             _logger = loggerFactory?.CreateLogger(ScriptConstants.LogCategoryHostGeneral);
@@ -37,6 +42,8 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
             _functionsSyncManager = functionsSyncManager;
             _hostNameProvider = hostNameProvider;
             _functionMetadataManager = functionMetadataManager;
+            _hostFunctionMetadataProvider = hostFunctionMetadataProvider;
+            _languageWorkerOptions = languageWorkerOptions;
         }
 
         public async Task<IEnumerable<FunctionMetadataResponse>> GetFunctionsMetadata(bool includeProxies)
@@ -144,11 +151,21 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
                 await FileUtility.WriteAsync(dataFilePath, functionMetadata.TestData);
             }
 
+            var metadata = (await _hostFunctionMetadataProvider.GetFunctionMetadataAsync(_languageWorkerOptions.CurrentValue.WorkerConfigs, configChanged))
+                .FirstOrDefault(metadata => Utility.FunctionNamesMatch(metadata.Name, name));
+
+            var success = false;
+            FunctionMetadataResponse functionMetadataResult = null;
+            if (functionMetadata != null)
+            {
+                string routePrefix = await GetRoutePrefix(hostOptions.RootScriptPath);
+                var baseUrl = $"{request.Scheme}://{request.Host}";
+                functionMetadataResult = await metadata.ToFunctionMetadataResponse(hostOptions, routePrefix, baseUrl);
+                success = true;
+            }
+
             // we need to sync triggers if config changed, or the files changed
             await _functionsSyncManager.TrySyncTriggersAsync();
-
-            // Setting force refresh to false as host restart causes a refersh already
-            (var success, var functionMetadataResult) = await TryGetFunction(name, request, false);
 
             return (success, configChanged, functionMetadataResult);
         }
--- a/src/WebJobs.Script.WebHost/Management/WebFunctionsManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/WebFunctionsManager.cs
@@ -151,23 +151,23 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
                 await FileUtility.WriteAsync(dataFilePath, functionMetadata.TestData);
             }
 
-            var metadata = (await _hostFunctionMetadataProvider.GetFunctionMetadataAsync(_languageWorkerOptions.CurrentValue.WorkerConfigs, configChanged))
+            var updatedFunctionMetadata = (await _hostFunctionMetadataProvider.GetFunctionMetadataAsync(_languageWorkerOptions.CurrentValue.WorkerConfigs, configChanged))
                 .FirstOrDefault(metadata => Utility.FunctionNamesMatch(metadata.Name, name));
 
-            var success = false;
+            bool getFunctionMetadataSuccessful = false;
             FunctionMetadataResponse functionMetadataResult = null;
-            if (functionMetadata != null)
+            if (updatedFunctionMetadata != null)
             {
                 string routePrefix = await GetRoutePrefix(hostOptions.RootScriptPath);
                 var baseUrl = $"{request.Scheme}://{request.Host}";
-                functionMetadataResult = await metadata.ToFunctionMetadataResponse(hostOptions, routePrefix, baseUrl);
-                success = true;
+                functionMetadataResult = await updatedFunctionMetadata.ToFunctionMetadataResponse(hostOptions, routePrefix, baseUrl);
+                getFunctionMetadataSuccessful = true;
             }
 
             // we need to sync triggers if config changed, or the files changed
             await _functionsSyncManager.TrySyncTriggersAsync();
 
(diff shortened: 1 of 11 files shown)


src/WebJobs.Script/Host/IHostFunctionMetadataProvider.cs
```diff
@@ -12,13 +12,13 @@ namespace Microsoft.Azure.WebJobs.Script
     /// <summary>
     /// Defines an interface for fetching function metadata from function.json files
     /// </summary>
-    internal interface IHostFunctionMetadataProvider
+    public interface IHostFunctionMetadataProvider
```
liliankasem, line 15: Why is this no longer internal?

src/WebJobs.Script/Host/FunctionMetadataProvider.cs
```diff
@@ -32,12 +32,12 @@ public FunctionMetadataProvider(ILogger<FunctionMetadataProvider> logger, IWorke
 
         public ImmutableDictionary<string, ImmutableArray<string>> FunctionErrors { get; private set; }
 
-        public async Task<ImmutableArray<FunctionMetadata>> GetFunctionMetadataAsync(IEnumerable<RpcWorkerConfig> workerConfigs, IEnvironment environment, bool forceRefresh = false)
+        public async Task<ImmutableArray<FunctionMetadata>> GetFunctionMetadataAsync(IEnumerable<RpcWorkerConfig> workerConfigs, bool forceRefresh = false)
```
liliankasem, line 35: Just double checking that this is just clean up/tidy work since environment is never actually used?

src/WebJobs.Script.WebHost/Management/WebFunctionsManager.cs
```diff
@@ -144,12 +151,22 @@ await functionMetadata
                 await FileUtility.WriteAsync(dataFilePath, functionMetadata.TestData);
             }
 
+            var metadata = (await _hostFunctionMetadataProvider.GetFunctionMetadataAsync(_languageWorkerOptions.CurrentValue.WorkerConfigs, configChanged))
+                .FirstOrDefault(metadata => Utility.FunctionNamesMatch(metadata.Name, name));
+
+            var success = false;
+            FunctionMetadataResponse functionMetadataResult = null;
+            if (functionMetadata != null)
```
liliankasem: Can functionMetadata be null here? We are checking properties of it in code above this, so it would be odd for it to be null at this point? Would it be better to add a null check on `functionMetadata` at the start of this method?

src/WebJobs.Script.WebHost/Management/WebFunctionsManager.cs
```diff
@@ -144,12 +151,22 @@ await functionMetadata
                 await FileUtility.WriteAsync(dataFilePath, functionMetadata.TestData);
             }
 
+            var metadata = (await _hostFunctionMetadataProvider.GetFunctionMetadataAsync(_languageWorkerOptions.CurrentValue.WorkerConfigs, configChanged))
+                .FirstOrDefault(metadata => Utility.FunctionNamesMatch(metadata.Name, name));
+
+            var success = false;
```
liliankasem: Can we rename this to something more descriptive, what is successful here?

src/WebJobs.Script/Host/FunctionMetadataProvider.cs
soninaren: That is correct.

src/WebJobs.Script.WebHost/Management/WebFunctionsManager.cs
soninaren: This was a bug, i was supposed to check the updated function metadata here
https://github.com/Azure/azure-functions-host/pull/10678/files#diff-203384510505c08751e66abe5167ac5534fe369ed79ffa4d424d9a4eb8847fe8R154

soninaren: I have fixed it.

src/WebJobs.Script.WebHost/Management/WebFunctionsManager.cs
soninaren: done

src/WebJobs.Script/Host/IHostFunctionMetadataProvider.cs
soninaren: Needed to make it public to use within a public class. WebfunctionManager was public already.

liliankasem (approved):

src/WebJobs.Script.WebHost/Management/WebFunctionsManager.cs
```diff
@@ -144,13 +151,23 @@ await functionMetadata
...
+
+            bool getFunctionMetadataSuccessful = false;
+            FunctionMetadataResponse functionMetadataResult = null;
+            if (updatedFunctionMetadata != null)
+            {
+                string routePrefix = await GetRoutePrefix(hostOptions.RootScriptPath);
+                var baseUrl = $"{request.Scheme}://{request.Host}";
+                functionMetadataResult = await updatedFunctionMetadata.ToFunctionMetadataResponse(hostOptions, routePrefix, baseUrl);
+                getFunctionMetadataSuccessful = true;
+            }
```
kshyju, line 166: Should we log any relevant state/information in the `else` case here? The API endpoint which calls this method is going to return 500 status code in that case and any useful information being logged will be helpful to debug.

soninaren: @kshyju i looked through the call stack to see if we can provide any useful information. I couldn't identify anything additional we can log here.

kshyju (approved):

src/WebJobs.Script.WebHost/Management/WebFunctionsManager.cs
```diff
@@ -144,13 +151,23 @@ await functionMetadata
                 await FileUtility.WriteAsync(dataFilePath, functionMetadata.TestData);
             }
 
+            var updatedFunctionMetadata = (await _hostFunctionMetadataProvider.GetFunctionMetadataAsync(_languageWorkerOptions.CurrentValue.WorkerConfigs, configChanged))
```
mathewc: On the other path [here](https://github.com/Azure/azure-functions-host/blob/ba3e9ef990a9a5ebacaf47bbc3a71ed5f661e8ea/src/WebJobs.Script/Host/FunctionMetadataProvider.cs#L60) where this provider is used, we set FunctionErrors. We'll be bypassing that here, but maybe that's OK? Assuming that if the host restarts after this call, it'll be reindexing anyways and set the errors if any at that time.

You'll also be bypassing the AllowList that is normally applied [here](https://github.com/Azure/azure-functions-host/blob/ba3e9ef990a9a5ebacaf47bbc3a71ed5f661e8ea/src/WebJobs.Script/Host/FunctionMetadataManager.cs#L97). Again might not be an issue, just calling out the differences.

mathewc: If we're going to make these changes to this method, I'd like to put a comment // TEMP HACK on this line pointing to a work item that describes the reason for this workaround, and references the work tracking the actual refactoring work (if we have that work item).

Basically ensuring that anyone looking at this code can understand why we're not using IFunctionMetadataManager as we should be

src/WebJobs.Script.WebHost/Management/WebFunctionsManager.cs
```diff
@@ -144,13 +151,23 @@ await functionMetadata
...
+                .FirstOrDefault(metadata => Utility.FunctionNamesMatch(metadata.Name, name));
+
+            bool getFunctionMetadataSuccessful = false;
```
mathewc: nit: can we just keep the bool named "success"?

src/WebJobs.Script.WebHost/Management/WebFunctionsManager.cs
```diff
@@ -144,13 +151,23 @@ await functionMetadata
...
+            FunctionMetadataResponse functionMetadataResult = null;
+            if (updatedFunctionMetadata != null)
+            {
+                string routePrefix = await GetRoutePrefix(hostOptions.RootScriptPath);
```
mathewc: I think the following 3 lines are a copy of the code [here](https://github.com/Azure/azure-functions-host/blob/ba3e9ef990a9a5ebacaf47bbc3a71ed5f661e8ea/src/WebJobs.Script.WebHost/Management/WebFunctionsManager.cs#L182) in this same file. Can you extract into a common method `GetFunctionMetadataResponseAsync(hostOptions, functionMetadata, request)`?

src/WebJobs.Script.WebHost/Management/WebFunctionsManager.cs
mathewc: nit: maybe just name the var "functionMetadata"

src/WebJobs.Script/Host/IHostFunctionMetadataProvider.cs
mathewc: Yes, the visibility on these classes wasn't managed very well. WebFunctionsManager should have been internal

src/WebJobs.Script.WebHost/Management/WebFunctionsManager.cs
soninaren: already have one with that name in the scope

mathewc (approved):

Issue #7864, opened 2021-11-17
FunctionAppLogs from Diagnostic settings have different format at "properties" field by OS/SKU


georgeOsdDev:
#### Investigative information

Please provide the following:

- Timestamp:
- Function App version: 4.0.1.16815
- Function App name: 2111040060000473ELASTIC-LINUX
- Function name(s) (as appropriate): N/A
- Invocation ID: N/A
- Region: JapanEast

#### Repro steps

Provide the steps required to reproduce the problem:

- Save FunctionAppLogs to storage account from Diagnostic settings.

#### Expected behavior

All OS and SKU should have same format.

#### Actual behavior

■Linux Elastic Premium
"properties" is quoted with "(double quote),  
key-values in stringifyed object are quoted with '(single quote)

```
{ "level": "Informational", "resourceId": "/SUBSCRIPTIONS/---/RESOURCEGROUPS/---/PROVIDERS/MICROSOFT.WEB/SITES/2111040060000473ELASTIC-LINUX", "operationName": "Microsoft.Web/sites/functions/log", "category": "FunctionAppLogs", "time": "11/15/2021 11:36:06", "properties": "{'appName':'2111040060000473elastic-linux','roleInstance':'4de66c2ce3805624b7e310460ca35c2789c27a85279b183e8c95aef70dd8d30a','message':'Starting JobHost','category':'Microsoft.Azure.WebJobs.Hosting.JobHostService','hostVersion':'4.0.1.16815','hostInstanceId':'3d7b63ad-2230-4902-9d52-195b26396d5f','level':'Information','levelId':2,'processId':19}", "EventStampType": "Stamp", "EventPrimaryStampName": "waws-prod-ty1-021", "EventStampName": "waws-prod-ty1-021e", "Host": "RD00155D7B6F5E", "EventIpAddress": "10.0.4.43"}
```

■Linux Consumption 
"properties" is not quoted, key-value in object are quoted with '(single quote)
Single quotation in json object is invalid

```
[...]

PR #10684, opened 2024-12-10
Adding support to emit azmon logs in iso time format for Dedicated sku based on a flag.

vivekjilla:

resolves timestamp issue under #7864

Validated with these changes in a custom image:
After adding Featureflag, time format is returned in iso format.
image

### Pull request checklist

    * Otherwise: Link to backporting PR https://github.com/Azure/azure-functions-host/pull/10810
    * [x] Otherwise: Created [internal work item](https://dev.azure.com/msazure/Antares/_workitems/edit/31324062) to track the docs update.

### Additional information

Additional PR information


src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
```diff
@@ -110,13 +110,28 @@ private void WriteEvent(string eventPayload)
...
 
         public static void LogUnhandledException(Exception e)
         {
             // Pipe the unhandled exception to stdout as part of docker logs.
             Console.WriteLine($"Unhandled exception on {DateTime.UtcNow}: {e?.ToString()}");
         }
+
+        private bool IsAzureMonitorTimeIsoFormatEnabled()
+        {
+            string enabledString = Environment.GetEnvironmentVariable(EnvironmentSettingNames.AzureMonitorTimeIsoFormatEnabled);
```
jviau: Please use `IOptions` pattern and not reading env variables directly.

jviau (changes requested):

src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
```diff
@@ -71,6 +71,7 @@ public static class EnvironmentSettingNames
         public const string HealthPingEnabled = "WEBSITE_FUNCTIONS_HEALTH_PING_ENABLED";
         public const string TestDataCapEnabled = "WEBSITE_FUNCTIONS_TESTDATA_CAP_ENABLED";
         public const string AzureMonitorCategories = "WEBSITE_FUNCTIONS_AZUREMONITOR_CATEGORIES";
+        public const string AzureMonitorTimeIsoFormatEnabled = "FUNCTIONS_AZUREMONITOR_TIME_ISOFORMAT_ENABLED";
```
jviau: Couple issues with the env name here:

1. This only impacts linux, but the name does not reflect that
2. The class that consumes this appears more generic then "azure monitor". Is it accurate to include that naming here, or should the name also be more generic? Or does this class directly correlate to AzureMonitor and no other telemetry technology?

pragnagopa: > This only impacts linux, but the name does not reflect that

good point! fix applies to Linux Dedicated and EP SKUs only -`WEBSITE_FUNCTIONS_AZUREMONITOR_LINUX_APPSERVICE_TIME_ISOFORMAT_ENABLED` ?

> The class that consumes this appears more generic then "azure monitor". Is it accurate to include that naming here, or should the name also be more generic? Or does this class directly correlate to AzureMonitor and no other telemetry technology?

This would apply to `AZUREMONITOR` only; @vivekjilla - do confirm by looking at Kusto logs if we see any difference in FunctionsLogs

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
pragnagopa: @vivekjilla - here is an example - https://github.com/Azure/azure-functions-host/blob/ba3e9ef990a9a5ebacaf47bbc3a71ed5f661e8ea/src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs#L89

vivekjilla: I see we have `IOptions<FunctionsHostingConfigOptions>` which seems to read the config from `FunctionsHostingConfig` section. Have few questions here:
1. Can customers set this value in FunctionsHostingConfig, and if so, how? or otherwise
2. I believe the app settings set by the customers can be accessed via environment variables. Do we have any example of IOptions which gets populated based on environment variables.

Asking because, I think we want this to be configurable by end users, as it can potentially break things.

Or if I'm completely off the direction suggested here, I can catch up with @jviau offline.

src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
vivekjilla: Yes, we're using this only in the LogAzureMonitorDiagnosticLogEvent method (in that class) which is used to generate azmon logs. Verified the kusto logs and FunctionsLogs are not impacted by this change.

Was just wondering if the long names can cause any size limits etc, but otherwise I can go with this as Pragna suggested: WEBSITE_FUNCTIONS_AZUREMONITOR_LINUX_APPSERVICE_TIME_ISOFORMAT_ENABLED

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
jviau: @vivekjilla, IOptions pattern reads from IConfiguration (and other custom ways). And IConfiguration is built from many sources - including environment variables. So, app settings (which are env vars at runtime) are available in IOptions and IConfiguration pattern.

As for which options object to use - you can see if there is a fitting one, or you can always add your own. So, you don't have to specifically use the any existing IOptions object if they are not fitting and instead create a new one.

That is the purpose of options pattern: use an options POCO which is specific to the _usage_ of the values and abstract away how the values are populated (in the options pipeline).

src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
jviau: Yeah, that name is quite long. Reading the issue more I think I am less worried about confusion: ISO format is already enabled on other skus/platforms, so having a customer set this to "true" for those other skus won't be concerning. I am not too concerned about a customer setting this to "false" and we still emit ISO.

This brings up another question: I assume we are only going with the config value here to avoid a breaking change? But ideally this would be the default format?

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
```diff
@@ -110,13 +110,28 @@ private void WriteEvent(string eventPayload)
 
         public override void LogAzureMonitorDiagnosticLogEvent(LogLevel level, string resourceId, string operationName, string category, string regionName, string properties)
         {
-            _writeEvent($"{ScriptConstants.LinuxAzureMonitorEventStreamName} {(int)ToEventLevel(level)},{resourceId},{operationName},{category},{regionName},{NormalizeString(properties.Replace("'", string.Empty))},{DateTime.UtcNow}");
+            var timeStr = IsAzureMonitorTimeIsoFormatEnabled() ? DateTime.UtcNow.ToString("s") : DateTime.UtcNow.ToString();
+            _writeEvent($"{ScriptConstants.LinuxAzureMonitorEventStreamName} {(int)ToEventLevel(level)},{resourceId},{operationName},{category},{regionName},{NormalizeString(properties.Replace("'", string.Empty))},{timeStr}");
```
jviau: Not sure if this is applicable, but you can supply formats directly in the interpolated string:

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/interpolated

src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
pragnagopa: that is correct. We want to avoid breaking change and hence making it opt-in

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
vivekjilla: Yeah, it seems little different from what we want (and we also need to configure format based on flag)

src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
vivekjilla: Thanks. Keeping the env name as is then

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
vivekjilla: Thanks @jviau . Updated the PR using IOptions.

src/WebJobs.Script.WebHost/Configuration/AzureMonitorOptionsSetup.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        public AzureMonitorOptionsSetup(IConfiguration configuration)
+        {
+            _configuration = configuration;
+        }
+
+        public void Configure(AzureMonitorOptions options)
+        {
+            var isAzureMonitorTimeIsoFormatEnabled = _configuration.GetValue<bool?>(EnvironmentSettingNames.AzureMonitorTimeIsoFormatEnabled);
+            options.IsAzureMonitorTimeIsoFormatEnabled = isAzureMonitorTimeIsoFormatEnabled ?? false;
```
vivekjilla: It doesn't seem to be working after moving to IOptions. 
Trying to debug further, maybe the environment variable is not being read correctly. 
Any thoughts/suggestion wrt initialization etc @jviau ?

vivekjilla: It worked after I moved the `services.ConfigureOptions<AzureMonitorLoggingOptionsSetup>();` to a different place now, looks like. Please let me know if the current changes look good.

src/WebJobs.Script.WebHost/Configuration/AzureMonitorLoggingOptionsSetup.cs
```diff
@@ -0,0 +1,37 @@
...
+using System;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Configuration
+{
+    public class AzureMonitorLoggingOptionsSetup : IConfigureOptions<AzureMonitorLoggingOptions>
+    {
+        private readonly IEnvironment _env;
+
+        public AzureMonitorLoggingOptionsSetup(IEnvironment env)
```
jviau: Use `IConfiguration` instead, it will include environment variables in it (and is a replacement for `IEnvironment`)

src/WebJobs.Script.WebHost/Configuration/AzureMonitorLoggingOptionsSetup.cs
```diff
@@ -0,0 +1,37 @@
...
+        }
+
+        public void Configure(AzureMonitorLoggingOptions options)
+        {
+            options.IsAzureMonitorTimeIsoFormatEnabled = IsAzureMonitorTimeIsoFormatEnabled();
+        }
+
+        private bool IsAzureMonitorTimeIsoFormatEnabled()
+        {
+            string enabledString = _env.GetEnvironmentVariable(EnvironmentSettingNames.AzureMonitorTimeIsoFormatEnabled);
```
jviau: With `IConfiguration` you can call `.GetValue<bool>("<value_name>")`

src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceEventGenerator.cs
```diff
@@ -110,7 +115,8 @@ private void WriteEvent(string eventPayload)
 
         public override void LogAzureMonitorDiagnosticLogEvent(LogLevel level, string resourceId, string operationName, string category, string regionName, string properties)
         {
-            _writeEvent($"{ScriptConstants.LinuxAzureMonitorEventStreamName} {(int)ToEventLevel(level)},{resourceId},{operationName},{category},{regionName},{NormalizeString(properties.Replace("'", string.Empty))},{DateTime.UtcNow}");
+            var timeStr = _azureMonitorLoggingOptions.Value.IsAzureMonitorTimeIsoFormatEnabled ? DateTime.UtcNow.ToString("s") : DateTime.UtcNow.ToString();
```
jviau: Something to consider: with options objects I like to make them smart. Instead of simply holding config value, you can also add methods to them to _act_ on the config value. In this case you could add a `AzureMonitorLoggingOptions.GetUtcDateTime()` and it will return the date time with the correct format based on the config values.

jviau (changes requested):
Please us `IConfiguration` instead of `IEnvironment`. But other than that, looks great! Thank you for switching to options pattern.

jacobcolvin-8451:
Does this solve all of the issues identified in #7864 or just the timestamp? I ask because this pr has "resolves #7864" and will close that issue, but #7864 also pointed out other inconsistencies in JSON encoding besides just timestamps.

vivekjilla:
> Does this solve all of the issues identified in #7864 or just the timestamp? I ask because this pr has "resolves #7864" and will close that issue, but #7864 also pointed out other inconsistencies in JSON encoding besides just timestamps.

You're right. It's only fixing the timestamp part. Updated the description.

jviau (approved):

pragnagopa:
Thanks @vivekjilla  @jviau 

@vivekjilla - please update the release notes to call out introduction of new app setting `"FUNCTIONS_AZUREMONITOR_TIME_ISOFORMAT_ENABLED`

No doc updates until the feature roll out; Please track updating docs in internal WI

Do send backport PR to inproc branch -- @jviau  can you please share the right branch for sending cherrypick PR for inproc?

vivekjilla:
> Thanks @vivekjilla @jviau
> 
> @vivekjilla - please update the release notes to call out introduction of new app setting `"FUNCTIONS_AZUREMONITOR_TIME_ISOFORMAT_ENABLED`
> 
> No doc updates until the feature roll out; Please track updating docs in internal WI
> 
> Do send backport PR to inproc branch -- @jviau can you please share the right branch for sending cherrypick PR for inproc?

Updated the description with release notes, doc work item link, and backport PR to in-proc. Thanks @pragnagopa

src/WebJobs.Script.WebHost/Configuration/AzureMonitorLoggingOptionsSetup.cs
```diff
@@ -0,0 +1,24 @@
...
+        private readonly IConfiguration _configuration;
+
+        public AzureMonitorLoggingOptionsSetup(IConfiguration configuration)
+        {
+            _configuration = configuration;
+        }
+
+        public void Configure(AzureMonitorLoggingOptions options)
+        {
+            var isAzureMonitorTimeIsoFormatEnabled = _configuration.GetValue<bool?>(EnvironmentSettingNames.AzureMonitorTimeIsoFormatEnabled);
```
fabiocav: Assuming this will be customer defined (can you please confirm?), and if so, have you considered using a feature flag instead? (supported by https://github.com/Azure/azure-functions-host/blob/f03b0ec4b0e73b84f3505a9ccc41c1bd8054a6fd/src/WebJobs.Script/Config/FeatureFlags.cs#L13). 

If possible, let's avoid introducing a new environment variable for this.

vivekjilla: Thanks @fabiocav , updated.

src/WebJobs.Script.WebHost/Configuration/AzureMonitorLoggingOptionsSetup.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Azure.WebJobs.Script.Config;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Configuration
+{
+    public class AzureMonitorLoggingOptionsSetup : IConfigureOptions<AzureMonitorLoggingOptions>
```
kshyju: Should this be `public`? Can it be `internal` ?

src/WebJobs.Script.WebHost/Configuration/AzureMonitorLoggingOptions.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Configuration
+{
+    public class AzureMonitorLoggingOptions
```
kshyju:
```suggestion
    internal sealed class AzureMonitorLoggingOptions
```

src/WebJobs.Script.WebHost/Configuration/AzureMonitorLoggingOptionsSetup.cs
kshyju: Mark the class as `sealed`

src/WebJobs.Script.WebHost/Configuration/AzureMonitorLoggingOptions.cs
```diff
@@ -0,0 +1,17 @@
...
+    {
+        public bool IsAzureMonitorTimeIsoFormatEnabled { get; set; }
+
+        public string GetUtcDateTime()
+        {
+           return IsAzureMonitorTimeIsoFormatEnabled ? DateTime.UtcNow.ToString("s") : DateTime.UtcNow.ToString();
```
kshyju, line 14: This would not produce milli seconds in the time stamp. Is that okay for this use case?

src/WebJobs.Script.WebHost/Configuration/AzureMonitorLoggingOptionsSetup.cs
```diff
@@ -0,0 +1,16 @@
...
+    {
+        public void Configure(AzureMonitorLoggingOptions options)
+        {
+            options.IsAzureMonitorTimeIsoFormatEnabled = FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagEnableAzureMonitorTimeIsoFormat);
```
kshyju, line 13: The feature flag value will be different after specialization. Correct? (If customer is setting this as app setting). In that case, we want to refresh this option property after specialization. I don't see a test covering specialization use case in this PR, so I am unsure this is being handled. can you confirm?

src/WebJobs.Script.WebHost/Configuration/AzureMonitorLoggingOptions.cs
vivekjilla: We were not sending milliseconds previously also, might be getting truncated somewhere in the logging pipeline. Added screenshots in the description.

src/WebJobs.Script.WebHost/Configuration/AzureMonitorLoggingOptionsSetup.cs
vivekjilla: This support is mainly for linux dedicated/EP. For consumption and Windows SKUs, the time is already printed in the iso format. And this change won't do anything. So should be fine I guess.

kshyju (approved):

PR #10688, opened 2024-12-10
Update http proxy to handle client disconnect scenario

### LINKED ISSUE #10600: DefaultHttpProxyService does not take into account cancellation (ForwarderError.RequestCanceled)
Proxy not forwarding cancellation token correctly to function app

Canada Central example @pmerineau 
first reported https://github.com/Azure/azure-functions-dotnet-worker/issues/2417#issuecomment-2455793340

```
Sending InvocationCancel request for invocation: '0577cf46-de3f-47ba-9751-0750cc28ba18'
And your function invocation is failing, likely due to this proxy that is not forwarding the cancellation request to your app properly:
Exception while executing function: Functions.<YourFunctionName> ---> System.InvalidOperationException : Failed to proxy request with ForwarderError: RequestCanceled. System.IO.IOException : Unable to read data from the transport connection: The I/O operation has been aborted because of either a thread exit or an application request
```

North Europe example @stuartf-aida
first reported https://github.com/Azure/azure-functions-dotnet-worker/issues/2417#issuecomment-2455857336

```
Sending InvocationCancel request for invocation: 'a11b8045-59cf-4db2-90eb-356d561e2fb2'
Exception while executing function: Functions.<YourFunctionName> ---> System.InvalidOperationException : Failed to proxy request with ForwarderError: RequestCanceled. System.IO.IOException : Unable to read data from the transport connection: The I/O operation has been aborted because of either a thread exit or an application request
```

As mentioned by @liliankasem 
>The error is actually happening on the Functions Host side in the [RetryProxyHandler](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.Grpc/Server/RetryProxyHandler.cs) so I would suggest opening the issue there and I will continue that investigation there as well.

I hope this is what you need, Lilian!

liliankasem:

resolves #10600

### Pull request checklist

    * [x] Otherwise: Link to backporting PR #10731

### Additional information

- Introduce HttpForwardingException as this makes more sense than an invalid operation when a forwarding error occurs
- Update the proxy service to check for client disconnect and log a helpful message instead of throwing an exception
- Introduce `HandleCancellationMiddleware` to return `Status499ClientClosedRequest` on client disconnect
- Update `RetryProxyHandler` to stop retries when the request is cancelled

### Testing

See comments below for test scenarios completed. Behaviour change from this PR have been compared to a) what is in dev today and b) what aspnet does


jviau:
For the scenarios you tested, can you compare how this would behave in a regular AspNetCore app? I think we should align with that behavior as best we can. If client disconnects, is an exception thrown? Is the request considered failed or successful?

src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
```diff
@@ -62,7 +65,15 @@ public async Task EnsureSuccessfulForwardingAsync(ScriptInvocationContext contex
                     forwarderException = request.HttpContext.GetForwarderErrorFeature()?.Exception;
                 }
 
-                throw new InvalidOperationException($"Failed to proxy request with ForwarderError: {httpProxyTaskResult}", forwarderException);
+                if (request.HttpContext.RequestAborted.IsCancellationRequested)
```
jviau, line 70: We may want to log the forwarding error code for all failure cases (if it isn't logged already)

liliankasem:
> For the scenarios you tested, can you compare how this would behave in a regular AspNetCore app? I think we should align with that behavior as best we can. If client disconnects, is an exception thrown? Is the request considered failed or successful?

Here's what I found using a plain aspnet app:

**RequestAborted**
- **Not utilizing the CT**: work gets to complete without interuption, no errors or exceptions; no response sent as client does not exist
- **Utilizes CT, but not catching exception**: work stops, no exception is seen (handled by the framework maybe?); no response sent as client does not exist
- **Utilizes CT & catches exceptions**: OperationCanceledException is thrown and is handled by user code, no errors or exceptions; no response sent as client does not exist

**CancellationTokenSource cancelled**
- **Not utilizing the CT**: work gets to complete without interuption, no errors or exceptions; successful completion with user defined status code
- **Utilizes CT, but not catching exception**: TaskCancelled exception is thrown; fails with 500 internal server error
- **Utilizes CT & catches exceptions**: OperationCanceledException is thrown and handled by user code, no errors or exceptions; successful completion with user defined status code

Here's the app I used, or do you think I need to try and replicate what we're doing with the YARP as well?

```csharp
var builder = WebApplication.CreateBuilder(args);
var app = builder.Build();

app.MapGet("/process", async (HttpContext context) =>
{
    var cancellationToken = context.RequestAborted; // Client disconnect
    // var cancellationToken = new CancellationTokenSource(TimeSpan.FromSeconds(5)).Token; // Cancel after 5 seconds

    try
    {
        // Simulate some work that can be cancelled
        Console.WriteLine("Starting work...");
        for (int i = 0; i < 10; i++)
        {
            // Simulate work
            await Task.Delay(1000, cancellationToken);
            Console.WriteLine($"Work iteration {i + 1} completed.");
        }
        Console.WriteLine("Work completed.");

        context.Response.StatusCode = StatusCodes.Status200OK;
        await context.Response.WriteAsync("Processing completed successfully.");
    }
    catch (OperationCanceledException)
    {
        Console.WriteLine("Request was cancelled.");
        context.Response.StatusCode = StatusCodes.Status499ClientClosedRequest; // Status code for client closed request
        await context.Response.WriteAsync("Request was cancelled.");
    }
});

app.Run();
```

jviau:
Don't need to test with YARP directly, but I would like us to compare with your found results for AspNetCore with what happens in the functions YARP scenario today.

Also: `StatusCodes.Status499ClientClosedRequest` - good to know there is a specific HTTP code for that! Do you know if we set this status code in functions host for client disconnect scenarios?

liliankasem:
> Don't need to test with YARP directly, but I would like us to compare with your found results for AspNetCore with what happens in the functions YARP scenario today.
> 
> Also: `StatusCodes.Status499ClientClosedRequest` - good to know there is a specific HTTP code for that! Do you know if we set this status code in functions host for client disconnect scenarios?

@jviau `Status499ClientClosedRequest` is technically an unofficial status code, and no, we don't use it anywhere in the host/

In comparison, here is the result to the exact same code as above but in a function app with what we have today in dev:

**RequestAborted**
- **Not utilizing the CT**: work gets to complete without interuption, DefaultHttpProxyService retry handler throws exception, function invocation fails with forwarder error from host, status code 500, no response to client
  - Exception the moment we cancel: `Unsupported exception type in RetryProxyHandler. Request will not be retried. Exception: System.Threading.Tasks.TaskCanceledException: The operation was canceled.`
  - Exception after the function completes: `Microsoft.Azure.WebJobs.Host.FunctionInvocationException: Exception while executing function: Functions.MyHttpTriggerNoCT` - `Failed to proxy request with ForwarderError: RequestCanceled`
- **Utilizes CT, but not catching exception**: work stops, DefaultHttpProxyService retry handler throws exception, function invocation fails with exception, status code 500, no response to client
  - Exception the moment we cancel: `Unsupported exception type in RetryProxyHandler. Request will not be retried. Exception: System.Threading.Tasks.TaskCanceledException: The operation was canceled.`
  - User code gets unhandled exception: `Microsoft.Azure.WebJobs.Script.Workers.Rpc.RpcException: Result: Function 'MyHttpTriggerCTNoCatch', Invocation id 'e33ab37a-069f-4a89-b64c-2e7152946142': An exception was thrown by the invocation.`
  - Exception after the function completes: `Microsoft.Azure.WebJobs.Host.FunctionInvocationException: Exception while executing function: Functions.MyHttpTriggerNoCT` - `Failed to proxy request with ForwarderError: RequestCanceled`
- **Utilizes CT & catches exceptions**: OperationCanceledException is thrown and is handled by user code, DefaultHttpProxyService retry handler throws exception, function invocation fails with forwarder error from host, status code 500, no response to client
  - Exception the moment we cancel: `Unsupported exception type in RetryProxyHandler. Request will not be retried. Exception: System.Threading.Tasks.TaskCanceledException: The operation was canceled.`
  - Exception after the function completes: `Microsoft.Azure.WebJobs.Host.FunctionInvocationException: Exception while executing function: Functions.MyHttpTriggerNoCT` - `Failed to proxy request with ForwarderError: RequestCanceled`

**CancellationTokenSource cancelled**
- **Not utilizing the CT**: work gets to complete without interuption, no errors or exceptions, function invocation successful with user defined status code
- **Utilizes CT, but not catching exception**: work stops, TaskCancelled exception is thrown, function invocation fails with unhandled cancellation exception, status code 500 returned to client
- **Utilizes CT & catches exceptions**: OperationCanceledException is thrown and handled by user code, no errors or exceptions; function invocation successful with user defined status code

### And with my changes in this PR:

**RequestAborted**
- **Not utilizing the CT**: work gets to complete without interruption, function invocation successful, status code 499 (our http logs/trace), no response to client
  - New Log: `The client disconnected while the function was processing the request.`
- **Utilizes CT, but not catching exception**: work stops, TaskCancelled exception is thrown, function invocation fails with unhandled exception, status code 499, no response to client
  - User code gets unhandled exception: `Microsoft.Azure.WebJobs.Script.Workers.Rpc.RpcException: Result: Function 'MyHttpTriggerCTNoCatch', Invocation id 'e33ab37a-069f-4a89-b64c-2e7152946142': An exception was thrown by the invocation.`
  - New Log: `The client disconnected while the function was processing the request.`
- **Utilizes CT & catches exceptions**: OperationCanceledException is thrown and is handled by user code, function invocation successful, status code 499, no response to client
  - New Log: `The client disconnected while the function was processing the request.`


**CancellationTokenSource cancelled**
- **Not utilizing the CT**: work gets to complete without interruption, no errors or exceptions, function invocation successful with user defined status code
- **Utilizes CT, but not catching exception**: work stops, TaskCancelled exception is thrown, function invocation fails with unhandled exception, status code 500 returned to client
- **Utilizes CT & catches exceptions**: OperationCanceledException is thrown and handled by user code, no errors or exceptions; function invocation successful with user defined status code

src/WebJobs.Script.WebHost/Middleware/SystemTraceMiddleware.cs
```diff
@@ -36,6 +37,11 @@ public async Task Invoke(HttpContext context)
 
             await _next.Invoke(context);
 
+            if (context.RequestAborted.IsCancellationRequested)
```
jviau:
```suggestion
            if (context.RequestAborted.IsCancellationRequested && !context.Response.HasStarted)
```

Got this from the PR which added this handling to AspNetCore: https://github.com/dotnet/aspnetcore/pull/46330

May be useful to skim this PR to see if there is anything else we should be handling/doing

jviau: Should we also be catching exceptions incase some later middleware already threw for request aborted?

liliankasem: I wonder if the changes here might be greater and it's worth its own PR?

src/WebJobs.Script.WebHost/Middleware/SystemTraceMiddleware.cs
```diff
@@ -34,7 +35,32 @@ public async Task Invoke(HttpContext context)
...
+            {
+                await _next.Invoke(context);
+
+                if (context.RequestAborted.IsCancellationRequested && !context.Response.HasStarted)
+                {
+                    _logger.RequestAborted(requestId);
+                    context.Response.StatusCode = StatusCodes.Status499ClientClosedRequest;
+                }
+            }
+            catch (Exception ex)
```
jviau: Can you use a `when` statement?

src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
```diff
@@ -188,6 +188,12 @@ public static class ScriptHostServiceLoggerExtension
                 new EventId(530, nameof(LogHostInitializationSettings)),
                 "{hostInitializationSettings}");
 
+        private static readonly Action<ILogger, string, Exception> _requestAborted =
+            LoggerMessage.Define<string>(
+                LogLevel.Debug,
+                new EventId(531, nameof(RequestAborted)),
+                "The request was aborted by the client (requestId: '{mS_ActivityId}').");
```
jviau, line 195: Is the capitalization of `mS` intentional here?

jviau (approved):

src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
liliankasem: Yeah I was copying what `ExecutedHttpRequest` already had, but I don't know if this was intentional by the original author. I change my reference to it, but I think for `ExecutedHttpRequest`, something in `Properties.Resources.ExecutingHttpRequest)` is expecting that format

https://github.com/Azure/azure-functions-host/blob/c84ce0dd2033e016a9445ec4515b14374ba7c20f/src/WebJobs.Script.WebHost/Properties/Resources.resx#L126-L139

src/WebJobs.Script.WebHost/Middleware/SystemTraceMiddleware.cs
liliankasem: Yup

src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
liliankasem: But I think we either change them all, or none. I'd rather not have the inconsistency in naming

jviau: If it is an existing way we case it, that works for me.

src/WebJobs.Script.Grpc/Server/RetryProxyHandler.cs
```diff
@@ -58,6 +63,9 @@ protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage
                 }
             }
 
+            // If the loop exits without returning, propagate the cancellation
+            cancellationToken.ThrowIfCancellationRequested();
```
fabiocav: With the original implementation, execution would never reach here. Relying on the `break` above changes that and reevaluates the token cancellation state (which you're already doing when catching). Can you please replace the `break` and throw an `OperationCanceledException` directly from there instead?

src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
fabiocav: Interesting. I was going to ask a similar question.

src/WebJobs.Script.WebHost/Middleware/SystemTraceMiddleware.cs
```diff
@@ -34,7 +35,25 @@ public async Task Invoke(HttpContext context)
             string userAgent = context.Request.GetHeaderValueOrDefault("User-Agent");
             _logger.ExecutingHttpRequest(requestId, context.Request.Method, userAgent, context.Request.Path);
 
-            await _next.Invoke(context);
+            try
```
fabiocav: This logic is overloading the `SystemTraceMiddleware` (its responsibility is to handle request tracing). Can you please move this to a new piece of middleware and have it added to the pipeline right after this one?

liliankasem: That makes sense, I can do that!

fabiocav (approved):

PR #10690, opened 2024-12-11
Adding support to set `WorkingDirectory` while running a worker

ghost:
Logic app needs this supper because PAL worker we are running as part of our PAL project, looks for `.spf` libraries in either in `CurrentDirectory` or `CurrentDirectory/lib`.
So we need a support in functions app to set the working directory of a worker process. Currently it defaults to function app's location.

### Pull request checklist

### Additional information

We have tested the changes using a worker config like below to run a pal process in Linux. Test branch `la-hybrid/v4.636.1`
```json
{
  "description": {
    "language": "dotnet-isolated",
    "extensions": [
      ".dll"
    ],
    "defaultExecutablePath": "{workerDirectoryPath}lapalfe",
    "executableWorkingDirectory": "{workerDirectoryPath}",
    "defaultWorkerPath": "lapalfe"
  },
  "processOptions": {
    "initializationTimeout": "00:00:30",
    "environmentReloadTimeout": "00:00:30"
  }
}

```

### DIFF
--- a/src/WebJobs.Script/Workers/ProcessManagement/WorkerDescription.cs
+++ b/src/WebJobs.Script/Workers/ProcessManagement/WorkerDescription.cs
@@ -19,6 +19,11 @@ namespace Microsoft.Azure.WebJobs.Script.Workers
         public string DefaultExecutablePath { get; set; }
 
         /// <summary>
+        /// Gets or sets the working direfory for executing worker.
+        /// </summary>
+        public string ExecutableWorkingDirectory { get; set; }
+
+        /// <summary>
         /// Gets or sets the default path to the worker.
         /// </summary>
         public string DefaultWorkerPath { get; set; }
--- a/src/WebJobs.Script/Workers/ProcessManagement/WorkerDescription.cs
+++ b/src/WebJobs.Script/Workers/ProcessManagement/WorkerDescription.cs
@@ -19,7 +19,7 @@ namespace Microsoft.Azure.WebJobs.Script.Workers
         public string DefaultExecutablePath { get; set; }
 
         /// <summary>
-        /// Gets or sets the working direfory for executing worker.
+        /// Gets or sets the working directory for executing worker.
         /// </summary>
         public string ExecutableWorkingDirectory { get; set; }
 
--- a/src/WebJobs.Script/Workers/ProcessManagement/WorkerDescription.cs
+++ b/src/WebJobs.Script/Workers/ProcessManagement/WorkerDescription.cs
@@ -20,6 +20,8 @@ namespace Microsoft.Azure.WebJobs.Script.Workers
 
         /// <summary>
         /// Gets or sets the working directory for worker executable. Defaults to the Function App root directory.
+        /// If set, it will be used as the working directory for the worker executable.
+        /// If not set, the Function App root directory will be used as the working directory for the worker executable.
         /// </summary>
         public string ExecutableWorkingDirectory { get; set; }
 
--- a/src/WebJobs.Script/Workers/ProcessManagement/WorkerDescription.cs
+++ b/src/WebJobs.Script/Workers/ProcessManagement/WorkerDescription.cs
@@ -19,7 +19,7 @@ namespace Microsoft.Azure.WebJobs.Script.Workers
         public string DefaultExecutablePath { get; set; }
 
         /// <summary>
-        /// Gets or sets the working directory for executing worker.
+        /// Gets or sets the working directory for worker executable. Defaults to the Function App root directory.
         /// </summary>
         public string ExecutableWorkingDirectory { get; set; }
 
--- a/src/WebJobs.Script/Workers/Rpc/RpcWorkerDescription.cs
+++ b/src/WebJobs.Script/Workers/Rpc/RpcWorkerDescription.cs
@@ -251,5 +251,10 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
 
             return match.Value;
         }
+
+        internal void FormatWorkingDirectoryIfNeeded()
+        {
+            ExecutableWorkingDirectory = ExecutableWorkingDirectory?.Replace(RpcWorkerConstants.WorkerDirectoryPath, WorkerDirectory);
+        }
     }
 }
--- a/src/WebJobs.Script/Workers/Rpc/RpcWorkerProcess.cs
+++ b/src/WebJobs.Script/Workers/Rpc/RpcWorkerProcess.cs
@@ -20,6 +20,7 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
         private readonly string _workerId;
         private readonly Uri _serverUri;
         private readonly string _scriptRootPath;
+        private readonly string _workerExecutableWorkingDirectory;
         private readonly WorkerProcessArguments _workerProcessArguments;
         private readonly string _workerDirectory;
         private readonly IOptions<FunctionsHostingConfigOptions> _hostingConfigOptions;
@@ -50,7 +51,8 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
             _workerProcessLogger = workerProcessLogger;
             _workerId = workerId;
             _serverUri = serverUri;
-            _scriptRootPath = rpcWorkerConfig.Description.ExecutableWorkingDirectory ?? rootScriptPath;
+            _scriptRootPath = rootScriptPath;
+            _workerExecutableWorkingDirectory = rpcWorkerConfig.Description.ExecutableWorkingDirectory;
             _workerProcessArguments = rpcWorkerConfig.Arguments;
             _workerDirectory = rpcWorkerConfig.Description.WorkerDirectory;
             _hostingConfigOptions = hostingConfigOptions;
@@ -59,7 +61,7 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
 
         internal override Process CreateWorkerProcess()
         {
-            var workerContext = new RpcWorkerContext(Guid.NewGuid().ToString(), RpcWorkerConstants.DefaultMaxMessageLengthBytes, _workerId, _workerProcessArguments, _scriptRootPath, _serverUri);
+            var workerContext = new RpcWorkerContext(Guid.NewGuid().ToString(), RpcWorkerConstants.DefaultMaxMessageLengthBytes, _workerId, _workerProcessArguments, _workerExecutableWorkingDirectory ?? _scriptRootPath, _serverUri);
             workerContext.EnvironmentVariables.Add(WorkerConstants.FunctionsWorkerDirectorySettingName, _workerDirectory);
             workerContext.EnvironmentVariables.Add(WorkerConstants.FunctionsApplicationDirectorySettingName, _scriptRootPath);
             workerContext.EnvironmentVariables.Add(RpcWorkerConstants.FunctionWorkerRuntimeVersionSettingName, _environment.GetEnvironmentVariable(RpcWorkerConstants.FunctionWorkerRuntimeVersionSettingName));
--- a/src/WebJobs.Script/Workers/Rpc/RpcWorkerProcess.cs
+++ b/src/WebJobs.Script/Workers/Rpc/RpcWorkerProcess.cs
@@ -24,6 +24,7 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
         private readonly string _workerDirectory;
         private readonly IOptions<FunctionsHostingConfigOptions> _hostingConfigOptions;
         private readonly IEnvironment _environment;
+        private readonly RpcWorkerDescription _rpcWorkerDescription;
 
         internal RpcWorkerProcess(string runtime,
                                        string workerId,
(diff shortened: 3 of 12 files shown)


jviau:
> So we need a support in functions app to set the working directory of a worker process. Currently it defaults to function app's location.

Have you tried changing the working directory of your worker app once during startup from within your worker app? https://learn.microsoft.com/en-us/dotnet/api/system.io.directory.setcurrentdirectory?view=net-9.0

ghost:
> > So we need a support in functions app to set the working directory of a worker process. Currently it defaults to function app's location.
> 
> Have you tried changing the working directory of your worker app once during startup from within your worker app? https://learn.microsoft.com/en-us/dotnet/api/system.io.directory.setcurrentdirectory?view=net-9.0

We are trying to run PAL executable (C++ project coming from a Microsoft Reasearch to run  `.exe` on `Linux`). Those executable link libraries under `cwd` or `cwd/lib`. So the app will not start until it finds and links all dependencies

kshyju:
To clarify, is the goal here to start a Linux executable from the host? If so, where in the environment is it located? Is this a .NET executable, and what is its specific purpose once it starts? For example, how does it initiate the dotnet isolated payload?  

Could we leverage the `AzureWebJobsScriptRoot` environment variable to point to this executable?

ghost:
> To clarify, is the goal here to start a Linux executable from the host? If so, where in the environment is it located? Is this a .NET executable, and what is its specific purpose once it starts? For example, how does it initiate the dotnet isolated payload?
> 
> Could we leverage the `AzureWebJobsScriptRoot` environment variable to point to this executable?

Yes we want to run a linux worker on linux host. The worker will start a windows sandbox and run a .NET Framework exe.
Currently all this info is coming from `worker.config.json`. Like below:
```json
{
  "description": {
    "language": "dotnet-isolated",
    "extensions": [
      ".dll"
    ],
    "defaultExecutablePath": "{workerDirectoryPath}lapalfe",
    "executableWorkingDirectory": "{workerDirectoryPath}",
    "defaultWorkerPath": "lapalfe"
  },
  "processOptions": {
    "initializationTimeout": "00:00:30",
    "environmentReloadTimeout": "00:00:30"
  }
}
```

The thing is, PAL, which is published by MS Research expects the dependent libraries to be part of `pwd/lib` or `pwd`. Thus we need a mechanism to set the working directory of an executable.

Everything else, as suggested in last meeting, can be done by the configs, and we have removed the flag we added before `palEmulated` as we don't want SKU specific code in the repo.

Regarding `AzureWebjobsScriptRoot`: We support multi language workers. So, we run more than one worker in single host. Like a single host can run Java, Js, and other languages as well. Thus, we want to avoid setting it for all the workers through this env var.

jviau:
@siddharth-ms the worker.config.json allows for setting of executable working directing. Is this not sufficient for your scenario? If not, can you explain why?

ghost:
> @siddharth-ms the worker.config.json allows for setting of executable working directing. Is this not sufficient for your scenario? If not, can you explain why?

Ohh, does it? How can I set the working directory?

jviau:
> > @siddharth-ms the worker.config.json allows for setting of executable working directing. Is this not sufficient for your scenario? If not, can you explain why?
> 
> Ohh, does it? How can I set the working directory?

Ah I misread an existing property - nevermind.

src/WebJobs.Script/Workers/ProcessManagement/WorkerDescription.cs
```diff
@@ -18,6 +18,11 @@ public abstract class WorkerDescription
         /// </summary>
         public string DefaultExecutablePath { get; set; }
 
+        /// <summary>
+        /// Gets or sets the working direfory for executing worker.
```
mathewc: spelling

src/WebJobs.Script/Workers/Rpc/RpcWorkerProcess.cs
```diff
@@ -59,7 +61,7 @@ internal RpcWorkerProcess(string runtime,
 
         internal override Process CreateWorkerProcess()
         {
-            var workerContext = new RpcWorkerContext(Guid.NewGuid().ToString(), RpcWorkerConstants.DefaultMaxMessageLengthBytes, _workerId, _workerProcessArguments, _scriptRootPath, _serverUri);
+            var workerContext = new RpcWorkerContext(Guid.NewGuid().ToString(), RpcWorkerConstants.DefaultMaxMessageLengthBytes, _workerId, _workerProcessArguments, _workerExecutableWorkingDirectory ?? _scriptRootPath, _serverUri);
```
mathewc: Consider doing this defaulting in the ctor at the time when you set _workerExecutableWorkingDirectory

test/WebJobs.Script.Tests/Workers/Rpc/RpcWorkerProcessTests.cs
```diff
@@ -62,6 +63,23 @@ public RpcWorkerProcessTests()
                 testEnv,
                 scriptApplicationHostOptions,
                 new LoggerFactory());
+
+            _rpcWorkerProcessWithExecutablePath = new RpcWorkerProcess("node",
```
mathewc: Any reason why this needs to be a field? Seems like you can just define this locally in your CreateWorkerProcess_AddsExecutableWorkingDirectory test below, since it's not used anywhere else

test/WebJobs.Script.Tests/Workers/Rpc/RpcWorkerProcessTests.cs
```diff
@@ -75,7 +93,7 @@ public void Constructor_RegistersHostStartedEvent()
                 return eventType == typeof(HostStartEvent);
             };
 
-            _eventManager.Verify(_ => _.Subscribe(It.Is<IObserver<ScriptEvent>>(p => validate(p))), Times.Once());
+            _eventManager.Verify(_ => _.Subscribe(It.Is<IObserver<ScriptEvent>>(p => validate(p))), Times.AtLeastOnce());
```
mathewc: Why this change?

test/WebJobs.Script.Tests/Workers/Rpc/RpcWorkerProcessTests.cs
ghost: I allows re-using other mocks defined in scope

test/WebJobs.Script.Tests/Workers/Rpc/RpcWorkerProcessTests.cs
ghost: Subscribe is called twice due to addition of `_rpcWorkerProcessWithExecutablePath ` in ctor

src/WebJobs.Script/Workers/ProcessManagement/WorkerDescription.cs
```diff
@@ -18,6 +18,11 @@ public abstract class WorkerDescription
         /// </summary>
         public string DefaultExecutablePath { get; set; }
 
+        /// <summary>
+        /// Gets or sets the working directory for executing worker.
```
fabiocav: nit: `/// Gets or sets the working directory for worker executable.

test/WebJobs.Script.Tests/Workers/Rpc/RpcWorkerProcessTests.cs
fabiocav: This invalidates the test as it wouldn't catch a regression that ends up resulting multiple events. Can you update this to use `Exactly` instead?

Refactoring this test to avoid the code duplication with the new instance of the `RpcWorkerProcess` would be preferred over this, though.

src/WebJobs.Script/Workers/ProcessManagement/WorkerDescription.cs
mathewc: So there are 3 directories that can be configured for the worker: ExecutableWorkingDirectory, DefaultWorkerPath and WorkerDirectory. How do this new property and WorkerDirectory differ?

mathewc: Also on your help text for this property, you might add "Defaults to the Function App root directory".

test/WebJobs.Script.Tests/Workers/Rpc/RpcWorkerProcessTests.cs
mathewc: Unless you're reusing it, just define it locally in the test. If/when we ever need to reuse in other tests, it can be hoisted at that time.

src/WebJobs.Script/Workers/ProcessManagement/WorkerDescription.cs
ghost: `WorkerDirectory` is the base directory of worker.

But the executables that are used to run the worker code can be anywhere in the machine, and `ExecutableWorkingDirectory` is the working directory of the worker process executing the worker, it might not be same for some of PAL scenarios, where `ExecutableWorkingDirectory` should be set to PAL executable binaries

`WorkerDirectory` doesn't update the working directory of worker process

test/WebJobs.Script.Tests/Workers/Rpc/RpcWorkerProcessTests.cs
```diff
@@ -18,39 +20,34 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.Workers.Rpc
 {
     public class RpcWorkerProcessTests
     {
-        private readonly Mock<IWorkerProcessFactory> _workerProcessFactory;
-        private RpcWorkerProcess _rpcWorkerProcess;
-        private Mock<IScriptEventManager> _eventManager;
-        private Mock<IHostProcessMonitor> _hostProcessMonitorMock;
-        private TestLogger _logger = new TestLogger("test");
-        private IOptions<FunctionsHostingConfigOptions> _functionsHostingConfigOptions;
+        private readonly Mock<IWorkerProcessFactory> _workerProcessFactory = new Mock<IWorkerProcessFactory>();
```
kshyju, line 23: Are these changes unrelated / needed? Can we fix the inconsistencies in private variable names (some has _ prefix and some does not)

src/WebJobs.Script/Workers/ProcessManagement/WorkerDescription.cs
```diff
@@ -18,6 +18,11 @@ public abstract class WorkerDescription
         /// </summary>
         public string DefaultExecutablePath { get; set; }
 
+        /// <summary>
+        /// Gets or sets the working directory for worker executable. Defaults to the Function App root directory.
+        /// </summary>
+        public string ExecutableWorkingDirectory { get; set; }
```
kshyju, line 26: To help future readers, can we add some comment about how this will be different from `WorkerDirectory` property which also exist in this class?

src/WebJobs.Script/Workers/Rpc/RpcWorkerDescription.cs
```diff
@@ -251,5 +251,10 @@ private string GetSanitizedRuntimeVersion(string version)
 
             return match.Value;
         }
+
+        internal void FormatWorkingDirectoryIfNeeded()
+        {
+            ExecutableWorkingDirectory = ExecutableWorkingDirectory?.Replace(RpcWorkerConstants.WorkerDirectoryPath, WorkerDirectory);
```
kshyju, line 257: Can you remind me why do we need this? Can you share a sample worker.config entry which is applicable here?

src/WebJobs.Script/Workers/ProcessManagement/WorkerDescription.cs
mathewc: Yes, I find it confusing that there's an ExecutableWorkingDirectory, WorkerDirectory and DefaultWorkerPath.

DefaultWorkerPath I assume is the path to the worker executable? What's the difference between the other two?

fabiocav: In agreement. We need to make sure the description here is very clear so there's no confusion about what each one of those properties control.

src/WebJobs.Script/Workers/Rpc/RpcWorkerDescription.cs
kshyju: Ah never mind. I missed the worker directory in the PR description. Thanks!

src/WebJobs.Script/Workers/ProcessManagement/WorkerDescription.cs
ghost: `WorkerDirectory` is the base directory of worker.

But the executables that are used to run the worker code can be anywhere in the machine, and `ExecutableWorkingDirectory` is the working directory of the worker process executing the worker, it might not be same for some of PAL scenarios, where `ExecutableWorkingDirectory` should be set to PAL executable binaries

`WorkerDirectory` doesn't update the working directory of forked worker process

---

Any suggestion on the comment is welcome, current comment came from suggestion above in the PR Comments

test/WebJobs.Script.Tests/Workers/Rpc/RpcWorkerProcessTests.cs
ghost: These came from another comment in PR to refactor the code so that we don't create 2 instances of `RpcWorkerProcess`.
Updated the private naming in next iteration

kshyju (approved):

src/WebJobs.Script/Workers/ProcessManagement/WorkerDescription.cs
kshyju: @siddharth-ms  I meant to make the clarifying note/comments in the code itself (WorkerDescription class). Feel free to be more descriptive with relevant examples as our goal is to help future readers.

ghost: Got it, added some more remarks here

fabiocav (approved):

src/WebJobs.Script/Workers/ProcessManagement/WorkerDescription.cs
kshyju: @siddharth-ms I still don’t see any detailed comments highlighting the differences in the file, similar to the one you posted earlier. Please make sure to add them before merging the PR. This will help future readers understand the difference between these properties and when to use what. Thank you!

PR #10698, opened 2024-12-13
Allow for an output binding value of an invocation result to be `null`

satvu:

Part of the fix for https://github.com/Azure/azure-functions-dotnet-worker/issues/2682, related to https://github.com/Azure/azure-functions-dotnet-worker/pull/2896.

Allows for output binding values to be `null`, which is interpreted as the type `ParameterBinding.None` when receiving the values over gRPC.

### Pull request checklist

### Additional information

[Here](https://github.com/Azure/azure-functions-host/pull/6836) is the original PR adding the switch statement. The logic was originally added to differentiate between data transferred by the worker using shared memory vs RPC. It doesn't appear to purposefully prevent output bindings from having a `null` value. 

Invocation results are nullable (see [here](https://github.com/Azure/azure-functions-host/blob/de44db4bf69f533c3e30da41f989ddb318a548f6/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs#L1157)), and the HTTP Proxying design sends a null result over the gRPC pipeline.


src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -1157,6 +1159,8 @@ private async Task<ScriptInvocationResult> CreateScriptInvocationResult(Invocati
                 Return = invokeResponse?.ReturnValue?.ToObject()
             };
 
+            _workerChannelLogger.LogDebug("Successfully parsed the result of the invocation with invocationId: '{invocationId}'", invokeResponse.InvocationId);
```
satvu: Added for testing purposes - most of the other unit tests for this class rely on logs (showing that function metadata has been processed, showing that invocations were received, etc), so followed the existing architecture.

test/WebJobs.Script.Tests/Workers/Rpc/GrpcWorkerChannelTests.cs
```diff
@@ -1546,6 +1546,21 @@ await TestHelpers.Await(() => GetInvocationLogs().Length == logLoop,
             }
         }
 
+        [Fact]
+        public async Task NullOutputBinding_DoesNotThrow()
```
satvu, line 1550: Fails without change (crashes before logging success), but succeeds with the new change.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
jviau: Can you timebox looking into an alternative way to test? I am not a fan of using logs for testing as this impacts production logs. Even though you use a debug log, we aren't filtering much in production and this will show up in our tables and possible customers.

I see the default switch statement throws an exception. An acceptable unit test to me is a before & after test. Where you write the test and run it without your changes and see it throw an exception (failing the test). Then with your changes it no longer throws and the test passes. That is good enough to me, no need to verify a log. This is all provided you can see the exception be thrown in the test to begin with.

test/WebJobs.Script.Tests/Workers/Rpc/GrpcWorkerChannelTests.cs
jviau: Ah my other comment touched on this: since this test will fail via thrown exception without your changes, I don't think you need any of the logs to validate the change. Just a lack of thrown exception (and thus test passes) is good enough.

jviau (approved):

liliankasem (approved):

kshyju:
Is this a breaking change for customers?

satvu:
> Is this a breaking change for customers?

I don't believe this will _break_ existing customers explicitly. It will allow for a successful invocation to have an output binding with a `null` value. Without this change, the scenario would throw an exception.

I think that currently, cases where we can expect `null` output binding values are usually when invocations fail, and that is handled in a different code path starting from [here](https://github.com/Azure/azure-functions-host/blob/c84ce0dd2033e016a9445ec4515b14374ba7c20f/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs#L1121). 

This could definitely open the door for some new scenarios though. If a customer expects to have a value for their output binding but their function invocation status is "successful" despite an erroneous `null` value, this could break another one of their components.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -1068,6 +1068,8 @@ private async Task<object> GetBindingDataAsync(ParameterBinding binding, string
                 case ParameterBindingType.Data:
                     // Data was transferred by the worker using RPC
                     return binding.Data.ToObject();
+                case ParameterBindingType.None:
+                    return null;
                 default:
                     throw new InvalidOperationException("Unknown ParameterBindingType");
```
brettsam: I know this isn't your change... but can you update this exception to include the type in case we ever see this?

brettsam (approved):

PR #10731, opened 2025-01-09
[in-proc backport] Update http proxy to handle client disconnect scenario 

liliankasem:
Backport of #10688

### DIFF
--- a/src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
+++ b/src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
@@ -8,6 +8,7 @@ using System.Net.Http;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Azure.WebJobs.Script.Description;
+using Microsoft.Azure.WebJobs.Script.Grpc.Exceptions;
 using Microsoft.Azure.WebJobs.Script.Workers;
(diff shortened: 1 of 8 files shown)


src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
```diff
@@ -56,13 +59,23 @@ public async Task EnsureSuccessfulForwardingAsync(ScriptInvocationContext contex
 
             if (httpProxyTaskResult is not ForwarderError.None)
             {
+                _logger.LogDebug($"The function invocation failed to proxy the request with ForwarderError: {httpProxyTaskResult}");
```
jviau: Was this in the original PR? We shouldn't be using string interpolation in log messages.

jviau (approved):

src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
liliankasem: Ah crap I totally missed this, and yes it was - I will fix it in both branches

liliankasem: Also there are a bunch of string interpolation in log messages across the host that we should probably fix at some point too.

satvu (approved):

PR #10734, opened 2025-01-10
WIP: Update all logs to remove use of string interpolation

liliankasem:
Updating logs to remove use of string interpolation. I know it's a lot of files, but the changes are very minimal.

I intentionally did not fix some formatting in these log messages to avoid breaking detectors and other items that depend on the current format. With the exception of things like:

`_logger.LogError($"Error reading previous start context: {e.ToString()}");`

Fixed to:

`_logger.LogError(e, "Error reading previous start context.");`

Is there are concerns about this, I can follow the same principal for this and keep the exception string in the message itself.

TODO: Update tests

### DIFF
--- a/src/WebJobs.Script.WebHost/ContainerManagement/LinuxContainerActivityPublisher.cs
+++ b/src/WebJobs.Script.WebHost/ContainerManagement/LinuxContainerActivityPublisher.cs
@@ -142,13 +142,13 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.ContainerManagement
                         {
                             if (_lastHeartBeatTime.AddMinutes(5) < DateTime.UtcNow)
                             {
-                                _logger.LogDebug($"Current activities count = {currentActivities.Count}");
+                                _logger.LogDebug("Current activities count = {currentActivitiesCount}", currentActivities.Count);
                                 _lastHeartBeatTime = DateTime.UtcNow;
                             }
 
                             if (currentActivities.Any())
                             {
-                                _logger.LogDebug($"Flushing {currentActivities.Count} function activities");
+                                _logger.LogDebug("Flushing {currentActivitiesCount} function activities", currentActivities.Count);
                                 await _meshServiceClient.PublishContainerActivity(currentActivities);
                             }
                         }
@@ -241,7 +241,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.ContainerManagement
             {
                 if (!PublishActivity(activity))
                 {
-                    _logger.LogWarning($"Failed to add activity {activity}");
+                    _logger.LogWarning("Failed to add activity {containerFunctionExecutionActivity}", activity);
                 }
             }
         }
--- a/src/WebJobs.Script.WebHost/Controllers/HostController.cs
+++ b/src/WebJobs.Script.WebHost/Controllers/HostController.cs
@@ -276,7 +276,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Controllers
             {
                 ScriptHostState currentState = scriptHostManager.State;
 
-                _logger.LogDebug($"Received request to resume a draining host - host status: {currentState.ToString()}");
+                _logger.LogDebug("Received request to resume a draining host - host status: {currentHostStatus}", currentState);
 
                 if (currentState != ScriptHostState.Running
                     || !Utility.TryGetHostService(scriptHostManager, out IDrainModeManager drainModeManager))
@@ -285,7 +285,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Controllers
                     return StatusCode(StatusCodes.Status409Conflict);
                 }
 
-                _logger.LogDebug($"Drain mode enabled: {drainModeManager.IsDrainModeEnabled}");
+                _logger.LogDebug("Drain mode enabled: {isDrainModeEnabled}", drainModeManager.IsDrainModeEnabled);
 
                 if (drainModeManager.IsDrainModeEnabled)
                 {
--- a/src/WebJobs.Script.WebHost/Standby/StandbyManager.cs
+++ b/src/WebJobs.Script.WebHost/Standby/StandbyManager.cs
@@ -197,7 +197,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
             }
 
             string scriptPath = options.ScriptPath;
-            _logger.LogInformation($"Creating StandbyMode placeholder function directory ({scriptPath})");
+            _logger.LogInformation("Creating StandbyMode placeholder function directory ({scriptPath})", scriptPath);
 
             await FileUtility.DeleteDirectoryAsync(scriptPath, true);
             FileUtility.EnsureDirectoryExists(scriptPath);
--- a/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
+++ b/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
@@ -755,13 +755,13 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
             {
                 hisMode = "Strict";
                 _metricsLogger.LogEvent(MetricEventNames.HISStrictModeEnabled);
-                _logger.LogDebug($"HIS Strict mode enabled.");
+                _logger.LogDebug("HIS Strict mode enabled.");
             }
             else if (FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagStrictHISModeWarn))
             {
                 hisMode = "Warn";
                 _metricsLogger.LogEvent(MetricEventNames.HISStrictModeWarn);
-                _logger.LogDebug($"HIS Warn mode enabled.");
+                _logger.LogDebug("HIS Warn mode enabled.");
             }
 
             logger.LogHostInitializationSettings(_originalFunctionsWorkerRuntime, functionWorkerRuntime, _originalFunctionsWorkerRuntimeVersion, functionWorkerRuntimeVersion,
--- a/src/WebJobs.Script/StorageProvider/HostAzureBlobStorageProvider.cs
+++ b/src/WebJobs.Script/StorageProvider/HostAzureBlobStorageProvider.cs
@@ -61,7 +61,7 @@ namespace Microsoft.Azure.WebJobs.Script
             if (_storageOptions?.CurrentValue.InternalSasBlobContainer != null)
             {
                 blobContainerClient = new BlobContainerClient(new Uri(_storageOptions.CurrentValue.InternalSasBlobContainer));
-                _logger.LogDebug($"Using storage account {blobContainerClient.AccountName} and container {blobContainerClient.Name} for hosting BlobContainerClient.");
+                _logger.LogDebug("Using storage account {blobAccountName} and container {blobContainerName} for hosting BlobContainerClient.", blobContainerClient.AccountName, blobContainerClient.Name);
                 return true;
             }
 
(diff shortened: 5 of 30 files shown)


src/WebJobs.Script.WebHost/ContainerManagement/LinuxContainerActivityPublisher.cs
```diff
@@ -92,7 +92,7 @@ public Task StartAsync(CancellationToken cancellationToken)
 
         public Task StopAsync(CancellationToken cancellationToken)
         {
-            _logger.LogInformation($"Stopping {nameof(LinuxContainerActivityPublisher)}");
+            _logger.LogInformation("Stopping {nameof(LinuxContainerActivityPublisher)}", nameof(LinuxContainerActivityPublisher));
```
jviau: This one should be reverted - it is a constant interpolation, no need to use a structured log

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -276,7 +276,7 @@ public async Task<IActionResult> Resume([FromServices] IScriptHostManager script
             {
                 ScriptHostState currentState = scriptHostManager.State;
 
-                _logger.LogDebug($"Received request to resume a draining host - host status: {currentState.ToString()}");
+                _logger.LogDebug("Received request to resume a draining host - host status: {currentState}", currentState.ToString());
```
jviau: `.ToString()` calls should be removed. The log system will call `ToString()` only when the log is actually being written

src/WebJobs.Script.WebHost/Management/LinuxSpecialization/RunFromPackageHandler.cs
```diff
@@ -193,10 +193,10 @@ public async Task<bool> MountAzureFileShare(HostAssignmentContext assignmentCont
                 using (_metricsLogger.LatencyEvent(MetricEventNames.LinuxContainerSpecializationMountCifs))
                 {
                     var targetPath = _environment.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebsiteHomePath);
-                    _logger.LogDebug($"Mounting {EnvironmentSettingNames.AzureFilesContentShare} at {targetPath}");
+                    _logger.LogDebug("Mounting {azureFilesContentShare} at {targetPath}", EnvironmentSettingNames.AzureFilesContentShare, targetPath);
                     bool succeeded = await _meshServiceClient.MountCifs(assignmentContext.AzureFilesConnectionString,
                         assignmentContext.AzureFilesContentShare, targetPath);
-                    _logger.LogInformation($"Mounted {EnvironmentSettingNames.AzureFilesContentShare} at {targetPath} Success = {succeeded}");
+                    _logger.LogInformation("Mounted {azureFilesContentShare} at {targetPath} Success = {succeeded}", EnvironmentSettingNames.AzureFilesContentShare, targetPath, succeeded);
```
jviau: `EnvironmentSettingNames.AzureFilesContentShare` is a constant. Not sure if we want that to be parameterized

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -390,18 +390,18 @@ public async Task<IDictionary<string, WorkerStatus>> GetWorkerStatusesAsync()
 
         public async Task ShutdownAsync()
         {
-            _logger.LogDebug($"Waiting for {nameof(RpcFunctionInvocationDispatcher)} to shutdown");
+            _logger.LogDebug("Waiting for {name} to shutdown", nameof(RpcFunctionInvocationDispatcher));
```
jviau: This is effectively a constant, shouldn't be structured. Same with lines 400 and 404

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -289,15 +289,15 @@ public async Task InitializeAsync(IEnumerable<FunctionMetadata> functions, Cance
...
+                _logger.LogDebug("Languages present in WorkerConfig: {allLanguageNamesFromWorkerConfigs}, allLanguageNamesFromWorkerConfigs");
 
                 throw new InvalidOperationException($"WorkerConfig for runtime: {_workerRuntime} not found");
             }
 
             if (functions == null || !functions.Any())
             {
                 // do not initialize function dispatcher if there are no functions, unless the worker is indexing
-                _logger.LogDebug($"{nameof(RpcFunctionInvocationDispatcher)} received no functions");
+                _logger.LogDebug("{name} received no functions", nameof(RpcFunctionInvocationDispatcher));
```
jviau: No need to parameterize this constant.

src/WebJobs.Script/Workers/Http/HttpFunctionInvocationDispatcher.cs
```diff
@@ -185,7 +185,7 @@ protected virtual void Dispose(bool disposing)
         {
             if (!_disposed && disposing)
             {
-                _logger.LogDebug($"Disposing {nameof(HttpFunctionInvocationDispatcher)}");
+                _logger.LogDebug("Disposing {name}", nameof(HttpFunctionInvocationDispatcher));
```
jviau: No need to parameterize this constant.

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptionsSetup.cs
```diff
@@ -37,7 +37,8 @@ public void Configure(HttpWorkerOptions options)
 
             if (httpWorkerSection.Exists() && customHandlerSection.Exists())
             {
-                _logger.LogWarning($"Both {ConfigurationSectionNames.HttpWorker} and {ConfigurationSectionNames.CustomHandler} sections are spefified in {ScriptConstants.HostMetadataFileName} file. {ConfigurationSectionNames.CustomHandler} takes precedence.");
+                _logger.LogWarning("Both {httpWorker} and {customHandler} sections are spefified in {hostMetadata} file. {customHandler} takes precedence.",
```
jviau: These are all constants, don't think we should parameterize it.

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptionsSetup.cs
```diff
@@ -47,7 +48,7 @@ public void Configure(HttpWorkerOptions options)
                 if (options.Type == CustomHandlerType.None)
                 {
                     // CustomHandlerType.None is only for maitaining backward compatibilty with httpWorker section.
-                    _logger.LogWarning($"CustomHandlerType {CustomHandlerType.None} is not supported. Defaulting to {CustomHandlerType.Http}.");
+                    _logger.LogWarning("CustomHandlerType {typeNone} is not supported. Defaulting to {typeHttp}.", CustomHandlerType.None, CustomHandlerType.Http);
```
jviau: This is effectively constant, don't think we should parameterize it.

src/WebJobs.Script/StorageProvider/HostAzureBlobStorageProvider.cs
```diff
@@ -61,13 +61,13 @@ public virtual bool TryCreateHostingBlobContainerClient(out BlobContainerClient
...
                 blobContainerClient = new BlobContainerClient(new Uri(_storageOptions.CurrentValue.InternalSasBlobContainer));
-                _logger.LogDebug($"Using storage account {blobContainerClient.AccountName} and container {blobContainerClient.Name} for hosting BlobContainerClient.");
+                _logger.LogDebug("Using storage account {accountName} and container {containerName} for hosting BlobContainerClient.", blobContainerClient.AccountName, blobContainerClient.Name);
                 return true;
             }
 
             if (!TryCreateBlobServiceClientFromConnection(ConnectionStringNames.Storage, out BlobServiceClient blobServiceClient))
             {
-                _logger.LogDebug($"Could not create BlobContainerClient using Connection: {ConnectionStringNames.Storage}");
+                _logger.LogDebug("Could not create BlobContainerClient using Connection: {storageConnectionName}", ConnectionStringNames.Storage);
```
jviau: Static/constant, don't think it should be parameterized.

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -909,12 +909,12 @@ private void DisposeDependencyTrackingModule(IHost instance)
 
                 module?.Dispose();
 
-                _logger.LogDebug($"{nameof(DependencyTrackingTelemetryModule)} disposed.");
+                _logger.LogDebug("{name} disposed.", nameof(DependencyTrackingTelemetryModule));
```
jviau: Constant here as well and line 917

src/WebJobs.Script.WebHost/Standby/StandbyManager.cs
```diff
@@ -63,7 +63,7 @@ public StandbyManager(IScriptHostManager scriptHostManager, IWebHostRpcWorkerCha
             _configuration = configuration as IConfigurationRoot ?? throw new ArgumentNullException(nameof(configuration));
             _rpcWorkerChannelManager = rpcWorkerChannelManager ?? throw new ArgumentNullException(nameof(rpcWorkerChannelManager));
             _hostNameProvider = hostNameProvider ?? throw new ArgumentNullException(nameof(hostNameProvider));
-            _changeTokenCallbackSubscription = ChangeToken.RegisterChangeCallback(_ => _logger.LogDebug($"{nameof(StandbyManager)}.{nameof(ChangeToken)} callback has fired."), null);
+            _changeTokenCallbackSubscription = ChangeToken.RegisterChangeCallback(_ => _logger.LogDebug("{name}.{changeToken} callback has fired.", nameof(StandbyManager), nameof(ChangeToken)), null);
```
jviau: Constants here as well

jviau (changes requested):
Thanks for doing this! We should discuss a bit more offline though as there are some customer impact here.

1. This may be a breaking change as the log structure is changing. Particularly with moving exceptions into a different property and out of the log message.
2. Log properties should be considered in a _global_ namespace. All structured log properties are grouped together with any scope variables, so names may conflict / overwrite there causing information loss. We need to put thought into the exact property names we use.
3. We are going to be doing deeper telemetry changes to comply with OTel, I wonder how this fits into that?

I also left comments regarding parameterization of static/constants. I should clarify the avoiding of string interpolation rule is actually "avoid templates which may change over the lifetime of an application". String interpolation with constants does not violate that rule.

liliankasem:
> Thanks for doing this! We should discuss a bit more offline though as there are some customer impact here.
> 
> 1. This may be a breaking change as the log structure is changing. Particularly with moving exceptions into a different property and out of the log message.
> 2. Log properties should be considered in a _global_ namespace. All structured log properties are grouped together with any scope variables, so names may conflict / overwrite there causing information loss. We need to put thought into the exact property names we use.
> 3. We are going to be doing deeper telemetry changes to comply with OTel, I wonder how this fits into that?
> 
> I also left comments regarding parameterization of static/constants. I should clarify the avoiding of string interpolation rule is actually "avoid templates which may change over the lifetime of an application". String interpolation with constants does not violate that rule.

@jviau  Thanks for taking a look!

1. Yeah agree, I was wondering this as I was making this change but wasn't 100% sure. We can undo changes all of the exception logs.
2. Noted
3. Good question, I not sure what the work around OTel will be, but this PR was intended as low hanging fruit - so if it's not worth it or ends up being more than this, I can abort the PR

fabiocav:
What is the state of this PR? Still something you want to consider?

I agree with the concerns raised, but the vast majority of the changes here would expand what we end up logging (additional properties), while preserving the original log value. 

I agree that part of the OTel and observability work will lead to a review of all logging, so if we want to wait for that instead, happy with that as well.

liliankasem:
> What is the state of this PR? Still something you want to consider?
> 
> I agree with the concerns raised, but the vast majority of the changes here would expand what we end up logging (additional properties), while preserving the original log value.
> 
> I agree that part of the OTel and observability work will lead to a review of all logging, so if we want to wait for that instead, happy with that as well.

I think let's just wait for the observability work, closing for now!

PR #10735, opened 2025-01-10
Fix for ISO string modification during function metadata binding parsing

### LINKED ISSUE #10732: Parsing bindings into a JObject modifies strings in ISO format
Related to: https://github.com/Azure/azure-functions-dotnet-worker/issues/2882

In [this line](https://github.com/Azure/azure-functions-host/blob/cb0a62c51dd800ee1919795b4a933b26c74ed8de/src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs#L253), the binding and properties are parsed using Newtonsoft.JSON, which has [this known bug](https://github.com/JamesNK/Newtonsoft.Json/issues/862) regarding ISO strings. Binding property values in ISO format are unintentionally modified.

satvu:

resolves #10732 (and https://github.com/Azure/azure-functions-dotnet-worker/issues/2882)

### Pull request checklist

    * [] Otherwise: Backport tracked by issue/PR #issue_or_pr

### Additional information

Tested locally E2E the following scenarios:

1. `StartFromTime` set to a past time started triggering and processing existing records immediately after startup (older records than start time untouched)
2. `StartFromTime` set to a future date and will process any new changes following startup. Discussed with original engineers - this feature was only meant to go into the past, so there was never a design set for future dates. This setting will not ignore changes until that future date (_note: need to test InProc to see if behavior is the same there_).


src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
```diff
@@ -244,13 +247,26 @@ internal IEnumerable<FunctionMetadata> ValidateMetadata(IEnumerable<RawFunctionM
             return validatedMetadata;
         }
 
+        private static BindingMetadata GetBindingMetadata(string binding)
```
liliankasem: Can this be `CreateBindingMetadata` ?

src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
```diff
@@ -244,13 +247,26 @@ internal IEnumerable<FunctionMetadata> ValidateMetadata(IEnumerable<RawFunctionM
...
+        {
+            using (StringReader stringReader = new StringReader(binding))
+            using (JsonTextReader jsonTextReader = new JsonTextReader(stringReader))
+            {
+                JsonSerializer serializer = JsonSerializer.Create(_jsonSerializerSettings);
```
liliankasem: Can we DI the JsonSerializer so we're not creating one every time this method is called?

src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
```diff
@@ -28,6 +30,7 @@ internal class WorkerFunctionMetadataProvider : IWorkerFunctionMetadataProvider
         private readonly IScriptHostManager _scriptHostManager;
         private string _workerRuntime;
         private ImmutableArray<FunctionMetadata> _functions;
+        private static JsonSerializerSettings _jsonSerializerSettings = new JsonSerializerSettings { DateParseHandling = DateParseHandling.None };
```
liliankasem: Can we setup the serializer + settings as part of the ctor?

src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
liliankasem: nit: can the private method be underneath the internal methods?

release_notes.md
```diff
@@ -10,3 +10,4 @@
   - [Added net9 prelaunch app.](https://github.com/Azure/azure-functions-dotnet-worker/pull/2898)
 - Update the `DefaultHttpProxyService` to better handle client disconnect scenarios (#10688)
   - Replaced `InvalidOperationException` with `HttpForwardingException` when there is a ForwarderError
+- Fix the accidental modification of ISO strings during function metadata binding parsing. This fixes the listener errors related to the CosmosDB trigger's `StartFromTime` parameter. (#10735)
```
liliankasem:
```suggestion
- Fix modification of ISO strings during the parsing of function metadata bindings.
  - This fixes the listener errors related to the CosmosDB trigger's `StartFromTime` parameter. (#10735)
```

src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
kshyju: We already have another instance of `JsonSerializerSettings` with the exact same properties in [another class](https://github.com/Azure/azure-functions-host/blob/969f76134a60c1b5bb13b77d56169bf324a8f1c6/src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs#L25). I think we should use that shared instance (move it to another class and use it in both places). You could have the shared serializer there.

For example, something like this.

```csharp
internal static class Shared
{
    internal static class JsonNetSerializerSettings
    {
        // Shared JsonSerializerSettings instance with DateParseHandling set to None
        internal static readonly JsonSerializerSettings NoDateParsing = new()
        {
            DateParseHandling = DateParseHandling.None
        };
    }

    internal static class JsonNetSerializers
    {
        // Shared JsonSerializer instance using the NoDateParsing settings
        internal static readonly JsonSerializer NoDateParsingSerializer = 
                    JsonSerializer.Create(JsonNetSerializerSettings.NoDateParsing);
    }
}
```
Now in the `GrpcMessageConversionExtensions`, you can use it like `Shared.JsonNetSerializerSettings.NoDateParsing` and in your new method,

```csharp
var parsedBinding = Shared.JsonNetSerializers
                              .NoDateParsingSerializer.Deserialize<JObject>(jsonTextReader);
```

src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
```diff
@@ -244,13 +247,26 @@ internal IEnumerable<FunctionMetadata> ValidateMetadata(IEnumerable<RawFunctionM
...
+        }
+
         internal static FunctionMetadata ValidateBindings(IEnumerable<string> rawBindings, FunctionMetadata function)
         {
             HashSet<string> bindingNames = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
 
             foreach (string binding in rawBindings)
             {
-                var functionBinding = BindingMetadata.Create(JObject.Parse(binding));
+                var functionBinding = GetBindingMetadata(binding);
```
kshyju: Should this new behavior be an explicit opt-in for customers? With this change, all customers and bindings will experience the performance impact of the additional serialization for any bindings that pass through this flow or method.

satvu:
Discussed offline with @liliankasem - to incorporate suggestions we will need to refactor `ValidateBindings` (see [here](https://github.com/Azure/azure-functions-host/blob/969f76134a60c1b5bb13b77d56169bf324a8f1c6/src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs#L247)) and change it from a static method. Investigation shows that the only places where this method is called is in tests and within the same class. This should be a safe change.

src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
satvu: I don't know if we should have an explicit opt-in for a behavior that should've already been working. I believe that the use of ISO strings is only for the CosmosDB extension, so we _could_ only check for the trigger or only for this property.

satvu: Note: we will benchmark the two approaches and compare the results to decide how to proceed.

satvu: Benchmarking results [here](https://gist.github.com/satvu/d7b900777ae7ba000660a0c432c174e7).

src/WebJobs.Script/Shared.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.WebJobs.Script
+{
+    internal static class Shared
```
liliankasem: I don't love this class. The name is really vague. Could you create a folder called `Serializers` and have a file with both of these classes in it?

Also do the settings need to be their own separate class? 

how about something like this?

```
using Newtonsoft.Json;

namespace Microsoft.Azure.WebJobs.Script.Serialization
{
    /// <summary>
    /// Provides shared JSON serialization settings and serializers.
    /// </summary>
    internal static class JsonSerialization
    {
        /// <summary>
        /// JSON serializer settings with no date parsing.
        /// </summary>
        public static readonly JsonSerializerSettings NoDateParsingSettings = new JsonSerializerSettings
        {
            DateParseHandling = DateParseHandling.None
        };

        /// <summary>
        /// JSON serializer instance configured with no date parsing settings.
        /// </summary>
        public static readonly JsonSerializer NoDateParsingSerializer =
            JsonSerializer.Create(NoDateParsingSettings);
    }
}
```

liliankasem: Second I am concerned about the thread safety of this, and testability? 

Is there a second we didn't go with DI for this?

src/WebJobs.Script/Shared.cs
```diff
@@ -1,24 +0,0 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
```
liliankasem: I think the file needs to be deleted?

src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs
```diff
@@ -1,4 +1,5 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
```
satvu: White space at the end was cleaned up automatically - I think it's ok to keep the change.

src/WebJobs.Script/Shared.cs
satvu: Discussed offline - for the scope of this PR, static serializer settings and using `JsonConvert` is sufficient. Review of serializers across the host will be saved for a larger refactor/review of how we use Newtonsoft.Json and System.Text.Json serializers across the host.

src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs
```diff
@@ -22,7 +23,7 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
 {
     internal static class GrpcMessageConversionExtensions
     {
-        private static readonly JsonSerializerSettings _datetimeSerializerSettings = JsonSerializers.NoDateParsingSettings;
+        private static readonly Newtonsoft.Json.JsonSerializerSettings _datetimeSerializerSettings = JsonSerializers.NoDateParsingSettings;
```
liliankasem: I would say revert all changes to this file?

`git checkout origin/dev src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs`

liliankasem (approved):

kshyju (approved):

test/WebJobs.Script.Tests/WorkerFunctionMetadataProviderTests.cs
```diff
@@ -180,5 +196,19 @@ public async void ValidateFunctionMetadata_Logging()
...
             // The third log is Host is running without any initialized channels, restarting the JobHost. This is not relevant to this test.
         }
+
+        [Fact]
+        public void ValidateFunctionMetadata_IsoStringNotAltered()
+        {
+            FunctionMetadata functionMetadata = new FunctionMetadata();
+            List<string> rawBindings = new List<string>();
+            var isoString = "2025-02-10T22:45:33Z";
+            rawBindings.Add("{\"type\": \"cosmosDBTrigger\",\"name\": \"cosmosTrigger\",\"direction\": \"in\",\"databaseName\":\"databaseName\"," +
```
kshyju, line 206: May be raw literal string can be used to store the JSON and can avoid these many escape characters.

PR #10741, opened 2025-01-14
Adding check for service and instances count.

### LINKED ISSUE #10780: ArgumentOutOfRangeException from CloudWorkFlow extension
CreateChildContainer is throwing an exception while loading CloudWorkFlow extension. 

If a service registration issue occurs due to an exception, it can be challenging for the host to consistently capture it. This is because the extension likely catches the exception during registration and doesn't propagate it, even though the host is already set up to log initialization failures within the extension's DI path.

Enhance the logging to pinpoint the specific service causing the exception. The fix will likely need to be made in the extension where those services are initialized, rather than in the host.

System.ArgumentOutOfRangeException : Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')
at System.ThrowHelper.ThrowArgumentOutOfRange_IndexMustBeLessException()
at System.SZArrayHelper.get_Item[T](Int32 index)
at System.Linq.Enumerable.ElementAt[TSource](IEnumerable`1 source,Int32 index)
at Microsoft.Azure.WebJobs.Script.WebHost.DependencyInjection.ServiceProviderExtensions.CreateChildContainer(IServiceProvider serviceProvider,IServiceCollection serviceCollection)
at Microsoft.Azure.WebJobs.Script.WebHost.DependencyInjection.JobHostScopedServiceProviderFactory.CreateServiceProvider(IServiceCollection services)
at //src/WebJobs.Script.WebHost/DependencyInjection/JobHostScopedServiceProviderFactory.cs : 68
at Microsoft.Extensions.Hosting.HostBuilder.InitializeServiceProvider()
at Microsoft.Extensions.Hosting.HostBuilder.Build()
at async Microsoft.Azure.WebJobs.Script.WebHost.WebJobsScriptHostService.UnsynchronizedStartHostAsync(ScriptHostStartupOperation activeOperation,Int32 attemptCount,JobHostStartupMode startupMode)
at //src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs : 346

RohitRanjanMS:
Adding a check in the `CreateChildContainer` method to validate the counts of services and instances. Validation ensures that the number of services matches the number of instances, preventing runtime errors when processing them together. Some applications are encountering the following error, and this change will provide additional logs to aid in investigating the issue.

> System.ArgumentOutOfRangeException : Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')
   at System.ThrowHelper.ThrowArgumentOutOfRange_IndexMustBeLessException()
   at System.SZArrayHelper.get_Item[T](Int32 index)
   at System.Linq.Enumerable.ElementAt[TSource](IEnumerable`1 source,Int32 index)
   at Microsoft.Azure.WebJobs.Script.WebHost.DependencyInjection.ServiceProviderExtensions.CreateChildContainer(IServiceProvider serviceProvider,IServiceCollection serviceCollection)
   at Microsoft.Azure.WebJobs.Script.WebHost.DependencyInjection.JobHostScopedServiceProviderFactory.CreateServiceProvider(IServiceCollection services) 
   at /_/src/WebJobs.Script.WebHost/DependencyInjection/JobHostScopedServiceProviderFactory.cs : 68
   at Microsoft.Extensions.Hosting.HostBuilder.InitializeServiceProvider()
   at Microsoft.Extensions.Hosting.HostBuilder.Build()
   at async Microsoft.Azure.WebJobs.Script.WebHost.WebJobsScriptHostService.UnsynchronizedStartHostAsync(ScriptHostStartupOperation activeOperation,Int32 attemptCount,JobHostStartupMode startupMode) 
   at /_/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs : 346

**Sample log**
> Mismatch detected for type Microsoft.AspNetCore.Mvc.ApplicationModels.IPageApplicationModelProvider:
services.Count = 6, instances.Count = 5
Services:
        Microsoft.AspNetCore.Mvc.ApplicationModels.IPageApplicationModelProvider (Singleton),
         Microsoft.AspNetCore.Mvc.ApplicationModels.IPageApplicationModelProvider (Singleton),
         Microsoft.AspNetCore.Mvc.ApplicationModels.IPageApplicationModelProvider (Singleton),
         Microsoft.AspNetCore.Mvc.ApplicationModels.IPageApplicationModelProvider (Singleton),
         Microsoft.AspNetCore.Mvc.ApplicationModels.IPageApplicationModelProvider (Singleton),
         Microsoft.AspNetCore.Mvc.ApplicationModels.IPageApplicationModelProvider (Singleton)
Instances:
        Microsoft.AspNetCore.Mvc.ApplicationModels.DefaultPageApplicationModelProvider,
        Microsoft.AspNetCore.Mvc.ApplicationModels.AutoValidateAntiforgeryPageApplicationModelProvider,
        Microsoft.AspNetCore.Mvc.ApplicationModels.AuthorizationPageApplicationModelProvider,
        Microsoft.AspNetCore.Mvc.ApplicationModels.TempDataFilterPageApplicationModelProvider,
        Microsoft.AspNetCore.Mvc.ApplicationModels.ViewDataAttributePageApplicationModelProvider

This will help investigate 

resolves #10780 
### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/DependencyInjection/ServiceProviderExtensions.cs
+++ b/src/WebJobs.Script.WebHost/DependencyInjection/ServiceProviderExtensions.cs
@@ -155,7 +155,6 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.DependencyInjection
                                   {string.Join(",\n\t", instancesArray.Select(i => i?.GetType()?.ToString() ?? "null"))}
                                 """;
 
-
                 throw new InvalidOperationException(message);
             }
         }
--- a/src/WebJobs.Script.WebHost/DependencyInjection/ServiceProviderExtensions.cs
+++ b/src/WebJobs.Script.WebHost/DependencyInjection/ServiceProviderExtensions.cs
@@ -7,6 +7,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
@@ -98,11 +99,16 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.DependencyInjection
                 else if (services.All(s => s.Lifetime == ServiceLifetime.Singleton))
                 {
                     // We can resolve them from the main container.
-                    var instances = serviceProvider.GetServices(services.Key);
+                    // Materialize services and instances to arrays once.
+                    var servicesArray = services.ToArray();
+                    var instancesArray = serviceProvider.GetServices(services.Key)?.ToArray() ?? Array.Empty<object>();
 
-                    for (var i = 0; i < services.Count(); i++)
+                    // Validate counts
+                    ValidateServiceAndInstanceCounts(servicesArray, instancesArray, services.Key);
+
+                    for (var i = 0; i < servicesArray.Length; i++)
                     {
-                        clonedCollection.CloneSingleton(services.ElementAt(i), instances.ElementAt(i));
+                        clonedCollection.CloneSingleton(servicesArray[i], instancesArray[i]);
                     }
                 }
 
@@ -112,24 +118,44 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.DependencyInjection
                     // We need a service scope to resolve them.
                     using var scope = serviceProvider.CreateScope();
 
-                    var instances = scope.ServiceProvider.GetServices(services.Key);
+                    // Materialize services and instances to arrays once.
+                    var servicesArray = services.ToArray();
+                    var instancesArray = scope.ServiceProvider.GetServices(services.Key)?.ToArray() ?? Array.Empty<object>();
+
+                    ValidateServiceAndInstanceCounts(servicesArray, instancesArray, services.Key);
 
                     // Then we only keep singleton instances.
-                    for (var i = 0; i < services.Count(); i++)
+                    for (var i = 0; i < servicesArray.Length; i++)
                     {
-                        if (services.ElementAt(i).Lifetime == ServiceLifetime.Singleton)
+                        if (servicesArray[i].Lifetime == ServiceLifetime.Singleton)
                         {
-                            clonedCollection.CloneSingleton(services.ElementAt(i), instances.ElementAt(i));
+                            clonedCollection.CloneSingleton(servicesArray[i], instancesArray[i]);
                         }
                         else
                         {
-                            clonedCollection.Add(services.ElementAt(i));
+                            clonedCollection.Add(servicesArray[i]);
                         }
                     }
                 }
             }
 
             return clonedCollection;
+        }
+
+        private static void ValidateServiceAndInstanceCounts<TService, TInstance>(TService[] servicesArray, TInstance[] instancesArray, Type serviceType)
+        {
+            if (servicesArray.Length != instancesArray.Length)
+            {
+                var messageBuilder = new StringBuilder()
+                    .AppendLine($"Mismatch detected for type {serviceType}:")
+                    .AppendLine($"services.Count = {servicesArray.Length}, instances.Count = {instancesArray.Length}")
+                    .Append("Services:\n\t")
+                    .Append(string.Join(",\n\t", servicesArray.Select(s => s.ToString())))
+                    .Append("\nInstances:\n\t")
+                    .AppendLine(string.Join(",\n\t", instancesArray.Select(i => i?.GetType()?.ToString() ?? "null")));
+                Console.WriteLine(messageBuilder.ToString());
+                throw new InvalidOperationException(messageBuilder.ToString());
+            }
         }
     }
 
--- a/src/WebJobs.Script.WebHost/DependencyInjection/ServiceProviderExtensions.cs
+++ b/src/WebJobs.Script.WebHost/DependencyInjection/ServiceProviderExtensions.cs
@@ -146,15 +146,17 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.DependencyInjection
         {
             if (servicesArray.Length != instancesArray.Length)
             {
-                var messageBuilder = new StringBuilder()
-                    .AppendLine($"Mismatch detected for type {serviceType}:")
-                    .AppendLine($"services.Count = {servicesArray.Length}, instances.Count = {instancesArray.Length}")
-                    .Append("Services:\n\t")
-                    .Append(string.Join(",\n\t", servicesArray.Select(s => s.ToString())))
-                    .Append("\nInstances:\n\t")
-                    .AppendLine(string.Join(",\n\t", instancesArray.Select(i => i?.GetType()?.ToString() ?? "null")));
-                Console.WriteLine(messageBuilder.ToString());
-                throw new InvalidOperationException(messageBuilder.ToString());
+                var message = $"""
+                                Mismatch detected for type {serviceType}:
+                                services.Count = {servicesArray.Length}, instances.Count = {instancesArray.Length}
+                                Services:
+                                  {string.Join(",\n\t", servicesArray.Select(s => s.ToString()))}
+                                Instances:
+                                  {string.Join(",\n\t", instancesArray.Select(i => i?.GetType()?.ToString() ?? "null"))}
+                                """;
+
+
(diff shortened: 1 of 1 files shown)


src/WebJobs.Script.WebHost/DependencyInjection/ServiceProviderExtensions.cs
```diff
@@ -112,25 +118,45 @@ public static IServiceCollection CreateChildContainer(this IServiceProvider serv
...
+
+        private static void ValidateServiceAndInstanceCounts<TService, TInstance>(TService[] servicesArray, TInstance[] instancesArray, Type serviceType)
+        {
+            if (servicesArray.Length != instancesArray.Length)
+            {
+                var messageBuilder = new StringBuilder()
+                    .AppendLine($"Mismatch detected for type {serviceType}:")
+                    .AppendLine($"services.Count = {servicesArray.Length}, instances.Count = {instancesArray.Length}")
+                    .Append("Services:\n\t")
+                    .Append(string.Join(",\n\t", servicesArray.Select(s => s.ToString())))
```
surgupta-msft: Will this be printed only in kusto logs or also in AppInsights logs? I am wondering if this could reveal too much information about our system. Generally what is the reason for this mismatch between services and instances and how can we fix it by knowing this additional info?

RohitRanjanMS: This won't be sent to AppInsights because the exception occurs before the AppInsights pipeline is initialized. However, it will appear on the portal. I don't see any issues with this, but I'll confirm with Fabio.

In one specific app, an exception was thrown while attempting to load the CloudWorkFlow extension, though it recovered later. We reviewed the logs and memory dump but couldn't determine the root cause. We were unable to reproduce this scenario.

src/WebJobs.Script.WebHost/DependencyInjection/ServiceProviderExtensions.cs
```diff
@@ -98,11 +99,16 @@ public static IServiceCollection CreateChildContainer(this IServiceProvider serv
                 else if (services.All(s => s.Lifetime == ServiceLifetime.Singleton))
                 {
                     // We can resolve them from the main container.
-                    var instances = serviceProvider.GetServices(services.Key);
+                    // Materialize services and instances to arrays once.
+                    var servicesArray = services.ToArray();
```
jviau, line 103: `ToList()` is a hair more efficient. (`ToArray()` will basically allocate a list, then resize to the actual count. `ToList()` skips the resize)

src/WebJobs.Script.WebHost/DependencyInjection/ServiceProviderExtensions.cs
```diff
@@ -98,11 +99,16 @@ public static IServiceCollection CreateChildContainer(this IServiceProvider serv
...
+                    var instancesArray = serviceProvider.GetServices(services.Key)?.ToArray() ?? Array.Empty<object>();
```
jviau, line 104: hmm why are we realizing these instances at all? Wonder if we can construct a descriptor which has captured the parent `IServiceProvider` to construct this instance only when requested. IE:

``` CSharp
public static IServiceCollection CloneSingleton(
    this IServiceCollection collection, ServiceDescriptor descriptor, IServiceProvider parent)
{
    return descriptor switch
    {
        { ImplementationInstance: { } instance } => collection.CloneSingleton(descriptor, instance),
        { ImplementationFactory: { } factory } => collection.CloneSingleton(descriptor, _ => descriptor.ImplementationFactory(parent)),
        { ImplementationType: { } type } => collection.CloneSingleton(descriptor, _ => ActivatorUtilities.CreateInstance(parent, type)),
        _ => throw new InvalidOperationException($"Invalid {nameof(ServiceDescriptor)} for service type {descriptor.ServiceType}."),
    };
}
```

@fabiocav do you know why this approach was not taken originally? Is there something I am missing for why the above won't work?

jviau (commented):
I am scratching my head at how this happens. I thought every registered `ServiceDescriptor` would result in a new instance? How are we getting only 5 instances back for 6 descriptors?

src/WebJobs.Script.WebHost/DependencyInjection/ServiceProviderExtensions.cs
fabiocav: Singletons are captured, eagerly instantiated and forwarded to the child container. We could modify the approach here and defer initialization, but that's outside of the scope of this changes.

src/WebJobs.Script.WebHost/DependencyInjection/ServiceProviderExtensions.cs
```diff
@@ -112,25 +118,45 @@ public static IServiceCollection CreateChildContainer(this IServiceProvider serv
...
+            if (servicesArray.Length != instancesArray.Length)
+            {
+                var messageBuilder = new StringBuilder()
+                    .AppendLine($"Mismatch detected for type {serviceType}:")
+                    .AppendLine($"services.Count = {servicesArray.Length}, instances.Count = {instancesArray.Length}")
+                    .Append("Services:\n\t")
+                    .Append(string.Join(",\n\t", servicesArray.Select(s => s.ToString())))
+                    .Append("\nInstances:\n\t")
+                    .AppendLine(string.Join(",\n\t", instancesArray.Select(i => i?.GetType()?.ToString() ?? "null")));
+                Console.WriteLine(messageBuilder.ToString());
```
fabiocav: We shouldn't write this out to console.

src/WebJobs.Script.WebHost/DependencyInjection/ServiceProviderExtensions.cs
fabiocav: Do we have an opportunity to replace this with raw string literals here (https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/raw-string)?

src/WebJobs.Script.WebHost/DependencyInjection/ServiceProviderExtensions.cs
RohitRanjanMS: Remove this.

src/WebJobs.Script.WebHost/DependencyInjection/ServiceProviderExtensions.cs
jviau: I also realized the error in my approach: it will create a 2nd copy of the singleton.

src/WebJobs.Script.WebHost/DependencyInjection/ServiceProviderExtensions.cs
RohitRanjanMS: Keeping it `ToArray()` as there's no additional allocation for `ICollection` but otherwise there will be additional allocation. Thanks, Jacob for explaining it.

src/WebJobs.Script.WebHost/DependencyInjection/ServiceProviderExtensions.cs
RohitRanjanMS: Removed `Console.WriteLine`

src/WebJobs.Script.WebHost/DependencyInjection/ServiceProviderExtensions.cs
RohitRanjanMS: Removed stringbuilder and replaced it with string literals.

@surgupta-msft , This log will primarily help identify the service responsible for the exception. The fix will likely be required in the extension being loaded, possibly because the extension isn't surfacing the exception to the host. 
I also consulted Fabio regarding the message content, and he has no concerns about it.

jviau:
> resolves https://github.com/Azure/azure-functions-host/issues/10780

Should we remove this as this is adding logs to investigate this further, and doesn't outright resolve the issue?

RohitRanjanMS:
> > resolves #10780
> 
> Should we remove this as this is adding logs to investigate this further, and doesn't outright resolve the issue?

I have updated the issue and set the expectation. 
`Enhance the logging to pinpoint the specific service causing the exception. The fix will likely need to be made in the extension where those services are initialized, rather than in the host.`

jviau (approved):

surgupta-msft (approved):

PR #10750, opened 2025-01-16
Linux Consumption metrics publisher for Legion

vivekjilla:
This PR pulls in the new metrics component for Linux Consumption in Legion environments
This is exactly  same as: https://github.com/Azure/azure-functions-host/pull/9741, except for few changes which I highlighted as Qs in comments below.

resolves #issue_for_this_pr

### Pull request checklist

### Additional information

Additional PR information


src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -335,6 +335,41 @@ private static bool IsLinuxConsumptionOnLegion(this IEnvironment environment)
...
         }
 
+        /// <summary>
+        /// Returns a value indicating whether the app is V1 Linux Consumption running on Legion.
+        /// </summary>
+        /// <param name="environment">The environment to verify.</param>
+        /// <returns><see cref="true"/> if the app is V1 Linux Consumption running on Legion; otherwise, <see cref="false"/>.</returns>
+        public static bool IsV1LinuxConsumptionOnLegion(this IEnvironment environment)
+        {
+            return IsLinuxConsumptionOnLegion(environment); //&& environment.WebsiteSkuIsDynamic();
```
vivekjilla: Need to figure out how to distinguish between cv2 vs cv1 in legion based on environment variables.

src/WebJobs.Script.WebHost/Metrics/LinuxContainerLegionMetricsPublisher.cs
```diff
@@ -0,0 +1,239 @@
...
+        private Timer _metricsPublisherTimer;
+        private bool _initialized = false;
+
+        public LinuxContainerLegionMetricsPublisher(IEnvironment environment, IOptionsMonitor<StandbyOptions> standbyOptions, ILogger<LinuxContainerLegionMetricsPublisher> logger, IFileSystem fileSystem, ILinuxConsumptionMetricsTracker metricsTracker, int? metricsPublishIntervalMS = null)
+        {
+            _standbyOptions = standbyOptions ?? throw new ArgumentNullException(nameof(standbyOptions));
+            _environment = environment ?? throw new ArgumentNullException(nameof(environment));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            _metricsTracker = metricsTracker ?? throw new ArgumentNullException(nameof(metricsTracker));
+            //_metricsLogger = metricsLogger ?? throw new ArgumentNullException(nameof(metricsLogger));
```
vivekjilla: Removed the metricsLogger dependency here (as metricsLogger depends on metricsPublisher and is causing cyclic dependency). The original code: https://github.com/Azure/azure-functions-host/pull/9741/files#diff-b40529fc8edfa035050463add05ab73052020dd74d615cdd5e939b860e739d5b. 

Seems like it was being used to log the OnMetricsDiagnosticEvent. So we need to figure out a way to add it back and refactor it out to remove the cyclic dependency. @mathewc any suggestions here.

src/WebJobs.Script.WebHost/WebJobs.Script.WebHost.csproj
```diff
@@ -58,6 +58,7 @@
     <PackageReference Include="Microsoft.AspNetCore.Mvc.NewtonsoftJson" Version="8.0.1" />
     <PackageReference Include="Microsoft.Azure.AppService.Middleware.Functions" Version="1.5.5" />
     <PackageReference Include="Microsoft.Azure.AppService.Proxy.Client" Version="2.3.20240307.67" />
+    <PackageReference Include="Microsoft.Azure.Functions.Platform.Metrics.LinuxConsumption" Version="1.0.4" />
```
vivekjilla, line 66: We need to setup official build pipeline to publish this package. Currently this one was manually uploaded as per: https://github.com/Azure/azure-functions-host/pull/9741#discussion_r1425862249

src/WebJobs.Script.WebHost/Metrics/LinuxContainerLegionMetricsPublisher.cs
```diff
@@ -0,0 +1,239 @@
...
+            _containerName = _environment.GetEnvironmentVariable(EnvironmentSettingNames.ContainerName);
+            _metricPublishInterval = TimeSpan.FromMilliseconds(metricsPublishIntervalMS ?? 30 * 1000);
```
mathewc: @anjanaNandagopal Had asked a question in the original PR that we should address here. She asked "curious, why is this 30 seconds instead of 5 seconds?"

mathewc: Answer: 30s is what is used in [LinuxContainerMetricsPublisher](https://github.com/Azure/azure-functions-host/blob/cfa63f6c4e149363fc3792608b6c3581e3501fda/src/WebJobs.Script.WebHost/Metrics/LinuxContainerMetricsPublisher.cs#L36) so I just retained that interval

src/WebJobs.Script.WebHost/Metrics/LinuxContainerLegionMetricsPublisher.cs
mathewc: Ah, I see. One way to break this cycle is to employ the [Service Locator](https://deviq.com/antipatterns/service-locator) antipattern as we've done in other places in the code to resolve these, e.g. [here](https://github.com/Azure/azure-functions-host/blob/cfa63f6c4e149363fc3792608b6c3581e3501fda/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs#L321).

So you could add an IScriptHostManager ctor param then use it to resolve the metrics logger. You'd define a helper property MetricsLogger that would do the service resolution.

@fabiocav in case he has another idea :)

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
```diff
@@ -280,7 +286,15 @@ private static void AddLinuxContainerServices(this IServiceCollection services)
             services.AddSingleton<IMetricsPublisher>(s =>
             {
                 var environment = s.GetService<IEnvironment>();
-                if (environment.IsFlexConsumptionSku())
+                if (environment.IsV1LinuxConsumptionOnLegion())
+                {
+                    var logger = s.GetService<ILogger<LinuxContainerLegionMetricsPublisher>>();
+                    var metricsTracker = s.GetService<ILinuxConsumptionMetricsTracker>();
+                    var standbyOptions = s.GetService<IOptionsMonitor<StandbyOptions>>();
+                    //var metricsLogger = s.GetService<IMetricsLogger>();
```
mathewc: TODO: Once we fix the cyclic ref update this code

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
```diff
@@ -280,7 +286,15 @@ private static void AddLinuxContainerServices(this IServiceCollection services)
             services.AddSingleton<IMetricsPublisher>(s =>
             {
                 var environment = s.GetService<IEnvironment>();
-                if (environment.IsFlexConsumptionSku())
+                if (environment.IsV1LinuxConsumptionOnLegion())
```
mathewc: I think we should rename this env helper to IsLinuxConsumptionOnLegion, which matches IsLinuxConsumptionOnAtlas

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
mathewc: There should be a unique environmental value/flag that we're specifying for Cv1 that isn't present for Flex apps. Perhaps the SKU value? @balag0 what would you recommend we use to distinguish here? We need this check to return True for Cv1 on Legion and False for Flex apps.

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -335,6 +335,41 @@ private static bool IsLinuxConsumptionOnLegion(this IEnvironment environment)
...
+        }
+
+        /// <summary>
+        /// Checks both WEBSITE_SKU and WEBSITE_SKU_NAME and returns true IFF one is
+        /// set to "Dynamic".
+        /// </summary>
+        /// <param name="environment">The environment to check.</param>
+        /// <returns><see cref="true"/> if the sku is Dynamic; otherwise, <see cref="false"/>.</returns>
+        public static bool WebsiteSkuIsDynamic(this IEnvironment environment)
```
mathewc, line 353: TBD if we need this - when I did the first PR this was a work in progress and we were thinking we could check the sku in this way to determine

test/WebJobs.Script.Tests/Extensions/EnvironmentExtensionsTests.cs
```diff
@@ -273,6 +273,33 @@ public void IsManagedAppEnvironment_ReturnsExpectedResult(bool isManagedAppEnvir
...
+        [InlineData(false, "", "", false)]
+        [InlineData(false, ScriptConstants.DynamicSku, null, false)]
+        [InlineData(false, null, ScriptConstants.DynamicSku, false)]
+        [InlineData(true, null, null, false)]
+        [InlineData(true, "", "", false)]
+        [InlineData(true, ScriptConstants.FlexConsumptionSku, null, false)]
+        [InlineData(true, null, ScriptConstants.DynamicSku, true)]
+        [InlineData(true, ScriptConstants.DynamicSku, null, true)]
+        [Trait(TestTraits.Group, TestTraits.LinuxConsumptionMetricsTests)]
+        public void IsV1LinuxConsumptionOnLegion_ReturnsExpectedResult(bool isLinuxConsumptionOnLegion, string websiteSku, string websiteSkuName, bool expectedValue)
```
mathewc, line 287: After rename, fix this to IsLinuxConsumptionOnLegion_ReturnsExpectedResult

src/WebJobs.Script.WebHost/Metrics/LinuxContainerLegionMetricsPublisher.cs
jainharsh98: Changed as per suggestion.

NuGet.config
```diff
@@ -8,5 +8,6 @@
     <add key="AzureFunctionsRelease" value="https://azfunc.pkgs.visualstudio.com/e6a70c92-4128-439f-8012-382fe78d6396/_packaging/AzureFunctionsRelease/nuget/v3/index.json" />
     <add key="AzureFunctionsPreRelease" value="https://azfunc.pkgs.visualstudio.com/e6a70c92-4128-439f-8012-382fe78d6396/_packaging/AzureFunctionsPreRelease/nuget/v3/index.json" />
     <add key="AzureFunctionsTempStaging" value="https://azfunc.pkgs.visualstudio.com/e6a70c92-4128-439f-8012-382fe78d6396/_packaging/AzureFunctionsTempStaging/nuget/v3/index.json" />
+    <add key="AzureFunctions" value="https://pkgs.dev.azure.com/azfunc/e6a70c92-4128-439f-8012-382fe78d6396/_packaging/AzureFunctions/nuget/v3/index.json" />
```
jviau: What package required adding this feed? I would prefer we use an existing feed if possible.

src/WebJobs.Script.WebHost/Metrics/LegionMetricsFileManager.cs
```diff
@@ -0,0 +1,109 @@
...
+            }
+
+            // we're at or over limit
+            // delete enough files that we have space to write a new one
+            int numToDelete = files.Count - _maxFileCount + 1;
+            var filesToDelete = files.Take(numToDelete).ToArray();
+
+            _logger.LogDebug($"Deleting {filesToDelete.Length} metrics file(s).");
+
+            foreach (var file in filesToDelete)
```
jviau: This could benefit from a Parrallel.ForEach - but this is just a code move so no worries.

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
balag0: CV2
LEGION_SERVICE_HOST = 1
WEBSITE_SKU = FlexConsumption

CV1 on Legion
LEGION_SERVICE_HOST = 1
WEBSITE_SKU_NAME or WEBSITE_SKU = Dynamic

src/WebJobs.Script.WebHost/Metrics/LegionMetricsFileManager.cs
```diff
@@ -0,0 +1,109 @@
...
+
+                string metricsContent = JsonConvert.SerializeObject(metrics);
+                _logger.PublishingMetrics(metricsContent);
+
+                if (metricsPublishEnabled)
+                {
+                    fileName = $"{Guid.NewGuid().ToString().ToLower()}.json";
+                    string filePath = Path.Combine(MetricsFilePath, fileName);
+
+                    using (var streamWriter = _fileSystem.File.CreateText(filePath))
```
jviau: Nit: can use concise using statement:

``` CSharp
using var streamWriter = _fileSystem.File.CreateText(filePath);
await streamWriter.WriteAsync(metricsContext);
```

src/WebJobs.Script.WebHost/Metrics/LinuxContainerLegionMetricsPublisher.cs
```diff
@@ -0,0 +1,251 @@
...
+        {
+            _standbyOptions = standbyOptions ?? throw new ArgumentNullException(nameof(standbyOptions));
+            _environment = environment ?? throw new ArgumentNullException(nameof(environment));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            _metricsTracker = metricsTracker ?? throw new ArgumentNullException(nameof(metricsTracker));
+            if (scriptHostManager == null || !Utility.TryGetHostService<IMetricsLogger>(scriptHostManager, out _metricsLogger))
+            {
+                throw new InvalidOperationException("Unable to get IMetricsLogger Service.");
+            }
+            _containerName = _environment.GetEnvironmentVariable(EnvironmentSettingNames.ContainerName);
```
jviau: Is there no options object with container name on it? Or can we create one? You can combine all different env vars you are retrieving onto it (such as metrics file path).

src/WebJobs.Script.WebHost/Metrics/LinuxContainerLegionMetricsPublisher.cs
```diff
@@ -0,0 +1,251 @@
...
+            _metricPublishInterval = TimeSpan.FromMilliseconds(metricsPublishIntervalMS ?? 30 * 1000);
```
jviau: Recommend putting this on the options object as well if moving to that pattern.

src/WebJobs.Script.WebHost/Metrics/LinuxContainerLegionMetricsPublisher.cs
```diff
@@ -0,0 +1,251 @@
...
+
+            // Default this to 15 minutes worth of files
+            int maxFileCount = 15 * (int)Math.Ceiling(1.0 * 60 / _metricPublishInterval.TotalSeconds);
```
jviau, line 50: The comment says "default", which implies configurability. But I don't see any configurability - is that missing? Or should the comment be updated?

src/WebJobs.Script.WebHost/Metrics/LinuxContainerLegionMetricsPublisher.cs
```diff
@@ -0,0 +1,251 @@
...
+            _metricsFileManager = new LegionMetricsFileManager(metricsFilePath, fileSystem, logger, maxFileCount);
+
+            _processMonitorTimer = new Timer(OnProcessMonitorTimer, null, Timeout.Infinite, Timeout.Infinite);
+            _metricsPublisherTimer = new Timer(OnFunctionMetricsPublishTimer, null, Timeout.Infinite, Timeout.Infinite);
+
+            _metricsTracker.OnDiagnosticEvent += OnMetricsDiagnosticEvent;
+
+            if (_standbyOptions.CurrentValue.InStandbyMode)
+            {
+                _standbyOptionsOnChangeSubscription = _standbyOptions.OnChange(o => OnStandbyOptionsChange());
```
jviau:
```suggestion
                _standbyOptionsOnChangeSubscription = _standbyOptions.OnChange(o => OnStandbyOptionsChange(o));
```

Might as well pass in options to the method as it is available.

src/WebJobs.Script.WebHost/Metrics/LinuxContainerLegionMetricsPublisher.cs
```diff
@@ -0,0 +1,251 @@
...
+
+        public void Start()
+        {
+            Initialize();
+
+            // start the timers by setting the due time
+            SetTimerInterval(_processMonitorTimer, _timerStartDelay);
+            SetTimerInterval(_metricsPublisherTimer, _metricPublishInterval);
+
+            _logger.LogInformation(string.Format("Starting metrics publisher for container : {0}.", _containerName));
```
jviau: Don't use string format or interpolation in log statements:

```suggestion
            _logger.LogInformation("Starting metrics publisher for container: '{containerName}'.", _containerName);
```

src/WebJobs.Script.WebHost/Metrics/LinuxContainerLegionMetricsPublisher.cs
```diff
@@ -0,0 +1,251 @@
...
+
+            _metricsTracker.AddMemoryActivity(memoryActivity);
+        }
+
+        private async void OnFunctionMetricsPublishTimer(object state)
+        {
+            await OnPublishMetrics();
+        }
+
+        internal async Task OnPublishMetrics()
```
jviau:
```suggestion
        internal async Task OnPublishMetricsAsync()
```

NuGet.config
jainharsh98: This feed was added to test it with the package Microsoft.Azure.Functions.Platform.Metrics.LinuxConsumption for internal consumption. Please refer to https://github.com/Azure/azure-functions-host/pull/10750#discussion_r1933261486. Is it okay to push it in the tempStaging or PreRelease feed?

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
jainharsh98: Done

NuGet.config
jviau: Yes you can put it in one of the other feeds.

src/WebJobs.Script.WebHost/Metrics/LinuxContainerLegionMetricsPublisher.cs
jainharsh98: Done

src/WebJobs.Script.WebHost/Metrics/LinuxContainerLegionMetricsPublisher.cs
jainharsh98: Done

src/WebJobs.Script.WebHost/Metrics/LinuxContainerLegionMetricsPublisher.cs
jainharsh98: Done

src/WebJobs.Script.WebHost/Metrics/LinuxContainerLegionMetricsPublisher.cs
jainharsh98: Removed.

src/WebJobs.Script.WebHost/Metrics/LinuxContainerLegionMetricsPublisher.cs
jainharsh98: Done

src/WebJobs.Script.WebHost/Metrics/LinuxContainerLegionMetricsPublisher.cs
jainharsh98: Added.

src/WebJobs.Script.WebHost/Metrics/LegionMetricsFileManager.cs
jainharsh98: Done.

src/WebJobs.Script.WebHost/Metrics/LegionMetricsFileManager.cs
jainharsh98: Done.

test/WebJobs.Script.Tests/Extensions/EnvironmentExtensionsTests.cs
jainharsh98: Done.

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
jainharsh98: Kept here as per Bala's suggestion.

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
jainharsh98: Done.

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
jainharsh98: Changed as per suggestion.

src/WebJobs.Script.WebHost/Metrics/LinuxContainerLegionMetricsPublisher.cs
```diff
@@ -0,0 +1,251 @@
...
+            _processMonitorTimer?.Dispose();
+            _processMonitorTimer = null;
+
+            _metricsPublisherTimer?.Dispose();
+            _metricsPublisherTimer = null;
+
+            _metricsTracker.OnDiagnosticEvent -= OnMetricsDiagnosticEvent;
+        }
+
+        internal class Metrics
```
liliankasem: Why would this not include all of the metrics that the other publisher does?

https://github.com/Azure/azure-functions-host/blob/ed038bed405924d4736b5448025c8a518a0ad48f/src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs#L346-L395

Would it not make more sense to extract `Metrics` out to it's own class and utilize that in all metric publishers?

jainharsh98: The metrics are specific to the sku - particularly Flex consumption metrics are different from Linux Consumption which is sort of fixed. Extracting out to a common class would not be very beneficial as the current overlap is very little.

liliankasem: It is currently little, but we are planning to eventually utilize these metrics else where, and not just in Flex. It would make more sense to have one `Metrics` class that consolidated the host metrics captured, rather than ever publisher having it's own version. Our goal is to standardize across skus, and not to further fragment.

src/WebJobs.Script.WebHost/WebJobs.Script.WebHost.csproj
jainharsh98: As per offline discussion we are going ahead with a manual upload of the the package in the AzureFunctionsTempStaging feed.

NuGet.config
jainharsh98: We have uploaded the package to AzureFunctionsTempStaging feed.

src/WebJobs.Script.WebHost/Configuration/LinuxConsumptionLegionMetricsPublisherOptions.cs
```diff
@@ -0,0 +1,21 @@
...
+    public class LinuxConsumptionLegionMetricsPublisherOptions
+    {
+        internal const int DefaultMetricsPublishIntervalMS = 30 * 1000;
+
+        public LinuxConsumptionLegionMetricsPublisherOptions()
+        {
+            MetricsPublishIntervalMS = DefaultMetricsPublishIntervalMS;
+        }
+
+        public int MetricsPublishIntervalMS { get; set; }
```
kshyju, line 15:
```suggestion
        public int MetricsPublishIntervalMs { get; set; }
```

src/WebJobs.Script.WebHost/Configuration/LinuxConsumptionLegionMetricsPublisherOptions.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Configuration
+{
+    public class LinuxConsumptionLegionMetricsPublisherOptions
+    {
+        internal const int DefaultMetricsPublishIntervalMS = 30 * 1000;
```
kshyju, line 8:
```suggestion
        internal const int DefaultMetricsPublishIntervalMs = 30 * 1000;
```

src/WebJobs.Script.WebHost/Configuration/LinuxConsumptionLegionMetricsPublisherOptions.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Configuration
+{
+    public class LinuxConsumptionLegionMetricsPublisherOptions
```
kshyju: Should this class be `public`, or can it be `internal` instead?

kshyju: Mark the class `sealed`

src/WebJobs.Script.WebHost/Configuration/LinuxConsumptionLegionMetricsPublisherOptionsSetup.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Configuration
+{
+    public class LinuxConsumptionLegionMetricsPublisherOptionsSetup : IConfigureOptions<LinuxConsumptionLegionMetricsPublisherOptions>
+    {
+        private IEnvironment _environment;
```
kshyju:
```suggestion
        private readonly IEnvironment _environment;
```

src/WebJobs.Script.WebHost/Metrics/LegionMetricsFileManager.cs
```diff
@@ -0,0 +1,143 @@
...
+using System.IO.Abstractions;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Script.Diagnostics.Extensions;
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Metrics
+{
+    public class LegionMetricsFileManager
```
kshyju: Can this be `internal sealed class` ?

src/WebJobs.Script.WebHost/Metrics/LegionMetricsFileManager.cs
```diff
@@ -0,0 +1,143 @@
...
+                }
+            }
+            catch (Exception ex) when (!ex.IsFatal())
+            {
+                // TODO: consider using a retry strategy here
+                _logger.LogError(ex, $"Error writing metrics file '{fileName}'.");
+            }
+        }
+
+        public async Task PublishMetricsAsync(object metrics)
```
kshyju: I see we have another method `PublsihMetricsAsync` which takes a `PublishMetrics` type as parameter. What is the usecase where we are not able to use that and need to use this `object` parameter one?

src/WebJobs.Script.WebHost/Metrics/LegionMetricsFileManager.cs
```diff
@@ -0,0 +1,143 @@
...
+                }
+            });
+
+            files = metricsDirectoryInfo.GetFiles().OrderBy(p => p.CreationTime).ToList();
+
+            // return true if we have space for a new file
+            return files.Count < _maxFileCount;
+        }
+
+        public async Task PublsihMetricsAsync(PublishMetrics metrics)
```
kshyju: typo

src/WebJobs.Script.WebHost/Metrics/LinuxContainerLegionMetricsPublisher.cs
```diff
@@ -0,0 +1,232 @@
...
+using System.Threading.Tasks;
+using Microsoft.Azure.Functions.Platform.Metrics.LinuxConsumption;
+using Microsoft.Azure.WebJobs.Script.Diagnostics;
+using Microsoft.Azure.WebJobs.Script.WebHost.Configuration;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Metrics
+{
+    public class LinuxContainerLegionMetricsPublisher : IMetricsPublisher, IDisposable
```
kshyju: mark the class `sealed`

src/WebJobs.Script.WebHost/Metrics/PublishMetrics.cs
```diff
@@ -0,0 +1,67 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Metrics
+{
+    public class PublishMetrics
```
kshyju: 1. mark as `sealed`. 
2. Can this class be `internal` ?

src/WebJobs.Script.WebHost/Metrics/LinuxContainerLegionMetricsPublisher.cs
jainharsh98: Added. Please take a look.

src/WebJobs.Script.WebHost/Metrics/LegionMetricsFileManager.cs
jainharsh98: As per the comment https://github.com/Azure/azure-functions-host/pull/10750#discussion_r1960822768 we are trying to take out metrics from being internal classes in the respective publishers to a public class available outside as well. The earlier method was a relic from that. Cleaned it up.

src/WebJobs.Script.WebHost/Metrics/LegionMetricsFileManager.cs
jainharsh98: Corrected.

src/WebJobs.Script.WebHost/Metrics/PublishMetrics.cs
jainharsh98: Marked as sealed. As per the comment https://github.com/Azure/azure-functions-host/pull/10750#discussion_r1960822768 we are trying to take out metrics from being internal classes in the respective publishers to a public class available outside as well.

src/WebJobs.Script.WebHost/Configuration/LinuxConsumptionLegionMetricsPublisherOptions.cs
jainharsh98: This follows the same pattern as the rest of the functions host. Leaving unchanged.

src/WebJobs.Script.WebHost/Metrics/LegionMetricsFileManager.cs
jainharsh98: Updated. Thanks

src/WebJobs.Script.WebHost/Configuration/LinuxConsumptionLegionMetricsPublisherOptionsSetup.cs
jainharsh98: Updated.

src/WebJobs.Script.WebHost/Configuration/LinuxConsumptionLegionMetricsPublisherOptions.cs
jainharsh98: This class has to be public in order to be configured in the WebHostServiceCollectionExtensions. Sealed the class.

src/WebJobs.Script.WebHost/Metrics/LinuxContainerLegionMetricsPublisher.cs
jainharsh98: Sealed the class.

src/WebJobs.Script.WebHost/Configuration/LinuxConsumptionLegionMetricsPublisherOptions.cs
jainharsh98: This follows the same pattern as the rest of the functions host. Leaving unchanged.

src/WebJobs.Script.WebHost/Metrics/PublishMetrics.cs
fabiocav: Can you please expand on the following?
> we are trying to take out metrics from being internal classes in the respective publishers to a public class available outside as well.

What is the effort? Available outside to what? What do you expect would be consuming those types?

src/WebJobs.Script.WebHost/Metrics/PublishMetrics.cs
```diff
@@ -0,0 +1,67 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Metrics
+{
+    public sealed class PublishMetrics
```
liliankasem, line 8: I don't love this name. I think HostMetrics makes the most sense but that class name is already taken here: 

https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Metrics/HostMetrics.cs

Some ideas:

HostExecutionMetrics
HostActivityMetrics 
HostRuntimeMetrics
HostTelemetryData

@fabiocav @mathewc @jviau thoughts?

jainharsh98: The reason for the naming is to distinguish it from the other Metrics extracted for the Host while keeping it generic enough. Any suggestions on alternate naming would be really helpful.

src/WebJobs.Script.WebHost/Metrics/PublishMetrics.cs
jainharsh98: I think my comment above caused some confusion. There is no ongoing effort to make the metrics available outside that I am aware of. Rather from the earlier changes where we were trying to have separate internal classes for each of the metrics publisher - We changed to having just one consolidated metrics class that can be used and shared across multiple skus in the future. I am assuming from the comment https://github.com/Azure/azure-functions-host/pull/10750#discussion_r1960822768 that there could be use cases for the class outside the host for other skus in the future. Please correct me if that is not the case.

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -317,7 +316,7 @@ public static bool IsManagedAppEnvironment(this IEnvironment environment)
         /// <returns><see cref="true"/> if running in a Linux Consumption App Service app; otherwise, false.</returns>
         public static bool IsAnyLinuxConsumption(this IEnvironment environment)
         {
-            return (environment.IsLinuxConsumptionOnAtlas() || environment.IsFlexConsumptionSku()) && !environment.IsManagedAppEnvironment();
+            return (environment.IsLinuxConsumptionOnAtlas() || environment.IsLinuxConsumptionOnLegion()) && !environment.IsManagedAppEnvironment();
```
vivekjilla, line 319: Why are we not checking for Flex Consumption SKU here?

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -230,15 +230,14 @@ public static bool IsConsumptionSku(this IEnvironment environment)
         /// <returns><see cref="true"/> if running in the FlexConsumption Sku, false otherwise.</returns>
         public static bool IsFlexConsumptionSku(this IEnvironment environment)
         {
-            string value = environment.GetEnvironmentVariable(AzureWebsiteSku);
-            if (string.Equals(value, ScriptConstants.FlexConsumptionSku, StringComparison.OrdinalIgnoreCase))
+            //This will be true for both CV1 and CV2
```
vivekjilla, line 233: Is it safe to change this order:
Previously we used to return `true` when it is `ConsumptionOnlegion` but `WebsiteSku` env is not set. 
Now, we will start returning `false` for the same scenario.

jainharsh98: @balag0 any suggestions on the above?

src/WebJobs.Script.WebHost/Metrics/PublishMetrics.cs
jviau: Agree with @liliankasem, we should use a name which is more specific. `Publish` is vague - is this class publishing the metrics, is it a metrics payload? What metrics does it represent?

I propose `HostMetricData`

liliankasem: HostMetricData works for me

mathewc: I had previously made this metrics class an internal class to LinuxContainerLegionMetricsPublisher in my original PR [here](https://github.com/Azure/azure-functions-host/pull/9741/files#diff-b40529fc8edfa035050463add05ab73052020dd74d615cdd5e939b860e739d5bR229) because it defines a file format that is only used by LinuxContainerLegionMetricsPublisher only. Basically the class defines the file format used when writing the metrics and isn't intended to be shared externally, since inadvertent changes to the class would break the file format. The pattern is similar to what we're doing for FlexConsumptionMetricsPublisher [here](https://github.com/Azure/azure-functions-host/blob/a6aac436248aecbf48b7746d984ce4aefa592d94/src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs#L346).

To avoid confusion and accidental file format breaks, I really think we should just move this back to where it was, and name it Metrics as it was before. Each of FlexConsumptionMetricsPublisher and LinuxContainerLegionMetricsPublisher should have their own internal/private Metrics class that encodes their file formats, as in my original PR https://github.com/Azure/azure-functions-host/pull/9741.

@vivekjilla I synced with Lilian/Jacob offline and they agree based on my explanation that the existing internal classes are best.

src/WebJobs.Script.WebHost/Configuration/LinuxConsumptionLegionMetricsPublisherOptions.cs
mathewc: Yeah, agree that we don't need to change this - looks like we're using MS elsewhere.

src/WebJobs.Script.WebHost/Configuration/LinuxConsumptionLegionMetricsPublisherOptions.cs
mathewc: Yeah, agree that we don't need to change this - looks like we're using MS elsewhere.

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
```diff
@@ -112,10 +113,10 @@ internal async Task OnPublishMetrics(DateTime now)
 
                 // we've been accumulating function activity for the entire period
                 // publish this activity and reset
-                Metrics metrics = null;
+                PublishMetrics metrics = null;
                 lock (_lock)
                 {
-                    metrics = new Metrics
+                    metrics = new PublishMetrics
```
mathewc, line 119: Please revert back to having the existing Metrics class scoped to just this class

src/WebJobs.Script.WebHost/Metrics/PublishMetrics.cs
jviau: @mathewc that sounds good to me. If this schema is specific to the publisher, it should be named/located to clarify as such.

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
```diff
@@ -342,57 +270,5 @@ public void Dispose()
             _standbyOptionsOnChangeSubscription?.Dispose();
             _standbyOptionsOnChangeSubscription = null;
         }
-
-        internal class Metrics
```
mathewc, line 346: Please revert back to having the existing Metrics class scoped to just this class

src/WebJobs.Script.WebHost/Metrics/LinuxContainerLegionMetricsPublisher.cs
mathewc: The previous comment "// Default this to 15 minutes worth of files" that I had in place served to explain the calculation I feel. How about "Set this to 15 minutes worth of files"? Or "Configure this to ..."

We're passing this value into the LegionMetricsFileManager ctor below

src/WebJobs.Script.WebHost/WebJobs.Script.WebHost.csproj
mathewc: Yes. We can then address this as a separate issue after this initial PR is merged.

src/WebJobs.Script.WebHost/Metrics/PublishMetrics.cs
liliankasem: Thanks for the context on this, I agree and okay to revert.

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
mathewc: Great question - didn't you make this change? The preexisting code was checking for both LinuxConsumption and Flex. Since we're adding yet another flavor (LinuxConsumptionOnLegion) to honor the intent of this method, we should check for all 3 flavors.

The name of the method IsAnyLinuxConsumption is somewhat loose, but that decision was already made and you shouldn't worry about renaming it now. E.g. Flex isn't really LinuxConsumption - so the name of the method is already a bit confusing. Again not your problem in this PR :)

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -327,14 +326,49 @@ public static bool IsLinuxConsumptionOnAtlas(this IEnvironment environment)
                    string.IsNullOrEmpty(environment.GetEnvironmentVariable(LegionServiceHost));
         }
 
-        private static bool IsLinuxConsumptionOnLegion(this IEnvironment environment)
+        private static bool IsConsumptionOnLegion(this IEnvironment environment)
```
mathewc, line 329: Agree with the rename on this, since it returns true for both Linux Consumption and Flex Consumption SKUs running on Legion.

I think you should retain a similar comment on the method to the one I added in my original PR "Returns a value indicating whether the app is either Linux Consumption or Flex Consumption running on Legion ."

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
mathewc: So @balag0  can you confirm that this check is the best check for us to be performing at this point? I believe it is. Checking that the SKU is Dynamic rather than FlexConsumption seems to be the right differentiator.

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
mathewc: These changes look risky to me. Instead can't you do the following, which is just a small modification to what we have already?

```
public static bool IsFlexConsumptionSku(this IEnvironment environment)
{
    string value = environment.GetEnvironmentVariable(AzureWebsiteSku);
    if (string.Equals(value, ScriptConstants.FlexConsumptionSku, StringComparison.OrdinalIgnoreCase))
    {
        return true;
    }

    return environment.IsLinuxConsumptionOnLegion() && !environment.WebsiteSkuIsDynamic();
}
```

test/WebJobs.Script.Tests/Extensions/EnvironmentExtensionsTests.cs
mathewc: Looks like the name still isn't correct. The method name is IsLinuxConsumptionOnLegion, so the test should be named IsLinuxConsumptionOnLegion_ReturnsExpectedResult

mathewc:
Closing this PR in favor of https://github.com/Azure/azure-functions-host/pull/10932 which is where we'll finalize things.

PR #10764, opened 2025-01-23
Allow snake case for binding names

### LINKED ISSUE #9917: Timer Trigger arg_name cannot use snake case
### Investigative information

```
func --version
4.0.5530
```

###### Provide the steps required to reproduce the problem:

* Create a new Azure function 

```
func new --template "Timer Trigger" --name another_trigger
```

* Change the generated code from

```python
@app.timer_trigger(schedule="0 */5 * * * *", arg_name="myTimer", run_on_startup=True,
              use_monitor=False) 
def another_trigger(myTimer: func.TimerRequest) -> None:
```

to

```python
@app.timer_trigger(schedule="0 */5 * * * *", arg_name="my_timer", run_on_startup=True,
              use_monitor=False) 
def another_trigger(my_timer: func.TimerRequest) -> None:
```

* Run the function, which results in an error of the form:

```
The 'another_trigger' function is in error: The binding name my_timer is invalid. Please assign a valid name to the binding.
```

### Expected behaviour

It should be possible to use a binding name that uses snake case, since this is the normal convention for Python programs.

### Actual behaviour

```
The 'another_trigger' function is in error: The binding name my_timer is invalid. Please assign a valid name to the binding.
```

### Known workarounds

Remove the `_` character from the binding and variable name.

### Contents of the requirements.txt file:

```txt
azure-functions==1.18.0
```

satvu:

resolves #9917. Allows for binding names to use snake case.

Related documentation issue: #10781

### Pull request checklist

### Additional information

Additional tests were run on the Python side for v1 and v2 [here](https://github.com/Azure/azure-functions-python-worker/pull/1633/files#diff-4c7ce59db8e50eb27c1f0ba7fc7b4f62e7414e949bbc39a97b3425d6c39cafd7) as well.

Additional sign-offs were acquired offline from other language worker owners (Powershell, Java, Node).


test/WebJobs.Script.Tests/UtilityTests.cs
```diff
@@ -531,7 +531,7 @@ public void ValidateFunctionName_DoesNotThrowOnValidName(string functionName)
         [Theory]
         [InlineData(null)]
         [InlineData("")]
-        [InlineData("_binding")]
+        [InlineData("__binding__")]
```
satvu: Per discussion with python worker team, we want to disallow names with double underscore.

test/WebJobs.Script.Tests/UtilityTests.cs
```diff
@@ -550,6 +550,9 @@ public void ValidateBinding_InvalidName_Throws(string bindingName)
         }
 
         [Theory]
+        [InlineData("_binding")]
```
satvu, line 555: Per discussion with python worker team, these new cases should be allowed.

test/WebJobs.Script.Tests/Description/FunctionDescriptorProviderTests.cs
```diff
@@ -203,54 +203,6 @@ public async Task CreateTriggerParameter_WithNoBindingMatch_ThrowsExpectedExcept
             Assert.Contains("someInvalidTrigger", ex.Message);
         }
 
-        [Theory]
```
satvu, line 206: This was testing a wrapper (see [here](https://github.com/Azure/azure-functions-host/blob/e6b3d16b104d535cbb2519e9c5c2c6318246612f/src/WebJobs.Script/Description/FunctionDescriptorProvider.cs#L215)) that directly called `Utility.ValidateBinding(binding)` using the same test cases that the unit tests for the `Utility` method had. Seemed repetitive.

Also considering cleaning up `FunctionDescriptorProvider` to use `Utility.ValidateBinding` directly instead of with the wrapper method linked above.

src/WebJobs.Script/Host/HostFunctionMetadataProvider.cs
```diff
@@ -116,14 +116,6 @@ private FunctionMetadata ReadFunctionMetadata(string functionDirectory, IFileSys
             }
         }
 
-        internal void ValidateName(string name, bool isProxy = false)
```
satvu, line 119: Cleanup request from @mathewc during code review of binding names - this was not being used as there was a previous refactor to directly call `Utility.IsValidFunctionName(name)`.

test/WebJobs.Script.Tests/WorkerFunctionMetadataProviderTests.cs
```diff
@@ -85,7 +85,7 @@ public void ValidateBindings_NoTriggerBinding_Throws()
         [Theory]
         [InlineData(null)]
         [InlineData("")]
-        [InlineData("_binding")]
+        [InlineData("__binding__")]
```
mathewc, line 88: Your regex disallows __ anywhere, so perhaps you should add cases:

__binding
binding__binding
binding__

src/WebJobs.Script/Utility.cs
```diff
@@ -57,7 +57,7 @@ public static class Utility
 #endif
 
         private static readonly Regex FunctionNameValidationRegex = new Regex(@"^[a-z][a-z0-9_\-]{0,127}$(?<!^host$)", RegexOptions.IgnoreCase | RegexOptions.CultureInvariant);
-        private static readonly Regex BindingNameValidationRegex = new Regex(string.Format("^([a-zA-Z][a-zA-Z0-9]{{0,127}}|{0})$", Regex.Escape(ScriptConstants.SystemReturnParameterBindingName)));
+        private static readonly Regex BindingNameValidationRegex = new Regex(string.Format("^(?!.*__)([a-zA-Z_][a-zA-Z0-9_]{{0,127}}|{0})$", Regex.Escape(ScriptConstants.SystemReturnParameterBindingName)));
```
mathewc: Curious what the `|{0}` part is giving us? Seems like `^(?!.*__)([a-zA-Z_][a-zA-Z0-9_]{0,127})$` is logically equivalent?

test/WebJobs.Script.Tests/Description/FunctionDescriptorProviderTests.cs
mathewc: I think these tests should remain. They're testing the virtual FunctionDescriptorProvider.ValidateBinding method impl which right now is basically just Utility.ValidateBinding, but it does also have some additional logic. And in the future more might be added that we need to test.

Is there a test in this class for the "Bindings must specify a Type." error? That's really the custom bit beyond the utility method that I'd expect this class to be testing.

test/WebJobs.Script.Tests/UtilityTests.cs
mathewc: Your regex disallows __ anywhere, so perhaps you should add cases:

`__binding`
`bind__ing`
`binding__`

test/WebJobs.Script.Tests/UtilityTests.cs
```diff
@@ -550,6 +550,9 @@ public void ValidateBinding_InvalidName_Throws(string bindingName)
...
+        [InlineData("binding_name")]
+        [InlineData("_")]
```
mathewc, line 561: So a parameter named "_" is valid C#, but what's the actual Python scenario for this? People want to do it?

mathewc:
Please link to the relevant work items for this, and also fill out the template section above with all the details, etc.

test/WebJobs.Script.Tests/UtilityTests.cs
hallvictoria: Yes, `_` is a valid naming convention for python. It means an unused / throwaway variable -- we've received requests for this for triggers like timers, where the arg might not be used in the function

hallvictoria (approved):

test/WebJobs.Script.Tests/Description/FunctionDescriptorProviderTests.cs
satvu: Yes, there's a test for the error as well. Will revert this then.

src/WebJobs.Script/Utility.cs
satvu: The `|{0}` part was there previously, and it allows us to use the `$return` binding name. The check is "match the constraints or match argument 0", and in our case, argument 0 is the next parameter of `String.Format()`, which is `Regex.Escape(ScriptConstants.SystemReturnParameterBindingName)`. See [here](https://learn.microsoft.com/en-us/dotnet/api/system.string.format?view=net-9.0#system-string-format(system-string-system-object-system-object-system-object)).

Removing that part makes this test case fail: `[InlineData(ScriptConstants.SystemReturnParameterBindingName)]` (the constant resolves to `$return`).

src/WebJobs.Script/Utility.cs
```diff
@@ -591,7 +590,7 @@ public static void ValidateBinding(BindingMetadata bindingMetadata)
         {
             if (bindingMetadata.Name == null || !BindingNameValidationRegex.IsMatch(bindingMetadata.Name))
             {
-                throw new ArgumentException($"The binding name {bindingMetadata.Name} is invalid. Please assign a valid name to the binding.");
+                throw new ArgumentException($"The binding name {bindingMetadata.Name} is invalid. Please assign a valid name to the binding. See https://aka.ms/azure-functions-binding-name-rules for more details.");
```
satvu, line 593: Fwlink has already been created. Documentation will go live once the host release for this is complete.

mathewc (approved):

liliankasem (approved):

src/WebJobs.Script/Utility.cs
fabiocav: Do we have an issue tracking that? The link would need to be updated to point to the right anchor, right?

fabiocav (approved):

src/WebJobs.Script/Utility.cs
satvu: https://github.com/Azure/azure-functions-host/issues/10781 - yes, this link will be updated to point to the right anchor as soon as we publish the docs officially. Working with @mattchenderson and @ggailey777 on this on the side.

fabiocav (approved):

liliankasem (approved):

PR #10767, opened 2025-01-24
Fix sync trigger error when AzureWebjobsStorage is not set in ManagedApp Environment

ishank12:
… ManagedAppEnvironment for Hybrid Logic Apps

Issue : https://github.com/Azure/azure-functions-host/issues/10686

resolves #issue_for_this_pr

### Pull request checklist

### Additional information

Additional PR information


fabiocav (commented):
@ishank12 can you please make sure the release notes are updated to capture this change?

Can you also please have some validation for the new behavior to ensure this does not regress in the future?

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -135,7 +135,7 @@ public async Task<SyncTriggersResult> TrySyncTriggersAsync(bool isBackgroundSync
 
                 bool shouldSyncTriggers = true;
                 string newHash = null;
-                if (isBackgroundSync && !_environment.IsKubernetesManagedHosting())
+                if (isBackgroundSync && !_environment.IsKubernetesManagedHosting() && hashBlobClient != null)
```
jviau: Want to confirm if this behavior change is intentional or not: we will now always call `SetTriggersAsync` if there is no `hashBlobClient`. Is that intentional?

ishank12: yes its intentional. 
hashBlobClient can be null in case of for hybrid logic app, which runs in managed app environment, where AzureWebJobsStorage is not set and that use customers local SMB filesystem

mathewc: My read on these changes is that previously for Kubernetes Managed Hosting, we'd force a background sync even if we didn't have a Blob Client to check the hash. With your changes, you're now expanding this to Managed App Environments.

The skipping of background syncs is an optimization - we don't want each instance that comes up to issue a sync, so we use a hash to determine if anything has actually changed.

Just want to confirm that you're OK with this. What is the endpoint for these non-optimized SetTriggers calls in the Managed App Environment case? Looks like the destination is given by the env var FUNCTIONS_API_SERVER, so it's not the Antares Front End?

mathewc: Seems like you should also update this logic to `if (isBackgroundSync && hashBlobClient != null && !_environment.IsAnyKubernetesEnvironment())`, the goal being to skip the hash check across the board for Kubernetes, even if we DO have a blob client? Also reorder the _hashBlobClient check to be before the env var checks for performance.

That makes this and the above clause symmetric. If they're not symmetric, it's hard to reason why.

test/WebJobs.Script.Tests.Integration/Management/FunctionsSyncManagerTests.cs
```diff
@@ -311,6 +311,21 @@ public async Task TrySyncTriggers_LocalEnvironment_ReturnsFalse()
             }
         }
 
+        [Fact]
+        public async Task TrySyncTriggers_ManagedAppEnv_withNo_AzureWebJobsStorage_ReturnsTrue()
```
mathewc: nit: fix the camel casing here

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
ishank12: is the goal to skip the hash check for all Kubernetes even if hash blob client is set?
Frankly I wanted to skip the hash check for cases only when hash blob client is not set.
This is because, this hash blob can be set as well as not set in managed environment. We will miss this optimization when hash blob client is set. 

So the second clause should correctly be like. 
`if (isBackgroundSync && hashBlobClient != null)`

And the first clause is like that because we dont want to short circuit in those environments where hash blob client can be null.

let me know your opinion, either way is fine to me.
1. check hash when hash blob is set.
2. skip the hash check for all Kubernetes even if hash blob client is set

mathewc: I asked some questions above that I don't think you answered? Basically just confirming that in these new scenarios you're OK with the fact that every single pod that comes up will result in a Sync call going to the FUNCTIONS_API_SERVER endpoint

mathewc: Yes, let's go with your number 1 then and simplify to `if (isBackgroundSync && hashBlobClient != null)`. My main goal here was to simplify these difficult to understand checks. Thanks.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -135,7 +135,7 @@ public async Task<SyncTriggersResult> TrySyncTriggersAsync(bool isBackgroundSync
 
                 bool shouldSyncTriggers = true;
                 string newHash = null;
-                if (isBackgroundSync && !_environment.IsKubernetesManagedHosting())
+                if (isBackgroundSync && hashBlobClient != null && !_environment.IsKubernetesManagedHosting())
```
mathewc: As noted in the other open discussion thread, I agree with you that changing this to `if (isBackgroundSync && hashBlobClient != null)` would be best.

test/WebJobs.Script.Tests.Integration/Management/FunctionsSyncManagerTests.cs
```diff
@@ -311,6 +311,21 @@ public async Task TrySyncTriggers_LocalEnvironment_ReturnsFalse()
...
+        {
+            using (var env = new TestScopedEnvironmentVariable(_vars))
+            {
+                _mockEnvironment.Setup(p => p.GetEnvironmentVariable(EnvironmentSettingNames.ManagedEnvironment)).Returns("true");
+                _mockEnvironment.Setup(p => p.GetEnvironmentVariable("FUNCTIONS_API_SERVER")).Returns("https://appname.azurewebsites.net");
+                _mockEnvironment.Setup(p => p.GetEnvironmentVariable("CONTAINER_APP_NAME")).Returns("appname");
+                _mockEnvironment.Setup(p => p.GetEnvironmentVariable("CONTAINER_APP_NAMESPACE")).Returns("appns");
+                _mockEnvironment.Setup(p => p.GetEnvironmentVariable("CONTAINER_APP_REVISION")).Returns("appname--r1");
+                var result = await _functionsSyncManager.TrySyncTriggersAsync(isBackgroundSync: true);
+                Assert.True(result.Success);
```
mathewc, line 327: This assert doesn't seem sufficient. E.g. If everything is short circuited [here](https://github.com/Azure/azure-functions-host/blob/46194b9893def8e31a76a69e6dfbfeac5530834f/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs#L122), your test would still pass, right? You need to verify that the sync actually happened, as other tests are doing, e.g. ensure _mockHttpClient sends the expected request.

test/WebJobs.Script.Tests.Integration/Management/FunctionsSyncManagerTests.cs
```diff
@@ -311,6 +311,21 @@ public async Task TrySyncTriggers_LocalEnvironment_ReturnsFalse()
             }
         }
 
+        [Fact]
+        public async Task TrySyncTriggers_ManagedAppEnv_WithNo_AzureWebJobsStorage_ReturnsTrue()
```
mathewc, line 315: Can you add a test for the existing Kubernetes Managed Hosting case as well? That way we have test coverage for all these Kubernetes scenarios.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
ishank12: Since there is no blob storage in this case, we cannot set the hash of last synced trigger.  So currently its fine that every single pod would initiate a sync trigger in managed env. 
However in my previous PR, I tried to make this optimized by storing the last hash on a filesystem which is mounted by each pod. But I think that change still need some more discussion on how to proceed with that change for all the kubernetes environments and the types of storage that can be used.

here was my previous attempt of optimised sync trigger for our case. https://github.com/Azure/azure-functions-host/pull/10335

test/WebJobs.Script.Tests.Integration/Management/FunctionsSyncManagerTests.cs
```diff
@@ -311,6 +311,57 @@ public async Task TrySyncTriggers_LocalEnvironment_ReturnsFalse()
...
+                }
+                else
+                {
+                    VerifyResultWithCacheOff(durableVersion: "V1");
+                }
+            }
+        }
+
+        [Theory]
+        [InlineData(false)]
```
ishank12: Test is failing for cacheEnabled True in kubernetes managed environment while reading the host.json file.
Not sure why, 
I don't have much insights for this. 
Hence adding test only for cacheEnabled False.
should be checked separately.

fabiocav: > should checked separately.

Can you please clarify what you mean?

ishank12: @fabiocav @mathewc 
The issue is not with the managed app environment that I was working for.  Its only with the kubernetes hosting environment.

while adding tests for kubernetes environment, I found a bug.
there is an exception in one path. 
it fails in kubernetes hosting when arm cache is enabled.

its failing here in this block
[azure-functions-host/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs at dev · Azure/azure-functions-host](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs#L345-L354) 
and exactly while reading the host.json file. here https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Extensions/FileUtility.cs#L111
because the fileStream cannot be read.
 
Although the host.json exists.

So I am not adding test for that as that will block the build.
This should be fixed separatetly and then test can be added for that

mathewc: I see. For these tests, you don't need to test cache enabled/true false at all. I'd just remove the Theory and make them Fact tests, with the cache validation hard coded to whatever. The point of your new tests is just to verify that the sync triggers call is made.

ishank12: hey @mathewc , @fabiocav 
fixed the tests to use cache false, both tests are pass now.

mathewc (approved):

kshyju (approved):

PR #10773, opened 2025-01-28
Run benchmarking in dedicated agents (application and load)

kshyju:
With this change, the application and load generators run on separate agents. To share the application agent's IP address with the load generator agent, we use Azure Table Storage. Job output variables are not suitable as they become available only after the job finishes, whereas we need the job (agent) to remain running.

Sample pipeline run: 
1. https://azfunc.visualstudio.com/internal/_build/results?buildId=199693&view=results

### DIFF
--- a/eng/ci/templates/official/jobs/run-benchmarks.yml
+++ b/eng/ci/templates/official/jobs/run-benchmarks.yml
@@ -59,10 +59,10 @@ jobs:
     displayName: Install Microsoft.Crank.Agent tool
 
   - script: dotnet tool install -g Microsoft.Crank.Controller --version "0.2.0-*"
-    displayName: Install Microsoft.Crank.Controller
+    displayName: Install Microsoft.Crank.Controller tool
 
   - pwsh: Start-Process powershell -ArgumentList '-NoExit', '-Command', 'crank-agent'
-    displayName: 'Start crank-agent'
+    displayName: Start crank-agent
 
   - task: CopyFiles@2
     displayName: Copy benchmark apps to temp location
@@ -90,6 +90,9 @@ jobs:
       scriptType: pscore
       scriptLocation: inlineScript
       inlineScript: |
+        $maxAttempts = 60
+        $attempt = 0
+
         do {
           $Entity = az storage entity show `
             --account-name $(StorageAccount) `
@@ -106,6 +109,12 @@ jobs:
           } else {
             Write-Host "Entity not found. Retrying in 30 seconds..."
             Start-Sleep -Seconds 30
+            $attempt++
+          }
+
+          if ($attempt -ge $maxAttempts) {
+            Write-Host "Maximum attempts reached. Exiting..."
+            exit 1
           }
         } while (-not $Entity)
 
@@ -158,4 +167,3 @@ jobs:
       crank compare "$(baselineBenchmarkResultFilePath)" "$(benchmarkResultsJsonPath)"
     condition: and(succeeded(), ne(variables['baselineBenchmarkResultFilePath'], ''))
     displayName: 'Compare Results with Baseline'
-    
--- a/eng/ci/templates/official/jobs/run-benchmarks.yml
+++ b/eng/ci/templates/official/jobs/run-benchmarks.yml
@@ -120,7 +120,7 @@ jobs:
         }
 
         Write-Host "Running command: $command"
-        Invoke-Expression $command -ErrorAction Stop
+        Invoke-Expression $command
     displayName: 'Run Benchmark'
 
   # Retrieve function host logs
--- a/eng/ci/templates/official/jobs/run-benchmarks.yml
+++ b/eng/ci/templates/official/jobs/run-benchmarks.yml
@@ -9,22 +9,25 @@ parameters:
 - name: additionalCrankArgs
   type: string
   default: ''
+- name: agentName
+  type: string
 
 jobs:
 - job: ${{ parameters.functionAppName }}
 
   variables:
+    agentName: ${{ parameters.agentName }}
     runDescription: ${{ parameters.description }}
     functionApp: ${{ parameters.functionAppName }}
     benchmarkArtifactName: benchmark_results_$(functionApp)
     functionAppOutputPath: $(Build.ArtifactStagingDirectory)/FunctionApps/$(functionApp)
     benchmarkResultsJsonPath: $(Build.ArtifactStagingDirectory)/BenchmarkResults/$(Build.BuildNumber)_$(functionApp).json
     functionsWorkerRuntime: 'dotnet-isolated'
-    crankAgentUrl: "http://localhost:5010" # Default crank agent URL.
     configFilePath: "./eng/perf/http.benchmarks.yml"
     hostLocation: "./../../"
     baselineBenchmarkResultFilePath: ''
     baselineBenchmarkResultsDownloadDir: $(Pipeline.Workspace)/BenchmarkBaselineResult
+    appAgentIpAddress: ''
 
   templateContext:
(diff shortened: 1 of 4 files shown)


eng/ci/templates/official/jobs/setup-benchmark-agents.yml
```diff
@@ -0,0 +1,103 @@
...
+  - task: PowerShell@2
+    displayName: Get Agent IP Address
+    inputs:
+      targetType: 'inline'
+      script: |
+        $ipAddress = (Test-Connection -ComputerName (hostname) -Count 1).IPv4Address.IPAddressToString
+        Write-Host "IP Address: $ipAddress"
+        Write-Host "##vso[task.setvariable variable=agentIp]$ipAddress"
+
+  - task: AzureCLI@2
```
kshyju, line 61: We need to share the agent info(IP address) to the other job(agent) so that It can communicate with this agent (communication from load to application). Job output variable won't work as it is available only after the job completes. We want the job(agent) running.

jviau (approved):

eng/ci/templates/official/jobs/run-benchmarks.yml
```diff
@@ -9,22 +9,25 @@ parameters:
...
   type: string
   default: ''
+- name: agentName
+  type: string
 
 jobs:
 - job: ${{ parameters.functionAppName }}
 
   variables:
+    agentName: ${{ parameters.agentName }}
```
jviau, line 19: Just so you are aware, you don't need to assign parameters to variables. You can use `${{ parameters.agentName }}` anywhere in this file - including in a pwsh script.

It is fine to use variables if you want to. A small nuance to them though: params are readonly. Variables can be updated between steps. Probably not important here, but something to consider.

eng/ci/templates/official/jobs/run-benchmarks.yml
```diff
@@ -55,6 +58,9 @@ jobs:
   - script: dotnet tool install -g Microsoft.Crank.Agent --version "0.2.0-*"
     displayName: Install Microsoft.Crank.Agent tool
 
+  - script: dotnet tool install -g Microsoft.Crank.Controller --version "0.2.0-*"
+    displayName: Install Microsoft.Crank.Controller
```
jviau: small nit: you end the previous step with `tool` but not this one. It's completely irrelevant in the end, just a slight eye twitch from me that's all.

eng/ci/templates/official/jobs/run-benchmarks.yml
```diff
@@ -77,11 +83,35 @@ jobs:
...
+            --partition-key $(Build.BuildNumber) `
+            --row-key $(agentName) `
+            --select AgentIPAddress
+
+          if ($Entity) {
+            $AgentIPAddress = ($Entity | ConvertFrom-Json).AgentIPAddress
+            Write-Host "##vso[task.setvariable variable=appAgentIpAddress]$AgentIPAddress"
+            break
+          } else {
+            Write-Host "Entity not found. Retrying in 30 seconds..."
```
jviau, line 110: may want to add a timeout and failure to this step so it doesn't spin for an hour.

eng/ci/templates/official/jobs/run-benchmarks.yml
```diff
@@ -128,3 +158,4 @@ jobs:
       crank compare "$(baselineBenchmarkResultFilePath)" "$(benchmarkResultsJsonPath)"
     condition: and(succeeded(), ne(variables['baselineBenchmarkResultFilePath'], ''))
     displayName: 'Compare Results with Baseline'
+
```
jviau: is there whitespace here or something? Wonder why it is giving the incorrect end-of-file icon.

eng/ci/templates/official/jobs/run-benchmarks.yml
```diff
@@ -55,6 +58,9 @@ jobs:
   - script: dotnet tool install -g Microsoft.Crank.Agent --version "0.2.0-*"
```
jviau, line 58: Is the crank agent still needed on this job? Or is only the controller needed?

eng/ci/templates/official/jobs/setup-benchmark-agents.yml
jviau: Interesting - there isn't a more pipelines-native way to do this?

eng/ci/templates/official/jobs/setup-benchmark-agents.yml
```diff
@@ -0,0 +1,103 @@
...
+        az storage entity insert `
+          --auth-mode login `
+          --account-name $(StorageAccount) `
+          --table-name $(BenchmarkAgentInfoTableName) `
+          --entity PartitionKey=$(Build.BuildNumber) RowKey=$(agentName) AgentName=$(agentName) AgentIPAddress=$(agentIp)
+
+  - pwsh: |
+      $url = "http://localhost:5010/jobs/all"
+      Write-Host "Calling $url to check benchmark job status"
+      while ($true) {
```
jviau: maybe a timeout for this step also. The build itself will timeout at 60minutes, but shorter may be better.

eng/ci/templates/official/jobs/setup-benchmark-agents.yml
```diff
@@ -0,0 +1,103 @@
...
+              Write-Host "Found at least 1 job with 'state' = 'Deleted'. Exiting task."
+              exit 0
+          }
+
+          Write-Host "No completed jobs found. Polling again in 30 seconds..."
+          Start-Sleep -Seconds 30
+      }
+    displayName: Wait for job completion
+
+  - task: AzureCLI@2
```
jviau, line 99: Consider adding `condition: always()` to this step so it will cleanup even on failure of previous steps.

jviau (approved):

eng/ci/templates/official/jobs/run-benchmarks.yml
kshyju: yes. We are using this agent for load generation as well. (If we want to create an agent for that, we could, but I chose this approach for now). So agent is needed.

eng/ci/templates/official/jobs/run-benchmarks.yml
kshyju: Good call. Added a retry exit mechanism after 30 minutes.

eng/ci/templates/official/jobs/setup-benchmark-agents.yml
kshyju: There is pipeline output variables. But that is not available before the job is finished. In our case, we need to keep the job running (so agent is up to take crank jobs).

eng/ci/templates/official/jobs/setup-benchmark-agents.yml
kshyju: Added `condition: always()`

eng/ci/templates/official/jobs/run-benchmarks.yml
kshyju: there was whitespace, which I removed in latest iteration.

eng/ci/templates/official/jobs/setup-benchmark-agents.yml
kshyju: Added retry exit mechanism.

eng/ci/templates/official/jobs/run-benchmarks.yml
kshyju: Fixed to be consistent with other displayName.

satvu (approved):

PR #10776, opened 2025-01-29
Cold Start Optimization

### LINKED ISSUE #10453: Cold Start Optimization
When the Host gets initialized in placeholder mode, all needed assemblies get JIT compiled during placeholder mode so we will not pay the cost for JIT during Specialization. We want to exercise as much of the code path as we can during placeholder mode to avoid JIT during specialization.
Right now, none of the OTEL specific code runs during placeholder mode and OTEL will have a huge JIT time during specialization. 

Cold start is a critical parameter that we closely monitor. Optimizing cold start times will enhance the user experience and encourage OTel adoption among customers.

RohitRanjanMS:
When the Host gets initialized in placeholder mode, all needed assemblies get JIT compiled during placeholder mode so we will not pay the cost for JIT during Specialization. We want to exercise as much of the code path as we can during placeholder mode to avoid JIT during specialization.

resolves #10453

### Pull request checklist

ditional PR information


src/WebJobs.Script/Diagnostics/OpenTelemetry/OpenTelemetryConfigurationExtensions.cs
```diff
@@ -42,82 +62,96 @@ internal static void ConfigureOpenTelemetry(this ILoggingBuilder loggingBuilder,
...
+                               o.FilterHttpRequestMessage = static (httpRequestMessage) =>
+                               {
+                                   if (httpRequestMessage.RequestUri?.AbsoluteUri is not { Length: > 0 } uri)
+                                   {
+                                       return false;
+                                   }
+
+                                   foreach (string substring in ExcludedRequestSubstrings)
+                                   {
+                                       if (uri.IndexOf(substring, StringComparison.Ordinal) >= 0)
```
jviau: I think this will need to be made more robust:

1. This looks like it might not be cheap to perform on a hotpath like this
2. These substrings may have a chance of false-positives.

Possible solutions:

1. Is there something besides URI on the `HttpRequestMessage` we can use to identify these calls?
2. Can we _add_ something to the `HttpRequestMessage` to identify these calls? Such as adding a value to `.Options` to indicate this call is to be filtered out.
3. Can we use an AsyncLocal to suppress Http client telemetry collection?

src/WebJobs.Script/Diagnostics/OpenTelemetry/OpenTelemetryConfigurationExtensions.cs
```diff
@@ -42,82 +62,96 @@ internal static void ConfigureOpenTelemetry(this ILoggingBuilder loggingBuilder,
...
-            {
-                // Log all warnings and above by default.
-                loggingBuilder.Services.AddHostedService(service => new OpenTelemetryEventListenerService(EventLevel.Warning));
-            }
+        private static ILoggingBuilder AddDefaultOpenTelemetryFilters(this ILoggingBuilder loggingBuilder)
+        {
+            return loggingBuilder
+                // These are messages piped back to the host from the worker - we don't handle these anymore if the worker has OpenTelemetry enabled.
+                // Instead, we expect the user's own code to be logging these where they want them to go.
+                .AddFilter<OpenTelemetryLoggerProvider>("Function.*", _ => !ScriptHost.WorkerOpenTelemetryEnabled)
```
jviau, line 103: Curious why we explicitly care about `WorkerOptenTelemetryEnabled`? What if worker is using app-insights. Don't ask _why_ an app would use OTel in host and AppInsights in the worker, but what if they did?

What I am getting at is, should there be a general flag of worker is managing its own telemetry, and agnostic to exactly how it is doing so?

src/WebJobs.Script/Diagnostics/OpenTelemetry/TelemetryMode.cs
```diff
@@ -7,6 +7,7 @@ internal enum TelemetryMode
     {
         None = 0, // or Default
         ApplicationInsights = 1,
-        OpenTelemetry = 2
+        OpenTelemetry = 2,
+        Placeholder = 3
```
jviau: small nit: I would put placeholder as `1`. Just feels right...

src/WebJobs.Script/Diagnostics/OpenTelemetry/OpenTelemetryConfigurationExtensions.cs
RohitRanjanMS: Removed HTTPClient instrumentation. I don't see much value in having HTTPClient instrumentation on the host. It captures all internal calls, including those made to AppInsights ingestion endpoint (Az Mon Exporter) and by Azure SDKs like Storage, which already provide distributed traces. 
Removing this would eliminate the need for filtering out unnecessary traces.

src/WebJobs.Script/Diagnostics/OpenTelemetry/OpenTelemetryConfigurationExtensions.cs
RohitRanjanMS: I agree with you. We should give customers the flexibility to configure it without requiring them to use AppInsights or OTel. I'll create a work item to discuss this further.

jviau (approved):

soninaren (approved):

PR #10777, opened 2025-01-29
Filtering out non-essential telemetry

### LINKED ISSUE #10712: Reduce telemetry
The Functions host communicates with Blob Storage and other external services for internal operations, such as managing locks and performing various background tasks. These interactions result in a significant amount of telemetry data being generated. To improve clarity and focus in the monitoring system, it's essential to minimize this noise by filtering out non-essential telemetry related to such internal processes. 

Image

RohitRanjanMS:
The Functions host communicates with Blob Storage and other external services for internal operations, such as managing locks and performing various background tasks. These interactions result in a significant amount of telemetry data being generated. This pull request filters out non-essential telemetry data to reduce the noise in the traces.

resolves #10712 

image

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script/StorageProvider/StorageClientProvider.cs
+++ b/src/WebJobs.Script/StorageProvider/StorageClientProvider.cs
@@ -76,6 +76,9 @@ namespace Microsoft.Azure.WebJobs.Script
         private TClientOptions CreateClientOptions(IConfiguration configuration)
         {
             var clientOptions = (TClientOptions)_componentFactory.CreateClientOptions(typeof(TClientOptions), null, configuration);
+
+            // Disable distributed tracing by default to reduce the noise in the traces.
+            clientOptions.Diagnostics.IsDistributedTracingEnabled = false;
             return clientOptions;
         }
     }


src/WebJobs.Script/StorageProvider/StorageClientProvider.cs
```diff
@@ -76,6 +76,9 @@ protected bool IsConnectionStringPresent(IConfiguration configuration)
         private TClientOptions CreateClientOptions(IConfiguration configuration)
         {
             var clientOptions = (TClientOptions)_componentFactory.CreateClientOptions(typeof(TClientOptions), null, configuration);
+
+            // Disable distributed tracing by default to reduce the noise in the traces.
+            clientOptions.Diagnostics.IsDistributedTracingEnabled = false;
```
kshyju, line 81: If a customer app relies on storage-related triggers/bindings, would disabling distributed tracing here affect any of their current data usage? Also, should we provide support with the ability to override the bool value, allowing customers to access these logs if needed? or that is not a valid case to support?

jviau: Agreed. I think we need to discuss the impact of this a bit more. Does it need to be configurable? Does it need to be enabled but _filtered_? Do we filter out success, but keep failures?

RohitRanjanMS: Hi @kshyju, this client is specifically intended for internal use. The binding and trigger will use a different client, and these settings will not affect those interactions. I have tested this scenario.

We don't process this telemetry in our ApplicationInsights pipeline, but in OpenTelemetry, it inadvertently gets consumed due to its design. These dependencies don't seem particularly valuable. We've managed without these logs so far.

RohitRanjanMS: @jviau, I missed your comment before resolving it—I'll reopen it to continue the discussion. 
We do receive failure logs if there's an issue accessing the storage account. The only added value I see from these logs is the response time. Let me know if you think we should still retain this telemetry.

jviau: Ah okay this is not for triggers and bindings. Thanks for the clarification.

jviau (approved):

kshyju (approved):

PR #10782, opened 2025-01-30
Add support for running benchmarks in Linux environment.

### LINKED ISSUE #10729: Build & Validate benchmark execution flow on Linux
Validate benchmark execution flow on Linux

kshyju:
resolves #10729

Adding support for running benchmarks on Linux agents. Also, other minor cleanups.

Sample run: https://azfunc.visualstudio.com/internal/_build/results?buildId=200596&view=results

### Pull request checklist

* [x ] My changes **do not** need to be backported to a previous version

### DIFF
--- a/eng/ci/templates/official/jobs/run-benchmarks.yml
+++ b/eng/ci/templates/official/jobs/run-benchmarks.yml
@@ -3,23 +3,28 @@ parameters:
   type: string
 - name: description
   type: string
-- name: storeResultsInDatabase
-  type: boolean
-  default: false
-- name: additionalCrankArgs
+- name: os
   type: string
-  default: ''
-- name: agentName
-  type: string
-
+  default: 'Windows'
 jobs:
-- job: ${{ parameters.functionAppName }}
+- job: ${{ parameters.functionAppName }}_${{ parameters.os }}
 
+  ${{ if eq(parameters.os, 'Linux') }}:
+    pool:
+      name: 1es-pool-azfunc-benchmarking-large
+      image: 1es-ubuntu-22.04-benchmark-runner-vanilla
+      os: linux 
+  ${{ else }}:
+    pool:
+      name: 1es-pool-azfunc-benchmarking-large
+      image: 1es-windows-2022-benchmark-runner-vanilla
+      os: windows
+  
   variables:
-    agentName: ${{ parameters.agentName }}
+    agentId: ${{ parameters.functionAppName }}${{ parameters.os }}
     runDescription: ${{ parameters.description }}
     functionApp: ${{ parameters.functionAppName }}
-    benchmarkArtifactName: benchmark_results_$(functionApp)
+    benchmarkArtifactName: benchmark_results_$(Agent.OS)_$(functionApp)
     functionAppOutputPath: $(Build.ArtifactStagingDirectory)/FunctionApps/$(functionApp)
     benchmarkResultsJsonPath: $(Build.ArtifactStagingDirectory)/BenchmarkResults/$(Build.BuildNumber)_$(functionApp).json
     functionsWorkerRuntime: 'dotnet-isolated'
@@ -27,7 +32,7 @@ jobs:
     hostLocation: "./../../"
     baselineBenchmarkResultFilePath: ''
     baselineBenchmarkResultsDownloadDir: $(Pipeline.Workspace)/BenchmarkBaselineResult
-    appAgentIpAddress: ''
+    appAgentHostName: ''
 
   templateContext:
     inputs:
@@ -46,7 +51,7 @@ jobs:
   steps:
 
   - task: AzureKeyVault@2
-    condition: and(succeeded(), eq('${{ parameters.storeResultsInDatabase }}', true))
+    condition: and(succeeded(), eq(variables['storeBenchmarkResultsInDatabase'], true))
     inputs:
       azureSubscription: Azure-Functions-Host-CI-internal
       KeyVaultName: functions-perf-crank-kv
@@ -62,26 +67,13 @@ jobs:
     displayName: Install Microsoft.Crank.Controller tool
 
   - pwsh: Start-Process powershell -ArgumentList '-NoExit', '-Command', 'crank-agent'
-    displayName: Start crank-agent
+    displayName: Start crank-agent(Windows)
+    condition: eq(variables['Agent.OS'], 'Windows_NT')
 
-  - task: CopyFiles@2
-    displayName: Copy benchmark apps to temp location
-    inputs:
-      SourceFolder: '$(Build.SourcesDirectory)/test/Performance/Apps'
-      Contents: '**/*'
-      TargetFolder: '$(Build.ArtifactStagingDirectory)/PerformanceTestApps'
-      CleanTargetFolder: true
-
-  - task: DotNetCoreCLI@2
-    displayName: Publish $(functionApp) app
-    inputs:
-      command: publish
-      publishWebProjects: false
-      zipAfterPublish: false
-      modifyOutputPath: false
-      projects: '$(Build.ArtifactStagingDirectory)/PerformanceTestApps/$(functionApp)/HelloHttp.csproj'
-      arguments: -c Release -o $(functionAppOutputPath) -f net9.0
-      workingDirectory: $(Build.ArtifactStagingDirectory)/PerformanceTestApps/$(functionApp)
+  - script: |
+      nohup crank-agent &
+    displayName: Start crank-agent(Linux)
+    condition: eq(variables['Agent.OS'], 'Linux')
 
   - task: AzureCLI@2
     displayName: Get Remote Agent IP Address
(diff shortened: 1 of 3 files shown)


eng/ci/templates/official/jobs/run-benchmarks.yml
```diff
@@ -3,31 +3,36 @@ parameters:
   type: string
 - name: description
   type: string
-- name: storeResultsInDatabase
```
kshyju, line 6: inlined these

eng/ci/templates/official/jobs/run-benchmarks.yml
```diff
@@ -62,26 +67,13 @@ jobs:
     displayName: Install Microsoft.Crank.Controller tool
 
   - pwsh: Start-Process powershell -ArgumentList '-NoExit', '-Command', 'crank-agent'
-    displayName: Start crank-agent
+    displayName: Start crank-agent(Windows)
+    condition: eq(variables['Agent.OS'], 'Windows_NT')
 
-  - task: CopyFiles@2
-    displayName: Copy benchmark apps to temp location
-    inputs:
```
kshyju, line 69: These 2 tasks are not needed on this job. Forgot to remove it in previous PR.

eng/ci/templates/official/jobs/setup-benchmark-agents.yml
```diff
@@ -1,29 +1,51 @@
...
-    agentName: ${{ parameters.agentName }}
+    agentId: ${{ parameters.functionAppName }}${{ parameters.os }}
     functionApp: ${{ parameters.functionAppName }}
     functionAppOutputPath: $(Build.ArtifactStagingDirectory)/FunctionApps/$(functionApp)
 
   steps:
 
   - template: /eng/ci/templates/install-dotnet.yml@self
 
+  # Adds the Ubuntu 20.04 security repository to Ubuntu 22.04 and installs libssl1.1, which is unavailable by default.
```
kshyju: Without this libssl version, the publish step fails with a "No usable version of libssl was found" error from the metadata generator.
https://azfunc.visualstudio.com/internal/_build/results?buildId=200558&view=logs&j=ca6baa9e-9efc-5f7e-211f-565f0d3d6d9f&t=1fb2d2a4-6dcd-5ba1-ab80-1d330a38e4b0&l=1518

The `1es-ubuntu-22.04` image includes this version, likely installed as a dependency of another package.

eng/ci/templates/official/jobs/run-benchmarks.yml
```diff
@@ -120,25 +112,30 @@ jobs:
...
+      $url = "http://$(appAgentHostName):5010/jobs/1/output"
       Write-Host "Fetching logs from: $url"
-      $response = Invoke-WebRequest -Uri $url -Method GET -UseBasicParsing
-      Write-Host $response.Content
-    displayName: 'Fetch Function Host Logs'
+      
+      if ($IsWindows) {
+          $response = Invoke-WebRequest -Uri $url -Method GET -UseBasicParsing
+      } else {
+          $response = curl -s $url
```
jviau: Why not `Invoke-WebRequest` on Linux?

eng/ci/templates/official/jobs/run-benchmarks.yml
```diff
@@ -152,14 +149,17 @@ jobs:
...
+      if (Test-Path $baselineDir -PathType Container) {
+          $baselineFile = Get-ChildItem -Path $baselineDir -Filter $fileNamePattern | Select-Object -First 1
+
+          if ($baselineFile) {
+              Write-Host "Found baseline benchmark result file: $($baselineFile.FullName)"
+              Write-Host "##vso[task.setvariable variable=baselineBenchmarkResultFilePath]$($baselineFile.FullName)"
+          }
       }
+
+      Write-Host "No baseline benchmark result file found."
```
jviau: Looks like this will currently print even if baseline file is found above. Need either a condition here or a `return` above to short-circuit.

eng/ci/templates/official/jobs/setup-benchmark-agents.yml
jviau: If we are maintaining the base image, can you add this as part of image prep?

eng/ci/templates/official/jobs/setup-benchmark-agents.yml
```diff
@@ -1,29 +1,51 @@
 parameters:
 - name: functionAppName
   type: string
-- name: agentName
+- name: os
```
jviau, line 4: You can add an `values` section to restrict to linux/windows.

eng/ci/templates/official/jobs/setup-benchmark-agents.yml
```diff
@@ -1,29 +1,51 @@
...
+  - script: |
+      echo "deb http://security.ubuntu.com/ubuntu focal-security main" | sudo tee /etc/apt/sources.list.d/focal-security.list
+      sudo apt update && sudo apt install -y libssl1.1
+    displayName: "Install libssl1.1 on Ubuntu 22.04"
+    condition: eq(variables['Agent.OS'], 'Linux')
+
   - script: dotnet tool install -g Microsoft.Crank.Agent --version "0.2.0-*"
     displayName: Install Microsoft.Crank.Agent tool
 
   - pwsh: Start-Process powershell -ArgumentList '-NoExit', '-Command', 'crank-agent'
```
jviau: Consider using templating to if/else these two steps:

``` yaml
- ${{ if eq(parameters.os, 'Windows') }}:
  - pwsh: ... # windows step

- ${{ else }}:
  - script: ... # linux step
```

The difference between this approach and using `condition` is that this will occur during templating. The resulting yaml will only include the appropriate step for the OS. The `conditon` approach is lazily evaluated at the time of step execution. So, the templating approach is more consistent (no mid-job variable changes will impact the decision!) and also gives a cleaner pipeline view, as there won't be the skipped steps (they won't even show up!).

I always use templating when possible. Only using `condition` when I absolutely must dynamically decide run or not based on a previous step.

eng/ci/templates/official/jobs/run-benchmarks.yml
kshyju: Yea, I didn't realize that would work for both platforms. Fixed in the latest iteration.

eng/ci/templates/official/jobs/setup-benchmark-agents.yml
```diff
@@ -1,29 +1,53 @@
...
   steps:
 
   - template: /eng/ci/templates/install-dotnet.yml@self
 
+  # Adds the Ubuntu 20.04 security repository to Ubuntu 22.04 and installs libssl1.1, which is unavailable by default.
+  - script: |
+      echo "deb http://security.ubuntu.com/ubuntu focal-security main" | sudo tee /etc/apt/sources.list.d/focal-security.list
+      sudo apt update && sudo apt install -y libssl1.1
+    displayName: "Install libssl1.1 on Ubuntu 22.04"
+    condition: eq(variables['Agent.OS'], 'Linux')
```
jviau: This task can also be switched to template format

eng/ci/templates/official/jobs/run-benchmarks.yml
kshyju: Updated it to print more specific error message and only when the file / artifact directory not found.

eng/ci/templates/official/jobs/setup-benchmark-agents.yml
kshyju: Good idea. Added values in both places.

eng/ci/templates/official/jobs/setup-benchmark-agents.yml
kshyju: Good call. Switched to this approach. Was doing similar thing for the pool already.

eng/ci/templates/official/jobs/setup-benchmark-agents.yml
kshyju: Switched in the latest iteration.

eng/ci/templates/official/jobs/setup-benchmark-agents.yml
kshyju: We need to build an artifact package with that code in the central repo so that we can bake it into the image (It takes a list of artifacts).
https://dev.azure.com/devdiv/XlabImageFactory/_wiki/wikis/XlabImageFactory.wiki/6337/Artifacts?anchor=contributing-artifacts. I am waiting on some permissions to contribute to that repo and see they accept it. Will revisit this based on how it goes. Should not be a blocker for this now.

They also have a [provisioning script approach](https://eng.ms/docs/more/developer-guides/cloudtest-user-guide/cloudtest/how-tos/update-provisioning-script), but that gets executed only when the image is provisoned (during the build) . so, it's not any different than having it in the YAML here. Also, we do not want this in all agents. The one where we are doing the `dotnet publish` of the function app is where this is needed.

kshyju:
Build from latest version- https://azfunc.visualstudio.com/internal/_build/results?buildId=200919&view=results

jviau (approved):

satvu (approved):

eng/ci/templates/official/jobs/setup-benchmark-agents.yml
kshyju: Made this an image factory artifact now ([PR](https://devdiv.visualstudio.com/XlabImageFactory/_git/XlabImageFactory/pullrequest/608527)). Opened PR #10849 is removing this pipeline step.

Issue #10560, opened 2024-10-23
Review and update OTel semantic conventions


lmolkova:
I see a few minor discrepancies:

1. `faas.trigger` is set to some Azure Function constant like `eventHubTrigger`

![Image](https://github.com/user-attachments/assets/baf80c3a-6cf7-4979-8a30-58caead38cdc)

Based on semantic conventions, I believe it should be `pubsub` - https://github.com/open-telemetry/semantic-conventions/blob/main/docs/faas/faas-spans.md and the fact that it's an Event Hub is captured in the corresponding `messaging.system`. 

`faas.trigger` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.

| Value  | Description | Stability |
|---|---|---|
| `datasource` | A response to some data source operation such as a database or filesystem read/write | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
| `http` | To provide an answer to an inbound HTTP request | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
| `other` | If none of the others apply | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
| `pubsub` | A function is set to be executed when messages are sent to a messaging system | ![Experimental](https://img.shields.io/badge/-experimental-blue) |
| `timer` | A function is scheduled to be executed regularly | ![Experimental](https://img.shields.io/badge/-experimental-blue) |

2. Very minor: `ai.sdk.prefix = azurefunctions:4.1036.0.0` or `dotnetiso:1.1.0.0` is populated on resources even when I don't use Application Insights SDK and use pure OTel like here https://github.com/lmolkova/testfuncotel. We probably should not set it if AppInsights is not enabled. 

3. `faas.version = 4.1036.0.0` is set on host resources, but it should not be set for Azure functions according to semconv 

   [Azure Functions: Not applicable. Do not set this attribute.](https://github.com/open-telemetry/semantic-conventions/blob/35664be5d835b2f2e8a2e414dea5cef3b04d045a/docs/attributes-registry/faas.md?plain=1#L67)

   Let's fix one side or the other.

4. `faas.execution` is used instead of `faas.invocation_id`. It was right according to OTel schema 1.17 which Functions support, but it was renamed later. Can we update the schema version for OTel case while it's all still experimental?

[...]

PR #10783, opened 2025-01-31
Adding a `http.route` Activity tag for HTTP Triggers when OTel is enabled

stebet:
### Making OTel span names more clear for HTTP triggers and adding the `http.route` tag to the root Activity

This PR Adds the `http.route` as discussed in #10560 and makes the span name more clear. Currently there aren't any test specifically for the OTel tags, but I have tested this locally with an existing function app.

This give us a much clearer root span name like this:
image

Earlier the root span name would have just been `GET`. Now the span has the `HttpTrigger.Route` added to make it more clear which function is being hit as well as the OTel attribute.

Tagging @RohitRanjanMS since he seems to have been involved with most of the OTel PRs so far and @lmolkova for visibility as well.

### Pull request checklist

* [X] Backporting to the `in-proc` branch is not required
    * [X] Otherwise: I've added my notes to `release_notes.md`


src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -1723,6 +1723,12 @@ private void AddAdditionalTraceContext(MapField<string, string> attributes, Scri
             {
                 Activity.Current?.AddTag(ResourceSemanticConventions.FaaSName, context.FunctionMetadata.Name);
                 Activity.Current?.AddTag(ResourceSemanticConventions.FaaSTrigger, OpenTelemetryConstants.ResolveTriggerType(context.FunctionMetadata?.Trigger?.Type));
+                if (context.FunctionMetadata?.Trigger?.Raw != null && context.FunctionMetadata.Trigger.Type == "httpTrigger" && context.FunctionMetadata.Trigger.Raw.TryGetValue("route", StringComparison.OrdinalIgnoreCase, out var value))
+                {
+                    string routeValue = value.ToString();
+                    Activity.Current.DisplayName = $"{Activity.Current.DisplayName} {routeValue}";
```
lmolkova: is `Activity.Current` guaranteed here?
Also, wonder if we can rely on its value being http method (something sets it to `{METHOD}`, but what if it starts to set route too? Or, on older .NET versions it'd likely default to activity name, which would be `HttpRequestIn` or something similar, but not so friendly.

The tl;dr: do we have method name around? can we format the display name that's guaranteed to comply?

stebet: Thanks for taking a look @lmolkova :)

The Host that executes this part of the code for httpTriggers is a .NET 8 Web host, so I'm pretty sure if the activity exists and the trigger is an httpTrigger we can count on it being correct. Also, you can see that the activity originates from the Microsoft.AspNetCore OTel library.

I hope @RohitRanjanMS can correct me if I'm wrong.

I'll add a null check on the Activity.Current, I missed that one :)

RohitRanjanMS: Thanks, @stebet , I will have a look at this today.

lmolkova: right, so what if in the future version Functions would use routing on the worker? Also, if I'm not mistaken, on .NET 8 OTel instrumentations sets ASP.NET Core activity name, but in .NET 9 (or 10, don't remember), it will be the ASP.NET Core itself. Plus some users may enable ASP.NET Core instrumentation and others won't, so it'd be safer to format the whole thing and and not rely on DisplayName matching method.

stebet: End users don't have control of this host process except through host.json config (and otel currently is just an on/off setting) AFAIK. It'll always be at the root since it handles the reverse proxying and routing of incoming function requests.

Also the OTEL host telemetry is in preview ATM and I'm just looking to make the existing spans it puts out a little more readable.

RohitRanjanMS:
Hi @stebet , let me figure out why http.route is null. If I can fix it, it would be ideal.

RohitRanjanMS:
In our middleware, we're using [IRoutingFeature](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.WebHost/Middleware/FunctionInvocationMiddleware.cs#L163), which is causing an issue since the instrumentation relies on [RouteEndpoint](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs#L248) to retrieve the route pattern. As a result, the http.route tag isn't being set, and the display name only shows the HTTP verb. Let me check what can we do here.

RohitRanjanMS:
@stebet , I have created an [issue ](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/2565) in `opentelemetry-dotnet-contrib`.

RohitRanjanMS:
@stebet , we will have to do this in Functions host. Let's continue with this PR.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
RohitRanjanMS: We can add this logic in `EnrichWithHttpResponse`. You can use IRoutingFeature to get the route information.

```
.AddAspNetCoreInstrumentation(o =>
{
    o.EnrichWithHttpResponse = (activity, httpResponse) =>
    {
        var routingFeature = httpResponse.HttpContext.Features.Get<AspNetCore.Routing.IRoutingFeature>();
        // Set the route and display name here
    };
});

stebet:
@RohitRanjanMS Is there anything more you need from me in this change?

RohitRanjanMS:
yes, my recommendation is to use [EnrichWithHttpResponse](https://github.com/Azure/azure-functions-host/pull/10783#discussion_r1987683670).

RohitRanjanMS:
Hi @stebet , I made this change while refactoring the code and adding support for metrics. I needed to include it in this [PR ](https://github.com/Azure/azure-functions-host/pull/10929)to meet the cutoff, and I apologize for any inconvenience this may have caused. I will go ahead and close the PR.

PR #10789, opened 2025-02-03
Skip HostIdValidator checks in placeholder mode

mathewc:
We're seeing cases in production where placeholder specializations are failing with host ID collision errors. Here's how this bug manifests itself:

- In placeholder mode the first time the host ID is accessed during startup code [here](https://github.com/Azure/azure-functions-host/blob/e6b3d16b104d535cbb2519e9c5c2c6318246612f/src/WebJobs.Script/Host/ScriptHostIdProvider.cs#L42) in the HostId provider will have scheduled the HostId validation for 5 seconds later (cold start optimization).
- this validation isn't supposed to run in placeholder mode since it requires a storage connection, and code [here](https://github.com/Azure/azure-functions-host/blob/e6b3d16b104d535cbb2519e9c5c2c6318246612f/src/WebJobs.Script/Host/HostIdValidator.cs#L72) on the validation path is supposed to prevent that. However, because of the coldstart delay, the validation can run **after** specialization with the app environment and storage connection, but the closure has captured the placeholder mode host ID e.g. `functionsv4node18placeholdertemp`.
- Due to race conditions between specialization, host ID access and validation, a blob entry can be written for the placeholder ID pointing to the actual host name (e.g. `myapp.azurewebsites.net`), rather than the placeholder hostname (e.g. `functionsv4node18placeholdertemplatesite.azurewebsites.net`).
- Depending on timing, this can cause the validation check [here](https://github.com/Azure/azure-functions-host/blob/29cf955ede30064df88e4c82b79967347087c2e2/src/WebJobs.Script/Host/HostIdValidator.cs#L105) to fail due to an expected hostname mismatch.

### Pull request checklist

    * [x] Inproc backport: https://github.com/Azure/azure-functions-host/pull/10801

### DIFF
--- a/test/WebJobs.Script.Tests.Integration/Host/StandbyManager/StandbyManagerE2ETests_Windows.cs
+++ b/test/WebJobs.Script.Tests.Integration/Host/StandbyManager/StandbyManagerE2ETests_Windows.cs
@@ -11,12 +11,15 @@ using System.IO;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using Azure.Storage.Blobs;
 using Microsoft.AspNetCore.Hosting;
+using Microsoft.Azure.WebJobs.Host.Storage;
 using Microsoft.Azure.WebJobs.Script.WebHost;
 using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.WebJobs.Script.Tests;
 using Xunit;
+using static Microsoft.Azure.WebJobs.Script.HostIdValidator;
 
 namespace Microsoft.Azure.WebJobs.Script.Tests
 {
@@ -28,6 +31,8 @@ namespace Microsoft.Azure.WebJobs.Script.Tests
 
         public StandbyManagerE2ETests_Windows()
         {
+            Utility.ColdStartDelayMS = 1000;
+
             _settings = new Dictionary<string, string>()
             {
                 { EnvironmentSettingNames.AzureWebsitePlaceholderMode, "1" },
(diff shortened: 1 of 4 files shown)


test/WebJobs.Script.Tests.Integration/Host/StandbyManager/StandbyManagerE2ETests_Windows.cs
```diff
@@ -168,6 +168,75 @@ await TestHelpers.Await(() =>
             Assert.NotSame(GetCachedTimeZoneInfo(), _originalTimeZoneInfoCache);
         }
 
+        [Fact]
+        public async Task StandbyModeE2E_Dotnet_HostIdValidator_DoesNotRunInPlaceholderMode()
```
mathewc, line 172: With the default 5s cold start delay, this test repro-ed the issue 100% of the time. It was resulting in host ID collisions.

test/WebJobs.Script.Tests/ScriptHostIdProviderTests.cs
```diff
@@ -53,10 +54,13 @@ public async Task GetHostIdAsync_WithConfigurationHostId_ReturnsConfigurationHos
         }
 
         [Theory]
-        [InlineData("TEST-FUNCTIONS--", "123123", "test-functions", false)]
-        [InlineData("TEST-FUNCTIONS-XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX", "123123", "test-functions-xxxxxxxxxxxxxxxxx", true)]
-        public async Task GetHostIdAsync_AzureHost_ReturnsExpectedResult(string siteName, string azureWebsiteInstanceId, string expected, bool validationExpected)
+        [InlineData("TEST-FUNCTIONS--", "123123", "test-functions", false, false)]
+        [InlineData("TEST-FUNCTIONS-XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX", "123123", "test-functions-xxxxxxxxxxxxxxxxx", true, false)]
+        [InlineData("functionsv4x86inproc8placeholdertemplatesite", "123123", "functionsv4x86inproc8placeholder", false, true)]
```
mathewc, line 59: Unit test for the same - ensuring that we don't schedule a background check when in placeholder mode

test/WebJobs.Script.Tests.Integration/Host/StandbyManager/StandbyManagerE2ETests_Windows.cs
```diff
@@ -168,6 +173,105 @@ await TestHelpers.Await(() =>
...
```
mathewc, line 177: This test reproduces the issue 100% of the time. I couldn't get a reliable repro since a race condition is involved, so as the comments below state, I resorted to forcing the condition via a direct blob write.

kshyju (approved):
LGTM

PR #10790, opened 2025-02-03
Support release channel settings in extension bundles resolution

### LINKED ISSUE #10778: Support Release Channel in Bundle Resolution
### Description

The Azure Functions host needs to consider the platform's release channel when resolving extension bundles. The release channel will be provided through the environment variable `WEBSITE_PlatformReleaseChannel`. This improvement ensures bundle resolution aligns with platform expectations.

#### Release Channel Definitions
The WEBSITE_PlatformReleaseChannel environment variable will have the following possible values:

- Latest: Behaves as it does today. Resolves to the latest available version of the bundle that meets the customer’s specified reference.
- Standard: Resolves to the version prior to the latest (n-1), if that version is available.
- Extended: Resolves to the version two releases prior to the latest (n-2), if that version is available.
    - However, Functions and Rapid Update should treat Standard and Extended the same, resolving to n-1.

#### Expected Behavior
- When resolving bundles, the host should check the value of WEBSITE_PlatformReleaseChannel and apply the corresponding versioning logic.
- If the variable is not set, default to Latest behavior.

satvu:

resolves #10778

### Pull request checklist


src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -36,6 +35,7 @@ public ExtensionBundleManager(ExtensionBundleOptions options, IEnvironment envir
             _cdnUri = _environment.GetEnvironmentVariable(EnvironmentSettingNames.ExtensionBundleSourceUri) ?? ScriptConstants.ExtensionBundleDefaultSourceUri;
             _options = options ?? throw new ArgumentNullException(nameof(options));
             _configOption = configOption ?? throw new ArgumentNullException(nameof(configOption));
+            _platformReleaseChannel = _environment.GetEnvironmentVariable(EnvironmentSettingNames.AntaresPlatformReleaseChannel) ?? ScriptConstants.LatestPlatformChannelName;
```
satvu: using "Latest" as the default - this matches with previous behavior before considering this setting.

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -263,7 +263,29 @@ internal static string FindBestVersionMatch(VersionRange versionRange, IEnumerab
...
+            if (string.Equals(_platformReleaseChannel, ScriptConstants.StandardPlatformChannelName, StringComparison.Ordinal) || string.Equals(_platformReleaseChannel, ScriptConstants.ExtendedPlatformChannelName, StringComparison.Ordinal))
+            {
+                if (bundleVersions.Count() > 1)
+                {
+                    matchingVersion = bundleVersions.OrderByDescending(version => version.Version).ElementAt(1); // if n is latest, should get version n-1
+                }
+                else
+                {
+                    // keep the latest version, log a warning
+                    _logger.LogInformation("Unable to retrieve the 2nd latest extension bundle version. The latest bundle version, {latestBundleVersion}, will be used instead.", latestBundleVersion);
```
satvu: Log warning or keep as information? Not sure how often this would happen since this is a new feature.

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -250,7 +250,7 @@ private async Task<string> GetLatestMatchingBundleVersionAsync(HttpClient httpCl
             return matchingBundleVersion;
         }
 
-        internal static string FindBestVersionMatch(VersionRange versionRange, IEnumerable<string> versions, string bundleId, FunctionsHostingConfigOptions configOption)
+        internal string FindBestVersionMatch(VersionRange versionRange, IEnumerable<string> versions, string bundleId, FunctionsHostingConfigOptions configOption)
```
satvu, line 254: Should be a safe change - this method is only called in this class and in the tests for this class.

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
soninaren: The wording here `2nd latest extension bundle version` can be confusing. We should try to keep the terms closer to the one used in the feature set itself.

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -263,7 +263,29 @@ internal static string FindBestVersionMatch(VersionRange versionRange, IEnumerab
                 return version;
             }).Where(v => v != null);
 
-            var matchingVersion = bundleVersions.OrderByDescending(version => version.Version).FirstOrDefault();
+            var latestBundleVersion = bundleVersions.OrderByDescending(version => version.Version).FirstOrDefault();
```
liliankasem: Can we move all this new logic to a method? i.e. `ResolvePlatformReleaseChannelVersion` which returns the version

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -263,7 +263,29 @@ internal static string FindBestVersionMatch(VersionRange versionRange, IEnumerab
...
+            var matchingVersion = latestBundleVersion;
+
+            if (string.Equals(_platformReleaseChannel, ScriptConstants.StandardPlatformChannelName, StringComparison.Ordinal) || string.Equals(_platformReleaseChannel, ScriptConstants.ExtendedPlatformChannelName, StringComparison.Ordinal))
```
liliankasem: nit: we could make this easier to follow along with something like this:

```
bool isStandardOrExtendedChannel = string.Equals(_platformReleaseChannel, ScriptConstants.StandardPlatformChannelName, StringComparison.Ordinal) ||
                                   string.Equals(_platformReleaseChannel, ScriptConstants.ExtendedPlatformChannelName, StringComparison.Ordinal);

bool isLatestOrEmptyChannel = string.Equals(_platformReleaseChannel, string.Empty, StringComparison.Ordinal) ||
                              string.Equals(_platformReleaseChannel, ScriptConstants.LatestPlatformChannelName, StringComparison.Ordinal);

```

test/WebJobs.Script.Tests/ExtensionBundle/ExtensionBundleManagerTests.cs
```diff
@@ -407,22 +401,120 @@ public void LimitMaxVersion(string versionRange, string hostConfigVersion, strin
...
+            var manager = GetExtensionBundleManager(options, GetTestAppServiceEnvironment());
+
+            var resolvedVersion = manager.FindBestVersionMatch(range, GetLargeVersionsList(), ScriptConstants.DefaultExtensionBundleId, hostingConfiguration);
+
+            Assert.Equal(expectedVersion, resolvedVersion);
+        }
+
+        [Theory]
+        [InlineData(ScriptConstants.LatestPlatformChannelName, "[4.*, 5.0.0)", "4.2.0", "4.2.0")]
+        [InlineData(ScriptConstants.StandardPlatformChannelName, "[4.*, 5.0.0)", "4.2.0", "4.2.0")]
```
liliankasem, line 414: I'm not sure I understand why standard and extended are not a version behind latest? If latest is 4.2.0, shouldn't standard nd extended by 4.1.0?

test/WebJobs.Script.Tests/ExtensionBundle/ExtensionBundleManagerTests.cs
```diff
@@ -407,22 +401,120 @@ public void LimitMaxVersion(string versionRange, string hostConfigVersion, strin
...
-        private ExtensionBundleManager GetExtensionBundleManager(ExtensionBundleOptions bundleOptions, TestEnvironment environment = null)
+        [Theory]
+        [InlineData(ScriptConstants.ExtendedPlatformChannelName)]
+        [InlineData(ScriptConstants.StandardPlatformChannelName)]
+        public void StandardExtendedReleaseChannel_OneBundleVersion_Handled(string platformReleaseChannelName)
+        {
+            // these release channels take version n-1 (where n is the latest)
+            // if there is only one bundle version, latest should be chosen and information logged
+
+            var versions = new List<string>() { "4.20.0" };
```
liliankasem: Can we test with more than one version in this list?

test/WebJobs.Script.Tests/ExtensionBundle/ExtensionBundleManagerTests.cs
```diff
@@ -407,22 +401,120 @@ public void LimitMaxVersion(string versionRange, string hostConfigVersion, strin
...
+
+            var resolvedVersion = manager.FindBestVersionMatch(range, GetLargeVersionsList(), ScriptConstants.DefaultExtensionBundleId, hostingConfiguration);
+
+            Assert.Equal(expectedVersion, resolvedVersion);
+        }
+
+        [Theory]
+        [InlineData(ScriptConstants.LatestPlatformChannelName, "[4.*, 5.0.0)", "4.2.0", "4.2.0")]
+        [InlineData(ScriptConstants.StandardPlatformChannelName, "[4.*, 5.0.0)", "4.2.0", "4.2.0")]
+        [InlineData(ScriptConstants.ExtendedPlatformChannelName, "[4.*, 5.0.0)", "4.2.0", "4.2.0")]
```
liliankasem: I'm not sure I understand why these don't result in 4.1.0?

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
liliankasem: nit:  move `_cdnUri` down so its between `_configOption` and `_platformReleaseChannel`

test/WebJobs.Script.Tests/ExtensionBundle/ExtensionBundleManagerTests.cs
satvu: The list of available bundles in this test (highest to lowest version) is: [ 4.3.0, 4.2.0, 4.1.0, ... 3.9.1].

The platform channel rules result in 4.2.0 (because latest is 4.3.0 and next is 4.2.0), and the max is also 4.2.0. It resolves to that value. 

There are cases below that have a max of 4.3.0, but resolves to 4.2.0 b/c the channel resolved to a lower value. I'll add a test for the reverse (max is 4.1.0, channel resolves to 4.2.0, so it should be 4.1.0).

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -293,6 +294,38 @@ internal static string FindBestVersionMatch(VersionRange versionRange, IEnumerab
             return matchingVersion?.ToString();
         }
 
+        private NuGetVersion ResolvePlatformReleaseChannelVersion(IEnumerable<NuGetVersion> bundleVersions, NuGetVersion currentMatchingVersion)
+        {
+            bool isStandardOrExtendedChannel = string.Equals(_platformReleaseChannel, ScriptConstants.StandardPlatformChannelName, StringComparison.Ordinal) ||
```
kshyju: This data a case sensitive comparison. Should we do a case insensitive comparison to be safe?

kshyju: > Extended: Resolves to the version two releases prior to the latest (n-2), if that version is available.
However, Functions and Rapid Update should treat Standard and Extended the same, resolving to n-1.

If we are intentionally returning n-1 for both cases, we should add a comment explaining "Why" we are doing what we are doing, for future readers.

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -293,6 +294,38 @@ internal static string FindBestVersionMatch(VersionRange versionRange, IEnumerab
...
+                }
+                else
+                {
+                    // keep the latest version, log a notice
+                    _logger.LogInformation("Unable to apply plaform release channel configuration {platformReleaseChannelName}. Only one matching bundle version is available. {latestBundleVersion} will be used", _platformReleaseChannel, currentMatchingVersion);
+                }
+            }
+            else if (isLatestOrEmptyChannel)
+            {
+                // no - op, latest version should be used.
```
kshyju: Shouldn't you be returning `return bundleVersions.OrderByDescending(version => version.Version).FirstOrDefault();` here?

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -293,6 +294,38 @@ internal static string FindBestVersionMatch(VersionRange versionRange, IEnumerab
...
+            if (isStandardOrExtendedChannel)
+            {
+                if (bundleVersions.Count() > 1)
+                {
+                   return bundleVersions.OrderByDescending(version => version.Version).ElementAt(1); // if n is latest, should get version n-1
+                }
+                else
+                {
+                    // keep the latest version, log a notice
+                    _logger.LogInformation("Unable to apply plaform release channel configuration {platformReleaseChannelName}. Only one matching bundle version is available. {latestBundleVersion} will be used", _platformReleaseChannel, currentMatchingVersion);
```
kshyju: typo in message.

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -293,6 +294,38 @@ internal static string FindBestVersionMatch(VersionRange versionRange, IEnumerab
...
+            bool isLatestOrEmptyChannel = string.Equals(_platformReleaseChannel, string.Empty, StringComparison.Ordinal) ||
+                                          string.Equals(_platformReleaseChannel, ScriptConstants.LatestPlatformChannelName, StringComparison.Ordinal);
+
+            if (isStandardOrExtendedChannel)
+            {
+                if (bundleVersions.Count() > 1)
+                {
+                   return bundleVersions.OrderByDescending(version => version.Version).ElementAt(1); // if n is latest, should get version n-1
+                }
+                else
```
kshyju: nit: can remove the `else {` lines since you are returning above.

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -293,6 +294,38 @@ internal static string FindBestVersionMatch(VersionRange versionRange, IEnumerab
             return matchingVersion?.ToString();
         }
 
+        private NuGetVersion ResolvePlatformReleaseChannelVersion(IEnumerable<NuGetVersion> bundleVersions, NuGetVersion currentMatchingVersion)
+        {
+            bool isStandardOrExtendedChannel = string.Equals(_platformReleaseChannel, ScriptConstants.StandardPlatformChannelName, StringComparison.Ordinal) ||
+                                   string.Equals(_platformReleaseChannel, ScriptConstants.ExtendedPlatformChannelName, StringComparison.Ordinal);
+
+            bool isLatestOrEmptyChannel = string.Equals(_platformReleaseChannel, string.Empty, StringComparison.Ordinal) ||
```
kshyju: Did you consider inlining this code? or moving closer to usage (with slight adjustments to if else)? That way we will execute this string comparison code only when the first one did return anything.

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
kshyju: Maybe we can move line 266 to this method. So, there is a single method which determines what version to use.

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -250,7 +250,7 @@ private async Task<string> GetLatestMatchingBundleVersionAsync(HttpClient httpCl
             return matchingBundleVersion;
         }
 
-        internal static string FindBestVersionMatch(VersionRange versionRange, IEnumerable<string> versions, string bundleId, FunctionsHostingConfigOptions configOption)
+        internal string FindBestVersionMatch(VersionRange versionRange, IEnumerable<string> versions, string bundleId, FunctionsHostingConfigOptions configOption)
         {
             var bundleVersions = versions.Select(p =>
```
jviau, line 256: I think we are going to be enumerating this list multiple times with these added checks. We should realize this to a list and then pass that around instead of `IEnumerable`. Make sure to include the `OrderByDescending` before calling `.ToList()`.

Also can this entire collection LINQ operations can be moved into `ResolvePlatformReleaseChannelVersion`?

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
jviau: Or we could ToUpper or ToLower it

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -263,7 +263,8 @@ internal static string FindBestVersionMatch(VersionRange versionRange, IEnumerab
                 return version;
             }).Where(v => v != null);
 
-            var matchingVersion = bundleVersions.OrderByDescending(version => version.Version).FirstOrDefault();
+            var latestBundleVersion = bundleVersions.OrderByDescending(version => version.Version).FirstOrDefault();
```
jviau: Can we move defaulting to this into `ResolvePlatformReleaseChannelVersion`? so we only fallback to it if it is needed.

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -293,6 +294,38 @@ internal static string FindBestVersionMatch(VersionRange versionRange, IEnumerab
...
 
+        private NuGetVersion ResolvePlatformReleaseChannelVersion(IEnumerable<NuGetVersion> bundleVersions, NuGetVersion currentMatchingVersion)
+        {
+            bool isStandardOrExtendedChannel = string.Equals(_platformReleaseChannel, ScriptConstants.StandardPlatformChannelName, StringComparison.Ordinal) ||
+                                   string.Equals(_platformReleaseChannel, ScriptConstants.ExtendedPlatformChannelName, StringComparison.Ordinal);
+
+            bool isLatestOrEmptyChannel = string.Equals(_platformReleaseChannel, string.Empty, StringComparison.Ordinal) ||
+                                          string.Equals(_platformReleaseChannel, ScriptConstants.LatestPlatformChannelName, StringComparison.Ordinal);
+
+            if (isStandardOrExtendedChannel)
```
jviau: instead of the bool checks, can we break the logic down into re-usable components, and then call the appropriate one given the channel? Can even use a switch statement for that:


```
return _platformReleaseChannel.ToUpperInvariant() switch
{
    "STANDARD" or "EXTENDED" => // call something
    "LATEST" or "" => // call something
    _ => // call something
}
```

liliankasem: You might be able to also convert the entire method to a switch

```
        private NuGetVersion ResolvePlatformReleaseChannelVersionTest(IEnumerable<NuGetVersion> bundleVersions, NuGetVersion currentMatchingVersion) => _platformReleaseChannel switch
        {
            "STANDARD" or "EXTENDED" => // call something,
            "LATEST" or "" => // call something,
            _ => // call something
        };
```

satvu: I combined the creation of the ordered bundles version list with the switch statement into one method based on another piece of feedback earlier (sorry, didn't see the second comment in time). I think it's alright to keep since it groups the reordering of the bundle versions and the decision-making based on the ordered list close together. 

Currently still using `ScriptConstants` file to store the known channel names, which is less readable, but given the broad name of this setting I think it could be used in other places eventually.

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -293,6 +282,54 @@ internal static string FindBestVersionMatch(VersionRange versionRange, IEnumerab
...
+            {
+                ScriptConstants.StandardPlatformChannelNameUpper or ScriptConstants.ExtendedPlatformChannelNameUpper => GetStandardOrExtendedBundleVersion(bundleVersions),
+                ScriptConstants.LatestPlatformChannelNameUpper or "" => GetLatestBundleVersion(bundleVersions),
+                _ => HandleUnknownPlatformReleaseChannelName(bundleVersions)
+            };
+        }
+
+        private NuGetVersion GetStandardOrExtendedBundleVersion(IList<NuGetVersion> orderedByDescBundlesList)
+        {
+            if (orderedByDescBundlesList.Count() > 1)
```
jviau: Since you are using `IList`, you can use `.Count` and indexer `[index]`

kshyju: Use the `Count` property since you are dealing with the IList now (The `Count()` method might be internally returning the `Count` in the implementation, but it's not best to make assumptions about the implementation).

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -293,6 +282,54 @@ internal static string FindBestVersionMatch(VersionRange versionRange, IEnumerab
...
+            {
+                return orderedByDescBundlesList.ElementAt(1); // if n is latest, should get version n-1 as this is how the platform release channel defines these channel names (resolve to version prior to latest)
+            }
+            else
```
kshyju: nit: Can remove the `else {` since you are returning above.

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -293,6 +282,54 @@ internal static string FindBestVersionMatch(VersionRange versionRange, IEnumerab
...
+            {
+                // keep the latest version, log a notice
+                var latest = orderedByDescBundlesList.FirstOrDefault();
+                _logger.LogInformation("Unable to apply platform release channel configuration {platformReleaseChannelName}. Only one matching bundle version is available. {latestBundleVersion} will be used", _platformReleaseChannel, latest);
+                return latest;
+            }
+        }
+
+        private NuGetVersion GetLatestBundleVersion(IList<NuGetVersion> orderedByDescBundlesList)
```
kshyju:
```suggestion
        private static NuGetVersion GetLatestBundleVersion(IList<NuGetVersion> orderedByDescBundlesList)
```

kshyju (approved):

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -293,6 +282,52 @@ internal static string FindBestVersionMatch(VersionRange versionRange, IEnumerab
...
+                ScriptConstants.LatestPlatformChannelNameUpper or "" => GetLatestBundleVersion(bundleVersions),
+                _ => HandleUnknownPlatformReleaseChannelName(bundleVersions)
+            };
+        }
+
+        private NuGetVersion GetStandardOrExtendedBundleVersion(IList<NuGetVersion> orderedByDescBundlesList)
+        {
+            if (orderedByDescBundlesList.Count > 1)
+            {
+                return orderedByDescBundlesList[1]; // if n is latest, should get version n-1 as this is how the platform release channel defines these channel names (resolve to version prior to latest)
```
jviau: This comment and the code doesn't line up. Why does it mention `n-1`, but you get `1`?

satvu: I'm trying to best translate the logic from here: https://github.com/Azure/azure-functions-host/issues/10778

We want the "next latest" or "version prior to latest". I can change the comment to: 
`For these release channel types, we should get the version prior to latest`? 

I'm thinking now that it's more confusing to go into the full logic and just keep it simple?

jviau: I read and understand the issue, as see what you are going for. The issue is written assuming latest would be `orderedByDescBundlesList[orderedByDescBundlesList.Count - 1]`, but we are ordering them in the opposite direction. We should update all comments in this PR to be based on how the code orders them (and not the ordering the issue assumes).

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -293,6 +293,38 @@ internal static string FindBestVersionMatch(VersionRange versionRange, IEnumerab
...
+        private NuGetVersion GetStandardOrExtendedBundleVersion(IList<NuGetVersion> orderedByDescBundlesList)
+        {
+            if (orderedByDescBundlesList.Count > 1)
+            {
+                return orderedByDescBundlesList[1]; // These channels should resolve to the version prior to latest. This list is in descending order, which makes latest [0], and prior-to-latest [1].
+            }
+
+            // keep the latest version, log a notice
+            var latest = orderedByDescBundlesList.FirstOrDefault();
+            _logger.LogInformation("Unable to apply platform release channel configuration {platformReleaseChannelName}. Only one matching bundle version is available. {latestBundleVersion} will be used", _platformReleaseChannel, latest);
```
jviau, line 317: The usage of `FirstOrDefault()` above, specifically the _OrDefault_ part means `latest` could be `null`. Or are there other safeguards in place to ensure there is at least 1 element in the list?

Should this be a `.First()`, or do you intentionally want this log to include `latestBundleVersion: null` for investigation purposes?

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -261,9 +261,9 @@ internal static string FindBestVersionMatch(VersionRange versionRange, IEnumerab
                     version = versionRange.Satisfies(version) ? version : null;
                 }
                 return version;
-            }).Where(v => v != null);
+            }).Where(v => v != null).OrderByDescending(version => version.Version).ToList();
 
-            var matchingVersion = bundleVersions.OrderByDescending(version => version.Version).FirstOrDefault();
+            var matchingVersion = ResolvePlatformReleaseChannelVersion(bundleVersions);
```
jviau, line 267: There are many null-handlers for `matchingVersion` below via `?.` operators. But I wonder what the actual impact of this being null is? Is this an error scenario? If so, should this be updated to log and maybe error out if `matchingVersion == null`?

jviau (approved):

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -293,6 +293,38 @@ internal static string FindBestVersionMatch(VersionRange versionRange, IEnumerab
             return matchingVersion?.ToString();
         }
 
+        private NuGetVersion ResolvePlatformReleaseChannelVersion(IList<NuGetVersion> orderedByDescBundles) => _platformReleaseChannel switch
+        {
+            ScriptConstants.StandardPlatformChannelNameUpper or ScriptConstants.ExtendedPlatformChannelNameUpper => GetStandardOrExtendedBundleVersion(orderedByDescBundles),
+            ScriptConstants.LatestPlatformChannelNameUpper or "" => GetLatestBundleVersion(orderedByDescBundles),
```
liliankasem, line 299: can you use string.Empty here?

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -293,6 +293,38 @@ internal static string FindBestVersionMatch(VersionRange versionRange, IEnumerab
...
+            _ => HandleUnknownPlatformReleaseChannelName(orderedByDescBundles)
+        };
+
+        private NuGetVersion GetStandardOrExtendedBundleVersion(IList<NuGetVersion> orderedByDescBundlesList)
+        {
+            if (orderedByDescBundlesList.Count > 1)
+            {
+                return orderedByDescBundlesList[1]; // These channels should resolve to the version prior to latest. This list is in descending order, which makes latest [0], and prior-to-latest [1].
```
liliankasem, line 307: super nit: comment above the line rather than on the same line

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
liliankasem, line 300: can you use string.Empty here?

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
liliankasem: super nit: comment above line rather than on the same line as the code; esp for a long comment like this

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -293,6 +293,38 @@ internal static string FindBestVersionMatch(VersionRange versionRange, IEnumerab
...
         }
 
+        private NuGetVersion ResolvePlatformReleaseChannelVersion(IList<NuGetVersion> orderedByDescBundles) => _platformReleaseChannel switch
+        {
+            ScriptConstants.StandardPlatformChannelNameUpper or ScriptConstants.ExtendedPlatformChannelNameUpper => GetStandardOrExtendedBundleVersion(orderedByDescBundles),
+            ScriptConstants.LatestPlatformChannelNameUpper or "" => GetLatestBundleVersion(orderedByDescBundles),
+            _ => HandleUnknownPlatformReleaseChannelName(orderedByDescBundles)
+        };
+
+        private NuGetVersion GetStandardOrExtendedBundleVersion(IList<NuGetVersion> orderedByDescBundlesList)
```
liliankasem, line 307: Can you add a comment here stating what Fabio said in the original work item:

> Standard: Resolves to the version prior to the latest (n-1), if that version is available.
Extended: Resolves to the version two releases prior to the latest (n-2), if that version is available.
However, Functions and Rapid Update should treat Standard and Extended the same, resolving to n-1.

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -293,6 +293,38 @@ internal static string FindBestVersionMatch(VersionRange versionRange, IEnumerab
...
+                return orderedByDescBundlesList[1]; // These channels should resolve to the version prior to latest. This list is in descending order, which makes latest [0], and prior-to-latest [1].
+            }
+
+            // keep the latest version, log a notice
+            var latest = orderedByDescBundlesList.FirstOrDefault();
+            _logger.LogInformation("Unable to apply platform release channel configuration {platformReleaseChannelName}. Only one matching bundle version is available. {latestBundleVersion} will be used", _platformReleaseChannel, latest);
+            return latest;
+        }
+
+        private static NuGetVersion GetLatestBundleVersion(IList<NuGetVersion> orderedByDescBundlesList)
```
liliankasem, line 321: Why does this need to be static?

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -293,6 +293,38 @@ internal static string FindBestVersionMatch(VersionRange versionRange, IEnumerab
...
+        {
+            return orderedByDescBundlesList.FirstOrDefault();
```
liliankasem, line 323: nit: If this is all that happens here, it might be more readable to have this be part of the switch directly 

> ScriptConstants.LatestPlatformChannelNameUpper or "" => orderedByDescBundlesList.FirstOrDefault(),

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
satvu: Suggested by @kshyju - it doesn't use any instance fields or props and provides some perf benefits.

liliankasem: How much perf benefits are we talking here 😭

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
soninaren: If the matching bundle version turns out to be null, then ultimately we end up logging an error statement, saying error loading bundle. The host continues to start up as expected. If there are functions that require extensions from a bundle to execute will log an error.

soninaren (approved):

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
satvu: Logging `null` is intentional as having a `null` value resolved for bundle resolution is currently supported behavior (see @soninaren's [comment](https://github.com/Azure/azure-functions-host/pull/10790/#discussion_r1945589167)).

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
kshyju: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1822
The gain might be minor in this specific case. But It's a good idea to stick with that pattern in our code base when possible.

Keep in mind that performance is the result of thousands of small optimizations.

PR #10802, opened 2025-02-06
fixing some test races (hopefully)

brettsam:
Hopefully fixing a handful of failures:

1. When asking for the HostInstanceId in tests that did restarts -- it would occasionally have a null ActiveHost, meaning there'd be some exception during DI service resolution. This adds a retry for a bit and a better error when this happens.
2. An error in Metrics tests with "InvocationId has already been used ". Looking through this, I saw that we manually call `Flush()` at the end of some tests as well as set the internal FlushTimer to 2 seconds. Sometimes these will happen simultaneously and it appears we flush the same event multiple times, leading to this weird error. Locking around this should fix it.
3. StackOverflow in one of our powershell scripts. There's a few issues (https://github.com/Azure/azure-powershell/issues/26623) on this around github. Most say switcing to `pwsh` fixes the issue, so giving that a try.
4. Some of our multi-process tests were very flaky. They're checking for all processes named "node" on the machine and not really waiting the right amount of time for processes to spin up, etc. I had recently taken a different approach in https://github.com/Azure/azure-functions-host/pull/9820 where, rather than check externally for processes to spin-up, I looked directly at the channels to be ready before continuing. This seems to be a lot more stable, so I re-wrote a few tests to use this pattern.

### Pull request checklist

    * [x] Otherwise:  https://github.com/Azure/azure-functions-host/pull/10818

### DIFF
--- a/eng/ci/templates/official/jobs/run-integration-tests.yml
+++ b/eng/ci/templates/official/jobs/run-integration-tests.yml
@@ -46,6 +46,7 @@ jobs:
     inputs:
       azureSubscription: Azure-Functions-Host-CI-internal
       azurePowerShellVersion: 'LatestVersion'
+      pwsh: true
       ScriptPath: eng/script/checkout-secrets.ps1
 
   - task: AzureKeyVault@1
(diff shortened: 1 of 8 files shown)


liliankasem (approved):

cjaliaga (approved):

jviau (approved):
💯 🔥 

You are a saint for fixing these. I have been slowly losing my mind to them.

brettsam:
Things looked stable on this last run -- but going to kick this off a few more times to see if anything else crops up (or maybe I didn't fix the issues I thought I did).

I'm not entirely sure why this all got worse recently -- maybe something hardware-wise changed meaning timing was subtly different than it has always been? Or test ordering caused instability? No clue, but there were obvious issues with these tests, so this should be a good step in the right direction nonetheless.

PR #10848, opened 2025-02-13
[Bug Fix] Refactor Linux container management to avoid race condition that leads the host to initialize placeholder (warmup) function

### LINKED ISSUE #10725: Race condition leads the host to initialize placeholder (`warmup`) function in Linux environments
On Linux environments, a race condition may lead the host to initialize placeholder (`warmup`) function.

#### Repro steps

- Deploy a function app in Flex Consumption with a worker that will fail initialization (e.g., crash at startup)
- The platform will restart the host. During the restart, the LinuxInstanceManager.StartAssignment logic runs, which pushes site assignment to a background task (see the line of code [here](https://github.com/Azure/azure-functions-host/blob/cb0a62c51dd800ee1919795b4a933b26c74ed8de/src/WebJobs.Script.WebHost/Management/LinuxInstanceManager.cs#L94)).

- Because this assignment call is “fire and forget,” the environment variable indicating the host is no longer in placeholder mode sometimes isn’t set before the host initialization proceeds.

- The host detects placeholder mode environment settings and incorrectly indexes the placeholder (warmup) function.

#### Expected behavior

The host should wait for the assignment to complete before initializing the `JobHost` so placeholder information (as well as application settings) is correctly applied.  As a result, it should not index the placeholder/warmup function if the assignment step has actually completed.

#### Actual behavior

Due to the race condition, the host sometimes sees the environment as still in placeholder mode because the assignment logic hasn’t finished setting the appropriate environment variables. This leads the host to index the placeholder function incorrectly.

satvu:

resolves #10725 

### Pull request checklist

    *** Otherwise: Link to backporting PR** - [backport PR](https://github.com/Azure/azure-functions-host/pull/10958)

This change introduces a new, asynchronous method for assigning an instance in order to resolve the race condition described in #10725. It also maintains the existing "fire and forget" method which is necessary for the HTTP call [here](https://github.com/Azure/azure-functions-host/blob/e23bd89f14f2b9c2d55025a179d33f3d6ec4684f/src/WebJobs.Script.WebHost/Controllers/InstanceController.cs#L60) in the instance controller. 

This PR also includes a refactor of flaky tests, removing calls of `Task.Delay()` and switching to the use of the async method for multiple E2E and integration scenarios.


src/WebJobs.Script.WebHost/ContainerManagement/LinuxContainerInitializationHostedService.cs
```diff
@@ -51,8 +51,15 @@ private async Task ApplyStartContextIfPresent(CancellationToken cancellationToke
                 var assignmentContext = _startupContextProvider.SetContext(encryptedAssignmentContext);
                 await SpecializeMSISideCar(assignmentContext);
 
-                bool success = _instanceManager.StartAssignment(assignmentContext);
-                _logger.LogDebug($"StartAssignment invoked (Success={success})");
+                if (_instanceManager.IsValidEnvironment(assignmentContext))
```
jviau: Can the `IsValidEnvironment` check be added to `StartAssignment`? Or we can add a new method `Task<bool> InstanceManager.TryStartAssignmentAsync`? This way the callers don't need to know to call `IsValidEnvironment` before

src/WebJobs.Script.WebHost/Management/LinuxInstanceManager.cs
```diff
@@ -91,19 +70,42 @@ public bool StartAssignment(HostAssignmentContext context)
                 _webHostEnvironment.DelayRequests();
 
                 // start the specialization process in the background
-                Task.Run(async () => await AssignAsync(context));
-
-                return true;
+                return Task.Run(async () =>
+                {
+                    await AssignAsync(context);
+                    return true;
```
satvu: Could technically also use `return await AssignAsync(context)` which has a `Task<bool>` return type, but I think `return true` is closer to the previous "fire and forget" pattern - still evaluating the downstream effects of this change.

src/WebJobs.Script.WebHost/Management/LinuxInstanceManager.cs
```diff
@@ -44,38 +36,25 @@ public LinuxInstanceManager(IHttpClientFactory httpClientFactory, IScriptWebHost
...
-                return false;
-            }
-
             if (context.IsWarmupRequest)
             {
                 // Based on profiling download code jit-ing holds up cold start.
                 // Pre-jit to avoid paying the cost later.
-                Task.Run(async () => await DownloadWarmupAsync(context.GetRunFromPkgContext()));
-                return true;
+                return Task.Run(async () =>
```
jviau: The `Task.Run` is no longer needed.

src/WebJobs.Script.WebHost/Management/LinuxInstanceManager.cs
```diff
@@ -91,19 +70,42 @@ public bool StartAssignment(HostAssignmentContext context)
                 _webHostEnvironment.DelayRequests();
 
                 // start the specialization process in the background
-                Task.Run(async () => await AssignAsync(context));
-
-                return true;
+                return Task.Run(async () =>
```
jviau: Same here, `Task.Run` is no longer needed. All it will do is force a yield as the work inside it gets posted to the thread pool.

src/WebJobs.Script.WebHost/Management/LinuxInstanceManager.cs
jviau: Make the method itself `async`, and then make this change:

``` diff
- return Task.Run(async () =>
- {
-     await AssignAsync(context);
-     return true;
- }
+ await AssignAsync(context);
+ return true;
```

You will then need to remove all the `return Task.FromResult(x);`, instead just have `return x;`

src/WebJobs.Script.WebHost/Management/IInstanceManager.cs
```diff
@@ -13,7 +13,7 @@ public interface IInstanceManager
 
         Task<string> ValidateContext(HostAssignmentContext assignmentContext);
 
-        bool StartAssignment(HostAssignmentContext assignmentContext);
+        Task<bool> StartAssignment(HostAssignmentContext assignmentContext);
```
jviau: Maybe we should rename this. It is no longer a `Start` anymore with the `Task`. Maybe `AssignInstanceAsync`?

test/WebJobs.Script.Tests.Integration/Management/InstanceManagerTests.cs
```diff
@@ -67,7 +66,7 @@ public InstanceManagerTests()
         }
 
         [Fact]
-        public async Task StartAssignment_AppliesAssignmentContext()
```
satvu: Bringing this PR back to draft following the refactor. Current approach:

1. Keep all tests that reference `StartAssignment` as they are.
2. Review other tests to understand which scenario they are more likely to fall under (synchronous fire and forget call through `InstanceController` or async call to `AssignInstanceAsync` through `LinuxContainerInitializationHostedService`. These reviews will likely need to be done in `LegionInstanceManagerTests` and `AtlasInstanceManagereTests` as well.

satvu: Resolved - there are tests for the async path under `AtlasContainerInitializationHostedServiceTests` and `LegionContainerInitializationHostedServiceTests`.

src/WebJobs.Script.WebHost/Management/LinuxInstanceManager.cs
```diff
@@ -44,65 +36,106 @@ public LinuxInstanceManager(IHttpClientFactory httpClientFactory, IScriptWebHost
...
+            // that any subsequent incoming requests while the assign is in progress
+            // will be delayed until complete
+            _webHostEnvironment.DelayRequests();
+
+            // start the specialization process in the background
+            await AssignAsync(context);
+            return true;
+        }
+
+        public async Task<bool> HandleWarmupRequestAsync(HostAssignmentContext context)
```
jviau: can this be `private`? Also please move to below `public` if making `private`.

src/WebJobs.Script.WebHost/Management/LinuxInstanceManager.cs
```diff
@@ -44,65 +36,106 @@ public LinuxInstanceManager(IHttpClientFactory httpClientFactory, IScriptWebHost
...
-                _logger.LogError("Assign called while container is marked as specialized.");
                 return false;
             }
 
             if (context.IsWarmupRequest)
             {
-                // Based on profiling download code jit-ing holds up cold start.
-                // Pre-jit to avoid paying the cost later.
-                Task.Run(async () => await DownloadWarmupAsync(context.GetRunFromPkgContext()));
+                _ = HandleWarmupRequestAsync(context);
```
jviau: Do we know if this is guaranteed to be async? If not, may be safer to leave it in a `Task.Run`.

src/WebJobs.Script.WebHost/Management/LinuxInstanceManager.cs
```diff
@@ -44,65 +36,106 @@ public LinuxInstanceManager(IHttpClientFactory httpClientFactory, IScriptWebHost
 
         public abstract Task<string> SpecializeMSISidecar(HostAssignmentContext context);
 
-        public bool StartAssignment(HostAssignmentContext context)
+        private async Task<bool> AssignInstanceAsyncCore(HostAssignmentContext context)
```
jviau: wonder if this logic can be rolled into `AssignAsync` at this point.

test/WebJobs.Script.Tests.Integration/ContainerManagement/AtlasContainerInitializationHostedServiceTests.cs
```diff
@@ -155,7 +155,7 @@ public async Task Assigns_Even_If_MSI_Specialization_Fails()
                 m.SpecializeMSISidecar(It.Is<HostAssignmentContext>(context =>
                     hostAssignmentContext.Equals(context) && !context.IsWarmupRequest))).Returns(Task.FromResult("msi-specialization-error"));
 
-            _instanceManagerMock.Setup(manager => manager.StartAssignment(It.Is<HostAssignmentContext>(context => hostAssignmentContext.Equals(context) && !context.IsWarmupRequest))).Returns(true);
+            _instanceManagerMock.Setup(manager => manager.AssignInstanceAsync(It.Is<HostAssignmentContext>(context => hostAssignmentContext.Equals(context) && !context.IsWarmupRequest))).Returns(Task.FromResult(true));
```
jviau, line 113: Moq has a small helper for this:

```suggestion
            _instanceManagerMock.Setup(manager => manager.AssignInstanceAsync(It.Is<HostAssignmentContext>(context => hostAssignmentContext.Equals(context) && !context.IsWarmupRequest))).ReturnsAsync(true);
```

test/WebJobs.Script.Tests.Integration/Management/InstanceManagerTests.cs
```diff
@@ -67,7 +66,7 @@ public InstanceManagerTests()
...
+          public async Task StartAssignment_AppliesAssignmentContext()
```
jviau: nit:
```suggestion
        public async Task StartAssignment_AppliesAssignmentContext()
```

test/WebJobs.Script.Tests.Integration/Management/InstanceManagerTests.cs
jviau, line 1: I would prefer these tests to be switched to primarily use `AssignInstanceAsync`. Keep only some for `StartAssignment` and use a `TaskCompletionSource` in a moq to identify concretely when it reaches the point you need.

These tests are flaky. Had `Does_Not_Mount_Invalid_BYOS_Accounts` fail in a random PR for me today.

src/WebJobs.Script.WebHost/Management/LinuxInstanceManager.cs
```diff
@@ -62,24 +98,28 @@ public bool StartAssignment(HostAssignmentContext context)
...
+                catch (Exception ex)
                 {
-                    if (_assignmentContext != null)
-                    {
-                        return _assignmentContext.Equals(context);
-                    }
-                    _assignmentContext = context;
+                    _logger.LogError(ex, "Warmup download failed");
+                    await _meshServiceClient.NotifyHealthEvent(ContainerHealthEventType.Warning, GetType(), "Warmup download failed");
+                    throw;
```
satvu: There wasn't any error handling for this previously - mostly tried to match the same pattern as an assignment failure, but not sure if this should be as aggressive (`ContainerHealthEventType.Fatal` and throw).

src/WebJobs.Script.WebHost/ContainerManagement/LinuxContainerInitializationHostedService.cs
```diff
@@ -51,8 +52,15 @@ private async Task ApplyStartContextIfPresent(CancellationToken cancellationToke
...
-                _logger.LogDebug($"StartAssignment invoked (Success={success})");
+                try
+                {
+                    bool success = await _instanceManager.AssignInstanceAsync(assignmentContext);
+                    _logger.LogDebug("AssignInstanceAsync was invoked (Success={success})", success);
+                }
+                catch (Exception ex)
+                {
+                    _logger.LogError(ex, "Failed to assign instance.");
+                }
```
satvu, line 63: Added tests for this in `AtlasContainerInitializationHostedServiceTests` and `LegionContainerInitializationHostedServiceTests` - previous fire and forget pattern didn't observe exceptions related to assignment failure.

test/WebJobs.Script.Tests.Integration/Management/InstanceManagerTests.cs
fabiocav: @satvu this is aligned with what we had previously discussed. Some of the tests might need some changes once you use the new API, but the overall behavior should be verifiable with that.

src/WebJobs.Script.WebHost/Management/IInstanceManager.cs
```diff
@@ -13,6 +13,18 @@ public interface IInstanceManager
...
+        /// AssignInstanceAsync will asynchronously assign an instance.
+        /// </summary>
+        /// <param name="assignmentContext">Takes in a <see cref="HostAssignmentContext"/> which will be applied to an assigned instance.</param>
+        /// <returns>Returns true if assiging an instance succeeds, false otherwise.</returns>
+        Task<bool> AssignInstanceAsync(HostAssignmentContext assignmentContext);
+
+        /// <summary>
+        /// StartAssinment will validate the current environment then assign an instance in a "fire and forget" pattern. Should be used synchronously.
+        /// </summary>
+        /// <param name="assignmentContext">Takes in a <see cref="HostAssignmentContext"/> which will be applied to an assigned instance</param>
```
fabiocav: nit: replace `Takes in a` with `The`:

Suggestion (including minor verbiage adjustments):
``` csharp
/// <param name="assignmentContext">The <see cref="HostAssignmentContext"/> that will be applied to the instance being assigned to the application.</param>
```

src/WebJobs.Script.WebHost/Management/IInstanceManager.cs
```diff
@@ -13,6 +13,18 @@ public interface IInstanceManager
 
         Task<string> ValidateContext(HostAssignmentContext assignmentContext);
 
+        /// <summary>
+        /// AssignInstanceAsync will asynchronously assign an instance.
+        /// </summary>
+        /// <param name="assignmentContext">Takes in a <see cref="HostAssignmentContext"/> which will be applied to an assigned instance.</param>
```
fabiocav: nit: replace this description with the one mentioned below.

src/WebJobs.Script.WebHost/Management/IInstanceManager.cs
```diff
@@ -13,6 +13,18 @@ public interface IInstanceManager
...
+        /// <returns>Returns true if assiging an instance succeeds, false otherwise.</returns>
```
fabiocav:
```suggestion
        /// <returns><see langword="true"/> if environment validation succeeds; otherwise <see langword="false"/>.</returns>
```

src/WebJobs.Script.WebHost/Management/IInstanceManager.cs
```diff
@@ -13,6 +13,18 @@ public interface IInstanceManager
...
+        Task<bool> AssignInstanceAsync(HostAssignmentContext assignmentContext);
+
+        /// <summary>
+        /// StartAssinment will validate the current environment then assign an instance in a "fire and forget" pattern. Should be used synchronously.
```
fabiocav: spelling: `StartAssignment`

fabiocav: Another thought, let's simply remove the method name here:

`Validates the assignment context and begins the assignment process in a "fire and forget" pattern.`

src/WebJobs.Script.WebHost/Management/IInstanceManager.cs
```diff
@@ -13,6 +13,18 @@ public interface IInstanceManager
...
+        /// </summary>
+        /// <param name="assignmentContext">Takes in a <see cref="HostAssignmentContext"/> which will be applied to an assigned instance</param>
+        /// <returns>Returns true if environment validation succeeds, false otherwise.</returns>
```
fabiocav:
```suggestion
        /// <returns><see langword="true"/> if environment validation succeeds; otherwise <see langword="false"/>.</returns>
```

src/WebJobs.Script.WebHost/Management/IInstanceManager.cs
```diff
@@ -13,6 +13,18 @@ public interface IInstanceManager
 
         Task<string> ValidateContext(HostAssignmentContext assignmentContext);
 
+        /// <summary>
+        /// AssignInstanceAsync will asynchronously assign an instance.
```
fabiocav: Remove method name (e.g., `Asynchronously assigns a host instance.`)

src/WebJobs.Script.WebHost/Management/LinuxInstanceManager.cs
```diff
@@ -62,24 +98,28 @@ public bool StartAssignment(HostAssignmentContext context)
...
-            else if (_assignmentContext == null)
+            return true;
+        }
+
+        private async Task AssignAsync(HostAssignmentContext assignmentContext)
+        {
+            if (assignmentContext.IsWarmupRequest)
             {
-                lock (_assignmentLock)
+                try
```
fabiocav: Can you please move the logic to handle a warmup request into its own method (e.g., `HandleWarmupRequestAsync` or something similar).

src/WebJobs.Script.WebHost/Management/LinuxInstanceManager.cs
```diff
@@ -62,24 +98,28 @@ public bool StartAssignment(HostAssignmentContext context)
...
+                }
+                catch (Exception ex)
                 {
-                    if (_assignmentContext != null)
-                    {
-                        return _assignmentContext.Equals(context);
-                    }
-                    _assignmentContext = context;
+                    _logger.LogError(ex, "Warmup download failed");
+                    await _meshServiceClient.NotifyHealthEvent(ContainerHealthEventType.Warning, GetType(), "Warmup download failed");
```
fabiocav: Wanted to callout this health metric/event being sent here (in the context of the metrics work). This is specific to Linux Consumption and would be good to track as part of existing metrics.

src/WebJobs.Script.WebHost/Management/LinuxInstanceManager.cs
```diff
@@ -62,24 +98,28 @@ public bool StartAssignment(HostAssignmentContext context)
...
+                    throw;
                 }
+                return;
+            }
 
+            try
+            {
                 _logger.LogInformation($"Starting Assignment. Cloud Name: {_environment.GetCloudName()}");
```
fabiocav, line 125: Debating whether we should change this to take in a log argument. It would technically a break of this log statement.

src/WebJobs.Script.WebHost/Management/LinuxInstanceManager.cs
```diff
@@ -36,43 +36,13 @@ public LinuxInstanceManager(IHttpClientFactory httpClientFactory, IScriptWebHost
...
-        }
-
         public async Task<bool> AssignInstanceAsync(HostAssignmentContext context)
         {
             if (!IsValidEnvironment(context))
             {
                 return false;
             }
 
-            if (context.IsWarmupRequest)
```
jviau: I think the explicit warmup calls need to remain how you had them in the previous iteration - separate and before the `lock (_assignmentLock)` calls.

src/WebJobs.Script.WebHost/Management/LinuxInstanceManager.cs
```diff
@@ -139,8 +103,24 @@ private bool IsValidEnvironment(HostAssignmentContext context)
 
         private async Task AssignAsync(HostAssignmentContext assignmentContext)
         {
+            if (assignmentContext.IsWarmupRequest)
```
jviau: As mentioned in the other comment, what you had before is preferred (keeping this before the lock statements in `StartAssignment` and `AssignInstanceAsync`.)

src/WebJobs.Script.WebHost/Management/LinuxInstanceManager.cs
jviau: Also this is a standardized piece of telemetry, so we wouldn't want a different name for it than the standards

test/WebJobs.Script.Tests.Integration/ContainerManagement/AtlasContainerInitializationHostedServiceTests.cs
```diff
@@ -120,26 +98,89 @@ public async Task Does_Not_Assign_If_Context_Not_Available()
...
-            secrets.Host = new FunctionAppSecrets.HostSecrets
+            var secrets = new FunctionAppSecrets
             {
-                Master = "test-key",
-                Function = new Dictionary<string, string>
+                Host = new FunctionAppSecrets.HostSecrets
+                {
+                    Master = "test-key",
+                    Function = new Dictionary<string, string>
                 {
```
jviau: nit: indent the braces as well to match the line above.

src/WebJobs.Script.WebHost/Management/LinuxInstanceManager.cs
```diff
@@ -44,70 +36,97 @@ public LinuxInstanceManager(IHttpClientFactory httpClientFactory, IScriptWebHost
...
+                return true;
+            }
+
+            lock (_assignmentLock)
+            {
+                if (_assignmentContext != null)
                 {
-                    _logger.LogError("Assign called while host is not in placeholder mode and start context is not present.");
-                    return false;
+                    return _assignmentContext.Equals(context);
```
jviau: If we have 2 concurrent callers to this, the caller which _does not_ win the assignment may return before the assignment is complete. The caller which won the assignment may still be in `AssignAsync` when the other returns at this line.

Will that be an issue?

fabiocav: For your exact scenario, yes, that would be a problem as it would lead to a similar race to the one PR aims to address (i.e., caller B would move on assuming the context was already applied).

The original assignment task should be awaited on. Returning here in the error state (when the context doesn't match) is fine, but the valid case should await on the same task.

src/WebJobs.Script.WebHost/Management/LinuxInstanceManager.cs
```diff
@@ -44,70 +37,102 @@ public LinuxInstanceManager(IHttpClientFactory httpClientFactory, IScriptWebHost
...
             }
-        }
 
-        public abstract Task<string> ValidateContext(HostAssignmentContext assignmentContext);
+            return true;
+        }
 
         private async Task AssignAsync(HostAssignmentContext assignmentContext)
         {
+            await Task.Yield();
```
jviau:
```suggestion
            await Task.Yield(); // This may be called from within a lock. This will force the lock to be released.
```

jviau: Also a side benefit: this guarantees `StartAssignment` will be fire & forget.

jviau (approved):

src/WebJobs.Script.WebHost/Management/LinuxInstanceManager.cs
jviau: Also @fabiocav , can you confirm if that is correct? A lock will be released once a thread yields, right?

src/WebJobs.Script.WebHost/Management/LinuxInstanceManager.cs
```diff
@@ -44,70 +37,102 @@ public LinuxInstanceManager(IHttpClientFactory httpClientFactory, IScriptWebHost
...
-                        return _assignmentContext.Equals(context);
-                    }
-                    _assignmentContext = context;
+                    return _assignmentContext.Equals(context);
                 }
+                _assignmentContext = context;
+            }
 
-                _logger.LogInformation($"Starting Assignment. Cloud Name: {_environment.GetCloudName()}");
+            _ = AssignAsync(context);
```
fabiocav: We should this call into the `lock` scope and assign it to `_assignment`.

No need to await in this method, but doing that will ensure the `Task` is assigned and subsequent calls to `AssignInstanceAsync` would await on the same `Task`. Skipping this might create a condition where you might run into a null reference in line 66

src/WebJobs.Script.WebHost/Management/LinuxInstanceManager.cs
fabiocav: The lock will be released, but I'm wondering if we should adjust the comment for clarity, as it isn't `Task.Yield` forcing that to happen, but the fact that when this `await` is hit, control flow will return to the caller and the lock will be released when it exits the lock scope.

fabiocav (approved):

PR #10850, opened 2025-02-14
test improvements

brettsam:
Fixing and improving some test:
- xunit v2 doesn't support `IAsyncDisposable`, but it does support `IAsyncLifetime` for test classes. One test wasn't properly disposing b/c `DisposeAsync()` was never being called.
- Moving the build step for tests outside of the secrets lock we use for resources. This saves 4-5 minutes of lock time per run.
- Overriding the step that creates storage entities for some tests that don't need them. Should speed things up a little bit.

### Pull request checklist

    * Otherwise: https://github.com/Azure/azure-functions-host/pull/10851

### DIFF
--- a/test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SamplesEndToEndTests_Node_MultipleProcesses.cs
+++ b/test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SamplesEndToEndTests_Node_MultipleProcesses.cs
@@ -21,13 +21,7 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.EndToEnd
     [Trait(TestTraits.Group, TestTraits.SamplesEndToEnd)]
     public class SamplesEndToEndTests_Node_MultipleProcesses : IAsyncLifetime
     {
-        private readonly MultipleProcessesTestFixture _fixture;
-
-        public SamplesEndToEndTests_Node_MultipleProcesses()
-        {
-            // We want a new fixture for each test
-            _fixture = new MultipleProcessesTestFixture();
-        }
+        private readonly MultipleProcessesTestFixture _fixture = new();
 
         [Fact]
         public async Task NodeProcess_Different_AfterHostRestart()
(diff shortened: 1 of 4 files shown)


test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SamplesEndToEndTests_Node_MultipleProcesses.cs
```diff
@@ -19,20 +19,19 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.EndToEnd
...
+    public class SamplesEndToEndTests_Node_MultipleProcesses : IAsyncLifetime
     {
-        private readonly MultiplepleProcessesTestFixture _fixture;
+        private readonly MultipleProcessesTestFixture _fixture;
 
         public SamplesEndToEndTests_Node_MultipleProcesses()
         {
             // We want a new fixture for each test
-            _fixture = new MultiplepleProcessesTestFixture();
+            _fixture = new MultipleProcessesTestFixture();
```
jviau: This is intentional to have it be an instance per test, instead of `IClassFixture<T>`?

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SamplesEndToEndTests_Node_MultipleProcesses.cs
```diff
@@ -19,20 +19,19 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.EndToEnd
 {
     [Trait(TestTraits.Category, TestTraits.EndToEnd)]
     [Trait(TestTraits.Group, TestTraits.SamplesEndToEnd)]
-    public class SamplesEndToEndTests_Node_MultipleProcesses : IAsyncDisposable
+    public class SamplesEndToEndTests_Node_MultipleProcesses : IAsyncLifetime
     {
-        private readonly MultiplepleProcessesTestFixture _fixture;
+        private readonly MultipleProcessesTestFixture _fixture;
```
jviau: Can inline this and remove the ctor
```suggestion
        private readonly MultipleProcessesTestFixture _fixture = new();
```

jviau (approved):

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SamplesEndToEndTests_Node_MultipleProcesses.cs
brettsam: yeah, each of these tests are restarting the host (and worker(s)) several times. i refactored this in my last PR so that they were starting from a stable starting point rather than restarting after a bunch of timeouts and other restarts forced by other tests.

liliankasem (approved):

PR #10854, opened 2025-02-18
removing static ActiveOperations

### LINKED ISSUE #10831: Unstable test: SamplesEndToEndTests_CustomHandlerRetry.HttpTrigger_CustomHandlerRetry_Get_Succeeds
https://github.com/Azure/azure-functions-host/blob/b9214a2345a3becd6b67a7f7da2e0cf7a0050ac5/test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SamplesEndToEndTests_CustomHandlerRetry.cs#L32

https://dev.azure.com/azfunc/internal/_test/analytics?definitionId=463&contextType=build

brettsam:
fixes #10831

previous PR #10850 helped, but the ultimate issue was that ActiveOperations was static, which allowed unrelated tests to interact with each other

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.WebHost/ScriptHostStartupOperation.cs
+++ b/src/WebJobs.Script.WebHost/ScriptHostStartupOperation.cs
@@ -4,49 +4,50 @@
 using System;
 using System.Threading;
 
-namespace Microsoft.Azure.WebJobs.Script.WebHost;
-
-/// <summary>
-/// Used to track all outstanding startup operations.
-/// </summary>
-internal class ScriptHostStartupOperation : IDisposable
+namespace Microsoft.Azure.WebJobs.Script.WebHost
 {
-    private bool _disposed = false;
-
     /// <summary>
-    /// Initializes a new instance of the <see cref="ScriptHostStartupOperation"/> class.
+    /// Used to track all startup operations.
     /// </summary>
-    /// <param name="cancellationToken">A CancellationToken that will be linked and exposed via the <see cref="CancellationTokenSource"/> property.</param>
-    /// <param name="parentId">The parent operation Id, if applicable.</param>
-    public ScriptHostStartupOperation(CancellationToken cancellationToken, Guid? parentId = null)
+    internal class ScriptHostStartupOperation : IDisposable
     {
-        Id = Guid.NewGuid();
-        CancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
-        ParentId = parentId;
-    }
+        private bool _disposed = false;
 
-    /// <summary>
-    /// Gets the Id of the operation, used for tracking through logs.
-    /// </summary>
-    public Guid Id { get; }
-
-    /// <summary>
-    /// Gets the parent operation that started this one. Used when restarting during exceptions.
-    /// </summary>
-    public Guid? ParentId { get; }
-
-    /// <summary>
-    /// Gets the CancellationTokenSource used to cancel this operation. This is CancellationTokenSource is linked to the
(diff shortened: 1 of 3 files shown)


src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -1008,6 +1022,24 @@ private void FlushOpenTelemetry(IHost host)
             }
         }
 
+        private ScriptHostStartupOperation BeginStartupOperation(CancellationToken parentToken, Guid? parentId = null)
+        {
+            var operation = new ScriptHostStartupOperation(parentToken, parentId);
```
fabiocav, line 1030: For clarity, can you plase document the scenarios where `parentId` would be `null`?

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -1008,6 +1022,24 @@ private void FlushOpenTelemetry(IHost host)
...
+            _activeStartupOperations.TryAdd(operation, byte.MinValue);
```
fabiocav, line 1031: Do we want any special handling here when this fails? (main thing in mind is the `parentId` set `null` case)

src/WebJobs.Script.WebHost/ScriptHostStartupOperation.cs
```diff
@@ -2,108 +2,51 @@
...
-                CancellationTokenSource?.Dispose();
-                Log.StartupOperationCompleted(_logger, Id);
-                _disposed = true;
-            }
-        }
+    /// <summary>
+    /// Gets the CancellationTokenSource used to cancel this operation. This is CancellationTokenSource is linked to the
+    /// CancellationToken passed via the <see cref="Create(CancellationToken, ILogger, Guid?)"/> method.
+    /// </summary>
+    public CancellationTokenSource CancellationTokenSource { get; }
```
fabiocav, line 83: 👍🏼

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
brettsam: We only use a parent id when we're re-trying due to some other startup failure (like [here](https://github.com/Azure/azure-functions-host/blob/46194b9893def8e31a76a69e6dfbfeac5530834f/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs#L499)). It lets us flow back via logs to where everything started and follow the chain of startup attempts. Most of the time this will be `null` as it was not created as a result of some other failure.

None of this has changed since it was introduced 6 years ago, btw

brettsam: Added some docs in a few places to note that these come in via errors only.

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
brettsam: I can't think of why it'd fail as the key here is the object itself. If this object already existed in our tracking collection, then we'd still be in an okay state.

This behavior didn't really change with this; it originally lived inside the `ScriptHostStartupOperation` type itself (and static). So it's been like this for quite a while.

brettsam:
FYI -- my second commit undid some whitespace changes that made it look like I'd effectively re-written the StartupOperation type. I really just moved the collection-tracking out of there and cleaned up the unused logger. Everything else should remain the same as it has since this was introduced years ago.

jviau (approved):

fabiocav (approved):

PR #10860, opened 2025-02-19
Adding coldstart data collection support from dev inner loop

### LINKED ISSUE #10859: Capture cold start latency from dev inner loop
Build the infra to capture cold start latency from dev build in continuously (nightly schedule). This will allow us to 

1. Detect cold start regression within the dev inner loop.
2. Allows developers to test their cold start improvements and collect the numbers to understand the impact of the change. Allows developers to collect profiling data s needed for deeper investigation.
3. Gives us better insights to understand the root cause is from host changes if we observed cold start regression in the overall mission report numbers.

kshyju:
resolves #10859

Sample run: https://azfunc.visualstudio.com/internal/_build/results?buildId=203300&view=results

Each run will capture data for both ASP.NET Core and non-ASP.NET Core applications on Windows and Linux. A JSON file containing relevant details about the data collection process will be published as an artifact. For scheduled nightly runs, the collected data will be stored in a SQL table, enabling reporting and alerting.  

Perfview profile collection from windows agent is also enabled.  To capture the profile, set the  pipeline variable `CollectPerfviewProfile` to `true`

Sample run with perfview profile collection: https://azfunc.visualstudio.com/internal/_build/results?buildId=203913&view=results

### Pull request checklist

### DIFF
--- a/eng/ci/host.coldstart.yml
+++ b/eng/ci/host.coldstart.yml
@@ -1,0 +1,107 @@
+# No triggers for code push to any branch.
+trigger: none
+
+# No PR triggers for now
+pr: none
+
+schedules:
+  - cron: "0 1 * * *"
+    displayName: Nightly Build
+    branches:
+      include:
+      - dev
+    always: true
+
+resources:
+  repositories:
+  - repository: 1es
+    type: git
+    name: 1ESPipelineTemplates/1ESPipelineTemplates
+    ref: refs/tags/release
+  - repository: eng
+    type: git
+    name: engineering
+    ref: refs/tags/release
+
+variables:
+  - template: /eng/ci/templates/variables/coldstart.yml@self
+  - template: /ci/variables/cfs.yml@eng
+
+extends:
+  template: v1/1ES.Unofficial.PipelineTemplate.yml@1es
+  parameters:
+    pool:
+      name: 1es-pool-azfunc-benchmarking-large
+      image: 1es-windows-2022-benchmark-runner-vanilla
+      os: windows
+
+    stages:
+    - stage: RunWindows
+      displayName: Record latency(Windows)
+      jobs:
+      - ${{ each appId in split(variables.runInstances, ',') }}:
+        - template: /eng/ci/templates/official/jobs/run-coldstart.yml@self
+          parameters:
+            description: .NET9 Web Application
+            functionAppName: HelloHttpNet9
+            instanceId: ${{ appId }} 
+        - template: /eng/ci/templates/official/jobs/run-coldstart.yml@self
+          parameters:
+            description: .NET9 Worker Application
+            functionAppName: HelloHttpNet9NoProxy
+            instanceId: ${{ appId }}  
+
+    - stage: CalculateWindows
+      displayName: Process results(Windows)
+      dependsOn: RunWindows
+      jobs:
+      - template: /eng/ci/templates/official/jobs/process-coldstart.yml@self
+        parameters:
+          functionAppName: HelloHttpNet9
+          description: .NET9 Web Application
+          runInstances: 
+            items: ${{ split(variables.runInstances, ',') }}
+      - template: /eng/ci/templates/official/jobs/process-coldstart.yml@self
+        parameters:
+          functionAppName: HelloHttpNet9NoProxy
+          description: .NET9 Worker Application
+          runInstances: 
+            items: ${{ split(variables.runInstances, ',') }}
+
+# # LINUX
+    - stage: RunLinux
+      dependsOn: []
+      displayName:  Record latency(Linux)
+      jobs:
+      - ${{ each appId in split(variables.runInstances, ',') }}:
+        - template: /eng/ci/templates/official/jobs/run-coldstart.yml@self
+          parameters:
+            description: .NET9 Web Application
+            functionAppName: HelloHttpNet9
+            instanceId: ${{ appId }}
+            os: Linux
+        - template: /eng/ci/templates/official/jobs/run-coldstart.yml@self
+          parameters:
+            description: .NET9 Worker Application
+            functionAppName: HelloHttpNet9NoProxy
+            instanceId: ${{ appId }}
+            os: Linux
+  
+    - stage: CalculateLinux
+      displayName:  Process results(Linux)
+      dependsOn: RunLinux
+      jobs:
+      - template: /eng/ci/templates/official/jobs/process-coldstart.yml@self
+        parameters:
+          functionAppName: HelloHttpNet9
+          description: .NET9 Web Application
+          os: Linux
+          runInstances: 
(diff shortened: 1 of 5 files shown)


eng/ci/host.coldstart.yml
```diff
@@ -0,0 +1,107 @@
...
+      name: 1es-pool-azfunc-benchmarking-large
+      image: 1es-windows-2022-benchmark-runner-vanilla
+      os: windows
+
+    stages:
+    - stage: RunWindows
+      displayName: Record latency(Windows)
+      jobs:
+      - ${{ each appId in split(variables.runInstances, ',') }}:
+        - template: /eng/ci/templates/official/jobs/run-coldstart.yml@self
```
surgupta-msft, line 53: nit: I think if I recall correctly, yml files for unofficial pipeline will be under `templates/jobs` folder instead of `templates/official/jobs`

kshyju: what makes one pipeline official and one not?

surgupta-msft: We are using Unofficial template in line 31 - `template: v1/1ES.Unofficial.PipelineTemplate.yml@1es`. But I just noticed that the pipeline run you have shared is in `azfunc/internal` project. Line 58 and 64 also using official templates.

kshyju: This pipeline is extending "1ES.Unofficial.PipelineTemplate" as this is considered as a non-prod pipeline. It appears that we have other pipelines which does the same thing (Extending "1ES.Unofficial.PipelineTemplate", but using jobs stored in  "official/jobs"). For example, this one: https://github.com/Azure/azure-functions-host/blob/dev/eng/ci/integration-tests.yml#L28-L38

We could reorganize them if needed. I am not opposed to that idea. Thoughts @jviau ?

jviau: The `official` vs non-official differentiation is something I borrowed from how dotnet laid out their templates. The primary idea was the separate what is and isn't safe to run in `azfunc/public`. Needing service connections is one such reason. But it can also be a business decision that a given template is for internal builds only. For example, generating artifacts (for now) is not expected to be ran in public. 

Might be better to rename that folder `internal`.

eng/ci/templates/official/jobs/run-coldstart.yml
```diff
@@ -0,0 +1,129 @@
...
+      arguments: -c Release -o $(hostOutputPath) -f net8.0 -p:PlaceholderSimulation=true 
+      workingDirectory: $(Build.SourcesDirectory)/src/WebJobs.Script.WebHost
+
+  - pwsh: |
+      Write-Host "Creating log directory: $(logsDirectory)"
+      New-Item -ItemType Directory -Path $(logsDirectory)
+    displayName: Create log directory
+
+  - ${{ if eq(parameters.os, 'Windows') }}:
+    - pwsh: |
```
jviau, line 97: Is there an issue using `pwsh` on Linux as well?

I see that comment below, but wonder if you can use the `env` parameter to workaround the issue.

eng/ci/templates/official/jobs/run-coldstart.yml
```diff
@@ -0,0 +1,129 @@
...
+  default: Windows
+  values:
+    - Windows
+    - Linux
+
+jobs:
+- job: ${{ parameters.functionAppName }}_${{ parameters.instanceId }}
+  displayName: ${{ parameters.description }} ${{ parameters.instanceId }}
+  pool:
+    name: 1es-pool-azfunc-benchmarking-large
```
jviau: We may want a smaller pool with more agents available if we will be running this with like 20 machines at a time. Unless you plan on making these run in sequence? Will take longer but will use far less VMs overall.

eng/ci/templates/official/jobs/run-coldstart.yml
kshyju: I couldn't get it working to start a **detached** new process (so that it continues to run after this task is done), using pwsh. This version appears to do it.

eng/ci/templates/official/jobs/run-coldstart.yml
kshyju: You meant large pool (size) with small VMs. correct? The current images used(-sd4) are 4 core machines. I am creating 2 core ones now and will push an iteration later today after testing on that version.

kshyju: Pushed an iteration which uses images with 2 cores.

kshyju:
Pipeline run with recent iteration: https://azfunc.visualstudio.com/internal/_build/results?buildId=203981&view=results

eng/ci/templates/official/jobs/process-coldstart.yml
```diff
@@ -0,0 +1,112 @@
...
+          | Metric  | Value     |
+          | :------ | :-------- |
+          | P50     | $p50 ms   |
+          | P75     | $p75 ms   |
+          | P99     | $p99 ms   |
+        "@
+        $summaryFilePath = "$(Build.ArtifactStagingDirectory)/summary.md"
+        $summaryContent | Out-File -FilePath $summaryFilePath
+        $summaryName = "ColdStart latency - $($env:OS) $($env:FUNCTION_APP_NAME)"
+        Write-Host "##vso[task.addattachment type=Distributedtask.Core.Summary;name=$summaryName;]$summaryFilePath"
```
kshyju, line 85: This will add a new tab to the pipeline run with this markup table. Example: https://azfunc.visualstudio.com/internal/_build/results?buildId=203981&view=ms.vss-build-web.run-extensions-tab

jviau: ❗ its that easy? Can we make a similar change to the crank pipeline?

eng/ci/templates/official/jobs/run-coldstart.yml
jviau: The image does not control core count, the pool does. `1es-pool-azfunc-benchmarking-large` is an 8 core pool (standard D8as v5)

jviau (approved):

eng/ci/templates/official/jobs/process-coldstart.yml
kshyju: Yea, I will update that pipeline for similar changes.

surgupta-msft (approved):

eng/ci/templates/official/jobs/run-coldstart.yml
kshyju: Oh, I didn't realize that. Yea, I updated the images now to be the originals (1es-ubuntu-22.04-benchmark-runner-vanilla, 1es-windows-2022-benchmark-runner-vanilla) and update the pool size to d4.

PR #10870, opened 2025-02-21
Add Response compression support

### LINKED ISSUE #7285: GZip on Azure Function Not Working For JSON Response
Hey everyone,

"I feel like I'm taking crazy pills", but I can't find the answer to my problem... I have a group of Azure Functions V3 that are serving up a REST API.  They are on an "Always On" App service plan of B1.

In front of our Azure Functions, we are using Azure API Management.  (Although this problem occurs when we make calls directly to the functions)

We have a larger API response that returns about 375k of JSON, and noticed that it wasn't being compressed (GZIP).  

image

Documents online seems to say it should just work by default, but mine isn't:

https://stackoverflow.com/questions/49339268/how-can-i-enable-gzip-compression-on-azure-functions-v2

Also other people seemed to have this issue previously: https://github.com/Azure/azure-functions-host/issues/4000

I have been searching online for hours, and I have even searched for ways to disable gzip, so I could find the location of the settings so I could verify it wasn't disabled, but to no avail.

I even thought maybe it's compressed but somehow I'm missing it, so I tested it on webpagetest.org:

image

Related question post: https://docs.microsoft.com/en-us/answers/questions/353996/gzip-on-azure-functions.html

  [1]: /answers/storage/attachments/86997-image.png
  [2]: /answers/storage/attachments/87022-image.png

kshyju:
resolves #7285

Currently, response compression is enabled on platform level for app service windows because Windows workers still use IIS. For Linux use case, Kestrel serves as the web server. This PR adds support for enabling response compression on host by leveraging the ASP.NET response compression middleware. Customers can opt in by setting the feature flag **"EnableResponseCompression"** to enable this behavior.  

For additional context, see [Compression on App Service Linux](https://azureossd.github.io/2022/07/13/Compression-on-App-Service-Linux).

### Pull request checklist

    * Otherwise: Link to backporting PR - will follow
    * [x] Otherwise: Backport tracked by issue/PR #issue_or_pr  - will follow

### Additional information

Additional PR information


src/WebJobs.Script.WebHost/Configuration/ResponseCompressionOptions.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Azure.WebJobs.Hosting;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Configuration
+{
+    internal sealed class ResponseCompressionOptions : IOptionsFormatter
```
surgupta-msft, line 10: curious if there is a specific reason for making it a `sealed` class?

surgupta-msft (approved):

surgupta-msft:
I have a question here. Could there be a scenario where users want to compress only those responses where Json is large and not for the smaller ones? In other words, could there be a need to enable this feature per function basis. Similar example here - https://github.com/Azure/azure-functions-host/issues/7285#issuecomment-2000410861

kshyju:
> I have a question here. Could there be a scenario where users want to compress only those responses where Json is large and not for the smaller ones? In other words, could there be a need to enable this feature per function basis. Similar example here - [#7285 (comment)](https://github.com/Azure/azure-functions-host/issues/7285#issuecomment-2000410861)

We are introducing this feature at the app level for the host. If a customer wants to implement compression for individual functions, they can do so within their function app, provided the language worker supports it. We do not plan to add support for per-function compression to the functions host.

src/WebJobs.Script.WebHost/Configuration/ResponseCompressionOptions.cs
kshyju: Sealed classes provide a slight performance benefit.  Any new classes we write should be sealed unless there is a need for a derived class. a sealed class can be unsealed when/if needed in the future.

surgupta-msft: awesome, that's great to know.

surgupta-msft (approved):

src/WebJobs.Script.WebHost/WebJobsApplicationBuilderExtension.cs
```diff
@@ -56,6 +61,12 @@ public static IApplicationBuilder UseWebJobsScriptHost(this IApplicationBuilder
                 builder.UseMiddleware<PlaceholderSpecializationMiddleware>();
             }
 
+            // Enable response compression only after PlaceholderSpecializationMiddleware, as it requires customer opt-in feature flag value.
+            builder.UseWhen(_ => responseCompressionOptions.EnableResponseCompression, config =>
```
jviau: Any reason you are using `OnChange` hooks and not just accessing `.CurrentValue.EnableResponseCompression` in here?

kshyju: Was trying to use same pattern used for other options above. I pushed an iteration in which I switched to using CurrentValue instead of onchange and local variable (only updated the new code to use this pattern)

jviau (approved):

kshyju:
@mattchenderson I don't believe we need to backport this to in-proc. .NET in-proc customers can enable response compression directly in their app code. Let me know if you disagree.

cajuncoding:
Why doesn't this explain (even a little) about how this is supposed to be enabled for a vanilla Azure Function running in the Isolated Process Model? How are we supposed to enable this feature flag? Will it work when running locally? Does it work when using `HttpResponseData`?

I have been very unsuccessful at getting this to work after days of trial and error, and of zero help from AI or any documentation. 😞

watri:
Hi @kshyju , I have add this EnableResponseCompression = true in my function app environment. Compression still does not work. are my steps correct?

kshyju:
To enable this feature, add `EnableResponseCompression` to the `AzureWebJobsFeatureFlags` app setting.

https://learn.microsoft.com/en-us/azure/azure-functions/functions-app-settings#azurewebjobsfeatureflags

image

plaa:
@kshyju Can you please update the `AzureWebJobsFeatureFlags` documentation you linked to?  This parameter works perfectly, but it's documented nowhere.  

The PR has checked "My changes do not require documentation changes" but I feel this should definitely be documented.  (Even better would be to enable it by default, could save huge amounts of bandwidth globally.)

PR #10884, opened 2025-02-26
Trigger code-mirror on all release branches

jviau:
Normalize code-mirror to trigger off all release branches.

### DIFF
--- a/eng/ci/code-mirror.yml
+++ b/eng/ci/code-mirror.yml
@@ -3,8 +3,7 @@ trigger:
     include:
     - dev
     - in-proc
-    - release/4.*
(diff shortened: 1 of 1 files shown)


liliankasem (approved):

liliankasem:
Can we include feature/ branches as well? We may eventually start using those again as we start our refactor work across our repos

jviau:
@liliankasem I would like to discuss feature branch usage before adding it as a glob. They aren't blocked from mirroring either - every branch can add themselves to this trigger list.

surgupta-msft (approved):

PR #10913, opened 2025-03-07
Patch/4.38.200

brettsam:
hotfix to bring #10911 and #10916 into 4.1038


jviau (approved):

liliankasem:
Let me know if you don't want this, but should we bring your other hotfix into this patch as well?

brettsam:
> Let me know if you don't want this, but should we bring your other hotfix into this patch as well?

Yes, just pushed it. Test issues were holding things up.

liliankasem (approved):

PR #10921, opened 2025-03-10
Platform release channel bundles resolution fix and additional logging

satvu:

Resolves bug in initial implementation (https://github.com/Azure/azure-functions-host/pull/10790) and adds additional logging.

### Pull request checklist

    * **Otherwise: Link to backporting PR** - https://github.com/Azure/azure-functions-host/pull/10922

### DIFF
--- a/src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
+++ b/src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
@@ -317,21 +317,24 @@ namespace Microsoft.Azure.WebJobs.Script.ExtensionBundle
             }
 
             // keep the latest version, log a notice
-            _logger.LogInformation("Unable to apply platform release channel configuration {platformReleaseChannelName}. Only one matching bundle version is available. {latestBundleVersion} will be used", _platformReleaseChannel, latest);
+            _logger.LogWarning("Unable to apply platform release channel configuration {platformReleaseChannelName}. Only one matching bundle version is available. {latestBundleVersion} will be used", _platformReleaseChannel, latest);
             return latest;
         }
 
         private NuGetVersion GetLatestBundleVersion(IList<NuGetVersion> orderedByDescBundlesList)
         {
             var latest = orderedByDescBundlesList.FirstOrDefault();
-            _logger.LogInformation("Applying platform release channel configuration {platformReleaseChannelName}. Bundle version {latest} will be used", _platformReleaseChannel, latest);
+            if (string.Equals(_platformReleaseChannel.ToUpper(), ScriptConstants.LatestPlatformChannelNameUpper))
+            {
+                _logger.LogInformation("Applying platform release channel configuration {platformReleaseChannelName}. Bundle version {latest} will be used", _platformReleaseChannel, latest);
+            }
             return latest;
         }
 
         private NuGetVersion HandleUnknownPlatformReleaseChannelName(IList<NuGetVersion> orderedByDescBundlesList)
         {
             var latest = GetLatestBundleVersion(orderedByDescBundlesList);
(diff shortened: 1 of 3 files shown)


src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -308,8 +308,11 @@ private NuGetVersion GetStandardOrExtendedBundleVersion(IList<NuGetVersion> orde
         {
             if (orderedByDescBundlesList.Count > 1)
             {
+                var previous = orderedByDescBundlesList[1];
+                _logger.LogInformation("Applying platform release channel configuration {platformReleaseChannelName}. Bundle version {previous} will be used", _platformReleaseChannel, previous);
```
soninaren: Should we print the latest one here too? 
something like using bundle version x instead of y.

soninaren (approved):

soninaren:
Left a nit comment.

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
satvu: Maybe:

`Applying platform release channel configuration {platformReleaseChannelName}. Previous bundle version {previous} will be used instead of latest version {latest}.`

soninaren: That looks good to me.

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -264,6 +263,7 @@ internal string FindBestVersionMatch(VersionRange versionRange, IEnumerable<stri
                 return version;
             }).Where(v => v != null).OrderByDescending(version => version.Version).ToList();
 
+            _logger.LogInformation("Platform release channel is set to: {platformReleaseChannelVersion}. Attempting to resolve bundle version for {bundleId} using the platform release channel configuration.", _platformReleaseChannel, bundleId);
```
satvu: Note: The platform release channel env var will be set by a specific subset of customers, but most won't use it. All logs related to platform release channel are `Information` level to avoid concern.

liliankasem: > but most won't use it

Can we only log when it is used to avoid noise?

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
liliankasem: Can we change all of these logsto debug or trace? Why do they need to be info

satvu: Here's the levels we currently use for the different extensions bundles logs (information and higher): https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Diagnostics/Extensions/ExtensionBundleLoggerExtension.cs. 

It logs non-fatal errors as Information as well (we have some non-fatal error scenarios covered in this PR's logs) and some additional things like the path the bundles were found at. These logs are also only logged on startup when extension bundles are loaded (so they won't be excessively noisy). 

I think we could lower the level for non-error logs that I'm adding here but wanted to note the existing logging behavior of this class.

jviau: To @liliankasem's point though, we have an over-usage of `LogInformation`. We should start considering lower verbosity for many of our logs.

soninaren (approved):

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
satvu: Results of offline discussion:

Keeping the logs informational in this case after tightening the scenarios/cases where they are logged. Bumped the non-fatal errors to warning.

jviau (approved):

PR #10932, opened 2025-03-17
Linux Consumption metrics publisher for Legion

mathewc:
This PR pulls in the new metrics component for Linux Consumption in Legion environments.

_History_: The original PR for this was [here](https://github.com/Azure/azure-functions-host/pull/9741), which was rebased by Vivek and team in the PR [here](https://github.com/Azure/azure-functions-host/pull/10750) which was iterated on for a while. The current PR is the culmination of all this - basically I took it back and resolved the final outstanding issues.

### Pull request checklist

    * https://github.com/Azure/azure-functions-host/pull/10942


src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -238,7 +238,15 @@ public static bool IsFlexConsumptionSku(this IEnvironment environment)
...
-            return environment.IsLinuxConsumptionOnLegion();
+            if (!environment.IsConsumptionOnLegion())
+            {
+                // not running on Legion, so not Flex Consumption
+                return false;
+            }
+
+            // If we're running on Legion and the app isn't CV1 on Legion,
+            // then it's Flex
+            return !environment.WebsiteSkuIsDynamic();
```
mathewc, line 250: Question for @balag0 on this - can we be certain that in all cases (placeholder or not), Linux Consumption on Legion will have the SKU set? We need to rely on this or some other environmental discriminator to distinguish between Flex and Linux Consumption on Legion.

The existing comments above on the IsConsumptionOnLegion check indicate that at least for Flex, the SKU might not always be set which necessitated that additional check rather than just relying on SKU. Do we have the same problem for Linux Consumption on Legion?

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -327,14 +335,49 @@ public static bool IsLinuxConsumptionOnAtlas(this IEnvironment environment)
...
         }
 
+        /// <summary>
+        /// Returns a value indicating whether the app is V1 Linux Consumption running on Legion.
+        /// </summary>
+        /// <param name="environment">The environment to verify.</param>
+        /// <returns><see cref="true"/> if the app is V1 Linux Consumption running on Legion; otherwise, <see cref="false"/>.</returns>
+        public static bool IsLinuxConsumptionOnLegion(this IEnvironment environment)
+        {
+            return IsConsumptionOnLegion(environment) && environment.WebsiteSkuIsDynamic();
```
mathewc, line 354: @balag0 Same question here - here we're also relying on SKU to make the determination. Is this correct and the best check we can be performing?

test/WebJobs.Script.Tests/Microsoft.Azure.WebJobs.Script.WebHost.deps.json
```diff
@@ -21,12 +21,19 @@
           "Microsoft.AspNetCore.Mvc.NewtonsoftJson": "8.0.1",
           "Microsoft.Azure.AppService.Middleware.Functions": "1.5.5",
           "Microsoft.Azure.AppService.Proxy.Client": "2.3.20240307.67",
-          "Microsoft.Azure.Functions.PythonWorker": "4.34.0",
+          "Microsoft.Azure.Functions.DotNetIsolatedNativeHost": "1.0.12",
+          "Microsoft.Azure.Functions.JavaWorker": "2.18.1",
+          "Microsoft.Azure.Functions.NodeJsWorker": "3.10.1",
+          "Microsoft.Azure.Functions.Platform.Metrics.LinuxConsumption": "1.0.4",
```
mathewc: There were deps file changes, so I took the new deps file output from CI and copied it into this baseline file. The only dependency change I made was to add a dependency to the new nuget package **Microsoft.Azure.Functions.Platform.Metrics.LinuxConsumption**. However, there are lots of additional changes in this file - need to determine if these are all expected.

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
vivekjilla: Thanks Mathew. For the SKU env variable, if needed, we can start adding the sku name env as part of docker image itself (only for Linux consumption on Legion images), so that placeholders will also have that value set. But just want a confirmation that this should be okay to do. cc @balag0

balag0: @yojagad is driving LC on Legion. so adding him here as well

>can we be certain that in all cases (placeholder or not)

all containers start in placeholder mode only in legion. 

>Linux Consumption on Legion will have the SKU set?

Yes 
WEBSITE_SKU_NAME = Dynamic
and
LEGION_SERVICE_HOST = 1
will be set on all Linux Consumption containers on legion

>at least for Flex, the SKU might not always be set

Correct. in flex the env vars set are slightly different
LEGION_SERVICE_HOST = 1
and
**WEBSITE_SKU** = FlexConsumption

@vivekjilla 
Agree it makes sense to just set this in the image directly. 
I think to standardize we can set

FlexConsumption Images
WEBSITE_SKU = FlexConsumption
WEBSITE_SKU_NAME = FlexConsumption
LEGION_SERVICE_HOST = 1

and
LinuxConsumption on Legion images
WEBSITE_SKU = Dynamic
WEBSITE_SKU_NAME = Dynamic
LEGION_SERVICE_HOST = 1

Note: Its a little confusing currently because
in other ( app service ) skus WEBSITE_SKU is the env. variable to represent sku so we started adding it, but it is only available after specialization.
since we need something in placeholder mode, we added WEBSITE_SKU. but we only did that for FlexConsumption since there was no scenario for LinuxConsumption

With the above changes to set WEBSITE_SKU from the start of container, its a small in change in behavior for LinuxConsumption images, so could you please do a quick check on how it is used in LinuxConsumption images to see if it can cause any issues :)

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -327,14 +335,49 @@ public static bool IsLinuxConsumptionOnAtlas(this IEnvironment environment)
...
+        /// <returns><see cref="true"/> if the sku is Dynamic; otherwise, <see cref="false"/>.</returns>
+        public static bool WebsiteSkuIsDynamic(this IEnvironment environment)
+        {
+            string value = environment.GetEnvironmentVariable(AzureWebsiteSku);
+            if (string.Equals(value, ScriptConstants.DynamicSku, StringComparison.OrdinalIgnoreCase))
+            {
+                return true;
+            }
+
+            // Linux Consumption v1 uses WEBSITE_SKU_NAME but is migrating to use WEBSTIE_SKU.
```
balag0: typo
WEBSTIE_SKU

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
balag0: based on my comment here
https://github.com/Azure/azure-functions-host/pull/10932#discussion_r2001235008

WEBSITE_SKU will be available in LinuxConsumption images only after specialization. using WEBSITE_SKU_NAME is better. but if we move the environment variables to docker files like @vivekjilla suggested, it would make things simpler/consistent

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
mathewc: Got it. So the following env vars are currently set:

- CV1 on Legion: WebsiteSkuName WEBSITE_SKU_NAME = Dynamic (since May 2023, code [here](https://dev.azure.com/msazure/One/_git/AAPT-Antares-Websites?path=/src/Hosting/Functions/FunctionsPodService/PodBuilder/FunctionsWorkerPodBuilderStrategy.cs&version=GBdev&line=60&lineEnd=61&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents))
- Flex on Legion: AzureWebsiteSku WEBSITE_SKU = FlexConsumption (since Jan 2024, code [here](https://dev.azure.com/msazure/One/_git/AAPT-Antares-Websites?path=/src/Hosting/Functions/FunctionsPodService/RegionalPool/AntaresConfig/PoolGroupDefinitionExtensions.cs&version=GBdev&line=301&lineEnd=302&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents))

Therefore, it seems the code as I have now in the PR for distinguishing between Linux Consumption and Flex on Legion is valid, correct? The code is actually a little more defensive than it needs to be, since at this point some of the extra checks I'm doing may no longer be necessary, but I'll leave them as is for now.

So why are you saying we need to add SKU to the Docker image, since it's already being added by the platform to the placeholders?

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
mathewc: But you shared with me the placeholder env vars for Linux Consumption on Legion [here](https://dev.azure.com/msazure/One/_git/AAPT-Antares-Websites?path=/src/Hosting/Functions/FunctionsPodService/PodBuilder/FunctionsWorkerPodBuilderStrategy.cs&version=GBdev&line=60&lineEnd=61&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents), so SKU is already present and can be relied on as I'm doing here?

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
mathewc: Fixed, thanks

src/WebJobs.Script.WebHost/Configuration/LinuxConsumptionLegionMetricsPublisherOptions.cs
```diff
@@ -0,0 +1,21 @@
...
+    public sealed class LinuxConsumptionLegionMetricsPublisherOptions
+    {
+        internal const int DefaultMetricsPublishIntervalMS = 30 * 1000;
+
+        public LinuxConsumptionLegionMetricsPublisherOptions()
+        {
+            MetricsPublishIntervalMS = DefaultMetricsPublishIntervalMS;
+        }
+
+        public int MetricsPublishIntervalMS { get; set; }
```
jviau:
```suggestion
        public int MetricsPublishIntervalMS { get; set; } = DefaultMetricsPublishIntervalMS;
```

Can remove the ctor then

src/WebJobs.Script.WebHost/Configuration/LinuxConsumptionLegionMetricsPublisherOptionsSetup.cs
```diff
@@ -0,0 +1,23 @@
...
+
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Configuration
+{
+    public class LinuxConsumptionLegionMetricsPublisherOptionsSetup : IConfigureOptions<LinuxConsumptionLegionMetricsPublisherOptions>
+    {
+        private readonly IEnvironment _environment;
+
+        public LinuxConsumptionLegionMetricsPublisherOptionsSetup(IEnvironment environment)
```
jviau, line 12: Please use `IConfiguration` over `IEnvironment` whenever possible. It will have environment variables mapped to it.

`configuration[EnvironmentSettingNames.ContainerName]`

src/WebJobs.Script.WebHost/Metrics/LegionMetricsFileManager.cs
```diff
@@ -0,0 +1,112 @@
...
+
+        public LegionMetricsFileManager(string metricsFilePath, IFileSystem fileSystem, ILogger logger, int maxFileCount)
+        {
+            _fileSystem = fileSystem ?? throw new ArgumentNullException(nameof(fileSystem));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            MetricsFilePath = metricsFilePath;
+            _maxFileCount = maxFileCount;
+        }
+
+        internal string MetricsFilePath { get; set;  }
```
jviau, line 36: Is there a reason, outside of testing, for this to not be readonly?

If it is just about configuring this during testing, can you not add custom options configuration in testing to get this path mapped?

I would prefer we make this readonly if possible. If not we need to ensure there are not race conditions in this class, which there currently are. `MetricsFilePath` may be changed after the `string.IsNullOrEmpty(MetricsFilePath)` checks pass. To guard against this you will need to assign `MetricsFilePath` to a local variable, and then use that for the entire publish sequence.

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
```diff
@@ -267,6 +269,11 @@ private static void AddStandbyServices(this IServiceCollection services)
...
         private static void AddLinuxContainerServices(this IServiceCollection services)
         {
+            if (SystemEnvironment.Instance.IsLinuxConsumptionOnLegion())
+            {
+                services.AddLinuxConsumptionMetricsServices();
+            }
+
             services.AddSingleton<IHostedService>(s =>
             {
                 var environment = s.GetService<IEnvironment>();
```
vivekjilla, line 279: Don't we need to return anything here for IHostedService for LinuxConsumptionOnLegion?

mathewc: Yes, good point. Currently in dev, the IsFlexConsumptionsSku check below will pass for both Flex and Linux Consumption on Legion. In this PR I made IsFlexConsumptionSku a little more restrictive so it returns true ONLY for Flex.

vivekjilla (approved):

src/WebJobs.Script.WebHost/Metrics/LegionMetricsFileManager.cs
```diff
@@ -0,0 +1,112 @@
...
+        {
+            if (string.IsNullOrEmpty(MetricsFilePath))
+            {
+                return false;
+            }
+
+            // ensure the directory exists
+            _fileSystem.Directory.CreateDirectory(MetricsFilePath);
+
+            var metricsDirectoryInfo = _fileSystem.DirectoryInfo.FromDirectoryName(MetricsFilePath);
```
jviau: The `CreateDirectory` call above returns the directory info already, does it not?

src/WebJobs.Script.WebHost/Metrics/LegionMetricsFileManager.cs
```diff
@@ -0,0 +1,112 @@
...
+
+        private bool PrepareDirectoryForFile()
+        {
+            if (string.IsNullOrEmpty(MetricsFilePath))
+            {
+                return false;
+            }
+
+            // ensure the directory exists
+            _fileSystem.Directory.CreateDirectory(MetricsFilePath);
```
jviau: Have you verified this doesn't throw if it already exists?

src/WebJobs.Script.WebHost/Metrics/LegionMetricsFileManager.cs
```diff
@@ -0,0 +1,112 @@
...
+        public async Task PublishMetricsAsync(object metrics)
+        {
+            string fileName = string.Empty;
+
+            try
+            {
+                bool metricsPublishEnabled = !string.IsNullOrEmpty(MetricsFilePath);
+                if (metricsPublishEnabled && !PrepareDirectoryForFile())
+                {
+                    return;
```
jviau, line 90: Should we log something if we couldn't publish metrics?

src/WebJobs.Script.WebHost/Metrics/LegionMetricsFileManager.cs
```diff
@@ -0,0 +1,112 @@
...
+                }
+
+                var settings = new JsonSerializerSettings
```
jviau: Recommend moving these to a readonly field and re-using.

jviau: Recommend moving these to a readonly field and re-using.

src/WebJobs.Script.WebHost/Metrics/LegionMetricsFileManager.cs
```diff
@@ -0,0 +1,112 @@
...
+                {
+                    NullValueHandling = NullValueHandling.Ignore
+                };
+
+                string metricsContent = JsonConvert.SerializeObject(metrics, settings);
+                _logger.PublishingMetrics(metricsContent);
+
+                if (metricsPublishEnabled)
+                {
+                    fileName = $"{Guid.NewGuid().ToString().ToLower()}.json";
```
jviau: Any particular reason this needs to be in lowercase?

```suggestion
                    fileName = $"{Guid.NewGuid().ToString().ToLowerInvariant()}.json";
```

(or remove the ToLower call altogether)

src/WebJobs.Script.WebHost/Metrics/LegionMetricsFileManager.cs
```diff
@@ -0,0 +1,112 @@
...
+            {
+                return true;
+            }
+
+            // we're at or over limit
+            // delete enough files that we have space to write a new one
+            int numToDelete = files.Count - _maxFileCount + 1;
+            var filesToDelete = files.Take(numToDelete).ToArray();
+
+            _logger.LogDebug($"Deleting {filesToDelete.Length} metrics file(s).");
```
jviau, line 60: Do we want to use structured logging?

src/WebJobs.Script.WebHost/Metrics/LinuxContainerLegionMetricsPublisher.cs
```diff
@@ -0,0 +1,265 @@
...
+        private readonly string _containerName;
+
+        private IMetricsLogger _metricsLogger;
+        private TimeSpan _metricPublishInterval;
+        private Process _process;
+        private Timer _processMonitorTimer;
+        private Timer _metricsPublisherTimer;
+        private bool _initialized = false;
+
+        public LinuxContainerLegionMetricsPublisher(IEnvironment environment, IOptionsMonitor<StandbyOptions> standbyOptions, IOptions<LinuxConsumptionLegionMetricsPublisherOptions> options, ILogger<LinuxContainerLegionMetricsPublisher> logger, IFileSystem fileSystem, ILinuxConsumptionMetricsTracker metricsTracker, IScriptHostManager scriptHostManager, int? metricsPublishIntervalMS = null)
```
jviau: nit: please break up among multiple lines

src/WebJobs.Script.WebHost/Metrics/LinuxContainerLegionMetricsPublisher.cs
```diff
@@ -0,0 +1,265 @@
...
+        {
+            _standbyOptions = standbyOptions ?? throw new ArgumentNullException(nameof(standbyOptions));
+            _environment = environment ?? throw new ArgumentNullException(nameof(environment));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            _metricsTracker = metricsTracker ?? throw new ArgumentNullException(nameof(metricsTracker));
+            _scriptHostManager = scriptHostManager ?? throw new ArgumentNullException(nameof(scriptHostManager));
+            _containerName = options.Value.ContainerName;
+
+            // Set this to 15 minutes worth of files
+            _metricPublishInterval = TimeSpan.FromMilliseconds(metricsPublishIntervalMS ?? options.Value.MetricsPublishIntervalMS);
```
jviau, line 50: Can we just take in a `TimeSpan` directly in the ctor?

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
```diff
@@ -66,13 +68,13 @@ public FlexConsumptionMetricsPublisher(IEnvironment environment, IOptionsMonitor
...
 
-        internal string MetricsFilePath { get; set; }
+        internal LegionMetricsFileManager MetricsFileManager => _metricsFileManager;
 
         public void Start()
         {
             Initialize();
 
-            _logger.LogInformation($"Starting metrics publisher (AlwaysReady={IsAlwaysReady}, MetricsPath='{MetricsFilePath}').");
+            _logger.LogInformation($"Starting metrics publisher (AlwaysReady={IsAlwaysReady}, MetricsPath='{_metricsFileManager.MetricsFilePath}').");
```
jviau, line 77: Any reason we aren't using structured logging?

jviau (approved):

src/WebJobs.Script.WebHost/Configuration/LinuxConsumptionLegionMetricsPublisherOptions.cs
mathewc: Done

src/WebJobs.Script.WebHost/Configuration/LinuxConsumptionLegionMetricsPublisherOptionsSetup.cs
mathewc: Gonna leave this as is for now to avoid having to rewrite tests. Many of our setups are using IEnvironment.

src/WebJobs.Script.WebHost/Metrics/LegionMetricsFileManager.cs
mathewc: No, it's only for testing. We are using internal access in many other places in these metrics tests right now. While you're technically correct that it does open the possibility of race conditions, the tests are only changing this value on a new instance, prior to writing any data. So it's up to the test to be well behaved.

I've added a note on the member that internal access is for testing only

src/WebJobs.Script.WebHost/Metrics/LegionMetricsFileManager.cs
mathewc: Good catch. Fixed

src/WebJobs.Script.WebHost/Metrics/LegionMetricsFileManager.cs
mathewc: Yes I have. Also the method doc says the semantics is "create if not exists"

src/WebJobs.Script.WebHost/Metrics/LegionMetricsFileManager.cs
mathewc: The PrepareDirectoryForFile method already logs in cases when it needs to (e.g. if we can't log because we're at file limit and file deletions are failing).

src/WebJobs.Script.WebHost/Metrics/LegionMetricsFileManager.cs
mathewc: Done

mathewc: Done

src/WebJobs.Script.WebHost/Metrics/LegionMetricsFileManager.cs
mathewc: Made the change you suggested. I can't remember now why I lower cased it, but I'll leave as is since we have Legion consuming these files.

src/WebJobs.Script.WebHost/Metrics/LinuxContainerLegionMetricsPublisher.cs
mathewc: Done

src/WebJobs.Script.WebHost/Metrics/LinuxContainerLegionMetricsPublisher.cs
mathewc: We're conditionally defaulting this based on options which is expressed in MS, so I used the same for the ctor param

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
mathewc: No reason, other than I usually don't I guess. Gonna leave these as is for now.

src/WebJobs.Script.WebHost/Metrics/LegionMetricsFileManager.cs
mathewc: Not right now - see my other comment :)

PR #10938, opened 2025-03-22
making GetFunctionMetadataAsync idempotent

brettsam:
Bug found during CRI investigation that can lead to "Binding names must be unique" error.

Scenario was that customer had one instance out of many that, during startup, hit "Binding names must be unique" across all of their functions on that instance. This meant that we didn't enable any functions on this one instance and they ended up getting 404s any time the FE sent requests to it (all other instances were fine).

What I found on this one instance was that the platform specialized the app while it was in the middle of running the "infrastructure warmup" call. This caused specialization to occur when the process was not completely ready, which resulted in one extra restart request.

We seem to only ever expect a restart to happen once, but since it happened twice, we ran through the GetFunctionMetadataAsync() method twice with the same worker-indexed functions. This method has a bug where it modifies existing objects and creates bindings every time you call it. This resulted in duplicate bindings here.

This entire flow is being actively redesigned right now, so I did a very targeted fix for this one scenario so we don't hit it again while working on the new design.

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
+++ b/src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
@@ -251,6 +251,10 @@ namespace Microsoft.Azure.WebJobs.Script
         {
             HashSet<string> bindingNames = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
 
+            // This method takes the RawBindings and adds them to the FunctionMetadata object. It's possible
+            // to call this twice, and we don't want to duplicate the bindings in that case.
+            function.Bindings.Clear();
+
             foreach (string binding in rawBindings)
             {
                 var deserializedObj = JsonConvert.DeserializeObject<JObject>(binding, _dateTimeSerializerSettings);
--- a/test/WebJobs.Script.Tests/WorkerFunctionMetadataProviderTests.cs
+++ b/test/WebJobs.Script.Tests/WorkerFunctionMetadataProviderTests.cs
@@ -269,15 +269,27 @@ namespace Microsoft.Azure.WebJobs.Script.Tests
 
             var workerConfigs = new List<RpcWorkerConfig>();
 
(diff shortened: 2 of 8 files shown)


src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
```diff
@@ -251,6 +251,10 @@ internal FunctionMetadata ValidateBindings(IEnumerable<string> rawBindings, Func
         {
             HashSet<string> bindingNames = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
 
+            // This method takes the RawBindings and adds them to the FunctionMetadata object. It's possible
+            // to call this twice, and we don't want to duplicate the bindings in that case.
+            function.Bindings.Clear();
```
brettsam, line 256: this is the fix... as noted, there's better ways (design-wise) to do this, but this is all actively being redesigned so I didn't want to go too far here. This is a targeted fix for this one scenario that we know customers have hit.

liliankasem (approved):

test/WebJobs.Script.Tests/WorkerFunctionMetadataProviderTests.cs
```diff
@@ -221,5 +222,74 @@ public void ValidateFunctionMetadata_IsoStringNotAltered()
...
+                Assert.Equal("node", function.Language);
+                Assert.Collection(function.Bindings, binding =>
+                {
+                    Assert.Equal("httpTrigger", binding.Type);
+                    Assert.Equal("req", binding.Name);
+                    Assert.Equal(BindingDirection.In, binding.Direction);
+                });
+            }
+
+            AssertFunction(function1);
```
brettsam, line 291: this test failed before the fix... there would have been multiple "httpTrigger" bindings, but we now check that there's only one

jviau (approved):

PR #10967, opened 2025-03-27
Warn if .azurefunctions folder does not exist

### LINKED ISSUE #2759: Warn if worker starts and cannot find the .azurefunctions folder
### Description

If the metadata cannot be located due to a problem with artifact generation, deployment, etc. The worker should issue a warning indicating a problem with the application payload.

surgupta-msft:

resolves https://github.com/Azure/azure-functions-dotnet-worker/issues/2759

Creating a new Hosting Service to asynchronously validate the functionapp payload and check the existence of `.azurefunctions` folder

### Pull request checklist

### Additional information

Additional PR information


src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
```diff
@@ -132,6 +132,13 @@ public async Task<IEnumerable<Type>> GetExtensionsStartupTypesAsync()
             }
             else
             {
+                string systemPath = Path.Combine(_rootScriptPath, ScriptConstants.AzureFunctionsSystemDirectoryName);
+
+                if (!Directory.Exists(systemPath) && isDotnetIsolatedApp)
```
RohitRanjanMS: Consider checking isDotnetIsolatedApp first, if it's false, there is no need to compute the path or check if the directory exists

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
```diff
@@ -132,6 +132,13 @@ public async Task<IEnumerable<Type>> GetExtensionsStartupTypesAsync()
             }
             else
             {
+                string systemPath = Path.Combine(_rootScriptPath, ScriptConstants.AzureFunctionsSystemDirectoryName);
+
+                if (isDotnetIsolatedApp && !Directory.Exists(systemPath))
+                {
+                    _logger.NoAzureFunctionsFolder(systemPath);
```
RohitRanjanMS: `systemPath` might be confusing since it includes _rootScriptPath, and customers may not understand what it represents. Can we use something like "package root" instead? I used this for the host validation error.

src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
```diff
@@ -184,14 +184,19 @@ internal static class LoggerExtension
...
         private static readonly Action<ILogger, Exception> _noHostJsonFile =
-    LoggerMessage.Define(LogLevel.Information,
-        new EventId(339, nameof(NoHostJsonFile)),
-        "No functions were found. This can occur before you deploy code to your function app or when the host.json file is missing from the most recent deployment. Make sure that your deployment package includes the host.json file in the root of the package. For deployment package requirements, see https://aka.ms/functions-deployment-technologies.");
+            LoggerMessage.Define(LogLevel.Information,
+            new EventId(339, nameof(NoHostJsonFile)),
+            "No functions were found. This can occur before you deploy code to your function app or when the host.json file is missing from the most recent deployment. Make sure that your deployment package includes the host.json file in the root of the package. For deployment package requirements, see https://aka.ms/functions-deployment-technologies.");
+
+        private static readonly Action<ILogger, string, Exception> _noAzureFunctionsFolderIssue =
+            LoggerMessage.Define<string>(LogLevel.Warning,
```
RohitRanjanMS, line 197: Is Warning the appropriate level here, or should we log this as an error instead?

test/WebJobs.Script.Tests/ScriptStartupTypeDiscovererTests.cs
```diff
@@ -790,6 +790,58 @@ public async Task GetExtensionsStartupTypes_WorkerRuntimeNotSetForNodeApp_LoadsE
...
+                factory.AddProvider(testLoggerProvider);
+                var testLogger = factory.CreateLogger<ScriptStartupTypeLocator>();
+
+                var mockExtensionBundleManager = new Mock<IExtensionBundleManager>();
+                mockExtensionBundleManager.Setup(e => e.IsExtensionBundleConfigured()).Returns(false);
+                mockExtensionBundleManager.Setup(e => e.GetExtensionBundleBinPathAsync()).Returns(Task.FromResult(binPath));
+                mockExtensionBundleManager.Setup(e => e.IsLegacyExtensionBundle()).Returns(false);
+                mockExtensionBundleManager.Setup(e => e.GetExtensionBundleDetails()).Returns(Task.FromResult(GetBundleDetails()));
+
+                RpcWorkerConfig nodeWorkerConfig = new RpcWorkerConfig() { Description = TestHelpers.GetTestWorkerDescription("node", "none", false) };
```
RohitRanjanMS: `nodeWorkerConfig ` isn't used in the test. I like the idea of including non-Isolated worker as well.

test/WebJobs.Script.Tests/ScriptStartupTypeDiscovererTests.cs
```diff
@@ -790,6 +790,58 @@ public async Task GetExtensionsStartupTypes_WorkerRuntimeNotSetForNodeApp_LoadsE
...
+
+                // Act
+                var types = await discoverer.GetExtensionsStartupTypesAsync();
+
+                //Assert
+                var traces = testLoggerProvider.GetAllLogMessages();
+                var traceMessage = traces.FirstOrDefault(val => val.EventId.Name.Equals("ScriptStartNotLoadingExtensionBundle"));
+                bool loadingExtensionBundle = traceMessage == null;
+
+                Assert.True(traces.Any(m => m.FormattedMessage.Contains($"Could not find the .azurefunctions folder in the path")));
```
RohitRanjanMS: Instead of using `Contains`, we can assert on the eventId `NoAzureFunctionsFolder`.

test/WebJobs.Script.Tests/ScriptStartupTypeDiscovererTests.cs
```diff
@@ -790,6 +790,58 @@ public async Task GetExtensionsStartupTypes_WorkerRuntimeNotSetForNodeApp_LoadsE
...
+                Directory.Move(directory.Path, directory.Path + "test");
+                var discoverer = new ScriptStartupTypeLocator(directory.Path, testLogger, mockExtensionBundleManager.Object, mockFunctionMetadataManager, testMetricsLogger, languageWorkerOptions, optionsWrapper);
+
+                // Act
+                var types = await discoverer.GetExtensionsStartupTypesAsync();
+
+                //Assert
+                var traces = testLoggerProvider.GetAllLogMessages();
+                var traceMessage = traces.FirstOrDefault(val => val.EventId.Name.Equals("ScriptStartNotLoadingExtensionBundle"));
+                bool loadingExtensionBundle = traceMessage == null;
```
RohitRanjanMS: `traceMessage` and `loadingExtensionBundle ` are not used for assertions.

test/WebJobs.Script.Tests/ScriptStartupTypeDiscovererTests.cs
```diff
@@ -790,6 +790,58 @@ public async Task GetExtensionsStartupTypes_WorkerRuntimeNotSetForNodeApp_LoadsE
...
 
+        [Fact]
+        public async Task GetExtensionsStartupTypes_DotnetIsolated_InnerBuild()
+        {
+            var vars = new Dictionary<string, string>();
+
+            using (var directory = GetTempDirectory())
+            using (var env = new TestScopedEnvironmentVariable(vars))
+            {
+                Environment.SetEnvironmentVariable(EnvironmentSettingNames.FunctionWorkerRuntime, "dotnet-isolated");
```
RohitRanjanMS: Consider cleaning up env variable after test.

RohitRanjanMS (dismissed):

src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
RohitRanjanMS: I just realized that this log occurs before the AppInsights pipeline is initialized, meaning it will go through the deferred logger flow. To ensure it reaches AppInsights, it needs to be logged as an error—anything below that level is ignored.

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
```diff
@@ -132,6 +132,13 @@ public async Task<IEnumerable<Type>> GetExtensionsStartupTypesAsync()
             }
             else
             {
+                string systemPath = Path.Combine(_rootScriptPath, ScriptConstants.AzureFunctionsSystemDirectoryName);
```
kshyju: No need to execute this line if `isDotnetIsolatedApp` is false. You may inline this in line 137.

src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
```diff
@@ -184,14 +184,19 @@ internal static class LoggerExtension
...
+            new EventId(338, nameof(NoAzureFunctionsFolder)),
+            "Could not find the .azurefunctions folder in the path: {path}. Make sure that your deployment package includes the .azurefunctions folder. For deployment package requirements, see https://aka.ms/functions-deployment-technologies.");
```
kshyju: If customers might see this in their AI logs, we should consider changing the value of the "path" variable in the message to something more generic—like "deployed function app artifacts"

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
kshyju: Will this get logged during placeholder mode?

kshyju: Yea, I think the message customer sees should include more generic term like "package root" or "deployed function app artifacts".

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
surgupta-msft: I need this path to check if the directory exists in the next line. Are you suggesting to do something else?

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
surgupta-msft: Yes, good catch. Added a check to avoid logging in placeholder mode.

src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
surgupta-msft: @mattchenderson @fabiocav Do you have any thoughts here? Is there any concern with making it as an error log (as the [issue](https://github.com/Azure/azure-functions-dotnet-worker/issues/2759) says about warning)? Or should we log this in a different place where it can come up as warning log?

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
RohitRanjanMS: I think this is what @kshyju is recommending-

`if (isDotnetIsolatedApp && 
    !SystemEnvironment.Instance.IsPlaceholderModeEnabled() && !Directory.Exists(Path.Combine(_rootScriptPath, ScriptConstants.AzureFunctionsSystemDirectoryName)))
{
    _logger.NoAzureFunctionsFolder();
}`

src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
RohitRanjanMS: Considering the impact of the folder being missing, I believe it should be logged as an error instead.

src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
```diff
@@ -184,14 +184,19 @@ internal static class LoggerExtension
...
-    LoggerMessage.Define(LogLevel.Information,
-        new EventId(339, nameof(NoHostJsonFile)),
-        "No functions were found. This can occur before you deploy code to your function app or when the host.json file is missing from the most recent deployment. Make sure that your deployment package includes the host.json file in the root of the package. For deployment package requirements, see https://aka.ms/functions-deployment-technologies.");
+            LoggerMessage.Define(LogLevel.Information,
+            new EventId(339, nameof(NoHostJsonFile)),
+            "No functions were found. This can occur before you deploy code to your function app or when the host.json file is missing from the most recent deployment. Make sure that your deployment package includes the host.json file in the root of the package. For deployment package requirements, see https://aka.ms/functions-deployment-technologies.");
+
+        private static readonly Action<ILogger, Exception> _noAzureFunctionsFolderIssue =
+            LoggerMessage.Define(LogLevel.Warning,
+            new EventId(338, nameof(NoAzureFunctionsFolder)),
```
liliankasem: ID should be new i.e. 340

surgupta-msft: Good catch. Fixed it.

liliankasem (approved):

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
```diff
@@ -129,6 +129,13 @@ public async Task<IEnumerable<Type>> GetExtensionsStartupTypesAsync()
             }
             else
             {
+                if (isDotnetIsolatedApp &&
```
fabiocav: Discussed offline. We need to make sure we're not logging warnings/errors for supported scenarios.

Portal creation flows will initialize a Function App without a deployed payload, and we want to make sure users are not greeted with an error upon creation. I've discussed some options with @surgupta-msft, leveraging some of the validation work done by @RohitRanjanMS

fabiocav (dismissed):

src/WebJobs.Script/Host/FunctionAppValidationService.cs
```diff
@@ -0,0 +1,56 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Script.Diagnostics.Extensions;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.WebJobs.Script.Host
+{
+    public sealed class FunctionAppValidationService : IHostedService
```
kshyju: Use `BackgroundService` and override `ExecuteAsync` method. The service will be disposed after `ExecuteAsync` returns. No need to keep the service after the folder check is done.

https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.hosting.backgroundservice?view=net-9.0-pp

src/WebJobs.Script/Host/FunctionAppValidationService.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            _scriptOptions = scriptOptions ?? throw new ArgumentNullException(nameof(scriptOptions));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            _environment = environment ?? throw new ArgumentNullException(nameof(environment));
+        }
+
+        public async Task StartAsync(CancellationToken cancellationToken)
+        {
+            // Adding a delay to ensure that this validation does not impact the cold start performance
+            _ = Task.Delay(_initializationDelay, cancellationToken).ContinueWith(t => Validate());
```
kshyju: Will the `Validate` gets executed after specialization? Is this service created/started after specialization or only when webhost starts (or in placeholder mode)?

src/WebJobs.Script/Host/FunctionAppValidationService.cs
```diff
@@ -0,0 +1,56 @@
...
+            IEnvironment environment)
+        {
+            _scriptOptions = scriptOptions ?? throw new ArgumentNullException(nameof(scriptOptions));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            _environment = environment ?? throw new ArgumentNullException(nameof(environment));
+        }
+
+        public async Task StartAsync(CancellationToken cancellationToken)
+        {
+            // Adding a delay to ensure that this validation does not impact the cold start performance
```
kshyju: Please use `Utility.ExecuteAfterColdStartDelay`

test/WebJobs.Script.Tests/Description/FunctionAppValidationServiceTests.cs
```diff
@@ -0,0 +1,149 @@
...
+        {
+            _testLoggerProvider.ClearAllLogMessages();
+
+            var service = new FunctionAppValidationService(
+                _testLogger,
+                _scriptOptionsMock.Object,
+                new TestEnvironment());
+
+            // Act
+            await service.StartAsync(CancellationToken.None);
```
kshyju, line 55: I think a more valuable test would be to add an E2E test which simulates specialization. Then you can remove this test class altogether (I am not convinced this test class will add any additional value after adding the above-mentioned test - I could be wrong)

src/WebJobs.Script/Host/FunctionAppValidationService.cs
surgupta-msft: it will get executed after specialization.

test/WebJobs.Script.Tests/Description/FunctionAppValidationServiceTests.cs
surgupta-msft: Added the test. I think unit tests are good too along with E2E tests.

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -175,6 +176,7 @@ public static IHostBuilder AddScriptHostCore(this IHostBuilder builder, ScriptAp
 
             builder.ConfigureServices((context, services) =>
             {
+                services.AddSingleton<IHostedService, FunctionAppValidationService>();
```
kshyju: Do we need this service to be running in placeholder mode? You can register this only after specialization.

```
if (!SystemEnvironment.Instance.IsPlaceholderModeEnabled())
{
    services.AddHostedService<FunctionAppValidationService>();
}
```

src/WebJobs.Script/Host/FunctionAppValidationService.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+            _scriptOptions = scriptOptions ?? throw new ArgumentNullException(nameof(scriptOptions));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            _environment = environment ?? throw new ArgumentNullException(nameof(environment));
+        }
+
+        protected override async Task ExecuteAsync(CancellationToken cancellationToken)
+        {
+            // Adding a delay to ensure that this validation does not impact the cold start performance
+            Utility.ExecuteAfterColdStartDelay(_environment, Validate, cancellationToken);
```
kshyju: You can move some of the checks currently done in the `Validate` method (e.g., `IsPlaceholderModeEnabled`) to this method, before calling `Utility.ExecuteAfterColdStartDelay`. If those conditions aren't met, there's no need to create and run a Timer (`ExecuteAfterColdStartDelay` uses `Task.Delay`, which internally uses a `Timer` instance).

src/WebJobs.Script/Host/FunctionAppValidationService.cs
```diff
@@ -0,0 +1,51 @@
...
+
+namespace Microsoft.Azure.WebJobs.Script.Host
+{
+    public sealed class FunctionAppValidationService : BackgroundService
+    {
+        private readonly IEnvironment _environment;
+        private readonly ILogger<FunctionAppValidationService> _logger;
+        private readonly IOptions<ScriptJobHostOptions> _scriptOptions;
+
+        public FunctionAppValidationService(
```
mattchenderson, line 26: This name is confusingly general. Can we make it more specific to the scenario?

src/WebJobs.Script/Host/FunctionAppValidationService.cs
```diff
@@ -0,0 +1,51 @@
...
+            await Task.CompletedTask;
+        }
+
+        internal void Validate()
+        {
+            if (!_environment.IsPlaceholderModeEnabled() &&
+                !_scriptOptions.Value.IsDefaultHostConfig &&
+                !_scriptOptions.Value.IsStandbyConfiguration &&
+                Utility.IsDotnetIsolatedApp(environment: _environment) &&
+                !Directory.Exists(Path.Combine(_scriptOptions.Value.RootScriptPath, ScriptConstants.AzureFunctionsSystemDirectoryName)))
```
mattchenderson: Should there be any binding between this and some other version info, config, or legacy indication that the isolated worker model is in use? If an app has been deployed with a much older version of the isolated worker model, is it possible that they might still run afoul of this even though they have a valid deployed payload?

RohitRanjanMS: A common cause of this issue is an incorrect zip file. If the zip wasn’t created at the root, .azurefunctions folder may exist, but not at the correct path. The log should clearly indicate the expected path, may be something like "Make sure that your deployment package includes the .azurefunctions folder at the root of the package"

RohitRanjanMS: Another option is to check the subfolders for the presence of the .azurefunctions folder. If it exists, we can confidently inform the customer that the deployment package is incorrect and provide guidance on how to fix it.

src/WebJobs.Script/Host/FunctionAppValidationService.cs
RohitRanjanMS: Could we make this run conditionally? For example, only trigger this validation if loading the app/functions fails.

src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
```diff
@@ -184,14 +184,19 @@ internal static class LoggerExtension
...
+            LoggerMessage.Define(LogLevel.Information,
+            new EventId(339, nameof(NoHostJsonFile)),
+            "No functions were found. This can occur before you deploy code to your function app or when the host.json file is missing from the most recent deployment. Make sure that your deployment package includes the host.json file in the root of the package. For deployment package requirements, see https://aka.ms/functions-deployment-technologies.");
+
+        private static readonly Action<ILogger, Exception> _noAzureFunctionsFolder =
+            LoggerMessage.Define(LogLevel.Warning,
+            new EventId(340, nameof(NoAzureFunctionsFolder)),
+            "Could not find the .azurefunctions folder in the deployed function app artifacts. Make sure that your deployment package includes the .azurefunctions folder. For deployment package requirements, see https://aka.ms/functions-deployment-technologies.");
```
RohitRanjanMS: We should update this log to cover the in-proc scenario you described. If an in-proc setup is mistakenly configured as isolated, the message could mislead the user.

src/WebJobs.Script/Host/FunctionAppValidationService.cs
surgupta-msft: I think the idea here is that this service can now be used for any function app payload related validations that don't need to be done synchronously. Hence, I kept a generic name. I can try other options such as `FunctionAppPayloadValidationService` or `FunctionAppArtifactsValidationService` or something similar, if you have any suggestions.

src/WebJobs.Script/Host/FunctionAppValidationService.cs
surgupta-msft: Do you mean that there were older versions of the isolated model that did not require the `.azurefunctions` folder to work correctly? I'm not aware of this scenario. Do you have any additional information that can help me repro the scenario and then add appropriate checks for it.

surgupta-msft: I don't think that's a valid scenario for this work. We create .azurefunctions directory when building an isolated app as part of dotnet worker sdk inner build - [link](https://github.com/Azure/azure-functions-dotnet-worker/blob/cfb4d9cba288df1bf1e14572ac5af46600cfd6d3/sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.targets#L25). Also, as per this comment from Matthew - [link](https://github.com/Azure/azure-functions-dotnet-worker/issues/2759#issuecomment-2718880253), the cause of this issue is the breaking change to upload-artifact in GitHub Action which is omitting the hidden files.

src/WebJobs.Script/Host/FunctionAppValidationService.cs
surgupta-msft: @mattchenderson the comment [here](https://github.com/Azure/azure-functions-dotnet-worker/issues/2759#issuecomment-2718880253) says that the app behaves unexpectedly due to this issue. Are there any specific symptoms of this issue, like functions not found, that you are aware of? One symptom I noticed is that I was getting no functions found in my local testing but I am not sure if that is the only symptom for all the scenarios.

src/WebJobs.Script/Host/FunctionAppValidationService.cs
kshyju: IMO, This service is the right place to perform both validations about the deployed payload, 
1. Verifying that the ZIP was created from the correct location
2. ensuring it contains a valid .azurefunctions subfolder.

mattchenderson: From a quick check, I think I was wrong here. It looks like that has been present since the earliest versions, so this should be safe.

src/WebJobs.Script/Host/FunctionAppValidationService.cs
mattchenderson: That statement is grounded in the scenario that was being described. The user learns of the issue because none of the things the app is supposed to do are happening. And then it's incredibly difficult to diagnose since there is a lack of information from the system, and they don't see the payload that was actually deployed or any errors indicating that it doesn't mirror their original build.

We would expect a zero functions found error raised, but the _why_ of that is the problem here.

surgupta-msft: > Could we make this run conditionally? For example, only trigger this validation if loading the app/functions fails.

@RohitRanjanMS I tried this approach and one issue I noticed is that we call `FunctionMetadataManager.GetFunctionMetadata()` from a few places in code. If we trigger the validation service when functions are not found then we will end up validating this everytime `GetFunctionMetadata()` is called and logging it a few times in kusto and AppInsights. So, this is leading to extra log lines and unnecessary IO operations.

src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
```diff
@@ -387,5 +397,17 @@ public static void NoHostJsonFile(this ILogger logger)
         {
             _noHostJsonFile(logger, null);
         }
+
+        public static void MissingAzureFunctionsFolder(this ILogger logger)
+        {
+            string functionWorkerRuntime = EnvironmentSettingNames.FunctionWorkerRuntime;
```
kshyju: nit: This is the env variable name. the name suggests it's the actual FWR value (because rest of the codebase uses variables like this to store the FWR value). Could improve the name or just inline it.

src/WebJobs.Script/Host/FunctionAppValidationService.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        protected override async Task ExecuteAsync(CancellationToken cancellationToken)
+        {
+            // Adding a delay to ensure that this validation does not impact the cold start performance
+            Utility.ExecuteAfterColdStartDelay(_environment, Validate, cancellationToken);
+
+            await Task.CompletedTask;
+        }
+
+        internal void Validate()
```
kshyju: Always consider using the least necessary visibility to reduce the surface area for misuse and improve maintainability. This method appears to be a candidate for `private`

src/WebJobs.Script/Host/FunctionAppValidationService.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            try
+            {
+                if (_scriptOptions.Value.RootScriptPath is not null &&
+                    !_scriptOptions.Value.IsDefaultHostConfig &&
+                    !_scriptOptions.Value.IsStandbyConfiguration &&
```
kshyju: Any thoughts on this comment - https://github.com/Azure/azure-functions-host/pull/10967/files#r2076036678 ?

src/WebJobs.Script/Host/FunctionAppValidationService.cs
```diff
@@ -0,0 +1,77 @@
...
+                    Utility.IsDotnetIsolatedApp(environment: _environment) &&
+                    !Directory.Exists(Path.Combine(_scriptOptions.Value.RootScriptPath, ScriptConstants.AzureFunctionsSystemDirectoryName)))
+                {
+                    string azureFunctionsDirPath = Path.Combine(_scriptOptions.Value.RootScriptPath, ScriptConstants.AzureFunctionsSystemDirectoryName);
+
+                    // Search for the .azurefunctions directory within nested directories to verify scenarios where it isn't located at the root. This situation occurs when a function app has been improperly zipped.
+                    IEnumerable<string> azureFunctionsDirectories = Directory.GetDirectories(_scriptOptions.Value.RootScriptPath, ScriptConstants.AzureFunctionsSystemDirectoryName, SearchOption.AllDirectories)
+                        .Where(dir => !dir.Equals(azureFunctionsDirPath, StringComparison.OrdinalIgnoreCase));
+
+                    if (azureFunctionsDirectories is not null && azureFunctionsDirectories.Any())
```
kshyju: null check is not needed as the LINQ expression returns an enumerable.

test/DotNetIsolatedInvalidApp/Program.cs
```diff
@@ -0,0 +1,13 @@
+using Microsoft.Extensions.Hosting;
+
+Console.WriteLine("Console Out from worker on startup.");
+
+//Debugger.Launch();
```
kshyju: Remove comment.

test/DotNetIsolatedInvalidApp/Program.cs
```diff
@@ -0,0 +1,13 @@
...
+
+var hostBuilder = new HostBuilder();
+
```
kshyju: Extra line break.

test/WebJobs.Script.Tests/Description/FunctionAppValidationServiceTests.cs
```diff
@@ -0,0 +1,151 @@
...
+
+            var service = new FunctionAppValidationService(
+                _testLogger,
+                _scriptOptionsMock.Object,
+                environment);
+
+            // Act
+            await service.StartAsync(CancellationToken.None);
+
+            await Task.Delay(_initializationDelay);
```
kshyju: This may lead to unreliable behavior. There's a helper method used in other tests that waits for specific log messages from the host, we can use that instead.

test/DotNetIsolatedInvalidApp/DotNetIsolatedInvalidApp.csproj
```diff
@@ -0,0 +1,33 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
kshyju: This app itself is not invalid. All we care about is using an invalid zip package. Can't we use an existing test app?

test/DotNetIsolatedInvalidApp/Properties/serviceDependencies.json
```diff
@@ -0,0 +1,8 @@
+{
```
kshyju: These files under "Properties" should not be checked in.

test/DotNetIsolatedInvalidApp/DotNetIsolatedInvalidApp.csproj
surgupta-msft: Found a way to test this without needing a new app. Deleted it.

test/WebJobs.Script.Tests/Description/FunctionAppValidationServiceTests.cs
surgupta-msft: Makes sense. Updated.

src/WebJobs.Script/Host/FunctionAppValidationService.cs
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        private void Validate()
+        {
+            try
+            {
+                if (_scriptOptions.Value.RootScriptPath is not null &&
+                    !_scriptOptions.Value.IsDefaultHostConfig &&
+                    Utility.IsDotnetIsolatedApp(environment: _environment) &&
+                    !Directory.Exists(Path.Combine(_scriptOptions.Value.RootScriptPath, ScriptConstants.AzureFunctionsSystemDirectoryName)))
```
kshyju: You can store the result of `Path.Combine(_scriptOptions.Value.RootScriptPath, ScriptConstants.AzureFunctionsSystemDirectoryName)` to a variable and use that again in line 56.

src/WebJobs.Script/Host/FunctionAppValidationService.cs
```diff
@@ -0,0 +1,79 @@
...
+                    }
+                    else
+                    {
+                        _logger.MissingAzureFunctionsFolder();
+                    }
+                }
+            }
+            catch
+            {
+                // Ignore any exceptions.
```
kshyju: Is there a specific reason for silently ignoring exceptions here? If so, we should add a comment about that here. Would it be better to log the errors, so we’re aware of unexpected behavior (if any) ?.

kshyju (approved):

src/WebJobs.Script/Host/FunctionAppValidationService.cs
```diff
@@ -0,0 +1,79 @@
...
+                    }
+                    else
+                    {
+                        _logger.MissingAzureFunctionsFolder();
+                    }
+                }
+            }
+            catch
+            {
+                // Ignore any exceptions such as if the path in RootScriptPath is invalid or unavailable, the recursive directory search may throw an exception
```
liliankasem: Can we at least put a trace log here? It would be good for us to gave something in logs in case this cause other issues down the line

src/WebJobs.Script/Host/FunctionAppValidationService.cs
```diff
@@ -0,0 +1,79 @@
...
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.WebJobs.Script.Host
+{
+    /// <summary>
+    /// A background service responsible for validating function app payload.
+    /// </summary>
+    public sealed class FunctionAppValidationService : BackgroundService
```
liliankasem: Does this need to be public? Can it be internal?

liliankasem (approved):
Two minor comments other looks good

liliankasem (approved):

PR #10979, opened 2025-04-01
Fail executions when worker restarts due to function timeout

### LINKED ISSUE #10936: Worker restart due to function timeout does not fail current executions
When an execution times out and the worker channel restarts as a result ([here](https://github.com/Azure/azure-functions-host/blob/cb16901e58789c7aed5db8e8c913d71b9c8fc2a6/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs#L679)), no exception is passed into the [call to shut down the existing channel ](https://github.com/Azure/azure-functions-host/blob/cb16901e58789c7aed5db8e8c913d71b9c8fc2a6/src/WebJobs.Script/Workers/Rpc/JobHostRpcWorkerChannelManager.cs#L42). This results in `TryFailExecutions()` [exiting early](https://github.com/Azure/azure-functions-host/blob/cb16901e58789c7aed5db8e8c913d71b9c8fc2a6/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs#L1554), and mid-flight executions are not failed properly as the worker channel shuts down.

If a timer trigger function is mid-execution during a worker channel restart like this, this scenario is particularly bad because the timer trigger won't run again until the previous execution finishes (which in this case won't happen). This can be mitigated by restarting the app itself.

satvu:

resolves #10936

### Pull request checklist

    * Otherwise: Link to backporting PR - TODO: will link

### DIFF
--- a/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
@@ -1556,16 +1556,16 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
 
         public bool TryFailExecutions(Exception workerException)
         {
-            if (workerException == null)
-            {
-                return false;
-            }
-
             foreach (var invocation in _executingInvocations?.Values)
             {
                 string invocationId = invocation.Context?.ExecutionContext?.InvocationId.ToString();
                 _workerChannelLogger.LogDebug("Worker '{workerId}' encountered a fatal error. Failing invocation: '{invocationId}'", _workerId, invocationId);
-                invocation.Context?.ResultSource?.TrySetException(workerException);
+
+                if (workerException is not null)
+                {
+                    invocation.Context?.ResultSource?.TrySetException(workerException);
+                }
+
                 RemoveExecutingInvocation(invocationId);
             }
             return true;
--- a/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
@@ -1565,6 +1565,10 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
                 {
                     invocation.Context?.ResultSource?.TrySetException(workerException);
                 }
+                else
+                {
+                    invocation.Context?.ResultSource?.TrySetCanceled();
+                }
 
                 RemoveExecutingInvocation(invocationId);
             }
--- a/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
+++ b/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
@@ -677,7 +677,7 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
             Dispose(true);
         }
 
-        public async Task<bool> RestartWorkerWithInvocationIdAsync(string invocationId)
+        public async Task<bool> RestartWorkerWithInvocationIdAsync(string invocationId, Exception e = default)
         {
             // Dispose and restart errored channel with the particular invocation id
             var channels = await GetInitializedWorkerChannelsAsync();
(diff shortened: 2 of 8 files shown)


test/WebJobs.Script.Tests/Workers/Rpc/RpcFunctionInvocationDispatcherTests.cs
```diff
@@ -549,6 +549,111 @@ public async Task FunctionDispatcher_Restart_ErroredChannels_OnWorkerRestart_Not
...
+            Assert.True(result);
+            mockWebHostChannelManager.Verify(m => m.ShutdownChannelIfExistsAsync(
+                It.IsAny<string>(),
+                It.Is<string>(id => id == "testChannelId"),
+                It.Is<Exception>(ex => ex is TimeoutException && ex.Message == $"Executing invocation `{invocationId}` timed out")),
+                Times.Once);
+        }
+
+        [Fact]
+        public async Task FunctionDispatcher_RestartOfTimedOutChannels_JobHostFailsCurrentExecutions()
```
satvu, line 582: The WebHost will try to shut down the channel first and fail executions, but in the case that returns false we want to make sure the JobHost attempts to. See [here](https://github.com/Azure/azure-functions-host/blob/49e1420b87b9d4a867089c968cddeb5eca288cc0/src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs#L540C15-L544C18).

src/WebJobs.Script/Workers/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -685,7 +686,7 @@ public async Task<bool> RestartWorkerWithInvocationIdAsync(string invocationId)
                 if (channel.IsExecutingInvocation(invocationId))
                 {
                     _logger.LogDebug($"Restarting channel with workerId: '{channel.Id}' that is executing invocation: '{invocationId}' and timed out.");
-                    await DisposeAndRestartWorkerChannel(_workerRuntime, channel.Id);
+                    await DisposeAndRestartWorkerChannel(_workerRuntime, channel.Id, new TimeoutException($"Executing invocation `{invocationId}` timed out"));
```
brettsam: This won't give us the right stack -- I'd say just modify this interface to take an Exception and pass it in from the caller (it looks like we'll have it). I know this is a public interface, but it's only used by us so it shouldn't matter too much to modify.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -1556,16 +1556,20 @@ public bool IsExecutingInvocation(string invocationId)
...
+                    invocation.Context?.ResultSource?.TrySetException(workerException);
+                }
+                else
+                {
+                    invocation.Context?.ResultSource?.TrySetCanceled();
+                }
+
                 RemoveExecutingInvocation(invocationId);
             }
             return true;
```
liliankasem, line 1575: When will this function ever return false? Is there is no scenario for it, why is it a bool?

satvu: Looked into this - the bool returned also isn't used anywhere (in all references, this method is called with no action taken based on the result). Will refactor.

jviau:
@satvu I have some questions about what 'failing' executions here entails. Specifically, how will this failure propagate to the individual WebJobs extensions which triggered these other invocations? I want to ensure we don't start causing other invocations to be truly 'failed' instead of being retried on next startup (if they support it). Queue based messages are an example - durable, service bus, storage queue. All of these triggers should be retried in the event of a host restart due to failure. If this change now propagates an exception to them, will the individual extensions know now to persist that failure? Will they know it ignore it and let the queue message lock expire and be retried again later?

satvu:
> @satvu I have some questions about what 'failing' executions here entails. Specifically, how will this failure propagate to the individual WebJobs extensions which triggered these other invocations? I want to ensure we don't start causing other invocations to be truly 'failed' instead of being retried on next startup (if they support it). Queue based messages are an example - durable, service bus, storage queue. All of these triggers should be retried in the event of a host restart due to failure. If this change now propagates an exception to them, will the individual extensions know now to persist that failure? Will they know it ignore it and let the queue message lock expire and be retried again later?

@jviau To make sure we're on the same page, the changes are related to _worker channel restarts_ and the host remains online the entire time while one worker channel is shut down and a new one is started (and handles incoming invocations). The pre-existing `TryFailExecutions` method uses the same strategy of failing executions as the `GrpcWorkerChannel` handling of failing invocation responses (see [here](https://github.com/Azure/azure-functions-host/blob/4ba45d7c6906f0e3655b7e22d650ada349433cb5/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs#L1098)) so for the caller it should be similar to an individual function invocation failing (but with the propogated timeout exception included).

I'll test one of your examples e2e for investigation/confirmation, and I'm also still looking to see in what cases `TryFailExecution` has been called in the past with an exception to observe the failure behavior and see if this method was every actually used as intended.

satvu:
Closing this - needs more design discussion and investigation.

PR #10980, opened 2025-04-01
Fix invocation timeout when incoming request contains "x-ms-invocation-id" header

### LINKED ISSUE #10934: Invocation can time out if incoming request contains "x-ms-invocation-id" header
If you look at the [`DefaultHttpProxyService`](https://github.com/Azure/azure-functions-host/blob/d1e06f1d9816105eefc4f50a78a1a43e63a936de/src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs#L100), we call `TryAdd()` to the `x-ms-invocation-id` header. This means that if it's already there, we do not add it.

This means that the header id is incorrect and the worker cannot find the corresponding grpc invocation, resulting in a timeout error on the worker side in dotnet-isolated.

This can happen if one dotnet-isolated function app is directly forwarding or cloning a request and sending it to a second dotnet-isolated function app. This carries the header from the original app over to the second app and causes a timeout. We've had one CRI on this.

The error is `Timed out waiting for the function start call.`

We should probably just be overwriting this header whenever we add it as it's really for host-worker communication. The fix should be fairly straightforward, but we'll need a test to protect against regressions.

satvu:

resolves #10934

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
+++ b/src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
@@ -101,7 +101,7 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
             // add invocation id as correlation id
             if (!httpRequest.Headers.TryAdd(ScriptConstants.HttpProxyCorrelationHeader, invocationId))
             {
-                httpRequest.Headers[ScriptConstants.HttpProxyingEnabled] = invocationId;
+                httpRequest.Headers[ScriptConstants.HttpProxyCorrelationHeader] = invocationId;
             }
 
             var forwardingTask = _httpForwarder.SendAsync(httpContext, httpUri.ToString(), _messageInvoker, _forwarderRequestConfig).AsTask();
--- a/src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
+++ b/src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
@@ -2,10 +2,8 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Generic;
 using System.Linq;
 using System.Net.Http;
-using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Azure.WebJobs.Script.Description;
@@ -13,7 +11,6 @@ using Microsoft.Azure.WebJobs.Script.Grpc.Exceptions;
 using Microsoft.Azure.WebJobs.Script.Workers;
 using Microsoft.Extensions.Logging;
 using Yarp.ReverseProxy.Forwarder;
-using static Microsoft.Azure.AppService.Proxy.Common.Debug.DebugLogHelper;
 
 namespace Microsoft.Azure.WebJobs.Script.Grpc
 {
--- a/src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
+++ b/src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Net.Http;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Azure.WebJobs.Script.Description;
@@ -12,6 +13,7 @@ using Microsoft.Azure.WebJobs.Script.Grpc.Exceptions;
 using Microsoft.Azure.WebJobs.Script.Workers;
 using Microsoft.Extensions.Logging;
 using Yarp.ReverseProxy.Forwarder;
+using static Microsoft.Azure.AppService.Proxy.Common.Debug.DebugLogHelper;
 
 namespace Microsoft.Azure.WebJobs.Script.Grpc
 {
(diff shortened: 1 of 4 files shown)


src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
```diff
@@ -97,7 +97,12 @@ public void StartForwarding(ScriptInvocationContext context, Uri httpUri)
             httpContext.Items[ScriptConstants.HttpProxyingEnabled] = bool.TrueString;
 
             // add invocation id as correlation id
-            httpRequest.Headers.TryAdd(ScriptConstants.HttpProxyCorrelationHeader, context.ExecutionContext.InvocationId.ToString());
+            if (!httpRequest.Headers.TryAdd(ScriptConstants.HttpProxyCorrelationHeader, context.ExecutionContext.InvocationId.ToString()))
```
brettsam: Rather than call `ToString()` twice -- call it once and use that later. In general, we'll need to sweep through and clean all of these `ToString()` calls up in the future, but let's not add to them.

src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
```diff
@@ -97,7 +97,12 @@ public void StartForwarding(ScriptInvocationContext context, Uri httpUri)
...
+            {
+                // try to remove the header if it already exists
+                httpRequest.Headers.Remove(ScriptConstants.HttpProxyCorrelationHeader);
+                httpRequest.Headers.TryAdd(ScriptConstants.HttpProxyCorrelationHeader, context.ExecutionContext.InvocationId.ToString());
```
brettsam: Rather than do this, can we do `httpRequest.Headers[ScriptConstants.HttpProxyingEnabled] = ...` since we're going to overwrite it anyway? That'd be less operations.

src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
```diff
@@ -97,7 +97,12 @@ public void StartForwarding(ScriptInvocationContext context, Uri httpUri)
...
+            }
 
             var forwardingTask = _httpForwarder.SendAsync(httpContext, httpUri.ToString(), _messageInvoker, _forwarderRequestConfig).AsTask();
```
brettsam, line 101: The only other thing that I was wondering is if it'd make sense to use an `HttpTransformer` instead here. But I think that since this is so straightforward it doesn't buy us much.

brettsam (commented):
This looks good -- I'd recommend sync'ing with @jviau or @fabiocav -- I recall one of them mentioning that we may need to store the original header? I'm not sure that we do, but they would likely have more details.

src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
```diff
@@ -96,8 +96,13 @@ public void StartForwarding(ScriptInvocationContext context, Uri httpUri)
             HttpContext httpContext = httpRequest.HttpContext;
             httpContext.Items[ScriptConstants.HttpProxyingEnabled] = bool.TrueString;
 
+            string invocationId = context.ExecutionContext.InvocationId.ToString();
+
             // add invocation id as correlation id
-            httpRequest.Headers.TryAdd(ScriptConstants.HttpProxyCorrelationHeader, context.ExecutionContext.InvocationId.ToString());
+            if (!httpRequest.Headers.TryAdd(ScriptConstants.HttpProxyCorrelationHeader, invocationId))
+            {
+                httpRequest.Headers[ScriptConstants.HttpProxyingEnabled] = invocationId;
```
RohitRanjanMS: I could be mistaken, but are you certain you want to set HttpProxyingEnabled instead of HttpProxyCorrelationHeader?

test/WebJobs.Script.Tests/HttpProxyService/DefaultHttpProxyServiceTests.cs
```diff
@@ -0,0 +1,82 @@
...
+                Properties = new Dictionary<string, object>()
+            };
+
+            var httpUri = new Uri("http://localhost");
+            var existingCorrelationId = Guid.NewGuid().ToString();
+            var httpRequest = (HttpRequest)context.Inputs.First().Val;
+            httpRequest.Headers[ScriptConstants.HttpProxyCorrelationHeader] = existingCorrelationId;
+
+            _proxyService.StartForwarding(context, httpUri);
+            Assert.NotEqual<string>(existingCorrelationId, httpRequest.Headers[ScriptConstants.HttpProxyCorrelationHeader]);
```
RohitRanjanMS: It might be preferable to assert that `httpRequest.Headers[ScriptConstants.HttpProxyCorrelationHeader]` equals `InvocationId`, as this implicitly indicates that it no longer matches `existingCorrelationId` but now aligns with `InvocationId`.

src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
satvu: Ah yes, typo.

RohitRanjanMS (approved):

src/WebJobs.Script.Grpc/Server/DefaultHttpProxyService.cs
```diff
@@ -96,8 +96,13 @@ public void StartForwarding(ScriptInvocationContext context, Uri httpUri)
             HttpContext httpContext = httpRequest.HttpContext;
             httpContext.Items[ScriptConstants.HttpProxyingEnabled] = bool.TrueString;
 
+            string invocationId = context.ExecutionContext.InvocationId.ToString();
+
             // add invocation id as correlation id
-            httpRequest.Headers.TryAdd(ScriptConstants.HttpProxyCorrelationHeader, context.ExecutionContext.InvocationId.ToString());
+            if (!httpRequest.Headers.TryAdd(ScriptConstants.HttpProxyCorrelationHeader, invocationId))
```
brettsam: Just remove this whole `if` -- we're gonna overwrite whatever is there anyway, no need to check for it.

jviau:
> This looks good -- I'd recommend sync'ing with @jviau or @fabiocav -- I recall one of them mentioning that we may need to store the original header? I'm not sure that we do, but they would likely have more details.

I am struggling to recall this exactly. But my _guess_ would be something along the lines of putting the existing header into a new header, like `x-ms-orig-invocation-id`, so the function still has access to it, but it doesn't interfere with our system.

fabiocav (approved):

fabiocav:
I'm OK with us scoping the changes to what is currently in place. If we feel there's a need for us to maintain the original header (I don't think we've been able to identify valid scenarios for that), we can introduce that in another PR. That would also lead to questions about hops across multiple Function Apps.

liliankasem (approved):

PR #10982, opened 2025-04-07
Filter Azure monitor logs when customer does not subscribe for the categories

manikantanallagatla:
For a few SKUs in functions, we want to filter azure monitor logs based on the categories subscribed by the user. This PR checks for the env variable for the categories and skips logging.

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/Diagnostics/AzureMonitorDiagnosticLogger.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/AzureMonitorDiagnosticLogger.cs
@@ -54,8 +54,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 
         public bool IsEnabled(LogLevel logLevel)
         {
-            if ((_environment.IsFlexConsumptionSku() || _environment.IsLinuxConsumptionOnLegion())
-                && !_environment.IsAzureMonitorEnabledOnLegionSkus())
+            if (_environment.IsLegionBasedSku() && !_environment.IsAzureMonitorEnabled(useCache:true))
             {
                 return false;
             }
--- a/src/WebJobs.Script.WebHost/Diagnostics/AzureMonitorDiagnosticLogger.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/AzureMonitorDiagnosticLogger.cs
@@ -5,15 +5,17 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using System.Linq;
 using Microsoft.Azure.WebJobs.Logging;
 using Microsoft.Azure.WebJobs.Script.Configuration;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Newtonsoft.Json;
+using static Microsoft.Azure.WebJobs.Script.Utility;
 
 namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 {
-    public class AzureMonitorDiagnosticLogger : ILogger
+    public class AzureMonitorDiagnosticLogger : ILogger, IDisposable
     {
         internal const string AzureMonitorCategoryName = "FunctionAppLogs";
         internal const string AzureMonitorOperationName = "Microsoft.Web/sites/functions/log";
@@ -29,7 +31,9 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
         private readonly IEventGenerator _eventGenerator;
         private readonly IEnvironment _environment;
         private readonly IExternalScopeProvider _scopeProvider;
+        private readonly IDisposable _appServiceOptionsOnChangeListener;
         private AppServiceOptions _appServiceOptions;
+        private static bool isAzureMonitorLogsEnabled;
 
         public AzureMonitorDiagnosticLogger(string category, string hostInstanceId, IEventGenerator eventGenerator, IEnvironment environment, IExternalScopeProvider scopeProvider,
             HostNameProvider hostNameProvider, IOptionsMonitor<AppServiceOptions> appServiceOptionsMonitor)
@@ -42,8 +46,8 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
             _hostNameProvider = hostNameProvider ?? throw new ArgumentNullException(nameof(hostNameProvider));
             _ = appServiceOptionsMonitor ?? throw new ArgumentNullException(nameof(appServiceOptionsMonitor));
 
-            appServiceOptionsMonitor.OnChange(newOptions => _appServiceOptions = newOptions);
-            _appServiceOptions = appServiceOptionsMonitor.CurrentValue;
+            _appServiceOptionsOnChangeListener = appServiceOptionsMonitor.OnChange(UpdateAppServiceOptions);
+            UpdateAppServiceOptions(appServiceOptionsMonitor.CurrentValue);
 
             _roleInstance = _environment.GetInstanceId();
 
@@ -54,7 +58,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 
         public bool IsEnabled(LogLevel logLevel)
         {
-            if (_environment.IsLegionBasedSku() && !_environment.IsAzureMonitorEnabled(useCache:true))
+            if (_environment.IsConsumptionOnLegion() && !isAzureMonitorLogsEnabled)
             {
                 return false;
             }
@@ -138,6 +142,17 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
             }
 
             _eventGenerator.LogAzureMonitorDiagnosticLogEvent(logLevel, _hostNameProvider.Value, AzureMonitorOperationName, AzureMonitorCategoryName, _regionName, sw.ToString());
+        }
+
+        public void Dispose()
+        {
+            _appServiceOptionsOnChangeListener?.Dispose();
+        }
+
+        private void UpdateAppServiceOptions(AppServiceOptions newOptions)
+        {
+            _appServiceOptions = newOptions;
+            isAzureMonitorLogsEnabled = IsAzureMonitorLoggingEnabled(_appServiceOptions.AzureMonitorTraceCategory);
         }
 
         private static void WritePropertyIfNotNull<T>(JsonTextWriter writer, string propertyName, T propertyValue)
--- a/src/WebJobs.Script.WebHost/Diagnostics/AzureMonitorDiagnosticLogger.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/AzureMonitorDiagnosticLogger.cs
@@ -5,17 +5,15 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
-using System.Linq;
 using Microsoft.Azure.WebJobs.Logging;
 using Microsoft.Azure.WebJobs.Script.Configuration;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Newtonsoft.Json;
-using static Microsoft.Azure.WebJobs.Script.Utility;
 
 namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 {
-    public class AzureMonitorDiagnosticLogger : ILogger, IDisposable
+    public class AzureMonitorDiagnosticLogger : ILogger
     {
         internal const string AzureMonitorCategoryName = "FunctionAppLogs";
         internal const string AzureMonitorOperationName = "Microsoft.Web/sites/functions/log";
@@ -31,9 +29,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
         private readonly IEventGenerator _eventGenerator;
         private readonly IEnvironment _environment;
         private readonly IExternalScopeProvider _scopeProvider;
-        private readonly IDisposable _appServiceOptionsOnChangeListener;
-        private AppServiceOptions _appServiceOptions;
-        private static bool isAzureMonitorLogsEnabled;
+        private IOptionsMonitor<AppServiceOptions> _appServiceOptionsMonitor;
 
         public AzureMonitorDiagnosticLogger(string category, string hostInstanceId, IEventGenerator eventGenerator, IEnvironment environment, IExternalScopeProvider scopeProvider,
             HostNameProvider hostNameProvider, IOptionsMonitor<AppServiceOptions> appServiceOptionsMonitor)
@@ -44,10 +40,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
             _environment = environment ?? throw new ArgumentNullException(nameof(environment));
             _scopeProvider = scopeProvider ?? throw new ArgumentNullException(nameof(scopeProvider));
             _hostNameProvider = hostNameProvider ?? throw new ArgumentNullException(nameof(hostNameProvider));
-            _ = appServiceOptionsMonitor ?? throw new ArgumentNullException(nameof(appServiceOptionsMonitor));
-
-            _appServiceOptionsOnChangeListener = appServiceOptionsMonitor.OnChange(UpdateAppServiceOptions);
-            UpdateAppServiceOptions(appServiceOptionsMonitor.CurrentValue);
+            _appServiceOptionsMonitor = appServiceOptionsMonitor ?? throw new ArgumentNullException(nameof(appServiceOptionsMonitor));
 
             _roleInstance = _environment.GetInstanceId();
 
@@ -58,7 +51,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 
         public bool IsEnabled(LogLevel logLevel)
         {
-            if (_environment.IsConsumptionOnLegion() && !isAzureMonitorLogsEnabled)
+            if (_environment.IsConsumptionOnLegion() && !_appServiceOptionsMonitor.CurrentValue.AzureMonitorLoggingEnabled)
             {
                 return false;
             }
@@ -115,7 +108,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
             using (JsonTextWriter writer = new JsonTextWriter(sw) { Formatting = Formatting.None })
             {
                 writer.WriteStartObject();
-                WritePropertyIfNotNull(writer, "appName", _appServiceOptions.AppName);
+                WritePropertyIfNotNull(writer, "appName", _appServiceOptionsMonitor.CurrentValue.AppName);
                 WritePropertyIfNotNull(writer, "roleInstance", _roleInstance);
                 WritePropertyIfNotNull(writer, "message", formattedMessage);
                 WritePropertyIfNotNull(writer, "category", _category);
@@ -142,17 +135,6 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
             }
 
             _eventGenerator.LogAzureMonitorDiagnosticLogEvent(logLevel, _hostNameProvider.Value, AzureMonitorOperationName, AzureMonitorCategoryName, _regionName, sw.ToString());
-        }
-
-        public void Dispose()
-        {
-            _appServiceOptionsOnChangeListener?.Dispose();
-        }
-
-        private void UpdateAppServiceOptions(AppServiceOptions newOptions)
-        {
-            _appServiceOptions = newOptions;
-            isAzureMonitorLogsEnabled = IsAzureMonitorLoggingEnabled(_appServiceOptions.AzureMonitorTraceCategory);
         }
 
         private static void WritePropertyIfNotNull<T>(JsonTextWriter writer, string propertyName, T propertyValue)
--- a/src/WebJobs.Script.WebHost/Diagnostics/AzureMonitorDiagnosticLogger.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/AzureMonitorDiagnosticLogger.cs
@@ -54,6 +54,10 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 
         public bool IsEnabled(LogLevel logLevel)
         {
+            if ((_environment.IsFlexConsumptionSku() || _environment.IsLinuxConsumptionOnLegion()) 
+                && !_environment.IsAzureMonitorEnabledOnLegionSkus()) {
+                return false;
+            }
             // We want to instantiate this Logger in placeholder mode to warm it up, but do not want to log anything.
             return !string.IsNullOrEmpty(_hostNameProvider.Value) && !_environment.IsPlaceholderModeEnabled();
         }
--- a/src/WebJobs.Script.WebHost/Diagnostics/AzureMonitorDiagnosticLogger.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/AzureMonitorDiagnosticLogger.cs
@@ -51,7 +51,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 
         public bool IsEnabled(LogLevel logLevel)
         {
-            if (_environment.IsConsumptionOnLegion() && !_appServiceOptionsMonitor.CurrentValue.AzureMonitorLoggingEnabled)
+            if (_environment.IsConsumptionOnLegion() && !_appServiceOptionsMonitor.CurrentValue.IsAzureMonitorLoggingEnabled)
             {
                 return false;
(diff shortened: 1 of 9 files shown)


jviau (changes requested):
Can you give an example of these categories? Are they regular ILogger categories?

If so, we should use built in filtering with Microsoft.Extensions.Logging. We can either transform configuration, or we can simply call:

Part 1: log level filter
``` CSharp
// condition on if this is the right sku
services.AddOptions<LoggerFilterOptions>()
    .Configure<IConfiguration>((options, config) =>
    {
        string setting = config[EnvironmentSettingNames.AzureMonitorCategories];
        options.AddFilter<AzureMonitorDiagnosticLoggerProvider>((category, level) =>
        {
            // return false when AzureMonitorCategories does not contain FunctionAppLogs
        });
    });
```

Part 2: IsEnabled update

``` CSharp
// AzureMonitorDiagnosticLogger.cs
// Update the below method to return false when AzureMonitorCategories does not contain FunctionAppLogs
public bool IsEnabled(LogLevel logLevel)
{
    // We want to instantiate this Logger in placeholder mode to warm it up, but do not want to log anything.
    return !string.IsNullOrEmpty(_hostNameProvider.Value) && !_environment.IsPlaceholderModeEnabled();
}
```

~Although I have to ask - who sets `EnvironmentSettingNames.AzureMonitorCategories`? Is it from the platform, or does the user explicitly set it?~

RohitRanjanMS:
I agree with @jviau that providing more context about the scenario would be beneficial. 
A simple list of categories alone might not be enough; the log level will also be necessary. 
The environment variable is applicable only to `LinuxContainerEventGenerator` and does not extend to `AzureMonitor/Diagnostic Settings` integrations for other skus. 
If these are specified by the customer, can the existing logging configuration in host.json be leveraged instead?

manikantanallagatla:
> Can you give an example of these categories? Are they regular ILogger categories?
> 
> If so, we should use built in filtering with Microsoft.Extensions.Logging. We can either transform configuration, or we can simply call:
> 
> Part 1: log level filter
> 
> ```cs
> // condition on if this is the right sku
> services.AddOptions<LoggerFilterOptions>()
>     .Configure<IConfiguration>((options, config) =>
>     {
>         string setting = config[EnvironmentSettingNames.AzureMonitorCategories];
>         options.AddFilter<AzureMonitorDiagnosticLoggerProvider>((category, level) =>
>         {
>             // return false when AzureMonitorCategories does not contain FunctionAppLogs
>         });
>     });
> ```
> 
> Part 2: IsEnabled update
> 
> ```cs
> // AzureMonitorDiagnosticLogger.cs
> // Update the below method to return false when AzureMonitorCategories does not contain FunctionAppLogs
> public bool IsEnabled(LogLevel logLevel)
> {
>     // We want to instantiate this Logger in placeholder mode to warm it up, but do not want to log anything.
>     return !string.IsNullOrEmpty(_hostNameProvider.Value) && !_environment.IsPlaceholderModeEnabled();
> }
> ```
> 
> ~Although I have to ask - who sets `EnvironmentSettingNames.AzureMonitorCategories`? Is it from the platform, or does the user explicitly set it?~

I have removed all the changes. We already have category filter in the host code. I just had to called it in isenabled class

jviau:
@manikantanallagatla I believe you will need to also add the filter callback I suggested. It may seem redundant but they have different purposes:

1. filter: avoid running templating on the log message
2. `IsEnabled`: give Logger users a hook to wrap their log statements in incase they have pre-templating code they want to skip.

I also am wondering about the `IsAzureMonitorEnabled` implementation as is being used in this new code paths. These are hot-paths, we need to make sure they are as performant as possible. `IsAzureMonitorEnabled` will be allocating an array and perform string comparison on each run. Can we optimize that? Cache the results of it somewhere?

Also this change will affect all skus - is the behavior of that env variable consistent across all skus?

src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
```diff
@@ -91,7 +95,27 @@ public static IHostBuilder AddWebScriptHost(this IHostBuilder builder, IServiceP
...
+                                    if (setting == null)
+                                    {
+                                        return true;
+                                    }
+                                    string[] categories = setting.Split(',');
+                                    return categories.Contains(ScriptConstants.AzureMonitorTraceCategory);
+                                });
+                            });
+                        }
+                        else
```
vivekjilla: If we put this in `else`, would the provider still gets added as a singleton.

src/WebJobs.Script.WebHost/Diagnostics/AzureMonitorDiagnosticLogger.cs
```diff
@@ -54,6 +54,10 @@ public AzureMonitorDiagnosticLogger(string category, string hostInstanceId, IEve
 
         public bool IsEnabled(LogLevel logLevel)
         {
+            if (_environment.IsLegionBasedSku() && !_environment.IsAzureMonitorEnabled(useCache:true))
+            {
```
vivekjilla: Did we validate the logs are not being emitted in placeholder mode..

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -66,15 +68,26 @@ public static bool IsEasyAuthEnabled(this IEnvironment environment)
...
+            if (string.Equals(ScriptConstants.DefaultAzureMonitorCategories, value, StringComparison.Ordinal))
+            {
+                // Default value for the env variable is None.
+                // This is set when customer does not subscribe any category.
+                return false;
+            }
+            if (useCache && isAzureMonitorLogsSubscribed != null)
+            {
+                return isAzureMonitorLogsSubscribed.Value;
+            }
```
vivekjilla: Can this impact other SKUs?  should we scope down these changes only to Legion.

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -250,6 +263,16 @@ public static bool IsFlexConsumptionSku(this IEnvironment environment)
             return !environment.WebsiteSkuIsDynamic();
         }
 
+        /// <summary>
+        /// Gets a value indicating whether the application is running in Legion.
+        /// </summary>
+        /// <param name="environment">The environment to verify.</param>
+        /// <returns><see cref="true"/> if running on legion, false otherwise.</returns>
+        public static bool IsLegionBasedSku(this IEnvironment environment)
```
vivekjilla: we already have a method called `IsConsumptionOnLegion` which we can use instead of adding new method

src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
jviau: Yeah I don't think this should be in an `else`. This means it won't be registered for legion skus. Is that intentional?

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -66,15 +68,26 @@ public static bool IsEasyAuthEnabled(this IEnvironment environment)
         /// <summary>
         /// Returns true if any Functions AzureMonitor log categories are enabled.
         /// </summary>
-        public static bool IsAzureMonitorEnabled(this IEnvironment environment)
+        public static bool IsAzureMonitorEnabled(this IEnvironment environment, bool useCache=false)
```
jviau: nit: spacing
```suggestion
        public static bool IsAzureMonitorEnabled(this IEnvironment environment, bool useCache = false)
```

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -66,15 +68,26 @@ public static bool IsEasyAuthEnabled(this IEnvironment environment)
...
         {
             string value = environment.GetEnvironmentVariable(AzureMonitorCategories);
             if (value == null)
             {
                 return true;
             }
+            if (string.Equals(ScriptConstants.DefaultAzureMonitorCategories, value, StringComparison.Ordinal))
```
jviau: I think there is an opportunity to refactor to IOptions pattern here. If we have an options configuration based off `IConfiguration`, which parses this setting into the discrete values, then cache invalidation is handled for us. Anything which triggers an IConfiguration re-evaluation will lead to that options being re-evaluated. So long as you import `IOptionsMonitor<>` and use `.CurrentValue`.

By adding an options for this, we can leave this method as-is and selectively move places that use this method to the options object as we feel it is safe.

manikantanallagatla:
> Can you give an example of these categories? Are they regular ILogger categories?
> 
> If so, we should use built in filtering with Microsoft.Extensions.Logging. We can either transform configuration, or we can simply call:
> 
> Part 1: log level filter
> 
> ```cs
> // condition on if this is the right sku
> services.AddOptions<LoggerFilterOptions>()
>     .Configure<IConfiguration>((options, config) =>
>     {
>         string setting = config[EnvironmentSettingNames.AzureMonitorCategories];
>         options.AddFilter<AzureMonitorDiagnosticLoggerProvider>((category, level) =>
>         {
>             // return false when AzureMonitorCategories does not contain FunctionAppLogs
>         });
>     });
> ```
> 
> Part 2: IsEnabled update
> 
> ```cs
> // AzureMonitorDiagnosticLogger.cs
> // Update the below method to return false when AzureMonitorCategories does not contain FunctionAppLogs
> public bool IsEnabled(LogLevel logLevel)
> {
>     // We want to instantiate this Logger in placeholder mode to warm it up, but do not want to log anything.
>     return !string.IsNullOrEmpty(_hostNameProvider.Value) && !_environment.IsPlaceholderModeEnabled();
> }
> ```
> 
> ~Although I have to ask - who sets `EnvironmentSettingNames.AzureMonitorCategories`? Is it from the platform, or does the user explicitly set it?~

Addressed it

manikantanallagatla:
> @manikantanallagatla I believe you will need to also add the filter callback I suggested. It may seem redundant but they have different purposes:
> 
> 1. filter: avoid running templating on the log message
> 2. `IsEnabled`: give Logger users a hook to wrap their log statements in incase they have pre-templating code they want to skip.
> 
> I also am wondering about the `IsAzureMonitorEnabled` implementation as is being used in this new code paths. These are hot-paths, we need to make sure they are as performant as possible. `IsAzureMonitorEnabled` will be allocating an array and perform string comparison on each run. Can we optimize that? Cache the results of it somewhere?
> 
> Also this change will affect all skus - is the behavior of that env variable consistent across all skus?

Addressed it. Made changes to happen only for legion based SKUs

src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
manikantanallagatla: Yup. Removed else.

src/WebJobs.Script.WebHost/Diagnostics/AzureMonitorDiagnosticLogger.cs
manikantanallagatla: We are calling IsAzureMonitorEnabled which checks for env variable which is null during placeholder mode So, for placeholder mode, there is no change with and without change

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
manikantanallagatla: Nope. Made changes to only affect legion based SKUs

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
manikantanallagatla: Yup. Using it now

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
manikantanallagatla: Fixed

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
manikantanallagatla: Makes sense. Using options monitor and tested it thouroughly. We dont need functionapp restart for changes in env variable

manikantanallagatla:
Will fix/ update tests once changes look good to others

src/WebJobs.Script/Config/AppServiceOptions.cs
```diff
@@ -12,5 +12,7 @@ public class AppServiceOptions
         public string RuntimeSiteName { get; set; }
 
         public string SlotName { get; set; }
+
+        public string AzureMonitorTraceCategory { get; set; }
```
jviau: Optional suggestion: instead of having the users of this option class call `IsAzureMonitorLoggingEnabled`, you could calculate and set it as a bool directly on this object.

``` CSharp
// AppServiceOptions.cs
public bool IsAzureMonitorEnabled { get; set; }

// AppServiceOptionsSetup.cs
// Configure(AppServiceOptions options)
options.IsAzureMonitorEnabled = _environment.IsAzureMonitorEnabled();
```

You also won't need the `OnChange` subscription in the consumption of this new property - you can always reference `options.CurrentValue.IsAzureMonitorEnabled`.

src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
```diff
@@ -91,6 +94,18 @@ public static IHostBuilder AddWebScriptHost(this IHostBuilder builder, IServiceP
                     loggingBuilder.AddWebJobsSystem<SystemLoggerProvider>();
                     if (environment.IsAzureMonitorEnabled())
                     {
+                        if (environment.IsConsumptionOnLegion())
+                        {
+                            loggingBuilder.Services.AddOptions<LoggerFilterOptions>()
+                            .Configure<IConfiguration>((options, config) =>
```
jviau: nit: indentation. Lines 100-107 should be indented 1 level further. When splitting multi-line calls, the lines after the first should be indented 1 level further.
```suggestion
                                .Configure<IConfiguration>((options, config) =>
```

jviau: Now that you have options, if you take my suggestion to make `AppServiceOptions` have the bool directly, you could refactor this to:

``` CSharp
.Configure<IOptionsMonitor<AppServiceOptions>>((filters, options) =>
{
    filters.AddFilter<AzureMonitorDiagnosticLoggerProvider>((category, level) =>
    {
        return options.CurrentValue.IsAzureMonitorEnabled;
    });
});
```

jviau (commented):
Couple suggestions to embrace the options pattern a bit more.

src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
manikantanallagatla: Updated. thanks

manikantanallagatla: Done

src/WebJobs.Script/Config/AppServiceOptions.cs
manikantanallagatla: Added and tested it.

manikantanallagatla:
Addressed all comments and updates unit tests as well

src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
```diff
@@ -91,6 +93,18 @@ public static IHostBuilder AddWebScriptHost(this IHostBuilder builder, IServiceP
                     loggingBuilder.AddWebJobsSystem<SystemLoggerProvider>();
                     if (environment.IsAzureMonitorEnabled())
```
jviau, line 94: Well, this is awkward. Going back and forth on this PR and it looks like it is completely unnecessary. We already completely disable AzureMonitor / shoebox by simply not registering the `AzureMonitorDiagnosticLoggerProvider` - and this is applicable to all skus.

Which means, is this PR even necessary anymore?

manikantanallagatla: Yup. I looked at this code and tested the flow. Without the PR, the logs are not getting disabled. I think what was happening is that:

- Legion pods start in placeholder mode and this code is executed at that time. That time, we enable azure monitor.
- When legion pods are specialized, the env variable is set(lets say to None when categories are not subscribed) and we never disable the Azure monitor as we are not doing any options monitor or env variable monitor without the PR.
- Thats why I started with checking for this env variable in is_enabled flow in aAzureMonitorDiagnosticLogger.cs in the PR initially
- Then we went through this options monitor flow which is more efficient.

In summary, specialization is updating the env variable and we need options monitor on that env variable in order to enable/ disable azure monitor logs.

jviau: Are you saying AzureMonitor env variable is set during placeholder mode? I would expect it to be the opposite: no env variable set, logger not registered, specialization brings in the azure monitor variable, but it has no effect.

manikantanallagatla: I mean:

During placeholder mode, WEBSITE_FUNCTIONS_AZUREMONITOR_CATEGORIES env is set to null. it is not set to anything. So, logger is initialized as we are returning true here for null case: 
https://github.com/Azure/azure-functions-host/blob/e04ccef2b01c858d7ad6b293d9aad296607a012a/src/WebJobs.Script/Environment/EnvironmentExtensions.cs#L75

It is also mentioned here that the logger is initialized but the logs are filtered out in placeholder mode: 
https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.WebHost/Diagnostics/AzureMonitorDiagnosticLogger.cs#L58

So, specialization updates the env variable but it has no effect of disabling the logs. 

I think we both are saying samething.

As env variable update is not taking effect, we need this PR so we can disable logging when that env variable is set to None.

src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
```diff
@@ -91,6 +93,18 @@ public static IHostBuilder AddWebScriptHost(this IHostBuilder builder, IServiceP
...
                     {
+                        if (environment.IsConsumptionOnLegion())
+                        {
+                            loggingBuilder.Services.AddOptions<LoggerFilterOptions>()
+                                .Configure<IOptionsMonitor<AppServiceOptions>>((filters, options) =>
+                                {
+                                    bool isAzureMonitorLoggingEnabled = options.CurrentValue.AzureMonitorLoggingEnabled;
+                                    filters.AddFilter<AzureMonitorDiagnosticLoggerProvider>((category, level) =>
+                                    {
+                                        return isAzureMonitorLoggingEnabled;
```
jviau: Small nit: recommend not capturing `isAzureMonitorLoggingEnabled` into a bool. Just to avoid any potential inconsistencies with options and filter refresh.

```suggestion
                                        return options.CurrentValue.AzureMonitorLoggingEnabled;
```

jviau (approved):

src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
manikantanallagatla: Updated

src/WebJobs.Script.WebHost/Diagnostics/AzureMonitorDiagnosticLogger.cs
```diff
@@ -54,6 +51,10 @@ public AzureMonitorDiagnosticLogger(string category, string hostInstanceId, IEve
 
         public bool IsEnabled(LogLevel logLevel)
```
RohitRanjanMS, line 55: Is there any scenario where this could switch from enabled to disabled? I believe that once enabled, it will stay active. Would you like to store the state so that, after activation, all checks can be bypassed?

src/WebJobs.Script/Config/AppServiceOptions.cs
```diff
@@ -12,5 +12,7 @@ public class AppServiceOptions
         public string RuntimeSiteName { get; set; }
 
         public string SlotName { get; set; }
+
+        public bool AzureMonitorLoggingEnabled { get; set; }
```
RohitRanjanMS: IsAzureMonitorLoggingEnabled

src/WebJobs.Script.WebHost/Diagnostics/AzureMonitorDiagnosticLogger.cs
manikantanallagatla: Yes. Customer can unsubscribe the categories which they have subscribed in the past. Then, this would go from enabled to disabled. 

In PR, I was initially caching the isenabled value. But, the options monitor in essence handles it for us. We are not doing too much computation like string split or something. We are checking env variables in this change. So, again caching over the options monitor is redundant

src/WebJobs.Script/Config/AppServiceOptions.cs
manikantanallagatla: Updated

src/WebJobs.Script.WebHost/Diagnostics/AzureMonitorDiagnosticLogger.cs
jviau: @RohitRanjanMS my understanding of the changes is that this will correctly enable/disable dynamically as the config value is changed (due to the usage of `OptionsMonitor`).

With that said, the primary interaction point is an app setting, which when changed will always cause a site restart.

Do you see some issue or concern with how this will enable/disable?

jviau (approved):

src/WebJobs.Script.WebHost/Diagnostics/AzureMonitorDiagnosticLogger.cs
RohitRanjanMS: I’m concerned about the number of environment variable lookups in `IsEnabled`, especially since they occur for every log statement. I’ll run a benchmark on this branch to assess the impact. 

`AzureMonitorDiagnosticLoggerProvider` is added in `WebScriptHostBuilderExtension` and isn’t reinitialized during specialization. This is why AppInsights is configured in `ScriptHostBuilderExtensions`, ensuring it’s reconfigured when specialization happens (or when host.json is modified).

Here’s what I am thinking:
- Register `AzureMonitorDiagnosticLoggerProvider` in ScriptHost if categories are set. The deferred logger should forward all logs from the root logger to `AzureMonitorDiagnosticLogger`. If categories aren’t set, `AzureMonitorDiagnosticLoggerProvider` won’t be registered, and the deferred logger won’t forward logs to it. Deferred logger only forwards Error and above so this option will require a change in deferred logger as well.
- Determine and cache IsAzureMonitorLoggingEnabled when ScriptHost is built, using it in `IsEnabled` instead of relying on environment variable lookups.

RohitRanjanMS: We should also disable `AzureMonitorDiagnosticLoggerProvider` if categories are not set.

RohitRanjanMS: I spoke with @jviau, and here’s our proposed approach:
- Utilize OptionsMonitor to perform the environment variable lookup just once.
- When building ScriptHost, add logic to disable the logger provider if AzureMonitorDiagnosticLogger isn’t needed. Instead of AzureMonitorDiagnosticLogger, return NullLogger.Instance.
`return _isEnabled ? new AzureMonitorDiagnosticLogger(categoryName, _hostInstanceId, _eventGenerator, _environment, _scopeProvider, _hostNameProvider, _appServiceOptions) : NullLogger.Instance;`
- Can this be applied across all SKUs? This would be a solid improvement.

manikantanallagatla: Thanks for checking with Jacob.

Will try the first two things.

Regarding `Can this be applied across all SKUs?` -> We can do it if we have confidence on the change blast radius. As it is logging change which will have direct customer impact in case it has issues. Can we do for legion SKUs for now and add a task to remove legion skus check after deployments and change is working fine? That way we will reduce impact in case of issues.

CC: @vivekjilla

RohitRanjanMS: Sure, we can do this later.

RohitRanjanMS (approved):

src/WebJobs.Script.WebHost/Diagnostics/AzureMonitorDiagnosticLogger.cs
manikantanallagatla: Created task for future optimization: https://dev.azure.com/msazure/Antares/_workitems/edit/33255837

Issue #10995, opened 2025-04-11
Unable to get table reference or create table. Aborting write operation.


Mikkelcv:
Previous function runtime: 4.1037.1.23605
Current function runtime: 4.1038.300.25164

After the function runtime version has been updated to 4.1038 then our function app periodicaly every 10 minutes reports the exception:

_"Unable to get table reference or create table. Aborting write operation."_

Message:
```
This request is not authorized to perform this operation using this permission.
Time:2025-04-11T11:38:55.5555894Z
Status: 403 (Forbidden)
ErrorCode: AuthorizationPermissionMismatch
```

The execption originates in:
`{"assembly":"Microsoft.Azure.WebJobs.Script.WebHost, Version=4.1038.0.0, Culture=neutral, PublicKeyToken=null","method":"Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics.DiagnosticEventTableStorageRepository+<FlushLogs>d__30.MoveNext","level":34,"line":195,"fileName":"/_/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs"}`

Granting the function app: "StorageTableDataContributor" RBAC role on the storage enables the function app to create an empty table in the storage.

Why is it now a requirement that this empty table must be created? 
How do we disable this so we can have the old functionallity back?

It is NOT an option to grant this RBAC role to all function apps, we do not need this and wont grant access that is not required otherwise.

The guide: 
https://learn.microsoft.com/en-us/azure/azure-functions/storage-considerations?tabs=azure-cli

Does not explain that the access is required. Is it something that is on its way?

asos-chanhkiethuynh:
We have been experiencing the same issue as well out of no where, since 26/04/2025.

scrocquesel-ml150:
Neither is https://learn.microsoft.com/en-us/azure/azure-functions/functions-reference?tabs=blob&pivots=programming-language-csharp#connecting-to-host-storage-with-an-identity updated.
This is related to modification done for https://github.com/Azure/azure-functions-host/issues/9569

scrocquesel-ml150:
The role "Storage Table Data Contributor" on the storage account is enough.

rob-thijssen:
I totally agree on all statements made by @Mikkelcv. Especially on this one:
```It is NOT an option to grant this RBAC role to all function apps, we do not need this and wont grant access that is not required otherwise.```

[...]

PR #10996, opened 2025-04-11
Logging improvements to DiagnosticEventTableStorageRepository 

### LINKED ISSUE #10757: Improve error handling when DiagnosticEventsTableRepository cannot flush logs to table storage
When the Function App cannot access the table service (due to missing permissions, firewall/private endpoint restrictions, invalid SAS, etc.), attempts to flush logs in `DiagnosticEventsTableRepository` fail repeatedly. This generates excessive log entries and does not resolve itself without manual intervention.

We need to enhance error handling so that if table access remains unavailable, we can disable the service and stop flushing logs as we have done in https://github.com/Azure/azure-functions-host/issues/10539, reducing log noise and preventing endless retries.

cjaliaga:

This PR introduces improvements to the logging and error handling mechanisms within the `DiagnosticEventTableStorageRepository` class. The main changes include the addition of a dedicated `Logger` class for structured logging and enhancements to error handling to disable the service when it's not able to connect to the table storage service.

Resolves #10757 #10995 

### Pull request checklist

    * Otherwise: #11041

### DIFF
--- a/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.Log.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.Log.cs
@@ -1,0 +1,74 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
+{
+    public partial class DiagnosticEventTableStorageRepository
+    {
+        private static class Logger
+        {
+            private static readonly Action<ILogger, Exception> _serviceDisabledFailedToCreateClient =
+                LoggerMessage.Define(LogLevel.Warning, new EventId(1, nameof(ServiceDisabledFailedToCreateClient)), "An error occurred initializing the Table Storage Client. We are unable to record diagnostic events, so the diagnostic logging service is being stopped.");
+
+            private static readonly Action<ILogger, Exception> _serviceDisabledUnauthorizedClient =
+                LoggerMessage.Define(LogLevel.Warning, new EventId(2, nameof(ServiceDisabledUnauthorizedClient)), "The Table Storage Client is not authorized to access the table storage account. We are unable to record diagnostic events, so the diagnostic logging service is being stopped.");
+
+            private static readonly Action<ILogger, string, Exception> _purgingDiagnosticEvents =
+                LoggerMessage.Define<string>(LogLevel.Debug, new EventId(3, nameof(PurgingDiagnosticEvents)), "Purging diagnostic events with versions older than '{currentEventVersion}'.");
+
+            private static readonly Action<ILogger, string, Exception> _deletingTableWithoutEventVersion =
+                LoggerMessage.Define<string>(LogLevel.Debug, new EventId(4, nameof(DeletingTableWithoutEventVersion)), "Deleting table '{tableName}' as it contains records without an EventVersion.");
+
+            private static readonly Action<ILogger, string, Exception> _deletingTableWithOutdatedEventVersion =
+                LoggerMessage.Define<string>(LogLevel.Debug, new EventId(5, nameof(DeletingTableWithOutdatedEventVersion)), "Deleting table '{tableName}' as it contains records with an outdated EventVersion.");
+
+            private static readonly Action<ILogger, Exception> _errorPurgingDiagnosticEventVersions =
+                LoggerMessage.Define(LogLevel.Error, new EventId(6, nameof(ErrorPurgingDiagnosticEventVersions)), "Error occurred when attempting to purge previous diagnostic event versions.");
+
+            private static readonly Action<ILogger, Exception> _unableToGetTableReference =
+                LoggerMessage.Define(LogLevel.Error, new EventId(7, nameof(UnableToGetTableReference)), "Unable to get table reference. Aborting write operation.");
+
+            private static readonly Action<ILogger, Exception> _unableToGetTableReferenceOrCreateTable =
+                LoggerMessage.Define(LogLevel.Error, new EventId(8, nameof(UnableToGetTableReferenceOrCreateTable)), "Unable to get table reference or create table. Aborting write operation.");
+
+            private static readonly Action<ILogger, Exception> _unableToWriteDiagnosticEvents =
+                LoggerMessage.Define(LogLevel.Error, new EventId(9, nameof(UnableToWriteDiagnosticEvents)), "Unable to write diagnostic events to table storage.");
+
+            private static readonly Action<ILogger, Exception> _primaryHostStateProviderNotAvailable =
+                LoggerMessage.Define(LogLevel.Debug, new EventId(10, nameof(PrimaryHostStateProviderNotAvailable)), "PrimaryHostStateProvider is not available. Skipping the check for primary host.");
+
+            private static readonly Action<ILogger, Exception> _stoppingFlushLogsTimer =
+                LoggerMessage.Define(LogLevel.Information, new EventId(11, nameof(StoppingFlushLogsTimer)), "Stopping the flush logs timer.");
+
+            private static readonly Action<ILogger, Exception> _queueingBackgroundTablePurge =
+                LoggerMessage.Define(LogLevel.Debug, new EventId(12, nameof(QueueingBackgroundTablePurge)), "Queueing background table purge.");
+
+            public static void ServiceDisabledFailedToCreateClient(ILogger logger) => _serviceDisabledFailedToCreateClient(logger, null);
+
+            public static void ServiceDisabledUnauthorizedClient(ILogger logger, Exception exception) => _serviceDisabledUnauthorizedClient(logger, exception);
+
+            public static void PurgingDiagnosticEvents(ILogger logger, string currentEventVersion) => _purgingDiagnosticEvents(logger, currentEventVersion, null);
+
+            public static void DeletingTableWithoutEventVersion(ILogger logger, string tableName) => _deletingTableWithoutEventVersion(logger, tableName, null);
+
+            public static void DeletingTableWithOutdatedEventVersion(ILogger logger, string tableName) => _deletingTableWithOutdatedEventVersion(logger, tableName, null);
+
+            public static void ErrorPurgingDiagnosticEventVersions(ILogger<DiagnosticEventTableStorageRepository> logger, Exception exception) => _errorPurgingDiagnosticEventVersions(logger, exception);
+
+            public static void UnableToGetTableReference(ILogger logger) => _unableToGetTableReference(logger, null);
+
+            public static void UnableToGetTableReferenceOrCreateTable(ILogger logger, Exception exception) => _unableToGetTableReferenceOrCreateTable(logger, exception);
+
+            public static void UnableToWriteDiagnosticEvents(ILogger logger, Exception exception) => _unableToWriteDiagnosticEvents(logger, exception);
+
+            public static void PrimaryHostStateProviderNotAvailable(ILogger logger) => _primaryHostStateProviderNotAvailable(logger, null);
+
+            public static void StoppingFlushLogsTimer(ILogger logger) => _stoppingFlushLogsTimer(logger, null);
+
+            public static void QueueingBackgroundTablePurge(ILogger logger) => _queueingBackgroundTablePurge(logger, null);
+        }
+    }
+}
--- a/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.Log.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.Log.cs
@@ -11,44 +11,58 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
         private static class Logger
         {
             private static readonly Action<ILogger, Exception> _serviceDisabledFailedToCreateClient =
-                LoggerMessage.Define(LogLevel.Warning, new EventId(1, nameof(ServiceDisabledFailedToCreateClient)), "An error occurred initializing the Table Storage Client. We are unable to record diagnostic events, so the diagnostic logging service is being stopped.");
+                LoggerMessage.Define(
+                    LogLevel.Warning,
+                    new EventId(1, nameof(ServiceDisabledFailedToCreateClient)),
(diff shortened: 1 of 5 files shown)


cjaliaga:
TODO: 
~~1) Run the private site extension to test it E2E~~
~~2) Add additional test coverage to test for permissions. Probably I'll need to use the [Azurite Fixture](https://github.com/Azure/azure-functions-host/blob/dev/test/WebJobs.Script.Tests.Integration/Fixtures/AzuriteFixture.cs) to see if I can get a valid connection string with an invalid key to force a 403.~~

test/WebJobs.Script.Tests.Integration/Diagnostics/DiagnosticEventTableStorageRepositoryTests.cs
```diff
@@ -164,7 +164,8 @@ public void GetDiagnosticEventsTable_LogsError_StorageConnectionStringIsNotPrese
             var cloudTable = repository.GetDiagnosticEventsTable(dateTime);
             Assert.Null(cloudTable);
             var messages = _loggerProvider.GetAllLogMessages();
-            Assert.Equal(messages[0].FormattedMessage, "An error occurred initializing the Table Storage Client. We are unable to record diagnostic events, so the diagnostic logging service is being stopped.");
+            var errorIntializingPresent = messages.Any(m => m.FormattedMessage.Contains( "An error occurred initializing the Table Storage Client. We are unable to record diagnostic events, so the diagnostic logging service is being stopped."));
```
RohitRanjanMS: We need a more actionable message. What is the impact of the diagnostic logging service not running, and what steps can customers take to resolve this? I anticipate that customers might contact us upon encountering this error, believing there’s an issue with the Function App.

This applies to other messages as well.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -59,11 +60,33 @@ internal TableServiceClient TableClient
         {
             get
             {
-                if (!_environment.IsPlaceholderModeEnabled() && _tableClient == null && !_azureTableStorageProvider.TryCreateHostingTableServiceClient(out _tableClient))
+                if (!_environment.IsPlaceholderModeEnabled() && _tableClient == null)
```
RohitRanjanMS: Once the app is specialized, this will always be false. Can we cache this value and use it instead of calling `_environment.IsPlaceholderModeEnabled()` every time.

test/WebJobs.Script.Tests.Integration/Diagnostics/DiagnosticEventTableStorageRepositoryTests.cs
RohitRanjanMS: Can you check if these logs appear in AppInsights?

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -59,11 +60,33 @@ internal TableServiceClient TableClient
...
+                        return _tableClient;
+                    }
+
+                    try
+                    {
+                        // When using RBAC, we need "Storage Table Data Contributor" as we require to list, create and delete tables and query/insert/delete entities.
+                        // Testing permissions by listing tables and creating a test table.
+                        var testTable = _tableClient.GetTableClient($"{TableNamePrefix}Check");
+                        _ = TableStorageHelpers.TableExist(testTable, _tableClient);
+                        _ = testTable.CreateIfNotExists();
```
RohitRanjanMS, line 80: Do you want to delete this temp table?

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
RohitRanjanMS: if _tableClient is not null, we can short circuit and just return the client.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -59,11 +60,33 @@ internal TableServiceClient TableClient
...
+                        Logger.ServiceDisabledFailedToCreateClient(_logger);
+                        return _tableClient;
+                    }
+
+                    try
+                    {
+                        // When using RBAC, we need "Storage Table Data Contributor" as we require to list, create and delete tables and query/insert/delete entities.
+                        // Testing permissions by listing tables and creating a test table.
+                        var testTable = _tableClient.GetTableClient($"{TableNamePrefix}Check");
+                        _ = TableStorageHelpers.TableExist(testTable, _tableClient);
```
RohitRanjanMS: How's the behavior if the role is removed after the _tableClient is initialized?

src/WebJobs.Script.WebHost/Helpers/TableStorageHelpers.cs
```diff
@@ -120,5 +122,17 @@ internal static async Task<bool> TableExistAsync(TableClient table, TableService
...
         }
+
+        internal static bool TableExist(TableClient table, TableServiceClient tableClient)
+        {
+            var query = tableClient.Query(p => p.Name == table.Name);
+
+            foreach (var item in query)
+            {
+                return true;
+            }
```
RohitRanjanMS, line 133: Do we need the foreach loop? The intent is just to check if the exception is thrown or not.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
RohitRanjanMS, line 80: `testTable.CreateIfNotExists();` inherently requires both read and write permissions. Do we need to call `TableExist(...)`?

RohitRanjanMS (changes requested):

RohitRanjanMS:
> https://github.com/Azure/azure-functions-host/issues/10995#:~:text=It%20is%20NOT%20an%20option%20to%20grant%20this%20RBAC%20role%20to%20all%20function%20apps%2C%20we%20do%20not%20need%20this%20and%20wont%20grant%20access%20that%20is%20not%20required%20otherwise.

Do we have any publicly available documentation for this feature? If not, it would be helpful to create documentation outlining its behavior and the requirements needed to enable it.

src/WebJobs.Script.WebHost/Helpers/TableStorageHelpers.cs
cjaliaga: Unfortunately the query only materializes (requests are performed) when you go through it.

test/WebJobs.Script.Tests.Integration/Diagnostics/DiagnosticEventTableStorageRepositoryTests.cs
cjaliaga: Good catch! I don't see these warnings in App Insights. I don't actually see any log for the `Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics.DiagnosticEventTableStorageRepository` category other than errors.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
cjaliaga: Swapped conditions to check first for initialized client so we don't enter the if statement if it's already `not null`.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
cjaliaga: Would start failing to write events every ten minutes. I'm now disabling the service as well if we fail to write events due to permissions.

test/WebJobs.Script.Tests.Integration/Diagnostics/DiagnosticEventTableStorageRepositoryTests.cs
RohitRanjanMS: Only error and above are sent to AppInsights.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
cjaliaga: `CreateIfNotExists` internally tries to create the table and checks if the response is a 409 (conflict) and the error code is distinct than `TableBeingDeleted`. That way it knows the table already exists. You could create tables without having read permissions on the table service (custom role for example). Since we need to query the list of tables (read permissions) we test that with `TableStorageHelpers.TableExist`.

src/WebJobs.Script.WebHost/Helpers/TableStorageHelpers.cs
kshyju: _nit_: If you prefer a more concise version `return tableClient.Query(p => p.Name == table.Name).Any();`

src/WebJobs.Script.WebHost/Helpers/TableStorageHelpers.cs
```diff
@@ -120,5 +122,17 @@ internal static async Task<bool> TableExistAsync(TableClient table, TableService
 
             return false;
         }
+
+        internal static bool TableExist(TableClient table, TableServiceClient tableClient)
```
kshyju: Should it be `TableExists`? (like the File.Exist**s**)

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -59,11 +62,41 @@ internal TableServiceClient TableClient
...
+                    if (!_azureTableStorageProvider.TryCreateHostingTableServiceClient(out _tableClient))
+                    {
+                        DisableService();
+                        Logger.ServiceDisabledFailedToCreateClient(_logger);
+                        return _tableClient;
+                    }
+
+                    try
+                    {
+                        // When using RBAC, we need "Storage Table Data Contributor" as we require to list, create and delete tables and query/insert/delete entities.
```
kshyju, line 76: Is one of the log entries we log when this fails include this information (""Storage Table Data Contributor"" role is needed)?

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -59,11 +62,41 @@ internal TableServiceClient TableClient
         {
             get
```
kshyju, line 63: I am wondering we should be doing all these logic inside the property get access code path. The property has 30+ callers today (and could increase in the future). Should we do this check only once, may be during the startup or after specialization is done? Can't we expect user to restart their app once they fix the permission problem?

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
kshyju: Will these logs go to AI where customer can see it?

cjaliaga: Not right now as we're not currently emitting warnings to App Insights and I feel warning is the right log level for this scenario. They will start flowing when we enable that. Since Geomaster is not able yet to read the events for identity based connections, the priority right now is to disable the service. 

We're not explicitly mentioning the role in the log message, but we can totally do that:

https://github.com/Azure/azure-functions-host/blob/34ed87ea3786aa97c1440bcd92fbc0bca300a529/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.Log.cs#L23

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.Log.cs
```diff
@@ -0,0 +1,88 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
+{
+    public partial class DiagnosticEventTableStorageRepository
+    {
+        private static class Logger
```
jviau, line 11: We should have access to logger source gen if you want. 

https://learn.microsoft.com/en-us/dotnet/core/extensions/logger-message-generator

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
cjaliaga: > I am wondering we should be doing all these logic inside the property get access code path. The property has 30+ callers today (and could increase in the future). Should we do this check only once, may be during the startup or after specialization is done?

As soon as the `backing field` is `not null`, further access to the property won't check for permissions and will simply return the `backing field`. 

That said, I totally agree that refactoring the service to react to specialization, check the requirements once, and then either disable the service or enable the timer for flushing logs would be a more efficient and robust approach, as it would ensure that all requirements are properly validated while minimizing overhead during property access.

I actually discussed this approach with @fabiocav in the past. It seems like a good improvement to consider. I guess it depends on whether we want to prioritize addressing the current issue or take the opportunity to refactor now. I'm happy to follow whichever path we prefer.

> Can't we expect user to restart their app once they fix the permission problem?

We expect the customer to restart the application, even with the proposed changes to the property, the service would be disabled and only a restart of the app would make it try again.

jviau (approved):
LGTM. One minor concern is if the issues leading to the service failing are resolved, instances need to be restarted to re-enable this service. I can live with that for now, but something for us to be aware of.

cjaliaga:
> LGTM. One minor concern is if the issues leading to the service failing are resolved, instances need to be restarted to re-enable this service. I can live with that for now, but something for us to be aware of.

I have the same concern but I'm ok for now as well. We could have a timer that runs the checks every 30 minutes for example (without logging errors of course) and re-enables the service if the issues are resolved.

jviau:
> > LGTM. One minor concern is if the issues leading to the service failing are resolved, instances need to be restarted to re-enable this service. I can live with that for now, but something for us to be aware of.
> 
> I have the same concern but I'm ok for now as well. We could have a timer that runs the checks every 30 minutes for example (without logging errors of course) and re-enables the service if the issues are resolved.

That seems reasonable for a future improvement. I would want to make sure we don't log this as an error though, maybe a warning.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
cjaliaga: Added the role in the log message.

kshyju (approved):

RohitRanjanMS (approved):

PR #11012, opened 2025-04-17
Reduce allocations in ScriptStartupTypeLocator.GetExtensionsStartupTypesAsync

kshyju:
Optimization in `GetExtensionsStartupTypesAsync` method which gets executed during cold start. 

- Updated `ParseExtensions` sync method to use S.T.J for Json deserialization and made it async.

#### Before

|Name|Total \(Allocations\)|Self \(Allocations\)|Total Size \(Bytes\)|Self Size \(Bytes\)|
|-|-|-|-|-|
GetExtensionsStartupTypesAsync\(\)|11,611|8|1,650,762|352|

#### After

|Name|Total \(Allocations\)|Self \(Allocations\)|Total Size \(Bytes\)|Self Size \(Bytes\)|
|-|-|-|-|-|
GetExtensionsStartupTypesAsync\(\)|10,463|7|841,050|280|

#### Delta

- `ScriptStartupTypeLocator.GetExtensionsStartupTypesAsync` - Total allocations dropped by 1,148 (9.89%)
- `ScriptStartupTypeLocator.GetExtensionsStartupTypesAsync` - Total allocated size dropped by 809,712 bytes (49.05%).

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
+++ b/src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
@@ -7,6 +7,7 @@ using System.Collections.Immutable;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using System.Text.Json;
 using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Hosting;
 using Microsoft.Azure.WebJobs.Script.Config;
@@ -19,8 +20,6 @@ using Microsoft.Azure.WebJobs.Script.Models;
 using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
-using Newtonsoft.Json;
-using Newtonsoft.Json.Linq;
 
 namespace Microsoft.Azure.WebJobs.Script.DependencyInjection
 {
@@ -28,7 +27,7 @@ namespace Microsoft.Azure.WebJobs.Script.DependencyInjection
     /// An implementation of an <see cref="IWebJobsStartupTypeLocator"/> that locates startup types
     /// from extension registrations.
     /// </summary>
-    public class ScriptStartupTypeLocator : IWebJobsStartupTypeLocator
+    public sealed class ScriptStartupTypeLocator : IWebJobsStartupTypeLocator
     {
         private const string ApplicationInsightsStartupType = "Microsoft.Azure.WebJobs.Extensions.ApplicationInsights.ApplicationInsightsWebJobsStartup, Microsoft.Azure.WebJobs.Extensions.ApplicationInsights, Version=1.0.0.0, Culture=neutral, PublicKeyToken=9475d07f10cb09df";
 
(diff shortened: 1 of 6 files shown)


src/WebJobs.Script/Models/ExtensionReference.cs
```diff
@@ -1,38 +1,34 @@
...
         /// implementation.
         /// </summary>
-        public string HintPath { get; set; }
+        public string HintPath { get; init; }
 
         /// <summary>
-        /// Gets the binding exposed by the extension
+        /// Gets the binding exposed by the extension.
         /// </summary>
-        public ICollection<string> Bindings { get; } = new Collection<string>();
```
RohitRanjanMS, line 36: I’m assuming binding is optional, and this change is likely to have an effect. If the Binding is not defined in the JSON, it will now be null instead of an empty collection. I expect [this ](https://github.com/Azure/azure-functions-host/blob/7e3e7b78fa90ec18eaefd995a9e7797d326d2d00/src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs#L163)to result in an exception.
I believe all the tests we've rely on an extension file with binding information, and this scenario slipped through unnoticed.

RohitRanjanMS (changes requested):

src/WebJobs.Script/Models/ExtensionReference.cs
kshyju: Good find. thank you! There is no easy way to set this to an empty collecting when deserializing using STJ. So I added a null check in the line you linked above before using the `Bindings` property. That is the only place where this property is being used.

The other option is to do make this property `set`table. I don't like that approach since the class is otherwise immutable and would like to keep that way.

kshyju: The null check is to allow executing the code inside the block. I also added a new test which uses extension JSON content without "Bindings" property in it.

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
```diff
@@ -160,7 +159,7 @@ public async Task<IEnumerable<Type>> GetExtensionsStartupTypesAsync()
                 }
 
                 if (!bundleConfigured
-                    || extensionItem.Bindings.Count == 0
+                    || extensionItem.Bindings is null || extensionItem.Bindings.Count == 0
```
kshyju, line 162: With the new changes, `Bindings` property will be `null` when it is not present in the extension JSON content. (see the new test added in the PR). In that case, previously, the Bindings property was populated, but as an empty collection. With STJ, It will not be populated as empty collection, but as null when the source JSON does not have it.

RohitRanjanMS (approved):

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
```diff
@@ -327,7 +327,7 @@ void CollectError(Type extensionType, Version minimumVersion, ExtensionStartupTy
                 }
             }
 
-            if (errors.Count > 0)
+            if (errors != null && errors.Count > 0)
```
surgupta-msft, line 330: We can create a IsNullOrEmpty utility method for list and arrays. We are checking this a few times in this PR. Also, using `is null` and `is not null`

surgupta-msft (approved):

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
kshyju: We are checking that combination only here, in this PR. The other instance (line 162) is an or condition (not and).

RohitRanjanMS (approved):

PR #11016, opened 2025-04-21
Allow UpdateBuildNumber to be set to false

joelverhagen:

Some Azure DevOps organizations have `build.updatebuildnumber` disabled per policy. This causes a build warning. We self-host Azure Functions Host so it would like to eliminate these build warnings if possible.

If we can set `UpdateBuildNumber` to `false` then we can avoid the build warning.

This is used here:
https://github.com/Azure/azure-functions-host/blob/7e3e7b78fa90ec18eaefd995a9e7797d326d2d00/eng/build/Version.targets#L19-L21

### Pull request checklist

  - I am not sure if this should be done. I THINK so based on the template description, but please let me know.
  - N/A

### DIFF
--- a/src/Directory.Version.props
+++ b/src/Directory.Version.props
@@ -1,6 +1,6 @@
 <Project>
   <PropertyGroup>
     <VersionPrefix>4.1040.100</VersionPrefix>
-    <UpdateBuildNumber>true</UpdateBuildNumber>
+    <UpdateBuildNumber Condition="'$(UpdateBuildNumber)' == ''">true</UpdateBuildNumber>
   </PropertyGroup>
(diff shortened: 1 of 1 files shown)


jviau:
@joelverhagen I believe when supplying a parameter via cmdline (or env variable), it takes precedence over any declared in props. You should already be able to set `-p:UpdateBuildNumber=false` in your build command

joelverhagen:
Ah, nice. The environment variable (`$env:UpdateBuildNumber = "false"` in PowerShell) didn't work but the `-p:UpdateBuildNumber=false` did work. Thanks for your help!

PR #11019, opened 2025-04-22
Host Metrics heartbeat published every 30 seconds

liliankasem:

resolves #issue_for_this_pr

### Pull request checklist

### Additional information

The metric publisher currently triggers every 5 seconds, if there is activity, it will publish metrics. If there is no activity is will skip. With this change, the same thing occurs, however now every 30 seconds we publish regardless of if there is any activity or not. This acts as a sort of heartbeat signal so that the platform can know if we are still able to publish metrics or not.

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -4,8 +4,9 @@
 - My change description (#PR)
 -->
 - Improved memory metrics reporting using CGroup data for Linux consumption (#10968)
-- Memory allocation optimizations in `RpcWorkerConfigFactory.AddProviders` (#10959)
+- Memory allocation optimizations in `RpcWorkerConfigFactory.AddProviders` (#10959)
 - Fixing GrpcWorkerChannel concurrency bug (#10998)
 - Avoid circular dependency when resolving LinuxContainerLegionMetricsPublisher. (#10991)
 - Add 'unix' to the list of runtimes kept when importing PowerShell worker for Linux builds
 - Update PowerShell 7.4 worker to 4.0.4206
+- Update flex metric publisher to publish every 30 seconds (regardless of actvity) (#11019)
--- a/src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
+++ b/src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
@@ -36,6 +36,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Metrics
         private IDisposable _standbyOptionsOnChangeSubscription;
         private TimeSpan _metricPublishInterval;
         private TimeSpan _initialPublishDelay;
+        private DateTime _lastPublishTime = DateTime.MinValue;
 
         public FlexConsumptionMetricsPublisher(IEnvironment environment, IOptionsMonitor<StandbyOptions> standbyOptions, IOptions<FlexConsumptionMetricsPublisherOptions> options,
             ILogger<FlexConsumptionMetricsPublisher> logger, IFileSystem fileSystem, IHostMetricsProvider metricsProvider)
@@ -105,9 +106,12 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Metrics
                     }
                 }
 
-                if (FunctionExecutionCount == 0 && FunctionExecutionTimeMS == 0 && !IsAlwaysReady && !_metricsProvider.HasMetrics())
+                bool hasActivity = FunctionExecutionCount > 0 || FunctionExecutionTimeMS > 0 || IsAlwaysReady || _metricsProvider.HasMetrics();
+                bool shouldForcePublish = (now - _lastPublishTime) >= TimeSpan.FromMilliseconds(_options.KeepAliveIntervalMS);
+
+                if (!hasActivity && !shouldForcePublish)
                 {
-                    // no activity to report
+                    // No activity and not time for keep-alive publish
                     return;
                 }
 
@@ -135,6 +139,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Metrics
                     }
 
                     FunctionExecutionTimeMS = FunctionExecutionCount = 0;
+                    _lastPublishTime = now;
                 }
 
                 await _metricsFileManager.PublishMetricsAsync(metrics);
@@ -321,6 +326,12 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Metrics
             /// Gets or sets the total number of function invocations that have started.
             /// </summary>
             public long StartedInvocationCount { get; set; }
+
+            /// <summary>
+            /// Gets a value indicating whether the publisher is alive and sending metrics.
+            /// Will always be true. If not sent, the platform should assume the host is unable to publish metrics.
+            /// </summary>
+            public bool IsAlive { get; } = true;
         }
     }
 }
--- a/src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
+++ b/src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
@@ -36,7 +36,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Metrics
         private IDisposable _standbyOptionsOnChangeSubscription;
         private TimeSpan _metricPublishInterval;
         private TimeSpan _initialPublishDelay;
-        private DateTime _lastPublishTime = DateTime.MinValue;
+        private DateTime _lastPublishTime = DateTime.UtcNow;
 
         public FlexConsumptionMetricsPublisher(IEnvironment environment, IOptionsMonitor<StandbyOptions> standbyOptions, IOptions<FlexConsumptionMetricsPublisherOptions> options,
             ILogger<FlexConsumptionMetricsPublisher> logger, IFileSystem fileSystem, IHostMetricsProvider metricsProvider)
@@ -106,12 +106,12 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Metrics
                     }
                 }
 
-                bool hasActivity = FunctionExecutionCount > 0 || FunctionExecutionTimeMS > 0 || IsAlwaysReady || _metricsProvider.HasMetrics();
+                bool hasActivity = FunctionExecutionCount > 0 || FunctionExecutionTimeMS > 0 || _metricsProvider.HasMetrics();
                 bool shouldForcePublish = (now - _lastPublishTime) >= TimeSpan.FromMilliseconds(_options.KeepAliveIntervalMS);
 
-                if (!hasActivity && !shouldForcePublish)
+                if (!hasActivity && !shouldForcePublish && !IsAlwaysReady)
                 {
-                    // No activity and not time for keep-alive publish
+                    // No activity and not time for keep-alive publish & not always ready
                     return;
                 }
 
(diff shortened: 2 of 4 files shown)


liliankasem:
@mathewc are there any concerns with a change like this and impact to billing?

kshyju:
>This acts as a sort of heartbeat signal so that the platform can know if we are still able to publish metrics or not.

Would like to better understand the impact of this change on the platform layer and any other components that typically rely on this data. Why does the platform need this "metrics heartbeat" data? Will this "dummy" data be persisted downstream, or are there platform-level changes to ensure it is discarded after being identified as a heartbeat signal?

liliankasem:
> > This acts as a sort of heartbeat signal so that the platform can know if we are still able to publish metrics or not.
> 
> Would like to better understand the impact of this change on the platform layer and any other components that typically rely on this data. Why does the platform need this "metrics heartbeat" data? Will this "dummy" data be persisted downstream, or are there platform-level changes to ensure it is discarded after being identified as a heartbeat signal?

> Would like to better understand the impact of this change on the platform layer and any other components that typically rely on this data

For FPS, which consumes this, I don't expect any impact as they do not currently look for anything like this, they just assume  all activity is 0 anyway. The goal would be to later update FPS to care about not seeing any metrics after a given amount of time. As for MDM, they only publish every 30 seconds anyway so I don't expect impact there. And for billing, I also don't expect impact as the activity numbers is depends on are still 0 (but I would like confirmation from @mathewc on this).

> Why does the platform need this "metrics heartbeat" data? 

Currently, the publisher does not publish any metrics if there is no function activity. So FPS (which consumes these metrics for scale and host health) does not suspect any issues from the host when we do not send any metrics. However, we have found that when the host crashes, we also do not send any metrics, but FPS thinks that everything is normal in the host. This change is of one of two parts: 1) in the host we make sure that we still publish the metrics every 30s (could be longer) and 2) update FPS so that it can use that information to know that if it has not seen anything in over 1-2 minutes, something might be wrong with the host. There are also other changes on this front coming in with the Jacob's work on health checks.

> Will this "dummy" data be persisted downstream, or are there platform-level changes to ensure it is discarded after being identified as a heartbeat signal?

In FPS it will not be persisted; in MDM we aggregate the metrics (if activity is still 0 there should be no change). So I don't anticipate changes needed here.

jviau (approved):

kshyju (approved):
No release_notes updates?

liliankasem:
> No release_notes updates?

Didn't think it would be required for an internal process, but I can add a note

src/WebJobs.Script.WebHost/Metrics/FlexConsumptionMetricsPublisher.cs
```diff
@@ -105,9 +106,12 @@ internal async Task OnPublishMetrics(DateTime now)
                     }
                 }
 
-                if (FunctionExecutionCount == 0 && FunctionExecutionTimeMS == 0 && !IsAlwaysReady && !_metricsProvider.HasMetrics())
+                bool hasActivity = FunctionExecutionCount > 0 || FunctionExecutionTimeMS > 0 || _metricsProvider.HasMetrics();
+                bool shouldForcePublish = (now - _lastPublishTime) >= TimeSpan.FromMilliseconds(_options.KeepAliveIntervalMS);
```
mathewc, line 110: Rather than just force emitting empty metrics, isn't there a new metric that you can add to HostMetricsProvider such that HasMetrics returns true on your required interval and this class behaves as it does now without any forced writing?

E.g. if your goal is to have some heartbeat metric, why not have HostMetricsProvider return one?

liliankasem: That's a great suggestion, I like that idea a lot - will look into it

release_notes.md
```diff
@@ -4,8 +4,9 @@
...
 -->
 - Improved memory metrics reporting using CGroup data for Linux consumption (#10968)
-- Memory allocation optimizations in `RpcWorkerConfigFactory.AddProviders` (#10959)

+- Memory allocation optimizations in `RpcWorkerConfigFactory.AddProviders` (#10959)
 - Fixing GrpcWorkerChannel concurrency bug (#10998)
 - Avoid circular dependency when resolving LinuxContainerLegionMetricsPublisher. (#10991)
 - Add 'unix' to the list of runtimes kept when importing PowerShell worker for Linux builds
 - Update PowerShell 7.4 worker to 4.0.4206
+- Update flex metric publisher to publish every 30 seconds (regardless of actvity) (#11019)
```
mathewc: update the comment to reflect the new approach

src/WebJobs.Script.WebHost/Metrics/HostMetricsProvider.cs
```diff
@@ -19,21 +20,24 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Metrics
 {
     public class HostMetricsProvider : IHostMetricsProvider, IDisposable
     {
+        private const string HeartbeatMetricName = "azure.functions.host_metrics_heartbeat";
```
mathewc: Recommend adding this to the existing MetricEventNames class. Also, looking at the naming convention for the rest of the metrics, I don't see _ being used. Assuming your new metric is a peer to the others?

liliankasem: _ is okay for metric names, we do that for the other host metrics i.e.

https://github.com/Azure/azure-functions-host/blob/2557cf38e1de490f4ea423492ea2ec23fb78f24a/src/WebJobs.Script/Metrics/HostMetrics.cs#L23

I went with `azure.functions.host_metrics_heartbeat` instead of `azure.functions.heartbeat` to avoid taking up a name that might be utilized by other health check work coming up, but perhaps I am over thinking this. @RohitRanjanMS @jviau fyi

src/WebJobs.Script.WebHost/Metrics/HostMetricsProvider.cs
```diff
@@ -78,9 +82,21 @@ public void Start()
...
         }
 
+        private void OnHeartbeatTimer(object state)
+        {
+            if (!_started)
+            {
+                return;
+            }
+
+            AddOrUpdateMetricsCache(HeartbeatMetricName, 1);
```
mathewc: Thought: I keep wondering whether it might be better to publish a value with some meaning, like ProcessUptime that is currently returned by the Host status endpoint [here](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.WebHost/Controllers/HostController.cs#L86)? You'd still get your heartbeat, but the value has additional meaning that might be useful.

While generally in Flex instances come and go frequently meaning uptime might not be that long generally, if we did have a case where a particular instance was long running (e.g. it was draining for a long time, or perhaps became orphaned) this information might be useful.

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
```diff
@@ -227,6 +227,7 @@ public static void AddWebJobsScriptHost(this IServiceCollection services, IConfi
             services.ConfigureOptions<AzureMonitorLoggingOptionsSetup>();
             services.AddHostingConfigOptions(configuration);
             services.ConfigureOptions<ExtensionRequirementOptionsSetup>();
+            services.Configure<HostMetricsProviderOptions>(configuration.GetSection("HostMetricsProvider")); // todo: use options setup?
```
mathewc: Yes it might make sense to use an options setup, similar to FlexConsumptionMetricsPublisherOptionsSetup, to centralize the initialization logic.

What external entity will be configuring this value though? Are you doing this in case we need to configure via hosting config across the fleet? I don't think we need individual apps to configure this separately. For hosting config, we have FunctionsHostingConfigOptions.

src/WebJobs.Script.WebHost/Metrics/HostMetricsProvider.cs
jviau: Yes, `_` is the correct separator to use for multi-word names

src/WebJobs.Script.WebHost/Metrics/HostMetricsProvider.cs
jviau: OTel has a metric defined for that: https://opentelemetry.io/docs/specs/semconv/system/system-metrics/#general-metrics

I'd prefer to adopt that metric officially when we begin incorporating OTel for our own consumption (this work is being discussed with the legion team).

As for the goal of this metric: a constant `1` is great to me - this is a common pattern for a health check or heartbeat metric. It is easier to visualize and program against, as it provides a nice step-function to clearly see the health over time.

Example for using HealthChecks as a metric. It is slightly different, as it has 3 states and not 2.

image

mathewc: Ok, sounds good :) In accordance with the above, perhaps you might want to rename this metric to HostHealthHeartbeat to allow it to take on more than one status value in the future? Then maybe you return 2=Healthy now, and in the future we can extend to handle the other states?

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
jviau: `.Configure` _is_ an options setup. This adds an `IConfigureOptions` under the hood which calls `IConfiguration.Bind(options)`.

The question would then be: any reason we don't want config binding here?

jviau: Small nit: I prefer this approach,

```suggestion
            services.AddOptions<HostMetricsProviderOptions>().BindConfiguration("HostMetricsProvider");
```

There is a slight difference between the two. What you have now is getting "host" configuration only. No "app" configuration. `BindConfiguration` will get both host & app config and also support config reloads.

liliankasem: Definitely not for individual app usage, but I wanted to give us an option to change the interval value across for the fleet in case 30 seconds was too long or not enough. I just don't love the idea of having a hard coded value for something like this. 

If hosting config makes more sense, I can move the interval there and scrap `HostMetricsProviderOptions`

src/WebJobs.Script.WebHost/Metrics/HostMetricsProvider.cs
liliankasem: Edit: scratch that- its not worth the extra effort right given the otel work coming up.

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
kshyju: +1 on the hosting config suggestion.

liliankasem:
Sharing summary from Jacob and I's conversation on this topic:

> After reviewing the current design of metrics in flex we found that adding a meter and choosing the metric name in the functions host was irrelevant. Legion's code ignores all of that and just reads the metric payload we write and has hardcoded metric names it emits for each value on its own side.
> 
> Then there is the discussion of "we needed to emit a non-0 metric". That is actually incorrect. This was stated under the understanding of how Legion's processing of our metric file behaves, where it skips the metric if its value is 0. But that is a legion code limitation - we can change that!
> 
> So to the plan. All we actually need from the host code is to emit that metric file, even if it contains 0s, at a minimum interval. Because that file being emitted (its timestamp) is the heartbeat. And we can then update legion's code to emit a heartbeat metric itself each time it processes a metric file, regardless of the contents of that file.
>
> As for the name - well legion already has its own metric names. So we will just follow that naming for now, which means it will be FunctionsHeartbeat

I've reverted to the last state of the PR where we publish from the publisher every 30 seconds, and no new metrics.

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
liliankasem: @mathewc @kshyju per the latest conversations on this (see: https://github.com/Azure/azure-functions-host/pull/11019?notification_referrer_id=NT_kwDOACGNebMxNTk5MTk2OTEwNjoyMTk4OTA1#issuecomment-2843191639). I feel this is no longer needed as we're using the flex publisher options again instead of introducing a new one. However, if you still feel this needs to be in host config let me know.

jviau (approved):

PR #11021, opened 2025-04-22
new api to get triggers

patelchandni:

New GET admin/host/triggers API needed for Functions Deployment Manager will return sync trigger payload.
It is using ResourceContainsSecrets Attribute.

### Pull request checklist

    * Otherwise: Documentation issue linked to PR
    * Otherwise: I've added my notes to `release_notes.md`
    * Otherwise: Backport tracked by issue/PR #issue_or_pr

### DIFF
--- a/src/WebJobs.Script.WebHost/Controllers/HostController.cs
+++ b/src/WebJobs.Script.WebHost/Controllers/HostController.cs
@@ -432,7 +432,6 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Controllers
 
             var result = await _functionsSyncManager.GetTriggersPayloadAsync();
 
-            // Return a dummy body to make it valid in ARM template action evaluation
             return result.Success
                 ? Ok(result)
                 : StatusCode(StatusCodes.Status500InternalServerError, new { status = result.Error });
--- a/src/WebJobs.Script.WebHost/Controllers/HostController.cs
+++ b/src/WebJobs.Script.WebHost/Controllers/HostController.cs
@@ -434,7 +434,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Controllers
 
             return result.Success
                 ? Ok(result)
-                : StatusCode(StatusCodes.Status500InternalServerError, new { status = result.Error });
+                : StatusCode(StatusCodes.Status403Forbidden, new { status = result.Error });
         }
 
         [HttpPost]
--- a/src/WebJobs.Script.WebHost/Controllers/HostController.cs
+++ b/src/WebJobs.Script.WebHost/Controllers/HostController.cs
@@ -430,7 +430,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Controllers
         {
             _metricsLogger.LogEvent(MetricEventNames.GetTriggersInvoked);
 
-            var result = await _functionsSyncManager.GetTriggersPayloadAsync();
+            var result = await _functionsSyncManager.GetTriggersAsync();
 
             return result.Success
                 ? Ok(result)
--- a/src/WebJobs.Script.WebHost/Controllers/HostController.cs
+++ b/src/WebJobs.Script.WebHost/Controllers/HostController.cs
@@ -432,6 +432,9 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Controllers
 
             var result = await _functionsSyncManager.GetTriggersAsync();
 
+            // GetTriggersAsync() does not swallow exceptions,so any failures will still be a 500.
+            // The only result.Success == false we need to consideris when the environment
+            // does not support sync triggers.
             return result.Success
                 ? Ok(result)
                 : StatusCode(StatusCodes.Status403Forbidden, new { status = result.Error });
--- a/src/WebJobs.Script.WebHost/Controllers/HostController.cs
+++ b/src/WebJobs.Script.WebHost/Controllers/HostController.cs
@@ -421,6 +421,23 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Controllers
                 : StatusCode(StatusCodes.Status500InternalServerError, new { status = result.Error });
         }
 
+        [HttpGet]
+        [Route("admin/host/triggers")]
+        [Authorize(Policy = PolicyNames.AdminAuthLevelOrInternal)]
+        [ResourceContainsSecrets]
+        [RequiresRunningHost]
+        public async Task<IActionResult> GetTriggers()
+        {
+            _metricsLogger.LogEvent(MetricEventNames.GetTriggersInvoked);
+
+            var result = await _functionsSyncManager.GetTriggersPayloadAsync();
+
+            // Return a dummy body to make it valid in ARM template action evaluation
+            return result.Success
+                ? Ok(result)
+                : StatusCode(StatusCodes.Status500InternalServerError, new { status = result.Error });
+        }
+
         [HttpPost]
         [Route("admin/host/restart")]
         [Authorize(Policy = PolicyNames.AdminAuthLevel)]
--- a/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
@@ -111,8 +111,6 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
             {
                 await _syncSemaphore.WaitAsync();
 
-                PrepareSyncTriggers();
-
                 var hashBlobClient = await GetHashBlobAsync();
                 if (isBackgroundSync && hashBlobClient == null && !_environment.IsAnyKubernetesEnvironment())
                 {
@@ -308,6 +306,8 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
 
         private async Task<SyncTriggersPayload> GetSyncTriggersPayload()
         {
+            PrepareSyncTriggers();
+
             var hostOptions = _applicationHostOptions.CurrentValue.ToHostOptions();
             var functionsMetadata = _functionMetadataManager.GetFunctionMetadata().Where(m => !m.IsProxy());
 
@@ -803,10 +803,6 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
 
             try
             {
-                await _syncSemaphore.WaitAsync();
-
-                PrepareSyncTriggers();
-
                 var payload = await GetSyncTriggersPayload();
 
                 result.Content = payload.Content;
@@ -819,10 +815,6 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
                 result.Error = "GetTriggers operation failed.";
                 _logger.LogError(ex, result.Error);
                 return result;
-            }
-            finally
-            {
-                _syncSemaphore.Release();
             }
         }
 
--- a/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
@@ -786,7 +786,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
             }
         }
 
-        public async Task<TriggersPayload> GetTriggersPayloadAsync()
+        public async Task<TriggersPayload> GetTriggersAsync()
         {
             var result = new TriggersPayload
             {
--- a/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
@@ -786,6 +786,56 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
             }
         }
 
+        public async Task<GetTriggersResult> GetTriggersPayloadAsync()
+        {
+            var result = new GetTriggersResult
+            {
+                Success = true
+            };
+
+            if (!IsSyncTriggersEnvironment(_webHostEnvironment, _environment))
+            {
+                result.Success = false;
+                result.Error = "Invalid environment for GetTriggers operation.";
+                _logger.LogWarning(result.Error);
+                return result;
+            }
+
+            try
+            {
+                await _syncSemaphore.WaitAsync();
+
+                PrepareSyncTriggers();
+
+                var hashBlobClient = await GetHashBlobAsync();
+
+                var payload = await GetSyncTriggersPayload();
+                if (payload.Count == 0)
+                {
+                    // We don't do background sync for empty triggers.
+                    // We've seen error cases where a site temporarily gets into a situation
+                    // where it's site content is empty. Doing the empty sync can cause the app
+                    // to go idle when it shouldn't.
+                    _logger.LogDebug("No functions found in GetTriggers operation.");
+                }
+
+                result.Payload = payload;
+                return result;
+            }
+            catch (Exception ex)
+            {
+                // best effort - log error and continue
+                result.Success = false;
+                result.Error = "GetTriggers operation failed.";
+                _logger.LogError(ex, result.Error);
+                return result;
+            }
+            finally
+            {
+                _syncSemaphore.Release();
+            }
+        }
+
         public void Dispose()
         {
             _syncSemaphore.Dispose();
--- a/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
@@ -801,21 +801,9 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
                 return result;
             }
 
-            try
-            {
-                var payload = await GetSyncTriggersPayload();
-
-                result.Content = payload.Content;
-                return result;
-            }
-            catch (Exception ex)
-            {
-                // best effort - log error and continue
-                result.Success = false;
-                result.Error = "GetTriggers operation failed.";
-                _logger.LogError(ex, result.Error);
-                return result;
-            }
+            var payload = await GetSyncTriggersPayload();
+            result.Content = payload.Content;
+            return result;
         }
 
         public void Dispose()
--- a/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
@@ -786,9 +786,9 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
             }
         }
 
-        public async Task<TriggersPayload> GetTriggersAsync()
+        public async Task<TriggersResult> GetTriggersAsync()
         {
-            var result = new TriggersPayload
+            var result = new TriggersResult
             {
                 Success = true
             };
--- a/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
@@ -786,9 +786,9 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
             }
         }
 
-        public async Task<GetTriggersResult> GetTriggersPayloadAsync()
+        public async Task<TriggersPayload> GetTriggersPayloadAsync()
         {
-            var result = new GetTriggersResult
+            var result = new TriggersPayload
             {
                 Success = true
             };
--- a/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
@@ -807,15 +807,9 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
 
                 PrepareSyncTriggers();
 
-                var hashBlobClient = await GetHashBlobAsync();
-
                 var payload = await GetSyncTriggersPayload();
                 if (payload.Count == 0)
                 {
-                    // We don't do background sync for empty triggers.
-                    // We've seen error cases where a site temporarily gets into a situation
-                    // where it's site content is empty. Doing the empty sync can cause the app
-                    // to go idle when it shouldn't.
                     _logger.LogDebug("No functions found in GetTriggers operation.");
                 }
 
--- a/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
@@ -92,9 +92,9 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
             }
         }
 
-        public async Task<SyncTriggersResult> TrySyncTriggersAsync(bool isBackgroundSync = false)
+        public async Task<TriggersOperationResult> TrySyncTriggersAsync(bool isBackgroundSync = false)
         {
-            var result = new SyncTriggersResult
+            var result = new TriggersOperationResult
             {
                 Success = true
             };
@@ -808,12 +808,8 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
                 PrepareSyncTriggers();
 
                 var payload = await GetSyncTriggersPayload();
-                if (payload.Count == 0)
-                {
-                    _logger.LogDebug("No functions found in GetTriggers operation.");
-                }
 
-                result.Payload = payload;
+                result.Content = payload.Content;
                 return result;
             }
             catch (Exception ex)
--- a/src/WebJobs.Script.WebHost/Management/GetTriggersResult.cs
+++ b/src/WebJobs.Script.WebHost/Management/GetTriggersResult.cs
@@ -1,0 +1,25 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using static Microsoft.Azure.WebJobs.Script.WebHost.Management.FunctionsSyncManager;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
+{
+    public class GetTriggersResult
+    {
+        /// <summary>
+        /// Gets or sets a value indicating whether the sync operation was successful.
+        /// </summary>
+        public bool Success { get; set; }
+
+        /// <summary>
+        /// Gets or sets the error string in the case of a failed sync operation.
+        /// </summary>
+        public string Error { get; set; }
+
+        /// <summary>
+        /// Gets or sets the payload of the triggers.
+        /// </summary>
+        public SyncTriggersPayload Payload { get; set; }
+    }
+}
--- a/src/WebJobs.Script.WebHost/Management/GetTriggersResult.cs
+++ b/src/WebJobs.Script.WebHost/Management/GetTriggersResult.cs
@@ -1,25 +1,13 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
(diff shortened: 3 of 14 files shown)


eliaslopezgt (approved):
LGTM 👍

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -421,6 +421,23 @@ public async Task<IActionResult> SyncTriggers()
...
+        [Authorize(Policy = PolicyNames.AdminAuthLevelOrInternal)]
+        [ResourceContainsSecrets]
+        [RequiresRunningHost]
+        public async Task<IActionResult> GetTriggers()
+        {
+            _metricsLogger.LogEvent(MetricEventNames.GetTriggersInvoked);
+
+            var result = await _functionsSyncManager.GetTriggersPayloadAsync();
+
+            // Return a dummy body to make it valid in ARM template action evaluation
```
mathewc: this copied comment is invalid

patelchandni: done, thanks.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -786,6 +786,56 @@ internal HttpRequestMessage BuildSetTriggersRequest()
...
+                return result;
+            }
+
+            try
+            {
+                await _syncSemaphore.WaitAsync();
+
+                PrepareSyncTriggers();
+
+                var hashBlobClient = await GetHashBlobAsync();
```
mathewc: This is unused - why is it here?

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -786,6 +786,56 @@ internal HttpRequestMessage BuildSetTriggersRequest()
...
+
+                var payload = await GetSyncTriggersPayload();
+                if (payload.Count == 0)
+                {
+                    // We don't do background sync for empty triggers.
```
mathewc: We're not doing a background sync here. This whole if block can go away

src/WebJobs.Script.WebHost/Management/GetTriggersResult.cs
```diff
@@ -0,0 +1,25 @@
...
+        /// </summary>
+        public bool Success { get; set; }
+
+        /// <summary>
+        /// Gets or sets the error string in the case of a failed sync operation.
+        /// </summary>
+        public string Error { get; set; }
+
+        /// <summary>
+        /// Gets or sets the payload of the triggers.
```
mathewc: Gets or sets the triggers payload

src/WebJobs.Script.WebHost/Management/GetTriggersResult.cs
```diff
@@ -0,0 +1,25 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using static Microsoft.Azure.WebJobs.Script.WebHost.Management.FunctionsSyncManager;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
+{
+    public class GetTriggersResult
+    {
+        /// <summary>
+        /// Gets or sets a value indicating whether the sync operation was successful.
```
mathewc: Remove the "sync" here and below since this result isn't for sync.

Might just say "Gets or sets the error string in the case of a failed operation." and "Gets or sets a value indicating whether the operation was successful."

mathewc: What might even be better is for you to move the Success/Error members into a new base class TriggersOperationResult, and have both SyncTriggersResult and GetTrigersResult inherit from that

src/WebJobs.Script.WebHost/Management/GetTriggersResult.cs
```diff
@@ -0,0 +1,25 @@
...
+
+        /// <summary>
+        /// Gets or sets the error string in the case of a failed sync operation.
+        /// </summary>
+        public string Error { get; set; }
+
+        /// <summary>
+        /// Gets or sets the payload of the triggers.
+        /// </summary>
+        public SyncTriggersPayload Payload { get; set; }
```
mathewc: I think rather than using this type here, you should just have a "string Content" member. I'm not sure why we ever added the SyncTriggersPayload.Count member - looks like it's only used internally by FunctionsSyncManager. So Count doesn't need to be part of our external contract I think. Note that we only pass Content to FE on syncs anyways.

test/WebJobs.Script.Tests.Integration/Management/FunctionsSyncManagerTests.cs
```diff
@@ -256,6 +256,22 @@ private void ResetMockFileSystem(string hostJsonContent = null, string extension
             FileUtility.Instance = fileSystem;
         }
 
+        [Fact]
+        public async Task GetSyncTriggersPayload_ExpectedContent()
+        {
+            // Act
+            var syncResult = await _functionsSyncManager.GetTriggersPayloadAsync();
```
mathewc: nit: not a "sync" operation so, perhaps just rename to "result"

src/WebJobs.Script.WebHost/Management/GetTriggersResult.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using static Microsoft.Azure.WebJobs.Script.WebHost.Management.FunctionsSyncManager;
```
mathewc: can remove this using if you make the changes suggested below

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
patelchandni: removed it, thanks.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
patelchandni: removed the comments, but can we just print a debug log if count is 0?

src/WebJobs.Script.WebHost/Management/GetTriggersResult.cs
patelchandni: updated comment.

test/WebJobs.Script.Tests.Integration/Management/FunctionsSyncManagerTests.cs
patelchandni: Updated variable names.

src/WebJobs.Script.WebHost/Management/GetTriggersResult.cs
patelchandni: I saw this block of comments for sync trigger method:
```
// We don't do background sync for empty triggers.
// We've seen error cases where a site temporarily gets into a situation
// where it's site content is empty. Doing the empty sync can cause the app
// to go idle when it shouldn't.
```

@mathewc - The question is: If we get 0 count, should we overwrite the empty payload during deployment or not? 
If we should overwrite, then I can remove count.
If we should not, then we will need the count.

src/WebJobs.Script.WebHost/Management/GetTriggersResult.cs
patelchandni: That will make SyncTriggersResult = TriggersOperationResult i.e. SyncTriggersResult will not have any new members:

```
public class SyncTriggersResult : TriggersOperationResult
{
}
```

@mathewc - Will that be okay?

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
mathewc: No need to do that

src/WebJobs.Script.WebHost/Management/GetTriggersResult.cs
mathewc: Yes I was aware that it'd leave the empty SyncTriggersResult class. We could also just rename the existing SyncTriggersResult to TriggersOperationResult and update the 9 places where that's used. Then we can just derive your new GetTriggersResult : TriggersOperationResult. This way we don't have an empty class :)

Main goal in all this is to avoid the code duplication

src/WebJobs.Script.WebHost/Management/GetTriggersResult.cs
mathewc: You don't need the count regardless, since it's just the length of the triggers array which is already part of the Content. I actually don't remember why we added the Count member.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
patelchandni: ok, removed it

src/WebJobs.Script.WebHost/Management/GetTriggersResult.cs
patelchandni: Makes sense. Renamed SyncTriggersResult to TriggersOperationResult.
Updated GetTriggersResult to:

```
public class GetTriggersResult : TriggersOperationResult
{
    /// <summary>
    /// Gets or sets the triggers payload.
    /// </summary>
    public string Content { get; set; }
}

File: src/WebJobs.Script.WebHost/Management/GetTriggersResult.cs (hunk 8 above)
Comment (patelchandni) [reply src/WebJobs.Script.WebHost/Management/GetTriggersResult.cs]: removed it.

File: src/WebJobs.Script.WebHost/Management/GetTriggersResult.cs (hunk 6 above)
Comment (patelchandni) [reply src/WebJobs.Script.WebHost/Management/GetTriggersResult.cs]: Oh yes, the content is not encrypted. Updated GetTriggersResult to:

```
public class GetTriggersResult : TriggersOperationResult
{
    /// <summary>
    /// Gets or sets the triggers payload.
    /// </summary>
    public string Content { get; set; }
}

src/WebJobs.Script.WebHost/Management/GetTriggersResult.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
+{
+    public class GetTriggersResult : TriggersOperationResult
```
kshyju:
```suggestion
    public sealed class GetTriggersResult : TriggersOperationResult
```

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -421,6 +421,22 @@ public async Task<IActionResult> SyncTriggers()
...
+        [RequiresRunningHost]
+        public async Task<IActionResult> GetTriggers()
+        {
+            _metricsLogger.LogEvent(MetricEventNames.GetTriggersInvoked);
+
+            var result = await _functionsSyncManager.GetTriggersPayloadAsync();
+
+            return result.Success
+                ? Ok(result)
+                : StatusCode(StatusCodes.Status500InternalServerError, new { status = result.Error });
```
kshyju: If sync trigger isn't supported (`IsSyncTriggersEnvironment` returns false in `FunctionsSyncManager`), is a 500 response the right choice here? Would it make more sense to return a status code in the 4xx range instead?

src/WebJobs.Script.WebHost/Management/GetTriggersResult.cs
kshyju: Not a fan of this `Get` prefix for this type.

kshyju: May be call it "TriggerInfo" or "TriggersResult" or "TriggersResponse" ?

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -421,6 +421,22 @@ public async Task<IActionResult> SyncTriggers()
                 : StatusCode(StatusCodes.Status500InternalServerError, new { status = result.Error });
         }
 
+        [HttpGet]
+        [Route("admin/host/triggers")]
+        [Authorize(Policy = PolicyNames.AdminAuthLevelOrInternal)]
+        [ResourceContainsSecrets]
```
kshyju, line 427: Is the `ResourceContainsSecrets` needed for this API? Can you remind me why?

patelchandni: This API returns sync triggers payload which contains secrets as well. So treating it the same ways we treat Keycontroller APIs using ResourceContainsSecrets. I have added you to the Teams thread that has more context.

src/WebJobs.Script.WebHost/Management/GetTriggersResult.cs
patelchandni: Changed to TriggersPayload

patelchandni: done, thanks

src/WebJobs.Script.WebHost/Controllers/HostController.cs
patelchandni: @kshyju - Can you please elaborate in what cases will IsSyncTriggersEnvironment return false?
I just want to make sure how relevant it is for deployment scenarios and if we are missing anything else.

mathewc: The sync triggers API also returns a 500 in the same situation, so I'd say for now we should just follow suit here. A 4xx status code isn't really appropriate I think since they indicate a client error - i.e. generally the client should change their request and try again. That doesn't apply here. A 5xx status is most appropriate I think.

src/WebJobs.Script.WebHost/Management/TriggersOperationResult.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
 {
-    public class SyncTriggersResult
+    public class TriggersOperationResult
     {
         /// <summary>
         /// Gets or sets a value indicating whether the sync operation was successful.
```
mathewc: Gets or sets a value indicating whether the operation was successful.

mathewc: Fix below as well

src/WebJobs.Script.WebHost/Management/IFunctionsSyncManager.cs
```diff
@@ -11,7 +11,13 @@ public interface IFunctionsSyncManager
...
-        /// <returns>The <see cref="SyncTriggersResult"/> for the request.</returns>
-        Task<SyncTriggersResult> TrySyncTriggersAsync(bool isBackgroundSync = false);
+        /// <returns>The <see cref="TriggersOperationResult"/> for the request.</returns>
+        Task<TriggersOperationResult> TrySyncTriggersAsync(bool isBackgroundSync = false);
+
+        /// <summary>
+        /// Get function triggers payload.
+        /// </summary>
+        /// <returns>The <see cref="TriggersPayload"/> for the request.</returns>
+        Task<TriggersPayload> GetTriggersPayloadAsync();
```
mathewc: Name this method GetTriggersAsync

src/WebJobs.Script.WebHost/Management/TriggersPayload.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
+{
+    public sealed class TriggersPayload : TriggersOperationResult
```
mathewc: Name this class GetTriggersResult - I thought we had already agreed on that based on your other comments? It's not just a payload, it's also a result

patelchandni: But then Shyju asked to remove "Get" from the name. How about TriggersResult? or I'm fine with GetTriggersResult as well.

src/WebJobs.Script.WebHost/Management/TriggersOperationResult.cs
patelchandni: updated comments

src/WebJobs.Script.WebHost/Management/IFunctionsSyncManager.cs
patelchandni: Changed to GetTriggersAsync

src/WebJobs.Script.WebHost/Controllers/HostController.cs
jviau: I agree with Shyju that `500` is not ideal - there are better fitting codes. It would be helpful to change it from 500 to identify the difference between our implementation rejected the request vs we genuinely faulted.

Some suggestions:
`404 Not Found`: works because the environment is not supported, and thus sync triggers cannot be 'found'.
`405 Method Not Allowed`: we can argue that the `GET` method is not allowed (but also no other methods are allowed...)
`406 Not Acceptable`: the request is not acceptable as we do not support it in this environment.

`501 Not Implemented`: can also argue this method is not implemented for this environment. This is my least favorite though because overall we should avoid ever returning 5xx ourselves as part of an expected scenario.

src/WebJobs.Script.WebHost/Management/TriggersPayload.cs
```diff
@@ -0,0 +1,13 @@
...
+    {
+        /// <summary>
+        /// Gets or sets the triggers payload.
+        /// </summary>
+        public string Content { get; set; }
```
jviau: I am not a fan of using `string Content` here. I much prefer returning a serializable object and letting the MVC formatters serialize it. I also get the current implementation does manual serialization and `JObject` interactions.

Can you evaluate the work involved with changing this?

src/WebJobs.Script.WebHost/Management/TriggersPayload.cs
mathewc: Sure, TriggersResult is fine :)

src/WebJobs.Script.WebHost/Management/TriggersPayload.cs
mathewc: Agree usually, but in this case this was done because [GetSyncTriggersPayload](https://github.com/Azure/azure-functions-host/blob/7e3e7b78fa90ec18eaefd995a9e7797d326d2d00/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs#L309) exposes a string contract because internally it's doing size checks/truncation. I'd rather not go refactor all that right now. The work would be to change the contract of the existing SyncTriggersPayload, and make callers responsible for json formatting etc. Currently that logic is centralized in this method.

Another point here is that we need to maintain parity between this new GET API, and the existing SyncTriggers POST API, so we want to use the same serialization in both cases.

src/WebJobs.Script.WebHost/Management/TriggersPayload.cs
patelchandni: Changed to TriggersResult

src/WebJobs.Script.WebHost/Controllers/HostController.cs
patelchandni: Thank you for the feedback! 
To give some background, the primary purpose of this PR is to address deployment-related changes. It does not aim to modify or optimize the existing sync trigger functionality. This new API is only indented to be used by Functions Deployment Manager. 
Modifying the return status code would require updates to the GetSyncTriggersPayload() method to convey IsSyncTriggersEnvironment is some way, which is used across the current sync trigger API. Optimizing sync trigger functionality is outside the scope of this PR and can have unintended side effects. 
If there are any questions or need for further clarity, I’d be happy to schedule a quick sync to align.

mathewc: Agree - I don't think we should do that now. The SyncTriggers and new GetTriggers APIs should remain consistent w.r.t their status codes. Long ago I chose 500 for SyncTriggers, and I think we should just follow suite here. If we want to change this to another status code pattern, then I think that'd be a separate work item to address.

jviau: > The SyncTriggers and new GetTriggers APIs should remain consistent w.r.t their status codes

Why? This is a public API surface we will need to maintain backwards compatibility with for the foreseeable future. Why not take this opportunity to improve upon the problems with the initial API?

> I think that'd be a separate work item to address.

Given this is a public API, that is **significantly** more work to change it afterwards instead of at its conception.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -786,6 +786,46 @@ internal HttpRequestMessage BuildSetTriggersRequest()
...
+            {
+                result.Success = false;
+                result.Error = "Invalid environment for GetTriggers operation.";
+                _logger.LogWarning(result.Error);
+                return result;
+            }
+
+            try
+            {
+                await _syncSemaphore.WaitAsync();
```
mathewc: I was going to point out that if we're going to be synchronizing here as well, we'd rename the semaphore to _triggerOperationSemaphore or something. However, I don't think we even need syncronization here and can just remove it from this method. The synchronization is needed for SyncTriggers because it's performing multiple steps that need to be in sync. For the get API we don't need this.

src/WebJobs.Script.WebHost/Management/TriggersPayload.cs
jviau: > Another point here is that we need to maintain parity between this new GET API, and the existing SyncTriggers POST API, so we want to use the same serialization in both cases.

It makes no sense for a GET api to double serialize JSON. The POST API makes sense because we are writing that to storage, it is not part of the API response. Including the `string Content` in the API response though is a whole different story and we **should not** do that. You are going to have this response:

``` json
{
    "content": "<json-serialized-string>"
}
```

That to me is not ideal. Instead we could have:

``` json
{
    "content" : {
       // json serialized content of sync triggers. But its a proper JOBject that will deserialize when deserializing the entire payload.
    }
}
```

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -786,6 +786,46 @@ internal HttpRequestMessage BuildSetTriggersRequest()
...
+
+                PrepareSyncTriggers();
```
mathewc: I think we should move PrepareSyncTriggers into GetSyncTriggersPayload, since it needs to be done in all cases, and we ensure that it's always done any time GetSyncTriggers is called. Wasn't an issue when we had only one caller, but now we have multiple

test/WebJobs.Script.Tests.Integration/Management/FunctionsSyncManagerTests.cs
```diff
@@ -256,6 +256,22 @@ private void ResetMockFileSystem(string hostJsonContent = null, string extension
             FileUtility.Instance = fileSystem;
         }
 
+        [Fact]
+        public async Task GetTriggersPayload_ExpectedContent()
```
mathewc: GetTriggersAsync_ReturnsExpectedContent

src/WebJobs.Script.WebHost/Controllers/HostController.cs
patelchandni: This new API is only indented to be used by Functions Deployment Manager. So, it is not intended to be a public API. We are going to restrict it to Validation pod. If this API is called targeting a validation pod during deployment, only then it will return response. Or else it will return 404. This validation pod checks are going to be part of next PR: https://github.com/Azure/azure-functions-host/pull/11024

src/WebJobs.Script.WebHost/Management/TriggersPayload.cs
patelchandni: Mainly, we are using this API to perform sync triggers during deployment so we want the payload to be exactly what sync triggers is using.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
patelchandni: makes sense, removed it.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
patelchandni: Yes, moved PrepareSyncTriggers() call into GetSyncTriggersPayload()

test/WebJobs.Script.Tests.Integration/Management/FunctionsSyncManagerTests.cs
patelchandni: done

mathewc (approved):

src/WebJobs.Script.WebHost/Management/TriggersPayload.cs
patelchandni: Added remark:
```
/// <summary>
/// Gets or sets the triggers payload.
/// </summary>
/// <remarks>
/// Use of an already formatted JSON string here as opposed to a JObject is by design.
/// We want to ensure that the result we return here is the exact same JSON payload as is generated during a sync triggers operation (down to whitespacing).
/// </remarks>
public string Content { get; set; }

File: src/WebJobs.Script.WebHost/Controllers/HostController.cs (hunk 10 above)
Comment (patelchandni) [reply src/WebJobs.Script.WebHost/Controllers/HostController.cs]: Changed to Status403Forbidden. And GetTriggersAsync() will not catch exception, so when it throws exception, middleware will automatically make it a 500 status code.

File: src/WebJobs.Script.WebHost/Controllers/HostController.cs [hunk 19]
```diff
@@ -421,6 +421,22 @@ public async Task<IActionResult> SyncTriggers()
...
+        [RequiresRunningHost]
+        public async Task<IActionResult> GetTriggers()
+        {
+            _metricsLogger.LogEvent(MetricEventNames.GetTriggersInvoked);
+
+            var result = await _functionsSyncManager.GetTriggersAsync();
+
+            return result.Success
+                ? Ok(result)
+                : StatusCode(StatusCodes.Status403Forbidden, new { status = result.Error });
```
jviau, line 440: Can we add a comment explaining the 403 forbidden. Suggestion:

``` CSharp
// GetTriggersAsync() does not swallow exceptions, so any failures will still be a 500. The only result.Success == false we need to consider is when the environment does not support sync triggers.
```

jviau (approved):

mathewc (approved):

src/WebJobs.Script.WebHost/Management/TriggersResult.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
+{
+    public sealed class TriggersResult : TriggersOperationResult
+    {
+        /// <summary>
+        /// Gets or sets the triggers payload.
+        /// </summary>
+        /// <remarks>
+        /// Use of an already formatted JSON string here as opposed to a JObject is by design.
+        /// We want to ensure that the result we return here is the exact same JSON payload as is generated during a sync triggers operation (down to whitespacing).
```
mathewc: nit: Only if you have to make other changes - I'd add a line break here to even this out. Otherwise just fix in your next PR.

patelchandni: done :)

src/WebJobs.Script.WebHost/Controllers/HostController.cs
patelchandni: Added the comment :)

eliaslopezgt (approved):

fabiocav (approved):

PR #11024, opened 2025-04-23
In validation mode, disable background sync trigger and warmup function

patelchandni:

1. Detect validation mode using: `FunctionGroup = deployment:validation`
1. If not validation mode, GET admin/host/triggers API will return 403
1. If in validation mode:
- Updated ShouldSyncTriggers() to return false.
- Disable Warmup Function.

### Pull request checklist

    * Otherwise: Documentation issue linked to PR
    * Otherwise: I've added my notes to `release_notes.md`
    * Otherwise: Backport tracked by issue/PR #issue_or_pr

### DIFF
--- a/src/WebJobs.Script.WebHost/Controllers/HostController.cs
+++ b/src/WebJobs.Script.WebHost/Controllers/HostController.cs
@@ -435,8 +435,8 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Controllers
             {
                 var result = await _functionsSyncManager.GetTriggersAsync();
 
-                // GetTriggersAsync() does not swallow exceptions,so any failures will still be a 500.
-                // The only result.Success == false we need to consideris when the environment
+                // GetTriggersAsync() does not swallow exceptions, so any failures will still be a 500.
+                // The only result.Success == false we need to consider is when the environment
                 // does not support sync triggers.
                 return result.Success
                     ? Ok(result)
--- a/src/WebJobs.Script.WebHost/Controllers/HostController.cs
+++ b/src/WebJobs.Script.WebHost/Controllers/HostController.cs
@@ -430,14 +430,20 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Controllers
         {
             _metricsLogger.LogEvent(MetricEventNames.GetTriggersInvoked);
 
-            var result = await _functionsSyncManager.GetTriggersAsync();
+            // This endpoint is only available in validation mode
+            if (_environment.IsInValidationMode())
+            {
+                var result = await _functionsSyncManager.GetTriggersAsync();
 
-            // GetTriggersAsync() does not swallow exceptions,so any failures will still be a 500.
-            // The only result.Success == false we need to consideris when the environment
-            // does not support sync triggers.
-            return result.Success
-                ? Ok(result)
-                : StatusCode(StatusCodes.Status403Forbidden, new { status = result.Error });
+                // GetTriggersAsync() does not swallow exceptions,so any failures will still be a 500.
+                // The only result.Success == false we need to consideris when the environment
+                // does not support sync triggers.
+                return result.Success
+                    ? Ok(result)
+                    : StatusCode(StatusCodes.Status403Forbidden, new { status = result.Error });
+            }
+
+            return NotFound();
         }
 
         [HttpPost]
--- a/src/WebJobs.Script.WebHost/Controllers/HostController.cs
+++ b/src/WebJobs.Script.WebHost/Controllers/HostController.cs
@@ -443,7 +443,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Controllers
                     : StatusCode(StatusCodes.Status403Forbidden, new { status = result.Error });
             }
 
-            return NotFound();
+            return StatusCode(StatusCodes.Status403Forbidden, new { status = "GetTriggers operation is Forbidden." });
         }
 
         [HttpPost]
--- a/src/WebJobs.Script.WebHost/Controllers/HostController.cs
+++ b/src/WebJobs.Script.WebHost/Controllers/HostController.cs
@@ -443,7 +443,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Controllers
                     : StatusCode(StatusCodes.Status403Forbidden, new { status = result.Error });
             }
 
-            return StatusCode(StatusCodes.Status403Forbidden, new { status = "GetTriggers operation is Forbidden." });
+            return StatusCode(StatusCodes.Status403Forbidden, new { status = "GetTriggers operation is not supported in for this instance." });
         }
 
         [HttpPost]
--- a/src/WebJobs.Script.WebHost/FunctionsSyncService.cs
+++ b/src/WebJobs.Script.WebHost/FunctionsSyncService.cs
@@ -21,10 +21,11 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
         private readonly IScriptHostManager _scriptHostManager;
         private readonly IPrimaryHostStateProvider _primaryHostStateProvider;
         private readonly IFunctionsSyncManager _functionsSyncManager;
+        private readonly IEnvironment _environment;
         private Timer _syncTimer;
         private bool _disposed = false;
 
-        public FunctionsSyncService(ILoggerFactory loggerFactory, IScriptHostManager scriptHostManager, IPrimaryHostStateProvider primaryHostStateProvider, IFunctionsSyncManager functionsSyncManager)
+        public FunctionsSyncService(ILoggerFactory loggerFactory, IScriptHostManager scriptHostManager, IPrimaryHostStateProvider primaryHostStateProvider, IFunctionsSyncManager functionsSyncManager, IEnvironment environment)
         {
             ArgumentNullException.ThrowIfNull(loggerFactory);
 
@@ -32,6 +33,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
             _scriptHostManager = scriptHostManager;
             _primaryHostStateProvider = primaryHostStateProvider;
             _functionsSyncManager = functionsSyncManager;
+            _environment = environment;
             _logger = loggerFactory.CreateLogger(ScriptConstants.LogCategoryHostGeneral);
         }
 
@@ -43,7 +45,8 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
             get
             {
                 return _primaryHostStateProvider.IsPrimary &&
-                       (_scriptHostManager.State == ScriptHostState.Running);
+                       (_scriptHostManager.State == ScriptHostState.Running) &&
+                       !_environment.IsInValidationMode();
             }
         }
 
--- a/src/WebJobs.Script.WebHost/WebScriptHostHttpRoutesManager.cs
+++ b/src/WebJobs.Script.WebHost/WebScriptHostHttpRoutesManager.cs
@@ -69,7 +69,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
 
                     LogRouteMap(routesLogBuilder, function.Metadata.Name, route, httpTrigger.Methods, isProxy, _httpOptions.Value.RoutePrefix);
                 }
-                else if (warmupRouteBuilder != null && function.IsWarmupFunction())
+                else if (warmupRouteBuilder != null && function.IsWarmupFunction(_environment))
                 {
                     warmupRouteBuilder.MapFunctionRoute(function.Metadata.Name, "warmup", function.Metadata.Name);
                 }
--- a/src/WebJobs.Script.WebHost/WebScriptHostHttpRoutesManager.cs
+++ b/src/WebJobs.Script.WebHost/WebScriptHostHttpRoutesManager.cs
@@ -69,7 +69,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
 
                     LogRouteMap(routesLogBuilder, function.Metadata.Name, route, httpTrigger.Methods, isProxy, _httpOptions.Value.RoutePrefix);
                 }
-                else if (warmupRouteBuilder != null && function.IsWarmupFunction(_environment))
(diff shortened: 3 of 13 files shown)


src/WebJobs.Script/Description/FunctionGroups.cs
```diff
@@ -11,6 +11,9 @@ public static class FunctionGroups
         public const string Durable = "durable";
         public const string Blob = "blob";
 
+        // Function group for validation worker used by Functions Deployment Manager
+        public const string ValidationWorker = "deployment:validation";
```
eliaslopezgt, line 15:
```suggestion
        public const string ValidationWorkerFunctionGroup = "deployment:validation";
```

patelchandni: Wouldn't that be redundant? It is already `FunctionGroups.ValidationWorker`

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -430,14 +430,20 @@ public async Task<IActionResult> GetTriggers()
...
+                var result = await _functionsSyncManager.GetTriggersAsync();
 
-            // GetTriggersAsync() does not swallow exceptions,so any failures will still be a 500.
-            // The only result.Success == false we need to consideris when the environment
-            // does not support sync triggers.
-            return result.Success
-                ? Ok(result)
-                : StatusCode(StatusCodes.Status403Forbidden, new { status = result.Error });
+                // GetTriggersAsync() does not swallow exceptions,so any failures will still be a 500.
+                // The only result.Success == false we need to consideris when the environment
```
mathewc: Spelling error from previous PR

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -430,14 +430,20 @@ public async Task<IActionResult> GetTriggers()
...
+                // does not support sync triggers.
+                return result.Success
+                    ? Ok(result)
+                    : StatusCode(StatusCodes.Status403Forbidden, new { status = result.Error });
+            }
+
+            return NotFound();
```
mathewc: Hmmm, why not just also return 403 in this case? The endpoint does exist (so not a 404), however just as above due to host conditions we don't allow the call.

src/WebJobs.Script/Host/ScriptJobHostExtensions.cs
```diff
@@ -33,9 +33,9 @@ private static bool IsFunctionNameMatch(string functionName, string comparison)
         /// Checks if a function is a warmup function
         /// </summary>
         /// <returns>true if a warmup function or else false</returns>
-        public static bool IsWarmupFunction(this FunctionDescriptor function)
+        public static bool IsWarmupFunction(this FunctionDescriptor function, IEnvironment environment)
```
mathewc: Wondering why we need to modify this method. Instead, we can just change the code in one place if the goal is to prevent warmup functions from being called. E.g. in InitializeHttpFunctionRoutes we can just not add the mapping. That'd prevent it from being called, right?:

```
else if (warmupRouteBuilder != null && !_environment.IsInValidationMode() && function.IsWarmupFunction())
{
    warmupRouteBuilder.MapFunctionRoute(function.Metadata.Name, "warmup", function.Metadata.Name);
}
```

However can you explain to me why we need to even do that? If the validation pod is being created in an isolated way and FDM completely controls it, it won't send a warmup request, right? So why does the host need to guard against it?

src/WebJobs.Script.WebHost/FunctionsSyncService.cs
```diff
@@ -21,17 +21,19 @@ public class FunctionsSyncService : IHostedService, IDisposable
         private readonly IScriptHostManager _scriptHostManager;
         private readonly IPrimaryHostStateProvider _primaryHostStateProvider;
         private readonly IFunctionsSyncManager _functionsSyncManager;
+        private readonly IEnvironment _environment;
         private Timer _syncTimer;
         private bool _disposed = false;
 
-        public FunctionsSyncService(ILoggerFactory loggerFactory, IScriptHostManager scriptHostManager, IPrimaryHostStateProvider primaryHostStateProvider, IFunctionsSyncManager functionsSyncManager)
+        public FunctionsSyncService(ILoggerFactory loggerFactory, IScriptHostManager scriptHostManager, IPrimaryHostStateProvider primaryHostStateProvider, IFunctionsSyncManager functionsSyncManager, IEnvironment environment)
```
mathewc, line 28: Thinking we don't need to modify this ctor at all. Instead, we change [here ](https://github.com/Azure/azure-functions-host/blob/ce4fa5fe4b468123ed3419e106719ac614d10edb/src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs#L109)to not even register the service in validation mode. I.e. just modify FunctionsSyncManager.IsSyncTriggersEnvironment to do your new check.

src/WebJobs.Script/Description/FunctionGroups.cs
mathewc: Agree that adding "FunctionGroup" would be redundant. Look at all the other function groups.

src/WebJobs.Script.WebHost/FunctionsSyncService.cs
patelchandni: My understanding is that when we ConfigureServices, the pod can be in placeholder mode. If so, then we will not be able to detect if it is in validation mode or not. OR does it happen after specialization?

src/WebJobs.Script.WebHost/Controllers/HostController.cs
patelchandni: Main reason is to differentiate between:
case 1: it is not IsSyncTriggersEnvironment. This will result in 403.
case 2: it is not a validation mode, so the API itself is not supported. I'm open for any other suggestions if not 404 :)

mathewc: Why do we need to differentiate? FDM is the only caller of this API - why do you need to differentiate? In my mind, both of these cases are just saying that the host isn't in a state that permits this API, so we return 403 Forbidden.

src/WebJobs.Script/Description/FunctionGroups.cs
eliaslopezgt: sorry didn't see the name of the file :)

src/WebJobs.Script.WebHost/Controllers/HostController.cs
patelchandni: done :)

src/WebJobs.Script.WebHost/FunctionsSyncService.cs
mathewc: Ah, you may be right - I was thinking that these validation pods were special in some way and that this state would be known before specialization.

src/WebJobs.Script/Host/ScriptJobHostExtensions.cs
patelchandni: That's better. I have made the change. Thank you :)

src/WebJobs.Script.WebHost/Controllers/HostController.cs
patelchandni: you are right. We can say: it is forbidden if not validation mode. Updated the status code. Thank you!

mathewc (approved):

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -430,14 +430,20 @@ public async Task<IActionResult> GetTriggers()
...
-                : StatusCode(StatusCodes.Status403Forbidden, new { status = result.Error });
+                // GetTriggersAsync() does not swallow exceptions, so any failures will still be a 500.
+                // The only result.Success == false we need to consider is when the environment
+                // does not support sync triggers.
+                return result.Success
+                    ? Ok(result)
+                    : StatusCode(StatusCodes.Status403Forbidden, new { status = result.Error });
+            }
+
+            return StatusCode(StatusCodes.Status403Forbidden, new { status = "GetTriggers operation is Forbidden." });
```
jviau: Can we have a more description error message? Just so a caller won't think it is a permission related issue.

```suggestion
            return StatusCode(StatusCodes.Status403Forbidden, new { status = "GetTriggers operation is not supported in for this instance." });
```

src/WebJobs.Script.WebHost/FunctionsSyncService.cs
```diff
@@ -43,7 +45,8 @@ internal bool ShouldSyncTriggers
             get
             {
                 return _primaryHostStateProvider.IsPrimary &&
-                       (_scriptHostManager.State == ScriptHostState.Running);
+                       (_scriptHostManager.State == ScriptHostState.Running) &&
+                       !_environment.IsInValidationMode();
```
jviau, line 49: This looks like it will impact non-flex skus. We should ensure this doesn't happen.

src/WebJobs.Script.WebHost/WebScriptHostHttpRoutesManager.cs
```diff
@@ -69,7 +69,7 @@ public void InitializeHttpFunctionRoutes(IScriptJobHost host)
 
                     LogRouteMap(routesLogBuilder, function.Metadata.Name, route, httpTrigger.Methods, isProxy, _httpOptions.Value.RoutePrefix);
                 }
-                else if (warmupRouteBuilder != null && function.IsWarmupFunction())
+                else if (warmupRouteBuilder != null && !_environment.IsInValidationMode() && function.IsWarmupFunction())
```
jviau, line 72: Same here, looks like it will impact non-flex skus. We should ensure this doesn't happen.

jviau (changes requested):

src/WebJobs.Script.WebHost/FunctionsSyncService.cs
jviau, line 28: I prefer to avoid any new usage of `IEnvironment`. This type is entirely superseded by `IConfiguration`, but more preferably an options pattern.

src/WebJobs.Script.WebHost/Controllers/HostController.cs
patelchandni: Updated status text :)

eliaslopezgt (approved):

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -748,5 +749,14 @@ public static bool TryGetFunctionsTargetGroup(this IEnvironment environment, out
 
             return group is not null;
         }
+
+        /// <summary>
+        /// Returns true if the worker is running in validation mode.
+        /// </summary>
+        public static bool IsInValidationMode(this IEnvironment environment)
+        {
+            return string.Equals(environment.GetEnvironmentVariable(EnvironmentSettingNames.FunctionsTargetGroup),
```
eliaslopezgt: should we be using `_environment.TryGetFunctionsTargetGroup(out string targetGroup)`

patelchandni: good catch, thanks :)

src/WebJobs.Script.WebHost/FunctionsSyncService.cs
mathewc: Can you point us to an example extension method pattern and usage of IConfiguration in another service that you'd like her to follow here? If we're really serious about blocking all usage of IEnvironment, I think we need a team wide memo on this, and perhaps to mark the interface deprecated or something. Because I myself am not aware of any agreed upon edict here and PRs that are going in are not consistent in this regard.

We use IEnvironment all over the place along with its associated helper extensions, so if we really want to switch over, we should have a work item (with justification) to do that wholesale through the platform.

patelchandni: But everywhere else the FunctionsTargetGroup is being read from IEnvironment. Not sure how will option pattern work in this case?

patelchandni: Mainly, we do not want to divert on a path that is not commonly used by host and then have unexpected result.

jviau: @mathewc yes we should send out a team wide memo on it. We haven't taken to steps to obsolete `IEnvironment` because it is so extensively used that will be a major work item which we haven't gotten around to.

As for using `IConfiguration` in place of `IEnvironment`, it will contain all environment variables already. So you just need to index into it: `configuration["My-Environment-Variable"]`. The extra suggestion I have is to use an options pattern, which replaces the loosely typed access into configuration with a strongly typed object, which is typically populated from config. https://learn.microsoft.com/en-us/dotnet/core/extensions/options

Main highlights for options pattern from the above link:

> The options pattern uses classes to provide strongly typed access to groups of related settings. When [configuration settings](https://learn.microsoft.com/en-us/dotnet/core/extensions/configuration) are isolated by scenario into separate classes, the app adheres to two important software engineering principles:
> - The [Interface Segregation Principle (ISP) or Encapsulation](https://learn.microsoft.com/en-us/dotnet/architecture/modern-web-apps-azure/architectural-principles#encapsulation): Scenarios (classes) that depend on configuration settings depend only on the configuration settings that they use.
> - [Separation of Concerns](https://learn.microsoft.com/en-us/dotnet/architecture/modern-web-apps-azure/architectural-principles#separation-of-concerns): Settings for different parts of the app aren't dependent or coupled with one another.
> 
> Options also provide a mechanism to validate configuration data. For more information, see the [Options validation](https://learn.microsoft.com/en-us/dotnet/core/extensions/options#options-validation) section.

jviau: This is also just a preference, you don't have to make this change.

eliaslopezgt (approved):

src/WebJobs.Script.WebHost/FunctionsSyncService.cs
eliaslopezgt: being this concern a wider refactor effort do you think we can get this PR approved? @jviau

jviau (approved):

PR #11035, opened 2025-04-25
Custom Handler Proxying

### LINKED ISSUE #11025: Update custom handlers to proxy requests to custom workers

satvu:

resolves #11025

### Pull request checklist

    * [x] Otherwise: Documentation issue linked to PR - TODO


src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptions.cs
```diff
@@ -15,8 +15,16 @@ public class HttpWorkerOptions
 
         public int Port { get; set; }
 
+        /// <summary>
+        /// Setting that enables the host to rebuild the initial invocation HTTP Request and send the copy to the worker process.
```
kshyju: Are you not seeing [SA1623 ](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1623.md)warning in VS? VS should also offer a context menu option to fix it for you.

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptionsSetup.cs
```diff
@@ -46,7 +46,7 @@ public void Configure(HttpWorkerOptions options)
                 ConfigureWorkerDescription(options, customHandlerSection);
                 if (options.Type == CustomHandlerType.None)
                 {
-                    // CustomHandlerType.None is only for maitaining backward compatibilty with httpWorker section.
+                    // CustomHandlerType.None is only for maintaining backward compatibilty with httpWorker section.
```
kshyju: typo in "compatibilty"

src/WebJobs.Script/Workers/Http/DefaultHttpWorkerService.cs
```diff
@@ -165,6 +206,20 @@ internal void AddHeaders(HttpRequestMessage httpRequest, string invocationId)
...
+        {
+            httpRequest.Headers.TryAdd(HttpWorkerConstants.HostVersionHeaderName, ScriptHost.Version);
+
+            if (!httpRequest.Headers.TryAdd(HttpWorkerConstants.InvocationIdHeaderName, invocationId))
+            {
+                httpRequest.Headers[HttpWorkerConstants.InvocationIdHeaderName] = invocationId;
+            }
+
+            var userAgent = $"{HttpWorkerConstants.UserAgentHeaderValue}/{ScriptHost.Version}";
+            httpRequest.Headers.Remove("User-Agent");
```
kshyju: Can be simplified to 

```
httpRequest.Headers.UserAgent = userAgent;
```

src/WebJobs.Script/Workers/Http/DefaultHttpWorkerService.cs
```diff
@@ -165,6 +206,20 @@ internal void AddHeaders(HttpRequestMessage httpRequest, string invocationId)
             httpRequest.Headers.UserAgent.ParseAdd($"{HttpWorkerConstants.UserAgentHeaderValue}/{ScriptHost.Version}");
         }
 
+        internal void AddProxyingHeaders(HttpRequest httpRequest, string invocationId)
```
kshyju: should this be `private`?

src/WebJobs.Script/Workers/Http/DefaultHttpWorkerService.cs
```diff
@@ -71,6 +89,29 @@ public Task InvokeAsync(ScriptInvocationContext scriptInvocationContext)
...
+                    throw new InvalidOperationException($"Cannot proxy the HttpTrigger function {scriptInvocationContext.FunctionMetadata.Name} without an input of type {nameof(HttpRequest)}.");
+                }
+
+                AddProxyingHeaders(httpRequest, scriptInvocationContext.ExecutionContext.InvocationId.ToString());
+
+                // YARP only requires the destination prefix. The path and query string are added by the YARP proxy during SendAsync using info from the HttpContext.
+                _httpProxyService.StartForwarding(scriptInvocationContext, _destinationPrefix);
+
+                await _httpProxyService.EnsureSuccessfulForwardingAsync(scriptInvocationContext);
+                scriptInvocationContext.ResultSource.SetResult(_successfulInvocationResult);
```
kshyju: Does the downstream code updates this `_successfulInvocationResult` instance? I see we use a shared instance for all invocations.

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -289,6 +289,8 @@ public static IHostBuilder AddScriptHostCore(this IHostBuilder builder, ScriptAp
                 // Core WebJobs/Script Host services
                 services.AddSingleton<ScriptHost>();
 
+                services.AddSingleton<IHttpProxyService, DefaultHttpProxyService>();
```
kshyju: Could this be conditionally registered? Only when custom handlers is enabled?

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptions.cs
liliankasem: Yeah please switch to the correct docs for this "Gets or sets _"

src/WebJobs.Script/HttpProxyService/HttpForwardingException.cs
```diff
@@ -3,7 +3,7 @@
 
 using System;
 
-namespace Microsoft.Azure.WebJobs.Script.Grpc.Exceptions
+namespace Microsoft.Azure.WebJobs.Script
```
liliankasem: Why is the namespace changing for all these files?

src/WebJobs.Script/Workers/Http/DefaultHttpWorkerService.cs
liliankasem: Can we put strings like this in consts

src/WebJobs.Script/HttpProxyService/HttpForwardingException.cs
satvu: These files were all moved to the `WebJobs.Script` project (they were previously in the `Grpc` project).

src/WebJobs.Script/Workers/Http/DefaultHttpWorkerService.cs
satvu: No, the downstream code does not. This value is set to allow the function invocation to continue, but its value is never used (similar to how http proxying flows). We ignore the function invocation result at [this step](https://github.com/Azure/azure-functions-host/blob/2557cf38e1de490f4ea423492ea2ec23fb78f24a/src/WebJobs.Script.WebHost/Middleware/FunctionInvocationMiddleware.cs#L60) so that the actual response just flows through.

src/WebJobs.Script/HttpProxyService/HttpForwardingException.cs
liliankasem: Are they common items that will be used by others? Shouldn't this move to `src/WebJobs.Script/Exceptions/HttpForwardingException.cs`  and vice versa for the rest? If not, should you add `HttpProxyService` to namespace? i.e. `namespace Microsoft.Azure.WebJobs.Script.HttpProxyService`

src/WebJobs.Script/Workers/Http/DefaultHttpWorkerService.cs
```diff
@@ -47,6 +55,14 @@ internal DefaultHttpWorkerService(HttpClient httpClient, IOptions<HttpWorkerOpti
...
                 _httpClient.Timeout = scriptHostOptions.Value.FunctionTimeout.Value.Add(TimeSpan.FromMinutes(1));
             }
+
+            _successfulInvocationResult = new ScriptInvocationResult()
+            {
+                Outputs = new Dictionary<string, object>()
+            };
+
+            _destinationPrefix = new UriBuilder(WorkerConstants.HttpScheme, WorkerConstants.HostName, _httpWorkerOptions.Port).Uri;
+            _userAgentString = $"{HttpWorkerConstants.UserAgentHeaderValue}/{ScriptHost.Version}";
```
liliankasem, line 59: Does this ever get set outside of the ctor? Could this be a const? It doesnt look like its made up of anything that would require it to be init in here no?

satvu: Doesn't get set outside of the ctor, but `ScriptHost.Version` isn't a constant so this can't be init as a constant outside of the ctor at the top of the file.

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
satvu: With the cleanup and refactor, this service will also be used for HTTP Proxying for worker http streaming. Information about streaming support/capabilities won't be known until there is a worker init response (which occurs much later than this stage).

src/WebJobs.Script/HttpProxyService/DefaultHttpProxyService.cs
```diff
@@ -7,12 +7,12 @@
...
 using Microsoft.AspNetCore.Http;
 using Microsoft.Azure.WebJobs.Script.Description;
-using Microsoft.Azure.WebJobs.Script.Grpc.Exceptions;
+using Microsoft.Azure.WebJobs.Script.Exceptions;
 using Microsoft.Azure.WebJobs.Script.Workers;
 using Microsoft.Extensions.Logging;
 using Yarp.ReverseProxy.Forwarder;
 
-namespace Microsoft.Azure.WebJobs.Script.Grpc
+namespace Microsoft.Azure.WebJobs.Script.HttpProxyService
```
fabiocav: For the namespace, let's rename this to `Microsoft.Azure.WebJobs.Script.Http` instead.

src/WebJobs.Script/HttpProxyService/IHttpProxyService.cs
```diff
@@ -5,7 +5,7 @@
 using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Script.Description;
 
-namespace Microsoft.Azure.WebJobs.Script.Grpc
+namespace Microsoft.Azure.WebJobs.Script.HttpProxyService
```
fabiocav: Same comment about the namespace.

src/WebJobs.Script/Workers/Http/DefaultHttpWorkerService.cs
```diff
@@ -47,6 +55,14 @@ internal DefaultHttpWorkerService(HttpClient httpClient, IOptions<HttpWorkerOpti
                 // Set 1 minute greater than FunctionTimeout to ensure invoction failure due to timeout is raised before httpClient raises operation cancelled exception
                 _httpClient.Timeout = scriptHostOptions.Value.FunctionTimeout.Value.Add(TimeSpan.FromMinutes(1));
             }
+
+            _successfulInvocationResult = new ScriptInvocationResult()
+            {
+                Outputs = new Dictionary<string, object>()
```
fabiocav: Can you use https://learn.microsoft.com/en-us/dotnet/api/system.collections.immutable.immutabledictionary-2.empty?view=netcore-2.0 here instead?

A good option here would be to expose a `Success` property in `ScriptInvocationResult`

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptions.cs
```diff
@@ -15,8 +15,16 @@ public class HttpWorkerOptions
 
         public int Port { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether the host will rebuild the initial invocation HTTP Request and send the copy to the worker process.
+        /// </summary>
         public bool EnableForwardingHttpRequest { get; set; }
```
kshyju, line 24: A bit unclear about the property name and the summary comment. From the name, it sounds like the request is simply being forwarded as it is. But the comment suggests that a new copy of the request is being rebuilt from the original and that is sent to the worker process. Should we consider updating the property name to reflect that behavior?

src/WebJobs.Script/Workers/Http/DefaultHttpWorkerService.cs
```diff
@@ -47,6 +54,9 @@ internal DefaultHttpWorkerService(HttpClient httpClient, IOptions<HttpWorkerOpti
                 // Set 1 minute greater than FunctionTimeout to ensure invoction failure due to timeout is raised before httpClient raises operation cancelled exception
                 _httpClient.Timeout = scriptHostOptions.Value.FunctionTimeout.Value.Add(TimeSpan.FromMinutes(1));
             }
+
+            _destinationPrefix = new UriBuilder(WorkerConstants.HttpScheme, WorkerConstants.HostName, _httpWorkerOptions.Port).Uri;
```
kshyju, line 58: Is it a valid scenario where a customer configures their web app to only accept HTTPS traffic, but this code affects request forwarding because it explicitly uses HTTP? Should we allow the customer to specify or override the scheme?

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptions.cs
satvu: This is an existing property (see [schema store](https://github.com/SchemaStore/schemastore/blob/6a8d70c729f559da8455984f40e4615e9a1620a9/src/schemas/json/host.json#L1060)) and we cannot change this without having breaking changes. It's also already [documented](https://learn.microsoft.com/en-us/azure/azure-functions/functions-custom-handlers#http-only-function). 

This feature currently works by handling http requests [here](https://github.com/Azure/azure-functions-host/blob/265424c38d4f1aa453121735ae69fa8423acc290/src/WebJobs.Script/Workers/Http/DefaultHttpWorkerService.cs#L74) and calling[ this extension method](https://github.com/Azure/azure-functions-host/blob/265424c38d4f1aa453121735ae69fa8423acc290/src/WebJobs.Script/Extensions/HttpRequestExtensions.cs#L155).

It's a really unfortunate name - I believe the plan is to phase this setting out in favor of the new one (for streaming). I added the comments to try to make clear the difference between the two very similar property names.

src/WebJobs.Script/Workers/Http/DefaultHttpWorkerService.cs
satvu: I think your scenario could be valid - @fabiocav, is this an issue customers run into today? I borrowed from the existing pattern (for the `EnableForwardingHttpRequest` prop) in `DefaultHttpWorkerService` (see [here](https://github.com/Azure/azure-functions-host/blob/265424c38d4f1aa453121735ae69fa8423acc290/src/WebJobs.Script/Workers/Http/DefaultHttpWorkerService.cs#L291)) which only uses HTTP.

src/WebJobs.Script/Workers/Http/DefaultHttpWorkerService.cs
```diff
@@ -71,6 +86,29 @@ public Task InvokeAsync(ScriptInvocationContext scriptInvocationContext)
             return ProcessDefaultInvocationRequest(scriptInvocationContext);
```
kshyju, line 86: line 84 can be removed (since 86 does the same)?

src/WebJobs.Script/Workers/Http/DefaultHttpWorkerService.cs
```diff
@@ -71,6 +86,29 @@ public Task InvokeAsync(ScriptInvocationContext scriptInvocationContext)
...
+
+                // YARP only requires the destination prefix. The path and query string are added by the YARP proxy during SendAsync using info from the HttpContext.
+                _httpProxyService.StartForwarding(scriptInvocationContext, _destinationPrefix);
+
+                await _httpProxyService.EnsureSuccessfulForwardingAsync(scriptInvocationContext); // this will throw if forwarding is unsuccessful
+                scriptInvocationContext.ResultSource.SetResult(ScriptInvocationResult.Success);
+            }
+            catch (Exception exc)
+            {
+                scriptInvocationContext.ResultSource.TrySetException(exc);
```
kshyju, line 108: Should this exception be logged? (` EnsureSuccessfulForwardingAsync` could throw)

test/WebJobs.Script.Tests/HttpWorker/DefaultHttpWorkerServiceTests.cs
```diff
@@ -510,12 +516,128 @@ public async Task IsWorkerReady_Returns_True()
...
+
+            await _defaultHttpWorkerService.ProxyInvocationRequest(testScriptInvocationContext);
+
+            Assert.True(testScriptInvocationContext.ResultSource.Task.IsFaulted);
+            Assert.IsType<InvalidOperationException>(testScriptInvocationContext.ResultSource.Task.Exception.InnerException);
+            Assert.Equal("Cannot proxy the HttpTrigger function TestFunction without an input of type HttpRequest.", testScriptInvocationContext.ResultSource.Task.Exception.InnerException.Message);
+        }
+
+        [Fact]
+        public async Task ProxyInvocationRequest_HandlesForwardingError()
```
kshyju, line 598: Are we missing a test which covers the scenarios where the request either times out or receives a non-2xx response from the custom handler app?

src/WebJobs.Script/Workers/Http/DefaultHttpWorkerService.cs
satvu: This follows the same pattern we use in the [gRPC worker channel](https://github.com/Azure/azure-functions-host/blob/265424c38d4f1aa453121735ae69fa8423acc290/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs#L1141) (for both gRPC and proxying invocations). I believe the result source will be handled [here](https://github.com/Azure/azure-functions-host/blob/0622c73d98cf1aa46023b41f61653d8f144731f4/src/WebJobs.Script/Description/Workers/WorkerFunctionInvoker.cs#L101) and the [exception middleware](https://github.com/Azure/azure-functions-host/blob/0622c73d98cf1aa46023b41f61653d8f144731f4/src/WebJobs.Script.WebHost/Middleware/ExceptionMiddleware.cs) should help with logging the exception.

test/WebJobs.Script.Tests/HttpWorker/DefaultHttpWorkerServiceTests.cs
satvu: The timeout would register as a `ForwarderError` (see types of [forwarder errors](https://learn.microsoft.com/en-us/dotnet/api/yarp.reverseproxy.forwarder.forwardererror?view=yarp-2.2)) which is handled by `DefaultHttpProxyService` and covered in this test case.

For non-2xx responses, as long as forwarding succeeds, we do not handle the final response in the host.

kshyju (approved):

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptions.cs
liliankasem: I think you can update the comment to 

```
/// <summary>
/// Gets or sets a value indicating whether the host will forward the request to the worker process. 
/// </summary>
/// <remarks>
/// The host will rebuild the initial invocation HTTP Request and send the copy to the worker process.
/// </remarks>
```

src/WebJobs.Script/Workers/Http/DefaultHttpWorkerService.cs
```diff
@@ -162,7 +200,15 @@ internal void AddHeaders(HttpRequestMessage httpRequest, string invocationId)
...
             httpRequest.Headers.Add(HttpWorkerConstants.HostVersionHeaderName, ScriptHost.Version);
             httpRequest.Headers.Add(HttpWorkerConstants.InvocationIdHeaderName, invocationId);
-            httpRequest.Headers.UserAgent.ParseAdd($"{HttpWorkerConstants.UserAgentHeaderValue}/{ScriptHost.Version}");
+            httpRequest.Headers.UserAgent.ParseAdd(_userAgentString);
+        }
+
+        private void AddProxyingHeaders(HttpRequest httpRequest, string invocationId)
+        {
+            // if there are existing headers, override them
+            httpRequest.Headers[HttpWorkerConstants.HostVersionHeaderName] = ScriptHost.Version;
```
liliankasem, line 209: Curious why we are not following the same '.Add' pattern as above? Is it because we're overriding? If not could we use Add or AndOrUpdate (if exists)?

liliankasem, line 209: Does AddOrUpdate exist on `Headers`?

src/WebJobs.Script/Description/Workers/ScriptInvocationResult.cs
```diff
@@ -2,11 +2,17 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System.Collections.Generic;
+using System.Collections.Immutable;
 
 namespace Microsoft.Azure.WebJobs.Script.Description
 {
     public class ScriptInvocationResult
     {
+        public static readonly ScriptInvocationResult Success = new ScriptInvocationResult
```
liliankasem, line 11: Not sure I understand why this is named as it is, or why it exists? - aren't outputs captured the prop on line 18 already? If you're trying to reflect state, could we use a bool or enum?

src/WebJobs.Script/Workers/Http/DefaultHttpWorkerService.cs
satvu: Yes, but following the pattern we implemented here to fix a bug: https://github.com/Azure/azure-functions-host/pull/10980. In some cx scenarios, such as when a request makes it from one function app to another, we want the headers of the last host.

src/WebJobs.Script/Description/Workers/ScriptInvocationResult.cs
satvu: This is a suggestion from [this comment](https://github.com/Azure/azure-functions-host/pull/11035#discussion_r2136629958) - similar to how we can grab an "Empty" array (ex. `ImmutableDictionary<string, object>.Empty`), this allows us to circumvent initializing and storing an empty, successful `ScriptInvocationResult` in `DefaultHttpWorkerService`. 

Do you think a name like "SuccessfulResult" is clearer?

liliankasem (approved):
lgtm

src/WebJobs.Script/Description/Workers/ScriptInvocationResult.cs
liliankasem: Naming is ok

src/WebJobs.Script/Workers/Http/DefaultHttpWorkerService.cs
satvu: Discussed with @fabiocav offline, this is outside of the scope of the current changes, but it is a valid comment and may be something we want to address in a future update.

kshyju (approved):

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptionsSetup.cs
```diff
@@ -46,7 +46,7 @@ public void Configure(HttpWorkerOptions options)
                 ConfigureWorkerDescription(options, customHandlerSection);
                 if (options.Type == CustomHandlerType.None)
                 {
-                    // CustomHandlerType.None is only for maitaining backward compatibilty with httpWorker section.
+                    // CustomHandlerType.None is only for maintaining backward compatability with httpWorker section.
```
fabiocav, line 49: This introduced a spelling error here.

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptions.cs
```diff
@@ -15,8 +15,19 @@ public class HttpWorkerOptions
...
+        /// </summary>
+        /// <remarks>
+        /// The host will rebuild the initial invocation HTTP Request and send the copy to the worker process.
+        /// </remarks>
         public bool EnableForwardingHttpRequest { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether the host will proxy the invocation HTTP request to the worker process.
+        /// </summary>
+        public bool EnableProxyingHttpRequest { get; set; }
```
fabiocav, line 29: This came up in R2D.

Doesn't look like `HttpWorkerOptions` implements `IOptionsFormatter`. Can we please follow up with a change to make sure we have telemetry indicading when this feature is used?

PR #11036, opened 2025-04-28
Fixes to improve the reliability of the LinuxContainerActivityPublisher

fabiocav:
Fixes to improve the reliability of the LinuxContainerActivityPublisher. The key issue being resolved is a change to move away from a timer callback that was using the `async void` pattern.

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.WebHost/ContainerManagement/LinuxContainerActivityPublisher.cs
+++ b/src/WebJobs.Script.WebHost/ContainerManagement/LinuxContainerActivityPublisher.cs
@@ -14,7 +14,7 @@ using Microsoft.Extensions.Options;
 
 namespace Microsoft.Azure.WebJobs.Script.WebHost.ContainerManagement
 {
-    public class LinuxContainerActivityPublisher : IHostedService, IDisposable, ILinuxContainerActivityPublisher
+    public sealed class LinuxContainerActivityPublisher : IHostedService, IAsyncDisposable, ILinuxContainerActivityPublisher
     {
         public const string SpecializationCompleteEvent = "SpecializationCompleted";
         private const int InitialFlushIntervalMs = 5 * 1000; // 5 seconds
@@ -30,9 +30,10 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.ContainerManagement
         private readonly HashSet<ContainerFunctionExecutionActivity> _uniqueActivities;
         private IDisposable _standbyOptionsOnChangeSubscription;
         private DateTime _lastHeartBeatTime = DateTime.MinValue;
-        private Timer _timer;
         private int _flushInProgress;
         private bool _initialPublish;
+        private CancellationTokenSource _publishingCts;
+        private Task _publishingTask;
 
         public LinuxContainerActivityPublisher(IOptionsMonitor<StandbyOptions> standbyOptions,
             IMeshServiceClient meshServiceClient, IEnvironment environment,
@@ -49,7 +50,6 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.ContainerManagement
             _logger = logger;
             _flushIntervalMs = flushIntervalMs;
             _initialFlushIntervalMs = initialFlushIntervalMs;
-            _timer = new Timer(OnTimer, null, Timeout.Infinite, Timeout.Infinite);
             _uniqueActivities = new HashSet<ContainerFunctionExecutionActivity>();
             _flushInProgress = 0;
             _initialPublish = true;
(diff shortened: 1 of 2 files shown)


src/WebJobs.Script.WebHost/ContainerManagement/LinuxContainerActivityPublisher.cs
```diff
@@ -94,27 +133,11 @@ public Task StopAsync(CancellationToken cancellationToken)
...
             _logger.LogInformation($"Stopping {nameof(LinuxContainerActivityPublisher)}");
 
-            // stop the timer if it has been started
-            _timer?.Change(Timeout.Infinite, Timeout.Infinite);
+            _publishingCts.Cancel();
 
             return Task.CompletedTask;
         }
 
-        private async void OnTimer(object state)
```
fabiocav, line 103: This is the key problem being addressed by this PR.

src/WebJobs.Script.WebHost/ContainerManagement/LinuxContainerActivityPublisher.cs
```diff
@@ -59,8 +59,47 @@ private void Start()
...
+                        else
+                        {
+                            await FlushFunctionExecutionActivities();
+                        }
+
+                        await Task.Delay(nextFlushDelay, _publishingCts.Token);
+                    }
+                    catch (Exception ex)
+                    {
+                        _logger.LogError(ex,$"Error in {nameof(LinuxContainerActivityPublisher)} publishing loop.");
```
jviau:
```suggestion
                        _logger.LogError(ex, $"Error in {nameof(LinuxContainerActivityPublisher)} publishing loop.");
```

src/WebJobs.Script.WebHost/ContainerManagement/LinuxContainerActivityPublisher.cs
```diff
@@ -169,81 +192,72 @@ private async Task FlushFunctionExecutionActivities()
...
+                _logger.LogWarning($"Failed to add activity {activity}");
+            }
         }
 
-        public void PublishFunctionExecutionActivity(ContainerFunctionExecutionActivity activity)
+        public async ValueTask DisposeAsync()
         {
-            if (!_standbyOptions.CurrentValue.InStandbyMode)
+            // Wait for the publishing task to complete
+            if (_publishingTask != null)
```
jviau, line 258: Do you need to cancel `_publishCts` before `await _publishingTask`? Otherwise if this is disposed without stopping, this would never complete.

src/WebJobs.Script.WebHost/ContainerManagement/LinuxContainerActivityPublisher.cs
```diff
@@ -59,8 +59,47 @@ private void Start()
...
             _logger.LogInformation($"Starting {nameof(LinuxContainerActivityPublisher)}");
 
-            // start the timer by setting the due time
-            SetTimerInterval(_initialFlushIntervalMs);
+            _publishingTask = StartPublishingAsync();
+        }
+
+        private async Task StartPublishingAsync()
+        {
+            _publishingCts = new CancellationTokenSource();
```
jviau, line 67: Since this is only set in one place, should it be done ctor and made `readonly`?

src/WebJobs.Script.WebHost/ContainerManagement/LinuxContainerActivityPublisher.cs
```diff
@@ -30,9 +30,10 @@ public class LinuxContainerActivityPublisher : IHostedService, IDisposable, ILin
         private readonly HashSet<ContainerFunctionExecutionActivity> _uniqueActivities;
         private IDisposable _standbyOptionsOnChangeSubscription;
         private DateTime _lastHeartBeatTime = DateTime.MinValue;
-        private Timer _timer;
         private int _flushInProgress;
         private bool _initialPublish;
+        private CancellationTokenSource _publishingCts;
+        private Task _publishingTask;
```
jviau, line 36: Can this be `readonly`? I don't see the one place (in the ctor) it is set.

src/WebJobs.Script.WebHost/ContainerManagement/LinuxContainerActivityPublisher.cs
fabiocav: Cancellation is expected in `StopAsync`, but I'll add some logic to handle a potential disposal where cancellation was not called.

src/WebJobs.Script.WebHost/ContainerManagement/LinuxContainerActivityPublisher.cs
fabiocav: Debated on that. Though this helps defer initialization to when needed only, in practice, that will almost never be the case.

src/WebJobs.Script.WebHost/ContainerManagement/LinuxContainerActivityPublisher.cs
fabiocav: Based on the previous comment, I assume you're talking about the `_publishingTask`. If so, that is set when start is called, so we can't make this `readonly`.

src/WebJobs.Script.WebHost/ContainerManagement/LinuxContainerActivityPublisher.cs
fabiocav: I'd feel better making that change if we actually prevented multiple start/stop calls in this type, though.

src/WebJobs.Script.WebHost/ContainerManagement/LinuxContainerActivityPublisher.cs
```diff
@@ -59,8 +59,47 @@ private void Start()
...
+                            nextFlushDelay -= _initialFlushIntervalMs;
+                        }
+                        else
+                        {
+                            await FlushFunctionExecutionActivities();
+                        }
+
+                        await Task.Delay(nextFlushDelay, _publishingCts.Token);
+                    }
+                    catch (Exception ex)
```
mathewc, line 91: This catch is the actual fix, right? All the rest is just a refactoring away from a Timer based solution?

Looks like we're using async void timer callbacks in at least 2 other places (GrpcWorkerChannel and WorkerConcurrencyManager) and it looks like the the problem exists in the former as well (uncaught exceptions).

fabiocav: I'm already following up on the other instances.

PR #11075, opened 2025-05-22
add warning log for end of support extension bundles versions

manvkaur:

resolves #https://github.com/Azure/azure-functions-pyfx-planning/issues/713
screenshot of working log in app insights - 
image

### Pull request checklist

    * [ ] Otherwise: Documentation issue linked to PR - https://learn.microsoft.com/en-us/azure/azure-functions/extension-bundles#supported-extension-bundles 

### Additional information
This pull request introduces changes to enhance validation mechanisms for function apps, improve logging for outdated extension bundles.

### Validation Enhancements:
* Added the `IFunctionAppValidator` interface to support modular validation logic for function apps. (`src/WebJobs.Script/Host/IFunctionAppValidator.cs`)
* Introduced `ExtensionBundleManagerValidator` and `MissingAzureFunctionsFolderValidator` classes to validate outdated extension bundles and improper folder structures, respectively. (`src/WebJobs.Script/Host/ExtensionBundleManagerValidator.cs`, `src/WebJobs.Script/Host/MissingAzureFunctionsFolderValidator.cs`) [[1]](diffhunk://#diff-24e75d79a66a16d8e0348cf218d8705cbc9dacd2c16bfa6ffe7e588bccbc6b21R1-R28) [[2]](diffhunk://#diff-b5159c57c930b097519ed350581becd1c14a6b4ac013224252ca0c48cbb37d78R1-R39)
* Updated `FunctionAppValidationService` to use a collection of validators for streamlined validation and improved error handling. (`src/WebJobs.Script/Host/FunctionAppValidationService.cs`) [[1]](diffhunk://#diff-1668a7b6edca7c3f42059cec0fdc4058592e791d9920da617ca7f060eff5d5d0R22-R33) [[2]](diffhunk://#diff-1668a7b6edca7c3f42059cec0fdc4058592e791d9920da617ca7f060eff5d5d0L43-R57)

### Logging Improvements:
* Added a warning log for outdated extension bundles, including localized messages for future and past deprecation dates. (`src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs`, `src/WebJobs.Script/Properties/Resources.Designer.cs`, `src/WebJobs.Script/Properties/Resources.resx`) [[1]](diffhunk://#diff-374e3d774adf38015c92a7b4077f75f152c0b2d19f4f0d3ecebfd2835ae57ce9R209-R213) [[2]](diffhunk://#diff-38b9958d6cd65643e22aff8924b2dce1f5f77949bbe6c4e3447740704483bab6R212-R229) [[3]](diffhunk://#diff-ac7e2526baf2b7bc0492f3bb6b058e7f9a1fcaecaaa54a58b0636ac49f930e7bR182-R187)
* Registered the new logging functionality in the `ExtensionBundleManager` class to generate warnings based on bundle version and deprecation timelines. (`src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs`)

### Platform Configuration Updates:
* Corrected the `x86` bitness mapping in `New-PrivateSiteExtension.ps1` to reflect "32bit" instead of "64bit." (`src/WebJobs.Script.SiteExtension/New-PrivateSiteExtension.ps1`)

### Test Suite Updates:
* Modified test classes to accommodate new validation logic and ensure compatibility with the updated `FunctionAppValidationService`. (`test/WebJobs.Script.Tests.Integration/TestFunctionHost.cs`, `test/WebJobs.Script.Tests/Description/FunctionAppValidationServiceTests.cs`) [[1]](diffhunk://#diff-d62bf451975eb564458955bfbce916b5674295b15d7af22832158544ac73e323R534) [[2]](diffhunk://#diff-d7915eb80fb660b7dbc48c8dcc5ad2bb5fa3cdffb86c0a5721c7f0b408e35a8cL39-R66)

### Miscellaneous:
* Added a warning log entry for outdated extension bundles in the release notes. (`release_notes.md`)


src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -375,5 +377,32 @@ public async Task<string> GetExtensionBundleBinPathAsync()
             // if no bin directory is present something is wrong
             return FileUtility.DirectoryExists(binPath) ? binPath : null;
         }
+
+        public void CompareWithLatestMajorVersion()
```
kshyju: nit: Could the naming be improved to indicate we are validating the version to be outdated or not and logging?

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -375,5 +377,32 @@ public async Task<string> GetExtensionBundleBinPathAsync()
...
+        {
+            string majorVersionStr = _extensionBundleVersion?.Split('.')?.FirstOrDefault() ?? string.Empty;
```
kshyju: You can remove the string array allocation here.
```
ReadOnlySpan<char> versionSpan = _extensionBundleVersion;
var dotIndex = versionSpan.IndexOf('.');
if (dotIndex <= 0 || !int.TryParse(versionSpan.Slice(0, dotIndex), out var majorVersion))
{
    return;
}
```

With this you can also remove the ` majorVersion != 0` check in below if statement.

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -375,5 +377,32 @@ public async Task<string> GetExtensionBundleBinPathAsync()
...
+            int majorVersion = int.TryParse(majorVersionStr, out int result) ? result : 0;
+
+            int latestMajorVersion = ScriptConstants.ExtensionBundleV4MajorVersion;
+            if (string.Compare(_options?.Id, ScriptConstants.DefaultExtensionBundleId, StringComparison.OrdinalIgnoreCase) == 0
+                && majorVersion != 0
+                && majorVersion < latestMajorVersion)
+            {
+                DateTime currentTime = DateTime.UtcNow;
```
kshyju: nit: Consider using implicit types.

src/WebJobs.Script/Host/ExtensionBundleManagerValidator.cs
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Azure.WebJobs.Script.ExtensionBundle;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Script.Host
+{
+    internal class ExtensionBundleManagerValidator : IFunctionAppValidator
```
kshyju:
```suggestion
    internal sealed class ExtensionBundleManagerValidator : IFunctionAppValidator
```

src/WebJobs.Script/Host/FunctionAppValidationService.cs
```diff
@@ -40,39 +40,21 @@ protected override async Task ExecuteAsync(CancellationToken cancellationToken)
...
-                    }
-                    else
-                    {
-                        _logger.MissingAzureFunctionsFolder();
-                    }
+                    validator.Validate(_scriptOptions.Value, _environment, _logger);
+                }
+                catch (Exception ex)
+                {
+                    _logger.LogTrace($"Validator {validator.GetType().Name} failed", ex);
```
kshyju: Should this be a Warning level log entry?

kshyju: Consider using structured logging

```
_logger.LogTrace(ex, "Validator {ValidatorType} failed", validator.GetType().Name);
```

src/WebJobs.Script/Host/MissingAzureFunctionsFolderValidator.cs
```diff
@@ -0,0 +1,39 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Microsoft.Azure.WebJobs.Script.Diagnostics.Extensions;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Script.Host
+{
+    internal class MissingAzureFunctionsFolderValidator : IFunctionAppValidator
```
kshyju:
```suggestion
    internal sealed class MissingAzureFunctionsFolderValidator : IFunctionAppValidator
```

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
kshyju: Could we add tests for this method?

manvkaur: There are tests in `FunctionAppValidationServiceTests.cs` that test this method

manvkaur: updated to `VerifyAndWarnIfBundleOutdated`

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
manvkaur: valuable suggestion, thanks Shyju

src/WebJobs.Script/Host/FunctionAppValidationService.cs
manvkaur: adopted the structured logging,
for change to warning, this is the original implementation - 

https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Host/FunctionAppValidationService.cs#L75

If the host team agrees it should be changed to warning, I can update it.

cc: @surgupta-msft , @RohitRanjanMS

src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
```diff
@@ -206,6 +206,11 @@ internal static class LoggerExtension
         private static readonly Action<ILogger, string, Exception> _publishingMetrics =
             LoggerMessage.Define<string>(LogLevel.Debug, new EventId(338, nameof(PublishingMetrics)), "{metrics}");
 
+        private static readonly Action<ILogger, string, Exception> _outdatedExtensionBundle =
+           LoggerMessage.Define<string>(LogLevel.Warning,
+           new EventId(342, nameof(OutdatedExtensionBundle)),
+           "{message}");
```
jviau: This is not using high perf logging correctly. The point is to have the template defined here and only fill in the non-constant parts at runtime. (Also`_publishingMetrics` is a bad example, it shouldn't be doing that either).

I see you have 2 different templates, what I suggest is:

1. Define two different templates via `LoggerMessage.Define`.
2. Define a _single_ extension method on `ILogger` for this. Move all of the templating logic into this message. But you will just be trying to get the individual values to pass to the messages defined in step 1.
3. Make sure to check `logger.IsEnabled(LogLevel.Warning);` before doing any calculations so you can no-op as an optimization.

A rough idea of the end `ILogger` extension method API they would call is:

``` CSharp
public static void OutdatedExtensionBundle(this ILogger logger, string currentVersion, string suggestedMin, string suggestedMax)
```

Feel free to play with those params however you need! The goal is to keep all the formatting logic in that method and not needing to add custom formatting, especially in a separate API

manvkaur: Thanks Jacob for this feedback, I have updated the PR with the recommended changes, please review it again.

src/WebJobs.Script/Host/FunctionAppValidationService.cs
surgupta-msft: For my case, the impact of this log was very minimal, so I kept it as a trace log. If we believe that now this log will add more value, then I have no objection to making it as a warning log.

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -172,6 +172,11 @@ public static IHostBuilder AddScriptHostCore(this IHostBuilder builder, ScriptAp
             {
                 if (!SystemEnvironment.Instance.IsPlaceholderModeEnabled())
                 {
+                    // Register all validators here
+                    services.AddSingleton<IFunctionAppValidator, ExtensionBundleManagerValidator>();
```
kshyju, line 176: Was talking to @surgupta-msft about this. I think we need both validations to happen in non-placeholder mode as well (think about dedicated SKU scenario). so, I think we should register the `FunctionAppValidationService` service and the validators always.

manvkaur: done.

src/WebJobs.Script/Host/FunctionAppValidationService.cs
manvkaur: for me too, it is not critical if this validator fails, warning log of validator failing can create panic if it shows up in customer logs

RohitRanjanMS: Let’s just keep this as info. Customers often worry when they see random warnings or errors and open support tickets. We can improve the error message and explain what it means and it's okay even if it fails.

src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
```diff
@@ -407,5 +417,25 @@ public static void IncorrectAzureFunctionsFolderPath(this ILogger logger, string
         {
             _incorrectAzureFunctionsFolderPath(logger, path, EnvironmentSettingNames.FunctionWorkerRuntime, null);
         }
+
+        public static void OutdatedExtensionBundle(this ILogger logger, string currentVersion, int suggestedMinVersion, int suggestedMaxVersion)
+        {
+            if (!logger.IsEnabled(LogLevel.Warning))
+            {
+                return;
+            }
```
RohitRanjanMS: This can be confusing. IsEnabled checks all the providers to see if at least one will log the message based on its level. We can skip this step and let each provider decide on its own.

src/WebJobs.Script/Host/ExtensionBundleManagerValidator.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            _extensionBundleManager = extensionBundleManager;
+        }
+
+        public void Validate(ScriptJobHostOptions options, IEnvironment environment, ILogger logger)
+        {
+            if (!logger.IsEnabled(LogLevel.Warning))
+            {
+                return;
+            }
```
RohitRanjanMS, line 24: This can be removed—it’s already covered.

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -375,5 +377,31 @@ public async Task<string> GetExtensionBundleBinPathAsync()
...
+            int dotIndex = _extensionBundleVersion.IndexOf('.');
+            if (dotIndex <= 0 || !int.TryParse(_extensionBundleVersion.AsSpan(0, dotIndex), out var majorVersion) || majorVersion == 0)
+            {
+                return string.Empty;
+            }
+
+            int latestMajorVersion = ScriptConstants.ExtensionBundleV4MajorVersion;
+
+            // Only show warnings for the default extension bundle when it's outdated
+            if (string.Equals(_options?.Id, ScriptConstants.DefaultExtensionBundleId, StringComparison.OrdinalIgnoreCase)
```
RohitRanjanMS: Can we move this check up? If it returns false, just return string.Empty and avoid the logic to extract bundle version. 

`string.Equals(_options?.Id, ScriptConstants.DefaultExtensionBundleId, StringComparison.OrdinalIgnoreCase)`

src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
manvkaur: This was introduced following recommendations from PR comment - https://github.com/Azure/azure-functions-host/pull/11075#discussion_r2157623309

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
manvkaur: done, thank for this recommendation

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -375,5 +377,33 @@ public async Task<string> GetExtensionBundleBinPathAsync()
...
         }
+
+        public string GetOutdatedBundleVersion()
+        {
+            // If the extension bundle version is not set or if the extension bundle is not the default one,
+            // return empty string
+            if (string.IsNullOrEmpty(_extensionBundleVersion) ||
+                !string.Equals(_options?.Id, ScriptConstants.DefaultExtensionBundleId, StringComparison.OrdinalIgnoreCase))
+            {
+                return string.Empty;
```
kshyju: nit: my personal preference would be to return `null` in these cases, instead of `string.empty`.

src/WebJobs.Script/Host/IFunctionAppValidator.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Script.Host
+{
+    /// <summary>
+    /// A validator interface for function app payload validation.
```
kshyju: The summary is incorrect (it appears to be specific to function app payload validation)

test/WebJobs.Script.Tests/Description/FunctionAppValidationServiceTests.cs
```diff
@@ -147,5 +161,53 @@ await TestHelpers.Await(() =>
...
+            // Arrange
+            _testLoggerProvider.ClearAllLogMessages();
+
+            var options = new ExtensionBundleOptions { Id = bundleId };
+            var env = new TestEnvironment();
+            var config = new FunctionsHostingConfigOptions();
+            var manager = new ExtensionBundleManager(options, env, _loggerFactory, config);
+
+            // Set the private _extensionBundleVersion field using reflection
+            typeof(ExtensionBundleManager)
```
kshyju, line 184: You may also consider mocking the http client. Ex:  https://github.com/Azure/azure-functions-host/blob/649c38edd0144439feb73091e40e1e67741723d6/test/WebJobs.Script.Tests/ExtensionBundle/ExtensionBundleManagerTests.cs#L338

test/WebJobs.Script.Tests/Description/FunctionAppValidationServiceTests.cs
```diff
@@ -147,5 +161,53 @@ await TestHelpers.Await(() =>
...
+        [InlineData("Microsoft.Azure.Functions.ExtensionBundle", "2.25.0", true)]
+        [InlineData("Microsoft.Azure.Functions.ExtensionBundle", "4.22.0", false)]
+        [InlineData("Microsoft.Azure.Functions.ExtensionBundle.Preview", "4.29.0", false)]
+        [InlineData("Microsoft.Azure.Functions.ExtensionBundle.Preview", "3.2.0", false)]
+        [InlineData(null, null, false)]
+        [InlineData("", "", false)]
+        public void GetOutdatedBundleWarningMessage_LogsWarning(string bundleId, string bundleVersion, bool shouldLogEvent)
+        {
+            // Arrange
+            _testLoggerProvider.ClearAllLogMessages();
```
kshyju, line 176: Why do we need this? Won't each test run gets its own instances. rite? No shared data state between tests.

kshyju (approved):

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
manvkaur: done

src/WebJobs.Script/Host/IFunctionAppValidator.cs
manvkaur: Good catch, updated now.

test/WebJobs.Script.Tests/Description/FunctionAppValidationServiceTests.cs
manvkaur: It's a best practice to have each test start with a clean state, especially when dealing with logs that are used for assertions, to ensure test reliability and prevent hard-to-debug test failures.

All the existing test cases are doing the same. ClearAllLogMessages() has 208 references in Test projects

src/WebJobs.Script/Host/ExtensionBundleManagerValidator.cs
manvkaur: removed at one place and changed category to LogCategoryHostGeneral

src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
manvkaur: removed at one place and changed category to LogCategoryHostGeneral

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
manvkaur: as tested, this logic gets executed during specialisation

src/WebJobs.Script/Host/FunctionAppValidationService.cs
```diff
@@ -20,17 +17,20 @@ namespace Microsoft.Azure.WebJobs.Script.Host
...
-            ILogger<FunctionAppValidationService> logger,
+            ILoggerFactory loggerFactory,
             IOptions<ScriptJobHostOptions> scriptOptions,
-            IEnvironment environment)
+            IEnvironment environment,
+            IEnumerable<IFunctionAppValidator> validators)
         {
             _scriptOptions = scriptOptions ?? throw new ArgumentNullException(nameof(scriptOptions));
-            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            _logger = loggerFactory.CreateLogger(ScriptConstants.LogCategoryHostGeneral) ?? throw new ArgumentNullException(nameof(loggerFactory));
```
manvkaur: @surgupta-msft , changing the category of logger to ScriptConstants.LogCategoryHostGeneral as per @RohitRanjanMS 's recommendation.
Let me know if you have any concerns.

surgupta-msft: No concerns

RohitRanjanMS (approved):

Frulfump:
This is nice when is it (roughly) expected to be deployed?

PR #11090, opened 2025-05-27
Avoiding exporter set-up in the placeholder mode.

### LINKED ISSUE #11073: Increased 429s in 4.1038+
> @fabiocav 
> We got host version 4.1038.300 in our production on April 11. 
> Since then a lot of our functions are on that version.
> But since that day we have had many functions that refuse to start and the response code is 429. 
> In total 7 200 requests coming into APIM just fail with 429 when the APIM makes the BACKEND call.
> So the function does not start.
> And many of the functions do not have a high load, and are on a consumption plan with just 11 apps on the same plan.
> When I inspect the "Diagnose and solve problems/Availability and Performance/HTTP 4xx Errors" i see the 429:s and they are classified as FrontEnd. The policy in the APIM has a retry policy element but that does not kick in, the request just dies.
> 
> Any clue on what might be ongoing? 

 _Originally posted by @peterno in [#2610](https://github.com/Azure/azure-functions-dotnet-worker/issues/2610#issuecomment-2802182948)_

RohitRanjanMS:
OpenTelemetry and AzureMonitor exporters are enabled in the Placeholder mode to optimize for cold start. If OpenTelemetry exporter isn't configured with an endpoint, it defaults to localhost:4317, leading to errors since no valid endpoint is available. Similarly, AzureMonitor fails due to an invalid IKey, generating status code 400 errors.

`StatusCode="Unavailable", Detail="Error connecting to subchannel."`
`Status Code: 400. Code: InvalidInput. Message: . Exception: iKey is garbage or not in a valid GUID format.`

These exceptions are frequently logged while apps run in Placeholder mode but stop after specialization. The sandbox issues numerous warnings, and customers may later hit connection limits, even though no signs indicate reaching the threshold. When this happens, throttling occurs with 429 responses.

The [sandbox ](https://github.com/projectkudu/kudu/wiki/Azure-Web-App-sandbox#local-address-requests)blocks connection attempts to local addresses (e.g., localhost, 127.0.0.1) and the machine's own IP, logging a warning when this occurs.

resolves #11073

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script/Diagnostics/OpenTelemetry/OpenTelemetryConfigurationExtensions.cs
+++ b/src/WebJobs.Script/Diagnostics/OpenTelemetry/OpenTelemetryConfigurationExtensions.cs
@@ -28,9 +28,8 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics.OpenTelemetry
     {
         internal static void ConfigureOpenTelemetry(this ILoggingBuilder loggingBuilder, HostBuilderContext context, TelemetryMode telemetryMode)
         {
-            var (otlpEndpoint, azMonConnectionString) = (
-                GetConfigurationValue(EnvironmentSettingNames.OtlpEndpoint, context.Configuration),
-                GetConfigurationValue(EnvironmentSettingNames.AppInsightsConnectionString, context.Configuration));
+            var otlpEndpoint = GetConfigurationValue(EnvironmentSettingNames.OtlpEndpoint, context.Configuration);
+            var azMonConnectionString = GetConfigurationValue(EnvironmentSettingNames.AppInsightsConnectionString, context.Configuration);
 
             bool enableOtlp = !string.IsNullOrWhiteSpace(otlpEndpoint);
             bool enableAzureMonitor = !string.IsNullOrWhiteSpace(azMonConnectionString);
@@ -43,8 +42,9 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics.OpenTelemetry
                 return;
             }
 
-            loggingBuilder
-                .ConfigureLogging(enableOtlp, enableAzureMonitor, azMonConnectionString, credential, telemetryMode).Services
+            loggingBuilder.ConfigureLogging(enableOtlp, enableAzureMonitor, azMonConnectionString, credential, telemetryMode);
+
+            loggingBuilder.Services
                 .AddOpenTelemetry()
                 .ConfigureResource(r => ConfigureResource(r))
                 .ConfigureMetrics(enableOtlp, enableAzureMonitor, azMonConnectionString, credential, telemetryMode)
--- a/src/WebJobs.Script/Diagnostics/OpenTelemetry/OpenTelemetryConfigurationExtensions.cs
+++ b/src/WebJobs.Script/Diagnostics/OpenTelemetry/OpenTelemetryConfigurationExtensions.cs
@@ -28,42 +28,34 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics.OpenTelemetry
     {
         internal static void ConfigureOpenTelemetry(this ILoggingBuilder loggingBuilder, HostBuilderContext context, TelemetryMode telemetryMode)
         {
-            var connectionString = GetConfigurationValue(EnvironmentSettingNames.AppInsightsConnectionString, context.Configuration);
-            var (azMonConnectionString, credential, enableOtlp, enableAzureMonitor) = telemetryMode switch
-            {
-                // Initializing OTel services during placeholder mode as well to avoid the cost of JITting these objects during specialization.
-                // Azure Monitor Exporter requires a connection string to be initialized. Use placeholder connection string if in placeholder mode.
-                TelemetryMode.Placeholder => (
-                    "InstrumentationKey=00000000-0000-0000-0000-000000000000;",
-                    null,
-                    true,
-                    true),
-                _ => (
-                    connectionString,
-                    GetTokenCredential(context.Configuration),
-                    !string.IsNullOrEmpty(GetConfigurationValue(EnvironmentSettingNames.OtlpEndpoint, context.Configuration)),
-                    !string.IsNullOrEmpty(connectionString))
-            };
+            var (otlpEndpoint, azMonConnectionString) = (
+                GetConfigurationValue(EnvironmentSettingNames.OtlpEndpoint, context.Configuration),
+                GetConfigurationValue(EnvironmentSettingNames.AppInsightsConnectionString, context.Configuration));
 
-            // If neither OTLP nor Azure Monitor is enabled, don't configure OpenTelemetry.
-            if (!enableOtlp && !enableAzureMonitor)
+            bool enableOtlp = !string.IsNullOrWhiteSpace(otlpEndpoint);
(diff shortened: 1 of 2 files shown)


jviau (approved):
Is it possible to add unit tests to verify we don't add exporters in placeholder mode?

src/WebJobs.Script/Diagnostics/OpenTelemetry/OpenTelemetryConfigurationExtensions.cs
```diff
@@ -28,42 +28,34 @@ internal static class OpenTelemetryConfigurationExtensions
...
-                _ => (
-                    connectionString,
-                    GetTokenCredential(context.Configuration),
-                    !string.IsNullOrEmpty(GetConfigurationValue(EnvironmentSettingNames.OtlpEndpoint, context.Configuration)),
-                    !string.IsNullOrEmpty(connectionString))
-            };
-
-            // If neither OTLP nor Azure Monitor is enabled, don't configure OpenTelemetry.
-            if (!enableOtlp && !enableAzureMonitor)
+            var (otlpEndpoint, azMonConnectionString) = (
```
fabiocav: nit: I think this is actually hurting readability as opposed to improving it. Can you split this up, please?

fabiocav: Not part of the changes, but it would be good to follow up on the need to expose `GetConfigurationValue` and have the fallback to environment variables handled in this class. Ideally, that logic would be removed from here.

src/WebJobs.Script/Diagnostics/OpenTelemetry/OpenTelemetryConfigurationExtensions.cs
```diff
@@ -28,42 +28,34 @@ internal static class OpenTelemetryConfigurationExtensions
...
+
+            // If placeholder mode is disabled and both OTLP and Azure Monitor are not enabled, avoid configuring OpenTelemetry.
+            if (!enableOtlp && !enableAzureMonitor && telemetryMode != TelemetryMode.Placeholder)
             {
                 return;
             }
 
             loggingBuilder
-                .ConfigureLogging(enableOtlp, enableAzureMonitor, azMonConnectionString, credential).Services
+                .ConfigureLogging(enableOtlp, enableAzureMonitor, azMonConnectionString, credential, telemetryMode).Services
```
fabiocav: Using the `Services` property from the method return here hurts readability a bit.

Can you please switch to:

``` csharp
loggingBuilder.ConfigureLogging(enableOtlp, enableAzureMonitor, azMonConnectionString, credential, telemetryMode);

loggingBuilder.Services
    .AddOpenTelemetry()
    .ConfigureResource(r => ConfigureResource(r))
    .ConfigureMetrics(enableOtlp, enableAzureMonitor, azMonConnectionString, credential, telemetryMode)
    .ConfigureTracing(enableOtlp, enableAzureMonitor, azMonConnectionString, credential, telemetryMode)
    .ConfigureEventLogLevel(context.Configuration);
```

fabiocav (approved):
Approving, but @jviau had a recommendation of unit tests that would be great to follow up on.

src/WebJobs.Script/Diagnostics/OpenTelemetry/OpenTelemetryConfigurationExtensions.cs
```diff
@@ -28,42 +28,34 @@ internal static class OpenTelemetryConfigurationExtensions
...
+            var otlpEndpoint = GetConfigurationValue(EnvironmentSettingNames.OtlpEndpoint, context.Configuration);
+            var azMonConnectionString = GetConfigurationValue(EnvironmentSettingNames.AppInsightsConnectionString, context.Configuration);
+
+            bool enableOtlp = !string.IsNullOrWhiteSpace(otlpEndpoint);
+            bool enableAzureMonitor = !string.IsNullOrWhiteSpace(azMonConnectionString);
+
+            TokenCredential credential = enableAzureMonitor ? GetTokenCredential(context.Configuration) : null;
+
+            // If placeholder mode is disabled and both OTLP and Azure Monitor are not enabled, avoid configuring OpenTelemetry.
+            if (!enableOtlp && !enableAzureMonitor && telemetryMode != TelemetryMode.Placeholder)
```
kshyju, line 40: Do we need the `telemetryMode` value check? If the endpoints are not configured, can't we just return/exit from here? (Also if the telemetryMode check is confusing to me. Should that be `==` ? (Because in placeholder mode, the value of `telemetryMode` is  `Placeholder`. Isn't it?).

src/WebJobs.Script/Diagnostics/OpenTelemetry/OpenTelemetryConfigurationExtensions.cs
```diff
@@ -28,42 +28,34 @@ internal static class OpenTelemetryConfigurationExtensions
...
-
-            // If neither OTLP nor Azure Monitor is enabled, don't configure OpenTelemetry.
-            if (!enableOtlp && !enableAzureMonitor)
+            var otlpEndpoint = GetConfigurationValue(EnvironmentSettingNames.OtlpEndpoint, context.Configuration);
+            var azMonConnectionString = GetConfigurationValue(EnvironmentSettingNames.AppInsightsConnectionString, context.Configuration);
+
+            bool enableOtlp = !string.IsNullOrWhiteSpace(otlpEndpoint);
+            bool enableAzureMonitor = !string.IsNullOrWhiteSpace(azMonConnectionString);
+
+            TokenCredential credential = enableAzureMonitor ? GetTokenCredential(context.Configuration) : null;
```
kshyju, line 37: Can this be moved closer to usage? In the use case where we are exiting, there is no need to execute this line, I guess

src/WebJobs.Script/Diagnostics/OpenTelemetry/OpenTelemetryConfigurationExtensions.cs
fabiocav: That's a great catch here, @kshyju . I do agree this check adds a bit of confusion here. 

Given the fact the methods below seem to evaluate `enableOtlp` and `enableAzureMonitor` independently, I believe @RohitRanjanMS's intention is to have the logic flow in placeholder mode (not short-circuit), so OTel is still configured for warmups, but without exporters (@RohitRanjanMS , can you confirm)? If this is accurate, we do need to make sure we have a better comment  is in place to explain the reason/intent, we should perhaps consider refactoring this logic to avoid confusion as well.

RohitRanjanMS: The intent is to continue OTel in placeholder mode while configuring everything except the exporters. I'm upgrading the OTel/AzMon package, and in the newer library, one line of code sets up the trace, log, and metrics exporters. The code will be much simpler after the version update. There are customers asking for it, so I am going to do it soon.

src/WebJobs.Script/Diagnostics/OpenTelemetry/OpenTelemetryConfigurationExtensions.cs
RohitRanjanMS: I will refactor this when I update the nuget package.

PR #11101, opened 2025-05-31
Improved metadata binding parsing and validation.

kshyju:
Improving how metadata binding is being read and used. Handles both worker indexing and host indexing cases.

### Pull request checklist

    * [x] Otherwise: Backport tracked by issue/PR #issue_or_pr - will follow


src/WebJobs.Script.WebHost/Extensions/FunctionMetadataExtensions.cs
```diff
@@ -175,7 +176,7 @@ private static async Task<JObject> GetFunctionConfig(FunctionMetadata metadata,
 
         private static async Task<JObject> GetFunctionConfigFromFile(string path)
         {
-            return JObject.Parse(await FileUtility.ReadAsync(path));
+            return JObject.Parse(Sanitizer.Sanitize(await FileUtility.ReadAsync(path)));
```
kshyju: synctrigger endpoint and /functions/{functionname} endpoint code reads the file if it is present. (Valid for custom handlers /non worker indexing use case).

src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
```diff
@@ -257,7 +258,7 @@ internal FunctionMetadata ValidateBindings(IEnumerable<string> rawBindings, Func
 
             foreach (string binding in rawBindings)
             {
-                var deserializedObj = JsonConvert.DeserializeObject<JObject>(binding, _dateTimeSerializerSettings);
+                var deserializedObj = JsonConvert.DeserializeObject<JObject>(Sanitizer.Sanitize(binding), _dateTimeSerializerSettings);
```
kshyju: This method gets executed in worker indexing use case.

test/WebJobs.Script.Tests/SanitizerTests.cs
```diff
@@ -41,6 +41,9 @@ public class SanitizerTests
         [InlineData("test,aaa://aaa:aaaaaa1111aa@aaa.aaa.io:1111,test", "test,[Hidden Credential],test")]
         [InlineData(@"some text abc://abc:aaaaaa1111aa@aaa.abc.io:1111 some text abc://abc:aaaaaa1111aa@aaa.abc.io:1111 text", @"some text [Hidden Credential] some text [Hidden Credential] text")]
         [InlineData(@"some text abc://abc:aaaaaa1111aa@aaa.abc.io:1111 some text AccountKey=heyyyyyyy text", @"some text [Hidden Credential] some text [Hidden Credential]")]
+        [InlineData("""{"queueName":"my-q-items","connection":"MyConnection","type":"queueTrigger","name":"qTrigger1","direction":"in"}""", "{\"queueName\":\"my-q-items\",\"connection\":\"MyConnection\",\"type\":\"queueTrigger\",\"name\":\"qTrigger1\",\"direction\":\"in\"}")]
+        [InlineData("""{"queueName":"my-q-items","connection":"DefaultEndpointsProtocol=https;AccountName=a;AccountKey=b/c==;EndpointSuffix=core.windows.net","type":"queueTrigger","name":"queueTrigger1","direction":"in"}""", "{\"queueName\":\"my-q-items\",\"connection\":\"[Hidden Credential]\",\"type\":\"queueTrigger\",\"name\":\"queueTrigger1\",\"direction\":\"in\"}")]
+
```
surgupta-msft: Could this be a breaking change in some scenarios? Let's say, queue name is `"key=myqueue"` then we will replace it with `"[Hidden Credential]"`. Will the function stop executing since we changed the queue name or is it only the Sync trigger response that is being sanitized (and it won't break any other component that depends on Sync trigger response)?

RohitRanjanMS (approved):

test/WebJobs.Script.Tests/SanitizerTests.cs
kshyju: The functions are already in a broken state (not executing anything). The extension listener code throws when it is not able to resolve a valid app setting value from the value of these attributes (example code for service bus extension [here](https://github.com/Azure/azure-sdk-for-net/blob/fc7d07aa1e638f4dbb50b0001fd08e31ee59559b/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Config/ServiceBusClientFactory.cs#L59-L74)). So this is not breaking anything.

surgupta-msft: We are sanitizing the entire binding object for all scenarios and not just the exception cases. If any extension introduces (or already has) a binding property with name `"password="` or `"key="` (may not be good examples), they will be modified to `[Hidden Credential]` and the app can break.

In terms of binding values, I checked that the queue name and connection setting name cannot have `"="` (but other bindings value may have `"="`), but they can have other characters like ":", so may be, we need to ensure that the Sanitizer class should never have tokens such as `"key:"`, otherwise we will modify that value to `[Hidden Credential]`. A unit test should work for this scenario.

kshyju: I am not sure I follow the last part of the comment. But for inputs like `"one:two;three"`, Sanitizer returns the same value output.

| Input              | Output             |
|-------------------|--------------------|
| one:two;three     | one:two;three      |
| one:a;two=three   | one:a;two=three    |
| foo=bar           | foo=bar            |

If you meant something else, can you share an example input?

surgupta-msft: Today, `key=myvalue` will return `[HiddenCredential]`.  If in future, someone adds a token `"key:"` in Sanitizer, then
`key:myvalue` will also return `[HiddenCredential]`. This `key:myvalue` is a valid connection setting name. Let's also discuss this offline to make sure we don't have any confusion here.

jviau: Right because `:` is the connection separator, so it won't be unexpected for connection names to include it. Now, if customer's have a section literally called "key" or "secret", or something else covered by our sanitizer that is another question.

kshyju: Maybe we should reconsider the approach taken here and sanitize only the binding values when we serve the response from the /functions/ endpoints.  Let's follow up offline.

kshyju:
Had offline sync with @surgupta-msft and agreed to mask only the "connection" binding value.

src/WebJobs.Script.WebHost/Extensions/FunctionMetadataExtensions.cs
```diff
@@ -175,7 +176,21 @@ private static async Task<JObject> GetFunctionConfig(FunctionMetadata metadata,
 
         private static async Task<JObject> GetFunctionConfigFromFile(string path)
         {
-            return JObject.Parse(await FileUtility.ReadAsync(path));
+            var fileContent = await FileUtility.ReadAsync(path);
+            var jObject = JObject.Parse(fileContent);
+
+            if (jObject.TryGetValue("bindings", StringComparison.OrdinalIgnoreCase, out JToken bindingsToken) && bindingsToken is JArray bindings)
+            {
+                var bindingObjects = bindings.OfType<JObject>().ToList();
```
surgupta-msft: Is it possible to avoid converting JArray to List here? In HostFunctionMetadataProvider, we are able to do similar thing with JArray.

src/WebJobs.Script/Host/MetadataJsonHelper.cs
```diff
@@ -0,0 +1,76 @@
...
+        public static JObject CreateJObjectWithSanitizedPropertyValue(JObject jsonObject, IEnumerable<string> propertyNames)
+        {
+            ArgumentNullException.ThrowIfNull(jsonObject, nameof(jsonObject));
+            ArgumentNullException.ThrowIfNull(propertyNames, nameof(propertyNames));
+
+            foreach (var prop in jsonObject.Properties())
+            {
+                if (propertyNames.Contains(prop.Name, StringComparer.OrdinalIgnoreCase))
+                {
+                    jsonObject[prop.Name] = Sanitizer.Sanitize(prop.Value.ToString());
```
surgupta-msft: Connection value could be null. So, I believe we need a null check for prop.Value.

src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
```diff
@@ -257,8 +258,8 @@ internal FunctionMetadata ValidateBindings(IEnumerable<string> rawBindings, Func
 
             foreach (string binding in rawBindings)
             {
-                var deserializedObj = JsonConvert.DeserializeObject<JObject>(binding, _dateTimeSerializerSettings);
```
surgupta-msft, line 292: We need to still leverage `_dateTimeSerializerSettings`. If we don't use it, it will break some datetime stuff. 

For example, current serializer setting says to not parse the datetime. If we don't use it, it will start parsing it. I added an example in the test class below. I worked on similar issue before - https://github.com/Azure/azure-functions-dotnet-worker/issues/2442

test/WebJobs.Script.Tests/HostFunctionMetadataProviderTests.cs
```diff
@@ -350,5 +350,59 @@ public void ParseFunctionMetadata_ResolvesCorrectDotNetLanguage(string scriptFil
...
+                                      ""direction"": ""in"",
+                                      ""queueName"": ""test-input-node"",
+                                      ""connection"": ""DefaultEndpointsProtocol=https;AccountName=a;AccountKey=b/c==;EndpointSuffix=core.windows.net""
+                                    },
+                                    {
+                                      ""name"": ""$return"",
+                                      ""type"": ""queue"",
+                                      ""direction"": ""out"",
+                                      ""queueName"": ""test-output-node"",
+                                      ""connection"": ""MyConnection""
```
surgupta-msft, line 372: Can we check for connection null and empty too?

test/WebJobs.Script.Tests/WorkerFunctionMetadataProviderTests.cs
```diff
@@ -291,5 +291,32 @@ static void AssertFunction(FunctionMetadata function)
...
             AssertFunction(function2);
         }
+
+        [Fact]
+        public void ValidateBindings_MasksSensitiveDataInBindings()
+        {
+            var functionMetadata = new FunctionMetadata();
+            List<string> rawBindings =
+            [
+                """{"type": "queueTrigger","name": "myQueueItem","direction": "in","queueName": "test-input-node","connection": "DefaultEndpointsProtocol=https;AccountName=a;AccountKey=b/c==;EndpointSuffix=core.windows.net"}""",
```
surgupta-msft, line 301: Add a test with queue name (or any other field) as `"2024-04-29T11:35:00+08:00"`. You will see that the datetime is parsed to some other value. This won't happen if we use 

`new JsonSerializerSettings { DateParseHandling = DateParseHandling.None }`

src/WebJobs.Script.WebHost/Extensions/FunctionMetadataExtensions.cs
kshyju: Good call. The List is not needed. Updated to remove the allocation.  We still need to check the instance is `JObject` type before continuing to use it.

src/WebJobs.Script/Host/MetadataJsonHelper.cs
kshyju: It will not be `null`, It will be of type `JToken`
https://www.newtonsoft.com/json/help/html/P_Newtonsoft_Json_Linq_JProperty_Value.htm

I added a new test to ensure this does not throw.

test/WebJobs.Script.Tests/HostFunctionMetadataProviderTests.cs
kshyju: The new test verifies both cases.

src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
kshyju: BindingMetadata does not have DateTime attributes today.  https://github.com/Azure/azure-functions-host/blob/c7192ec2ea087f30b617527b173859ab3a4eeb80/src/WebJobs.Script.Abstractions/Description/Binding/BindingMetadata.cs#L16

If we ever introduce a `DateTime` property in binding data, we should clearly define the allowed formats, how it affects other layers of the system. 

Since this helper method is not specific to binding data, I added a `DateParseHandling` parameter with a default value of `DateParseHandling.None`. This allows the caller to override the behavior if needed and makes the date parsing behavior explicit. We are using `JObject.Load`, which is more performant than the `DeserializeObject` method.

This new method is called in only one place where we pass the bindingMetdata JSON. Do you see a use case where this will still break any existing behavior?

test/WebJobs.Script.Tests/WorkerFunctionMetadataProviderTests.cs
kshyju: see CreateJObjectWithSanitizedPropertyValue_StringInput_DateTimeWithTimezoneOffset_RemainsUnchanged test

src/WebJobs.Script.WebHost/Extensions/FunctionMetadataExtensions.cs
```diff
@@ -175,7 +176,21 @@ private static async Task<JObject> GetFunctionConfig(FunctionMetadata metadata,
...
+            var fileContent = await FileUtility.ReadAsync(path);
+            var jObject = JObject.Parse(fileContent);
+
+            if (jObject.TryGetValue("bindings", StringComparison.OrdinalIgnoreCase, out JToken bindingsToken) && bindingsToken is JArray bindingsArray)
+            {
+                for (int i = 0; i < bindingsArray.Count; i++)
+                {
+                    if (bindingsArray[i] is JObject binding)
+                    {
+                        bindingsArray[i] = MetadataJsonHelper.CreateJObjectWithSanitizedPropertyValue(binding, ScriptConstants.SensitiveMetadataBindingPropertyNames);
```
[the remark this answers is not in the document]
kshyju: InternalsVisibleTo is enabled (our test project is calling this method).

src/WebJobs.Script/Host/MetadataJsonHelper.cs
```diff
@@ -0,0 +1,98 @@
...
+            ArgumentNullException.ThrowIfNull(propertyNames, nameof(propertyNames));
+
+            if (propertyNames.Count == 0)
+            {
+                return jsonObject;
+            }
+
+            foreach (var prop in jsonObject.Properties())
+            {
+                if (propertyNames.Contains(prop.Name, StringComparer.OrdinalIgnoreCase))
```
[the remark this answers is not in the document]
kshyju: This method takes an ImmutableHashSet as a parameter. Since the caller might not use a case-insensitive comparer when creating the set, we cannot assume that the input is case-insensitive. Therefore, this change is needed.

src/WebJobs.Script/Host/MetadataJsonHelper.cs
```diff
@@ -0,0 +1,98 @@
...
+                    {
+                        continue;
+                    }
+
+                    var valueToSanitize = prop.Value.Type == JTokenType.String ? (string)prop.Value : prop.Value.ToString();
+                    jsonObject[prop.Name] = Sanitizer.Sanitize(valueToSanitize);
+                }
+            }
+
+            return jsonObject;
```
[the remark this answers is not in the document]
kshyju: Not cloning due to performance concerns. Instead, the method name and summary were updated to be reflect that the input param will be mutated.

src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
```diff
@@ -289,8 +290,8 @@ internal FunctionMetadata ValidateBindings(IEnumerable<string> rawBindings, Func
...
-                var functionBinding = BindingMetadata.Create(deserializedObj);
+                var sanitizedBinding = MetadataJsonHelper.CreateJObjectWithSanitizedPropertyValue(binding, ScriptConstants.SensitiveMetadataBindingPropertyNames);
```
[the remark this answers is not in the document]
kshyju: This new method (`CreateJObjectWithSanitizedPropertyValue`) takes a parameter which specifies how date parsing should be handled. I passed explicit `DateParseHandling.None` as the value in this case. This will produce same output as the previous version.

src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
surgupta-msft: The changes look good.

src/WebJobs.Script/Host/MetadataJsonHelper.cs
```diff
@@ -0,0 +1,99 @@
...
+        /// <returns>
+        /// The modified <see cref="JObject"/> with the specified properties' values sanitized if found.
+        /// </returns>
+        /// <exception cref="ArgumentNullException">
+        /// Thrown if <paramref name="jsonObject"/> or <paramref name="propertyNames"/> is <c>null</c>.
+        /// </exception>
+        public static JObject SanitizeProperties(JObject jsonObject, ImmutableHashSet<string> propertyNames)
+        {
+            ArgumentNullException.ThrowIfNull(jsonObject, nameof(jsonObject));
+            ArgumentNullException.ThrowIfNull(propertyNames, nameof(propertyNames));
```
surgupta-msft, line 33: The changes look good. The only thing I did not understand is why we are throwing these exceptions. Can customers see these exceptions? Should we consider just returning from the methods if the parameters are null? Same comment for the other method in this class - `CreateJObjectWithSanitizedPropertyValue`

surgupta-msft (approved):

src/WebJobs.Script/Host/MetadataJsonHelper.cs
kshyju: Throwing an argument null exception when required parameters are null is considered a good practice in .NET. Many .NET APIs does that. It clearly signals to the caller that they have made an error and makes debugging easier as the exception points directly to the misuse. Returning an empty object instead can cause confusion when diagnosing issues.

This newly added method is called from 3 places in code.

1) `FunctionMetadataExtensions.GetFunctionConfigFromFile`
We are using pattern matching there. So, binding will not be `null` when calling `SanitizeProperties` method.

2) `HostFunctionMetadataProvider.ParseFunctionMetadata`

The previous version of code in this method was 
```csharp
BindingMetadata bindingMetadata = BindingMetadata.Create(binding);
```
and the new version is
```csharp
var sanitizedJObject = MetadataJsonHelper.SanitizeProperties(binding, ScriptConstants.SensitiveMetadataBindingPropertyNames);
BindingMetadata bindingMetadata = BindingMetadata.Create(sanitizedJObject);
```

The previous version called `BindingMetadata.Create` which would throw if `binding` is null. So that behavior is same as the new version (except, now the first method will throw).

https://github.com/Azure/azure-functions-host/blob/c7192ec2ea087f30b617527b173859ab3a4eeb80/src/WebJobs.Script.Abstractions/Description/Binding/BindingMetadata.cs#L106-L109

3) `MetadataJsonHelper.CreateJObjectWithSanitizedPropertyValue`
This is another newly added method. This method already validates the input used to build the JObject. JObject.Load throws if the json string passed in is not valid json.  This new method is called in only one place.

`WorkerFunctionMetadataProvider.ValidateBindings`


The previous version of code in `ValidateBindings` was 
```csharp
var deserializedObj = JsonConvert.DeserializeObject<JObject>(binding, _dateTimeSerializerSettings);
var functionBinding = BindingMetadata.Create(deserializedObj);
```
and the new version is
```csharp
var sanitizedBinding = MetadataJsonHelper.CreateJObjectWithSanitizedPropertyValue(binding, ScriptConstants.SensitiveMetadataBindingPropertyNames, DateParseHandling.None);
var functionBinding = BindingMetadata.Create(sanitizedBinding);
```
Here also behavior is same as 2. `JsonConvert.DeserializeObject` will throw if `binding` is `null`. `JsonConvert.DeserializeObject` will return `null` if `binding` is empty string. But the next line (`BindingMetadata.Create`) will throw when the parameter value is null.

So the existing callers should not see any new behavior with this new APIs.

src/WebJobs.Script/Host/MetadataJsonHelper.cs
```diff
@@ -0,0 +1,99 @@
...
+
+            if (propertyNames.Count == 0)
+            {
+                return jsonObject;
+            }
+
+            foreach (var prop in jsonObject.Properties())
+            {
+                if (propertyNames.Contains(prop.Name, StringComparer.OrdinalIgnoreCase))
```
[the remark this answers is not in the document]
kshyju: This method takes `propertyNames` as a parameter and we cannot assume that he caller is passing the hashset which was created with StringComparer.OrdinalIgnoreCase. So this suggestion is not valid.

RohitRanjanMS (approved):

surgupta-msft (approved):

PR #11116, opened 2025-06-13
RpcException handling.

RohitRanjanMS:

resolves #issue_for_this_pr

### Pull request checklist

* [] My changes **should not** be added to the release notes for the next release
    * [x ] Otherwise: I've added my notes to `release_notes.md`

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/Diagnostics/SystemLogger.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/SystemLogger.cs
@@ -7,6 +7,7 @@ using Microsoft.Azure.WebJobs.Host;
 using Microsoft.Azure.WebJobs.Host.Executors.Internal;
 using Microsoft.Azure.WebJobs.Host.Indexers;
 using Microsoft.Azure.WebJobs.Logging;
+using Microsoft.Azure.WebJobs.Script.Config;
 using Microsoft.Azure.WebJobs.Script.Configuration;
 using Microsoft.Azure.WebJobs.Script.Eventing;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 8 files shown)


Frulfump:
The PR description contains nothing only template text

src/WebJobs.Script.WebHost/Diagnostics/SystemLogger.cs
```diff
@@ -173,12 +189,29 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
                     functionName = string.IsNullOrEmpty(fex.MethodName) ? string.Empty : fex.MethodName.Replace("Host.Functions.", string.Empty);
                 }
 
-                (innerExceptionType, innerExceptionMessage, details) = exception.GetExceptionDetails();
-                formattedMessage = Sanitizer.Sanitize(formattedMessage);
-                innerExceptionMessage = innerExceptionMessage ?? string.Empty;
+                var exceptionDetails = exception.GetExceptionDetails();
+
+                if (_logRpcExceptionDetails || !exceptionDetails.ExceptionType.Equals(_rpcExceptionName, StringComparison.Ordinal))
```
jviau, line 194: Can we just check the exception type? The type should be accessible to this class.

```suggestion
                if (_logRpcExceptionDetails || exception is not RpcException)
```

jviau (changes requested):

src/WebJobs.Script.WebHost/Diagnostics/SystemLogger.cs
RohitRanjanMS: We determine the type of the innermost exception and use that for comparison. If the innerException is null, we default to `exception`.

jviau: Can we do the same logic here then, but comparing by type and not string? Add an exception extension method if needed:

``` CSharp
public static bool IsCausedBy<T>(this Exception ex)
{
    while (ex?.InnerException is not null)
    {
      ex = ex.InnerException;
    }
    
    return ex is T;
}
```

fabiocav (changes requested):
Blocking this until we have made teams with potential dependency on this data aware of the changes.

pagariyaalok (commented):
Pending discussion related to exception-based Anomaly detection.

PR #11138, opened 2025-06-18
Use -or operator in New-PrivateSiteExtension.ps1

castrodd:

Our release pipeline for the Host started failing to create the PrivateSiteExtension. When we looked at the logs, we saw the following line:
```
Published private site extension to .zip
```
Normally, this would be something like:
```
Published private site extension to Functions.Private.4.1040.300.win-x86.zip
```
I could see that this log was coming from the last line of the script. This suggested that there was an issue with the `$OutputPath`.

When I ran the script locally, I could see that `$OutputPath` was never getting set. After some trial and error, I was able to pin it down to the line being modified in this PR. This condition was not resolving to true, even though both `$AppendOutputName` and `$OutputPath` are null at that point in the script and therefore `!$OutputPath` should be equal to true.

When I ran the script with the `-or` operator instead of `||`, the script worked as expected and the `$OutputPath` was set correctly.

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.SiteExtension/New-PrivateSiteExtension.ps1
+++ b/src/WebJobs.Script.SiteExtension/New-PrivateSiteExtension.ps1
@@ -65,7 +65,7 @@ if (-not (Join-Path $InputPath "extension.xml" | Test-Path))
     exit 1
 }
 
-if ($AppendOutputName || !$OutputPath)
+if ($AppendOutputName -or !$OutputPath)
 {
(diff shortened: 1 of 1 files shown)


jviau (approved):
Interesting - are you running this with an old powershell version?

castrodd:
> Interesting - are you running this with an old powershell version?

I would need to check what the pipeline is running, but locally I am on 7.5.1.

Francisco-Gamino (approved):

PR #11142, opened 2025-06-20
fixing startup deadlock

### LINKED ISSUE #10766: Error during host startup can cause a deadlock in the restart flow, leaving the host unhealthy until a manual restart
When a host startup fails (e.g., due to storage connectivity issues), sets the host to **Error** and initiates a new host startup. This new startup acquires `_hostStartSemaphore` and calls `BuildHost()`. During that process, `WorkerFunctionMetadataProvider.GetFunctionMetadataAsync()` detects no active worker channels and calls `RestartHostAsync()`. However, `RestartHostAsync()` attempts to cancel the same in-progress startup, and because there’s no `ThrowIfCancellationRequested`, `_hostStartSemaphore` is never released. The restart remains blocked, leaving the host in **Error** state until it is manually restarted.

### Repro steps

1. **First Host Start**  
   The host begins to initialize (loads metadata, starts worker channels, etc.).  

2. **Failure Connecting to Storage**  
   A transient error (e.g., DNS or storage connection issue) leads to an aborted startup and moves the host state to **Error**.  

3. **Host Startup Canceled**  
   Because of the error, the system transitions the existing startup to a canceled state (shutting down worker channels).  

4. **New Host Scheduled**  
   The system schedules a new host startup after a short delay.  

5. **Second Host Startup**  
   This new host startup acquires the `_hostStartSemaphore` and begins another `BuildHost()` process.  

6. **Metadata Provider Finds No Channels**  
   Inside `WorkerFunctionMetadataProvider.GetFunctionMetadataAsync()`, the code detects that no channels exist (they were previously shut down). https://github.com/Azure/azure-functions-host/blob/dae16f96500d825b16d9dd5d3887f83925d67b88/src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs#L81-L97 

7. **`RestartHostAsync()` Called**  
   Since no channels are active, `RestartHostAsync()` is invoked to re-initialize the host.  

8. **Cancellation Attempt**  
   `RestartHostAsync()` attempts to cancel the active startup operation by calling `Cancel()` on its `CancellationTokenSource`. However, because this cancellation call happens within the same call stack/async flow as the current `BuildHost()`, there is no `ThrowIfCancellationRequested` check or natural yield point to abort the build operation.  

https://github.com/Azure/azure-functions-host/blob/dae16f96500d825b16d9dd5d3887f83925d67b88/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs#L562-L577

9. **Semaphore Deadlock**  
   With the second startup still holding the `_hostStartSemaphore`—and never releasing it due to the ineffective cancellation—the new restart attempt blocks indefinitely when trying to reacquire that semaphore.  

At this point, the host remains in an **Error** state until it is manually restarted, since the restart logic is effectively deadlocked.

### Example Call Stack

```
Microsoft.Azure.WebJobs.Script.WebHost.WebJobsScriptHostService.RestartHostAsync
Microsoft.Azure.WebJobs.Script.WorkerFunctionMetadataProvider.GetFunctionMetadataAsync
Microsoft.Azure.WebJobs.Script.WebHost.FunctionMetadataProvider.GetFunctionMetadataAsync
Microsoft.Azure.WebJobs.Script.FunctionMetadataManager.LoadFunctionMetadata
Microsoft.Azure.WebJobs.Script.DependencyInjection.ScriptStartupTypeLocator.GetExtensionsStartupTypesAsync
Microsoft.Azure.WebJobs.Script.WebHost.DefaultScriptHostBuilder.BuildHost
Microsoft.Azure.WebJobs.Script.WebHost.WebJobsScriptHostService.BuildHost
Microsoft.Azure.WebJobs.Script.WebHost.WebJobsScriptHostService.StartHostAsync
```

brettsam:

resolves #10766

### Pull request checklist

### Additional information

The problem here (outlined in the issue) is that:
1. Even in transient error situations, we were setting the `State` to `Error`.
2. If we tried to recover, the `WorkerFunctionMetadataProvider` would look at that state, think that the host was in a stable state without any channels, and issue a restart here: https://github.com/Azure/azure-functions-host/blob/71738799181ebd3c8f7104ba2c881c618cd1e631/src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs#L99
3. That restart would deadlock waiting for the current startup to complete... but it was part of the current startup so it'd never complete.
4. Deadlock.

After looking through all the code, I've concluded that:
- ~~`Error` and `Offline` (and to some extent `Running`) are considered "terminal" events... that's the state things will remain in. We have code that immediately returns if it sees `Error`, etc.~~
- ~~However, during a transient error, this isn't the case. We are recovering and need to communicate that somehow.~~
- ~~In this specific case, it's untrue that the `ScriptHost` is in an `Error` state. We are retrying fresh, so the metadata manager needs to know that.~~
- ~~By resetting to `Default`, it communicates that we're still starting up.~~
- ~~Yes, this may go on forever as we'll back off but retry until the platform stops us. However, the places that are waiting on this also have timeouts applied, so they should be unaffected. And in fact, they should behave better under transient conditions.~~

My original attempt had added a `HandlingHostError` state that, while did communicate correctly, change the assumptions across other services. A handful of tests failed.

After seeing tests fail b/c they assumed `Error`, I've changed the approach to simply starting worker channels also while in the `Error` state. This seems fine the more I look at it -- we are performing a normal startup (or else these services wouldn't start), so we should behave as such. Every other service behaves normally here and we just happen to have this unique piece of code for an odd situation where we need to recover.

Other considerations:
- What we really should do is make sure that when `RestartAsync()` is called, the `CancellationToken` we call will flow all the way down, effectively cancelling itself. However, that would be a *large* refactoring of code that is already being completely rewritten. I consider this a strategic fix to stop the issues we're currently seeing and the next iteration of worker management will take care of this scenario in a cleaner way.

### DIFF
--- a/src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
+++ b/src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
@@ -83,9 +83,11 @@ namespace Microsoft.Azure.WebJobs.Script
                 // Start up GRPC channels if they are not already running.
                 if (channels?.Any() != true)
                 {
+                    // Note: Error can mean transient error, so we want to attempt a normal startup.
                     if (_scriptHostManager.State is ScriptHostState.Default
                         || _scriptHostManager.State is ScriptHostState.Starting
-                        || _scriptHostManager.State is ScriptHostState.Initialized)
+                        || _scriptHostManager.State is ScriptHostState.Initialized
+                        || _scriptHostManager.State is ScriptHostState.Error)
                     {
                         // We don't need to restart if the host hasn't even been created yet.
                         _logger.LogDebug("Host is starting up, initializing language worker channel");
--- a/src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
+++ b/src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
@@ -11,6 +11,8 @@ using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Script.Description;
 using Microsoft.Azure.WebJobs.Script.Diagnostics.Extensions;
 using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Newtonsoft.Json;
@@ -18,7 +20,7 @@ using Newtonsoft.Json.Linq;
 
 namespace Microsoft.Azure.WebJobs.Script
 {
-    internal class WorkerFunctionMetadataProvider : IWorkerFunctionMetadataProvider
+    internal class WorkerFunctionMetadataProvider : IWorkerFunctionMetadataProvider, IDisposable
     {
         private const string _metadataProviderName = "Worker";
         private readonly Dictionary<string, ICollection<string>> _functionErrors = new Dictionary<string, ICollection<string>>();
@@ -30,6 +32,7 @@ namespace Microsoft.Azure.WebJobs.Script
         private readonly JsonSerializerSettings _dateTimeSerializerSettings;
         private string _workerRuntime;
         private ImmutableArray<FunctionMetadata> _functions;
+        private IHost _currentJobHost = null;
 
         public WorkerFunctionMetadataProvider(
             IOptionsMonitor<ScriptApplicationHostOptions> scriptOptions,
@@ -45,6 +48,8 @@ namespace Microsoft.Azure.WebJobs.Script
             _scriptHostManager = scriptHostManager;
             _workerRuntime = _environment.GetEnvironmentVariable(EnvironmentSettingNames.FunctionWorkerRuntime);
             _dateTimeSerializerSettings = new JsonSerializerSettings { DateParseHandling = DateParseHandling.None };
+
+            _scriptHostManager.ActiveHostChanged += OnHostChanged;
         }
 
         public ImmutableDictionary<string, ImmutableArray<string>> FunctionErrors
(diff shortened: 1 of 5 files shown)


test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/WebHostStartupEndToEndTests.cs
```diff
@@ -0,0 +1,74 @@
...
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Options;
+using Xunit;
+
+namespace Microsoft.Azure.WebJobs.Script.Tests.Integration.WebHostEndToEnd;
+
+public class WebHostStartupEndToEndTests
+{
+    [Fact]
+    public async Task TransientError_DuringHostBuild_DoesNotDeadlock()
```
brettsam, line 18: Test failed previously... second host deadlocked and never recovered.

cjaliaga (approved):

RohitRanjanMS (approved):

src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
```diff
@@ -83,9 +83,11 @@ public async Task<FunctionMetadataResult> GetFunctionMetadataAsync(IEnumerable<R
                 // Start up GRPC channels if they are not already running.
                 if (channels?.Any() != true)
                 {
+                    // Note: Error can mean transient error, so we want to attempt a normal startup.
                     if (_scriptHostManager.State is ScriptHostState.Default
                         || _scriptHostManager.State is ScriptHostState.Starting
-                        || _scriptHostManager.State is ScriptHostState.Initialized)
+                        || _scriptHostManager.State is ScriptHostState.Initialized
+                        || _scriptHostManager.State is ScriptHostState.Error)
```
mathewc: Just looking at this code in isolation is confusing. We'll execute this branch when the host is in an error state, but below the code comment says "host hasn't even been created yet". If we're in an error state, won't we be initiating a restart elsewhere?

brettsam: It's 100% confusing. We're currently re-writing this layer b/c of how mangled this code has become over the years.  This is an attempted targeted fix to prevent deadlocked hosts on fairly normal transient errors (like a connection to storage on startup).

The biggest thing I feel is that the `Error` state doesn't give us enough information -- it can be `Error` while we're starting up or it can be `Error` as a "final" state. I tried to introduce a `HandlingStartupError` value but a bunch of tests failed -- and there's some assumptions elsewhere as to what `Error` means that were tough to understand. It was quickly becoming a very un-targeted fix (while we have a re-write in flight).

That being said -- I think I can make this read better than it does right now. Ultimately, we need to know which of these scenarios we're in:
- are we in the middle of starting up?
- or did we already complete startup and still have no channels?

Let me come back with a tweak...

brettsam:
@cjaliaga / @mathewc / @RohitRanjanMS -- this is ready for another look after a refactor to instead rely on the host lifetime to check whether it's completed startup or not

RohitRanjanMS (approved):

liliankasem (approved):

brettsam:
@mathewc -- merging but will follow-up if you have any further suggestions

PR #11145, opened 2025-06-23
Add Dependabot configuration for .NET SDK updates

JamieMagee:
This PR adds Dependabot configuration to automatically manage .NET SDK version updates in global.json.

This configuration will:
- Monitor the global.json file for .NET SDK version updates
- Create pull requests when new SDK versions are available
- Help keep the project secure with the latest SDK patches

For more information, see: https://devblogs.microsoft.com/dotnet/using-dependabot-to-manage-dotnet-sdk-updates/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+updates:
+  - package-ecosystem: "dotnet-sdk"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "wednesday"
+    ignore:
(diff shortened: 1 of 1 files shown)


jviau (changes requested):
Blocking for now as I believe there is some dependency between our repo and Antares with what SDK version we can use.

fabiocav:
This, unfortunately, is not a safe automation due to environment requirements. Closing, but I appreciate your help and keeping us in mind, @JamieMagee

PR #11159, opened 2025-07-02
Fail in-flight invocations when worker channel shuts down

### LINKED ISSUE #10936: Worker restart due to function timeout does not fail current executions

satvu:

resolves #10936 

### Pull request checklist

### Additional information

Additional PR information


src/WebJobs.Script/Http/ScriptInvocationRequestTransformer.cs
```diff
@@ -0,0 +1,25 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.Azure.WebJobs.Script;
+using Microsoft.Azure.WebJobs.Script.Description;
+using Yarp.ReverseProxy.Forwarder;
+
+public class ScriptInvocationRequestTransformer : HttpTransformer
+{
+    public override async ValueTask TransformRequestAsync(HttpContext httpContext, HttpRequestMessage proxyRequest, string destinationPrefix, CancellationToken cancellationToken)
+    {
+        await base.TransformRequestAsync(httpContext, proxyRequest, destinationPrefix, cancellationToken);
```
fabiocav: Discussed this offline, but I'm concerned this will not maintain the default behavior we previously had. 

Your best course of action is  here would likely be to call `HttpTransformer.Default.TransformRequestAsync` instead, then add our custom logic on top.

src/WebJobs.Script/Http/ScriptInvocationRequestTransformer.cs
```diff
@@ -0,0 +1,28 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.Azure.WebJobs.Script;
+using Microsoft.Azure.WebJobs.Script.Description;
+using Yarp.ReverseProxy.Forwarder;
+
+public class ScriptInvocationRequestTransformer : HttpTransformer
+{
+    public override async ValueTask TransformRequestAsync(HttpContext httpContext, HttpRequestMessage proxyRequest, string destinationPrefix, CancellationToken cancellationToken)
+    {
+        var defaultTransformer = HttpTransformer.Default;
```
fabiocav: nit: remove this local, use `HttpTransformer.Default` directly.

src/WebJobs.Script/Workers/Http/HttpFunctionInvocationDispatcher.cs
```diff
@@ -206,7 +206,7 @@ public Task ShutdownAsync()
             return Task.CompletedTask;
         }
 
-        public Task<bool> RestartWorkerWithInvocationIdAsync(string invocationId)
+        public Task<bool> RestartWorkerWithInvocationIdAsync(string invocationId, Exception exception = null)
```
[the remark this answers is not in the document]
satvu: The unused parameter is here because this class implements `IFunctionInvocationDispatcher` (which includes this method) - this parameter is used in other implementations of this interface.

test/WebJobs.Script.Tests/Workers/Rpc/GrpcWorkerChannelTests.cs
```diff
@@ -1561,6 +1547,72 @@ public async Task NullOutputBinding_DoesNotThrow()
...
+            // Should not throw even if there are no executing invocations
+            _workerChannel.Shutdown(workerException);
+        }
+
+        [Theory]
+        [InlineData(1, true)]
+        [InlineData(3, true)]
+        [InlineData(1, false)]
+        [InlineData(3, false)]
+        public async Task Shutdown_FailsInFlightInvocations(int numberOfInvocations, bool hasFailureException)
```
satvu: Prior to this change, calling `TryFailExecutions` (replaced by `Shutdown`) would not fail executions without an exception being passed in. This test makes sure that in-flight invocations are failed whether or not an exception is passed in.

src/WebJobs.Script/Exceptions/WorkerShutdownException.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.WebJobs.Script.Exceptions
+{
+    internal class WorkerShutdownException : Exception
```
kshyju: Mark it sealed. We can always unseal it if/when needed.
`internal sealed class WorkerShutdownException`

src/WebJobs.Script/Exceptions/WorkerShutdownException.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        public WorkerShutdownException() { }
+
+        public WorkerShutdownException(string message) : base(message) { }
+
```
kshyju: Have you thought about adding a constructor which takes the message & inner exception as well? Similar to this one in the base class.
https://learn.microsoft.com/en-us/dotnet/api/system.exception.-ctor?view=net-9.0#system-exception-ctor(system-string-system-exception)

src/WebJobs.Script/Http/RetryProxyHandler.cs
```diff
@@ -30,11 +32,22 @@ public RetryProxyHandler(HttpMessageHandler innerHandler, ILogger logger)
...
+            {
+                resultSource = scriptInvocationContext.ResultSource;
+            }
+
             var currentDelay = InitialDelay;
             for (int attemptCount = 1; attemptCount <= MaxRetries; attemptCount++)
             {
                 try
                 {
+                    if (resultSource is not null && resultSource.Task.IsFaulted)
```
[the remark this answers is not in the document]
satvu: This needs to be checked before every retry.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -1555,21 +1556,21 @@ public bool IsExecutingInvocation(string invocationId)
             return _executingInvocations.ContainsKey(invocationId);
         }
 
-        public bool TryFailExecutions(Exception workerException)
+        public void Shutdown(Exception workerException)
```
kshyju, line 1557: A thought: Should Shutdown also dispose this instance? The (worker) process is removed/killed in `Dispose` method code, and I see the callers of `Shutdown` is explicitly calling `Dispose` after calling `Shutdown`. Wondering that should be part of Shutdown method itself.

src/WebJobs.Script/Http/ScriptInvocationRequestTransformer.cs
```diff
@@ -0,0 +1,28 @@
...
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Azure.WebJobs.Script.Description;
+using Yarp.ReverseProxy.Forwarder;
+
+namespace Microsoft.Azure.WebJobs.Script.Http
+{
+    public class ScriptInvocationRequestTransformer : HttpTransformer
```
kshyju: Can this be a public class?

kshyju: Mark as `sealed`

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
satvu: They are distinct behaviors, and I think there may be places where we don't call `Shutdown` (prev. `TryFailExecutions`) but we do call `Dispose`. Will have to double check.

We could bundle it into the `Shutdown` call, but this should not be used as a replacement for `Dispose` without further investigation yet. This fix is only scoped to where `TryFailExecutions` was previously called. We would save some lines of code and make the shutdown scenario simpler, but I wonder if it would cause confusion for future code contributors if we bundle this.

liliankasem: Yeah +1 to Sarah, I don't think we should bundle these up. If folks calling Shutdown already are also calling Dispose then we know that is covered. I don't think we need to overload what shutdown does

src/WebJobs.Script/Http/RetryProxyHandler.cs
```diff
@@ -51,6 +64,11 @@ protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage
 
                     currentDelay = Math.Min(currentDelay * 2, MaximumDelay);
                 }
+                catch (WorkerShutdownException)
+                {
+                    _logger.LogInformation("Language worker channel is shutting down. Request will not be retried.");
```
liliankasem: Does this need to be info or should be debug?

liliankasem (approved):

src/WebJobs.Script/Http/RetryProxyHandler.cs
satvu: Downgrading to Debug to match with the grpc shutdown logs.

src/WebJobs.Script.WebHost/WebScriptHostExceptionHandler.cs
```diff
@@ -29,6 +29,13 @@ public async Task OnTimeoutExceptionAsync(ExceptionDispatchInfo exceptionInfo, T
         {
             FunctionTimeoutException timeoutException = exceptionInfo.SourceException as FunctionTimeoutException;
 
+            // this seems to happen when the worker channel is already shutting down. Ex. One timeout is being handled and another comes in during shutdown.
+            if (timeoutException is null)
+            {
+                _logger.LogDebug("A timeout exception has occurred, but worker channel is already shutting down", exceptionInfo.SourceException);
```
kshyju: For the overload you are using here, don't you need a placeholder in the message template?

src/WebJobs.Script.WebHost/WebScriptHostExceptionHandler.cs
```diff
@@ -29,6 +29,13 @@ public async Task OnTimeoutExceptionAsync(ExceptionDispatchInfo exceptionInfo, T
         {
             FunctionTimeoutException timeoutException = exceptionInfo.SourceException as FunctionTimeoutException;
 
+            // this seems to happen when the worker channel is already shutting down. Ex. One timeout is being handled and another comes in during shutdown.
+            if (timeoutException is null)
```
kshyju: If it's not `FunctionTimeoutException`, what exception the caller is sending to this method?

src/WebJobs.Script.WebHost/WebScriptHostExceptionHandler.cs
kshyju: So, this part of the message "_but worker channel is already shutting down_", we are logging that without actually checking the state of the worker channel but simply based on the exception here. Is that a reliable signal to make that conclusion? Should we update the message here to remove the channel state?

kshyju: Also check line 65 for the same.

jviau: Not for `Exception` types. There is a specific parameter for that on all log methods.

src/WebJobs.Script/Http/RetryProxyHandler.cs
```diff
@@ -30,11 +32,22 @@ public RetryProxyHandler(HttpMessageHandler innerHandler, ILogger logger)
...
+            }
+
             var currentDelay = InitialDelay;
             for (int attemptCount = 1; attemptCount <= MaxRetries; attemptCount++)
             {
                 try
                 {
+                    if (resultSource is not null && resultSource.Task.IsFaulted)
+                    {
+                        throw resultSource.Task.Exception?.InnerException ?? new InvalidOperationException("The function invocation tied to this HTTP request failed.");
```
jviau: Is there a better exception type for this? It isn't "invalid" to have called `SendAsync` here. Should we use `HttpRequestException`? You can add a condition below to not catch this exception when `isTaskFaulted` is `true`

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -1555,21 +1556,21 @@ public bool IsExecutingInvocation(string invocationId)
...
+                ? new WorkerShutdownException("Worker encountered a fatal error and is shutting down.", workerException)
+                : new WorkerShutdownException("Worker encountered a fatal error and is shutting down.");
 
             foreach (var invocation in _executingInvocations?.Values)
             {
                 string invocationId = invocation.Context?.ExecutionContext?.InvocationId.ToString();
                 _workerChannelLogger.LogDebug("Worker '{workerId}' encountered a fatal error. Failing invocation: '{invocationId}'", _workerId, invocationId);
-                invocation.Context?.ResultSource?.TrySetException(workerException);
+
+                invocation.Context?.ResultSource?.TrySetException(shutdownException);
```
jviau: Since you only new up this exception and never throw it, it won't have any stack trace on it. Is that acceptable? Or should you throw and catch it before calling `TrySetException`?

src/WebJobs.Script.WebHost/WebScriptHostExceptionHandler.cs
```diff
@@ -29,6 +29,13 @@ public async Task OnTimeoutExceptionAsync(ExceptionDispatchInfo exceptionInfo, T
         {
             FunctionTimeoutException timeoutException = exceptionInfo.SourceException as FunctionTimeoutException;
 
+            // this seems to happen when the worker channel is already shutting down. Ex. One timeout is being handled and another comes in during shutdown.
```
jviau: Nit: can use pattern matching here:

``` CSharp
if (exceptionInfo.SourceException is FunctionTimeoutException timeoutException)
{
    // non-null case
}
else
{
    // null case
}

```

src/WebJobs.Script.WebHost/WebScriptHostExceptionHandler.cs
jviau: Yeah this seems odd, the method is `OnTimeoutExceptionAsync`, but we are now checking if it was called for a non-timeout? Additionally, the log here just assumes that if `timeoutException is null`, then the channel _must_ be shutting down? Will that always be true? How future proof is that assumption?

src/WebJobs.Script/Http/RetryProxyHandler.cs
liliankasem: Are we saying we don't want to retry if the request failed? The second part of this statement will end up in the last exception catch (lines 72-80

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -1555,21 +1556,21 @@ public bool IsExecutingInvocation(string invocationId)
...
         }
 
-        public bool TryFailExecutions(Exception workerException)
+        public void Shutdown(Exception workerException)
         {
-            if (workerException == null)
-            {
-                return false;
-            }
+            WorkerShutdownException shutdownException = workerException is not null
```
jviau: Want to make sure I understand the core behavior change of this PR:

_Before_: some cases would call this with a `null` workerException, and thus this was a no-op.
_After_: those same places may still call with `null` workerException, but it is no longer a no-op and will shut the channel down.

Am I understanding this right?

src/WebJobs.Script/Http/RetryProxyHandler.cs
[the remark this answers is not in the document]
liliankasem: Will the value of `resultSource?.Task.IsFaulted` change with every retry when resultSource is set before the foreach loop?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
jviau: Is this the exception which ultimately gets exposed to extensions via the `TryExecuteAsync` and the `FunctionResult.Exception` property?

If so, changing the exception type on them is a breaking change.

jviau (changes requested):
My main concern is the potential breaking change by switching to a `WorkerShutdownException`. But maybe my assumption this exception makes it to extensions is wrong?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
liliankasem: IIRC the shutdown of the channel was happening anyway as we dispose right after "TryFailExecutions" is called:

- https://github.com/Azure/azure-functions-host/blob/48712e88f8f2a06653a78ed28cb3907d65bd2717/src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs#L272
- https://github.com/Azure/azure-functions-host/blob/48712e88f8f2a06653a78ed28cb3907d65bd2717/src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs#L298
- https://github.com/Azure/azure-functions-host/blob/48712e88f8f2a06653a78ed28cb3907d65bd2717/src/WebJobs.Script/Workers/Rpc/JobHostRpcWorkerChannelManager.cs#L52

liliankasem: Before `TryFailExecutions` itself was a no-op and this never happens

```
invocation.Context?.ResultSource?.TrySetException(workerException);
RemoveExecutingInvocation(invocationId);
```

But channel is still gone/disposed regardless

jviau: Right my bad. Not that the channel will shut down, but that failing the invocation was a potential no-op before.

liliankasem: Yeah correct! But Sarah let me know if you observed that behaviour to confirm?

liliankasem: I would assume it was a bug honestly that `RemoveExecutingInvocation` was not happening on shut down.

It also looks like every instance where `TryFailExecutions` was called we did provide a workerException and the null check was just defensive

src/WebJobs.Script.WebHost/WebScriptHostExceptionHandler.cs
kshyju: Yea, for that overload, Exception is the first parameter, I think.

src/WebJobs.Script.WebHost/WebScriptHostExceptionHandler.cs
satvu: I've only reproduced this twice, but timeoutException has ended up `null` in this method - still digging into why that is, but from my observation of the two events is as described in the comment. Let me continue to investigate this.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
satvu: The channel will shut down in both _before_ and _after_ scenarios.

_Before:_ channel is shutting down, and we will fail in-flight invocations ONLY if an exception is passed in
_After:_ channel is shutting down, we fail ALL in-flight invocations and wrap it in a WorkerShutdownException

We wrap it in a WorkerShutdownException because we often restart channels based on _one_ invocation timing out, and it's not exactly accurate to fail the other invocations as "timeouts" as well.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
satvu: I think it would be - not familiar with the extension side, but this is what we wait for then use in `BindOutputsAsync` inside `InvokeCore` in the host.

What if we added a check to just use `WorkerShutdownException` when the exception passed in is a `FunctionTimeoutException`? This wouldn't break the other scenarios we currently have with other exception types being passed in.

src/WebJobs.Script/Http/RetryProxyHandler.cs
satvu: We don't want to retry if the invocation was failed on the functions, non-http side - without this change, the proxying service will keep retrying even though the host has stopped tracking the invocation. This will align the http service with the host by making sure to check that the invocation is still considered "valid" or "executing".

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
jviau: We definitely can't change the type from `FunctionTimeoutException`. Extensions have direct type checks on that:
https://github.com/Azure/azure-functions-durable-extension/blob/0729864c6a472f01f599d5506c274b0f0eff9eea/src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs#L172

satvu: Discussed offline - writing another comment to explain this and other concerns, posting soon.

src/WebJobs.Script/Http/RetryProxyHandler.cs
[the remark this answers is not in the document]
satvu: `resultSource` is a task completion source we are observing

satvu:
The new approach:

To avoid a breaking change with Durable, we will introduce an exception that _derives from the existing FunctionTimeoutException_, and looks like this: `FunctionAbortException : FunctionTimeoutException`. **This will only be used when the exception passed in is** `null` or `FunctionTimeoutException` to preserve the current behavior for all other logs passed in (in case other components depend on any other type exceptions this method sets results to). 
 
In these cases, in the `Shutdown` method, we create a new `FunctionAbortException` to fail executions. For the `FunctionTimeoutException` cases, we additionally pass in the original `FunctionTimeoutException` as an inner exception (?). 
 
Since durable already has handling for `FunctineTimeoutException` in some places, this will ensure that the resulting behavior is as desired by the extension team. This also covers other extensions that perform any specific behavior based on this long-existing exception.

Edit: The new exception type is also being added to the WebJobs SDK (https://github.com/Azure/azure-webjobs-sdk/pull/3139) so extensions can handle this type of failure with more granularity.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
satvu: Resolved - scoped the changes of creating a new exception to only the cases of `null` or `FunctionTimeoutException`.

src/WebJobs.Script.WebHost/WebScriptHostExceptionHandler.cs
```diff
@@ -48,14 +48,16 @@ public async Task OnTimeoutExceptionAsync(ExceptionDispatchInfo exceptionInfo, T
...
             if (!functionInvocationDispatcher.State.Equals(FunctionInvocationDispatcherState.Default))
             {
                 _logger.LogWarning($"A function timeout has occurred. Restarting worker process executing invocationId '{timeoutException.InstanceId}'.", exceptionInfo.SourceException);
+
                 // If invocation id is not found in any of the workers => worker is already disposed. No action needed.
-                await functionInvocationDispatcher.RestartWorkerWithInvocationIdAsync(timeoutException.InstanceId.ToString());
-                _logger.LogWarning("Restart of language worker process(es) completed.", exceptionInfo.SourceException);
+                await functionInvocationDispatcher.RestartWorkerWithInvocationIdAsync(timeoutException.InstanceId.ToString(), timeoutException);
+
+                _logger.LogWarning("Attempt to restart language worker process(es) completed.", exceptionInfo.SourceException);
```
satvu, line 60: Changed this error message (which is why there's a failing test). Previously, invocation ids of functions that were no longer running would trigger the call above to `RestartWorkerWithInvocationIdAsync` and result in this log even though no restart took place.

This would happen less with the change introduced in the PR (since all invocations will be failed and no longer time out much later), but from local testing this can still be logged multiple times in durable scenarios even when no restart takes place. I think it can be confusing, so I added this change for review. I also updated logging in `RestartWorkerWithInvocationIdAsync` which would give more insight on whether or not a restart took place.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -1555,21 +1557,24 @@ public bool IsExecutingInvocation(string invocationId)
...
+                shutdownException = new FunctionTimeoutAbortException(workerException?.Message ?? "Worker channel is shutting down. Aborting function.", workerException);
             }
 
             foreach (var invocation in _executingInvocations?.Values)
             {
                 string invocationId = invocation.Context?.ExecutionContext?.InvocationId.ToString();
                 _workerChannelLogger.LogDebug("Worker '{workerId}' encountered a fatal error. Failing invocation: '{invocationId}'", _workerId, invocationId);
-                invocation.Context?.ResultSource?.TrySetException(workerException);
+
+                invocation.Context?.ResultSource?.TrySetException(shutdownException);
```
jviau: @RohitRanjanMS has a PR which does more work on setting this exception. #11313

I mentioned in it we should extract to a common helper method and this is a good justification for that. Can you sync with Rohit to coordinate ensuring the activity error tracking is performed here also?

jviau (approved):
Approved, but make sure to follow up on the PR from Rohit.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
satvu: Discussed w/ Rohit, he made a change to his PR that is reflected in this one.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -1552,11 +1554,13 @@ public bool IsExecutingInvocation(string invocationId)
...
-        public bool TryFailExecutions(Exception workerException)
+        public void Shutdown(Exception workerException)
         {
-            if (workerException == null)
+            var shutdownException = workerException;
+
+            if (workerException is null || workerException is FunctionTimeoutException)
             {
-                return false;
+                shutdownException = new FunctionTimeoutAbortException(workerException?.Message ?? "Worker channel is shutting down. Aborting function.", workerException);
```
RohitRanjanMS, line 1563: `shutdownException` is never used. You probably meant to pass `shutdownException` instead when SetException is called.

test/WebJobs.Script.Tests/Workers/Rpc/GrpcWorkerChannelTests.cs
```diff
@@ -1581,6 +1674,7 @@ public async Task Ensure_Failure_Status_On_CurrentActivity_WhenInvocationFailed(
 
             Assert.Equal(ActivityStatusCode.Error, activity.Status);
             Assert.Contains("Failure", activity.StatusDescription);
+
```
RohitRanjanMS: nit: remove the empty line

test/WebJobs.Script.Tests/Workers/Rpc/GrpcWorkerChannelTests.cs
```diff
@@ -1562,6 +1549,112 @@ public async Task NullOutputBinding_DoesNotThrow()
...
+            _workerChannel.Shutdown(workerException);
+
+            for (int i = 0; i < numberOfInFlightInvocations; i++)
+            {
+                Assert.False(_workerChannel.IsExecutingInvocation(invocationIds[i].ToString()),
+                    $"Invocation {i} should no longer be executing");
+
+                var resultSource = invocationContexts[i].ResultSource;
+                Assert.Equal(TaskStatus.Faulted, resultSource.Task.Status);
+                Assert.Equal(typeof(FunctionTimeoutAbortException), resultSource.Task.Exception.InnerException.GetType());
```
RohitRanjanMS: Should this be `FunctionTimeoutAbortException`? Assuming you are going to pass `shutdownException` instead of `workerException`.

test/WebJobs.Script.Tests/Workers/Rpc/GrpcWorkerChannelTests.cs
```diff
@@ -1562,6 +1549,112 @@ public async Task NullOutputBinding_DoesNotThrow()
...
+                Assert.False(_workerChannel.IsExecutingInvocation(invocationIds[i].ToString()),
+                    $"Invocation {i} should no longer be executing");
+
+                var resultSource = invocationContexts[i].ResultSource;
+                Assert.Equal(TaskStatus.Faulted, resultSource.Task.Status);
+            }
+        }
+
+        [Theory]
+        [InlineData(1, true)]
```
RohitRanjanMS, line 1612: Can you also add `false` scenario?

test/WebJobs.Script.Tests/Workers/Rpc/GrpcWorkerChannelTests.cs
```diff
@@ -1562,6 +1549,112 @@ public async Task NullOutputBinding_DoesNotThrow()
...
+            var workerException = hasFailureException ? new FunctionTimeoutException("Invocation timed out.") : null;
+
+            _workerChannel.Shutdown(workerException);
+
+            for (int i = 0; i < numberOfInFlightInvocations; i++)
+            {
+                Assert.False(_workerChannel.IsExecutingInvocation(invocationIds[i].ToString()),
+                    $"Invocation {i} should no longer be executing");
+
+                var resultSource = invocationContexts[i].ResultSource;
```
RohitRanjanMS, line 1657: You can also assert the logs.

RohitRanjanMS (changes requested):

jviau (approved):

RohitRanjanMS (approved):
Please update the release notes.

RohitRanjanMS (approved):

Issue #11171, opened 2025-07-08
[HealthChecks] Add health check HTTP endpoint


jviau:
Add an HTTP endpoint for health checks. Will fill out more details later.

Issue #11010, opened 2025-04-17
[HealthChecks] Add Diagnostic HealthChecks


jviau:
## Health Checks Today

Customers on dedicated/premium function apps can configure a health check today. This is an HTTP endpoint which the platform will poll to determine instance health. This work will not replace that but potentially augment it. The health checks today are purely customer implemented, only offering health for the customer side of the instance. This work will add platform / functions host health. Customer's will have the option to incorporate this health check with their own.

## Proposal

To improve observability of a functions app, we should incorporate a robust health checks system in the functions host.  This will give immediately feedback on the general health of a function app and can be incorporated into liveness probes as necessary (e.g. #9181).

## Steps

1. Adopt `Microsoft.Extensions.Diagnostics.HealthChecks` package.
2. Design and implement select health checks for different areas.
3. Expose a health check endpoint.
4. Publish health check results to telemetry.

Will expand on this issue at a later date with more details and a breakdown into sub-tasks.

PR #11173, opened 2025-07-09
[HealthChecks] Add health check middleware

jviau:

Part of #11010, #11171

### Pull request checklist

### Additional information

Adds supporting files for the eventual addition of a health check endpoint. The added files do two primary things:

1. `HealthCheckWaitMiddleware.cs` -- this file enables HTTP long polling of health probes by adding a `?wait={timeInSeconds}` query string to the health check calls.
2. `HealthCheckResponseWriter` -- this files writes the HTTP responses for health checks. It has 2 modes, a minimal or an expanded response.
    - The expanded response is provided by [AspNetCore.Diagnostics.HealthChecks](https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks). This ensures it is compatible with other components from that ecosystem of health check packages.
    - Minimal/expanded response is controlled by `?expand={true/false}`. Default is `false` if not included.
    - Note this intentionally does not set status code. That will be done by the AspNetCore health check middleware (which will be added later)

This PR also removes `ApiErrorModel` (it is unused) and introduces `ErrorResponse`, which is modeled after the ARM recommended error structure.


src/WebJobs.Script.WebHost/Diagnostics/HealthChecks/HealthCheckResponseWriter.cs
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        private static Task WriteMinimalResponseAsync(HttpContext httpContext, HealthReport report)
+        {
+            MinimalResponse body = new(report.Status);
+            return JsonSerializer.SerializeAsync(
+                httpContext.Response.Body, body, _options, httpContext.RequestAborted);
+        }
+
+        private static JsonSerializerOptions CreateJsonOptions()
```
kshyju: wondering if it would be better to move this into a separate file to improve discoverability. That way, anyone needing similar behavior in the future can easily find and reuse it. either by leveraging existing entries if they align with their needs, or by adding new ones.

Similar to this: https://github.com/dotnet/aspnetcore/blob/3e86bc0931085adca23787374ae4bed9dfcbfede/src/Components/Shared/src/JsonSerializerOptionsProvider.cs#L8

src/WebJobs.Script.WebHost/Diagnostics/HealthChecks/HealthCheckResponseWriter.cs
```diff
@@ -0,0 +1,60 @@
...
+using HealthChecks.UI.Client;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics.HealthChecks
+{
+    public class HealthCheckResponseWriter
+    {
+        private static readonly JsonSerializerOptions _options = CreateJsonOptions();
```
kshyju: Don't we use Pascal case for static readonly items?

src/WebJobs.Script.WebHost/Models/ApiErrorModel.cs
```diff
@@ -1,45 +0,0 @@
...
-// Licensed under the MIT License. See License.txt in the project root for license information.
-
-using System;
-using System.Collections.Generic;
-using System.Net;
-using Newtonsoft.Json;
-
-namespace Microsoft.Azure.WebJobs.Script.WebHost.Models
-{
-    public class ApiErrorModel
```
liliankasem, line 11: This is a public class, can we just get rid of it like that? Should we deprecate instead?

src/WebJobs.Script/JsonSerializerOptionsProvider.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// Provides constants related to JSON serialization options used.
+    /// </summary>
+    public static class JsonSerializerOptionsProvider
+    {
+        /// <summary>
+        /// Gets the default JSON serializer options used across the functions host.
+        /// </summary>
+        public static readonly JsonSerializerOptions Options = CreateJsonOptions();
+
+        private static JsonSerializerOptions CreateJsonOptions()
```
liliankasem, line 22: This would be great to reuse - we have a bunch of places where we create a new JsonSerializerOptions that should be consistent across the Host

liliankasem (approved):
lgtm, minor comments. deps json looks good, I don't see any major version bumps

src/WebJobs.Script.WebHost/Models/ApiErrorModel.cs
jviau: We are not a publicly consumable package, so we don't have strict breaking change requirements

src/WebJobs.Script/JsonSerializerOptionsProvider.cs
jviau: Yeah, we can work on re-using this class later

src/WebJobs.Script.WebHost/Diagnostics/HealthChecks/HealthCheckWaitMiddleware.cs
```diff
@@ -0,0 +1,40 @@
...
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Azure.WebJobs.Script.WebHost.Models;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics.HealthChecks
+{
+    public class HealthCheckWaitMiddleware(RequestDelegate next, IScriptHostManager manager)
```
kshyju: Where and how are we hooking this up to the request processing pipeline?  I did not see this middleware registered anywhere.

kshyju: mark `sealed`.

jviau: It is not hooked up yet, will be doing that in a later PR. I am 'staging' all the different parts of health checks first, then integrating them into startup as the last step.

src/WebJobs.Script.WebHost/Diagnostics/HealthChecks/HealthCheckResponseWriter.cs
```diff
@@ -0,0 +1,43 @@
...
+            }
+
+            return WriteMinimalResponseAsync(httpContext, report);
+        }
+
+        private static Task WriteMinimalResponseAsync(HttpContext httpContext, HealthReport report)
+        {
+            MinimalResponse body = new(report.Status);
+            return JsonSerializer.SerializeAsync(
+                httpContext.Response.Body, body, JsonSerializerOptionsProvider.Options, httpContext.RequestAborted);
```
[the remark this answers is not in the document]
jviau: This is intentional. No idea why copilot thinks this is an issue.

src/WebJobs.Script/JsonSerializerOptionsProvider.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Microsoft.Azure.WebJobs.Script
+{
+    /// <summary>
+    /// Provides constants related to JSON serialization options used.
+    /// </summary>
+    public static class JsonSerializerOptionsProvider
+    {
+        /// <summary>
+        /// Gets the default JSON serializer options used across the functions host.
```
kshyju: nit: Not sure we should mention "used across the functions host". I prefer something along the line of "Shared Json serializer with the following settings: 1) Trailing commas allowed, 2) camel casing ...."

test/WebJobs.Script.Tests/Diagnostics/HealthChecks/HealthCheckResponseWriterTests.cs
```diff
@@ -0,0 +1,120 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text.Json;
+using System.Text.Json.Nodes;
+using System.Threading.Tasks;
+using FluentAssertions;
+using HealthChecks.UI.Client;
```
kshyju: Remove unnecessary usings.

kshyju (approved):

src/WebJobs.Script.WebHost/Diagnostics/HealthChecks/HealthCheckWaitMiddleware.cs
```diff
@@ -0,0 +1,40 @@
...
+        public async Task InvokeAsync(HttpContext context)
+        {
+            ArgumentNullException.ThrowIfNull(context);
+
+            // If specified, the ?wait={seconds} query param will wait for an
+            // active script host for that duration. This is to avoid excessive polling
+            // when waiting for the initial readiness probe.
+            if (context.Request.Query.TryGetValue("wait", out StringValues wait))
+            {
+                if (!int.TryParse(wait.ToString(), out int waitSeconds) || waitSeconds < 0)
```
RohitRanjanMS, line 27: We can have upper bound check to prevent abuse.

src/WebJobs.Script.WebHost/Diagnostics/HealthChecks/HealthCheckWaitMiddleware.cs
```diff
@@ -0,0 +1,40 @@
...
+                {
+                    context.Response.StatusCode = StatusCodes.Status400BadRequest;
+                    await context.Response.WriteAsJsonAsync(
+                        ErrorResponse.BadArgument("'wait' query param must be a positive integer", $"wait={wait}"));
+                    return;
+                }
+
+                await _manager.DelayUntilHostReadyAsync(waitSeconds);
```
RohitRanjanMS: Similar to how you have used httpContext.RequestAborted while writing response, can we also use it here to avoid hanging on client disconnect.

src/WebJobs.Script/runtimeassemblies.json
```diff
@@ -67,6 +67,14 @@
       "name": "Grpc.Net.Common",
       "resolutionPolicy": "private"
     },
+    {
```
RohitRanjanMS, line 70: Should this be minorMatchOrLower? We may have customers using healthcheck in InProc.

jviau: They may be using health checks, but we don't need to interop these 2 assemblies with customer code. So we will keep it `private` so they get their own copy. Plus these changes are not planned for in-proc.

src/WebJobs.Script.WebHost/Diagnostics/HealthChecks/HealthCheckWaitMiddleware.cs
jviau: I opted to cap the wait time at 60s (and not error out).

RohitRanjanMS: Could this be confusing? If I set a higher wait time, it still only waits 60 seconds—and I won’t know why.

RohitRanjanMS (approved):

src/WebJobs.Script/runtimeassemblies.json
```diff
@@ -775,6 +787,10 @@
       "name": "Microsoft.Extensions.Primitives",
       "resolutionPolicy": "minorMatchOrLower"
     },
+    {
+      "name": "Microsoft.Extensions.Telemetry.Abstractions",
+      "resolutionPolicy": "minorMatchOrLower"
```
brettsam: Do any of these other additions interact with extension (customer) code? Our policy has been that anything new must be "private" in order to not break existing behavior for extensions -- they may already be referencing some other version than the host uses.

jviau: I actually wound up not using that assembly at all. I have removed this change.

src/WebJobs.Script.WebHost/Diagnostics/HealthChecks/HealthCheckWaitMiddleware.cs
jviau: That is true to an extent. Given HTTP client timeouts tend to be 30s or 60s by default, I found this to be a safe assumption that no one would expect a >60s long poll. Additionally, this isn't a crucial enough behavior to worry about. The caller is expected to perform this poll in a loop, so they should never have a hard dependency on how long this call will wait for.

brettsam (approved):

PR #11196, opened 2025-07-18
[WIP] Flow cancellation token to PendingItem

liliankasem:

resolves #issue_for_this_pr

### Pull request checklist

### Additional information

Whilst investigating the core tools `ctrl+c` [issue](https://github.com/Azure/azure-functions-core-tools/issues/4355) I was able to narrow down the problem to this area of the code.

Cancellation tokens are not passed through and so when we hit `Ctrl+C` and trigger the application stopping lifetime event, nothing happens because we are either waiting for a response from the worker (which never happens in this state since we kill the channel), or until timeout. The timeout is set to 1 minute and that's why it often takes a minute for the host to shut down (if the host is stopped during startup).

### DIFF
--- a/src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
+++ b/src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
@@ -29,6 +29,7 @@ namespace Microsoft.Azure.WebJobs.Script
         private readonly IEnvironment _environment;
         private readonly IWebHostRpcWorkerChannelManager _channelManager;
         private readonly IScriptHostManager _scriptHostManager;
+        private readonly IHostApplicationLifetime _applicationLifetime;
         private string _workerRuntime;
         private ImmutableArray<FunctionMetadata> _functions;
         private IHost _currentJobHost = null;
@@ -38,7 +39,8 @@ namespace Microsoft.Azure.WebJobs.Script
             ILogger<WorkerFunctionMetadataProvider> logger,
             IEnvironment environment,
             IWebHostRpcWorkerChannelManager webHostRpcWorkerChannelManager,
-            IScriptHostManager scriptHostManager)
+            IScriptHostManager scriptHostManager,
+            IHostApplicationLifetime applicationLifetime)
         {
             _scriptOptions = scriptOptions;
             _logger = logger;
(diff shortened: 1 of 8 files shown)


src/WebJobs.Script/Workers/Rpc/WebHostRpcWorkerChannelManager.cs
```diff
@@ -72,22 +73,24 @@ public WebHostRpcWorkerChannelManager(IScriptEventManager eventManager,
...
 
-        internal async Task<IRpcWorkerChannel> InitializeLanguageWorkerChannel(IEnumerable<RpcWorkerConfig> workerConfigs, string runtime, string scriptRootPath)
+        internal async Task<IRpcWorkerChannel> InitializeLanguageWorkerChannel(IEnumerable<RpcWorkerConfig> workerConfigs, string runtime, string scriptRootPath, CancellationToken cancellationToken = default)
         {
+            cancellationToken.ThrowIfCancellationRequested();
+
             IRpcWorkerChannel rpcWorkerChannel = null;
             string workerId = Guid.NewGuid().ToString();
-            _logger.LogDebug("Creating language worker channel for runtime:{runtime}", runtime);
+            _logger.LogWarning("Creating language worker channel for runtime:{runtime}", runtime);
```
liliankasem, line 88: revert

src/WebJobs.Script/Host/WorkerFunctionMetadataProvider.cs
```diff
@@ -89,7 +92,7 @@ public async Task<FunctionMetadataResult> GetFunctionMetadataAsync(IEnumerable<R
                     if (IsJobHostStarting())
                     {
                         _logger.LogDebug("JobHost is starting with state '{State}'. Initializing worker channel.", _scriptHostManager.State);
-                        await _channelManager.InitializeChannelAsync(workerConfigs, _workerRuntime);
+                        await _channelManager.InitializeChannelAsync(workerConfigs, _workerRuntime, _applicationLifetime.ApplicationStopping);
```
liliankasem, line 95: So my idea was to pass the `_applicationLifetime.ApplicationStopping` CT to `StartWorkerProcessAsync` -> PendingItem. (it doesn't have to come through `WorkerFunctionMetadataProvider` so let me know if you have better ideas)

The problem I am seeing with this is that when I cancel (exit the host) we don't see CT triggered until _after_ the timeout period of `PendingItem` has run its course. We are blocking in a weird way and we need to get around that block for the CT to flow properly.

cc: @jviau @fabiocav

liliankasem:
Closing for now as I am not actively working on this.

PR #11202, opened 2025-07-22
adding a read-only ContainerApps secret repository

### LINKED ISSUE #11166: Aspire: Add read-only secret repository for ACA
We currently have a read/write K8S secret repository, but it won't work as expected in ACA where we don't have write access. So we need a repository that supports only reading from the mounted volume, and not updating/writing.

brettsam:
fixes #11166 

Container Apps currently uses blob storage for it's secrets, which are not manageable easily.  We're transitioning secrets to be managed via [Container Apps itself](https://learn.microsoft.com/en-us/azure/container-apps/manage-secrets?tabs=azure-portal). In this case, the secrets will be in a drive in the container the same way as our K8S secrets are.

The main difference here is that these are read-only. We have no mechanism for auto-generating these in Container Apps, so they need to be created separately. 

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
+++ b/src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
@@ -1,4 +1,7 @@
-﻿using System;
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
--- a/src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
+++ b/src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
@@ -1,0 +1,123 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost;
+
+public class ContainerAppsSecretsRepository : ISecretsRepository
+{
+    internal const string ContainerAppsSecretsDir = "/run/secrets/functions-keys";
+
+    // host.master = value
+    private const string MasterKey = "host.master";
+    // host.function.{keyName} = value
+    private const string HostFunctionKeyPrefix = "host.function.";
+    // host.systemKey.{keyName} = value
+    private const string SystemKeyPrefix = "host.systemKey.";
+    // functions.{functionName}.{keyName} = value
+    private const string FunctionKeyPrefix = "functions.";
+
+    private readonly ILogger<ContainerAppsSecretsRepository> _logger;
+
+    public ContainerAppsSecretsRepository(ILogger<ContainerAppsSecretsRepository> logger)
+    {
+        _logger = logger;
+    }
+
+    public event EventHandler<SecretsChangedEventArgs> SecretsChanged;
+
+    public bool IsEncryptionSupported => false;
+
+    public string Name => nameof(ContainerAppsSecretsRepository);
+
+    public async Task<ScriptSecrets> ReadAsync(ScriptSecretsType type, string functionName)
+    {
+        if (type == ScriptSecretsType.Function && string.IsNullOrEmpty(functionName))
+        {
+            throw new ArgumentNullException(nameof(functionName), $"{nameof(functionName)} cannot be null or empty with {nameof(type)} = {nameof(ScriptSecretsType.Function)}");
+        }
+
+        return type == ScriptSecretsType.Host ? await ReadHostSecretsAsync() : await ReadFunctionSecretsAsync(functionName?.ToLowerInvariant());
+    }
+
+    public Task WriteAsync(ScriptSecretsType type, string functionName, ScriptSecrets secrets)
+        => throw new NotImplementedException();
+
+    private async Task<ScriptSecrets> ReadHostSecretsAsync()
+    {
+        var secrets = await GetFromFilesAsync(ContainerAppsSecretsDir);
+
+        HostSecrets hostSecrets = new HostSecrets()
+        {
+            FunctionKeys = [],
+            SystemKeys = []
+        };
+
+        foreach (var pair in secrets)
+        {
+            if (pair.Key.StartsWith(MasterKey, StringComparison.OrdinalIgnoreCase))
+            {
+                hostSecrets.MasterKey = new Key("master", pair.Value);
+            }
+            else if (pair.Key.StartsWith(HostFunctionKeyPrefix, StringComparison.OrdinalIgnoreCase))
+            {
+                hostSecrets.FunctionKeys.Add(ParseKeyWithPrefix(HostFunctionKeyPrefix, pair.Key, pair.Value));
+            }
+            else if (pair.Key.StartsWith(SystemKeyPrefix))
+            {
+                hostSecrets.SystemKeys.Add(ParseKeyWithPrefix(SystemKeyPrefix, pair.Key, pair.Value));
+            }
+        }
+
+        // Always return a HostSecrets object, even if empty. This will prevent the SecretManager from thinking
+        // it needs to create and persist new secrets, which is not supported in Container Apps.
+        return hostSecrets;
+    }
+
+    private async Task<ScriptSecrets> ReadFunctionSecretsAsync(string functionName)
+    {
+        var secrets = await GetFromFilesAsync(ContainerAppsSecretsDir);
+
+        var prefix = $"{FunctionKeyPrefix}{functionName}.";
+
+        var functionSecrets = new FunctionSecrets()
+        {
+            Keys = secrets
+                .Where(p => p.Key.StartsWith(prefix, StringComparison.OrdinalIgnoreCase))
+                .Select(p => ParseKeyWithPrefix(prefix, p.Key, p.Value))
+                .ToList()
+        };
+
+        // Always return a FunctionSecrets object, even if empty. This will prevent the SecretManager from thinking
+        // it needs to create and persist new secrets, which is not supported in Container Apps.
+        return functionSecrets;
+    }
+
+    private static async Task<IDictionary<string, string>> GetFromFilesAsync(string path)
+    {
+        string[] files = await FileUtility.GetFilesAsync(path, "*");
+        var secrets = new Dictionary<string, string>(files.Length);
+
+        foreach (var file in files)
+        {
+            secrets.Add(Path.GetFileName(file), await FileUtility.ReadAsync(file));
+        }
+
+        return secrets;
+    }
+
+    public Task WriteSnapshotAsync(ScriptSecretsType type, string functionName, ScriptSecrets secrets)
+        => throw new NotImplementedException();
+
+    public Task PurgeOldSecretsAsync(IList<string> currentFunctions, ILogger logger)
+        => throw new NotImplementedException();
+
+    public Task<string[]> GetSecretSnapshots(ScriptSecretsType type, string functionName)
+        => throw new NotImplementedException();
+
+    private static Key ParseKeyWithPrefix(string prefix, string key, string value)
+        => new(key.Substring(prefix.Length), value);
+}
--- a/src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
+++ b/src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
@@ -49,7 +49,7 @@ public class ContainerAppsSecretsRepository : ISecretsRepository
             throw new ArgumentNullException(nameof(functionName), $"{nameof(functionName)} cannot be null or empty with {nameof(type)} = {nameof(ScriptSecretsType.Function)}");
         }
 
-        return type == ScriptSecretsType.Host ? await ReadHostSecretsAsync() : await ReadFunctionSecretsAsync(functionName?.ToLowerInvariant());
+        return type == ScriptSecretsType.Host ? await ReadHostSecretsAsync() : await ReadFunctionSecretsAsync(functionName.ToLowerInvariant());
     }
 
     public Task WriteAsync(ScriptSecretsType type, string functionName, ScriptSecrets secrets)
@@ -122,14 +122,23 @@ public class ContainerAppsSecretsRepository : ISecretsRepository
         return secrets;
     }
 
+    /// <summary>
+    /// no-op - allow stale secrets to remain.
+    /// </summary>
+    public async Task PurgeOldSecretsAsync(IList<string> currentFunctions, ILogger logger)
+        => await Task.Yield();
+
+    /// <summary>
+    /// Runtime is not responsible for encryption so this code will never be executed.
+    /// </summary>
     public Task WriteSnapshotAsync(ScriptSecretsType type, string functionName, ScriptSecrets secrets)
-        => throw new NotImplementedException();
+        => throw new NotSupportedException();
 
-    public Task PurgeOldSecretsAsync(IList<string> currentFunctions, ILogger logger)
-        => throw new NotImplementedException();
-
+    /// <summary>
+    /// Runtime is not responsible for encryption so this code will never be executed.
+    /// </summary>
     public Task<string[]> GetSecretSnapshots(ScriptSecretsType type, string functionName)
-        => throw new NotImplementedException();
+        => throw new NotSupportedException();
 
     private static Key ParseKeyWithPrefix(string prefix, string key, string value)
         => new(key.Substring(prefix.Length), value);
--- a/src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
+++ b/src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 
@@ -104,16 +105,20 @@ public class ContainerAppsSecretsRepository : ISecretsRepository
         return functionSecrets;
     }
 
-    private static async Task<IDictionary<string, string>> GetFromFilesAsync(string path)
+    private async Task<IDictionary<string, string>> GetFromFilesAsync(string path)
     {
         string[] files = await FileUtility.GetFilesAsync(path, "*");
         var secrets = new Dictionary<string, string>(files.Length);
 
+        StringBuilder sb = new StringBuilder("Loaded secrets from files:");
+
         foreach (var file in files)
         {
             secrets.Add(Path.GetFileName(file), await FileUtility.ReadAsync(file));
+            sb.AppendLine($"  {file}");
         }
 
+        _logger.LogDebug(sb.ToString());
         return secrets;
     }
 
--- a/src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
+++ b/src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
@@ -30,7 +30,12 @@ public class ContainerAppsSecretsRepository : ISecretsRepository
         _logger = logger;
     }
 
-    public event EventHandler<SecretsChangedEventArgs> SecretsChanged;
+    // explicitly implementing this to avoid "unused" warnings on build
+    event EventHandler<SecretsChangedEventArgs> ISecretsRepository.SecretsChanged
+    {
+        add { }
+        remove { }
+    }
 
     public bool IsEncryptionSupported => false;
 
--- a/src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
+++ b/src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
@@ -44,22 +44,22 @@ public class ContainerAppsSecretsRepository : ISecretsRepository
 
     public async Task<ScriptSecrets> ReadAsync(ScriptSecretsType type, string functionName)
     {
-        if (type == ScriptSecretsType.Function && string.IsNullOrEmpty(functionName))
+        if (type == ScriptSecretsType.Function)
         {
-            throw new ArgumentNullException(nameof(functionName), $"{nameof(functionName)} cannot be null or empty with {nameof(type)} = {nameof(ScriptSecretsType.Function)}");
+            ArgumentException.ThrowIfNullOrEmpty(functionName);
         }
 
         return type == ScriptSecretsType.Host ? await ReadHostSecretsAsync() : await ReadFunctionSecretsAsync(functionName.ToLowerInvariant());
     }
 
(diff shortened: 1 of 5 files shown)


src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
```diff
@@ -0,0 +1,131 @@
...
+    public string Name => nameof(ContainerAppsSecretsRepository);
+
+    public async Task<ScriptSecrets> ReadAsync(ScriptSecretsType type, string functionName)
+    {
+        if (type == ScriptSecretsType.Function && string.IsNullOrEmpty(functionName))
+        {
+            throw new ArgumentNullException(nameof(functionName), $"{nameof(functionName)} cannot be null or empty with {nameof(type)} = {nameof(ScriptSecretsType.Function)}");
+        }
+
+        return type == ScriptSecretsType.Host ? await ReadHostSecretsAsync() : await ReadFunctionSecretsAsync(functionName?.ToLowerInvariant());
```
mathewc: nit: null conditional on functionName is misleading since we know it can't be null at this point

src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
```diff
@@ -0,0 +1,131 @@
...
+            {
+                hostSecrets.FunctionKeys.Add(ParseKeyWithPrefix(HostFunctionKeyPrefix, pair.Key, pair.Value));
+            }
+            else if (pair.Key.StartsWith(SystemKeyPrefix))
+            {
+                hostSecrets.SystemKeys.Add(ParseKeyWithPrefix(SystemKeyPrefix, pair.Key, pair.Value));
+            }
+        }
+
+        // Always return a HostSecrets object, even if empty. This will prevent the SecretManager from thinking
```
mathewc, line 84: So who is responsible for ensuring that require keys like the master key are present? What happens if the key isn't present but /admin APIs are called by control plane?

src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
```diff
@@ -0,0 +1,131 @@
...
+    // host.systemKey.{keyName} = value
+    private const string SystemKeyPrefix = "host.systemKey.";
+    // functions.{functionName}.{keyName} = value
+    private const string FunctionKeyPrefix = "functions.";
+
+    private readonly ILogger<ContainerAppsSecretsRepository> _logger;
+
+    public ContainerAppsSecretsRepository(ILogger<ContainerAppsSecretsRepository> logger)
+    {
+        _logger = logger;
```
mathewc, line 31: _logger not used so can remove? Actually, I'm wondering if it wouldn't be good to do some logging in GetFromFilesAsync so we have some diagnostics on whether any files are present, how many, etc.

src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
```diff
@@ -0,0 +1,131 @@
...
+            secrets.Add(Path.GetFileName(file), await FileUtility.ReadAsync(file));
+        }
+
+        return secrets;
+    }
+
+    public Task WriteSnapshotAsync(ScriptSecretsType type, string functionName, ScriptSecrets secrets)
+        => throw new NotImplementedException();
+
+    public Task PurgeOldSecretsAsync(IList<string> currentFunctions, ILogger logger)
```
mathewc, line 132: If SecretManager is periodically calling this, wouldn't we want to noop here rather than throw? Actually looking at SecretManager, I don't see PurgeOldSecretsAsync being called anymore - can you confirm? Is all this purge logic dead?

src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
```diff
@@ -0,0 +1,131 @@
...
+
+        foreach (var file in files)
+        {
+            secrets.Add(Path.GetFileName(file), await FileUtility.ReadAsync(file));
+        }
+
+        return secrets;
+    }
+
+    public Task WriteSnapshotAsync(ScriptSecretsType type, string functionName, ScriptSecrets secrets)
```
mathewc: So the write/get snapshot logic can't be triggered for this provider because it's only initiated by SM when reading of secrets results in a crypto exception attempting to deserialize the raw secrets, and the Read method above never does that?

Don't we want the files stored at ContainerAppsSecretsDir to be encrypted as we do in other providers?

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
```diff
@@ -740,7 +739,7 @@ private HostSecrets ReadHostSecrets(HostSecrets hostSecrets)
         {
             return new HostSecrets
             {
-                MasterKey = _keyValueConverterFactory.ReadKey(hostSecrets.MasterKey),
+                MasterKey = hostSecrets.MasterKey is null ? null : _keyValueConverterFactory.ReadKey(hostSecrets.MasterKey),
```
mathewc, line 742: It sucks that we're having to loosen our validity checks for this - it should never be the case that we're running the host with a null master key. Is there a way we can prevent that? E.g. make this a startup failure somehow on container apps scenarios if the host tries to run without requisite keys being present?

src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
```diff
@@ -0,0 +1,131 @@
...
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost;
+
+public class ContainerAppsSecretsRepository : ISecretsRepository
+{
+    internal const string ContainerAppsSecretsDir = "/run/secrets/functions-keys";
```
mathewc, line 16: Who will be creating these keys, and how will ensure that the values are "highly identifiable secrets" using the right seeds?

src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
brettsam: Ah yes. good catch. I had slightly modified how the K8S secret repo does this and didn't fix that part.

src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
brettsam: I actually had the plan to write out diagnostics and never did. Now I log out the files we read, which should be enough to debug scenarios without getting too verbose.

src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
brettsam: I don't think it is even being called anywhere. If you even look how it works, it's looking at file paths to determine current functions, which doesn't work at all anymore with worker indexing.

That being said, I agree the safe thing here is to no-op. I was trying to throw where possible to ensure that we could catch any attempt to *create* new keys -- which could lead to confusion in the SecretManager. But this doesn't fall into that category anyway so no-op makes more sense.

I'll make this no-op and have filed a follow-up to remove this Purge behavior completely: https://github.com/Azure/azure-functions-host/issues/11207

mathewc: Thanks for logging the cleanup work item

src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
brettsam: Encryption is not required here. The secrets are encrypted at-rest via the ACA Secrets feature, then mounted when the container starts. Similar behavior in our Kubernetes secrets. I've added comments to the code now to match the K8S secrets so it's more clear.

src/WebJobs.Script.WebHost/Security/KeyManagement/SecretManager.cs
brettsam: We discussed this in a separate chat, but re-iterating here for posterity. The host should be able to run with zero keys. If there's no need for /admin API calls (which the ACA RP doesn't natively support) nor any non-Anonymous HTTP triggers, the other Triggers can start and run without any key support. If you need any of those features, then yes, you need to add keys, which we will document. But starting up w/o keys and running anonymous HTTP triggers or any other triggers should be supported here while the key-generation story isn't quite ready.

This null-check was the only thing preventing that.

src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
brettsam: As of right now, it's up to customer to create them themselves via the ACA Secret feature (with docs). We're working with both the ACA control plane folks as well as tooling (Aspire, azd) to make sure we have flows covered where customers can at least be running with some minimal keys set up (like a `host.function.default`, for example) so they're not stuck upon first run with a non-anonymous http function. That's all being worked out separately and will follow the HIS guidelines.

This change is the first step to break us free from the default blob secret support and makes it a bit more usable in the ACA space -- which is hard to use without any RP support. You have to actually go look inside your blobs yourself to find your keys, changes aren't propagated to running instances, etc, etc.

src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
```diff
@@ -0,0 +1,145 @@
...
+        if (type == ScriptSecretsType.Function && string.IsNullOrEmpty(functionName))
+        {
+            throw new ArgumentNullException(nameof(functionName), $"{nameof(functionName)} cannot be null or empty with {nameof(type)} = {nameof(ScriptSecretsType.Function)}");
+        }
+
+        return type == ScriptSecretsType.Host ? await ReadHostSecretsAsync() : await ReadFunctionSecretsAsync(functionName.ToLowerInvariant());
+    }
+
+    public Task WriteAsync(ScriptSecretsType type, string functionName, ScriptSecrets secrets)
+        => throw new NotImplementedException();
```
jviau: Should this be `NotSupportedException` with a message saying this is a read-only repository?

src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
```diff
@@ -0,0 +1,145 @@
...
+
+    public bool IsEncryptionSupported => false;
+
+    public string Name => nameof(ContainerAppsSecretsRepository);
+
+    public async Task<ScriptSecrets> ReadAsync(ScriptSecretsType type, string functionName)
+    {
+        if (type == ScriptSecretsType.Function && string.IsNullOrEmpty(functionName))
+        {
+            throw new ArgumentNullException(nameof(functionName), $"{nameof(functionName)} cannot be null or empty with {nameof(type)} = {nameof(ScriptSecretsType.Function)}");
```
jviau: very minor nit: _technically_ you throw `ArgumentNullException` when `null` and `ArgumentException` when empty. 

There is a helper for that: `ArgumentException.ThrowIfNullOrEmpty`

``` CSharp
if (type == ScriptSecretsType.Function)
{
    ArgumentException.ThrowIfNullOrEmpty(functionName);
}
```

src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
```diff
@@ -0,0 +1,145 @@
...
+
+        return type == ScriptSecretsType.Host ? await ReadHostSecretsAsync() : await ReadFunctionSecretsAsync(functionName.ToLowerInvariant());
+    }
+
+    public Task WriteAsync(ScriptSecretsType type, string functionName, ScriptSecrets secrets)
+        => throw new NotImplementedException();
+
+    private async Task<ScriptSecrets> ReadHostSecretsAsync()
+    {
+        var secrets = await GetFromFilesAsync(ContainerAppsSecretsDir);
```
jviau: nit: prefer actual types over `var`

src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
```diff
@@ -0,0 +1,145 @@
...
+
+        HostSecrets hostSecrets = new HostSecrets()
```
jviau: nit: can be made more concise
```suggestion
        HostSecrets hostSecrets = new()
```

src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
```diff
@@ -0,0 +1,145 @@
...
+        };
+
+        // Always return a FunctionSecrets object, even if empty. This will prevent the SecretManager from thinking
+        // it needs to create and persist new secrets, which is not supported in Container Apps.
+        return functionSecrets;
+    }
+
+    private async Task<IDictionary<string, string>> GetFromFilesAsync(string path)
+    {
+        string[] files = await FileUtility.GetFilesAsync(path, "*");
```
jviau, line 110: For testability, can you inject a `IFileSystem`? Not sure if we have that type in the DI container. But would be helpful to add it if not there.

This way you can mock the file system and not need to perform any actual I/O in tests.

src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
```diff
@@ -0,0 +1,145 @@
...
+
+        StringBuilder sb = new StringBuilder("Loaded secrets from files:");
+
+        foreach (var file in files)
+        {
+            secrets.Add(Path.GetFileName(file), await FileUtility.ReadAsync(file));
+            sb.AppendLine($"  {file}");
+        }
+
+        _logger.LogDebug(sb.ToString());
```
jviau: nit: not sure if we have a performance concern here, but you could do a `_logger.IsEnabled(LogLevel.Debug)` check and no-op all logging if it is false. This will avoid the string builder allocation.

src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
```diff
@@ -0,0 +1,145 @@
...
+        return secrets;
+    }
+
+    /// <summary>
+    /// no-op - allow stale secrets to remain.
+    /// </summary>
+    public async Task PurgeOldSecretsAsync(IList<string> currentFunctions, ILogger logger)
+        => await Task.Yield();
```
jviau: nit: remove `async` and return `Task.CompletedTask`.

src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
```diff
@@ -0,0 +1,145 @@
...
+
+    /// <summary>
+    /// Runtime is not responsible for encryption so this code will never be executed.
+    /// </summary>
+    public Task WriteSnapshotAsync(ScriptSecretsType type, string functionName, ScriptSecrets secrets)
+        => throw new NotSupportedException();
```
jviau: Can we add a message explaining this is a read-only repository?

src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
fabiocav: This would be a platform expectation. When provisioning an app with the right app kind, the keys should be created. If they're not in place, there's still an experience to manage those in ACA, which is an improvement over what's available now.

src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
fabiocav: Can we update the `GetFromFilesAsync` method so we can pass a filter here that flows into the call to filter the files we operate on to limit to `host.*`  files and reduce IO here?

src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
```diff
@@ -0,0 +1,145 @@
...
+
+        HostSecrets hostSecrets = new HostSecrets()
+        {
+            FunctionKeys = [],
+            SystemKeys = []
+        };
+
+        foreach (var pair in secrets)
+        {
+            if (pair.Key.StartsWith(MasterKey, StringComparison.OrdinalIgnoreCase))
```
fabiocav: Our file name comparisons (particularly given the fact is will be on Linux) should be case sensitive)

fabiocav: I'm assuming we're using `StartsWith` here instead of `Equals` to deal with the fact the keys currently have an extension. Can we instead load the secretes without the file extension and use `Equals` to avoid false positives?

src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
```diff
@@ -0,0 +1,145 @@
...
+        }
+
+        // Always return a HostSecrets object, even if empty. This will prevent the SecretManager from thinking
+        // it needs to create and persist new secrets, which is not supported in Container Apps.
+        return hostSecrets;
+    }
+
+    private async Task<ScriptSecrets> ReadFunctionSecretsAsync(string functionName)
+    {
+        var secrets = await GetFromFilesAsync(ContainerAppsSecretsDir);
```
fabiocav, line 91: Same comment here about passing the prefix so we can limit this operation to reading function key files only.

src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
fabiocav: I'm curious to understand why @brettsam wanted to use `Yield` here, though, as simply changing to returning `Task.CompletedTask` will not have the same behavior.

src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
```diff
@@ -0,0 +1,145 @@
...
+    private async Task<IDictionary<string, string>> GetFromFilesAsync(string path)
+    {
+        string[] files = await FileUtility.GetFilesAsync(path, "*");
+        var secrets = new Dictionary<string, string>(files.Length);
+
+        StringBuilder sb = new StringBuilder("Loaded secrets from files:");
+
+        foreach (var file in files)
+        {
+            secrets.Add(Path.GetFileName(file), await FileUtility.ReadAsync(file));
```
fabiocav, line 117: Can we perform this enumeration/file reads in parallel (kick off the reads, and WaitAll on the tasks so we can have this IO done in parallel instead?

src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
jviau: Oh, if `Yield` is needed can still remove the `async` portion.

brettsam: Copied a bunch of this from the existing `KubernetesSecretsRepository` and then started modifying it for this new scenario. I'm not sure why that one is using `await Task.Yield();` -- I didn't even notice it. It's been this way since 2019 there :-) 

I'll update this to `Task.CompeltedTask`

src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
brettsam: Another copy/paste from K8S repository. Will update.

src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
brettsam: This mocks behind-the-scenes with `FileUtility.Instance`. It's a little weirdly indirect, but I'm following the pattern that's been established.

src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
brettsam: Yeah good call on the casing. I'm not sure why I added that (maybe copilot told me to?) -- K8S secret repository doesn't ignore case and I was trying to match that logic as much as possible so I must have done something weird.

For `StartsWith()` -- there are no file extensions. For master key, I can use an exact match, but for others it truly is a prefix (like `host.function.`). This logic came straight from how it already works in kubernetes so I didn't want to waiver too far from that behavior.

src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
brettsam: I'm going to punt on this one. I'm copying this logic straight from Kubernetes repo and this path should be restricted to one that only has our secret files in it anyway. Not expecting much perf impact here, but we can re-evaluate.

src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
brettsam: Copied logic from Kubernetes repository. We can come back to fix it up if we need to, but going to punt for now.

src/WebJobs.Script.WebHost/Security/KeyManagement/ContainerAppsSecretsRepository.cs
brettsam: Copied from Kubernetes repository. Directory should only contain our files if mounted correctly. Can come back to fix up later if needed.

test/WebJobs.Script.Tests.Integration/Host/SecretsRepositoryTests.cs
```diff
@@ -614,14 +614,19 @@ public ISecretsRepository GetNewSecretRepository()
...
+            public async Task DisposeAsync()
             {
                 try
                 {
                     // delete blob files
-                    ClearAllBlobSecrets().ContinueWith(t => { });
+                    await ClearAllBlobSecrets();
                     ClearAllFileSecrets();
-                    ClearAllKeyVaultSecrets().ContinueWith(t => { });
+                    await ClearAllKeyVaultSecrets();
```
brettsam, line 629: Bumped into a potential issue where it didn't seem like these secrets were getting cleaned up and some subsequent test failed. Wasn't able to repro but thought I'd make this more deterministic while I was in here.

fabiocav (approved):

mattchenderson (approved):

PR #11210, opened 2025-07-28
Stop retrying http requests when script invocation result is failed

### LINKED ISSUE #11237: RetryProxyHandler should stop retrying when invocation result fails
Sub-problem identified while investigating https://github.com/Azure/azure-functions-host/issues/10936. When `ScriptInvocationResult` fails, we should stop trying to proxy the HttpRequest tied to the invocation.

Example of this repro in https://github.com/Azure/azure-functions-host/issues/10580: 

```
[2024-10-30T03:18:47.651Z] A function timeout has occurred. Restarting worker process executing invocationId '6d2f2ee4-d1d4-4910-9821-b48986e8e3f9'.
[2024-10-30T03:18:48.032Z] 0: 32 (PID: 7404)
[2024-10-30T03:18:48.213Z] Worker process started and initialized.
[2024-10-30T03:18:48.579Z] 1: 13 (PID: 7404)
[2024-10-30T03:18:49.033Z] 0: 33 (PID: 7404)
...
[2024-10-30T03:18:52.063Z] 0: 36 (PID: 7404)
[2024-10-30T03:18:52.617Z] 1: 17 (PID: 7404)
[2024-10-30T03:18:52.686Z] Failed to proxy request to the worker. Retrying in 50ms. Attempt 1 of 3.
[2024-10-30T03:18:52.686Z] Failed to proxy request to the worker. Retrying in 50ms. Attempt 1 of 3.
[2024-10-30T03:18:56.798Z] Failed to proxy request to the worker. Retrying in 100ms. Attempt 2 of 3.
[2024-10-30T03:18:56.798Z] Failed to proxy request to the worker. Retrying in 100ms. Attempt 2 of 3.
```

satvu:

resolves #11237

Breaking down #11159 and putting the HTTP changes in a separate PR. These changes make sure we stop retrying an http request if the correlated function invocation is failed by the host elsewhere (in cases of worker channel shutdown, host-tracked function timeout).

### Pull request checklist

### Additional Info

When this change is _included with the propagation of `FunctionTimeoutException` and related exceptions_ in #11159, this is the before/after (images may not be accurate w/ final result, this is just a wip example).  

Before:
image

After:

image

Additionally, here is an example of the exception and innerException we could expect from a faulted `resultSource` (timeout):

image


src/WebJobs.Script/Http/RetryProxyHandler.cs
```diff
@@ -30,11 +32,22 @@ public RetryProxyHandler(HttpMessageHandler innerHandler, ILogger logger)
...
+            }
+
             var currentDelay = InitialDelay;
             for (int attemptCount = 1; attemptCount <= MaxRetries; attemptCount++)
             {
                 try
                 {
+                    if (resultSource is not null && resultSource.Task.IsFaulted)
+                    {
+                        throw resultSource.Task.Exception?.InnerException ?? new HttpRequestException("The function invocation tied to this HTTP request failed.");
```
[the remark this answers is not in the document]
satvu: Cancellation will be propagated to the HttpRequest and handled later in this file. I think we should leave the cancellation flow as is, and not add extra handling based on the script invocation result TCS.

Edit: [Here](https://github.com/Azure/azure-functions-host/blob/90d4ccf9f320e4f5c01238ba432727536fafab97/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs#L886) seems to be the only location this tcs is set to canceled, and this is a path that would lead to cancellation flowing to the worker and http request.

src/WebJobs.Script/Http/ScriptInvocationRequestTransformer.cs
```diff
@@ -0,0 +1,28 @@
...
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Azure.WebJobs.Script.Description;
+using Yarp.ReverseProxy.Forwarder;
+
+namespace Microsoft.Azure.WebJobs.Script.Http
+{
+    internal class ScriptInvocationRequestTransformer : HttpTransformer
```
kshyju: Can this be `sealed`?

src/WebJobs.Script/Http/ScriptInvocationRequestTransformer.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        public override async ValueTask TransformRequestAsync(HttpContext httpContext, HttpRequestMessage proxyRequest, string destinationPrefix, CancellationToken cancellationToken)
+        {
+            // this preserves previous behavior (which called the default transformer) - base method is also called inside of here
+            await HttpTransformer.Default.TransformRequestAsync(httpContext, proxyRequest, destinationPrefix, cancellationToken);
```
kshyju, line 23: I feel calling `await base.TransformRequestAsync`  instead here provides more clarity. you can also remove the comment.

src/WebJobs.Script/Http/RetryProxyHandler.cs
```diff
@@ -30,11 +32,23 @@ public RetryProxyHandler(HttpMessageHandler innerHandler, ILogger logger)
...
+            }
+
             var currentDelay = InitialDelay;
             for (int attemptCount = 1; attemptCount <= MaxRetries; attemptCount++)
             {
                 try
                 {
+                    if (resultSource is not null && resultSource.Task.IsFaulted)
+                    {
+                        throw resultSource.Task.Exception?.InnerException ??
```
kshyju: If `IsFaulted` is true, then the `Exception` will be not null. Should we use the overload of `HttpRequestException` which takes the innerexception and you can pass this exception to that. That way, it will have all the exceptions details(and inner exception) and callers can use it however they want.

```
 throw new HttpRequestException($"The...", resultSource.Task.Exception);
 ```

File: src/WebJobs.Script/Http/RetryProxyHandler.cs [hunk 5]
```diff
@@ -30,11 +32,23 @@ public RetryProxyHandler(HttpMessageHandler innerHandler, ILogger logger)
...
+                            new HttpRequestException($"The function invocation tied to this HTTP request failed. Invocation ID: {scriptInvocationContext.ExecutionContext.InvocationId}");
+                    }
+
                     return await base.SendAsync(request, cancellationToken);
                 }
                 catch (TaskCanceledException) when (cancellationToken.IsCancellationRequested)
```
kshyju, line 53: Also can you double check line 73 `LogWarning` method call is using the correct overload?  I think the exception should be the first param in that overload.

test/WebJobs.Script.Tests/HttpProxyService/ScriptInvocationRequestTransformerTests.cs
```diff
@@ -0,0 +1,135 @@
...
+            Assert.Contains(requestRemoteAddress, forwardedFor);
+
+            var forwardedHost = proxyRequest.Headers.GetValues("X-Forwarded-Host");
+            Assert.Contains(originalHost, forwardedHost);
+
+            var forwardedProto = proxyRequest.Headers.GetValues("X-Forwarded-Proto");
+            Assert.Contains(originalProto, forwardedProto);
+        }
+    }
+}
```
kshyju, line 199: Can you add/update one of these tests to include some other header other than the "X-Forwarded" ones? (ex: something the customer is sending in their request (Ex: "foo-bar") and ensure they remain present?

src/WebJobs.Script/Http/RetryProxyHandler.cs
```diff
@@ -30,11 +32,23 @@ public RetryProxyHandler(HttpMessageHandler innerHandler, ILogger logger)
...
+            {
+                resultSource = scriptInvocationContext.ResultSource;
+            }
+
             var currentDelay = InitialDelay;
             for (int attemptCount = 1; attemptCount <= MaxRetries; attemptCount++)
             {
                 try
                 {
+                    if (resultSource is not null && resultSource.Task.IsFaulted)
```
kshyju: Shouldn't we update RetryProxyHandlerTests to reflect this new code changes?

src/WebJobs.Script/Http/ScriptInvocationRequestTransformer.cs
satvu: It has to be `HttpTransformer.Default.TransformRequestAsync` - the base method does not add the `X-Forwarded` headers. I can remove the comment since tests should be able to catch if this is changed, but I wanted to make sure it was clear that choosing this method was intentional.

src/WebJobs.Script/Http/RetryProxyHandler.cs
satvu: If you mean [this line](https://github.com/Azure/azure-functions-host/blob/c62778b79e53786dce2852a39b9c1c73690be651/src/WebJobs.Script/Http/RetryProxyHandler.cs#L60), it is set up and logged correctly (additionally tested e2e).

src/WebJobs.Script/Http/RetryProxyHandler.cs
```diff
@@ -30,11 +32,22 @@ public RetryProxyHandler(HttpMessageHandler innerHandler, ILogger logger)
...
+                    {
+                        throw new HttpRequestException($"The function invocation tied to this HTTP request failed. Invocation ID: {scriptInvocationContext.ExecutionContext.InvocationId}", resultSource.Task.Exception);
```
kshyju: In this case, we should not be retrying. Correct? But looks like the exception thrown here will be handled by `catch (HttpRequestException) when (attemptCount < MaxRetries)` and that causes the retries?

src/WebJobs.Script/Http/RetryProxyHandler.cs
```diff
@@ -30,11 +32,23 @@ public RetryProxyHandler(HttpMessageHandler innerHandler, ILogger logger)
...
+
             var currentDelay = InitialDelay;
             for (int attemptCount = 1; attemptCount <= MaxRetries; attemptCount++)
             {
                 try
                 {
+                    if (resultSource is not null && resultSource.Task.IsFaulted)
+                    {
+                        throw resultSource.Task.Exception.InnerException ??
+                              new Exception($"The function invocation tied to this HTTP request failed. Invocation ID: {scriptInvocationContext.ExecutionContext.InvocationId}");
```
satvu: Still undecided on what exception to throw here - we don't want to throw something that will cause retries (see below lines), but what makes the most sense (some cancellation-related exception? do we change the check for HttpRequestException below so we don't retry in some cases?

satvu: Chose `HttpRequestException` and modified the `catch` blocks to separate out cases where the resultSource task is faulted.

src/WebJobs.Script/Http/ScriptInvocationRequestTransformer.cs
```diff
@@ -0,0 +1,28 @@
...
+
+            if (httpContext.Items.TryGetValue(ScriptConstants.HttpProxyScriptInvocationContext, out object result)
+                && result is ScriptInvocationContext scriptContext)
+            {
+                proxyRequest.Options.TryAdd(ScriptConstants.HttpProxyScriptInvocationContext, scriptContext);
+            }
+        }
+    }
+}
```
kshyju: Add a line break at the end of the file.

src/WebJobs.Script/Http/RetryProxyHandler.cs
```diff
@@ -30,18 +32,35 @@ public RetryProxyHandler(HttpMessageHandler innerHandler, ILogger logger)
...
+
             var currentDelay = InitialDelay;
             for (int attemptCount = 1; attemptCount <= MaxRetries; attemptCount++)
             {
                 try
                 {
+                    if (resultSource is not null && resultSource.Task.IsFaulted)
+                    {
+                        throw resultSource.Task.Exception.InnerException ??
+                              new HttpRequestException($"The function invocation tied to this HTTP request failed. Invocation ID: {scriptInvocationContext.ExecutionContext.InvocationId}");
```
kshyju: Is it a valid use case to have `resultSource.Task.IsFaulted` true and `resultSource.Task.Exception.InnerException` to be `null`? In that case, we are going to retry. Is that expected?

src/WebJobs.Script/Http/DefaultHttpProxyService.cs
```diff
@@ -21,6 +21,7 @@ internal class DefaultHttpProxyService : IHttpProxyService, IDisposable
         private readonly HttpMessageInvoker _messageInvoker;
         private readonly ForwarderRequestConfig _forwarderRequestConfig;
         private readonly ILogger<DefaultHttpProxyService> _logger;
+        private readonly HttpTransformer _httpTransformer;
```
jviau: nit: you can inline this.

```suggestion
        private readonly HttpTransformer _httpTransformer = new ScriptInvocationRequestTransformer();
```

src/WebJobs.Script/Http/RetryProxyHandler.cs
```diff
@@ -30,18 +32,35 @@ public RetryProxyHandler(HttpMessageHandler innerHandler, ILogger logger)
...
+            }
+
             var currentDelay = InitialDelay;
             for (int attemptCount = 1; attemptCount <= MaxRetries; attemptCount++)
             {
                 try
                 {
+                    if (resultSource is not null && resultSource.Task.IsFaulted)
+                    {
+                        throw resultSource.Task.Exception.InnerException ??
```
jviau: This will reset the stack trace of the inner exception. Is that desired? I am not entirely sure what the inner exception will be here. But if you want to preserve it you can use `ExceptionDispatchInfo.Capture(ex).Throw();`

src/WebJobs.Script/Http/RetryProxyHandler.cs
```diff
@@ -30,18 +32,35 @@ public RetryProxyHandler(HttpMessageHandler innerHandler, ILogger logger)
...
+            {
+                resultSource = scriptInvocationContext.ResultSource;
+            }
+
             var currentDelay = InitialDelay;
             for (int attemptCount = 1; attemptCount <= MaxRetries; attemptCount++)
             {
                 try
                 {
+                    if (resultSource is not null && resultSource.Task.IsFaulted)
```
jviau: I wonder if this logic also needs to handle `Task.IsCanceled`? You can also use a check of `!Task.IsCompletedSuccessfully`

src/WebJobs.Script/Http/ScriptInvocationRequestTransformer.cs
```diff
@@ -0,0 +1,28 @@
...
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Azure.WebJobs.Script.Description;
+using Yarp.ReverseProxy.Forwarder;
+
+namespace Microsoft.Azure.WebJobs.Script.Http
+{
+    internal sealed class ScriptInvocationRequestTransformer : HttpTransformer
```
jviau, line 14: Since this has no state, consider have a singleton you can re-use. Much like how there is `HttpTransformer.Default`

src/WebJobs.Script/Http/ScriptInvocationRequestTransformer.cs
jviau: @kshyju for context, `HttpTransformer.Default` is a derived instance, and not just a `HttpTransformer` instance.

https://github.com/dotnet/yarp/blob/main/src/ReverseProxy/Forwarder/HttpTransformer.cs#L26

jviau (changes requested):
@satvu can you update your PR description with a breakdown of how the exceptions will propagate with these changes? Examples of what the `InnerException` you are re-throwing will be helpful. Also address @kshyju comment about us having a faulted task, but a null inner exception. Is that possible?

src/WebJobs.Script/Http/RetryProxyHandler.cs
satvu: No - fixing this.

src/WebJobs.Script/Http/RetryProxyHandler.cs
satvu: I added handling for it - we already have some handling in this file related to the observation of the cancellation token, but I think handling `Task.IsCanceled` here would have the same end result (goes through `catch (TaskCanceledException) when (cancellationToken.IsCancellationRequested)` and stop retries or go into the last block and stop retries).

src/WebJobs.Script/Http/RetryProxyHandler.cs
satvu: Added an example in the description.

src/WebJobs.Script/Http/RetryProxyHandler.cs
```diff
@@ -30,20 +32,37 @@ public RetryProxyHandler(HttpMessageHandler innerHandler, ILogger logger)
...
                 }
                 catch (TaskCanceledException) when (cancellationToken.IsCancellationRequested)
                 {
                     _logger.LogDebug("Request was canceled. Stopping retries.");
                     throw new OperationCanceledException(cancellationToken);
                 }
-                catch (HttpRequestException) when (attemptCount < MaxRetries)
+                catch (HttpRequestException exception) when (attemptCount < MaxRetries)
                 {
+                    if (resultSource is not null && (resultSource.Task.IsFaulted || resultSource.Task.IsCanceled))
```
satvu, line 60: If the reason for the `resultSource` faulting is not an `HttpRequestException`, this check here can be removed so we just wait until the next retry to check/handle `resultSource` faulting. 

I don't know if there's any cases where the reason for failure is `HttpRequestException`, but if there is, we probably want to add a stop here too.

Also, had a placeholder here in line 63, but unsure which error we want to propagate here (`HttpRequestException` or the function invocation `resultSource` exception.

src/WebJobs.Script/Http/RetryProxyHandler.cs
```diff
@@ -30,20 +32,37 @@ public RetryProxyHandler(HttpMessageHandler innerHandler, ILogger logger)
...
+                    {
+                        _logger.LogWarning("HTTP requests will not be retried. The associated function invocation has failed.");
+                        ExceptionDispatchInfo.Capture(exception).Throw();
```
jviau:
```suggestion
                        throw;
```

A simple `throw;` statement is a shorthand to do exactly this from within a `catch` block. It will rethrow the caught exception without resetting the stack trace.

jviau (approved):

src/WebJobs.Script/Http/RetryProxyHandler.cs
```diff
@@ -55,7 +74,7 @@ protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage
                 {
                     var message = attemptCount == MaxRetries
                         ? "Reached the maximum retry count for worker request proxying. Error: {exception}"
-                        : $"Unsupported exception type in {nameof(RetryProxyHandler)}. Request will not be retried. Exception: {{exception}}";
+                        : $"HTTP requests will not be retried. Excpetion in {nameof(RetryProxyHandler)}: {{exception}}.";
```
liliankasem: request**s**? Isn't this a http retry handler?

```suggestion
                        : $"HTTP request will not be retried. Excpetion in {nameof(RetryProxyHandler)}: {{exception}}.";
```

liliankasem: Same on line 62?

src/WebJobs.Script/Http/RetryProxyHandler.cs
liliankasem: If you're in this catch block, then isn't the reason alrady determined as `HttpRequestException` given that you're throwing it on line 48? 

Or maybe I don't understand the question?

src/WebJobs.Script/Http/RetryProxyHandler.cs
satvu: Fixed!

src/WebJobs.Script/Http/RetryProxyHandler.cs
satvu: If:

```
  if (resultSource is not null && (resultSource.Task.IsFaulted || resultSource.Task.IsCanceled))
  {
      ExceptionDispatchInfo.Capture(resultSource.Task.Exception).Throw();
  }
```

throws an `HttpRequestException`, and this block is entered, there needs to be another check or a retry will be kicked off. Not sure how often this actually happens, but just in case.

liliankasem (approved):

PR #11229, opened 2025-08-01
Refactor code to move the current logic to search for WorkerConfigs to a default worker configuration resolver

### LINKED ISSUE #11218: Refactor code to move the current logic to search for WorkerConfigs to a default worker configuration resolver
Refactor code to move the current logic to search for WorkerConfigs to a default worker configuration resolver. The current logic finds workerConfigs by scanning workers directory within the Host. This resolver should use Options pattern to get the required input values.

These changes should make it easier to add a new Resolver in future which will resolve workerConfigs from specified worker probing paths.

surgupta-msft:

resolves https://github.com/Azure/azure-functions-host/issues/11218

### Pull request checklist

### Additional information

Additional PR information

Refactor code to move the current logic to search for WorkerConfigs to a default worker configuration resolver. The current logic finds workerConfigs by scanning workers directory within the Host. This resolver should use Options pattern to get the required input values.

These changes should make it easier to add a new Resolver in future which will resolve workerConfigs from specified worker probing paths.


src/WebJobs.Script/Workers/Rpc/IWorkerConfigurationResolver.cs
```diff
@@ -0,0 +1,15 @@
...
+namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
+{
+    /// <summary>
+    /// Interface to resolve the Worker Configs.
+    /// </summary>
+    internal interface IWorkerConfigurationResolver
+    {
+        List<string> GetWorkerConfigPaths();
+    }
+}
```
kshyju: Add a line break at the end of this file.

src/WebJobs.Script/Workers/Rpc/Configuration/WorkerConfigurationResolverOptionsSetup.cs
```diff
@@ -0,0 +1,77 @@
...
+            {
+                // Site Extension Path. Default to parent directory.
+                workersDirPath = Path.Combine(Directory.GetParent(assemblyDir).FullName, RpcWorkerConstants.DefaultWorkersDirectoryName);
+            }
+            return workersDirPath;
+        }
+
+        internal static string GetWorkersDirPath(IConfiguration configuration)
+        {
+            string workersDirPath = GetDefaultWorkersDirectory(Directory.Exists);
```
kshyju: Optimization: The current flow always calls `GetDefaultWorkersDirectory()`, which triggers file system access even when a custom path is set in IConfiguration. Suggest checking config first and falling back to the default only if it's missing.

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
```diff
@@ -228,12 +231,20 @@ public static void AddWebJobsScriptHost(this IServiceCollection services, IConfi
...
 
             services.TryAddSingleton<IDependencyValidator, DependencyValidator>();
             services.TryAddSingleton<IJobHostMiddlewarePipeline>(s => DefaultMiddlewarePipeline.Empty);
 
+            services.AddSingleton<IWorkerConfigurationResolver>(p =>
+            {
+                var workerConfigurationResolverOptions = p.GetService<IOptionsMonitor<WorkerConfigurationResolverOptions>>();
+                var loggerFactory = p.GetService<ILoggerFactory>();
+                return new DefaultWorkerConfigurationResolver(loggerFactory, workerConfigurationResolverOptions);
+            });
```
surgupta-msft: The DI of `WorkerConfigurationResolverOptionsSetup` has been done similar to `LanguageWorkerOptionsSetup` because the ultimate purpose of both is to get WorkerConfigs and `LanguageWorkerOptionsSetup` uses `WorkerConfigurationResolverOptions` as input.

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -378,6 +382,11 @@ private async Task UnsynchronizedStartHostAsync(ScriptHostStartupOperation activ
                     }
                 }
 
+                if (_workerConfigResolverOptionsChangeTokenSource is HostBuiltChangeTokenSource<WorkerConfigurationResolverOptions> { } hostBuiltChangeToken)
+                {
+                    hostBuiltChangeToken.TriggerChange();
+                }
```
surgupta-msft, line 388: Lines 385-388 are implemented similarly to lines 390-393

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -338,14 +339,19 @@ public static IHostBuilder AddScriptHostCore(this IHostBuilder builder, ScriptAp
 
                 if (applicationHostOptions.HasParentScope)
                 {
-                    // Forward the host LanguageWorkerOptions to the Job Host.
+                    // Forward the host WorkerConfigurationResolverOptions and LanguageWorkerOptions to the Job Host.
+                    var workerResolverOptions = applicationHostOptions.RootServiceProvider.GetService<IOptionsMonitor<WorkerConfigurationResolverOptions>>();
+                    services.AddSingleton(workerResolverOptions);
+                    services.AddSingleton<IOptions<WorkerConfigurationResolverOptions>>(s => new OptionsWrapper<WorkerConfigurationResolverOptions>(workerResolverOptions.CurrentValue));
+
```
surgupta-msft: Lines 342-345 are implemented similarly to lines 347-350

src/WebJobs.Script/Workers/Rpc/Configuration/DefaultWorkerConfigurationResolver.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        public DefaultWorkerConfigurationResolver(ILoggerFactory loggerFactory, IOptionsMonitor<WorkerConfigurationResolverOptions> workerConfigurationResolverOptions)
+        {
+            ArgumentNullException.ThrowIfNull(loggerFactory);
+            _logger = loggerFactory.CreateLogger(ScriptConstants.LogCategoryWorkerConfig);
+            _workerConfigurationResolverOptions = workerConfigurationResolverOptions ?? throw new ArgumentNullException(nameof(workerConfigurationResolverOptions));
+        }
+
+        public List<string> GetWorkerConfigPaths()
+        {
```
surgupta-msft, line 31: Code in this method is taken from `AddProviders()` method of `RPCWorkerConfigFactory`

src/WebJobs.Script/Workers/Rpc/Configuration/DefaultWorkerConfigurationResolver.cs
```diff
@@ -0,0 +1,44 @@
...
+            var workersDirPath = _workerConfigurationResolverOptions.CurrentValue.WorkersDirPath;
+            _logger.LogDebug("Workers Directory set to: {workersDirPath}", workersDirPath);
```
kshyju: We should consider switching to the source generated logging wherever applicable. This method is in the cold start path and is a good candidate.
 https://learn.microsoft.com/en-us/dotnet/core/extensions/logger-message-generator

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
```diff
@@ -47,14 +53,10 @@ public RpcWorkerConfigFactory(IConfiguration config,
...
+            _workerConfigurationResolverOptions = workerConfigurationResolverOptions ?? throw new ArgumentNullException(nameof(workerConfigurationResolverOptions));
 
-            WorkersDirPath = GetDefaultWorkersDirectory(Directory.Exists);
-            var workersDirectorySection = _config.GetSection($"{RpcWorkerConstants.LanguageWorkersSectionName}:{WorkerConstants.WorkersDirectorySectionName}");
-
-            if (!string.IsNullOrEmpty(workersDirectorySection.Value))
-            {
-                WorkersDirPath = workersDirectorySection.Value;
-            }
+            WorkersDirPath = _workerConfigurationResolverOptions.CurrentValue.WorkersDirPath;
```
surgupta-msft: Lines 51-57 and lines 71-84 moved to `WorkerConfigurationResolverOptionsSetup` class.

src/WebJobs.Script/Workers/Rpc/Configuration/DefaultWorkerConfigurationResolver.cs
```diff
@@ -0,0 +1,44 @@
...
+
+                if (File.Exists(workerConfigPath))
+                {
+                    workerConfigs.Add(workerDir);
+                }
+            }
+            return workerConfigs;
+        }
+    }
+}
```
kshyju: Check all files in the PR to ensure there is a line break at the end of the file.

src/WebJobs.Script/Workers/Rpc/Configuration/WorkerConfigurationResolverOptionsSetup.cs
```diff
@@ -0,0 +1,68 @@
...
+            if (_scriptHostManager is IServiceProvider scriptHostManagerServiceProvider)
+            {
+                var latestConfiguration = scriptHostManagerServiceProvider.GetService<IConfiguration>();
+                if (latestConfiguration is not null)
+                {
+                    configuration = new ConfigurationBuilder()
+                        .AddConfiguration(_configuration)
+                        .AddConfiguration(latestConfiguration)
+                        .Build();
+                }
```
surgupta-msft, line 59: Configuration is setup similar to `LanguageWorkerOptionsSetup` - https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Workers/Rpc/Configuration/LanguageWorkerOptionsSetup.cs#L63-L71

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
```diff
@@ -47,14 +53,10 @@ public RpcWorkerConfigFactory(IConfiguration config,
             _metricsLogger = metricsLogger ?? throw new ArgumentNullException(nameof(metricsLogger));
             _profileManager = workerProfileManager ?? throw new ArgumentNullException(nameof(workerProfileManager));
             _workerRuntime = _environment.GetEnvironmentVariable(RpcWorkerConstants.FunctionWorkerRuntimeSettingName);
+            _workerConfigurationResolver = workerConfigurationResolver ?? throw new ArgumentNullException(nameof(workerConfigurationResolver));
+            _workerConfigurationResolverOptions = workerConfigurationResolverOptions ?? throw new ArgumentNullException(nameof(workerConfigurationResolverOptions));
```
kshyju: This class currently depends on both the resolver and its options. Instead of injecting WorkerConfigurationResolverOptions directly, we should have IWorkerConfigurationResolver return an immutable type that exposes relevant config path info (e.g., base path, list of config paths). That way, we centralize the logic and keep it better encapsulated.

src/WebJobs.Script/Workers/Rpc/IWorkerConfigurationResolver.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
+{
+    /// <summary>
+    /// Interface to resolve the Worker Configs.
+    /// </summary>
+    internal interface IWorkerConfigurationResolver
+    {
+        List<string> GetWorkerConfigPaths();
+
+        IOptionsMonitor<WorkerConfigurationResolverOptions> GetWorkerConfigurationResolverOptions();
```
kshyju: I wouldd suggest not exposing `WorkerConfigurationResolverOptions` through the interface. It feels like an internal implementation detail specific to `DefaultWorkerConfigurationResolver` implementation.

Instead, we could define an **immutable** type like this to encapsulate the relevant data:
```csharp
internal sealed record WorkerConfigurationInfo
{
    public required string BasePath { get; init; }

    public required IReadOnlyList<string> ConfigPaths { get; init; }
}
```
and update the interface to return that

```
internal interface IWorkerConfigurationResolver
{
    WorkerConfigurationInfo GetConfiguration();
}
```    

This keeps the interface clean and focused on what it provides, not how it is implemented.

src/WebJobs.Script/Workers/Rpc/Configuration/DefaultWorkerConfigurationResolver.cs
kshyju: The previous implementation does not have a dependency on `IConfiguration`. Looks like the new code relies on `WorkerConfigurationResolverOptions` and `WorkerConfigurationResolverOptionsSetup` relies on `IConfiguration`. Is this additional change intentional? If so, can we add a comment explaining why?

https://github.com/Azure/azure-functions-host/blob/de373194d431f696220bf60ae4948c82b15e2d8e/src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs#L71-L83

surgupta-msft: The code in the constructor of `RPCWorkerConfigFactory` depends on `IConfiguration`. That code calls this method `GetDefaultWorkersDirectory()` to prepare `workersDirPath`.

kshyju: Got it. Do we know what scenario this is supporting? Who sets `languageWorkers:workersDirectory` or `languageWorkers__workersDirectory`, and in what use case?

src/WebJobs.Script/Workers/Rpc/Configuration/WorkerConfigurationResolverOptionsSetup.cs
```diff
@@ -0,0 +1,67 @@
...
+            _scriptHostManager = scriptHostManager ?? throw new ArgumentNullException(nameof(scriptHostManager));
+        }
+
+        public void Configure(WorkerConfigurationResolverOptions options)
+        {
+            var configuration = _configuration;
+            if (_scriptHostManager is IServiceProvider scriptHostManagerServiceProvider)
+            {
+                var latestConfiguration = scriptHostManagerServiceProvider.GetService<IConfiguration>();
+                if (latestConfiguration is not null)
```
kshyju: What use case requires the configuration rebuild here?

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
```diff
@@ -228,12 +231,20 @@ public static void AddWebJobsScriptHost(this IServiceCollection services, IConfi
...
             services.ConfigureOptionsWithChangeTokenSource<LanguageWorkerOptions, LanguageWorkerOptionsSetup, HostBuiltChangeTokenSource<LanguageWorkerOptions>>();
 
             services.TryAddSingleton<IDependencyValidator, DependencyValidator>();
             services.TryAddSingleton<IJobHostMiddlewarePipeline>(s => DefaultMiddlewarePipeline.Empty);
 
+            services.AddSingleton<IWorkerConfigurationResolver>(p =>
+            {
+                var workerConfigurationResolverOptions = p.GetService<IOptionsMonitor<WorkerConfigurationResolverOptions>>();
+                var loggerFactory = p.GetService<ILoggerFactory>();
+                return new DefaultWorkerConfigurationResolver(loggerFactory, workerConfigurationResolverOptions);
```
jviau: The manual factory here shouldn't be necessary. Both `ILoggerFactory` and `IOptionsMonitor<WorkerConfigurationResolverOptions>` are resolvable if you register directly:

``` CSharp
 services.AddSingleton<IWorkerConfigurationResolver, DefaultWorkerConfigurationResolver>();
```

src/WebJobs.Script/Workers/Rpc/Configuration/DefaultWorkerConfigurationResolver.cs
```diff
@@ -0,0 +1,53 @@
...
+        }
+
+        public List<string> GetWorkerConfigPaths()
+        {
+            var workersDirPath = _workerConfigurationResolverOptions.CurrentValue.WorkersDirPath;
+            _logger.DefaultWorkersDirectoryPath(workersDirPath);
+
+            List<string> workerConfigs = new();
+
+            foreach (var workerDir in Directory.EnumerateDirectories(workersDirPath))
```
jviau: Consider using file system abstractions here for better testability:

`IFileSystem`. Not sure if we have it in the DI container? Recommend testing it out.

src/WebJobs.Script/Workers/Rpc/Configuration/DefaultWorkerConfigurationResolver.cs
```diff
@@ -0,0 +1,53 @@
...
+        private readonly IOptionsMonitor<WorkerConfigurationResolverOptions> _workerConfigurationResolverOptions;
+
+        public DefaultWorkerConfigurationResolver(ILoggerFactory loggerFactory, IOptionsMonitor<WorkerConfigurationResolverOptions> workerConfigurationResolverOptions)
+        {
+            ArgumentNullException.ThrowIfNull(loggerFactory);
+            _logger = loggerFactory.CreateLogger(ScriptConstants.LogCategoryWorkerConfig);
+            _workerConfigurationResolverOptions = workerConfigurationResolverOptions ?? throw new ArgumentNullException(nameof(workerConfigurationResolverOptions));
+        }
+
+        public List<string> GetWorkerConfigPaths()
```
jviau: nit: can we change the contract and implementation to `IEnumerable<string>`? This way you can use `yield return`. If the caller needs a list, they can call `ToList()`

src/WebJobs.Script/Workers/Rpc/Configuration/WorkerConfigurationResolutionInfo.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc.Configuration
+{
+    internal sealed class WorkerConfigurationResolutionInfo
```
jviau: the way this is used in the PR, a struct may be better.

``` csharp
internal readonly struct WorkerConfigurationResolutionInfo(string workersDirPath)
{
    public string WorkersDirPath  => workersDirPath;
}
```

I recommend structs when the life of an instance is _mostly_ limited to the stack. This is when structs are most efficient generally.

Note: you could also do a record struct:

``` CSharp
internal record struct WorkerConfigurationResolutionInfo(string WorkersDirPath);
```

src/WebJobs.Script/Workers/Rpc/Configuration/WorkerConfigurationResolverOptions.cs
```diff
@@ -0,0 +1,25 @@
...
+
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Microsoft.Azure.WebJobs.Hosting;
+
+namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc.Configuration
+{
+    public sealed class WorkerConfigurationResolverOptions : IOptionsFormatter
+    {
+        // Gets or sets the workers directory path within the Host or defined by IConfiguration.
```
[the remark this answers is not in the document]
jviau: @surgupta-msft, please use XML doc when applicable.

src/WebJobs.Script/Workers/Rpc/Configuration/WorkerConfigurationResolverOptions.cs
```diff
@@ -0,0 +1,25 @@
...
+        public string WorkersDirPath { get; set; }
+
+        // Implements the Format method from IOptionsFormatter interface.
```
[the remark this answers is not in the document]
jviau: Same here, please use xmldoc. For interface and overridden methods you can use `/// <inheritdoc />` when the contract itself has xmldoc

src/WebJobs.Script/Workers/Rpc/Configuration/WorkerConfigurationResolverOptions.cs
```diff
@@ -0,0 +1,25 @@
...
+        public string Format()
+        {
+            return JsonSerializer.Serialize(this, typeof(WorkerConfigurationResolverOptions), ConfigResolverOptionsJsonSerializerContext.Default);
+        }
+    }
+
+    [JsonSourceGenerationOptions(WriteIndented = true)]
+    [JsonSerializable(typeof(WorkerConfigurationResolverOptions))]
+    internal partial class ConfigResolverOptionsJsonSerializerContext : JsonSerializerContext;
```
jviau, line 28: Since we only ever write this object, `GenerationMode = JsonSourceGenerationMode.Serialization` will optimize for that (but removes the deserialization ability, which we don't need)

src/WebJobs.Script/Workers/Rpc/Configuration/WorkerConfigurationResolverOptionsSetup.cs
jviau: Nothing consumes this configuration. Why is it being built? In `LanguageWorkerOptionsSetup` it is being passed to `RpcWorkerConfigFactory`, so it is actually used.

src/WebJobs.Script/Workers/Rpc/Configuration/WorkerConfigurationResolverOptionsSetup.cs
```diff
@@ -0,0 +1,67 @@
...
+        internal static string GetWorkersDirPath(IConfiguration configuration)
+        {
+            var workersDirectorySection = configuration?.GetSection($"{RpcWorkerConstants.LanguageWorkersSectionName}:{WorkerConstants.WorkersDirectorySectionName}");
+
+            if (!string.IsNullOrEmpty(workersDirectorySection?.Value))
+            {
+                return workersDirectorySection.Value;
+            }
+
+            return GetDefaultWorkersDirectory(Directory.Exists);
```
jviau: Consider using `IFileSystem` for better testability

src/WebJobs.Script/Workers/Rpc/Configuration/WorkerConfigurationResolverOptionsSetup.cs
surgupta-msft: line 38 in this class consumes this configuration -

`options.WorkersDirPath = GetWorkersDirPath(configuration);`

src/WebJobs.Script/Workers/Rpc/Configuration/WorkerConfigurationResolverOptionsSetup.cs
surgupta-msft: If I debug Host with a dotnet-isolated app, the `_configuration` has 108 sections. After adding `latestConfiguration`, the final `configuration` will have 124 sections. Some of these additional sections are also coming from dotnet worker - [Link](https://github.com/Azure/azure-functions-dotnet-worker/blob/fe00d7cf39fb9c635edc6a713b675689ead93b6f/sdk/FunctionMetadataLoaderExtension/Startup.cs#L28)

src/WebJobs.Script/Workers/Rpc/Configuration/DefaultWorkerConfigurationResolver.cs
surgupta-msft: Don't know the scenario on top of my mind. The code was added in 2018 - https://github.com/Azure/azure-functions-host/issues/2947. So, the ability to overwrite default workers dir path via configuration has been available for a long time.

src/WebJobs.Script/Workers/Rpc/Configuration/WorkerConfigurationResolverOptionsSetup.cs
kshyju: The code in this class specifically uses the `languageWorkers:workersDirectory` configuration entry. Is that entry only available after the configuration is rebuilt?  What use case is setting/adding that configuration entry? 

I think we should rebuild the configuration only if it is really needed.

surgupta-msft: Sharing my thought process on this. We have this ability to overwrite the default workers directory path via configuration available for a long time. Let's say there is a component/scenario which uses this (I asked Fabio too, we don't know exactly which scenario).

Before my changes, RPCWorkerConfigFactory always use the updated Configuration to find the WorkersDirPath.
After my changes, if I don't rebuild this configuration, I am breaking this behavior.

kshyju: @fabiocav I am wondering if we should use this opportunity to clean up any legacy code that was added early on but may no longer be needed/used. If we are unsure about the use case which requires this codes snippet, I would lean toward removing it in a safe manner. otherwise, it’s likely to stay indefinitely.

surgupta-msft: Addressed as per offline discussion. Instead of rebuilding the configuration, we will be now checking if `latestConfiguration` has our required config section. If so, we will just add that particular setting to the existing `configuration`.

src/WebJobs.Script/Workers/Rpc/Configuration/WorkerConfigurationResolverOptionsSetup.cs
```diff
@@ -0,0 +1,101 @@
...
+
+                string value = GetConfigurationSectionValue(latestConfiguration, nameof(latestConfiguration), requiredSection);
+
+                if (!string.IsNullOrEmpty(value))
+                {
+                    var inMemoryCollection = new Dictionary<string, string> { { requiredSection, value } };
+                    configuration = new ConfigurationBuilder()
+                                            .AddConfiguration(_configuration)
+                                            .AddInMemoryCollection(inMemoryCollection)
+                                            .Build();
```
kshyju: I thought we agreed that we don't need a rebuild of configuration based on our last sync since the plan was to check both configurations one by one (latest first).

src/WebJobs.Script/Workers/Rpc/Configuration/WorkerConfigurationResolutionInfo.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc.Configuration
+{
+    internal record struct WorkerConfigurationResolutionInfo(string WorkersDirectoryPath)
+    {
+        public readonly string WorkersDirPath => WorkersDirectoryPath;
```
kshyju: Consider adding the collection of worker directory paths also as a property to this type and keep a single method in the `IWorkerConfigurationResolver` interface as well to return that. It makes sense to encapsulate both piece of information to single immutable type.

src/WebJobs.Script/Workers/Rpc/Configuration/WorkerConfigurationResolverOptionsSetup.cs
surgupta-msft: My bad. Updated again.

src/WebJobs.Script/Workers/Rpc/Configuration/WorkerConfigurationResolutionInfo.cs
surgupta-msft: So, basically `GetWorkerConfigPaths()` in `IWorkerConfigurationResolver` uses the value provided by `WorkerConfigurationResolutionInfo` to prepare the output - `WorkerConfigPaths`.  Are you suggesting to move the output (`WorkerConfigPaths`) to `WorkerConfigurationResolutionInfo` ?  May be I did not understand the comment correctly?

Additionally, `RPCWorkerConfigFactory` also uses the value of `WorkerConfigurationResolutionInfo` to make certain decisions.

kshyju: yea, like what was suggested in this comment - https://github.com/Azure/azure-functions-host/pull/11229#discussion_r2252185114

surgupta-msft: If I take your example, `BasePath` will be used as input to determine the output `ConfigPaths`. Hence, they can be part of different structures but not the same structure. One is input to method `GetWorkerConfigPaths()` and another is output of this method.

```csharp
internal sealed record WorkerConfigurationInfo
{
    public required string BasePath { get; init; }

    public required IReadOnlyList<string> ConfigPaths { get; init; }
}
```

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
```diff
@@ -26,19 +27,22 @@ internal class RpcWorkerConfigFactory
...
         private readonly string _workerRuntime;
         private readonly IEnvironment _environment;
+        private readonly IWorkerConfigurationResolver _workerConfigurationResolver;
         private readonly JsonSerializerOptions _jsonSerializerOptions = new()
         {
             PropertyNameCaseInsensitive = true
         };
 
         private Dictionary<string, RpcWorkerConfig> _workerDescriptionDictionary = new Dictionary<string, RpcWorkerConfig>();
+        private WorkerConfigurationResolutionInfo _workerConfigurationResolutionInfo;
```
brettsam: Remove this. It's not really needed.

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
```diff
@@ -90,15 +71,18 @@ internal void BuildWorkerProviderDictionary()
...
-                {
-                    AddProvider(workerDir);
-                }
+                _logger.LogTrace("No worker configs found.");
+                return;
+            }
+
+            foreach (var workerConfig in workerConfigs)
+            {
+                AddProvider(workerConfig);
```
brettsam: Pass the root worker dir path into this method instead of storing it in a field.

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
```diff
@@ -122,13 +106,15 @@ internal void AddProvider(string workerDir)
             {
                 try
                 {
+                    string workersDirPath = _workerConfigurationResolutionInfo.WorkersDirPath;
```
brettsam: Remove this. It should be passed in instead of being a field.

src/WebJobs.Script/Workers/Rpc/Configuration/WorkerConfigurationResolutionInfo.cs
```diff
@@ -0,0 +1,14 @@
...
+
+using System.Collections.Generic;
+
+namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc.Configuration
+{
+    internal record struct WorkerConfigurationResolutionInfo(string WorkersDirectoryPath, IReadOnlyList<string> WorkerConfigPathsList)
+    {
+        public readonly string WorkersDirPath => WorkersDirectoryPath;
+
+        public readonly IReadOnlyList<string> WorkerConfigPaths => WorkerConfigPathsList;
```
brettsam: I'm pretty sure you don't need these properties. Records auto-generate properties based on the constructor params. This (note that I also removed `struct`) generates what we want, for example (init-only properties are auto-created):

```
internal record WorkerConfigurationResolutionInfo(string WorkersDirectoryPath, IReadOnlyList<string> WorkerConfigPathsList);
```

src/WebJobs.Script/Workers/Rpc/Configuration/WorkerConfigurationResolverOptions.cs
```diff
@@ -0,0 +1,29 @@
...
+using Microsoft.Azure.WebJobs.Hosting;
+
+namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc.Configuration
+{
+    public sealed class WorkerConfigurationResolverOptions : IOptionsFormatter
+    {
+        /// <summary>
+        /// Gets or sets the workers directory path within the Host or defined by IConfiguration.
+        /// </summary>
+        public string WorkersDirPath { get; set; }
```
brettsam: I think renaming this to `WorkersPathRoot` would be clearer for what this is giving us.

src/WebJobs.Script/Workers/Rpc/IWorkerConfigurationResolver.cs
```diff
@@ -0,0 +1,17 @@
...
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
+{
+    /// <summary>
+    /// Interface to resolve the Worker Configs.
+    /// </summary>
+    internal interface IWorkerConfigurationResolver
+    {
+        WorkerConfigurationResolutionInfo GetWorkerConfigPaths();
```
brettsam: Rename to `GetConfigurationInfo()`? -- also maybe drop the `Resolution` from that record to make it a little shorter?

src/WebJobs.Script/Workers/Rpc/Configuration/WorkerConfigurationResolutionInfo.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc.Configuration
+{
+    internal record struct WorkerConfigurationResolutionInfo(string WorkersDirectoryPath, IReadOnlyList<string> WorkerConfigPathsList)
```
brettsam: I'd name the first parameter here `WorkersPathRoot`. Secone one `WorkerConfigurationFiles`? I think it's giving us the actual full file paths, right?

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
```diff
@@ -217,6 +219,7 @@ public static void AddWebJobsScriptHost(this IServiceCollection services, IConfi
             services.ConfigureOptions<ScriptApplicationHostOptionsSetup>();
             services.AddSingleton<IOptionsChangeTokenSource<ScriptApplicationHostOptions>, ScriptApplicationHostOptionsChangeTokenSource>();
             services.ConfigureOptions<StandbyOptionsSetup>();
+            services.ConfigureOptions<WorkerConfigurationResolverOptionsSetup>();
```
brettsam: I don't believe this (or the next line, actually) are required. If you look at what `ConfigureOptionsWithChangeTokenSource` does -- it does this already.

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
brettsam: If you remove this, does it work? I think we did this with the LanguageWorkerOptions below b/c it was a complex setup. This new one is not. I *think* the registration will be copied to the child container automatically and all just work.

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
brettsam: I filed https://github.com/Azure/azure-functions-host/issues/11208 to make this easier in the future.

src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
```diff
@@ -216,6 +216,11 @@ internal static class LoggerExtension
            new EventId(342, nameof(OutdatedExtensionBundle)),
            "Your current bundle version {currentVersion} has reached end of support on Aug 4, 2026. Upgrade to [{suggestedMinVersion}.*, {suggestedMaxVersion}.0.0). For more information, see https://aka.ms/FunctionsBundlesUpgrade");
 
+        private static readonly Action<ILogger, string, Exception> _defaultWorkersDirectoryPath =
+           LoggerMessage.Define<string>(LogLevel.Debug,
+           new EventId(343, nameof(DefaultWorkersDirectoryPath)),
+           "Workers Directory set to: {workersDirPath}");
```
liliankasem, line 222: Suggested change
      
    
    
      
          
            
                       "Workers Directory set to: {workersDirPath}");
          
          
            
                       "Worker directory set to: {workerDirPath}");
          
      
    
    
  

Use “Worker Directory” — in noun–noun phrases, the first noun is usually singular unless it’s a fixed name, so “worker” is grammatically cleaner than “workers” here. I know its named "workers" because multiple workers would be found here, but "Worker directory” means “a directory for workers” — the fact that it holds multiple workers is already implied.

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
```diff
@@ -90,15 +71,18 @@ internal void BuildWorkerProviderDictionary()
...
 
-            foreach (var workerDir in Directory.EnumerateDirectories(WorkersDirPath))
+            if (workerConfigs is null || !workerConfigs.Any())
             {
-                string workerConfigPath = Path.Combine(workerDir, RpcWorkerConstants.WorkerConfigFileName);
-                if (File.Exists(workerConfigPath))
-                {
-                    AddProvider(workerDir);
-                }
+                _logger.LogTrace("No worker configs found.");
```
liliankasem: What happens upstream if no worker configs are found? We just return here so I am wondering if we should throw, is this a bad state?

src/WebJobs.Script/Workers/Rpc/IWorkerConfigurationResolver.cs
liliankasem: I also prefer `WorkerConfigurationInfo` here

src/WebJobs.Script/Workers/Rpc/Configuration/WorkerConfigurationResolverOptions.cs
liliankasem: I think I prefer `WorkerRootDir` - multiple workers is already implied

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
surgupta-msft: Hmm, I remember it wasn't working without this code when I was testing the big worker decoupling PR but it seems to be working fine now - may be due to all the refactoring and improvements we have done. Remove the code.

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
surgupta-msft: There is another method in this class `AddProvidersFromAppSettings()` which also needs `WorkersDirPath`. I have again added `WorkersDirPath` as global variable similar to dev branch - https://github.com/Azure/azure-functions-host/blob/f6cdf90796999c23cbe896b126ca2a754c021204/src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs#L60

src/WebJobs.Script/Workers/Rpc/Configuration/WorkerConfigurationResolutionInfo.cs
surgupta-msft: Updated naming and sample output is as below based on multiple feedback. Let me know if you have any concerns.

```csharp
WorkersRootDirPath - "C:\home\SiteExtensions\Functions\64bit\workers"
WorkerConfigPaths - ["C:\home\SiteExtensions\Functions\64bit\workers\java\", "C:\home\SiteExtensions\Functions\64bit\workers\node\".....]
```

src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
surgupta-msft: This was an existing log message in dev branch. Avoiding updating the log message as it may break detector or tooling depends on it - https://github.com/Azure/azure-functions-host/blob/f6cdf90796999c23cbe896b126ca2a754c021204/src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs#L93

src/WebJobs.Script/Workers/Rpc/Configuration/DefaultWorkerConfigurationResolver.cs
```diff
@@ -0,0 +1,50 @@
...
+        private readonly ILogger _logger;
+        private readonly IOptionsMonitor<WorkerConfigurationResolverOptions> _workerConfigurationResolverOptions;
+        private readonly IFileSystem _fileSystem;
+
+        public DefaultWorkerConfigurationResolver(ILoggerFactory loggerFactory,
+                                                    IFileSystem fileSystem,
+                                                    IOptionsMonitor<WorkerConfigurationResolverOptions> workerConfigurationResolverOptions)
+        {
+            ArgumentNullException.ThrowIfNull(loggerFactory);
+            _logger = loggerFactory.CreateLogger(ScriptConstants.LogCategoryWorkerConfig);
```
kshyju, line 25: Why do we use this custom category? Why not inject `ILogger<DefaultWorkerConfigurationResolver> logger` ?

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
kshyju: It seems that global variable is only used inside the `AddProvider` method, so you could pass it as a parameter instead. As a general best practice, try to avoid global variables whenever possible.

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
```diff
@@ -47,17 +50,10 @@ public RpcWorkerConfigFactory(IConfiguration config,
...
-
-            if (!string.IsNullOrEmpty(workersDirectorySection.Value))
-            {
-                WorkersDirPath = workersDirectorySection.Value;
-            }
+            _workerConfigurationResolver = workerConfigurationResolver ?? throw new ArgumentNullException(nameof(workerConfigurationResolver));
         }
 
-        public string WorkersDirPath { get; }
+        public string WorkersDirPath { get; private set; }
```
kshyju: This is also marked as `public`, but I don’t see it being used outside this class. As mentioned in the other comment, consider removing it entirely and passing it as a parameter instead of relying on a global variable to share data between methods.

src/WebJobs.Script/Workers/Rpc/IWorkerConfigurationResolver.cs
```diff
@@ -0,0 +1,17 @@
...
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
+{
+    /// <summary>
+    /// Interface to resolve the Worker Configs.
+    /// </summary>
+    internal interface IWorkerConfigurationResolver
+    {
+        WorkerConfigurationInfo GetConfigurationInfo();
```
kshyju, line 18: Add a summary for this method.

src/WebJobs.Script/Workers/Rpc/IWorkerConfigurationResolver.cs
```diff
@@ -0,0 +1,17 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Microsoft.Azure.WebJobs.Script.Workers.Rpc.Configuration;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.WebJobs.Script.Workers.Rpc
+{
+    /// <summary>
+    /// Interface to resolve the Worker Configs.
```
kshyju:
```suggestion
    /// Provides an abstraction for retrieving worker configuration resolution details.
```

src/WebJobs.Script/Workers/Rpc/Configuration/WorkerConfigurationResolverOptionsSetup.cs
```diff
@@ -0,0 +1,103 @@
...
+            return null;
+        }
+
+        private string GetConfigurationSectionValue(IConfiguration configuration, string configurationSource, string requiredSection)
+        {
+            var section = configuration?.GetSection(requiredSection);
+
+            if (!string.IsNullOrEmpty(section?.Value))
+            {
+                _logger.LogTrace("Found configuration section '{requiredSection}' in '{configurationSource}'.", requiredSection, configurationSource);
```
kshyju, line 95: Do you have a TODO item (work item) to inspect telemetry and cleanup this in the future?

src/WebJobs.Script/Workers/Rpc/Configuration/WorkerConfigurationResolverOptionsSetup.cs
```diff
@@ -0,0 +1,103 @@
...
+    {
+        private readonly IConfiguration _configuration;
+        private readonly IScriptHostManager _scriptHostManager;
+        private readonly IFileSystem _fileSystem;
+        private readonly ILogger _logger;
+
+        public WorkerConfigurationResolverOptionsSetup(ILoggerFactory loggerFactory, IConfiguration configuration, IScriptHostManager scriptHostManager, IFileSystem fileSystem)
+        {
+            ArgumentNullException.ThrowIfNull(loggerFactory);
+            _logger = loggerFactory.CreateLogger(ScriptConstants.LogCategoryWorkerConfig);
```
kshyju, line 24: Double checking, is this the log category we want here?

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
surgupta-msft: #### Code flow before my changes:
- `WorkersDirPath` is a global variable. It's value is getting set in the constructor - https://github.com/Azure/azure-functions-host/blob/f6cdf90796999c23cbe896b126ca2a754c021204/src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs#L51
- 2 methods will need this global variable after they call AddProvider() - https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs#L87-L88
- Ultimately, its the `AddProvider()` which uses this variable - https://github.com/Azure/azure-functions-host/blob/f6cdf90796999c23cbe896b126ca2a754c021204/src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs#L131


#### Code flow after my changes:

- We cannot set `WorkersDirPath` in the constructor because now it is part of `WorkerConfigurationInfo` which we get by calling `AddProviders()` in RPCWorkerConfigFactory - 
    - https://github.com/Azure/azure-functions-host/blob/0ee087135d89bde55c04640498862f2ba07ce21e/src/WebJobs.Script/Workers/Rpc/Configuration/WorkerConfigurationInfo.cs#L8
    - https://github.com/Azure/azure-functions-host/blob/0ee087135d89bde55c04640498862f2ba07ce21e/src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs#L75
- `WorkersDirPath` is still a global variable because 2 methods will need it - [Link](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs#L87-L88).  If I don't make it a global var, `AddProviders()` will have this information to pass to `AddProvider()` but `AddProvidersFromAppSettings()` will not have this information to pass to `AddProvider()`

src/WebJobs.Script/Workers/Rpc/Configuration/WorkerConfigurationResolverOptionsSetup.cs
surgupta-msft: Here is the issue - https://github.com/Azure/azure-functions-host/issues/11249. Do you also want me to add a comment and point to this issue?

src/WebJobs.Script/Workers/Rpc/Configuration/DefaultWorkerConfigurationResolver.cs
surgupta-msft: This was an existing code in `LanguageWorkerOptionsSetup` - https://github.com/Azure/azure-functions-host/blob/f6cdf90796999c23cbe896b126ca2a754c021204/src/WebJobs.Script/Workers/Rpc/Configuration/LanguageWorkerOptionsSetup.cs#L43

Before my changes, all the logs from `LanguageWorkerOptionsSetup` and `RPCWorkerConfigFactory` will have `Source = "Host.LanguageWorkerConfig"`. I just kept it same so that after my changes, the Source of the same log is unchanged.

**log from current Host 4.1041:**
image

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
kshyju: You can move the code ` _workerConfigurationResolver.GetConfigurationInfo()` to the caller of `AddProviders` which is `BuildWorkerProviderDictionary`. Now you can pass it to both.

src/WebJobs.Script/Workers/Rpc/Configuration/WorkerConfigurationResolverOptionsSetup.cs
kshyju: I think that is good enough.

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
```diff
@@ -128,7 +106,7 @@ internal void AddProvider(string workerDir)
                         string workerRuntime = Path.GetFileName(workerDir);
                         // Only skip worker directories that don't match the current runtime.
                         // Do not skip non-worker directories like the function app payload directory
-                        if (!workerRuntime.Equals(_workerRuntime, StringComparison.OrdinalIgnoreCase) && workerDir.StartsWith(WorkersDirPath))
+                        if (!workerRuntime.Equals(_workerRuntime, StringComparison.OrdinalIgnoreCase) && workerDir.StartsWith(workerConfigurationInfo.WorkersRootDirPath))
```
brettsam: It took me a while to understand why we're checking `StartsWith(workerConfigurationInfo.WorkersRootDirPath)` here. It seems the intent is to skip the "does the folder match the runtime" validation in the app setting scenario. Because in that scenario, the folder structure doesn't matter and we trust whatever the customer tells us to use. It's using the root dir as an indirect (and, I think, unclear) way of doing that.

If that's the case, can we instead pass a bool in like `AddProvider(string workerDir, string skipPathValidation)`? That seems much more clear about what's happening here. Then we don't need to pass this path around anymore, either.

If we think that's too risky now (maybe there's a scenario I'm not thinking about), it's fine to skip this, it just feels like overkill to pass this path around only for this one check that seems like it could be handled more clearly in other ways.

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
```diff
@@ -111,12 +89,12 @@ internal void AddProvidersFromAppSettings()
...
                 {
                     _workerDescriptionDictionary.Remove(languageSection.Key);
-                    AddProvider(workerDirectorySection.Value);
+                    AddProvider(workerDirectorySection.Value, workerConfigurationInfo);
                 }
             }
         }
 
-        internal void AddProvider(string workerDir)
+        internal void AddProvider(string workerDir, WorkerConfigurationInfo workerConfigurationInfo)
```
brettsam: nit: if you only need to use one property inside of a method, you can be more specific here. In this case, if you only need the `WorkersRootDirPath`, this parameter could change to a string `workersRootDirPath`. It makes things more clear. Although my other comment about making this a bool may be even more clear.

brettsam (approved):
I have a couple nits about internal methods that we can skip for now, but I wanted to capture them. They've been around a while so nothing was introduced here but they bugged me.

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
surgupta-msft: discussed offline

src/WebJobs.Script/Workers/Rpc/Configuration/LanguageWorkerOptionsSetup.cs
```diff
@@ -6,6 +6,7 @@
 using System.Diagnostics;
 using Microsoft.Azure.WebJobs.Script.Diagnostics;
 using Microsoft.Azure.WebJobs.Script.Workers.Profiles;
+using Microsoft.Azure.WebJobs.Script.Workers.Rpc.Configuration;
```
kshyju: Remove unused using statements.

src/WebJobs.Script/Workers/Rpc/Configuration/WorkerConfigurationResolverOptionsSetup.cs
```diff
@@ -0,0 +1,103 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
```
kshyju: Remove unused using statement.

kshyju (approved):

brettsam (approved):

PR #11230, opened 2025-08-01
Update outdated bundle message as per policy

### LINKED ISSUE #10941: Extension Bundles Retirement Messaging
The ask is to include warning messaging in the host that indicates to customers they should update their extension bundles to the latest available.

manvkaur:
Updated the outdated bundle warning message to be generic as the deprecation date is getting postponed.

resolves #10941 

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
+++ b/src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
@@ -206,15 +206,10 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics.Extensions
         private static readonly Action<ILogger, string, Exception> _publishingMetrics =
             LoggerMessage.Define<string>(LogLevel.Debug, new EventId(338, nameof(PublishingMetrics)), "{metrics}");
 
-        private static readonly Action<ILogger, string, int, int, Exception> _outdatedExtensionBundleFuture =
+        private static readonly Action<ILogger, string, int, int, Exception> _outdatedExtensionBundle =
            LoggerMessage.Define<string, int, int>(LogLevel.Warning,
            new EventId(342, nameof(OutdatedExtensionBundle)),
-           "Your current bundle version {currentVersion} will reach end of support on Aug 4, 2026. Upgrade to [{suggestedMinVersion}.*, {suggestedMaxVersion}.0.0). For more information, see https://aka.ms/FunctionsBundlesUpgrade");
-
-        private static readonly Action<ILogger, string, int, int, Exception> _outdatedExtensionBundlePast =
-           LoggerMessage.Define<string, int, int>(LogLevel.Warning,
-           new EventId(342, nameof(OutdatedExtensionBundle)),
-           "Your current bundle version {currentVersion} has reached end of support on Aug 4, 2026. Upgrade to [{suggestedMinVersion}.*, {suggestedMaxVersion}.0.0). For more information, see https://aka.ms/FunctionsBundlesUpgrade");
+           "Your app is using an older version - {currentVersion} of extension bundle. Upgrade to [{suggestedMinVersion}.*, {suggestedMaxVersion}.0.0). For more information, see https://aka.ms/FunctionsBundlesUpgrade");
 
         public static void PublishingMetrics(this ILogger logger, string metrics)
         {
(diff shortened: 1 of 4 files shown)


vrdmr (approved):

satvu (approved):

satvu (approved):

src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
```diff
@@ -206,15 +206,10 @@ internal static class LoggerExtension
...
+        private static readonly Action<ILogger, string, int, int, Exception> _outdatedExtensionBundle =
            LoggerMessage.Define<string, int, int>(LogLevel.Warning,
            new EventId(342, nameof(OutdatedExtensionBundle)),
-           "Your current bundle version {currentVersion} will reach end of support on Aug 4, 2026. Upgrade to [{suggestedMinVersion}.*, {suggestedMaxVersion}.0.0). For more information, see https://aka.ms/FunctionsBundlesUpgrade");
-
-        private static readonly Action<ILogger, string, int, int, Exception> _outdatedExtensionBundlePast =
-           LoggerMessage.Define<string, int, int>(LogLevel.Warning,
-           new EventId(342, nameof(OutdatedExtensionBundle)),
-           "Your current bundle version {currentVersion} has reached end of support on Aug 4, 2026. Upgrade to [{suggestedMinVersion}.*, {suggestedMaxVersion}.0.0). For more information, see https://aka.ms/FunctionsBundlesUpgrade");
+           "Your app is using an older version - {currentVersion} of extension bundle. Upgrade to [{suggestedMinVersion}.*, {suggestedMaxVersion}.0.0). For more information, see https://aka.ms/FunctionsBundlesUpgrade");
```
RohitRanjanMS: The suggestedMinVersion and suggestedMaxVersion values will become outdated as we move to newer versions, requiring a code change to update the constants. Do we want to include the suggested version in the log, given that we already provide a link to documentation that will always have the most up‑to‑date version recommendations?

manvkaur: host has hard coded v4majorversion, for now, we want to show this message only for anything older than v4.
In case v5 comes in future, we will need to revisit this message and also the timeline we want to start saying v4 is deprecated.
We will have to revisit all occurrences of this script constant - https://github.com/Azure/azure-functions-host/blob/25427b660bea5335f934454a64fa44bff2a3e9dc/src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs#L398

Suggested versions are coming from the version being compared against.

jviau: I am also not a big fan of having these details hardcoded. The behavior of this log message will change if a customer pins to an older version. This changes the customer's perception of what is supported or not.

Do we have any PM input on this? @mattchenderson 

While I know there is nothing we can do for already shipped versions. But if we make this change pull in information from some external config source (honestly not sure if we have anything set up for that), we could make control this log behavior without needing code changes.

manvkaur: Are we open to making an additional API call to fetch these configs, or is there already an existing mechanism that provides similar configuration details?

Also, I’m unclear on how the behavior changes if a customer pins to an older version. As I understand it, we’re simply checking if the version in use is below v4 and then recommending a more flexible version range. Is there something I’m missing?

Adding @MadhuraBharadwaj-MSFT for PM input on the support.

jviau: Updating from external conversation: my input is to not include suggested version range. Include a link to documentation only. This way the doc link can be updated independently and the range a customer gets suggested will not depend on what host version they are on.

This will also avoid any extra complexities with external config

vrdmr (approved):

src/WebJobs.Script/Diagnostics/Extensions/LoggerExtension.cs
manvkaur: resolved in internal discussion - removed hardcoding of suggested min and max versions and pointing to the short link

vrdmr (approved):

PR #11250, opened 2025-08-13
[LinuxConsumption] Mark Container Unhealthy During Failed Specialization

### LINKED ISSUE #9663: Azure File Mounting failures not treated as fatal even if App is not using Run-From-Package
// wawseas
FunctionsLogs
| where PreciseTimeStamp >= datetime(2023-10-11T14:08:29.803Z) and PreciseTimeStamp <= datetime(2023-10-11T14:08:44.804Z)
| where SourceNamespace == "SFLWAWS"
| where RoleInstance == "App-4E21DB8D-638326191442476902"

**Repro steps:**
Linux Consumption  app
No WEBSITE_RUN_FROM_PACKAGE appsetting configured
No remote build triggered,
Configure AzureFiles using WEBSITE_CONTENTAZUREFILECONNECTIONSTRING, WEBSITE_CONTENTSHARE - BUT use invalid account credentials so mounting AzureFiles will fail at runtime.
Trigger cold start

**Expected:** Instance should cold start and be removed right away since AzureFiles mount failed and the app has no other source of app content.
**Actual:** Instance continues to stay alive with 0 functions.

**Potential Fix:**
https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.WebHost/Management/AtlasInstanceManager.cs#L310C20-L310C89
The return value from mount operation is ignored here

Tonewall:
WI: https://msazure.visualstudio.com/Antares/_workitems/edit/32785404/

resolves #9663

### Pull request checklist

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -6,4 +6,4 @@
 - Add JitTrace Files for v4.1042
 - Updating OTel related nuget packages (#11216)
 - Moving to version 1.5.7 of Microsoft.Azure.AppService.Middleware.Functions (https://github.com/Azure/azure-functions-host/pull/11231)
-
+- Mark Failed Container Specialization as Fatal (https://github.com/Azure/azure-functions-host/pull/11250)
--- a/src/WebJobs.Script.WebHost/Management/AtlasInstanceManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/AtlasInstanceManager.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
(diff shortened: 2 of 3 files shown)


jviau (approved):

jviau:
@Tonewall is this change still needed? With the health checks recently added, the system will not report healthy until we get successful script host started. So if any point of specialization fails before starting the script host, the instance will remain unhealthy.

This should be tested though. Also if there are any components here worth their own health check (like checking if azure files is mounted successfully), you can open a new issue requesting they be implemented.

Tonewall:
> @Tonewall is this change still needed? With the health checks recently added, the system will not report healthy until we get successful script host started. So if any point of specialization fails before starting the script host, the instance will remain unhealthy.
> 
> This should be tested though. Also if there are any components here worth their own health check (like checking if azure files is mounted successfully), you can open a new issue requesting they be implemented.

Closing the PR since we're focused on migrating to legion and platform will handle specialization errors

PR #11264, opened 2025-08-25
Fixing /admin/host/resume ObjectDisposedException

### LINKED ISSUE #11263: /admin/host/resume throws ObjectDisposedException
There's a race here -- we set the RequestServices to be the current ScriptHost's services when the request starts... but then we fire-and-forget the disposal of those services when we go to resume (which restarts the host).

The error is just noise at this point. The error is returned to the infrastructure API call, and it handles the error properly. Everything behaves as expected.

brettsam:
Fixes #11263

The bug here is that during a request, we register a DI child scope container with the request... then we dispose it mid-request, so when anything in ASP.NET tries to get a service later, it throws. It's a race b/c it's possible that the request returns before the old container is disposed. Despite this, everything seems to be working b/c all the hard work has already been done by the time the exception is thrown. The JobHost is restarted and everything continues on happily even though we return a 500 to the caller.

I have a preliminary PR that I've just pushed up but it's not truly complete. It'll work for what we need but it's quite tricky to get right with how DI works.

Some background:
- For in-proc we use DryIoc and completely replace the DI engine to manage child scopes
- We added this code to prevent disposal of the DI container while a child scope was still active: https://github.com/Azure/azure-functions-host/blob/b0797762d1afb53b1c7093922c9d2bea2357d025/src/WebJobs.Script.WebHost/DependencyInjection/ScopedResolver.cs#L33-L37
- When we moved to out-of-proc, we removed DryIoc and use the default Microsoft.Extensions.DependencyInjection and did not add this "child scope tracking" behavior.
- There's not a great way to use the decorator pattern here. There's a bunch of ways to bypass the decorator (e.g. with constructor injection) and use the original IServiceProvider/IServiceScopeFactory because they're hard-coded and not able to be overwritten:
  - this issue links to other discussions: https://github.com/dotnet/runtime/issues/38240 
  - here's the hard-coding: https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.DependencyInjection/src/ServiceProvider.cs#L63-L66

~~What I've got *should* be good enough for the very specific scenario we need it for, but there still seems like there's gaps that may bite us later.~~

~~Want to discuss with @fabiocav later whether we should use this same approach? Or abandon it and just skip registering services for /admin calls? Or if there's other approaches.~~

~~Will leave it in Draft for now.~~

Edit:
This is now ready for review. Rather than use the non-optimal approach with the built-in DI, we're going to just skip replacing these services on calls that we know don't need them and can be problematic. We can add to this if we need to in the future.

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.WebHost/WebJobsApplicationBuilderExtension.cs
+++ b/src/WebJobs.Script.WebHost/WebJobsApplicationBuilderExtension.cs
@@ -90,9 +90,12 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
                 config.UseMiddleware<HostWarmupMiddleware>();
             });
 
-            // This middleware must be registered before any other middleware depending on
-            // JobHost/ScriptHost scoped services.
-            builder.UseMiddleware<ScriptHostRequestServiceProviderMiddleware>();
+            builder.UseWhen(context => !context.Request.IsAdminResumeRequest(), config =>
+            {
+                // This middleware must be registered before any other middleware depending on
+                // JobHost/ScriptHost scoped services.
+                config.UseMiddleware<ScriptHostRequestServiceProviderMiddleware>();
+            });
 
             if (environment.IsLinuxAzureManagedHosting())
             {
--- a/test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/DrainModeResumeEndToEndTests.cs
+++ b/test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/DrainModeResumeEndToEndTests.cs
@@ -1,9 +1,14 @@
 using System;
+using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Net;
 using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Script.Configuration;
 using Microsoft.Azure.WebJobs.Script.WebHost.Models;
 using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Logging;
 using Microsoft.WebJobs.Script.Tests;
 using Newtonsoft.Json;
 using Xunit;
@@ -43,9 +48,10 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.EndToEnd
 
             // Validate host is "Running" after resume is called
             response = await SamplesTestHelpers.InvokeResume(this);
+            Assert.True(HttpStatusCode.OK == response.StatusCode,
+                string.Join(Environment.NewLine, Host.GetWebHostLogMessages().Where(m => m.Level == LogLevel.Error).Select(m => m.ToString())));
             responseString = await response.Content.ReadAsStringAsync();
             var resumeStatus = JsonConvert.DeserializeObject<ResumeStatus>(responseString);
-
             Assert.Equal(ScriptHostState.Running, resumeStatus.State);
 
             // Validate the drain state is changed to "Disabled"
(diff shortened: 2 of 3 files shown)


test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/DrainModeResumeEndToEndTests.cs
```diff
@@ -88,18 +93,21 @@ public async Task DrainModeDisabled_RunningHost_ReturnsOk()
...
+        public override void ConfigureWebHost(IConfigurationBuilder configBuilder)
         {
-            base.ConfigureScriptHost(webJobsBuilder);
+            base.ConfigureWebHost(configBuilder);
+
+            configBuilder.AddInMemoryCollection(new Dictionary<string, string>
+            {
+                // This forces the hosts to be stopped and disposed before a new one starts.
+                // There was a bug hiding here originally, so we'll run all these tests this way.
+                { ConfigurationSectionNames.SequentialJobHostRestart, "true" }
```
jviau, line 110: I believe this behavior has some major differences. It doesn't call something regarding worker process cleanup. Any concern with missing that logic in this test?

brettsam: This is the only place that I see it's called

https://github.com/Azure/azure-functions-host/blob/1bb6bd88bf75a84d84c077681ff00c187b07acdd/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs#L633-L637

Which is awaiting the call to `Orphan()` so that we're guaranteed the previous host is disposed before we start another one. Otherwise we fire-and-forget that call.

The bug here is a race and it always passes for me locally (and apparently in CI) because the new host is started and the request returns while we're still waiting to dispose the orphaned one.

brettsam:
Added a bunch more context to the description; leaving this in Draft now until I can discuss with @fabiocav. It's possible this was a purposeful omission and he had another plan (or maybe not :-))

brettsam:
Updated approach after talking with @fabiocav -- we'll just skip the service replacement on /admin/host/resume calls

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/DrainModeResumeEndToEndTests.cs
```diff
@@ -43,9 +48,10 @@ public async Task DrainModeEnabled_RunningHost_StartsNewHost_ReturnsOk()
 
             // Validate host is "Running" after resume is called
             response = await SamplesTestHelpers.InvokeResume(this);
+            Assert.True(HttpStatusCode.OK == response.StatusCode,
```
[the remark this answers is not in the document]
brettsam: Did this on purpose so we could add additional logs upon failure.

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/DrainModeResumeEndToEndTests.cs
```diff
@@ -88,18 +94,21 @@ public async Task DrainModeDisabled_RunningHost_ReturnsOk()
...
-        public override void ConfigureScriptHost(IWebJobsBuilder webJobsBuilder)
+        public override void ConfigureWebHost(IConfigurationBuilder configBuilder)
         {
-            base.ConfigureScriptHost(webJobsBuilder);
+            base.ConfigureWebHost(configBuilder);
+
+            configBuilder.AddInMemoryCollection(new Dictionary<string, string>
+            {
+                // This forces the hosts to be stopped and disposed before a new one starts.
+                // There was a bug hiding here originally, so we'll run all these tests this way.
```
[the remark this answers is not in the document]
brettsam: It's okay. This is a general improvement to remove the race condition in our tests. Don't need an exact link.

src/WebJobs.Script.WebHost/WebJobsApplicationBuilderExtension.cs
```diff
@@ -90,9 +90,12 @@ public static IApplicationBuilder UseWebJobsScriptHost(this IApplicationBuilder
...
             });
 
-            // This middleware must be registered before any other middleware depending on
-            // JobHost/ScriptHost scoped services.
-            builder.UseMiddleware<ScriptHostRequestServiceProviderMiddleware>();
+            builder.UseWhen(context => !context.Request.IsAdminResumeRequest(), config =>
+            {
+                // This middleware must be registered before any other middleware depending on
+                // JobHost/ScriptHost scoped services.
+                config.UseMiddleware<ScriptHostRequestServiceProviderMiddleware>();
```
jviau, line 97: I find it interesting we would want this behavior on _any_ admin requests. But guess that is how it is, big risk to change that further.

jviau (approved):

src/WebJobs.Script.WebHost/WebJobsApplicationBuilderExtension.cs
brettsam: Yeah I was thinking about that too -- but this seems a lot safer right now. We may end up there.

liliankasem (approved):

PR #11302, opened 2025-09-05
Add support for host specialization with unencrypted payload

manikantanallagatla:
Current assign endpoint takes a encrypted payload for specializing. This is not memory efficient. Adding new API for taking the specialization unencrypted. This will be usedul in ATOL as well

### Pull request checklist

* [] I have added all required tests (Unit tests, E2E tests)


jviau (changes requested):
Instead of adding a new endpoint entirely, can we update `EncryptedHostAssignmentContext` as follows:

1. rename to `HostAssignmentRequest`
2. add properties as follows:

``` CSharp
public class HostAssignmentRequest
{
    [JsonProperty("encryptedContext")]
    public string EncryptedContext { get; set; }

    [JsonProperty("context")]
    public HostAssignmentContext Context { get; set; }
}
```

Now callers can choose to supply encrypted or not based on what JSON payload they send. The implementation will then check which of the two properties is not-null and use that.

manikantanallagatla:
> Instead of adding a new endpoint entirely, can we update `EncryptedHostAssignmentContext` as follows:
> 
> 1. rename to `HostAssignmentRequest`
> 2. add properties as follows:
> 
> ```cs
> public class HostAssignmentRequest
> {
>     [JsonProperty("encryptedContext")]
>     public string EncryptedContext { get; set; }
> 
>     [JsonProperty("context")]
>     public HostAssignmentContext Context { get; set; }
> }
> ```
> 
> Now callers can choose to supply encrypted or not based on what JSON payload they send. The implementation will then check which of the two properties is not-null and use that.

Hi, Thats a valid suggestion. But, that will break backward compatibility and makes it very difficult to deploy. We would need to change current flow of assignment in Antares outside ATOL. We would have to check which hostversion is supporting which endpoint and send object accordingly.

With new endpoint, we can remove old endpoint once we move to endpoint entirely.

In legion as well, we have seperate endppints for encrypted and unencypted payload to make it easier for maintenance

jviau:
@manikantanallagatla can you expand more on how my proposal limits maintenance? I am not quite understanding your point.

> But, that will break backward compatibility and makes it very difficult to deploy.

How so? One API will serve both formats of JSON. The existing encrypted, or the new unencrypted. The caller can send either of the two formats.

``` JSON
{
  "encryptedContext": "< some encrypted string >"
}
```
or
``` JSON
{
  "context": {
    "siteName": "MySiteName", // and the rest of the object
   }
}
```

> With new endpoint, we can remove old endpoint once we move to endpoint entirely.

And with my proposal we can remove the support for `encryptedContext` in the JSON payload. But it is highly unlikely we would remove that, as it would be a breaking change.

What this boils down to, is ATOL can call `admin/instance/assign` with an unencrypted payload. I don't see any benefit from making it explicitly a different endpoint.

manikantanallagatla:
> > Instead of adding a new endpoint entirely, can we update `EncryptedHostAssignmentContext` as follows:
> > 
> > 1. rename to `HostAssignmentRequest`
> > 2. add properties as follows:
> > 
> > ```cs
> > public class HostAssignmentRequest
> > {
> >     [JsonProperty("encryptedContext")]
> >     public string EncryptedContext { get; set; }
> > 
> >     [JsonProperty("context")]
> >     public HostAssignmentContext Context { get; set; }
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Now callers can choose to supply encrypted or not based on what JSON payload they send. The implementation will then check which of the two properties is not-null and use that.
> 
> Hi, Thats a valid suggestion. But, that will break backward compatibility and makes it very difficult to deploy. We would need to change current flow of assignment in Antares outside ATOL. We would have to check which hostversion is supporting which endpoint and send object accordingly.
> 
> With new endpoint, we can remove old endpoint once we move to endpoint entirely.
> 
> In legion as well, we have seperate endppints for encrypted and unencypted payload to make it easier for maintenance

Oh got it. There is a HostAssignmentContext inside HostAssignmentRequest. I missed that. Apologies.

We can make it a single request with this. I will give it a try. Thanks @jviau

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
```diff
@@ -79,5 +75,30 @@ public IActionResult GetHttpHealthStatus()
             // Reaching here implies that http health of the container is ok.
             return Ok();
         }
+
+        private async Task<IActionResult> AssignInternal(HostAssignmentContext assignmentContext)
```
jviau: nit: include `Async` suffix for `Task` returning methods.
```suggestion
        private async Task<IActionResult> AssignInternalAsync(HostAssignmentContext assignmentContext)
```

jviau:
@manikantanallagatla I see you re-requested my review, but I don't see the changes we discussed. There are still 2 separate APIs and encrypted/unencrypted models are not combined into the same top-level type. Are these changes still coming?

vivekjilla:
@jviau just for my curiosity, why do we want to hack into existing api and break the request model object structure from current flow and combining multiple properties where only one is used at any given time, instead of keeping the cleaner segregation with different APIs and their models, and not breaking/modifying the current API on the way.

@manikantanallagatla let's ensure any changes here are aligned by everyone and thought through, because I think we might be making some changes on Legion Platform side or Go Server code to call these APIs while maintaining backward compatibility. Might have to change all those if we have to retrofit new behavior into existing API without breaking the current flows.

manikantanallagatla:
@jviau , I merged the APIs into 1 API and requested review. Then when I discussed with Bala, we wanted two separate APIs so that it is cleaner. So, I again reverted that commit of merging into 1 API. Anyways I will start a thread in teams to have a conclusion on 1 API vs 2 APIs.

jviau:
> @jviau just for my curiosity, why do we want to hack into existing api and break the request model object structure from current flow and combining multiple properties where only one is used at any given time, instead of keeping the cleaner segregation with different APIs and their models, and not breaking/modifying the current API on the way.

I don't see this as a hack at all. It doesn't break anything. We are effectively turning the existing JSON payload into a "One-Of" structure. It is either the encrypted payload or the unencrypted payload, and the existing API will determine what assign flow to use based on the shape of the JSON input. We can add validation in the API handler to assert only one can be set at a time. I do not think having an `assign` and `assign2` API is cleaner in any ways. I think my approach is much cleaner. Because again, the API will **not break** for existing callers.

Existing callers. This is the same as today, and it will be the same after my suggested changes. There is **no change** for these callers.

``` http
POST "admin/instance/assign"

{
    "encryptedContext": "< encrypted context >"
}
```


**NEW** callers wanting unencrypted assign:

``` http
POST "admin/instance/assign"

{
    "context":  {
        // unencrypted host assignment context
    }
}
```

Additional scenario which will result in an HTTP 400. Callers will not be allowed to provide both encrypted and unencrypted context.

``` http
POST "admin/instance/assign"

{
    "encryptedContext": "< encrypted context >",
    "context":  {
        // unencrypted host assignment context
    }
}
```

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
```diff
@@ -79,5 +75,30 @@ public IActionResult GetHttpHealthStatus()
...
+            if (error != null)
+            {
+                return StatusCode(StatusCodes.Status500InternalServerError, error);
+            }
+
+            var succeeded = _instanceManager.StartAssignment(assignmentContext);
+
+            return succeeded
+                ? Accepted()
+                : StatusCode(StatusCodes.Status409Conflict, "Instance already assigned");
```
RohitRanjanMS: Is there any other reason for which `StartAssignment` can return false?

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
RohitRanjanMS: We need a better name for this method.

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
```diff
@@ -79,5 +75,30 @@ public IActionResult GetHttpHealthStatus()
             // Reaching here implies that http health of the container is ok.
             return Ok();
         }
+
+        private async Task<IActionResult> AssignInternal(HostAssignmentContext assignmentContext)
+        {
+            // before starting the assignment we want to perform as much
+            // up front validation on the context as possible
+            string error = await _instanceManager.ValidateContext(assignmentContext);
```
RohitRanjanMS: Returning a string where null means “valid” and a non‑null value means “error message” will work, but it’s not the most expressive design. We can make it more obvious.

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
```diff
@@ -41,27 +41,23 @@ public async Task<IActionResult> Assign([FromBody] EncryptedHostAssignmentContex
...
-            }
+            return await AssignInternal(assignmentContext);
+        }
 
-            // Wait for Sidecar specialization to complete before returning ok.
-            // This shouldn't take too long so ok to do this sequentially.
-            error = await _instanceManager.SpecializeMSISidecar(assignmentContext);
-            if (error != null)
+        [HttpPost]
+        [Route("admin/instance/assign2")]
```
RohitRanjanMS: can we have a better name instead of `assign2`?

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
```diff
@@ -41,27 +41,23 @@ public async Task<IActionResult> Assign([FromBody] EncryptedHostAssignmentContex
...
+        [Authorize(Policy = PolicyNames.AdminAuthLevel)]
+        public async Task<IActionResult> Assign2([FromBody] FunctionsWorkerContainerAssignmentContext workerAssignmentContext)
```
RohitRanjanMS: Is this a host assignment or a worker assignment?

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
```diff
@@ -41,27 +41,23 @@ public async Task<IActionResult> Assign([FromBody] EncryptedHostAssignmentContex
...
+        {
+            if (workerAssignmentContext?.AssignmentContext == null || workerAssignmentContext.AssignmentContext == null)
```
RohitRanjanMS: redundant null check.

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
```diff
@@ -41,27 +41,23 @@ public async Task<IActionResult> Assign([FromBody] EncryptedHostAssignmentContex
...
             {
-                return StatusCode(StatusCodes.Status500InternalServerError, error);
+                return BadRequest("Assignment context is missing.");
             }
 
-            var succeeded = _instanceManager.StartAssignment(assignmentContext);
+            _logger.LogDebug($"Starting container assignment for host : {Request?.Host}");
```
RohitRanjanMS: use structured logging.

RohitRanjanMS: Can `Request` ever be null?

src/WebJobs.Script.WebHost/Models/FunctionsWorkerContainerAssignmentContext.cs
```diff
@@ -0,0 +1,14 @@
+// -----------------------------------------------------------
```
RohitRanjanMS: // Copyright (c) .NET Foundation. All rights reserved.

src/WebJobs.Script.WebHost/Models/FunctionsWorkerContainerAssignmentContext.cs
```diff
@@ -0,0 +1,14 @@
...
+// Copyright (c) Microsoft Corporation.  All rights reserved.
+// -----------------------------------------------------------
+
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Models
+{
+    public class FunctionsWorkerContainerAssignmentContext
+    {
+        [JsonProperty("assignmentContext")]
```
RohitRanjanMS: Consider DefaultValueHandling.Ignore

RohitRanjanMS (changes requested):

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
mathewc: Agreed - we don't want this in our API surface area. If it ends up we actually need to support both encrypted and non-encrypted payloads, then we could just handle that on the existing API via an extra param or header that indicates whether the content is encrypted. We'd then just handle the deserialization in the method body conditionally.

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
mathewc: Right - this is a HostAssignmentContext, right? I guess this points to a lack of tests in this PR - as you were writing the test you'd realize the mistake

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
```diff
@@ -41,27 +41,23 @@ public async Task<IActionResult> Assign([FromBody] EncryptedHostAssignmentContex
 
             var assignmentContext = _startupContextProvider.SetContext(encryptedAssignmentContext);
 
-            // before starting the assignment we want to perform as much
```
mathewc, line 44: There are existing tests e.g. end to end test StandbyModeE2E_LinuxContainer as well as a suite of unit tests InstanceControllerTests for the assignment flow. We need new tests.

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
manikantanallagatla: I think so. When assign is called twice. Anyways, did not touch these existing code in latest changes

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
manikantanallagatla: Removed it

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
manikantanallagatla: Current code. Not modifying it in latest changes

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
manikantanallagatla: Having 1 API now

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
manikantanallagatla: Removed it

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
manikantanallagatla: Added tests.

src/WebJobs.Script.WebHost/Models/FunctionsWorkerContainerAssignmentContext.cs
manikantanallagatla: Added

src/WebJobs.Script.WebHost/Models/FunctionsWorkerContainerAssignmentContext.cs
manikantanallagatla: removed this file

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
manikantanallagatla: Mostly it should not be null. But, it was there even before my changes. So, did not edit it

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
manikantanallagatla: It is the same as existing admin/instance/assign but with unencrypted payload. Anyways removed this new API

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
manikantanallagatla: I see `_logger.LogDebug` being used in many places: https://github.com/search?q=repo%3AAzure%2Fazure-functions-host%20_logger.LogDebug&type=code

I see https://github.com/search?q=repo%3AAzure%2Fazure-functions-host+StructuredLogEntry&type=code in very few places.

We want to change this log as part of this change? This log was there before as well

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
```diff
@@ -35,11 +35,28 @@ public InstanceController(IEnvironment environment, IInstanceManager instanceMan
...
+        public async Task<IActionResult> Assign([FromBody] HostAssignmentRequest hostAssignmentContext)
         {
-            _logger.LogDebug($"Starting container assignment for host : {Request?.Host}. ContextLength is: {encryptedAssignmentContext.EncryptedContext?.Length}");
+            if (string.IsNullOrEmpty(hostAssignmentContext.EncryptedContext) &&
+                hostAssignmentContext.AssignmentContext == null)
+            {
+                return BadRequest("Atleast one of Assignment context and EncryptedContext needs to be set.");
+            }
+            if (!string.IsNullOrEmpty(hostAssignmentContext.EncryptedContext) &&
+                !(hostAssignmentContext.AssignmentContext == null))
```
RohitRanjanMS: nit: you can use pattern matching or consider changing it to `is not null`.

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
```diff
@@ -35,11 +35,28 @@ public InstanceController(IEnvironment environment, IInstanceManager instanceMan
...
         [Route("admin/instance/assign")]
         [Authorize(Policy = PolicyNames.AdminAuthLevel)]
-        public async Task<IActionResult> Assign([FromBody] EncryptedHostAssignmentContext encryptedAssignmentContext)
+        public async Task<IActionResult> Assign([FromBody] HostAssignmentRequest hostAssignmentContext)
         {
-            _logger.LogDebug($"Starting container assignment for host : {Request?.Host}. ContextLength is: {encryptedAssignmentContext.EncryptedContext?.Length}");
+            if (string.IsNullOrEmpty(hostAssignmentContext.EncryptedContext) &&
+                hostAssignmentContext.AssignmentContext == null)
+            {
+                return BadRequest("Atleast one of Assignment context and EncryptedContext needs to be set.");
```
RohitRanjanMS: `At least`

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
RohitRanjanMS: I'd appreciate it if you could use structured logging—especially when you're touching the code and there's an opportunity to improve it.

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
```diff
@@ -35,11 +35,28 @@ public InstanceController(IEnvironment environment, IInstanceManager instanceMan
...
+            }
+            if (!string.IsNullOrEmpty(hostAssignmentContext.EncryptedContext) &&
+                !(hostAssignmentContext.AssignmentContext == null))
+            {
+                return BadRequest("Only one of Assignment context and EncryptedContext needs to be set.");
```
jviau:
```suggestion
                return BadRequest("Only one of 'assignmentContext' or 'encryptedContext' may be set.");
```

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
jviau:
```suggestion
                return BadRequest("At least one of 'assignmentContext' or 'encryptedContext' must be provided.");
```

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
```diff
@@ -35,11 +35,28 @@ public InstanceController(IEnvironment environment, IInstanceManager instanceMan
...
+            }
+            if (!string.IsNullOrEmpty(hostAssignmentContext.EncryptedContext))
+            {
+                _logger.LogDebug($"Starting container assignment for host : {Request?.Host}. ContextLength is {hostAssignmentContext.EncryptedContext.Length}");
```
jviau: Use structured logging (and not string interpolation). We also don't need to include `Request.Host`, as the site name will already be a resource-level property on the logs. We can then infer unencrypted context by the absence of the ContextLength.


``` csharp
if (!string.IsNullOrEmpty(hostAssignmentContext.EncryptedContext)
    _logger.LogDebug("Starting container assignment. ContextLength is {ContextLength}", hostAssignmentContext.EncryptedContext.Length);
}
else
{
     _logger.LogDebug("Starting container assignment.");
}
```

test/WebJobs.Script.Tests.Integration/Host/StandbyManager/StandbyManagerE2ETests_Linux.cs
```diff
@@ -260,32 +262,36 @@ private async Task Assign(string encryptionKey)
...
             request.Content = new StringContent(json, Encoding.UTF8, "application/json");
             request.Headers.Add(AuthenticationLevelHandler.FunctionsKeyHeaderName, masterKey);
             var response = await _httpClient.SendAsync(request);
             Assert.Equal(HttpStatusCode.Accepted, response.StatusCode);
         }
 
-        private static EncryptedHostAssignmentContext CreateEncryptedContext(HostAssignmentContext context, string key)
+        private static HostAssignmentRequest CreateHostAssignmentRequest(HostAssignmentContext context, string key, bool useEncryptedPayload)
         {
             string json = JsonConvert.SerializeObject(context);
```
jviau, line 291: nit: can shortcircuit here:

``` CSharp
if (!useEncryptedPayload)
{
    return new() { AssignmentContext  context }:
}

var encryptionKey = Convert.FromBase64String(key);
string encrypted = EncryptionHelper.Encrypt(json, encryptionKey);
return new() { EncryptedContext = encrypted };
```

manikantanallagatla: Updated

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
manikantanallagatla: Updated thanks

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
manikantanallagatla: Updated

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
manikantanallagatla: Updated

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
manikantanallagatla: This is removed

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
manikantanallagatla: Updated

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
manikantanallagatla: Removed it

manikantanallagatla: Updated it

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
manikantanallagatla: Updated

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
```diff
@@ -35,11 +36,34 @@ public InstanceController(IEnvironment environment, IInstanceManager instanceMan
...
-        public async Task<IActionResult> Assign([FromBody] EncryptedHostAssignmentContext encryptedAssignmentContext)
+        public async Task<IActionResult> Assign([FromBody] HostAssignmentRequest hostAssignmentContext)
         {
-            _logger.LogDebug($"Starting container assignment for host : {Request?.Host}. ContextLength is: {encryptedAssignmentContext.EncryptedContext?.Length}");
+            if (string.IsNullOrEmpty(hostAssignmentContext.EncryptedContext) &&
+                hostAssignmentContext.AssignmentContext is null)
+            {
+                return BadRequest("At least one of 'assignmentContext' or 'encryptedContext' must be provided.");
+            }
+            if (!string.IsNullOrEmpty(hostAssignmentContext.EncryptedContext) &&
```
jviau: style nit: empty newline after a closing brace (except for when it is an `else` or `else if` block)

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
```diff
@@ -35,11 +36,34 @@ public InstanceController(IEnvironment environment, IInstanceManager instanceMan
...
+            {
+                _logger.LogDebug("Starting container assignment. ContextLength is {ContextLength}", hostAssignmentContext.EncryptedContext.Length);
+            }
+            else
+            {
+                if (User != null &&
+                    !User.HasClaim(c => c.Type == SecurityConstants.AssignUnencryptedClaimType && c.Value == "true"))
+                {
+                    _logger.LogWarning("Required claims missing for invoking unencrypted assignment");
+                    return StatusCode(StatusCodes.Status403Forbidden, new { status = "Required claims missing for invoking unencrypted assignment" });
```
jviau: No need to include the response object. With AuthN/AuthZ it is better to be vague / give no information in responses.

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
```diff
@@ -35,11 +36,34 @@ public InstanceController(IEnvironment environment, IInstanceManager instanceMan
...
+            {
+                return BadRequest("Only one of 'assignmentContext' or 'encryptedContext' may be set.");
+            }
+            if (!string.IsNullOrEmpty(hostAssignmentContext.EncryptedContext))
+            {
+                _logger.LogDebug("Starting container assignment. ContextLength is {ContextLength}", hostAssignmentContext.EncryptedContext.Length);
+            }
+            else
+            {
+                if (User != null &&
```
jviau: I would not check for `User != null`. While it will never be null, because we require admin auth (which then enforces user is authenticated and will exist), we still want to be clear in this code we expect it.

I would just check the claim and let it throw if User isn't present (as it is an unexpected case and is a true `500`).

```suggestion
                if (!User.HasClaim(c => c.Type == SecurityConstants.AssignUnencryptedClaimType && c.Value == "true"))
```

src/WebJobs.Script.WebHost/Security/Authorization/AuthUtility.cs
```diff
@@ -69,5 +69,10 @@ public static bool PrincipalHasInvokeClaim(ClaimsPrincipal principal, Authorizat
 
             return principal.HasClaim(SecurityConstants.InvokeClaimType, "true");
         }
+
+        public static bool PrincipalHasUnencryptedAssignClaim(ClaimsPrincipal principal, AuthorizationLevel requiredLevel)
```
jviau: This is not used anywhere. Can it be removed, or did you intend to use it in the AuthZ check in the controller?

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
manikantanallagatla: The user != null is needed for UTs. Where we are calling assign directly: https://github.com/Azure/azure-functions-host/blob/dev/test/WebJobs.Script.Tests.Integration/Management/InstanceControllerTests.cs#L90

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
manikantanallagatla: Updated

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
manikantanallagatla: Just "Required claims missing" is fine? Or you mean dont set anything?

src/WebJobs.Script.WebHost/Security/Authorization/AuthUtility.cs
manikantanallagatla: Removed

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
jviau: You can add a user to the test by assigning `instanceController.ControllerContext`

``` csharp
DefaultHttpContext context = new() { User = <create user> };
instanceController.ControllerContext = new() { HttpContext = context };
```

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
jviau: Don't send any body. Just a `return Forbid();`

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
manikantanallagatla: Updated

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
manikantanallagatla: Updated

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
```diff
@@ -35,11 +36,35 @@ public InstanceController(IEnvironment environment, IInstanceManager instanceMan
...
+            if (!string.IsNullOrEmpty(hostAssignmentContext.EncryptedContext))
+            {
+                _logger.LogDebug("Starting container assignment. ContextLength is {ContextLength}", hostAssignmentContext.EncryptedContext.Length);
+            }
+            else
+            {
+                if (!User.HasClaim(c => c.Type == SecurityConstants.AssignUnencryptedClaimType && c.Value == "true"))
+                {
+                    _logger.LogWarning("Required claims missing for invoking unencrypted assignment");
+                    return StatusCode(StatusCodes.Status403Forbidden);
```
jviau: Tiny nit. There are a bunch of helper methods from `ControllerBase`. Or is there a reason you don't want to use `Forbid()`?

```suggestion
                    return Forbid();
```

jviau (approved):

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
manikantanallagatla: Done

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
```diff
@@ -35,11 +36,35 @@ public InstanceController(IEnvironment environment, IInstanceManager instanceMan
...
         [Route("admin/instance/assign")]
         [Authorize(Policy = PolicyNames.AdminAuthLevel)]
-        public async Task<IActionResult> Assign([FromBody] EncryptedHostAssignmentContext encryptedAssignmentContext)
+        public async Task<IActionResult> Assign([FromBody] HostAssignmentRequest hostAssignmentContext)
         {
-            _logger.LogDebug($"Starting container assignment for host : {Request?.Host}. ContextLength is: {encryptedAssignmentContext.EncryptedContext?.Length}");
+            if (string.IsNullOrEmpty(hostAssignmentContext.EncryptedContext) &&
+                hostAssignmentContext.AssignmentContext is null)
+            {
+                return BadRequest("At least one of 'assignmentContext' or 'encryptedContext' must be provided.");
```
mathewc, line 44: I think you should verify that one and only one of these is specified. E.g. it'd also be invalid if BOTH are specified. So you should verify this, and error message when violated would be like "Exactly one of {nameof(HostAssignmentContext.EncryptedContext)} or {nameof(HostAssignmentContext.AssignmentContext)}" must be specified."

mathewc: can you use nameof expressions for these property names, rather than hardcoded strings, here and below?

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
```diff
@@ -35,11 +36,35 @@ public InstanceController(IEnvironment environment, IInstanceManager instanceMan
...
+
+            if (!string.IsNullOrEmpty(hostAssignmentContext.EncryptedContext) &&
+                hostAssignmentContext.AssignmentContext is not null)
+            {
+                return BadRequest("Only one of 'assignmentContext' or 'encryptedContext' may be set.");
+            }
+
+            if (!string.IsNullOrEmpty(hostAssignmentContext.EncryptedContext))
+            {
+                _logger.LogDebug("Starting container assignment. ContextLength is {ContextLength}", hostAssignmentContext.EncryptedContext.Length);
```
mathewc: Any reason why you removed the logging of Host that was there before?

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
```diff
@@ -35,11 +36,35 @@ public InstanceController(IEnvironment environment, IInstanceManager instanceMan
...
+            }
+            else
+            {
+                if (!User.HasClaim(c => c.Type == SecurityConstants.AssignUnencryptedClaimType && c.Value == "true"))
```
mathewc: Can't you use the overload HasClaim(SecurityConstants.AssignUnencryptedClaimType, "true")

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
```diff
@@ -35,11 +36,35 @@ public InstanceController(IEnvironment environment, IInstanceManager instanceMan
         [HttpPost]
         [Route("admin/instance/assign")]
         [Authorize(Policy = PolicyNames.AdminAuthLevel)]
-        public async Task<IActionResult> Assign([FromBody] EncryptedHostAssignmentContext encryptedAssignmentContext)
+        public async Task<IActionResult> Assign([FromBody] HostAssignmentRequest hostAssignmentContext)
```
mathewc: forgot to rename the param to match the new Type name

src/WebJobs.Script.WebHost/ContainerManagement/LinuxContainerInitializationHostedService.cs
```diff
@@ -48,8 +48,8 @@ private async Task ApplyStartContextIfPresent(CancellationToken cancellationToke
             {
                 _logger.LogDebug("Applying host context");
 
-                var encryptedAssignmentContext = JsonConvert.DeserializeObject<EncryptedHostAssignmentContext>(startContext);
-                var assignmentContext = _startupContextProvider.SetContext(encryptedAssignmentContext);
+                var hostAssignmentRequest = JsonConvert.DeserializeObject<HostAssignmentRequest>(startContext);
```
mathewc, line 51: The above TryGetStartContextOrNullAsync virtual call has a different implementation on Legion and Atlas. This same hosted service code is running on each SKU. Just want to confirm that there isn't any breaking format change here - the persisted assignment context is always encrypted.

src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
```diff
@@ -71,6 +71,10 @@ private static void ConfigureOptions(JwtBearerOptions options)
                     {
                         claims.Add(new Claim(SecurityConstants.InvokeClaimType, "true"));
                     }
+                    if (string.Equals(c.SecurityToken.Issuer, ScriptConstants.LegionCoreUri, StringComparison.OrdinalIgnoreCase))
```
mathewc, line 75: nit: missing a space here

src/WebJobs.Script/ScriptConstants.cs
```diff
@@ -151,6 +151,7 @@ public static class ScriptConstants
         public const string ScmSiteUriFormat = "https://{0}.scm.azurewebsites.net";
         public const string SiteUriFormat = "https://{0}.azurewebsites.net";
         public const string AppServiceCoreUri = "https://appservice.core.azurewebsites.net";
+        public const string LegionCoreUri = "https://legion.core.azurewebsites.net";
```
mathewc, line 155: Isn't Legion already calling /admin APIs with tokens? What issuer is it using currently - app service? I assume so.

So you're planning to change the issuer currently used by Legion to a new one? We'd have to make sure these host changes rolled out fully before the legion change, otherwise we could get auth failures.

test/WebJobs.Script.Tests.Integration/Host/StandbyManager/StandbyManagerE2ETests_Linux.cs
```diff
@@ -114,7 +116,7 @@ public async Task StandbyModeE2E_LinuxContainer()
             await VerifyWarmupSucceeds(restart: true);
 
             // now specialize the site
-            await Assign(encryptionKey);
+            await Assign(encryptionKey, useEncyptedPayload);
```
mathewc: spelling

src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
```diff
@@ -149,6 +153,7 @@ public static TokenValidationParameters CreateTokenValidationParameters()
                 result.ValidIssuers =
                 [
                     AppServiceCoreUri,
+                    LegionCoreUri, // ATOL: Specialization is invoked from Legion.
```
mathewc: nit: remove this comment. This general token auth isn't specific to any particular /admin operation that might be invoked from Legion. E.g., in addition to /assign, we'll also be invoking /drain and other APIs.

src/WebJobs.Script.WebHost/Security/Authentication/SecurityConstants.cs
```diff
@@ -20,5 +20,10 @@ public class SecurityConstants
         /// /admin/functions/{function} API.
         /// </summary>
         public const string InvokeClaimType = "http://schemas.microsoft.com/2017/07/functions/claims/invoke";
+
+        /// <summary>
+        /// Claim indicating whether a principal is authorized to invoke assign with unencrypted payload.
```
mathewc: nit: with an

src/WebJobs.Script.WebHost/StartupContextProvider.cs
```diff
@@ -135,12 +135,17 @@ private StartupContext GetStartupContextOrNull()
         /// Decrypt and deserialize the specified context, and apply values from it to the
         /// startup cache context.
         /// </summary>
-        /// <param name="encryptedContext">The encrypted assignment context.</param>
```
mathewc, line 138: Let's fix the summary comments as well. E.g. something like "Applies the values from the specified assignment request to the startup cache context, performing any required decryption."

src/WebJobs.Script.WebHost/StartupContextProvider.cs
```diff
@@ -135,12 +135,17 @@ private StartupContext GetStartupContextOrNull()
...
-        /// <returns>The decrypted assignment context</returns>
-        public virtual HostAssignmentContext SetContext(EncryptedHostAssignmentContext encryptedContext)
+        /// <param name="hostAssignmentRequest">The Host assignment request.</param>
+        /// <returns>The decrypted assignment context.</returns>
```
mathewc: returns: "The assignment context applied."

I.e. since we might not have performed decryption, I'd just omit that

src/WebJobs.Script.WebHost/StartupContextProvider.cs
```diff
@@ -135,12 +135,17 @@ private StartupContext GetStartupContextOrNull()
...
+        public virtual HostAssignmentContext SetContext(HostAssignmentRequest hostAssignmentRequest)
         {
-            string decryptedContext = EncryptionHelper.Decrypt(encryptedContext.EncryptedContext, environment: _environment);
-            var hostAssignmentContext = JsonConvert.DeserializeObject<HostAssignmentContext>(decryptedContext);
+            var hostAssignmentContext = hostAssignmentRequest.AssignmentContext;
+
+            if (!string.IsNullOrEmpty(hostAssignmentRequest.EncryptedContext))
```
mathewc: I wonder if you shouldn't add a clause hostAssignmentContext == null to this if, making it clear that only one of these should ever be specified. As opposed to the precedence handling you have here, where both can be specified, but encrypted wins.

I realize in the controller method you do validation, but there's another path where this method is called, where the request is pulled from storage.

src/WebJobs.Script/ScriptConstants.cs
mathewc: Got clarification that currently Legion does not yet form its own tokens, so this will all be green field. Still, you'll need to be sure all Flex pods are running this new version before you start sending tokens with the new issuer.

manikantanallagatla: yes, we will wait for host release before making images changes which use these changes

src/WebJobs.Script.WebHost/StartupContextProvider.cs
manikantanallagatla: I am not sure how the storage path works. But shall we move the handling that only one needs to be specified here?

src/WebJobs.Script.WebHost/StartupContextProvider.cs
manikantanallagatla: Updated

src/WebJobs.Script.WebHost/StartupContextProvider.cs
manikantanallagatla: Updated

src/WebJobs.Script.WebHost/Security/Authentication/SecurityConstants.cs
manikantanallagatla: Updated

src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
manikantanallagatla: Removed

test/WebJobs.Script.Tests.Integration/Host/StandbyManager/StandbyManagerE2ETests_Linux.cs
manikantanallagatla: Updated

src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
manikantanallagatla: Added

src/WebJobs.Script.WebHost/ContainerManagement/LinuxContainerInitializationHostedService.cs
manikantanallagatla: I tried for legion changes by giving encrypted and unencrypted payload for controller and it is working fine. Means  encrypted path of controller which is existing is working fine.

Any way to test this storage path?

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
manikantanallagatla: Updated

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
manikantanallagatla: Yes, updated it

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
manikantanallagatla: Removed based on Jacob's comment: https://github.com/Azure/azure-functions-host/pull/11302#discussion_r2361330189 `We also don't need to include Request.Host, as the site name will already be a resource-level property on the logs.`

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
manikantanallagatla: Updated. Hope String Interpolation is okay in this case as this is not log and it takes string

src/WebJobs.Script.WebHost/StartupContextProvider.cs
manikantanallagatla: If we are sure storage account path only takes encrypted payload, we can have method overloading for SetContext method. One which takes encrypted payload and other that takes unencrypted payload

src/WebJobs.Script.WebHost/ContainerManagement/LinuxContainerInitializationHostedService.cs
kshyju, line 51: Could we switch the deserialization here from JSON.NET to System.Text.Json [with source generation](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/source-generation)? We’ve also been gradually moving away from JSON.NET, so this would fit with that direction.

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
```diff
@@ -35,11 +36,35 @@ public InstanceController(IEnvironment environment, IInstanceManager instanceMan
         [HttpPost]
         [Route("admin/instance/assign")]
         [Authorize(Policy = PolicyNames.AdminAuthLevel)]
-        public async Task<IActionResult> Assign([FromBody] EncryptedHostAssignmentContext encryptedAssignmentContext)
+        public async Task<IActionResult> Assign([FromBody] HostAssignmentRequest hostAssignmentRequest)
         {
-            _logger.LogDebug($"Starting container assignment for host : {Request?.Host}. ContextLength is: {encryptedAssignmentContext.EncryptedContext?.Length}");
+            if (string.IsNullOrEmpty(hostAssignmentRequest.EncryptedContext) &&
```
kshyju: Is it possible that `hostAssignmentRequest` itself could be null? Should we null check before accessing the properties?

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
```diff
@@ -35,11 +36,35 @@ public InstanceController(IEnvironment environment, IInstanceManager instanceMan
...
+                return BadRequest($"Only one of {nameof(HostAssignmentRequest.AssignmentContext)} or {nameof(HostAssignmentRequest.EncryptedContext)} may be set.");
+            }
+
+            if (!string.IsNullOrEmpty(hostAssignmentRequest.EncryptedContext))
+            {
+                _logger.LogDebug("Starting container assignment. ContextLength is {ContextLength}", hostAssignmentRequest.EncryptedContext.Length);
+            }
+            else
+            {
+                if (!User.HasClaim(SecurityConstants.AssignUnencryptedClaimType, "true"))
```
kshyju: Does this method do case sensitive check? Would that be a concern?

src/WebJobs.Script.WebHost/Models/HostAssignmentRequest.cs
```diff
@@ -5,9 +5,12 @@
 
 namespace Microsoft.Azure.WebJobs.Script.WebHost.Models
 {
-    public class EncryptedHostAssignmentContext
+    public class HostAssignmentRequest
```
kshyju: If we don't have a use case today to inherit this class, let's mark it `sealed`. We can always open it up if needed in the future.

src/WebJobs.Script.WebHost/StartupContextProvider.cs
```diff
@@ -132,15 +132,20 @@ private StartupContext GetStartupContextOrNull()
...
+            {
+                string decryptedContext = EncryptionHelper.Decrypt(hostAssignmentRequest.EncryptedContext, environment: _environment);
+                hostAssignmentContext = JsonConvert.DeserializeObject<HostAssignmentContext>(decryptedContext);
```
kshyju, line 153: Is this line executed during cold start? See if you can switch this to STJ with source generation.

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
```diff
@@ -35,11 +36,35 @@ public InstanceController(IEnvironment environment, IInstanceManager instanceMan
         [HttpPost]
         [Route("admin/instance/assign")]
         [Authorize(Policy = PolicyNames.AdminAuthLevel)]
-        public async Task<IActionResult> Assign([FromBody] EncryptedHostAssignmentContext encryptedAssignmentContext)
+        public async Task<IActionResult> Assign([FromBody] HostAssignmentRequest hostAssignmentRequest)
         {
```
kshyju, line 43: Should we wrap this entire logic inside `IMetricsLogger.LatencyEvent`? @safihamid

src/WebJobs.Script.WebHost/ContainerManagement/LinuxContainerInitializationHostedService.cs
mathewc: I'm just pointing out how the system behaves. Likely things are OK, since up to this point only encrypted was supported, and your new deserialization supports both old and new formats. I'm just this out so you can think about this and ensure no breaks. Bala knows the context storage logic the best so you might have him sign off on this as well.

mathewc: I think we should handle any large serializer changes like that as a separate PR, to keep this PR focused.

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
mathewc: While the previous implementation also didn't check for null (so things are not any worse in this PR), you're right that it could be null. E.g. if the caller sends no body it could be null.

Can just add a simple check, like we do in some other controller methods, e.g.:

```
if (hostAssignmentContext == null)
{
    return BadRequest();
}
```

src/WebJobs.Script.WebHost/StartupContextProvider.cs
mathewc: The storage path where context is pulled from storage MUST be encrypted since data must be encrypted at rest.

src/WebJobs.Script.WebHost/ContainerManagement/LinuxContainerInitializationHostedService.cs
kshyju: Agree.

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
manikantanallagatla: Updated

src/WebJobs.Script.WebHost/StartupContextProvider.cs
manikantanallagatla: Can you create  a task for this? Touching all paths at once will increase blast radius. Out intention was to add unencryption support without touching current path(encryption path) which has been there for ages as part of this PR.

Anyways once we once we move to unencrypted path, we will anyways stop using encrypted path. So, making changes to old path will anyways be not usable

src/WebJobs.Script.WebHost/Models/HostAssignmentRequest.cs
manikantanallagatla: Updated

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
manikantanallagatla: We are adding the string in ScriptJwtBearerExtensions and checking here. So, we dont need case sensitive check right

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
manikantanallagatla: Updated

src/WebJobs.Script.WebHost/ContainerManagement/LinuxContainerInitializationHostedService.cs
manikantanallagatla: Yes

manikantanallagatla: Sure. Will check with Bala

src/WebJobs.Script.WebHost/StartupContextProvider.cs
jviau: > The storage path where context is pulled from storage MUST be encrypted since data must be encrypted at rest.

This is a bit besides the point, but is this the customer's storage account? The encrypted at rest requirement here would be the storage account itself satisfying that. _All_ data in a storage account is already "encrypted at rest": https://learn.microsoft.com/en-us/azure/storage/common/storage-service-encryption.

The only time we, the Functions Host team, would have an "encrypted at rest" requirement is if we were offering some storage solution which did not already have encrypted-at-rest features.

But back to the main point: I agree that if we expect storage scenario to always have client-side encryption (which is what this is), then we should assert that as necessary.

src/WebJobs.Script.WebHost/ContainerManagement/LinuxContainerInitializationHostedService.cs
manikantanallagatla: Storage one is only used in CV1 on ACI path(Atlas). We are not changing anything wrt encryption payload. So, storage path should not be affected by this repo. They should continue the encrypted payload and no changes needed.

For legion path, technically it should work as we are storing the volume in https://msazure.visualstudio.com/One/_git/AAPT-Antares-Functions-Docker?path=/images/flexconsumption/components/appserver/cv2/pkg/instance/instance.go&version=GBdev&line=76&lineEnd=77&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

I will check legion for both encrypted and unencrypted path

src/WebJobs.Script.WebHost/StartupContextProvider.cs
manikantanallagatla: Storage is only used in CV1 on ACI path. We are not touching that path. So, storage wise, no change with this PR

manikantanallagatla: I think no changes are needed for storage and current validations in [StartupContextProvider.cs](https://github.com/Azure/azure-functions-host/pull/11302/files/fded652230131dd1825ebba1e8b9c7ca91cf4550#diff-7577fa045b0a1f6071f1a5200789f169cd6adee86cfba37cd3e6daa3341fb6a0) seems fine

src/WebJobs.Script.WebHost/StartupContextProvider.cs
manikantanallagatla: will let shyju create a task for this. resolving this for this PR

src/WebJobs.Script.WebHost/ContainerManagement/LinuxContainerInitializationHostedService.cs
manikantanallagatla: Have tested both paths and they are working fine.

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
```diff
@@ -21,47 +23,81 @@ public class InstanceController : Controller
...
+            using (_metricsLogger.LatencyEvent(MetricEventNames.LinuxContainerSpecializationMSIInit))
+            {
+                if (hostAssignmentRequest == null)
+                {
+                    return BadRequest($"{nameof(hostAssignmentRequest)} cannot be null.");
+                }
 
-            var assignmentContext = _startupContextProvider.SetContext(encryptedAssignmentContext);
+                if (string.IsNullOrEmpty(hostAssignmentRequest.EncryptedContext) &&
+                hostAssignmentRequest.AssignmentContext is null)
```
mathewc: nit: formatting is off on the second line here

src/WebJobs.Script.WebHost/StartupContextProvider.cs
```diff
@@ -132,15 +132,20 @@ private StartupContext GetStartupContextOrNull()
...
+            }
```
liliankasem: I don't love assigning `hostAssignmentContext` to just potentially overwrite it down the line. I would suggest maybe something like

```csharp
HostAssignmentContext hostAssignmentContext = null;

// If EncryptedContext is set, use that
if (!string.IsNullOrEmpty(hostAssignmentRequest.EncryptedContext))
{
    string decryptedContext = EncryptionHelper.Decrypt(
        hostAssignmentRequest.EncryptedContext,
        environment: _environment);

    hostAssignmentContext = JsonConvert.DeserializeObject<HostAssignmentContext>(decryptedContext);
}
// Otherwise fall back to plain AssignmentContext if we didn't get one from decryption
else
{
    hostAssignmentContext = hostAssignmentRequest.AssignmentContext;
}
```

test/WebJobs.Script.Tests.Integration/Management/InstanceControllerTests.cs
```diff
@@ -238,7 +243,12 @@ public async Task Assignment_Invokes_InstanceManager_Methods_For_Warmup_Requests
             instanceManager.Reset();
 
             var instanceController = new InstanceController(environment, instanceManager.Object, loggerFactory,
-                startupContextProvider);
+                startupContextProvider, new TestMetricsLogger());
+            DefaultHttpContext context = new() { User = new ClaimsPrincipal(new ClaimsIdentity(new Claim[]
+            {
+                new Claim(SecurityConstants.AssignUnencryptedClaimType, "true")
```
mathewc, line 255: I don't see any tests where this claim is missing, to verify the 403 Forbidden codepath. You should add a new test for this

src/WebJobs.Script.WebHost/StartupContextProvider.cs
mathewc: I was simply suggesting something simple like below here, since as it stands, your code allows both to be specified, and defines a precedence if they are both specified (encrypted wins). Currently SetContext is called from two places in the code. We only expect one or the other to ever be specified, so why not be explicit about that.

```
HostAssignmentContext hostAssignmentContext = null;
if (hostAssignmentRequest.AssignmentContext != null)
{
    hostAssignmentContext = hostAssignmentRequest.AssignmentContext;
}
else if (!string.IsNullOrEmpty(hostAssignmentRequest.EncryptedContext))
{
    string decryptedContext = EncryptionHelper.Decrypt(hostAssignmentRequest.EncryptedContext, environment: _environment);
    hostAssignmentContext = JsonConvert.DeserializeObject<HostAssignmentContext>(decryptedContext);
}

```

test/WebJobs.Script.Tests.Integration/Host/StandbyManager/StandbyManagerE2ETests_Linux.cs
```diff
@@ -260,32 +262,36 @@ private async Task Assign(string encryptionKey)
...
+                { EnvironmentSettingNames.FunctionWorkerRuntime, "node" }
+            };
             var assignmentContext = new HostAssignmentContext
             {
                 SiteId = 1234,
                 SiteName = "TestApp",
                 Environment = environment
             };
-            var encryptedAssignmentContext = CreateEncryptedContext(assignmentContext, encryptionKey);
+            var encryptedAssignmentContext = CreateHostAssignmentRequest(assignmentContext, encryptionKey, useEncryptedPayload);
```
mathewc: Recommend renaming the var to hostAssignmentRequest to match the refactoring

test/WebJobs.Script.Tests.Integration/Management/InstanceControllerTests.cs
```diff
@@ -195,7 +198,7 @@ public async Task Assignment_Does_Not_Set_Secrets_Context_For_Warmup_Request()
 
             instanceManager.Reset();
 
-            var instanceController = new InstanceController(environment, instanceManager, loggerFactory, startupContextProvider);
+            var instanceController = new InstanceController(environment, instanceManager, loggerFactory, startupContextProvider, new TestMetricsLogger());
```
mathewc: Do you verify your new metric anywhere? That would be a simple thing to do in one of your existing tests. If you cared enough to add it, would be good to have a regression test for it.

test/WebJobs.Script.Tests.Integration/Management/InstanceControllerTests.cs
```diff
@@ -266,5 +280,43 @@ public async Task Assignment_Invokes_InstanceManager_Methods_For_Warmup_Requests
             instanceManager.Verify(i => i.StartAssignment(It.IsAny<HostAssignmentContext>()),
                 shouldInvokeMethod ? Times.Once() : Times.Never());
         }
+
+        [Fact]
+        public async Task Assignment_ErrorScenarios()
```
mathewc: nit: Assignment_InvalidInput_ReturnsBadRequest

test/WebJobs.Script.Tests/StartupContextProviderTests.cs
```diff
@@ -188,7 +188,7 @@ public void SetContext_AppliesHostAssignmentContext()
             };
             string json = JsonConvert.SerializeObject(context);
             string encrypted = EncryptionHelper.Encrypt(json, environment: _environment);
-            var encryptedContext = new EncryptedHostAssignmentContext { EncryptedContext = encrypted };
+            var encryptedContext = new HostAssignmentRequest { EncryptedContext = encrypted };
```
mathewc: rename vars to match type

manikantanallagatla: Updated

test/WebJobs.Script.Tests.Integration/Management/InstanceControllerTests.cs
manikantanallagatla: Updated

test/WebJobs.Script.Tests.Integration/Management/InstanceControllerTests.cs
manikantanallagatla: added verification

test/WebJobs.Script.Tests.Integration/Host/StandbyManager/StandbyManagerE2ETests_Linux.cs
manikantanallagatla: renamed

test/WebJobs.Script.Tests.Integration/Management/InstanceControllerTests.cs
manikantanallagatla: Added a forbid test now

src/WebJobs.Script.WebHost/StartupContextProvider.cs
manikantanallagatla: Sure, Updated it

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
manikantanallagatla: Updated it

src/WebJobs.Script.WebHost/StartupContextProvider.cs
manikantanallagatla: Ohh. got it. Lilian also suggested similar change. Updated it in this order

mathewc (approved):

RohitRanjanMS (approved):

src/WebJobs.Script.WebHost/StartupContextProvider.cs
pragnagopa: I see this is addressed --- resolving comment

PR #11313, opened 2025-09-10
Setting current activity status.

### LINKED ISSUE #11314: OTel - Report Failed Executions as "Failure"
Azure Function Apps when using the Service Bus trigger or output binding. In certain scenarios, if the function execution encounters an error, the associated request is still reported as successful in telemetry and monitoring systems. As a result, failures may be hidden from logs, dashboards, or alerting pipelines, leading to inaccurate success rates and potentially masking operational issues.

RohitRanjanMS:
Azure Function Apps when using the Service Bus trigger or output binding. In certain scenarios, if the function execution encounters an error, the associated request is still reported as successful in telemetry and monitoring systems. As a result, failures may be hidden from logs, dashboards, or alerting pipelines, leading to inaccurate success rates and potentially masking operational issues.

resolves #11314
### Pull request checklist

### Additional information

Additional PR information


test/WebJobs.Script.Tests/Workers/Rpc/GrpcWorkerChannelTests.cs
```diff
@@ -1705,6 +1727,18 @@ private static InvocationResponse BuildSuccessfulInvocationResponse(string invoc
             };
         }
 
+        private static InvocationResponse BuildSFailureInvocationResponse(string invocationId)
```
satvu: nit: typo?

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -1727,6 +1726,39 @@ private void AddAdditionalTraceContext(InvocationRequest invocationRequest, Scri
...
+            {
+                return;
+            }
+
+            // Restore the execution context from the original invocation. This allows AsyncLocal state to flow to loggers.
+            System.Threading.ExecutionContext.Run(
+                context.AsyncExecutionContext,
+                static state =>
+                {
+                    var invocationState = (InvocationState)state!;
```
jviau: `Activity.Current` is an AsyncLocal itself. Will setting `AsyncExecutionContext` not potentially change what the `Acitivty.Current` on line 1732 and within this delegate be?

brettsam: Yeah I think you're right -- the Activity.Current up there is different than here b/c they're running in different contexts. Probably need to remove the null check on 1732?

RohitRanjanMS: Removed the Activity.Current null check.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -1727,6 +1726,39 @@ private void AddAdditionalTraceContext(InvocationRequest invocationRequest, Scri
             }
         }
 
+        private void SetCurrentActivityStatus(ScriptInvocationContext context, Exception exception)
```
brettsam: nit -- I think I'd rename this method to something like `SetInvocationActivityError` b/c it's really only for errors (and it's not technically setting the *current* activity from the perspective of the caller).

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -1727,6 +1726,39 @@ private void AddAdditionalTraceContext(InvocationRequest invocationRequest, Scri
...
+        {
+            // If AsyncExecutionContext is null, skip running.
+            if (context?.AsyncExecutionContext is null)
```
brettsam: nit -- just to be defensive, maybe bail if `exception` is null? Or is there still value in setting the Activity status to `Error` w/o an exception? There may be...

brettsam (approved):
just a couple small comments

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -1727,6 +1726,39 @@ private void AddAdditionalTraceContext(InvocationRequest invocationRequest, Scri
...
+            // Restore the execution context from the original invocation. This allows AsyncLocal state to flow to loggers.
+            System.Threading.ExecutionContext.Run(
+                context.AsyncExecutionContext,
+                static state =>
+                {
+                    var invocationState = (InvocationState)state!;
+                    Activity.Current?.AddException(invocationState.Exception);
+                    Activity.Current?.SetStatus(ActivityStatusCode.Error, invocationState.Exception.Message);
+                },
+                new InvocationState(context, exception));
```
jviau: nit: you could just use a value tuple here to avoid the new class.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
jviau: I would potentially even make this an extension method for `ScriptInvocationContext` and perform both the Activity status work _and_ the `ResultSource` work.


``` CSharp
public static void SetException(this ScriptInvocationContext context, Exception exception)
{
    // null check
    // set ResultSource.SetException / TrySetException
    // perform work that is in SetCurrentActivityStatus here
}
```

This way you can unit test this specific method as well.

jviau (approved):

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
jviau: PR #11159 actually has a good reason for extracting this to a common method. There are other locations we are setting the contexts exception. We should do a pass in this PR or another to ensure all places also track the exception on the activity.

RohitRanjanMS: Added an extension method that handles both TrySetException and updating the activity status. I’m unsure whether a null check on the exception parameter is necessary, since TrySetException will throw an ArgumentNullException if it’s null. There’s no null check in the existing code, so I’m assuming exception is never actually null.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
RohitRanjanMS: `AddException` adds an event based on the provided exception, but the key step for accurate reporting is setting the activity status.
I’m uncertain whether a null check for the exception is needed — see my other [comment.](https://github.com/Azure/azure-functions-host/pull/11313#discussion_r2349570473)

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
RohitRanjanMS: I added it to avoid boxing.

src/WebJobs.Script.Grpc/Extensions/ScriptInvocationContextExtensions.cs
```diff
@@ -269,5 +270,40 @@ internal static void SetRetryContext(ScriptInvocationContext context, Invocation
                 }
             }
         }
+
+        public static void RecordException(this ScriptInvocationContext context, Exception exception)
```
satvu: It's not immediately clear from the name that this will also try to set the resultSource (not just record the exception). I'm not sure that line 276 (`context.ResultSource?.TrySetException(exception);`) needs to be bundled in (it can just always come before calling this method, and it seems to be set up already with current code). 

If we do want to include it, I think `RecordAndSetException` would be clearer.

jviau: I would go with just `SetException`. Recording can be an implementation detail the caller doesn't exactly need to know about.

jviau (approved):

jviau (approved):

src/WebJobs.Script.Grpc/Extensions/ScriptInvocationContextExtensions.cs
RohitRanjanMS: I thought a lot about the name, it's never easy. I will update to `SetException`

satvu (approved):

PR #11325, opened 2025-09-15
Implement Route Handling Options and Configuration

### LINKED ISSUE #11304: Configuration - Add route handling to support implicit catch-all route
## Problem statement

Certain hosting scenarios—such as hosting an MCP server or other single-endpoint custom handlers behind the Functions host—benefit from a single catch-all route instead of per-function routing. Current approaches require manual configuration or fragile workarounds. A first-class configuration option will:

- Make the desired behavior explicit and discoverable in `host.json` beneath `customHandler`.
- Allow the runtime to validate and consistently apply the behavior.
- Provide a clear override and precedence model so existing deployments are not surprised by implicit changes.

## Description
Introduce a new http section under customHandler in host.json to provide a simple, validated way to control how HTTP routes are mapped by the Functions host.

### Sample entry
```Jsonc
"customHandler": {
  "http": {
    "DefaultAuthorizationLevel": "function",  // defaults to function
    "Routes": [
      {
        "route": "{*route}",
        "AuthorizationLevel": "function"   // optional, defaults to DefaultAuthorizationLevel
      }
    ]
  }
}
```

## Requirements
1. `host.json` must support a http object under the customHandler section.
2. Each entry in the Routes array defines an HTTP-triggered function implicitly, using the specified route
3. Authorization level (optional): If AuthorizationLevel is omitted, the DefaultAuthorizationLevel is applied.
4. DefaultAuthorizationLevel by default is function.
5. The runtime must synthesize function metadata from the configured routes.
6. Validation must ensure reserved routes (e.g., /admin) are not overridden.

## For MCP scenarios "mcp-custom-handler"

The following configuration will be added to the mcp-custom-handler host profile:
```Jsonc
"http": {
    "Routes": [
        {
        "route": "{*route}",
       }
  ]
}
```

The above config would map all routes to `AuthorizationLevel: funciton`. Customer can override the AuthorizationLevel in mcp-custom-handler host profile via DefaultAuthorizationLevel. Below is sample config that override AuthorizationLevel to anonymous level

## Sample host 
```json
{
	"version": "2.0",
	"customHandler": {
		"description": {
			"defaultExecutablePath": ""
		},
		"enableForwardingHttpRequest": true,
		"http": {
			"DefaultAuthorizationLevel" : "anonymous",
		}
	}
}
```

### Unit Tests

- Validate binding behavior when customHandler.http is present or absent.
- Reject invalid combinations (e.g., reserved routes).
- Ensure correct route prefix selection and defaulting behavior.
- Verify synthesized function metadata matches expected structure.

### Regression Tests

- Confirm proxies ignore routePrefix as before.
- Ensure legacy function routing remains unchanged when customHandler.http is absent.

Acceptance Criteria

- host.json supports customHandler.http with Routes and DefaultAuthorizationLevel.
- Routes are validated and synthesized into function metadata.
- Explicit configuration always takes precedence.
- Tests cover both all and function modes, validation, and regression scenarios.

anthonychu:

resolves #11304

### Pull request checklist

#### Summary of changes
- Adds a new `RouteHandlingOptions` configuration binding to read `routeHandling` object from the `customHandler` section in `host.json` with the properties `mode` (enum) and (optionally) `authorizationLevel` (enum).
- Wires the options into both the web host and script host scopes so the runtime can validate and apply the configured routing behavior using standard .NET Options validation.
- Implements routing behavior for:
  - `mode = Function` (default): preserve existing per-function routing behavior. When `mode = Function`, providing `authorizationLevel` is invalid and will fail host startup with a clear error message.
  - `mode = All`: expose an implicit, discoverable catch-all HTTP handler (named `http-handler`) with route `{*route}`; when specified, `authorizationLevel` controls that implicit handler's auth level.
- Synthesizes function metadata and descriptor creation for the implicit handler and integrates validation during host initialization.
- Uses enum types for both `RouteHandlingMode` (Function/All) and `AuthorizationLevel` (Anonymous, Function, User, Admin, System) for type safety.

#### Recent updates
- **Property naming consistency**: Renamed `AuthenticationLevel` property to `AuthorizationLevel` throughout the codebase for consistency with the underlying `AuthorizationLevel` enum type.
- Updated all related method names (`GetAuthorizationLevelString`), XML documentation, error messages, and test configurations to use "authorization" terminology consistently.
- Added using alias to avoid naming conflicts while maintaining clean, non-fully-qualified references.

#### Configuration structure
The configuration is now located under the `customHandler` section:
```json
{
  "version": "2.0",
  "customHandler": {
    "description": {
      "defaultExecutablePath": "handler"
    },
    "routeHandling": {
      "mode": "all",
      "authorizationLevel": "anonymous"
    }
  }
}

### CODE REVIEW

File: src/WebJobs.Script/Host/ScriptHost.cs [hunk 1]
```diff
@@ -1001,23 +1054,58 @@ internal static bool TryGetFunctionFromException(ICollection<FunctionDescriptor>
...
+                {
+                    // Look for the directory name as a path segment in the stack. This helps match cases
+                    // where the stack includes an inner file (e.g. .../HttpTriggerNode/npm/lib/foo.js).
+                    string seg = "/" + directoryName + "/";
+                    if (errorStack.Contains(seg) || errorStack.Contains("/" + directoryName + ":"))
+                    {
+                        function = currFunction;
+                        return true;
+                    }
+                }
```
anthonychu: Needed this to run tests (maybe a mac issue?). Can remove if needed.

src/WebJobs.Script.WebHost/WebScriptHostHttpRoutesManager.cs
```diff
@@ -34,7 +37,14 @@ public void InitializeHttpFunctionRoutes(IScriptJobHost host)
...
             // TODO: FACAVAL Instantiation of the ScriptRouteHandler should be cleaned up
-            WebJobsRouteBuilder routesBuilder = _router.CreateBuilder(new ScriptRouteHandler(_loggerFactory, host, _environment, false), _httpOptions.Value.RoutePrefix);
+            // If routeHandling.mode is "all", map routes at root (no routePrefix). Otherwise honor configured routePrefix.
+            string routePrefixToUse = _httpOptions.Value.RoutePrefix;
+            if (string.Equals(_routeHandlingOptions?.Mode, "all", StringComparison.OrdinalIgnoreCase))
+            {
+                routePrefixToUse = null;
+            }
+
+            WebJobsRouteBuilder routesBuilder = _router.CreateBuilder(new ScriptRouteHandler(_loggerFactory, host, _environment, false), routePrefixToUse);
```
anthonychu: Not sure if we want to do this or if we still want user to explicitly remove the prefix. Can't think of a reason why you'd want a catch-all route that has a prefix.

src/WebJobs.Script/Config/RouteHandlingOptionsSetup.cs
```diff
@@ -0,0 +1,37 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Azure.WebJobs.Script;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.WebJobs.Script.Configuration
+{
+    internal class RouteHandlingOptionsSetup : IConfigureOptions<RouteHandlingOptions>
```
jviau: nit: you can use a single line as an alternative to this class:

``` CSharp
services.AddOptions<RouteHandlingOptions>()
    .BindConfiguration(ConfigurationPath.Combine(ConfigurationSectionNames.JobHost, ConfigurationSectionNames.RouteHandling));
```

note: if switching `AuthenticationLevel` to the enum directly, I _think_ this will handle enum conversion. But you would need to verify.

It is also fine to leave this class if you prefer. You can still simplify the implementation of `Configure` to be:

``` csharp
_configuration.GetSection(ConfigurationPath.Combine(
        ConfigurationSectionNames.JobHost, ConfigurationSectionNames.RouteHandling))
    ?.Bind(options);
```

src/WebJobs.Script/Binding/WebJobsCoreScriptBindingProvider.cs
```diff
@@ -50,7 +64,20 @@ public override Type DefaultType
 
             public override Collection<Attribute> GetAttributes()
             {
-                var authLevel = Context.GetMetadataEnumValue<AuthorizationLevel>("authLevel", AuthorizationLevel.Function);
+                AuthorizationLevel defaultLevel = AuthorizationLevel.Function;
+                if (!string.IsNullOrEmpty(_parentRouteHandlingOptions?.AuthenticationLevel))
+                {
+                    try
+                    {
+                        defaultLevel = (AuthorizationLevel)Enum.Parse(typeof(AuthorizationLevel), _parentRouteHandlingOptions.AuthenticationLevel, ignoreCase: true);
```
jviau: I recommend performing this parsing directly in `RouteHandlingOptions` during binding.

src/WebJobs.Script/Config/RouteHandlingOptions.cs
```diff
@@ -0,0 +1,33 @@
...
+    public class RouteHandlingOptions
+    {
+        private string _authenticationLevel;
+
+        /// <summary>
+        /// Mode determining how routes are mapped. "function" (default) maps httpTrigger functions
+        /// defined in function metadata. "all" creates a single catch-all route that handles all
+        /// requests and proxies it to the custom handler.
+        /// </summary>
+        public string Mode { get; set; } = "function";
```
jviau: Consider making this an enum for clarity on what values this is allowed to take on.

src/WebJobs.Script/Host/FunctionMetadataManager.cs
```diff
@@ -153,6 +165,43 @@ internal ImmutableArray<FunctionMetadata> LoadFunctionMetadata(bool forceRefresh
...
+            // This ensures the FunctionMetadataManager is aware of the implicit handler so it can be resolved at runtime
+            // for logging/metrics and other metadata lookups.
+            if (string.Equals(_routeHandlingOptions?.Mode, "all", StringComparison.OrdinalIgnoreCase))
+            {
+                var handler = new FunctionMetadata()
+                {
+                    Name = "http-handler"
+                };
+
+                var inputRaw = new global::Newtonsoft.Json.Linq.JObject
```
jviau: Does `JObject` not work here directly? Without the `global::`

src/WebJobs.Script/Host/FunctionMetadataManager.cs
```diff
@@ -153,6 +165,43 @@ internal ImmutableArray<FunctionMetadata> LoadFunctionMetadata(bool forceRefresh
...
+                };
+
+                handler.Bindings.Add(BindingMetadata.Create(inputRaw));
+                handler.Bindings.Add(BindingMetadata.Create(outputRaw));
+
+                // Ensure Errors is non-null so callers that enumerate it (ScriptHost.GetFunctionsMetadata)
+                // do not throw when routeHandling.mode == "all" and we return early.
+                Errors = _functionErrors.ToImmutableDictionary(kvp => kvp.Key, kvp => kvp.Value.ToImmutableArray());
+
+                return new[] { handler }.ToImmutableArray();
```
jviau: Should be able to use collection initializers:


```suggestion
                return [ handler ];
```

If not, at least avoid the extra array allocation:

``` csharp
return ImmutableArray.Create(handler);
```

src/WebJobs.Script/Host/ScriptHost.cs
```diff
@@ -279,6 +284,13 @@ public async Task InitializeAsync(CancellationToken cancellationToken = default)
             _stopwatch = ValueStopwatch.StartNew();
             using (_metricsLogger.LatencyEvent(MetricEventNames.HostStartupLatency))
             {
+                // Validate route handling configuration: authenticationLevel is only valid when mode = "all".
+                if (string.Equals(_routeHandlingOptions?.Mode, "function", StringComparison.OrdinalIgnoreCase) &&
+                    !string.IsNullOrEmpty(_routeHandlingOptions?.AuthenticationLevel))
+                {
+                    throw new HostInitializationException("Invalid configuration: 'routeHandling.authenticationLevel' cannot be set when 'routeHandling.mode' is 'function'.");
```
jviau: Recommend validating and throwing this as part of options configuration. https://learn.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-9.0#options-validation

src/WebJobs.Script/Host/FunctionMetadataManager.cs
```diff
@@ -153,6 +165,43 @@ internal ImmutableArray<FunctionMetadata> LoadFunctionMetadata(bool forceRefresh
             var functionMetadataList = new List<FunctionMetadata>();
             _functionErrors = new Dictionary<string, ICollection<string>>();
 
+            // If routeHandling.mode is "all", ignore any function metadata configured and expose a single implicit http-handler function.
+            // This ensures the FunctionMetadataManager is aware of the implicit handler so it can be resolved at runtime
+            // for logging/metrics and other metadata lookups.
+            if (string.Equals(_routeHandlingOptions?.Mode, "all", StringComparison.OrdinalIgnoreCase))
```
jviau: It looks like this same logic is defined twice. Consider extracting it to an instance method on `RouteHandlingOptions` itself.

src/WebJobs.Script/Config/RouteHandlingOptionsSetup.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        public RouteHandlingOptionsSetup(IConfiguration configuration)
+        {
+            _configuration = configuration;
+        }
+
+        public void Configure(RouteHandlingOptions options)
+        {
+            IConfigurationSection jobHostSection = _configuration.GetSection(ConfigurationSectionNames.JobHost);
+            var routeHandlingSection = jobHostSection.GetSection(ConfigurationSectionNames.RouteHandling);
```
jviau: The issue mentions the config should be under `customHandler` section. This is not under that section. Should it be?

src/WebJobs.Script/Config/RouteHandlingOptions.cs
jviau: I am curious what value we get with making this explicit. Technically we can already identify the difference between default / "function" based purely on the presence of this config value or not.

I am not saying to remove it, I just want to understand the customer benefit from needing to set 2 config values instead of one.

jviau (changes requested):
1. I think the config value should be _authorization_ Level and not authentication.
2. I believe the linked issue describes this config being under "customHandler" in host.json, not top level as implemented in this PR
3. Just want to confirm that the 2 independent config values are truly needed. Reasoning I could see for wanting them to be separate immediately is if we plan on adding more options in the future. If that is not planned, then mode of "all" vs "function" can be implied by the presence of the "routHandling:authorizationLevel" config value or not.

src/WebJobs.Script.WebHost/WebScriptHostHttpRoutesManager.cs
fabiocav: We don't want to couple those things. Profile based configuration will already make this much simpler.

src/WebJobs.Script/Config/RouteHandlingOptions.cs
fabiocav: +1

src/WebJobs.Script/Config/RouteHandlingOptionsSetup.cs
fabiocav: It should be under `customHandler`

src/WebJobs.Script/Host/ScriptHost.cs
fabiocav: +1

src/WebJobs.Script/Config/RouteHandlingOptions.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        /// <summary>
+        /// Only applicable to mode = "all". Determines the authentication level for the catch-all route.
+        /// Allowed values: "function", "anonymous".
```
fabiocav: > /// Allowed values: "function", "anonymous".

The comment limits to two possible allowed values, but default logic returns `string.Empty` when `function`

src/WebJobs.Script/Config/RouteHandlingOptions.cs
```diff
@@ -0,0 +1,33 @@
...
+        /// Defaults to "function" when Mode == "all". When Mode == "function" this defaults to an empty string.
+        /// </summary>
+
+        public string AuthenticationLevel
+        {
+            get => _authenticationLevel == null
+                ? (string.Equals(Mode, "all", StringComparison.OrdinalIgnoreCase) ? "function" : string.Empty)
+                : _authenticationLevel;
+            set => _authenticationLevel = value;
```
fabiocav: Validate within expected values?

src/WebJobs.Script/Config/RouteHandlingOptions.cs
fabiocav: Is there a reason why we wouldn't be able to support all our current AuthLevel options?

src/WebJobs.Script/Host/FunctionMetadataManager.cs
```diff
@@ -128,6 +136,10 @@ private void InitializeServices()
             _isHttpWorker = _serviceProvider.GetService<IOptions<HttpWorkerOptions>>()?.Value?.Description != null;
             _scriptOptions = _serviceProvider.GetService<IOptions<ScriptJobHostOptions>>();
 
+            // Refresh route handling options from the new host scope so routeHandling.mode
```
fabiocav: This comment here applies to the pattern used by all the options retrieved above. Either move the comment to the top and make it generic or remove.

src/WebJobs.Script/Host/FunctionMetadataManager.cs
```diff
@@ -128,6 +136,10 @@ private void InitializeServices()
...
+            // reflects the active host configuration (e.g. after specialization/host swap).
+            _routeHandlingOptions = _serviceProvider.GetService<IOptions<RouteHandlingOptions>>()?.Value;
```
fabiocav, line 138: Have you validated this with specialization?

src/WebJobs.Script/Host/FunctionMetadataManager.cs
fabiocav: We only want this logic to run when running custom handlers.

src/WebJobs.Script/Host/FunctionMetadataManager.cs
```diff
@@ -153,6 +165,43 @@ internal ImmutableArray<FunctionMetadata> LoadFunctionMetadata(bool forceRefresh
             var functionMetadataList = new List<FunctionMetadata>();
             _functionErrors = new Dictionary<string, ICollection<string>>();
 
+            // If routeHandling.mode is "all", ignore any function metadata configured and expose a single implicit http-handler function.
```
fabiocav: Let's refactor this to a separate method/remove from `LoadFunctionMetadata`

src/WebJobs.Script/Host/ScriptHost.cs
```diff
@@ -158,6 +161,8 @@ public ScriptHost(IOptions<JobHostOptions> options,
 
             _functionDataCache = functionDataCache;
             _hostingConfigOptions = hostingConfigOptions;
+
+            _routeHandlingOptions = routeHandlingOptions?.Value;
```
fabiocav, line 165: We'd not expect an `IOptions<T>` instance to be `null` here. Are you intruducing this as optional to avoid test impact?

fabiocav (commented):
Minor initial comments. Also had a sync with @soninaren to share context on the changes.

src/WebJobs.Script/Config/RouteHandlingOptions.cs
anthonychu: Converted to enum

src/WebJobs.Script/Host/FunctionMetadataManager.cs
anthonychu: Not sure how to test this, but this did fix an issue I had that might have resulted from a host restart

src/WebJobs.Script/Config/RouteHandlingOptions.cs
```diff
@@ -0,0 +1,128 @@
...
+        /// </summary>
+        Function,
+
+        /// <summary>
+        /// Creates a single catch-all route that handles all requests and proxies them to the custom handler.
+        /// </summary>
+        All
+    }
+
+    public class RouteHandlingOptions
```
soninaren, line 26: Since this is now in Custom handler section. We can simply add these properties [HttpWorkerOptions](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptions.cs) class. We don't need to create RouteHandlingOptions.

src/WebJobs.Script/Config/RouteHandlingOptions.cs
```diff
@@ -0,0 +1,128 @@
...
+        {
+            var authLevel = AuthorizationLevel;
+            return authLevel?.ToString().ToLowerInvariant();
+        }
+
+        /// <summary>
+        /// Creates HTTP handler metadata for the catch-all route when mode is "all".
+        /// </summary>
+        /// <returns>A FunctionMetadata object representing the HTTP handler.</returns>
+        public Description.FunctionMetadata CreateHttpHandlerMetadata()
```
soninaren, line 70: Can create an implementation of [IFunctionsProvider](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.Abstractions/Description/IFunctionProvider.cs) to generate and inject this metadata.

soninaren (commented):
An alternative approach here would be to create a config like this. This would handle the catch all scenario but allow us to create more than one customized route. 
```CSharp
{
  "customHandlers": {
    "routes": [ 
      {
        "route": "/my/route",
        "authorizationLevel": "function",
      },
      {
       "route": "{*route}", // Catch all
       "authorizationLevel": "function",
      },
    ]
  }
}
```

anthonychu:
Closing in favor of #11345

PR #11345, opened 2025-09-21
Adding route handling configuration for custom handlers

### LINKED ISSUE #11304: Configuration - Add route handling to support implicit catch-all route

soninaren:
This pull request introduces support for custom route handling in custom handler HTTP workers, allowing users to define HTTP routes and authorization levels directly in configuration. It adds new configuration options, updates service registrations to support the new provider, and implements logic to generate function metadata from route definitions. Several tests and documentation are updated to reflect these changes.

**Custom Handler HTTP Worker Route Support**

* Added a new `CustomHandlerHttpOptions` class and `HttpWorkerRoute` class to represent HTTP routes and authorization levels for custom handler workers (`src/WebJobs.Script/Workers/Http/CustomHandlerHttpOptions.cs`, `src/WebJobs.Script/Workers/Http/HttpWorkerRoute.cs`). [[1]](diffhunk://#diff-714275196413e568ad4f439aceefc5eac3e36f61e251b31a93371f8061528c49R1-R22) [[2]](diffhunk://#diff-3b0102da3f9907f0953242b1eaf25b56fdca20da178494cb36656a0bcb7b341aR1-R23)
* Updated `HttpWorkerOptions` to include `CustomRoutesEnabled` and an `Http` property for route configuration (`src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptions.cs`).
* Implemented `HttpWorkerFunctionProvider`, which generates function metadata from configured routes, validates route templates, and enforces configuration rules (`src/WebJobs.Script/Workers/Http/HttpWorkerFunctionProvider.cs`).

**Configuration and Service Registration**

* Registered `HttpWorkerFunctionProvider` as an `IFunctionProvider` and updated service registration to allow multiple providers (`src/WebJobs.Script/ScriptHostBuilderExtensions.cs`).
* Updated `HostConfigurationProfile` to add a default route mapping for custom handlers and adjusted related tests (`src/WebJobs.Script/Config/HostConfigurationProfile.cs`, `test/WebJobs.Script.Tests/Configuration/HostConfigurationProfileTests.cs`). [[1]](diffhunk://#diff-c6d0b6e2b0f3a8cfba20890b8d94591c588ddbfefc80977575799f0b823c99d7R33-R34) [[2]](diffhunk://#diff-fbaa5a8134225ed08cf37547e346544e10a21709176b7cf3f99e7192823e61d9L38-R46)
* Enhanced `HttpWorkerOptionsSetup` to set `CustomRoutesEnabled` for custom handler runtimes, populate route authorization levels, and support new constructor signature in tests (`src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptionsSetup.cs`, `test/WebJobs.Script.Tests/Configuration/HttpWorkerOptionsSetupTests.cs`). [[1]](diffhunk://#diff-7aba5109237a1688e4610e31720ab6134ac687e56c333ee312483e4c37d4f085R20-R35) [[2]](diffhunk://#diff-7aba5109237a1688e4610e31720ab6134ac687e56c333ee312483e4c37d4f085R75-R89) [[3]](diffhunk://#diff-467e5379eccec4a7b945df25e4059ff5d3c6f3ce0b4d936407cd6898ebe308b4L90-R90) [[4]](diffhunk://#diff-467e5379eccec4a7b945df25e4059ff5d3c6f3ce0b4d936407cd6898ebe308b4L127-R127) [[5]](diffhunk://#diff-467e5379eccec4a7b945df25e4059ff5d3c6f3ce0b4d936407cd6898ebe308b4L160-R160) [[6]](diffhunk://#diff-467e5379eccec4a7b945df25e4059ff5d3c6f3ce0b4d936407cd6898ebe308b4L215-R215) [[7]](diffhunk://#diff-467e5379eccec4a7b945df25e4059ff5d3c6f3ce0b4d936407cd6898ebe308b4L277-R277)

**Documentation**

* Updated release notes to announce route handling configuration for custom handlers (`release_notes.md`).

**Constants**

* Added a constant for the custom handler language worker name (`src/WebJobs.Script/Workers/Rpc/RpcWorkerConstants.cs`).

resolves #11304

### Pull request checklist

### Additional information

Additional PR information


src/WebJobs.Script/Workers/Http/HttpWorkerFunctionProvider.cs
```diff
@@ -0,0 +1,184 @@
...
+                _logger.LogInformation("Created function '{functionName}' for route '{routeTemplate}' (authLevel={auth}).", functionName, route.Route, route.AuthorizationLevel);
+            }
+
+            return [.. functions];
+        }
+
+        private static bool TryValidateHttpRoute(string route, out string error)
+        {
+            error = null;
+
```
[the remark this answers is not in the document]
soninaren: moved the null check from another file to this one to have all the error conditions in one place.

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptions.cs
```diff
@@ -33,6 +34,13 @@ public class HttpWorkerOptions : IOptionsFormatter
 
         public TimeSpan InitializationTimeout { get; set; } = TimeSpan.FromSeconds(30);
 
+        public string WorkerRuntime { get; set; }
```
fabiocav: Doesn't look like we have anything consuming this. I'm working on removing as many dependencies on this as possible in a separate set of changes (particularly where we're hydrating those dependencies directly from the environment variables).

Let's remove this to avoid expanding use of FWR.

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptionsSetup.cs
```diff
@@ -40,6 +44,8 @@ public void Configure(HttpWorkerOptions options)
                 _logger.LogWarning($"Both {ConfigurationSectionNames.HttpWorker} and {ConfigurationSectionNames.CustomHandler} sections are spefified in {ScriptConstants.HostMetadataFileName} file. {ConfigurationSectionNames.CustomHandler} takes precedence.");
             }
 
+            options.WorkerRuntime = _environment.GetFunctionsWorkerRuntime();
```
fabiocav: Expanding on the comment above about FWR. Doesn't look like this is really required, so we should be able to revert the changes made in this file.

src/WebJobs.Script/Workers/Http/HttpWorkerFunctionProvider.cs
```diff
@@ -0,0 +1,184 @@
...
+        }
+
+        private static FunctionMetadata CreateHttpFunctionMetadata(HttpWorkerRoute route, string functionName)
+        {
+            var trigger = new BindingMetadata
+            {
+                Raw = new JObject
+                {
+                    ["type"] = "httpTrigger",
+                    ["authLevel"] = route.AuthorizationLevel.ToString(),
```
fabiocav: `authLevel` -> `authorizationLevel`

src/WebJobs.Script/Workers/Http/HttpWorkerFunctionProvider.cs
```diff
@@ -0,0 +1,184 @@
...
+        private const string DoubleSlash = "//";
+
+        private readonly HttpWorkerOptions _httpWorkerOptions;
+        private readonly IHostFunctionMetadataProvider _hostFunctionMetadataProvider;
+        private readonly IOptionsMonitor<LanguageWorkerOptions> _languageWorkerOptions;
+        private readonly ILogger _logger;
+        private readonly Dictionary<string, List<string>> _errors = new();
+        private static readonly ImmutableArray<string> AllHttpMethods = ["get", "post", "put", "delete", "patch", "head", "options"];
+
+        public HttpWorkerFunctionProvider(IOptions<HttpWorkerOptions> httpWorkerOptions, IOptionsMonitor<LanguageWorkerOptions> languageWorkerOptions, IHostFunctionMetadataProvider hostFunctionMetadataProvider, IEnvironment environment, ILogger<HttpWorkerFunctionProvider> logger)
```
fabiocav: `environment` is unused.

src/WebJobs.Script/Workers/Http/HttpWorkerFunctionProvider.cs
```diff
@@ -0,0 +1,184 @@
...
+    internal class HttpWorkerFunctionProvider : IFunctionProvider
+    {
+        private const char SpaceChar = ' ';
+        private const string DoubleSlash = "//";
+
+        private readonly HttpWorkerOptions _httpWorkerOptions;
+        private readonly IHostFunctionMetadataProvider _hostFunctionMetadataProvider;
+        private readonly IOptionsMonitor<LanguageWorkerOptions> _languageWorkerOptions;
+        private readonly ILogger _logger;
+        private readonly Dictionary<string, List<string>> _errors = new();
```
fabiocav: nit: use collection expression here

src/WebJobs.Script/Workers/Http/HttpWorkerFunctionProvider.cs
```diff
@@ -0,0 +1,184 @@
...
+        private static readonly ImmutableArray<string> AllHttpMethods = ["get", "post", "put", "delete", "patch", "head", "options"];
```
fabiocav: This is not an exhaustive list (e.g., missing TRACE, CONNECT, etc.), Consider using `Microsoft.AspNetCore.Http.HttpMethods` and performing validation against the methods defined there.

src/WebJobs.Script/Workers/Http/HttpWorkerFunctionProvider.cs
```diff
@@ -0,0 +1,184 @@
...
+                functions.Add(CreateHttpFunctionMetadata(route, functionName));
+                _logger.LogInformation("Created function '{functionName}' for route '{routeTemplate}' (authLevel={auth}).", functionName, route.Route, route.AuthorizationLevel);
+            }
+
+            return [.. functions];
+        }
+
+        private static bool TryValidateHttpRoute(string route, out string error)
+        {
+            error = route switch
```
fabiocav: We should be able to rely on the `TemplateParser` class for validation here. Let's not perform our own validation.

src/WebJobs.Script/Workers/Http/HttpWorkerFunctionProvider.cs
```diff
@@ -0,0 +1,184 @@
...
+        {
+            var trigger = new BindingMetadata
+            {
+                Raw = new JObject
+                {
+                    ["type"] = "httpTrigger",
+                    ["authLevel"] = route.AuthorizationLevel.ToString(),
+                    ["direction"] = "in",
+                    ["name"] = "req",
+                    ["methods"] = new JArray(AllHttpMethods),
```
fabiocav: Reuse this array to avoid allocations.

src/WebJobs.Script/ScriptConstants.cs
```diff
@@ -262,6 +262,7 @@ public static class ScriptConstants
         public static readonly string HttpProxyCorrelationHeader = "x-ms-invocation-id";
         public static readonly string HttpProxyTask = "HttpProxyTask";
         public static readonly string HttpProxyScriptInvocationContext = "HttpProxyScriptInvocationContext";
+        public static readonly string CustomHandlerWorkerRuntime = "custom";
```
fabiocav: Other runtimes are defined in `RpcWorkerConstants`. We should have this there as well.

src/WebJobs.Script/Workers/Http/HttpWorkerFunctionProvider.cs
```diff
@@ -0,0 +1,184 @@
...
+        public HttpWorkerFunctionProvider(IOptions<HttpWorkerOptions> httpWorkerOptions, IOptionsMonitor<LanguageWorkerOptions> languageWorkerOptions, IHostFunctionMetadataProvider hostFunctionMetadataProvider, IEnvironment environment, ILogger<HttpWorkerFunctionProvider> logger)
+        {
+            _httpWorkerOptions = httpWorkerOptions?.Value ?? throw new ArgumentNullException(nameof(httpWorkerOptions));
+            _hostFunctionMetadataProvider = hostFunctionMetadataProvider ?? throw new ArgumentNullException(nameof(hostFunctionMetadataProvider));
+            _languageWorkerOptions = languageWorkerOptions ?? throw new ArgumentNullException(nameof(languageWorkerOptions));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        }
+
+        public ImmutableDictionary<string, ImmutableArray<string>> FunctionErrors =>
+           _errors.ToImmutableDictionary(k => k.Key, v => v.Value.ToImmutableArray());
```
jviau: This will create an immutable dictionary each time this property is retrieved. Recommend storing it as an immutable dictionary to begin with.

src/WebJobs.Script/Workers/Http/HttpWorkerFunctionProvider.cs
```diff
@@ -0,0 +1,184 @@
...
+
+        public async Task<ImmutableArray<FunctionMetadata>> GetFunctionMetadataAsync()
+        {
+            if (!string.Equals(_httpWorkerOptions.WorkerRuntime, ScriptConstants.CustomHandlerWorkerRuntime, StringComparison.OrdinalIgnoreCase))
```
jviau: this kind of check is a good candidate to be moved into some kind of options object, so we can have an `IsEnabled`.

Can we add something like `ConfigRoutesEnabled` to `HttpWorkerOptions`?

src/WebJobs.Script/Workers/Http/HttpWorkerFunctionProvider.cs
```diff
@@ -0,0 +1,184 @@
...
+                return [];
+            }
+
+            var routes = _httpWorkerOptions.HttpRoutes;
+            if (routes is null || !routes.Any())
+            {
+                return [];
+            }
+
+            var hostFunctionMetadata = await _hostFunctionMetadataProvider.GetFunctionMetadataAsync(_languageWorkerOptions.CurrentValue.WorkerConfigs, forceRefresh: false);
```
jviau, line 70: Does this need a force refresh? This will only ever be called after the metadata manager has already loaded host metadata. I think we would want to avoid reloading it here?

src/WebJobs.Script/Workers/Http/HttpWorkerFunctionProvider.cs
```diff
@@ -0,0 +1,184 @@
...
+                    "Detected both function.json files and custom handler HTTP route configuration definition in host.json" +
+                    "Only one configuration source is supported. Remove either the function.json files or the HTTP routes entries in host.json.");
+            }
+
+            return CreateFunctionsFromRoutes(routes);
+        }
+
+        private ImmutableArray<FunctionMetadata> CreateFunctionsFromRoutes(IEnumerable<HttpWorkerRoute> routes)
+        {
+            var functions = new Collection<FunctionMetadata>();
```
jviau: Use an `ImmutableArray.CreateBuilder<T>()`

src/WebJobs.Script/Workers/Http/HttpWorkerFunctionProvider.cs
```diff
@@ -0,0 +1,184 @@
...
+            int i = 0;
+            foreach (var route in routes)
+            {
+                var functionName = $"http-handler{i + 1}";
+                i++;
+
+                if (!TryValidateHttpRoute(route?.Route, out string error))
+                {
+                    AddFunctionError(functionName, error);
```
jviau: I think `_errors` needs to be cleared between invocations of `CreateFunctionsFromRoutes`. I recommend creating an `ImmutableDictionary` builder as well in this method, and then building & assigning it to the `FunctionErrors` property at the end. This will address my other comment of the property getter recreating the immutable on each get, and also clear out the errors on each run.

jviau (changes requested):

src/WebJobs.Script/Workers/Http/HttpWorkerFunctionProvider.cs
soninaren: Right, That is why force refresh is set to false.

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptionsSetup.cs
soninaren: @fabiocav  This is being used to check if the WorkerRuntime is set to custom in HttpWorkerFunctionProvider. Are you suggesting we skip that check?

src/WebJobs.Script/Workers/Http/HttpWorkerFunctionProvider.cs
jviau: .... 🤦

src/WebJobs.Script/Workers/Http/HttpWorkerFunctionProvider.cs
```diff
@@ -0,0 +1,184 @@
...
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Script.Description;
+using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs.Script.Workers.Http
+{
+    internal class HttpWorkerFunctionProvider : IFunctionProvider
```
kshyju:
```suggestion
    internal sealed class HttpWorkerFunctionProvider : IFunctionProvider
```

jviau: Is there a big reason for wanting this sealed? I tend to only use sealed in two cases:

1. Blocking inheritance for a public type (in a public library)
2. Simplifying the `IDisposable` pattern

Is there another reason for using `sealed`?

fabiocav: Performance and design intent.

Sealing types will have significant impact on how the runtime deals with the type.

From a design standpoint, unless a type is carefully designed for inheritance, we should default them to sealed.

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptionsSetup.cs
fabiocav: The check is still important, but we now have this sprinkled in several places, making some of the changes we need to make a bit harder. Are we able to leverage one of the existing places that already expose the worker runtime? If not, I'm wondering if an explicit check where we need to check this condition would be preferred over exposing this information again in another property.

src/WebJobs.Script/Workers/Http/HttpWorkerFunctionProvider.cs
soninaren: Not sure if you meant for me to use ["authorizationLevel"] in JObject. Doing that seems to not work.

src/WebJobs.Script/Workers/Http/HttpWorkerFunctionProvider.cs
```diff
@@ -0,0 +1,146 @@
...
+        public async Task<ImmutableArray<FunctionMetadata>> GetFunctionMetadataAsync()
+        {
+            var routes = _httpWorkerOptions.Http?.Routes;
+
+            if (routes is null || !routes.Any())
+            {
+                return [];
+            }
+
+            if (!_httpWorkerOptions.CustomRoutesEnabled)
```
jviau, line 65: Maybe this could be done in options validation? That would avoid needing to know that `CustomRoutesEnabled` is determined by functions worker runtime.

src/WebJobs.Script/Workers/Http/HttpWorkerFunctionProvider.cs
```diff
@@ -0,0 +1,146 @@
...
+
+        private ImmutableArray<FunctionMetadata> CreateFunctionsFromRoutes(IEnumerable<HttpWorkerRoute> routes)
+        {
+            var metadatabuilder = ImmutableArray.CreateBuilder<FunctionMetadata>();
+            var errorsBuilder = ImmutableDictionary.CreateBuilder<string, ImmutableArray<string>>();
+            int i = 0;
+
+            foreach (var route in routes)
+            {
+                var functionName = $"http-handler{i + 1}";
```
jviau: Don't want to start at 0?

Starting at 0:


```suggestion
                var functionName = $"http-handler{i++}";
```

Starting at 1:
```suggestion
                var functionName = $"http-handler{++i}";
```

Can remove line 88 with both suggestions (personally, I would start at 0)

test/WebJobs.Script.Tests/Workers/Http/HttpWorkerFunctionProviderTests.cs
```diff
@@ -0,0 +1,301 @@
...
+        public async Task GetFunctionMetadataAsync_NoHttpSection_ReturnsEmpty()
+        {
+            var hostMeta = new Mock<IHostFunctionMetadataProvider>(MockBehavior.Strict);
+            var provider = CreateProvider(
+                CreateHttpWorkerOptions(routes: null, customRoutesEnabled: true, includeHttpSection: false),
+                hostMeta.Object);
+
+            var result = await provider.GetFunctionMetadataAsync();
+
+            Assert.Empty(result);
```
jviau, line 71: If you are interested, I much prefer `AwesomeAssertions` (aka FluentAssertions) over Xunit.Assert.


```suggestion
            result.Should().BeEmpty();
```

jviau (approved):

src/WebJobs.Script/Workers/Http/HttpWorkerFunctionProvider.cs
soninaren: I am so conflicted. I would really want to start at 0 but a function name with 0 does not feel right either.

src/WebJobs.Script/Workers/Http/HttpWorkerFunctionProvider.cs
soninaren: I considered doing that, but it felt like the options was trying to validate things outside its scope. Worker runtime is not really host configuration. Though i am open to moving it to options.

src/WebJobs.Script/Workers/Http/HttpWorkerFunctionProvider.cs
```diff
@@ -0,0 +1,145 @@
...
+            int i = 0;
+
+            foreach (var route in routes)
+            {
+                var functionName = $"http-handler{++i}";
+
+                try
+                {
+                    // Template parser does not check for empty route.
+                    if (string.IsNullOrEmpty(route.Route))
```
RohitRanjanMS: Any possibility of `route` being null? You have used null-conditional-operator in line 97 that makes me think that it can be null.

RohitRanjanMS: Do you think we should also check for IsNullOrWhiteSpace()?

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptionsSetup.cs
```diff
@@ -14,19 +17,21 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Http
...
         private string workerArgumentsSectionName = $"{WorkerConstants.WorkerDescription}:workerArguments";
 
-        public HttpWorkerOptionsSetup(IOptions<ScriptJobHostOptions> scriptJobHostOptions, IConfiguration configuration, ILoggerFactory loggerFactory, IMetricsLogger metricsLogger)
+        public HttpWorkerOptionsSetup(IOptions<ScriptJobHostOptions> scriptJobHostOptions, IConfiguration configuration, ILoggerFactory loggerFactory, IMetricsLogger metricsLogger, IEnvironment environment)
         {
             _scriptJobHostOptions = scriptJobHostOptions.Value;
             _configuration = configuration;
             _metricsLogger = metricsLogger;
             _logger = loggerFactory.CreateLogger<HttpWorkerOptionsSetup>();
+            _environment = environment ?? throw new ArgumentNullException(nameof(environment));
```
RohitRanjanMS: nit: ArgumentNullException.ThrowIfNull(environment)

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptionsSetup.cs
```diff
@@ -66,6 +71,21 @@ public void Configure(HttpWorkerOptions options)
...
         {
             workerSection.Bind(options);
+
+            var workerRuntime = _environment.GetFunctionsWorkerRuntime();
+            if (string.Equals(workerRuntime, RpcWorkerConstants.CustomHandlerLanguageWorkerName, StringComparison.OrdinalIgnoreCase))
+            {
+                options.CustomRoutesEnabled = true;
+            }
+
+            if (options.Http?.Routes is not null && options.Http.Routes.Any())
```
RohitRanjanMS, line 82: I think `is not null` is redundant.

src/WebJobs.Script/Workers/Http/CustomHandlerHttpOptions.cs
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Collections;
+using System.Collections.Generic;
+using Microsoft.Azure.WebJobs.Extensions.Http;
+
+namespace Microsoft.Azure.WebJobs.Script.Workers.Http
+{
+    public class CustomHandlerHttpOptions
```
RohitRanjanMS: sealed

src/WebJobs.Script/Workers/Http/HttpWorkerFunctionProvider.cs
```diff
@@ -0,0 +1,145 @@
...
+        private readonly HttpWorkerOptions _httpWorkerOptions;
+        private readonly IHostFunctionMetadataProvider _hostFunctionMetadataProvider;
+        private readonly IOptionsMonitor<LanguageWorkerOptions> _languageWorkerOptions;
+        private readonly ILogger _logger;
+        private ImmutableDictionary<string, ImmutableArray<string>> _errors = ImmutableDictionary<string, ImmutableArray<string>>.Empty;
+        private static readonly JArray AllHttpMethods = BuildAllHttpMethods();
+
+        public HttpWorkerFunctionProvider(IOptions<HttpWorkerOptions> httpWorkerOptions, IOptionsMonitor<LanguageWorkerOptions> languageWorkerOptions, IHostFunctionMetadataProvider hostFunctionMetadataProvider, ILogger<HttpWorkerFunctionProvider> logger)
+        {
+            _httpWorkerOptions = httpWorkerOptions?.Value ?? throw new ArgumentNullException(nameof(httpWorkerOptions));
```
RohitRanjanMS: nit: use ArgumentNullException.ThrowIfNull(<>);

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptionsSetup.cs
soninaren: Need the `is not null` the check if routes is not empty.

src/WebJobs.Script/Workers/Http/HttpWorkerFunctionProvider.cs
soninaren: it is unlikely that route is going to be null. I tried several combinations on config. i can potentially remove null conditional operator from line 97

soninaren: not sure, TemplateParser checks for null. I wanted to do minimal validation on our part and pass that responsibility to template parser

soninaren: Confirmed route will never be null. I have removed null-conditional-operator in line 97

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptions.cs
```diff
@@ -33,6 +33,16 @@ public class HttpWorkerOptions : IOptionsFormatter
 
         public TimeSpan InitializationTimeout { get; set; } = TimeSpan.FromSeconds(30);
 
+        /// <summary>
+        /// Gets or sets a value indicating whether custom routes are enabled.
+        /// </summary>
+        public bool CustomRoutesEnabled { get; set; }
+
+        /// <summary>
+        /// Gets or sets route mapping for a HTTP worker.
```
fabiocav: nit: This description reflects the state today, but the `CustomHandlerHttpOptions` wouldn't be limited to route mapping (today, it already handles authorization level configuration).

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptionsSetup.cs
```diff
@@ -14,19 +17,22 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Http
...
         private string workerArgumentsSectionName = $"{WorkerConstants.WorkerDescription}:workerArguments";
 
-        public HttpWorkerOptionsSetup(IOptions<ScriptJobHostOptions> scriptJobHostOptions, IConfiguration configuration, ILoggerFactory loggerFactory, IMetricsLogger metricsLogger)
+        public HttpWorkerOptionsSetup(IOptions<ScriptJobHostOptions> scriptJobHostOptions, IConfiguration configuration, ILoggerFactory loggerFactory, IMetricsLogger metricsLogger, IEnvironment environment)
         {
             _scriptJobHostOptions = scriptJobHostOptions.Value;
             _configuration = configuration;
             _metricsLogger = metricsLogger;
             _logger = loggerFactory.CreateLogger<HttpWorkerOptionsSetup>();
+            ArgumentNullException.ThrowIfNull(environment);
```
fabiocav: Can you please move the argument validation to the top of this constructor?

src/WebJobs.Script/Workers/Http/HttpWorkerFunctionProvider.cs
```diff
@@ -0,0 +1,149 @@
...
+            foreach (var route in routes)
+            {
+                var functionName = $"http-handler{++i}";
+
+                try
+                {
+                    // Template parser does not check for empty route.
+                    if (string.IsNullOrEmpty(route.Route))
+                    {
+                        throw new ArgumentException("Route cannot be null or empty.");
```
fabiocav: This is not a valid exception here. The individual route is not an argument of this method.

src/WebJobs.Script/Workers/Http/HttpWorkerFunctionProvider.cs
soninaren: done

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptionsSetup.cs
```diff
@@ -66,6 +72,21 @@ public void Configure(HttpWorkerOptions options)
         private void ConfigureWorkerDescription(HttpWorkerOptions options, IConfigurationSection workerSection)
         {
             workerSection.Bind(options);
+
+            var workerRuntime = _environment.GetFunctionsWorkerRuntime();
+            if (string.Equals(workerRuntime, RpcWorkerConstants.CustomHandlerLanguageWorkerName, StringComparison.OrdinalIgnoreCase))
```
kshyju, line 77: Should we also check `options.Http?.Routes` is a valid collection before setting the `options.CustomRoutesEnabled = true` ?

kshyju: or is it optional?

src/WebJobs.Script/Workers/Http/HttpWorkerRoute.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        /// <summary>
+        /// Gets or sets route template.
+        /// </summary>
+        public string Route { get; set; }
+
+        /// <summary>
+        /// Gets or sets the authorization level.
+        /// </summary>
+        public AuthorizationLevel? AuthorizationLevel { get; set; }
```
mathewc, line 21: Thinking we want this property and it's host.json name to match our standard http trigger "authLevel" naming? https://github.com/Azure/azure-webjobs-sdk-extensions/blob/b8d5f24bb827ddb4e43410cd713cae3cbafc8bfb/src/WebJobs.Extensions.Http/HttpTriggerAttribute.cs#L69

Why have it differ?

src/WebJobs.Script/Config/HostConfigurationProfile.cs
```diff
@@ -30,6 +30,8 @@ public sealed class HostConfigurationProfile
                     ConfigurationSectionNames.CustomHandler, "enableHttpProxyingRequest"), "true"),
                 KeyValuePair.Create(ConfigurationPath.Combine(
                     ConfigurationSectionNames.Http, "routePrefix"), string.Empty),
+                KeyValuePair.Create(ConfigurationPath.Combine(
+                    ConfigurationSectionNames.CustomHandler, "http", "Routes", "0", "route"), "{*route}"),
```
jviau: Casing doesn't really matter, but I think lowercase would be nice for consistency.

```suggestion
                    ConfigurationSectionNames.CustomHandler, "http", "routes", "0", "route"), "{*route}"),
```

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptionsSetup.cs
jviau: _technically_ with options pattern the routes could be added by a later configuration. I don't think we get much benefit from using it as a condition for enabling custom routes. The collection being null or empty will do that naturally.

src/WebJobs.Script/Workers/Http/HttpWorkerFunctionProvider.cs
```diff
@@ -0,0 +1,149 @@
...
+                    metadatabuilder.Add(CreateHttpFunctionMetadata(route, functionName));
+                    _logger.LogInformation("Created function '{functionName}' for route '{routeTemplate}' (authLevel={auth}).",
+                        functionName, route.Route, route.AuthorizationLevel);
+                }
+                catch (ArgumentException ex)
+                {
+                    errorsBuilder.Add(functionName, [ex.Message]);
+                    _logger.LogError("Unable to create function '{functionName}' for route '{route}' due to invalid route: {reason}",
+                        functionName, route?.Route ?? "<null>", ex.Message);
+                    continue;
```
kshyju, line 104: nit: I do not think `continue` is needed here.

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptionsSetup.cs
jviau: You could do (I think)

```suggestion
            if (options.Http?.Routes is not null and not [])
```

or ...

```suggestion
            if (options.Http?.Routes?.Any() == true)
```

jviau (approved):

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptionsSetup.cs
soninaren: I had this `options.Http?.Routes?.Any() == true` before. For some reason it did not work. Have left it as is for now. Will update it another time.

src/WebJobs.Script/Workers/Http/CustomHandlerHttpOptions.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        /// <summary>
+        /// Gets or sets the default authorization level for custom handler HTTP worker routes.
+        /// </summary>
+        public AuthorizationLevel DefaultAuthorizationLevel { get; set; } = AuthorizationLevel.Function;
+
+        /// <summary>
+        /// Gets or sets route mapping for a HTTP worker.
+        /// </summary>
+        public IEnumerable<HttpWorkerRoute> Routes { get; set; }
```
brettsam, line 20: does anything prevent /admin from being overtaken? maybe that validation occurs elsewhere already. But it'd be good to maybe have a test to ensure it works.

fabiocav (approved):

src/WebJobs.Script/Workers/Http/CustomHandlerHttpOptions.cs
soninaren: I validated that admin endpoint is not overtaken by this config. Even with route prefix. Will add a test for this as well.

fabiocav (approved):

RohitRanjanMS (approved):

RohitRanjanMS (approved):

PR #11347, opened 2025-09-22
RpcException Handling 

RohitRanjanMS:

resolves #issue_for_this_pr

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/Diagnostics/SystemLogger.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/SystemLogger.cs
@@ -3,15 +3,12 @@
 
 using System;
 using System.Collections.Generic;
-using System.Text;
 using Microsoft.Azure.WebJobs.Host;
 using Microsoft.Azure.WebJobs.Host.Executors.Internal;
 using Microsoft.Azure.WebJobs.Host.Indexers;
 using Microsoft.Azure.WebJobs.Logging;
 using Microsoft.Azure.WebJobs.Script.Configuration;
 using Microsoft.Azure.WebJobs.Script.Eventing;
-using Microsoft.Azure.WebJobs.Script.WebHost.Security;
-using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 
@@ -19,7 +16,6 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 {
     public class SystemLogger : ILogger
     {
-        private const string RedactedMessage = "[Redacted]- Customers using AppInsights or OTel can view full details.";
         private readonly string _categoryName;
         private readonly string _functionName;
         private readonly string _hostInstanceId;
@@ -177,32 +173,10 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
                     functionName = string.IsNullOrEmpty(fex.MethodName) ? string.Empty : fex.MethodName.Replace("Host.Functions.", string.Empty);
                 }
 
-                formattedMessage = Sanitizer.Sanitize(formattedMessage);
-                var baseEx = exception.GetBaseException();
-                innerExceptionType = baseEx.GetType().ToString();
-
-                var originalMessage = baseEx.Message;
-                var formattedDetails = exception.ToFormattedString();
-
-                if (exception is FunctionInvocationException && baseEx is RpcException { RemoteMessage: { } remoteMsg })
-                {
-                    var replacement = GetRedactedExceptionMessage(remoteMsg);
-                    innerExceptionMessage = Sanitizer.Sanitize(originalMessage.Replace(remoteMsg, replacement));
-                    details = Sanitizer.Sanitize(formattedDetails.Replace(remoteMsg, replacement));
-                }
-                else
-                {
-                    innerExceptionMessage = Sanitizer.Sanitize(originalMessage);
-                    details = Sanitizer.Sanitize(formattedDetails);
-                }
+                (innerExceptionType, innerExceptionMessage, details, formattedMessage) = exception.GetSanitizedExceptionDetails(formattedMessage);
             }
 
             _eventGenerator.LogFunctionTraceEvent(logLevel, subscriptionId, appName, functionName, eventName, source, details, formattedMessage, innerExceptionType, innerExceptionMessage, invocationId, _hostInstanceId, activityId, runtimeSiteName, slotName, DateTime.UtcNow);
-        }
-
-        private static string GetRedactedExceptionMessage(string msg)
-        {
-            return $"{RedactedMessage} (Hash: {EncryptionHelper.GetSHA256Base64String(Encoding.UTF8.GetBytes(msg))})";
         }
     }
 }
--- a/src/WebJobs.Script.WebHost/Diagnostics/SystemLogger.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/SystemLogger.cs
@@ -19,6 +19,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
 {
     public class SystemLogger : ILogger
     {
+        private const string RedactedMessage = "[Redacted]- Customers using AppInsights or OTel can view full details.";
         private readonly string _categoryName;
         private readonly string _functionName;
         private readonly string _hostInstanceId;
@@ -176,20 +177,16 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
                     functionName = string.IsNullOrEmpty(fex.MethodName) ? string.Empty : fex.MethodName.Replace("Host.Functions.", string.Empty);
                 }
 
+                formattedMessage = Sanitizer.Sanitize(formattedMessage);
                 var baseEx = exception.GetBaseException();
-                innerExceptionType = baseEx.GetType().FullName;
+                innerExceptionType = baseEx.GetType().ToString();
 
                 var originalMessage = baseEx.Message;
                 var formattedDetails = exception.ToFormattedString();
 
-                const string RedactedMessage = "An exception occurred during invocation, but its details are redacted. Customers with AppInsights or OTel enabled can access full exception details.";
-
-                static string BuildReplacement(string msg)
-                    => $"{RedactedMessage} (Hash: {EncryptionHelper.GetSHA256Base64String(Encoding.UTF8.GetBytes(msg))})";
-
                 if (exception is FunctionInvocationException && baseEx is RpcException { RemoteMessage: { } remoteMsg })
                 {
-                    var replacement = BuildReplacement(remoteMsg);
+                    var replacement = GetRedactedExceptionMessage(remoteMsg);
                     innerExceptionMessage = Sanitizer.Sanitize(originalMessage.Replace(remoteMsg, replacement));
                     details = Sanitizer.Sanitize(formattedDetails.Replace(remoteMsg, replacement));
                 }
@@ -201,6 +198,11 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
             }
 
             _eventGenerator.LogFunctionTraceEvent(logLevel, subscriptionId, appName, functionName, eventName, source, details, formattedMessage, innerExceptionType, innerExceptionMessage, invocationId, _hostInstanceId, activityId, runtimeSiteName, slotName, DateTime.UtcNow);
+        }
+
+        private static string GetRedactedExceptionMessage(string msg)
+        {
+            return $"{RedactedMessage} (Hash: {EncryptionHelper.GetSHA256Base64String(Encoding.UTF8.GetBytes(msg))})";
         }
     }
 }
--- a/src/WebJobs.Script.WebHost/Diagnostics/SystemLogger.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/SystemLogger.cs
@@ -1,14 +1,17 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
 using System.Collections.Generic;
+using System.Text;
 using Microsoft.Azure.WebJobs.Host;
 using Microsoft.Azure.WebJobs.Host.Executors.Internal;
 using Microsoft.Azure.WebJobs.Host.Indexers;
 using Microsoft.Azure.WebJobs.Logging;
 using Microsoft.Azure.WebJobs.Script.Configuration;
 using Microsoft.Azure.WebJobs.Script.Eventing;
+using Microsoft.Azure.WebJobs.Script.WebHost.Security;
+using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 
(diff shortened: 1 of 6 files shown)


src/WebJobs.Script.WebHost/Security/EncryptionHelper.cs
```diff
@@ -94,7 +94,7 @@ public static string Decrypt(string value, IEnvironment environment = null)
             return Decrypt(key, value);
         }
 
-        private static string GetSHA256Base64String(byte[] key)
+        public static string GetSHA256Base64String(byte[] key)
```
surgupta-msft: internal access modifier should work here.

src/WebJobs.Script.WebHost/Diagnostics/SystemLogger.cs
```diff
@@ -173,9 +176,28 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
...
-                (innerExceptionType, innerExceptionMessage, details) = exception.GetExceptionDetails();
-                formattedMessage = Sanitizer.Sanitize(formattedMessage);
-                innerExceptionMessage = innerExceptionMessage ?? string.Empty;
+                var baseEx = exception.GetBaseException();
+                innerExceptionType = baseEx.GetType().FullName;
+
+                var originalMessage = baseEx.Message;
+                var formattedDetails = exception.ToFormattedString();
+
+                const string RedactedMessage = "An exception occurred during invocation, but its details are redacted. Customers with AppInsights or OTel enabled can access full exception details.";
```
surgupta-msft: We can either declare this string constant as global variable or keep in `ScriptConstants` file to keep string literals at one place.  

Also, is this message customer facing or just for our kusto logs?

test/WebJobs.Script.Tests/Eventing/SystemLoggerTests.cs
```diff
@@ -310,6 +310,50 @@ public void AppEnvironment_Reset_OnSpecialization()
             Assert.Equal("updatedruntimesitename", evt.RuntimeSiteName);
         }
 
+        [Fact]
+        public void Log_RpcException()
+        {
+            var ex = new RpcException("result", "stack", "type");
+            var details = "Microsoft.Azure.WebJobs.Script.Workers.Rpc.RpcException : Result: result\nType: \nException: An exception occurred during invocation, but its details are redacted. Customers with AppInsights or OTel enabled can access full exception details. (Hash: buCObrO8b0W8mfzTn8xHkoahvrDATTnyBMYXYjeAddY=)\nStack: type";
```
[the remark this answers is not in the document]
RohitRanjanMS: Good suggestion.

src/WebJobs.Script.WebHost/Diagnostics/SystemLogger.cs
```diff
@@ -173,9 +176,28 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
...
+
+                static string BuildReplacement(string msg)
```
surgupta-msft: Should we consider moving `BuildReplacement()` method outside instead of nested method, since the `Log()` method is already quite large?

Some naming ideas for `BuildReplacement()` method - 

`BuildReplacementExceptionMessage()`
`BuildRedactedExceptionMessage()`
`GetRedactedExceptionMessage()`

src/WebJobs.Script.WebHost/Diagnostics/SystemLogger.cs
```diff
@@ -173,9 +176,28 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
                     functionName = string.IsNullOrEmpty(fex.MethodName) ? string.Empty : fex.MethodName.Replace("Host.Functions.", string.Empty);
                 }
 
-                (innerExceptionType, innerExceptionMessage, details) = exception.GetExceptionDetails();
-                formattedMessage = Sanitizer.Sanitize(formattedMessage);
-                innerExceptionMessage = innerExceptionMessage ?? string.Empty;
+                var baseEx = exception.GetBaseException();
+                innerExceptionType = baseEx.GetType().FullName;
```
mathewc: Previous code ended up doing innerException.GetType().ToString(). Want to be sure we're not introducing any accidental changes here

src/WebJobs.Script.WebHost/Diagnostics/SystemLogger.cs
mathewc: Wondering if we can't just simplify this to "[Redacted]". Not sure we need all this log bloat.

src/WebJobs.Script.WebHost/Diagnostics/SystemLogger.cs
```diff
@@ -173,9 +176,28 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
...
+
+                var originalMessage = baseEx.Message;
```
mathewc: All these code changes are making it hard to see if you're regressing anything here. Can't you just keep the existing codepath as is and add if/else handling for these invocation exceptions?

E.g. currently the formatted message is being sanitized via Sanitizer.Sanitize(formattedMessage). In your refactoring you've lost that. Not sure if that is intentional.

RohitRanjanMS: This is intentional - formatted messages won't contain any secrets unless we explicitly choose to include them in the logs. I removed it for efficiency, but I can restore it if you believe there's any potential risk.

src/WebJobs.Script.WebHost/Diagnostics/SystemLogger.cs
RohitRanjanMS: I can, but I'm aiming to make it clear that we have the option to request these error messages from customers if needed for investigation.
I can change it to "Redacted - customers using AppInsights or OTel can view full details." to make it short.

src/WebJobs.Script.WebHost/Diagnostics/SystemLogger.cs
RohitRanjanMS: This is intentional. I believe it should be same, and I didn't see any difference during my testing, but I will revert it back just to avoid any concern.

src/WebJobs.Script.WebHost/Diagnostics/SystemLogger.cs
```diff
@@ -173,9 +176,28 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
                     functionName = string.IsNullOrEmpty(fex.MethodName) ? string.Empty : fex.MethodName.Replace("Host.Functions.", string.Empty);
                 }
 
-                (innerExceptionType, innerExceptionMessage, details) = exception.GetExceptionDetails();
```
jviau, line 176: Would it be possible to move the new logic into `GetExceptionDetails`?

src/WebJobs.Script.WebHost/Diagnostics/SystemLogger.cs
RohitRanjanMS: Message is only for kusto logs.

src/WebJobs.Script.WebHost/Diagnostics/SystemLogger.cs
RohitRanjanMS: GetExceptionDetails is shared, also used by AzureMonitorDiagnosticLogger.

surgupta-msft: We can create a new method and move the logic there.

src/WebJobs.Script.WebHost/Diagnostics/SystemLogger.cs
surgupta-msft: I am in favor of not making changes to the existing logic/messages wherever not required. Applens detectors and customers depend on log messages - even a small change can break their behavior. 

There is also difference in logic between `GetExceptionDetails()` and `GetBaseException()`

`GetExceptionDetails()` is using a while loop to find the innermost exception. Are we sure and have evidence that `GetBaseException()` has the exact same behavior in all scenarios? If not, we can create a new method and extract the required code (while loop to find the innermost exception) from `GetExceptionDetails()`

RohitRanjanMS: GetBaseException and GetExceptionDetails behave the same, with the only difference being that GetExceptionDetails applies sanitization. The logs and overall behavior are unchanged, apart from the redacted content. Have you noticed any changes in the log format or logic flow

`public virtual Exception GetBaseException()
{
    Exception? inner = InnerException;
    Exception back = this;

    while (inner != null)
    {
        back = inner;
        inner = inner.InnerException;
    }
    return back;
}`

src/WebJobs.Script.WebHost/Diagnostics/SystemLogger.cs
RohitRanjanMS: I don't think we need a helper method to do it, `GetBaseException()` returns the innermost exception.

jviau: Would be nice to at least move this logic into a separate method and add some comments.

RohitRanjanMS: Moved logic to a separate extension method.

src/WebJobs.Script.WebHost/Diagnostics/SystemLogger.cs
```diff
@@ -173,9 +173,7 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
...
-                formattedMessage = Sanitizer.Sanitize(formattedMessage);
-                innerExceptionMessage = innerExceptionMessage ?? string.Empty;
+                (innerExceptionType, innerExceptionMessage, details, formattedMessage) = exception.GetSanitizedExceptionDetails(formattedMessage);
```
surgupta-msft, line 176: I think we don't need to send `formattedMessage` to the method. We are mainly just returning it. We can keep the code related to `formattedMessage` unchanged.

src/WebJobs.Script.WebHost/Extensions/ExceptionExtensions.cs
```diff
@@ -27,5 +33,50 @@ public static (string ExceptionType, string ExceptionMessage, string ExceptionDe
...
+        /// A tuple containing:
+        /// (InnerExceptionType) The full CLR type name of the base exception.
+        /// (InnerExceptionMessage) The sanitized and safe base exception message.
+        /// (Details) The sanitized and safe formatted exception string.
+        /// (FormattedMessage) The sanitized version of the provided formattedMessage parameter.
+        /// </returns>
+        public static (string InnerExceptionType, string InnerExceptionMessage, string Details, string FormattedMessage)
+            GetSanitizedExceptionDetails(this Exception exception, string formattedMessage)
+        {
+            ArgumentNullException.ThrowIfNull(exception);
```
surgupta-msft, line 57: `exception` should not be null as we have a check before calling this method.

If we still prefer to add a check, then the previous code was not throwing on null. It was simply returning null. Not sure if we should change that behavior.

```csharp

GetExceptionDetails(this Exception exception)
{
    if (exception == null)
    {
        return (null, null, null);
    }
```

jviau (approved):

src/WebJobs.Script.WebHost/Extensions/ExceptionExtensions.cs
```diff
@@ -27,5 +33,50 @@ public static (string ExceptionType, string ExceptionMessage, string ExceptionDe
...
+
+            var baseException = exception.GetBaseException();
+            var innerType = baseException.GetType().ToString();
+            var originalMessage = baseException.Message;
+            var formattedDetails = exception.ToFormattedString();
+
+            return exception is FunctionInvocationException
+                && baseException is RpcException { RemoteMessage: var remoteMsg } && remoteMsg is not null
+                ? (innerType,
+                    Sanitizer.Sanitize(originalMessage.Replace(remoteMsg, GetRedactedExceptionMessage(remoteMsg), StringComparison.Ordinal)),
```
surgupta-msft: We are evaluating `GetRedactedExceptionMessage(remoteMsg)` twice.

src/WebJobs.Script.WebHost/Extensions/ExceptionExtensions.cs
```diff
@@ -27,5 +33,50 @@ public static (string ExceptionType, string ExceptionMessage, string ExceptionDe
...
+                    Sanitizer.Sanitize(formattedDetails.Replace(remoteMsg, GetRedactedExceptionMessage(remoteMsg), StringComparison.Ordinal)),
+                    formattedMessage)
+                : (innerType,
+                    Sanitizer.Sanitize(originalMessage),
+                    Sanitizer.Sanitize(formattedDetails),
+                    formattedMessage);
```
surgupta-msft: (may be nit) For me, it was more readable with if-else block but we can little bit simplify that 

```csharp
 if (<condition>)
 {
      var replacement = GetRedactedExceptionMessage(remoteMsg);
      originalMessage = originalMessage.Replace(remoteMsg, replacement, StringComparison.Ordinal);
      formattedDetails = formattedDetails.Replace(remoteMsg, replacement, StringComparison.Ordinal);
  }

  innerExceptionMessage = Sanitizer.Sanitize(originalMessage);
  details = Sanitizer.Sanitize(formattedDetails);

  return (innerType, innerExceptionMessage, details)
```

src/WebJobs.Script.WebHost/Extensions/ExceptionExtensions.cs
```diff
@@ -27,5 +33,50 @@ public static (string ExceptionType, string ExceptionMessage, string ExceptionDe
 
             return (exceptionType, exceptionMessage, exceptionDetails);
         }
+
+        /// <summary>
+        /// Returns sanitized exception details. The remote message segment is replaced with a redacted placeholder containing a stable hash so that
```
surgupta-msft: Consider updating the comment: 

`If RpcException, the remote message segment ...` 
or 
`For FunctionInvocationException with innermost exception as RpcException, the remote message segment ...`

src/WebJobs.Script.WebHost/Extensions/ExceptionExtensions.cs
surgupta-msft: Note: if you plan to return `(null, null, null)`, please add this check again as before

`innerExceptionMessage = innerExceptionMessage ?? string.Empty;`

src/WebJobs.Script.WebHost/Diagnostics/SystemLogger.cs
RohitRanjanMS: I thought about this before moving it to the ext method, the code is much cleaner now. The extension method takes care of all the sanitization and redaction.

src/WebJobs.Script.WebHost/Extensions/ExceptionExtensions.cs
RohitRanjanMS: This was intentional. As I moved the logic to an extension method so it can be reused, I added the null check.

Also, I didn’t like the old approach of returning `null, null, null` when the exception was `null`. Now it’s clearer that callers shouldn’t pass `null` and should handle it properly.

src/WebJobs.Script.WebHost/Extensions/ExceptionExtensions.cs
RohitRanjanMS: Good catch, I will fix this.

src/WebJobs.Script.WebHost/Extensions/ExceptionExtensions.cs
RohitRanjanMS: I will have anyway switch to `if else` to avoid calling GetRedactedExceptionMessage twice.

surgupta-msft (approved):

PR #11365, opened 2025-09-25
Add AppInsights module to publish metrics

### LINKED ISSUE #11364: [HealthChecks] Publish health check metrics to legacy AppInsights SDK
Currently customer's on the legacy (non-OTEL) app insights SDK will not get any health metrics as app insights does not listen to Meter API.

We should add a meter listener to bridge the gap from Meters to AppInsights SDK.

jviau:

resolves #11364

### Pull request checklist

* [ ] I have added all required tests (Unit tests, E2E tests) -- TODO

### Additional information

This PR adds an app insights telemetry module which exports Meter API instruments to app insights.

### Co-pilot summary

This pull request introduces a new system for exporting .NET Meter API metrics to Application Insights, allowing for improved observability of metrics in Azure Functions. The main changes include implementing a custom metric exporter, configuration options, and integration into the application's dependency injection setup. Additionally, supporting extensions and utility methods were added to ensure smooth operation and graceful shutdown.

**Metrics Exporter Integration:**

* Added `ApplicationInsightsMetricExporter`, a custom `ITelemetryModule` that listens to .NET Meter API instruments and exports their metrics to Application Insights. It handles various numeric types and supports asynchronous collection and graceful shutdown.
* Introduced `ApplicationInsightsMetricExporterOptions` for configuring which meters to listen to and the collection interval, with support for future extensibility.
* Integrated the new metric exporter into the DI container in `ScriptHostBuilderExtensions`, registering the exporter and configuring it to listen to the core Functions meter.

**Telemetry and Utility Enhancements:**

* Added `TelemetryClientExtensions` to efficiently map Meter instruments and their tags to Application Insights metrics, handling up to 10 dimensions and avoiding unnecessary allocations.
* Implemented `CancellationExtensions` to safely cancel `CancellationTokenSource` instances, handling disposed exceptions gracefully, which is used for robust exporter shutdown.

**Dependency and Usability Updates:**

* Updated using statements and minor method signature changes in `ScriptHostBuilderExtensions` to support the new metrics exporter integration. [[1]](diffhunk://#diff-0520ea7869aafe4adf4e46cee524515169549321187cc8ecf473409fe070b53cR32) [[2]](diffhunk://#diff-0520ea7869aafe4adf4e46cee524515169549321187cc8ecf473409fe070b53cL472-R474) [[3]](diffhunk://#diff-0520ea7869aafe4adf4e46cee524515169549321187cc8ecf473409fe070b53cL492-R495)

These changes collectively enable first-class support for exporting custom and built-in .NET metrics to Application Insights, improving monitoring and diagnostics for Azure Functions.


src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -479,40 +470,49 @@ internal static void ConfigureApplicationInsights(this ILoggingBuilder builder,
...
+                    });
 
                 builder.Services.ConfigureOptions<ApplicationInsightsLoggerOptionsSetup>();
                 builder.Services.AddSingleton<ISdkVersionProvider, FunctionsSdkVersionProvider>();
 
+                // Configure the meter listener, so we publish Meter API based metrics to application insights.
+                builder.Services.AddSingleton<ITelemetryModule, ApplicationInsightsMetricExporter>();
+                builder.Services.Configure<ApplicationInsightsMetricExporterOptions>(o =>
+                {
+                    o.Meters.Add(HostMetrics.FaasMeterName);
```
jviau, line 523: @RohitRanjanMS this will publish the health metrics as well as function invoke duration to app insights. Is that okay to publish func invoke duration? Or do we not want that to customer app insights?

src/WebJobs.Script/Diagnostics/TelemetryClientExtensions.cs
```diff
@@ -0,0 +1,133 @@
...
+                value9 = null;
+
+            int count = 0; // this will be how many 'valid' dimensions we have.
+            for (int i = 0; i < length; i++)
+            {
+                string? value = tags[i].Value?.ToString();
+                string name = tags[i].Key;
+
+                // Application Insights does not allow null/empty/whitespace dimension values.
+                if (!string.IsNullOrWhiteSpace(value))
```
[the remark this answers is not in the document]
jviau: Accessing an index for `ReadOnlySpan` is not expensive.

src/WebJobs.Script/Diagnostics/ApplicationInsightsMetricExporter.cs
```diff
@@ -0,0 +1,132 @@
...
+            _listener.Start();
+        }
+
+        /// <inheritdoc />
+        public async ValueTask DisposeAsync()
+        {
+            _listener.Dispose();
+
+            await _shutdown.CancelNoThrowAsync();
+            await _exportTask.ConfigureAwait(false);
```
RohitRanjanMS, line 90: Should we dispose _exportTask before disposing _listener?

src/WebJobs.Script/Diagnostics/ApplicationInsightsMetricExporter.cs
```diff
@@ -0,0 +1,132 @@
...
+            await _client.FlushAsync(default).ConfigureAwait(false);
```
RohitRanjanMS: Would be safe to guard `_client` against `null`.

src/WebJobs.Script/Diagnostics/ApplicationInsightsMetricExporter.cs
```diff
@@ -0,0 +1,132 @@
...
+
+        /// <summary>
+        /// Initializes this module, starting the meter listener and exporting process.
+        /// </summary>
+        /// <param name="configuration">The telemetry configuration.</param>
+        public void Initialize(TelemetryConfiguration configuration)
+        {
+            ArgumentNullException.ThrowIfNull(configuration);
+            _client = new TelemetryClient(configuration);
+            _exportTask = CollectAsync(_shutdown.Token);
```
RohitRanjanMS: Should we start the listener before `exportTask`?

src/WebJobs.Script/Diagnostics/ApplicationInsightsMetricExporter.cs
```diff
@@ -0,0 +1,132 @@
...
+        private readonly ApplicationInsightsMetricExporterOptions _options;
+        private readonly CancellationTokenSource _shutdown = new();
+
+        private Task _exportTask = Task.CompletedTask;
+        private TelemetryClient _client = null!;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ApplicationInsightsMetricExporter"/> class.
+        /// </summary>
+        /// <param name="lifetime">The application lifetime.</param>
```
RohitRanjanMS: This can be removed.

src/WebJobs.Script/Diagnostics/ApplicationInsightsMetricExporter.cs
```diff
@@ -0,0 +1,132 @@
...
+                if (state is not ApplicationInsightsMetricExporter listener)
+                {
+                    return;
+                }
+
+                listener.Publish(instrument, value.ToDouble(null), tags);
+            };
+        }
+
+        private async Task CollectAsync(CancellationToken cancellation)
```
RohitRanjanMS, line 125: Do we need `CollectAsync`? We don't have any observable metrics.

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
RohitRanjanMS: In AppInsights mode, we generate a set of metrics related to function invocations. It would be good to filter out the invoke duration metric. It's redundant and may cause confusion.

src/WebJobs.Script/Diagnostics/ApplicationInsightsMetricExporter.cs
```diff
@@ -0,0 +1,119 @@
...
+            }
+        }
+
+        private void Publish(Instrument instrument, double value, ReadOnlySpan<KeyValuePair<string, object?>> tags)
+        {
+            if (instrument is null)
+            {
+                return;
+            }
+
```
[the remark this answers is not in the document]
jviau: Leaving it in as it causes no harm

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
jviau: Will filter them out

src/WebJobs.Script/Diagnostics/ApplicationInsightsMetricExporter.cs
jviau: No - you typically don't dispose `Task` objects. https://devblogs.microsoft.com/dotnet/do-i-need-to-dispose-of-tasks/

src/WebJobs.Script/Diagnostics/ApplicationInsightsMetricExporter.cs
jviau: It is for future proofing / completeness if we do add them in.

src/WebJobs.Script/Diagnostics/ApplicationInsightsMetricExporter.cs
RohitRanjanMS: Thanks for the link. 

Since _exportTask relies on _listener, I think it's best to call _shutdown.CancelNoThrowAsync() first, then dispose _listener. I don’t expect _listener.RecordObservableInstruments() to throw an exception if called after disposal, but it’s safer to follow this order.

src/WebJobs.Script/Diagnostics/ApplicationInsightsMetricExporter.cs
RohitRanjanMS: Are you considering adding observable metrics? It seems like unnecessary overhead. 

If we decide to keep it, I believe we should call `_listener.RecordObservableInstruments();` during dispose to capture the current snapshot.

src/WebJobs.Script/Diagnostics/ApplicationInsightsMetricExporter.cs
```diff
@@ -0,0 +1,143 @@
...
+        /// <summary>
+        /// Initializes this module, starting the meter listener and exporting process.
+        /// </summary>
+        /// <param name="configuration">The telemetry configuration.</param>
+        public void Initialize(TelemetryConfiguration configuration)
+        {
+            ArgumentNullException.ThrowIfNull(configuration);
+            ObjectDisposedException.ThrowIf(_shutdown.IsCancellationRequested, this);
+
+            Events.Log.MeterListeningStarted();
```
RohitRanjanMS: Should we place this after the listener has started?

src/WebJobs.Script/Diagnostics/ApplicationInsightsMetricExporter.cs
```diff
@@ -0,0 +1,143 @@
...
+        {
+            Events.Log.MeterListeningStopped();
+            _listener.Dispose();
+
+            await _shutdown.CancelNoThrowAsync();
+            await _exportTask.ConfigureAwait(false);
+
+            if (_client is { } client)
+            {
+                await client.FlushAsync(default).ConfigureAwait(false);
```
RohitRanjanMS, line 96: Adding a log after `FlushAsync` would help in troubleshooting.

src/WebJobs.Script/Diagnostics/ApplicationInsightsMetricExporter.cs
jviau: I'd like to keep it, as I want to avoid the scenario where we do add an observable metric and then forget it isn't exposed to app insights SDK yet.

I can add the call in the dispose method.

src/WebJobs.Script/Diagnostics/ApplicationInsightsMetricExporter.EventSource.cs
```diff
@@ -0,0 +1,51 @@
...
+using System;
+using System.Diagnostics.Metrics;
+using System.Diagnostics.Tracing;
+
+namespace Microsoft.Azure.WebJobs.Script.Diagnostics
+{
+    public sealed partial class ApplicationInsightsMetricExporter
+    {
+        [EventSource(Name = $"{ScriptConstants.HostDiagnosticSourcePrefix}Diagnostics.{nameof(ApplicationInsightsMetricExporter)}")]
+        private sealed class Events : EventSource
```
RohitRanjanMS, line 13: If we switch to `DiagnosticListener` instead of `EventSource`, the events will begin flowing to Kusto right away. I expect we’ll need these logs for troubleshooting, so having a configurable feature flag to enable them would be very helpful. 

Alternatively, we could add an intermediate listener that aggregates the events and then writes them through the `DiagnosticListener`. I’ve implemented a similar approach in `OpenTelemetryEventListener`.

jviau: I am going to stick with `EventSource` here. This is the norm I see for low-level structured logging when `ILogger` is not available. Otel SDK also uses `EventSource` for the same reasons. `EventSource` is what we should use for all low-level instrumentation.

If we want, we can make a single `EventListener` which will subscribe to all the `EventSource`'s we care about and write them to the diag source, like your `OpenTelemetryEventListener`

RohitRanjanMS: I will create a feature request to extend the OpenTelemetryEventListener to listen to other sources as well.

RohitRanjanMS (approved):

brettsam (approved):

PR #11369, opened 2025-09-25
Enabling Admin Isolation for Flex Consumption

mathewc:
Work has already been done on the platform side to enable Admin Isolation to work on Flex. The right environment state is flowed to the container on specialization (FUNCTIONS_ADMIN_ISOLATION_ENABLED). The work remaining was to fix the "platform internal" environmental checks so they work in the Flex worker environment.

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script/Extensions/HttpRequestExtensions.cs
+++ b/src/WebJobs.Script/Extensions/HttpRequestExtensions.cs
@@ -78,7 +78,8 @@ namespace Microsoft.Azure.WebJobs.Script.Extensions
         public static bool IsAppServiceInternalRequest(this HttpRequest request, IEnvironment environment = null)
         {
             environment = GetEnvironment(request, environment);
-            if (!environment.IsAppService())
+
+            if (!FrontEndRoutingEnsured(environment))
             {
                 return false;
             }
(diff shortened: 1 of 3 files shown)


mattchenderson (approved):

larohra (approved):

RohitRanjanMS (approved):

src/WebJobs.Script/Extensions/HttpRequestExtensions.cs
```diff
@@ -102,6 +104,13 @@ public static bool IsPlatformInternalRequest(this HttpRequest request, IEnvironm
             return string.Compare(value, "true", StringComparison.OrdinalIgnoreCase) == 0;
         }
 
+        private static bool FrontEndRoutingEnsured(this IEnvironment environment)
+        {
+            // Verify we're running on a sku that ensures all requests go through the
+            // Antares Front End.
+            return environment.IsAppService() || environment.IsFlexConsumptionSku();
```
balag0, line 111: >environment.IsFlexConsumptionSku() [](http://example.com/codeflow?start=49&length=34)

this extension method is a little convoluted and seems risky to rely on that for scenario. (i always find it difficult to parse all those conditions :) )

may be we should add a new extension method IsFlexConsumptionSkuV2 and use it here (and over time expand to other places). it would be better if it doesnt rely on any env. vars since we have had CRIs where customer sets some internal env. var such as website_is_placeholdermode etc.

Its an existing problem, but other places dont have similar implications as this scenario.

what do you think?

PR #11371, opened 2025-09-25
Add duplicates check in Sync Triggers

### LINKED ISSUE #11284: Check for duplicate function names before creating the sync triggers payload
While we have checks for duplicates in function indexing and route names, we do not have a check in place for the sync triggers payload [here](https://github.com/Azure/azure-functions-host/blob/0622c73d98cf1aa46023b41f61653d8f144731f4/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs#L307).

We need to add a check because duplicate names in the sync triggers payload can lead to a fatal error that requires deletion/recreation of an app.

satvu:

resolves #11284 

### Pull request checklist


src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -309,7 +309,9 @@ private async Task<SyncTriggersPayload> GetSyncTriggersPayload()
             PrepareSyncTriggers();
 
             var hostOptions = _applicationHostOptions.CurrentValue.ToHostOptions();
-            var functionsMetadata = _functionMetadataManager.GetFunctionMetadata().Where(m => !m.IsProxy());
+            var functionsMetadata = _functionMetadataManager.GetFunctionMetadata().Where(m => !m.IsProxy())
```
mathewc: recommend using the DistinctBy operator rather than GroupBy. It'll be more efficient since it doesn't build full groups (which you don't need), and just uses a HashSet behind the scenes to efficiently track keys and skip as needed.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -309,7 +309,9 @@ private async Task<SyncTriggersPayload> GetSyncTriggersPayload()
...
+                .GroupBy(m => m.Name, StringComparer.OrdinalIgnoreCase)
+                .Select(g => g.First());
```
mathewc: There's a second GetFunctionMetadata call in this method that is also listing functions (with a slightly different query) and turning them into an array. This isn't causing problems currently because apparently we don't have any code doing a ToDictionary on the values like we do in the other case.

Initially I was going to suggest we de-dupe there as well, but perhaps we don't need to. The "functions" section is supposed to represent all the listable functions, similar to our other "source of truth" arguments.

test/WebJobs.Script.Tests.Integration/Management/FunctionsSyncManagerTests.cs
```diff
@@ -1021,6 +1026,132 @@ public void Managed_Kubernetes_Environment_SyncTrigger_Url_Validation(string kub
...
+            var optionsMonitorMock = new Mock<IOptionsMonitor<ScriptApplicationHostOptions>>();
+            optionsMonitorMock.Setup(m => m.CurrentValue).Returns(scriptAppHostOptions);
+            optionsMonitorMock.Setup(m => m.Get(It.IsAny<string>())).Returns(scriptAppHostOptions);
+
+            var hostingConfigOptions = new FunctionsHostingConfigOptions();
+            // Optionally set properties on hostingConfigOptions as needed for your test
+            var hostingConfigOptionsMock = new Mock<IOptions<FunctionsHostingConfigOptions>>();
+            hostingConfigOptionsMock.Setup(m => m.Value).Returns(hostingConfigOptions);
+
+            // Use all other dependencies from the test class, but replace IFunctionMetadataManager
```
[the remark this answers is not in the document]
mathewc: Yes, I was thinking the same thing :)

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
satvu: `listableFunctions` is only referenced twice in this file (when value is assigned, when added as an array to full syncTriggers payload as you mentioned). The one I de-deduped is used for triggers and secrets (which are more important to de-dupe). 

I think having the "source of truth" in this case is fine since it shouldn't block sync triggers updates like this bug did. Using the listable functions (probably used for display in portal?) could help customers figure out they have a problem more quickly.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -309,7 +309,23 @@ private async Task<SyncTriggersPayload> GetSyncTriggersPayload()
...
+                .GroupBy(m => m.Name, StringComparer.OrdinalIgnoreCase)
+                .Where(g => g.Count() > 1)
+                .ToList();
+
+            if (duplicateGroups.Count > 0)
+            {
+                var duplicateInfo = string.Join("; ",
+                    duplicateGroups.Select(g =>
+                        $"Name: '{g.Key}', Files: [{string.Join(", ", g.Select(f => f.ScriptFile ?? "<null>"))}]"));
+                _logger.LogDebug("Duplicate function metadata detected: {duplicateInfo}. Please review your functions again and remove duplicate names.", duplicateInfo);
```
satvu: Wondering if this is worth the trouble of logging since we definitely log it while validating function metadata already (at error level). 

Could also change wording to "Duplicate function metadata detected while preparing the sync triggers payload" so we have a better idea of when/where this check is performed?

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -309,7 +309,23 @@ private async Task<SyncTriggersPayload> GetSyncTriggersPayload()
             PrepareSyncTriggers();
 
             var hostOptions = _applicationHostOptions.CurrentValue.ToHostOptions();
-            var functionsMetadata = _functionMetadataManager.GetFunctionMetadata().Where(m => !m.IsProxy());
+            var allFunctionsMetadata = _functionMetadataManager.GetFunctionMetadata().Where(m => !m.IsProxy()).ToList();
+            var functionsMetadata = allFunctionsMetadata.Where(m => !m.IsProxy())
+                .DistinctBy(m => m.Name, StringComparer.OrdinalIgnoreCase);
+
+            // Log any duplicate function names and their script files
+            var duplicateGroups = allFunctionsMetadata
```
kshyju: Do you know what is actually causing the duplicate data? Should we be fixing there?

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -309,7 +309,23 @@ private async Task<SyncTriggersPayload> GetSyncTriggersPayload()
             PrepareSyncTriggers();
 
             var hostOptions = _applicationHostOptions.CurrentValue.ToHostOptions();
-            var functionsMetadata = _functionMetadataManager.GetFunctionMetadata().Where(m => !m.IsProxy());
+            var allFunctionsMetadata = _functionMetadataManager.GetFunctionMetadata().Where(m => !m.IsProxy()).ToList();
+            var functionsMetadata = allFunctionsMetadata.Where(m => !m.IsProxy())
```
surgupta-msft: wondering if we require this check `Where(m => !m.IsProxy())` again on `allFunctionsMetadata` as we filtered it in line 312?

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
surgupta-msft: I am also missing some context on these changes. As per my understanding, we cannot have functions with same names in the app. (Tried with VS, it did not add another function with the same name and different casing).

There is also this code which throws for duplicate name. I think it is not considering different casing - may be we can fix here or move it inside `GetFunctionMetadata()` - https://github.com/Azure/azure-functions-host/blob/920b01e79c0f5b709d0a433ecafb552d983dadfe/src/WebJobs.Script/Host/FunctionMetadataManager.cs#L270

One more point, if today in any scenario we allow functions with duplicate names due to different casing then this change will be breaking those scenarios. Below test succeeds with different casing -

https://github.com/Azure/azure-functions-host/blob/920b01e79c0f5b709d0a433ecafb552d983dadfe/test/WebJobs.Script.Tests/FunctionMetadataManagerTests.cs#L142

satvu: Tagged you both on a thread discussing why this happens - those checks do exist, and those errors are being thrown and logged, but there is no actual modification to the function metadata payload once these issues are found. That payload (with duplicates) is still used here in sync triggers leading to unintended consequences where an app is unable to update their sync triggers and the only resolution is to delete and recreate the app.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
satvu: Good point - will add that to this.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
liliankasem: If we already log something like this, I think `Duplicate function metadata detected while preparing the sync triggers payload` makes more sense here just as an FYI whist investigating

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -309,7 +309,22 @@ private async Task<SyncTriggersPayload> GetSyncTriggersPayload()
...
+                .Where(g => g.Count() > 1)
+                .ToList();
+
+            if (duplicateGroups.Count > 0)
+            {
+                var duplicateInfo = string.Join("; ",
+                    duplicateGroups.Select(g =>
+                        $"Name: '{g.Key}', Files: [{string.Join(", ", g.Select(f => f.ScriptFile ?? "<null>"))}]"));
+                _logger.LogDebug("Duplicate function metadata detected: {duplicateInfo}. Please review your functions again and remove duplicate names.", duplicateInfo);
+            }
```
surgupta-msft: We can move line 315-327 to a different method and call it here.

surgupta-msft: We can also optimize it by parsing list only once. One approach could be, instead of using linq, iterate on `allFunctionsMetadata` and as soon as you find a duplicate, append it to the output string.

Quick sample code from copilot -

```csharp
// Log any duplicate function names and their script files
var seenNames = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
var duplicateInfoBuilder = new StringBuilder();
foreach (var metadata in allFunctionsMetadata)
{
    if (!seenNames.Add(metadata.Name))
    {
        duplicateInfoBuilder.Append($"Name: '{metadata.Name}', Files: [{metadata.ScriptFile ?? "<null>"}]; ");
    }
}

if (duplicateInfoBuilder.Length > 0)
{
    var duplicateInfo = duplicateInfoBuilder.ToString().TrimEnd(' ', ';');
    _logger.LogDebug("Duplicate function metadata detected: {duplicateInfo}. Please review your functions again and remove duplicate names.", duplicateInfo);
}
```

satvu: I went with this approach: https://github.com/Azure/azure-functions-host/pull/11371#discussion_r2392548364.
The duplicates are already logged in a few other places (as error) so I am simplifying the message to use as an investigative helper.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -309,7 +309,21 @@ private async Task<SyncTriggersPayload> GetSyncTriggersPayload()
             PrepareSyncTriggers();
 
             var hostOptions = _applicationHostOptions.CurrentValue.ToHostOptions();
-            var functionsMetadata = _functionMetadataManager.GetFunctionMetadata().Where(m => !m.IsProxy());
+            var allFunctionsMetadata = _functionMetadataManager.GetFunctionMetadata().Where(m => !m.IsProxy()).ToList();
+            var functionsMetadata = allFunctionsMetadata.DistinctBy(m => m.Name, StringComparer.OrdinalIgnoreCase);
```
mathewc: You addressed my original comment and used the more performance DistinctBy, then readded GroupBy again below for the logging. Can we just forget this logging?

If you really want to log the fact that there are duplicates, then I think we'd do that in FunctionMetadataManager.LoadFunctionMetadata when we're building the metadata list. That way its only done once when the list is generated. Looks like that function is already doing a final validation pass over all functions, so it'd be a simple matter to use a HashSet to catch dupes and log.

Though I did share in the teams thread that ultimately we DO already log duplicates during indexing time already. The code [here](https://github.com/Azure/azure-webjobs-sdk/blob/a2e31999338ea7443aedb4105483dc98d0eff168/src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndex.cs#L30C55-L30C90) throws and exception that the host logs.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
mathewc: Yes, it's a shame that the code https://github.com/Azure/azure-functions-host/blob/920b01e79c0f5b709d0a433ecafb552d983dadfe/src/WebJobs.Script/Host/FunctionMetadataManager.cs#L268 that was added 5 years ago for catching duplicates was case sensitive incorrectly.

I think historically duplicates were naturally prevented because we didn't support worker indexing, and functions were determined based on directory names which prevented dupes, and also derived from .NET method names which again don't allow duplicates.

So based on what @surgupta-msft shared above, it currently IS possible for providers to register duplicates that differ only by case. And I'm not sure the worker side check Sarah is proposing would prevent dupes in all cases, since metadata comes from multiple providers as the test https://github.com/Azure/azure-functions-host/blob/790f2d84e25c0fac2434b41bd2588d1cebd4a473/test/WebJobs.Script.Tests/FunctionMetadataManagerTests.cs#L396 demonstrates.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -309,7 +309,7 @@ private async Task<SyncTriggersPayload> GetSyncTriggersPayload()
             PrepareSyncTriggers();
 
             var hostOptions = _applicationHostOptions.CurrentValue.ToHostOptions();
-            var functionsMetadata = _functionMetadataManager.GetFunctionMetadata().Where(m => !m.IsProxy());
+            var functionsMetadata = _functionMetadataManager.GetFunctionMetadata().Where(m => !m.IsProxy()).DistinctBy(m => m.Name, StringComparer.OrdinalIgnoreCase);
```
liliankasem, line 312: Just want to double check that `DistinctBy` sufficient for this? It keeps the first occurrence it sees and removes subsequent ones, does ordering matter here?

satvu: Yes, first one wins is also how the [indexer works](https://github.com/Azure/azure-webjobs-sdk/blob/a2e31999338ea7443aedb4105483dc98d0eff168/src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndex.cs#L26). @mathewc recommended this approach [here](https://github.com/Azure/azure-functions-host/pull/11371#discussion_r2383141200) since we wanted a first-one-wins strategy. 

The places that use function metadata (generator, indexer, in this file, etc) call `GetFunctionMetadata` from `FunctionMetadataManager` so the order should be the same (and the first one that wins will be synced).

liliankasem (approved):

soninaren (approved):

satvu:

jviau:

liliankasem:

satvu:

PR #11385, opened 2025-09-30
Multi-target WebJobs.Script

jviau:

resolves #issue_for_this_pr

### Pull request checklist

### Additional information

Updates `WebJobs.Script.csproj` to multi-target net6 and net8 (instead of being netstandard2.1). Addresses new warnings that come about as a result. Additionally embeds only the single `runtimeassemblies.json` we need for the given TFM.


liliankasem (approved):

src/WebJobs.Script/Extensions/AssemblyExtensions.cs
```diff
@@ -25,7 +25,7 @@ public static string GetCodeBase(this Assembly assembly)
                 return codeBase;
             }
 
-            return assembly.CodeBase;
+            return assembly.Location;
```
fabiocav: This a change in behavior and format. Will be fine in many scenarios, but have we validated we're OK with the change in this context?

fabiocav: The Code Base mapping already handles the scenarios for dynamic assemblies loaded from bytes, so we'll be OK there as we should never hit this fallback.

fabiocav: Was this even being hit?

jviau: This entire file is completely unused. I just removed it.

src/WebJobs.Script/Workers/Rpc/Configuration/RpcWorkerConfigFactory.cs
```diff
@@ -64,13 +64,15 @@ public IList<RpcWorkerConfig> GetConfigs()
 
         internal static string GetDefaultWorkersDirectory(Func<string, bool> directoryExists)
         {
-            string assemblyLocalPath = Path.GetDirectoryName(new Uri(typeof(RpcWorkerConfigFactory).Assembly.CodeBase).LocalPath);
-            string workersDirPath = Path.Combine(assemblyLocalPath, RpcWorkerConstants.DefaultWorkersDirectoryName);
+            string assemblyDir = AppContext.BaseDirectory;
```
jviau, line 67: Ported logic from dev: https://github.com/Azure/azure-functions-host/blob/800db8c08618fdeb3972a20d84419286bc01b974/src/WebJobs.Script/Workers/Rpc/Configuration/WorkerConfigurationResolverOptionsSetup.cs#L36

fabiocav (approved):

src/WebJobs.Script/Extensions/AssemblyExtensions.cs
fabiocav: Good. Comment applies to the change from `CodeBase` -> `Location` through the code. But looks like you've also updated RpcWorkerConfig, so we should be in a good state now.

surgupta-msft (approved):

liliankasem (approved):

PR #11398, opened 2025-10-01
Add keyed ILogger which forwards to ScriptHost when possible

### LINKED ISSUE #11400: [HealthChecks] TelemetryHealthCheckPublisher does not emit logs to ScriptHost/customer logs
`TelemetryHealthCheckPublisher` currently only emits logs to WebHost (Kusto) logs. We should emit these logs to customer logs as well.

This service lives in the WebHost, so importing `ILogger` does not give a logger for customers. To fix this, we will leverage keyed services and add a new `ForwardingLogger` which will emit to the active ScriptHost if it is available.

jviau:

resolves #11400

### Pull request checklist

### Additional information

This PR introduces a mechanism where WebHost services can opt into having their logs forwarded to the active ScriptHost `ILogger` when it is available. When a ScriptHost is not available, logs will go to the WebHost `ILogger`. The implementation automatically tracks changes to active ScriptHost, while avoiding memory leaks through the use of `WeakReference<T>` (allowed active ScriptHost `ILogger` to be cleaned up on ScriptHost shutdown).

#### Usage

Simply add `[ForwardingLogger]` attribute to `ILogger<T>` or `ILoggerFactory` import:

``` CSharp
public class MyService([ForwardingLogger] ILogger<MyService> logger)
{
}
```

#### TODO

- ~Evaluate how/if the added keyed services are being proxied to the ScriptHost services~ I have excluded `ILoggerFactory` and `ILogger<>` from being copied to script host service container. They were not actually used as ScriptHost overrode with their own, but the copying process did fail validation.


src/WebJobs.Script/Host/IScriptHostManager.cs
```diff
@@ -24,7 +26,12 @@ public interface IScriptHostManager
         /// <summary>
         /// Gets the last host <see cref="Exception"/> that has occurred.
         /// </summary>
-        Exception LastError { get; }
+        Exception? LastError { get; }
+
+        /// <summary>
+        /// Gets the current <see cref="IServiceProvider"/> for the active Script Host.
+        /// </summary>
+        IServiceProvider? Services { get; }
```
jviau, line 37: I have found a couple times recently that having access to `IServiceProvider` here is useful. Not sure if there is a reason we didn't expose it in the past?

jviau:
`serviceProvider.CreateChildContainer(..)` does not currently support keyed services. We will need to work on that. Or in this case, I can just block `ILoggerFactory` from propagating. It isn't used anyways - the ScriptHost overrides it.

RohitRanjanMS (approved):

src/WebJobs.Script/Host/IScriptHostManager.cs
fabiocav: When `IScriptHostManager` was initially introduced, this wasn't positioned as requirement implementers would need to meet. It's fine to add this here now

src/WebJobs.Script/Diagnostics/ForwardingLogger.cs
```diff
@@ -0,0 +1,110 @@
...
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+using Microsoft.Azure.WebJobs.Script;
+using Microsoft.Extensions.DependencyInjection;
+
+#nullable enable
+
+namespace Microsoft.Extensions.Logging
+{
+    internal class ForwardingLogger : ILogger
```
fabiocav: seal type

src/WebJobs.Script/Diagnostics/ForwardingLogger.cs
```diff
@@ -0,0 +1,110 @@
...
+                return _current.TryGetTarget(out logger);
+            }
+
+            logger = null;
+            return false;
+        }
+    }
+
+    [DebuggerDisplay("{_logger}")]
+    internal class ForwardingLogger<T> : ILogger<T>
```
fabiocav: seal

src/WebJobs.Script/Diagnostics/ForwardingLoggerFactory.cs
```diff
@@ -0,0 +1,68 @@
...
+
+#nullable enable
+
+namespace Microsoft.Extensions.Logging
+{
+    /// <summary>
+    /// A logger factory that creates loggers which track the current active ScriptHost (if any), falling
+    /// back to the WebHost logger if no ScriptHost is active.
+    /// </summary>
+    [DebuggerDisplay(@"InnerFactory = \{ {_inner} \}, ScriptHostState = {_manager.State}")]
```
fabiocav: Simplify this to `[DebuggerDisplay("InnerFactory = { _inner }, ScriptHostState = { _manager.State }")]`

src/WebJobs.Script/Diagnostics/ForwardingLoggerFactory.cs
```diff
@@ -0,0 +1,68 @@
...
+        public void AddProvider(ILoggerProvider provider)
+            => throw new NotSupportedException(
+                $"{nameof(ILoggerProvider)} can not be added to the {nameof(ForwardingLoggerFactory)}.");
+
+        /// <inheritdoc />
+        public ILogger CreateLogger(string categoryName)
+        {
+            ObjectDisposedException.ThrowIf(_disposed, this);
+
+            if (!_loggers.TryGetValue(categoryName, out ForwardingLogger? logger))
```
fabiocav: Swapping to using `GetOrAdd` here will avoid the need of the double check lock and simplify this a bit. In several scenarios, it will also give us a perf improvement.

fabiocav (approved):
Flagging as approved, but left some comments here.

RohitRanjanMS (approved):

RohitRanjanMS (approved):

PR #11408, opened 2025-10-08
Functions host to take a customer specified port in Custom Handler scenario

### LINKED ISSUE #11305: [Feature request] Have Functions host take a customer specified port for FUNCTIONS_CUSTOMHANDLER_PORT
When hosting BYO MCP servers on Functions, customers need to specify the server port for the host to listen to. This is the only thing code change that customers need to make for hosting on Functions. For example, for Python, they have to do this: 

```python
mcp_port = int(os.environ.get("FUNCTIONS_CUSTOMHANDLER_PORT", 8080)) 
mcp = FastMCP("weather", stateless_http=True, port=mcp_port)
```

Removing the need for this change would be value-add for customers who don't have direct control over the MCP server code - for example, they want to just host some off-the-shelf server. 

To do this, we need to have the Functions host take a port passed in by the customer during app deployment.

surgupta-msft:

resolves #11305 

### Pull request checklist

### Additional information

This PR enables Host to accept a customer-specified port in Custom Handler scenarios. Host will take the port specified via host.json and throws an exception if it is not able to bind to that port. If host.json does not specify any port then the Host will continue to use the logic of dynamic port allocation. The PR also has the logic to enforce sequential restarts when port is taken from configuration to avoid multiple processes binding to the same port.

### DIFF
--- a/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
+++ b/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
@@ -685,12 +685,9 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
 
         internal bool ShouldEnforceSequentialRestart(IHost host)
         {
-            var options = host?.Services?.GetService<IOptions<ScriptHostRecycleOptions>>();
-            var jobHostSequentialRestartRequired = options?.Value?.SequentialHostRestartRequired ?? false;
-
-            var applicationSequentialRestartRequired = _applicationHostOptions.CurrentValue.SequentialHostRestartRequired;
-
-            return jobHostSequentialRestartRequired || applicationSequentialRestartRequired;
+            var options = host?.Services?.GetService<IOptions<ScriptHostRecycleOptions>>().Value;
+            options ??= ScriptHostRecycleOptions.Create(_config);
+            return options.SequentialHostRestartRequired;
         }
 
         private void OnHostInitializing(object sender, EventArgs e)
--- a/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
+++ b/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
@@ -686,17 +686,13 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
         internal bool ShouldEnforceSequentialRestart(IHost host)
         {
             var options = host?.Services?.GetService<IOptions<ScriptHostRecycleOptions>>();
-            var sequentialRestartRequired = options?.Value?.SequentialHostRestartRequired ?? false;
-            if (sequentialRestartRequired)
+            var jobHostSequentialRestartRequired = options?.Value?.SequentialHostRestartRequired ?? false;
+
+            var applicationSequentialRestartRequired = _applicationHostOptions.CurrentValue.SequentialHostRestartRequired;
+
+            if (jobHostSequentialRestartRequired || applicationSequentialRestartRequired)
             {
                 return true;
-            }
-
-            var sequentialRestartSetting = _config.GetSection(ConfigurationSectionNames.SequentialJobHostRestart);
-            if (sequentialRestartSetting != null)
-            {
-                bool.TryParse(sequentialRestartSetting.Value, out bool enforceSequentialOrder);
-                return enforceSequentialOrder;
             }
 
             return false;
--- a/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
+++ b/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
@@ -690,12 +690,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
 
             var applicationSequentialRestartRequired = _applicationHostOptions.CurrentValue.SequentialHostRestartRequired;
 
-            if (jobHostSequentialRestartRequired || applicationSequentialRestartRequired)
-            {
-                return true;
-            }
-
-            return false;
+            return jobHostSequentialRestartRequired || applicationSequentialRestartRequired;
         }
 
         private void OnHostInitializing(object sender, EventArgs e)
--- a/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
+++ b/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
@@ -686,13 +686,10 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
         internal bool ShouldEnforceSequentialRestart(IHost host)
         {
             var options = host?.Services?.GetService<IOptions<ScriptHostRecycleOptions>>();
-            if (options is not null)
+            var sequentialRestartRequired = options?.Value?.SequentialHostRestartRequired ?? false;
+            if (sequentialRestartRequired)
             {
-                var sequentialRestartRequired = options.Value?.SequentialHostRestartRequired ?? false;
-                if (sequentialRestartRequired)
-                {
-                    return true;
-                }
+                return true;
             }
 
             var sequentialRestartSetting = _config.GetSection(ConfigurationSectionNames.SequentialJobHostRestart);
--- a/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
+++ b/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
@@ -683,7 +683,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
             dispatcher?.PreShutdown();
         }
 
-        internal bool ShouldEnforceSequentialRestart(IHost host)
+        internal bool ShouldEnforceSequentialRestart(IHost host = null)
         {
             var options = host?.Services?.GetService<IOptions<ScriptHostRecycleOptions>>().Value;
             options ??= ScriptHostRecycleOptions.Create(_config);
--- a/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
+++ b/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
@@ -630,7 +630,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
                         // If we are running in development mode with core tools, do not overlap the restarts.
                         // Overlapping restarts are problematic when language worker processes are listening
                         // to the same debug port
-                        if (ShouldEnforceSequentialRestart())
+                        if (ShouldEnforceSequentialRestart(previousHost))
                         {
                             stopTask = Orphan(previousHost, cancellationToken);
                             await stopTask;
@@ -683,8 +683,18 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
             dispatcher?.PreShutdown();
         }
 
-        internal bool ShouldEnforceSequentialRestart()
+        internal bool ShouldEnforceSequentialRestart(IHost host)
         {
+            var options = host?.Services?.GetService<IOptions<ScriptHostRecycleOptions>>();
+            if (options is not null)
+            {
+                var sequentialRestartRequired = options.Value?.SequentialHostRestartRequired ?? false;
+                if (sequentialRestartRequired)
+                {
+                    return true;
+                }
+            }
+
             var sequentialRestartSetting = _config.GetSection(ConfigurationSectionNames.SequentialJobHostRestart);
             if (sequentialRestartSetting != null)
             {
--- a/src/WebJobs.Script/Config/ScriptApplicationHostOptions.cs
+++ b/src/WebJobs.Script/Config/ScriptApplicationHostOptions.cs
@@ -1,4 +1,4 @@
-// Copyright (c) .NET Foundation. All rights reserved.
+﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
--- a/src/WebJobs.Script/Config/ScriptApplicationHostOptions.cs
+++ b/src/WebJobs.Script/Config/ScriptApplicationHostOptions.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
@@ -35,5 +35,7 @@ namespace Microsoft.Azure.WebJobs.Script
         public bool IsFileSystemReadOnly { get; set; }
 
         public bool IsScmRunFromPackage { get; set; }
+
+        public bool SequentialHostRestartRequired { get; set; }
     }
 }
--- a/src/WebJobs.Script/Config/ScriptApplicationHostOptions.cs
+++ b/src/WebJobs.Script/Config/ScriptApplicationHostOptions.cs
@@ -35,7 +35,5 @@ namespace Microsoft.Azure.WebJobs.Script
         public bool IsFileSystemReadOnly { get; set; }
 
         public bool IsScmRunFromPackage { get; set; }
-
-        public bool SequentialHostRestartRequired { get; set; }
     }
 }
--- a/src/WebJobs.Script/Config/ScriptHostRecycleOptions.cs
+++ b/src/WebJobs.Script/Config/ScriptHostRecycleOptions.cs
@@ -1,0 +1,29 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Microsoft.Azure.WebJobs.Hosting;
+
+namespace Microsoft.Azure.WebJobs.Script
+{
+    /// <summary>
+    /// Options that control Script Host recycling behavior.
+    /// </summary>
+    public sealed class ScriptHostRecycleOptions : IOptionsFormatter
+    {
+        /// <summary>
+        /// Gets or sets a value indicating whether sequential host restarts are required.
+        /// </summary>
+        public bool SequentialHostRestartRequired { get; set; }
+
+        public string Format()
+        {
+            return JsonSerializer.Serialize(this, typeof(ScriptHostRecycleOptions), ScriptHostRecycleOptionsJsonContext.Default);
+        }
+    }
+
+    [JsonSourceGenerationOptions(WriteIndented = true)]
+    [JsonSerializable(typeof(ScriptHostRecycleOptions))]
+    internal partial class ScriptHostRecycleOptionsJsonContext : JsonSerializerContext;
+}
--- a/src/WebJobs.Script/Config/ScriptHostRecycleOptions.cs
+++ b/src/WebJobs.Script/Config/ScriptHostRecycleOptions.cs
@@ -23,8 +23,6 @@ namespace Microsoft.Azure.WebJobs.Script
         {
             ScriptHostRecycleOptions options = new();
             options.Configure(configuration);
-
-            // set from configuration
             return options;
         }
 
--- a/src/WebJobs.Script/Config/ScriptHostRecycleOptions.cs
+++ b/src/WebJobs.Script/Config/ScriptHostRecycleOptions.cs
@@ -4,6 +4,8 @@
 using System.Text.Json;
 using System.Text.Json.Serialization;
 using Microsoft.Azure.WebJobs.Hosting;
+using Microsoft.Azure.WebJobs.Script.Configuration;
+using Microsoft.Extensions.Configuration;
 
 namespace Microsoft.Azure.WebJobs.Script
 {
@@ -17,9 +19,28 @@ namespace Microsoft.Azure.WebJobs.Script
         /// </summary>
         public bool SequentialHostRestartRequired { get; set; }
 
+        public static ScriptHostRecycleOptions Create(IConfiguration configuration)
+        {
+            ScriptHostRecycleOptions options = new();
+            options.Configure(configuration);
+
+            // set from configuration
+            return options;
+        }
+
         public string Format()
         {
             return JsonSerializer.Serialize(this, typeof(ScriptHostRecycleOptions), ScriptHostRecycleOptionsJsonContext.Default);
+        }
+
+        internal void Configure(IConfiguration configuration)
+        {
+            var sequentialRestartSetting = configuration.GetSection(ConfigurationSectionNames.SequentialJobHostRestart);
+            if (sequentialRestartSetting != null)
+            {
+                _ = bool.TryParse(sequentialRestartSetting.Value, out bool enforceSequentialOrder);
+                SequentialHostRestartRequired = enforceSequentialOrder;
+            }
         }
     }
 
--- a/src/WebJobs.Script/Config/ScriptHostRecycleOptions.cs
+++ b/src/WebJobs.Script/Config/ScriptHostRecycleOptions.cs
@@ -23,7 +23,7 @@ namespace Microsoft.Azure.WebJobs.Script
         }
     }
 
-    [JsonSourceGenerationOptions(WriteIndented = true)]
+    [JsonSourceGenerationOptions(GenerationMode = JsonSourceGenerationMode.Serialization, WriteIndented = true)]
     [JsonSerializable(typeof(ScriptHostRecycleOptions))]
     internal partial class ScriptHostRecycleOptionsJsonContext : JsonSerializerContext;
 }
--- a/src/WebJobs.Script/Config/ScriptHostRecycleOptionsSetup.cs
+++ b/src/WebJobs.Script/Config/ScriptHostRecycleOptionsSetup.cs
@@ -19,7 +19,7 @@ namespace Microsoft.Azure.WebJobs.Script.Configuration
         public void Configure(ScriptHostRecycleOptions options)
         {
             // Enforcing sequential host restarts when a user-specified custom handler port is configured to prevent multiple processes from attempting to bind to the same port concurrently.
-            if (_httpWorkerOptions.Value.IsPortManuallySet)
+            if (_httpWorkerOptions.Value.IsPortManuallySet.Value)
             {
                 options.SequentialHostRestartRequired = true;
             }
--- a/src/WebJobs.Script/Config/ScriptHostRecycleOptionsSetup.cs
+++ b/src/WebJobs.Script/Config/ScriptHostRecycleOptionsSetup.cs
@@ -3,6 +3,7 @@
 
 using System;
 using Microsoft.Azure.WebJobs.Script.Workers.Http;
+using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Options;
 
 namespace Microsoft.Azure.WebJobs.Script.Configuration
@@ -10,14 +11,20 @@ namespace Microsoft.Azure.WebJobs.Script.Configuration
     internal sealed class ScriptHostRecycleOptionsSetup : IConfigureOptions<ScriptHostRecycleOptions>
     {
         private readonly IOptions<HttpWorkerOptions> _httpWorkerOptions;
+        private readonly IConfiguration _configuration;
 
-        public ScriptHostRecycleOptionsSetup(IOptions<HttpWorkerOptions> httpWorkerOptions)
+        public ScriptHostRecycleOptionsSetup(
+            IOptions<HttpWorkerOptions> httpWorkerOptions, IConfiguration configuration)
         {
             _httpWorkerOptions = httpWorkerOptions ?? throw new ArgumentNullException(nameof(httpWorkerOptions));
+            _configuration = configuration ?? throw new ArgumentNullException(nameof(configuration));
         }
 
         public void Configure(ScriptHostRecycleOptions options)
         {
+            // check configuration
+            options.Configure(_configuration);
+
             // Enforcing sequential host restarts when a user-specified custom handler port is configured to prevent multiple processes from attempting to bind to the same port concurrently.
             if (_httpWorkerOptions.Value.IsPortManuallySet.Value)
             {
--- a/src/WebJobs.Script/Config/ScriptHostRecycleOptionsSetup.cs
+++ b/src/WebJobs.Script/Config/ScriptHostRecycleOptionsSetup.cs
@@ -22,7 +22,6 @@ namespace Microsoft.Azure.WebJobs.Script.Configuration
 
         public void Configure(ScriptHostRecycleOptions options)
         {
-            // check configuration
             options.Configure(_configuration);
 
             // Enforcing sequential host restarts when a user-specified custom handler port is configured to prevent multiple processes from attempting to bind to the same port concurrently.
--- a/src/WebJobs.Script/Config/ScriptHostRecycleOptionsSetup.cs
+++ b/src/WebJobs.Script/Config/ScriptHostRecycleOptionsSetup.cs
@@ -25,7 +25,7 @@ namespace Microsoft.Azure.WebJobs.Script.Configuration
             options.Configure(_configuration);
 
             // Enforcing sequential host restarts when a user-specified custom handler port is configured to prevent multiple processes from attempting to bind to the same port concurrently.
-            if (_httpWorkerOptions.Value.IsPortManuallySet.Value)
+            if (_httpWorkerOptions.Value.IsPortManuallySet)
             {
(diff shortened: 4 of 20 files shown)


src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -683,8 +683,18 @@ private static void NotifyHostStopping(IHost previousHost)
             dispatcher?.PreShutdown();
         }
 
-        internal bool ShouldEnforceSequentialRestart()
+        internal bool ShouldEnforceSequentialRestart(IHost host)
         {
+            var options = host?.Services?.GetService<IOptions<ScriptHostRecycleOptions>>();
+            if (options is not null)
```
soninaren: Can skip this null check by replacing the below statement with 
`var sequentialRestartRequired = options?.Value?.SequentialHostRestartRequired ?? false;`

src/WebJobs.Script/Config/ScriptHostRecycleOptions.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Microsoft.Azure.WebJobs.Hosting;
+
+namespace Microsoft.Azure.WebJobs.Script
+{
+    /// <summary>
+    /// Options that control Script Host recycling behavior.
+    /// </summary>
+    public sealed class ScriptHostRecycleOptions : IOptionsFormatter
```
soninaren, line 15: Why do we need `IOptionsFormatter` here?

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptions.cs
```diff
@@ -43,6 +49,11 @@ public class HttpWorkerOptions : IOptionsFormatter
         /// </summary>
         public CustomHandlerHttpOptions Http { get; set; }
 
+        internal int GetPortOrDefault()
+        {
+            return Port ?? 0;
```
soninaren: Do we have a scenario where this will ever be 0. If the port is not configured by the user, then we will assign a dynamic port.  This method can likely be removed.

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptionsSetup.cs
```diff
@@ -130,7 +130,21 @@ private void ConfigureWorkerDescription(HttpWorkerOptions options, IConfiguratio
 
             options.Arguments.ExecutableArguments.AddRange(options.Description.Arguments);
             options.Arguments.WorkerArguments.AddRange(options.Description.WorkerArguments);
-            options.Port = WorkerUtilities.GetUnusedTcpPort();
+
+            if (options.Port is not null)
+            {
+                if (options.Port == 0 || !WorkerUtilities.CanBindToPort(options.Port.Value))
```
soninaren: We might be able to skip the `options.Port == 0` check. if the user sets it to 0, the `CanBindToPort` will simply return false.

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptionsSetup.cs
```diff
@@ -130,7 +130,21 @@ private void ConfigureWorkerDescription(HttpWorkerOptions options, IConfiguratio
...
+                {
+                    throw new HostConfigurationException($"Unable to bind to port {options.Port} specified in configuration. Please specify a different port or remove the section to allow dynamic binding of port.");
+                }
+
+                options.IsPortConfigured = true;
+            }
+            else
+            {
+                // If the port is not specified in the custom handler configuration, use an unused port.
+                options.Port = WorkerUtilities.GetUnusedTcpPort();
```
jviau: Consider adding an `options.IsPortConfigured = false` here. _Technically_ `IsPortConfigured` is a valid host.json option now, so a user could have:

``` json
{
    "customHandler:isPortConfigured": true
}
```

and not actually set any port.

Or maybe there is an alternative way to express this configuration so we don't have the independently set _public_ properties. See my other comment for an approach for that.

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptions.cs
jviau: Will `Port` ever be `null` in practice? `HttpWorkerOptionsSetup` will set `Port` if it isn't already set. In what scenario will port be unset?

If that is the case, we could build this behavior into `Port`:

``` CSharp
private int? _port;

public int Port
{
    get => LazyInitializer.EnsureInitialized(ref _port, () => WorkerUtilities.GetUnusedTcpPort());
    set
    {
        if (value == 0)
        {
            // or throw if you want to
            IsPortManuallySet = false;
            _port = null;
            return;
        }

        IsPortManuallySet = true;
        _port = value;
    }
}

public bool IsPortManuallySet { get; private set; }
```

This approach will also prevent anyone from setting `IsPortManuallySet` via host.json (it is a private setter)

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptions.cs
```diff
@@ -16,7 +16,13 @@ public class HttpWorkerOptions : IOptionsFormatter
...
         public WorkerProcessArguments Arguments { get; set; }
 
-        public int Port { get; set; }
+        public int? Port { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether the <see cref="Port"/> property value is taken from configuration.
+        /// True value indicates that the host will use the configured port value rather than allocating a dynamic port.
+        /// </summary>
+        public bool IsPortConfigured { get; set; }
```
soninaren: May be change to `IsCustomPortConfigured` or something similar to add context.

test/WebJobs.Script.Tests/Configuration/HttpWorkerOptionsSetupTests.cs
```diff
@@ -257,20 +258,22 @@ public void HttpWorker_Config_ExpectedValues(string hostJsonContent, bool append
                                 'arguments': ['httpWorker.js'],
                                 'workingDirectory': 'c:/myWorkingDir',
                                 'workerDirectory': 'c:/myWorkerDir'
-                            }
+                            },
+                            'port': 1234
                         }
-                    }", false, false, false)]
+                    }", false, false, false, 1234)]
```
soninaren: we need additional tests where port is dynamically configured.

soninaren (dismissed):

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptionsSetup.cs
surgupta-msft: I tested with port 0, CanBindToPort() is not returning false. May be it is considering 0 to look for any available port.

src/WebJobs.Script/Config/ScriptHostRecycleOptions.cs
surgupta-msft: Shyju recommended that we use IOptionsFormatter. It helps with logging options in kusto.

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptions.cs
surgupta-msft: Yeah the port should never be null in practice. I updated the code based on my understanding of your comments. Let me know if further updates are needed.

src/WebJobs.Script/Config/ScriptHostRecycleOptionsSetup.cs
```diff
@@ -0,0 +1,25 @@
...
+
+        public ScriptHostRecycleOptionsSetup(IOptions<HttpWorkerOptions> httpWorkerOptions)
+        {
+            _httpWorkerOptions = httpWorkerOptions ?? throw new ArgumentNullException(nameof(httpWorkerOptions));
+        }
+
+        public void Configure(ScriptHostRecycleOptions options)
+        {
+            // Enforcing sequential host restarts when a user-specified custom handler port is configured to prevent multiple processes from attempting to bind to the same port concurrently.
+            options.SequentialHostRestartRequired = _httpWorkerOptions.Value?.IsPortManuallySet ?? false;
```
jviau: This will set it to `false` if someone else previously set to `true`. While I know we don't have anything that _would_ set it to true right now, a more future-proof approach would be:

(also, you can assume `.Value` is non-null.
```suggestion
            if (_httpWorkerOptions.Value.IsPortManuallySet)
            {
                // If statement to ensure we only set this to TRUE and never to FALSE.
                options.SequentialHostRestartRequired = true;
            }
```

this way the logic will only ever toggle sequential restart ON and never OFF

src/WebJobs.Script/Config/ScriptJobHostOptions.cs
```diff
@@ -145,5 +145,7 @@ public string RootScriptPath
         /// Gets or sets a value indicating the timeout duration for the function metadata provider.
         /// </summary>
         public TimeSpan MetadataProviderTimeout { get; set; } = TimeSpan.Zero;
+
+        public bool SequentialHostRestartRequired { get; set; }
```
jviau: Is this used? Can it be removed?

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptionsSetup.cs
jviau: Yeah, I think you are right @surgupta-msft, port `0` probably checks for first available port.

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptions.cs
jviau: I updated the sample to handle the `0` case. I see you updated to a different approach. I prefer to make the options object enforce the desired usage as much as possible. But ultimately up to you if you don't want to use the suggestion.

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -683,8 +683,15 @@ private static void NotifyHostStopping(IHost previousHost)
...
+        internal bool ShouldEnforceSequentialRestart(IHost host)
         {
+            var options = host?.Services?.GetService<IOptions<ScriptHostRecycleOptions>>();
+            var sequentialRestartRequired = options?.Value?.SequentialHostRestartRequired ?? false;
+            if (sequentialRestartRequired)
+            {
+                return true;
+            }
+
             var sequentialRestartSetting = _config.GetSection(ConfigurationSectionNames.SequentialJobHostRestart);
```
jviau: Could take this opportunity to move this config value parsing into `ScriptHostRecycleOptions` setup itself.

src/WebJobs.Script/Config/ScriptHostRecycleOptions.cs
```diff
@@ -0,0 +1,29 @@
...
+        /// </summary>
+        public bool SequentialHostRestartRequired { get; set; }
+
+        public string Format()
+        {
+            return JsonSerializer.Serialize(this, typeof(ScriptHostRecycleOptions), ScriptHostRecycleOptionsJsonContext.Default);
+        }
+    }
+
+    [JsonSourceGenerationOptions(WriteIndented = true)]
```
jviau: Consider using serialization optimized mode:
https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/source-generation#serialization-optimization-fast-path-mode-example


```suggestion
    [JsonSourceGenerationOptions(
        GenerationMode = JsonSourceGenerationMode.Serialization, WriteIndented = true)]
```

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptionsSetup.cs
```diff
@@ -130,7 +130,23 @@ private void ConfigureWorkerDescription(HttpWorkerOptions options, IConfiguratio
 
             options.Arguments.ExecutableArguments.AddRange(options.Description.Arguments);
             options.Arguments.WorkerArguments.AddRange(options.Description.WorkerArguments);
-            options.Port = WorkerUtilities.GetUnusedTcpPort();
+
+            if (options.Port is not null)
+            {
+                if (options.Port == 0 || !WorkerUtilities.CanBindToPort(options.Port.Value))
+                {
+                    throw new HostConfigurationException($"Unable to bind to port {options.Port} specified in configuration. Please specify a different port or remove the section to allow dynamic binding of port.");
```
jviau: See my other comment regarding handling setting of `0` for the port in the options object itself.

Also consider moving this to an options validator.

src/WebJobs.Script/WorkerUtilities.cs
```diff
@@ -17,5 +17,41 @@ public static int GetUnusedTcpPort()
...
             }
         }
+
+        /// <summary>
+        /// Determines whether the specified port is available.
+        /// </summary>
+        internal static bool CanBindToPort(int port)
+        {
+            // Try to bind to the port using IPv6 dual mode socket to cover both IPv4 and IPv6.
+            using (var tcpSocket = new Socket(AddressFamily.InterNetworkV6, SocketType.Stream, ProtocolType.Tcp) { DualMode = true })
```
jviau: We know for sure the port will be free immediately after disposal? Binding won't keep it locked for some duration?

I ask because sometimes that is the behavior with ports and connections at a system level, they are truly freed some period _after_ you release them.

test/WebJobs.Script.Tests/Configuration/HttpWorkerOptionsSetupTests.cs
```diff
@@ -140,6 +140,51 @@ public void InValid_HttpWorkerConfig_Throws_Exception(string hostJsonContent)
             }
         }
 
+        [Theory]
+        [InlineData(@"{
```
jviau: Consider using a member data. I like to use a dictionary + property to give the test a good name:

``` CSharp
private readonly Dictionary<string, string> MyTestCases = new()
{
    ["test_case_name1"] = "json",
    ["test_case_name2"] = "json",
}

public readonly IEnumerable<object[]> MyData => MyTestCases.Keys.Select(x => new object[] { x });

[Theory]
[MemberData(nameof(MyData))]
public void MyTest(string case)
{
    string json = MyTestCases[case];
}
```

I split it up like that because xunit has rules about how test cases get named in a test explorer. Only if the values of the member data are serializable do they show as individual cases. So you have the option of making everything an `IXUnitSerializable` in your data type, or use a dictionary for the indirection. And now you get to give each individual theory a human readable name.

test/WebJobs.Script.Tests/Configuration/HttpWorkerOptionsSetupTests.cs
```diff
@@ -140,6 +140,51 @@ public void InValid_HttpWorkerConfig_Throws_Exception(string hostJsonContent)
...
+                        'port': 1234567
+                        },
+                    }", 1234567)]
+        public void CustomHandlerConfig_InvalidPort_ExpandEnvVars(string hostJsonContent, int value)
+        {
+            HttpWorkerOptions options = new();
+
+            try
+            {
+                Environment.SetEnvironmentVariable("TestEnv", "TestVal");
```
jviau: Do not set environment variables in tests. This will potentially impact all other tests that run in parallel.

test/WebJobs.Script.Tests/Configuration/HttpWorkerOptionsSetupTests.cs
```diff
@@ -140,6 +140,51 @@ public void InValid_HttpWorkerConfig_Throws_Exception(string hostJsonContent)
...
+                File.WriteAllText(_hostJsonFile, hostJsonContent);
+                var configuration = BuildHostJsonConfiguration();
```
jviau: You can simplify this test by just building in-memory configuration. You shouldn't need to use explicitly host.json.

``` csharp
IConfiguration configuration = new ConfigurationBuilder()
    .AddInMemoryCollection(...)
    .Build();
```

give a dictionary to AddInMemoryCollection. Just make sure to use the config paths your options setup expect.

But I see all the other tests write a host.json. Seems fairly unnecessary to me - this options setup class is independent of `host.json` and our host json config source. Including this extra setup work goes against the spirit of a 'unit' test.

test/WebJobs.Script.Tests/Configuration/HttpWorkerOptionsSetupTests.cs
```diff
@@ -140,6 +140,51 @@ public void InValid_HttpWorkerConfig_Throws_Exception(string hostJsonContent)
...
+                HttpWorkerOptionsSetup setup = new(new OptionsWrapper<ScriptJobHostOptions>(_scriptJobHostOptions), configuration, _testLoggerFactory, _metricsLogger, _environment);
+                setup.Configure(options);
+                Assert.Equal("TestVal", options.Description.DefaultExecutablePath);
```
jviau: Consider using AwesomeAssertions


```suggestion
                options.Description.DefaultExecutablePath.Should().Be("TestVal");
```

test/WebJobs.Script.Tests/Configuration/HttpWorkerOptionsSetupTests.cs
```diff
@@ -140,6 +140,51 @@ public void InValid_HttpWorkerConfig_Throws_Exception(string hostJsonContent)
...
+                Assert.Contains("TestVal", options.Description.DefaultWorkerPath);
+            }
+            catch (HostConfigurationException ex)
+            {
+                Assert.True(ex.Message.Contains($"Unable to bind to port {value} specified in configuration. Please specify a different port or remove the section to allow dynamic binding of port."));
```
jviau: Consider using awesome assertions for this:

``` CSharp
Action act = () => setup.Configure(options);
act.Should().ThrowExactly<HostConfigurationException>().WithMessage(....);
```

test/WebJobs.Script.Tests/Configuration/HttpWorkerOptionsSetupTests.cs
```diff
@@ -248,6 +295,86 @@ public void HttpWorker_Config_ExpectedValues(string hostJsonContent, bool append
...
         }
 
+        [Theory]
+        [InlineData(@"{
+                    'version': '2.0',
+                    'customHandler': {
+                            'description': {
+                                'defaultExecutablePath': 'node',
+                                'arguments': ['httpWorker.js'],
+                                'workingDirectory': 'c:/myWorkingDir',
```
jviau, line 303: Any concern with trying to run this test on linux machines?

test/WebJobs.Script.Tests/Configuration/HttpWorkerOptionsSetupTests.cs
jviau: These assertions will actually never run as `.Configure` will throw.

Do we care about asserting these values if Configure ultimately throws? We give no guarantee on the state of the options object at that point.

test/WebJobs.Script.Tests/Configuration/HttpWorkerOptionsSetupTests.cs
jviau: Ah you can ignore this if you want. I see all the other test cases use inline data with raw json. Up to you if you want to migrate to member data.

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptions.cs
surgupta-msft: @jviau I slightly modified the get method for the scenario where no port is provided in the config as `configSection.Bind()` was assigning `IsPortManuallySet = true` for the dynamic binding case too. Also had to make it `bool?`. Let me know if there are any concerns.

src/WebJobs.Script/WorkerUtilities.cs
surgupta-msft: Tried implementing similar to the existing pattern used in the code - https://github.com/Azure/azure-functions-host/blob/04f0b2b2d2786e9e5aaa9287cb7d52dfcbec16a5/src/WebJobs.Script/WorkerUtilities.cs#L13.

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
surgupta-msft: Updated the code. 

The reason I cannot directly move the config check logic to `ScriptHostRecycleOptions` is - what if the method [ShouldEnforceSequentialRestart()](https://github.com/Azure/azure-functions-host/blob/04f0b2b2d2786e9e5aaa9287cb7d52dfcbec16a5/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs#L686) is called in a scenario where `ScriptHostRecycleOptions` service is not available. Today, we will directly check IConfiguration if options is null. But if I move this logic inside Options then we will not get the value from config too. This will change the behavior for this scenario. One example scenario is when JobHost is not yet up and can't provide the service. 

Hence I used this approach after briefly chatting with Fabio. Added a new property in `ScriptApplicationHostOptions` which represents Application level setting, which `ScriptHostRecycleOptions` is job host level.

src/WebJobs.Script/WorkerUtilities.cs
jviau: I don't have any immediate code-change ask. Just checking to see if a back-to-back host restart was tested and works or not.

I have dealt with bugs where a port remains locked after restart before (was a webjobs extension). Wasn't sure the exact cause, so I had some concern same thing may happen here.

src/WebJobs.Script/Config/ScriptApplicationHostOptions.cs
```diff
@@ -35,5 +35,7 @@ public class ScriptApplicationHostOptions
         public bool IsFileSystemReadOnly { get; set; }
 
         public bool IsScmRunFromPackage { get; set; }
+
+        public bool SequentialHostRestartRequired { get; set; }
```
jviau: It is a bit confusing to have this same value in two different options object, with different logic between them. Can we consolidate to only one? Either the new options object or here, whichever you prefer. Have the setup check all various sources for enabled sequential restart.

src/WebJobs.Script/Config/ScriptHostRecycleOptionsSetup.cs
```diff
@@ -0,0 +1,28 @@
...
+
+        public ScriptHostRecycleOptionsSetup(IOptions<HttpWorkerOptions> httpWorkerOptions)
+        {
+            _httpWorkerOptions = httpWorkerOptions ?? throw new ArgumentNullException(nameof(httpWorkerOptions));
+        }
+
+        public void Configure(ScriptHostRecycleOptions options)
+        {
+            // Enforcing sequential host restarts when a user-specified custom handler port is configured to prevent multiple processes from attempting to bind to the same port concurrently.
+            if (_httpWorkerOptions.Value.IsPortManuallySet.Value)
```
jviau: `.Value` throws if the nullable does not have a value.

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptions.cs
```diff
@@ -10,13 +10,39 @@ namespace Microsoft.Azure.WebJobs.Script.Workers.Http
...
+                IsPortManuallySet ??= true;
+                _port = value;
+            }
+        }
+
+        /// <summary>
+        /// Gets a value indicating whether the <see cref="Port"/> property value is taken from configuration.
+        /// True value indicates that the host will use the configured port value rather than allocating a dynamic port.
+        /// </summary>
+        public bool? IsPortManuallySet { get; private set; }
```
jviau: I see you made this a nullable. Can you explain why? What I see is a potential bug: if `.Port` is set _after_ `_port = WorkerUtilities.GetUnusedTcpPort();` runs we will not set `IsPortManuallySet` to `true` (when we should)

src/WebJobs.Script/WorkerUtilities.cs
jviau: Use the new using style:

```suggestion
            using Socket tcpSocket = new(AddressFamily.InterNetworkV6, SocketType.Stream, ProtocolType.Tcp) { DualMode = true };
```

src/WebJobs.Script/WorkerUtilities.cs
```diff
@@ -17,5 +17,41 @@ public static int GetUnusedTcpPort()
...
+                    return true;
+                }
+                catch (SocketException se) when (se.SocketErrorCode == SocketError.AddressAlreadyInUse)
+                {
+                    return false;
+                }
+                catch
+                {
+                    // Fall back to IPv4 only socket if IPv6 is not supported on the platform.
+                    using (var tcpSocketAny = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp))
```
jviau: Use the new using style:


```suggestion
                    using Socket tcpSocketAny = new(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp);
```

test/WebJobs.Script.Tests/Configuration/ScriptApplicationHostOptionsSetupTests.cs
```diff
@@ -13,6 +14,32 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.Configuration
...
+        [InlineData("True", true)]
+        [InlineData("false", false)]
+        public void SequentialHostRestartRequired_Configured(string value, bool expectedResult)
+        {
+            ScriptApplicationHostOptions options = new ScriptApplicationHostOptions();
+
+            IConfiguration config = new ConfigurationBuilder()
+                .AddInMemoryCollection(new Dictionary<string, string>
+                {
+                { "AzureFunctionsJobHost:SequentialRestart",  value },
```
jviau:
```suggestion
                    { "AzureFunctionsJobHost:SequentialRestart",  value },
```

Also tiniest nit, I like the newer dictionary initializer

``` csharp
{
    ["AzureFunctionsJobHost:SequentialRestart"] = value,
}
```

test/WebJobs.Script.Tests/Configuration/HttpWorkerOptionsSetupTests.cs
```diff
@@ -140,6 +142,31 @@ public void InValid_HttpWorkerConfig_Throws_Exception(string hostJsonContent)
...
+                        new("AzureFunctionsJobHost:customHandler:description:defaultExecutablePath", "%TestEnv%"),
+                        new("AzureFunctionsJobHost:customHandler:port", value.ToString())
+                        }).Build();
+
+            HttpWorkerOptionsSetup setup = new(new OptionsWrapper<ScriptJobHostOptions>(_scriptJobHostOptions), configuration, _testLoggerFactory, _metricsLogger, _environment);
+
+            Action act = () =>
+                    {
+                        setup.Configure(options);
+                        setup.Validate(nameof(options), options);
```
jviau: Small nit: you are testing two different methods here, I personally would break this up into two different tests.

One which validates only `Configure`, and asserts the values are parsed as expected.

One which validates `Validate` (and you just manually new up the options object with the values you want to verify, ie `options = new() { Port = 0 };`

src/WebJobs.Script/Config/ScriptApplicationHostOptions.cs
surgupta-msft: I added a comment here - https://github.com/Azure/azure-functions-host/pull/11408#discussion_r2415516399. Repasting and resolving old conversation to continue discussion here - 

The reason I cannot directly move the config check logic to `ScriptHostRecycleOptions` is - what if the method [ShouldEnforceSequentialRestart()](https://github.com/Azure/azure-functions-host/blob/04f0b2b2d2786e9e5aaa9287cb7d52dfcbec16a5/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs#L686) is called in a scenario where `ScriptHostRecycleOptions` service is not available. Today, we will directly check IConfiguration if options is null. But if I move this logic inside Options then we will not get the value from config too. This will change the behavior for this scenario. One example scenario is when JobHost is not yet up and can't provide the service. 

Hence I used this approach after briefly chatting with Fabio. Added a new property in `ScriptApplicationHostOptions` which represents Application level setting, which `ScriptHostRecycleOptions` is job host level.

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptions.cs
```diff
@@ -33,7 +33,12 @@ public int Port
...
             set
             {
-                IsPortManuallySet ??= true;
+                // During dynamic allocation of port, the get method will be called before set method and _port will be assigned dynamically.
+                // Adding a check here to make sure we don't override IsPortManuallySet flag in that case.
+                if (_port != value)
+                {
+                    IsPortManuallySet ??= true;
+                }
                 _port = value;
```
jviau: smallest of nits: this can go into the `if` block as well.

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptions.cs
```diff
@@ -33,7 +33,12 @@ public int Port
 
             set
             {
-                IsPortManuallySet ??= true;
+                // During dynamic allocation of port, the get method will be called before set method and _port will be assigned dynamically.
+                // Adding a check here to make sure we don't override IsPortManuallySet flag in that case.
+                if (_port != value)
+                {
+                    IsPortManuallySet ??= true;
```
jviau:
```suggestion
                    IsPortManuallySet = true;
```

src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptions.cs
jviau: We shouldn't need nullable anymore with the recent change.

src/WebJobs.Script/WorkerUtilities.cs
```diff
@@ -24,7 +24,7 @@ public static int GetUnusedTcpPort()
         internal static bool CanBindToPort(int port)
         {
             // Try to bind to the port using IPv6 dual mode socket to cover both IPv4 and IPv6.
-            using (var tcpSocket = new Socket(AddressFamily.InterNetworkV6, SocketType.Stream, ProtocolType.Tcp) { DualMode = true })
+            using var tcpSocket = new Socket(AddressFamily.InterNetworkV6, SocketType.Stream, ProtocolType.Tcp) { DualMode = true };
             {
```
jviau, line 29: with this using style you can remove the curly brace and move everything up an indent level

``` csharp
using var myDisposable = ....;
myDisposable.Something(); // this is inline. myDisposable is disposed when it loses scope.
```

jviau (commented):
Can you add tests for `ScriptHostRecycleOptions` and its setup.

test/WebJobs.Script.Tests/Configuration/ScriptHostRecycleOptionsSetupTests.cs
```diff
@@ -0,0 +1,47 @@
...
+    {
+        [Theory]
+        [InlineData("true", true, true)]
+        [InlineData("false", false, false)]
+        [InlineData("false", true, true)]
+        [InlineData("true", false, true)]
+        public void Configure_SequentialHostRestartRequired_Works(string configValue, bool isPortManuallySet, bool output)
+        {
+            var httpWorkerOptions = new HttpWorkerOptions();
+            typeof(HttpWorkerOptions)
```
jviau, line 25: You could set this indirectly by supplying port as a nullable to the test:

``` CSharp
public void Configure_SequentialHostRestartRequired_Works(string configValue, int? port, bool output)
{
    if (port is int p)
    {
        httpWorkerOptions.Port = p;
    }

    /// rest
}
```

test/WebJobs.Script.Tests/Configuration/ScriptHostRecycleOptionsSetupTests.cs
```diff
@@ -0,0 +1,47 @@
...
+                .GetProperty(nameof(HttpWorkerOptions.IsPortManuallySet))!
+                .SetValue(httpWorkerOptions, isPortManuallySet);
+
+            var mockOptions = new Mock<IOptions<HttpWorkerOptions>>();
```
jviau, line 29: There is a simple helper for creating `IOptions<T>`
```suggestion
            var options = Options.Create(httpWorkerOptions);
```

jviau (approved):

soninaren (approved):

PR #11410, opened 2025-10-08
Include HealthChecks from ScriptHost scope

### LINKED ISSUE #11403: [HealthChecks] Integration `IHealthCheck` implementations from active ScriptHost into health checks.
The current health check design in the host only resolves `IHealthCheck` implementations from the WebHost layer. This precludes extensions from registering their own health checks as they all live in the ScriptHost layer.

## Requirements

- Health checks registered to script host service using _existing_ `Microsoft.Extensions.HealthCheck` APIs will be included in the WebHost's `HealthCheckService`.
- Health checks from the script host layer are correctly resolved/invoked with a service scope from the ScriptHost.

## Implementation

We will introduce our own implementation of `HealthCheckService` which will aggregate health checks from the WebHost and then from the ScriptHost. This will work by internally invoking `CheckHealthAsync` on WebHost and ScriptHost layers separately, then combining the results for a final health check.

Considerations:

**Conflict Resolution**: since we will have two discrete health check configurations, we will have the possibility of conflicting health check names. In the case of a conflict, we will log a warning and take the check from the **WebHost**. We take the check from the WebHost as this will prevent extensions from stepping over host-specific health checks. Conflicts _should_ be rare as we will ensure to keep all of the hosts health checks in the `azure.functions.` namespace (and extensions should ideally be in `azure.webjobs.` namespace, if any namespace at all).

jviau:

resolves #11403

### Pull request checklist

### Additional information

This PR integrates `IHealthCheck` implementations from the ScriptHost scope. This is done by inserting a new `DynamicHealthCheckService`, which is effectively a wrapper around 2 health check services. (1) The existing WebJobs scoped health check service, and (2) the current active ScriptHost health check service (if one exists).

This approach is taken to ensure all health checks from the ScriptHost are ran in the correct service scope. An initial attempt was made to pull health check registrations from the active script host into the WebHost directly, but that will run them in the incorrect context and lead to wrong DI setup for them.


src/WebJobs.Script/Diagnostics/HealthChecks/DynamicHealthCheckService.cs
```diff
@@ -0,0 +1,103 @@
...
+        public override async Task<HealthReport> CheckHealthAsync(
+            Func<HealthCheckRegistration, bool> predicate,
+            CancellationToken cancellationToken = default)
+        {
+            Task<HealthReport> webHostReport = _webHostCheck.CheckHealthAsync(predicate, cancellationToken);
+            Task<HealthReport> scriptHostReport = CheckScriptHostHealthAsync(predicate, cancellationToken);
+            return MergeReports(await webHostReport, await scriptHostReport);
+        }
+
+        private HealthReport MergeReports(HealthReport left, HealthReport right)
```
RohitRanjanMS: Can both left and right be null?

RohitRanjanMS (approved):

src/WebJobs.Script/Diagnostics/HealthChecks/DynamicHealthCheckService.cs
jviau: By contract no. We guarantee the WebJobs `HealthCheckService` is non-null in the constructor. And that contract is always a non-null result. I think I will remove the null check on `left` as that should never happen.

src/WebJobs.Script/Diagnostics/HealthChecks/DynamicHealthCheckService.cs
```diff
@@ -0,0 +1,103 @@
...
+                return left;
+            }
+
+            // Merge the entries of both reports. If there are any duplicate keys, keep the left one.
+            Dictionary<string, HealthReportEntry> entries = new(left.Entries);
+            foreach ((string key, HealthReportEntry value) in right.Entries)
+            {
+                if (!entries.TryAdd(key, value))
+                {
+                    Log.DuplicateHealthCheckEntry(_logger, key);
```
surgupta-msft: Wondering, why can't we simply ignore duplicate entry instead of logging? Will this provide any additional insight?

I am also thinking what if there are frequent duplicate entries and we will log this statement multiple times.

release_notes.md
```diff
@@ -7,3 +7,6 @@
 - Add JitTrace Files for v4.1044
 - Remove duplicate function names from sync triggers payload(#11371)
 - Avoid emitting empty tag values for health check metrics (#11393)
+- Run health checks from the active ScriptHost (#11410)
+- Publish health check metrics to legacy AppInsights SDK (#11365)
+- Fix tag filter for health check live & ready endpoints (#11363)
```
surgupta-msft, line 13: Looks like some merge issue with release notes?

jviau: Intentional: I am adding release notes for previous PRs that I forgot to

src/WebJobs.Script/Diagnostics/HealthChecks/DynamicHealthCheckService.cs
jviau: We want to log to let the author/customer know their script-host entry is being ignored.

I can update this to log only once per duplicate

jviau: I opted to log once per duplicate _per hour_

surgupta-msft (approved):

surgupta-msft (approved):

PR #11471, opened 2025-11-03
[HealthChecks] Add AzureWebJobsStorage health check

### LINKED ISSUE #11461: [HealthChecks] Add health check for AzureWebJobsStorage connection
We will add a health check to verify we can connect to `AzureWebJobsStorage`. The check will be part of the `azure.functions.configuration` tag and **not** be part of live or ready probe.

jviau:

resolves #11461

### Pull request checklist

### Additional information

This PR adds a check for `AzureWebJobsStorage`, which is needed for basic app functionality (the functioning connection). The implementation takes a background-refresh approach. To avoid high latency health check calls, the health will be checked in the background, with the latest value being stored and immediately returned to `CheckHealthAsync` callers. An exception is the first call, which will block until a health result is available.

This PR also introduces a `HealthCheckData` type, which is a syntactic sugar wrapper around `IReadOnlyDictionary<string, object>` (which is what the health checks APIs use). The motivation is to provide a strongly-typed way to enforce a consistent schema for health check data. Having our checks follow a known schema will help automation consume our checks. This wrapper lets us do that via CSharp properties, which can enforce both key and value in one idiomatic way. This is much like `HeaderDictionary`, which wraps a dictionary but exposes many well-known headers as easy to use properties (that are just part of the dictionary under the hood).


src/WebJobs.Script/Diagnostics/HealthChecks/HealthCheckData.cs
jviau, line 1: A good area of discussion for this PR is the names of these properties I have here.

1. What style should they have? Currently it is pascal case (because it gets the name of the property)
2. Any names that should be changed?

src/WebJobs.Script/Diagnostics/HealthChecks/HealthCheckTags.cs
```diff
@@ -8,30 +8,36 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics.HealthChecks
     /// </summary>
     internal static class HealthCheckTags
     {
-        private const string Prefix = "azure.functions.";
+        private const string FuncPrefix = "azure.functions.";
+        private const string WebJobsPrefix = "azure.web_jobs.";
```
liliankasem: nit: I think I prefer webjobs as one, `azure.web_jobs.storage` feels off

```suggestion
        private const string WebJobsPrefix = "azure.webjobs.";
```

src/WebJobs.Script/Diagnostics/HealthChecks/HealthCheckExtensions.cs
```diff
@@ -126,6 +128,24 @@ public static IHealthChecksBuilder AddScriptHostHealthCheck(this IHealthChecksBu
...
+        /// <returns>The original builder, for call chaining.</returns>
+        public static IHealthChecksBuilder AddWebJobsStorageHealthCheck(this IHealthChecksBuilder builder)
+        {
+            ArgumentNullException.ThrowIfNull(builder);
+
+            // Ensure singleton as this health check refreshes in the background.
+            builder.Services.TryAddSingleton<WebJobsStorageHealthCheck>();
+            builder.AddCheck<WebJobsStorageHealthCheck>(
+                HealthCheckNames.WebJobsStorage,
+                tags: [HealthCheckTags.Configuration, HealthCheckTags.WebJobsStorage],
```
liliankasem: Isn't this more connectivity than config?

src/WebJobs.Script/Diagnostics/HealthChecks/HealthCheckTags.cs
jviau: I have the same underscore pattern for `script_host` and `web_host`. In this naming convention, different words are either namespace separated _or_ underscore separated. I don't want to violate that and have a single-word like `webjobs`.

src/WebJobs.Script/Diagnostics/HealthChecks/HealthCheckExtensions.cs
jviau: Good point. Why not both?

src/WebJobs.Script/Diagnostics/HealthChecks/HealthCheckTags.cs
liliankasem: If we're already doing this with the other "hosts", then I would prefer the consistancy here as well

src/WebJobs.Script/Diagnostics/HealthChecks/HealthCheckData.cs
liliankasem: I think this is fine as is. The only one where I'm like I wish there was a word more descriptive is "Area" but I also can't think of anything

liliankasem (approved):

src/WebJobs.Script/Diagnostics/HealthChecks/HealthCheckData.cs
jviau: I bounced around on that naming a few times. I had `Source` and `Cause` before. Maybe `Category`? or `ErrorCategory`?

src/WebJobs.Script/Diagnostics/HealthChecks/HealthCheckTags.cs
cjaliaga: Totally agree on keeping things consistent. That said, *WebJobs* feels more like a proper name to me, kind of like *Functions*, rather than something we'd split with an underscore.

Using `webjobs` without the underscore keeps that product identity and matches how it’s usually referenced across Azure.

That said, happy to follow the consistency rule if that’s the preferred convention.

src/WebJobs.Script/Diagnostics/HealthChecks/HealthCheckTags.cs
```diff
@@ -8,30 +8,44 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics.HealthChecks
     /// </summary>
     internal static class HealthCheckTags
```
RohitRanjanMS: We can introduce a hierarchical structure to health check tags like HealthCheckTags.Area.Connectivity and HealthCheckTags.ConfigurationSection.Storage

src/WebJobs.Script/Diagnostics/HealthChecks/HealthCheckTags.cs
mattchenderson: Because "WebJobs" is the brand term, I believe `azure.webjobs.` is the more correct approach here.

jviau: Switched to `webjobs`.

cjaliaga (approved):

PR #11486, opened 2025-11-19
Do Not Modify Function Language for dotnet-isolated worker runtime in workflowApp

pravagar:
Logic App runs as a multi-language Function App. Today, Logic Apps use different values for the FUNCTIONS_WORKER_RUNTIME app setting, though most are configured with DotNet. Some existing Logic Apps already use dotnet-isolated, and Flex Consumption Logic Apps are also planned to adopt dotnet-isolated.

When FUNCTIONS_WORKER_RUNTIME=dotnet-isolated and none of the indexed Logic App functions declare their language as dotnet-isolated, the current behavior rewrites functions with language DotNetAssembly to dotnet-isolated. However, functions marked as DotNetAssembly are meant to be processed by DotNetFunctionDescriptorProvider.
Once functions with the language  DotNetAssembly is changed to dotnet-isolated, the system incorrectly uses MultiLanguageFunctionDescriptorProvider instead of DotNetFunctionDescriptorProvider to create bindings, resulting in binding failures. 
Here is the error message:

Error indexing method 'Functions.WorkflowDispatcher'
      Microsoft.Azure.WebJobs.Host.Indexers.FunctionIndexingException: Error indexing method 'Functions.WorkflowDispatcher'
       ---> System.InvalidOperationException: Can't bind EdgeWorkflowRuntimeTrigger to type 'System.Object'.
         at Microsoft.Azure.WebJobs.Host.Bindings.GenericCompositeBindingProvider`1.TryCreateAsync(BindingProviderContext context) in c:\git1\azure-webjobs-sdk\src\Microsoft.Azure.WebJobs.Host\Bindings\BindingProviders\GenericCompositeBindingProvider.cs:line 89
         at Microsoft.Azure.WebJobs.Host.Bindings.CompositeBindingProvider.TryCreateAsync(BindingProviderContext context) in c:\git1\azure-webjobs-sdk\src\Microsoft.Azure.WebJobs.Host\Bindings\BindingProviders\CompositeBindingProvider.cs:line 25
         at Microsoft.Azure.WebJobs.Host.Indexers.FunctionIndexer.IndexMethodAsyncCore(MethodInfo method, IFunctionIndexCollector index, CancellationToken cancellationToken) in c:\git1\azure-webjobs-sdk\src\Microsoft.Azure.WebJobs.Host\Indexers\FunctionIndexer.cs:line 198
         at Microsoft.Azure.WebJobs.Host.Indexers.FunctionIndexer.IndexMethodAsync(MethodInfo method, IFunctionIndexCollector index, CancellationToken cancellationToken) in c:\git1\azure-webjobs-sdk\src\Microsoft.Azure.WebJobs.Host\Indexers\FunctionIndexer.cs:line 149
         --- End of inner exception stack trace ---
         at Microsoft.Azure.WebJobs.Host.Indexers.FunctionIndexer.IndexMethodAsync(MethodInfo method, IFunctionIndexCollector index, CancellationToken cancellationToken) in c:\git1\azure-webjobs-sdk\src\Microsoft.Azure.WebJobs.Host\Indexers\FunctionIndexer.cs:line 157
         at Microsoft.Azure.WebJobs.Host.Indexers.FunctionIndexer.IndexTypeAsync(Type type, IFunctionIndexCollector index, CancellationToken cancellationToken) in c:\git1\azure-webjobs-sdk\src\Microsoft.Azure.WebJobs.Host\Indexers\FunctionIndexer.cs:line 85

To avoid these failures, the language property of Logic App–indexed functions must not be modified during processing.

### DIFF
--- a/src/WebJobs.Script/Host/ScriptHost.cs
+++ b/src/WebJobs.Script/Host/ScriptHost.cs
@@ -799,7 +799,7 @@ namespace Microsoft.Azure.WebJobs.Script
                     bool payloadMatchesWorkerRuntime = ValidateAndLogRuntimeMismatch(functions, workerRuntime, _hostingConfigOptions, _logger);
                     if (!payloadMatchesWorkerRuntime)
                     {
-                        UpdateFunctionMetadataLanguageForDotnetAssembly(functions, workerRuntime);
+                        UpdateFunctionMetadataLanguageForDotnetAssembly(functions, workerRuntime, _environment.IsLogicApp());
                         throwOnWorkerRuntimeAndPayloadMetadataMismatch = false; // we do not want to throw an exception in this case
                     }
                 }
@@ -860,8 +860,14 @@ namespace Microsoft.Azure.WebJobs.Script
             return functionDescriptors;
         }
 
-        private static void UpdateFunctionMetadataLanguageForDotnetAssembly(IEnumerable<FunctionMetadata> functions, string workerRuntime)
+        private static void UpdateFunctionMetadataLanguageForDotnetAssembly(IEnumerable<FunctionMetadata> functions, string workerRuntime, bool isLogicApp)
         {
+            if (isLogicApp)
+            {
+                // For Logic Apps, we want to keep the language as DotNetAssembly to avoid breaking changes.
+                return;
+            }
+
             foreach (var function in functions)
             {
                 if (function.Language == DotNetScriptTypes.DotNetAssembly)
--- a/src/WebJobs.Script/Host/ScriptHost.cs
+++ b/src/WebJobs.Script/Host/ScriptHost.cs
@@ -791,15 +791,15 @@ namespace Microsoft.Azure.WebJobs.Script
             Collection<FunctionDescriptor> functionDescriptors = new Collection<FunctionDescriptor>();
             if (!cancellationToken.IsCancellationRequested)
             {
-                bool throwOnWorkerRuntimeAndPayloadMetadataMismatch = !(_environment.IsLogicApp() && _environment.IsLogicAppCodefulModeEnabled());
+                bool throwOnWorkerRuntimeAndPayloadMetadataMismatch = !_environment.IsLogicApp();
 
                 // this dotnet isolated specific logic is temporary to ensure in-proc payload compatibility with "dotnet-isolated" as the FUNCTIONS_WORKER_RUNTIME value.
-                if (string.Equals(workerRuntime, RpcWorkerConstants.DotNetIsolatedLanguageWorkerName, StringComparison.OrdinalIgnoreCase) && !_environment.IsPlaceholderModeEnabled())
+                if (!_environment.IsLogicApp() && string.Equals(workerRuntime, RpcWorkerConstants.DotNetIsolatedLanguageWorkerName, StringComparison.OrdinalIgnoreCase) && !_environment.IsPlaceholderModeEnabled())
                 {
                     bool payloadMatchesWorkerRuntime = ValidateAndLogRuntimeMismatch(functions, workerRuntime, _hostingConfigOptions, _logger);
                     if (!payloadMatchesWorkerRuntime)
                     {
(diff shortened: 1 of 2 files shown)


src/WebJobs.Script/Host/ScriptHost.cs
```diff
@@ -791,10 +791,10 @@ internal async Task<Collection<FunctionDescriptor>> GetFunctionDescriptorsAsync(
             Collection<FunctionDescriptor> functionDescriptors = new Collection<FunctionDescriptor>();
             if (!cancellationToken.IsCancellationRequested)
             {
-                bool throwOnWorkerRuntimeAndPayloadMetadataMismatch = !(_environment.IsLogicApp() && _environment.IsLogicAppCodefulModeEnabled());
+                bool throwOnWorkerRuntimeAndPayloadMetadataMismatch = !_environment.IsLogicApp();
```
apranaseth: Why are we removing the codeful mode enabled check from here?

mathewc, line 794: So here, we used to NOT throw for codeful logic apps. When you release this change we'll now start throwing. What does that mean for all the currently deployed codeful logic apps that aren't seeing exceptions currently because of this. Won't this be breaking?

src/WebJobs.Script/Host/ScriptHost.cs
```diff
@@ -791,10 +791,10 @@ internal async Task<Collection<FunctionDescriptor>> GetFunctionDescriptorsAsync(
...
 
                 // this dotnet isolated specific logic is temporary to ensure in-proc payload compatibility with "dotnet-isolated" as the FUNCTIONS_WORKER_RUNTIME value.
-                if (string.Equals(workerRuntime, RpcWorkerConstants.DotNetIsolatedLanguageWorkerName, StringComparison.OrdinalIgnoreCase) && !_environment.IsPlaceholderModeEnabled())
+                if (!_environment.IsLogicApp() && string.Equals(workerRuntime, RpcWorkerConstants.DotNetIsolatedLanguageWorkerName, StringComparison.OrdinalIgnoreCase) && !_environment.IsPlaceholderModeEnabled())
```
mathewc, line 797: Same question here - I'm wondering about release timing of this and whether it breaks anyone. Maybe I don't have enough context on this, and not a lot of context was given in the PR description.

src/WebJobs.Script/Host/ScriptHost.cs
pravagar: Codeful Logic Apps are not yet deployed or available to customers, so this change will not impact any existing Logic Apps workloads. Previously, the warning/error was suppressed only for Codeful Logic Apps; with this update, the warning/error will be suppressed for all Logic Apps, which is the intended behavior.

Currently Logic Apps FUNCTIONS_WORKER_RUNTIME app setting is set to different languages, though most of the logic apps has DotNet as its value. For Logic Apps, runtime and payload metadata mismatch warnings should not be provided.

src/WebJobs.Script/Host/ScriptHost.cs
pravagar: Currently Logic Apps used FunctionsInProc or FunctionsInProc8 site extension and not using 'Functions' site extension which is the latest function host. The plan is to move Logic Apps from using FunctionsInProc or FunctionsInProc8 to Functions site extension. This is prerequisite for Logic App's .NET 10 migration.

This change will not break any existing Logic App customer as any Logic App is not using latest function host.  Without this change, existing Logic Apps with dotnet-isolated as Worker runtime, fails during function host loading.

There are also changes to Antares-Websites repo to migrate Logic Apps to use latest function host. The release timeline is ASAP, though it will take time to gradually migrate all Logic Apps to use latest function host.

mathewc (approved):

fabiocav (approved):

rarayudu (approved):

PR #11506, opened 2025-12-08
Adding support for OTEL_SERVICE_NAME env variable

### LINKED ISSUE #11521: OTEL_SERVICE_NAME env variable is not applied to the telemetry
When setting the OTEL_SERVICE_NAME environment variable, the expected service name does not appear in the exported telemetry. Instead, the default or fallback name is used. This makes it difficult to correctly identify services in observability platforms.

Steps to Reproduce:
- Set OTEL_SERVICE_NAME=MyService in the environment.
- Run the application with OpenTelemetry enabled.
- Inspect the telemetry data (e.g., traces, metrics).
- Observe that the service name is not MyService.

RohitRanjanMS:

resolves #11521

### Pull request checklist

### Additional information

Additional PR information


src/WebJobs.Script/Diagnostics/OpenTelemetry/FunctionsResourceDetector.cs
```diff
@@ -10,22 +10,31 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics.OpenTelemetry
...
             List<KeyValuePair<string, object>> attributeList = new(9);
             try
             {
-                string serviceName = Environment.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebsiteName);
-                string version = typeof(ScriptHost).Assembly.GetName().Version.ToString();
+                // Determine service name with override hierarchy: OTEL_SERVICE_NAME > AzureWebsiteName > AssemblyName
+                string serviceName = GetServiceName();
+
+                // Determine service version with override hierarchy: OTEL_SERVICE_VERSION > AssemblyVersion
+                string serviceVersion = GetServiceVersion();
```
jviau: We have traditionally always wanted the functions host version for this telemetry. Are we sure we want to let this be customer configurable?

RohitRanjanMS: We enable this only when OTel mode is active; the host version is still the default value, but some customers prefer to set a custom value. The host version is appended to sdkVersion, ensuring it’s present in all telemetry.

ghost: Just wanted to start with that we are excited for these changes for OTEL.

So I have a question, will this also work if the `OTEL_RESOURCE_ATTRIBUTES` sets the `service.version` - I think this is the standard way to set the 'service version' based on the specification [here](https://opentelemetry.io/docs/languages/dotnet/resources/#resource-attributes)?

I tried setting `service.version` on the `OTEL_RESOURCE_ATTRIBUTES` env var in the current implementation and seems that its always be overridden by the function host version.

BTW I could not find any mention of an env var `OTEL_SERVICE_VERSION`, and based on this thread looks like it didn't make it into the specification: https://github.com/open-telemetry/opentelemetry-specification/issues/1901

RohitRanjanMS: Thank you for pointing this out, @pb-adrianbezzina . I've updated the code so that if the service name or version is already defined—either in OTEL_RESOURCE_ATTRIBUTES or the corresponding environment variable—it will be preserved.

src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
```diff
@@ -35,6 +35,7 @@ public static class EnvironmentSettingNames
         public const string AppInsightsAgent = "APPLICATIONINSIGHTS_ENABLE_AGENT";
         public const string FunctionsExtensionVersion = "FUNCTIONS_EXTENSION_VERSION";
         public const string FunctionWorkerRuntime = "FUNCTIONS_WORKER_RUNTIME";
+        public const string FunctionAppVersion = "FUNCTION_APP_VERSION";
```
kshyju: Is this intended to be user configurable so they can version their deployment? If so, should the app setting name align with how it appears in telemetry? Since this value is applied to `siteUpdateId` in telemetry, should we consider using a similar name here?

src/WebJobs.Script/Diagnostics/OpenTelemetry/FunctionsResourceDetector.cs
```diff
@@ -10,66 +12,120 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics.OpenTelemetry
...
 
-                attributeList.Add(new KeyValuePair<string, object>(ResourceSemanticConventions.ServiceVersion, version));
-                attributeList.Add(new KeyValuePair<string, object>(ResourceSemanticConventions.AISDKPrefix, $@"{OpenTelemetryConstants.SDKPrefix}:{version}"));
-                attributeList.Add(new KeyValuePair<string, object>(ResourceSemanticConventions.ProcessId, Process.GetCurrentProcess().Id));
+                // Priority: OTEL_SERVICE_NAME > OTEL_RESOURCE_ATTRIBUTES[service.name] > AzureWebsiteName > AssemblyName
+                // Only add service.name if not already configured
+                if (!IsServiceAttributeConfigured(ResourceSemConventions.ServiceName,
+                                               ResourceSemConventions.ServiceNameEnvVar))
+                {
+                    attributes.Add(new(ResourceSemConventions.ServiceName, azureWebsiteName ?? typeof(ScriptHost).Assembly.GetName().Name ?? "unknown"));
```
kshyju: Do we expect this method to be executed more than once? If so, we should use a shared cached version instead of executing `Assembly.GetName()` every time. 
```
internal static readonly string ScriptHostAssemblyName = typeof(ScriptHost).Assembly.GetName().Name!;
```

src/WebJobs.Script/Diagnostics/OpenTelemetry/ResourceSemConventions.cs
```diff
@@ -3,12 +3,15 @@
 
 namespace Microsoft.Azure.WebJobs.Script.Diagnostics.OpenTelemetry
 {
-    internal static class ResourceSemanticConventions
```
kshyju, line 6: Curious - why the rename here?

src/WebJobs.Script/Diagnostics/OpenTelemetry/FunctionsResourceDetector.cs
```diff
@@ -10,66 +12,120 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics.OpenTelemetry
...
+                };
+
+                string? azureWebsiteName = Environment.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebsiteName);
 
-                attributeList.Add(new KeyValuePair<string, object>(ResourceSemanticConventions.ServiceVersion, version));
-                attributeList.Add(new KeyValuePair<string, object>(ResourceSemanticConventions.AISDKPrefix, $@"{OpenTelemetryConstants.SDKPrefix}:{version}"));
-                attributeList.Add(new KeyValuePair<string, object>(ResourceSemanticConventions.ProcessId, Process.GetCurrentProcess().Id));
+                // Priority: OTEL_SERVICE_NAME > OTEL_RESOURCE_ATTRIBUTES[service.name] > AzureWebsiteName > AssemblyName
+                // Only add service.name if not already configured
+                if (!IsServiceAttributeConfigured(ResourceSemConventions.ServiceName,
```
kshyju: looks like `Environment.GetEnvironmentVariable(ResourceSemConventions.ResourceAttributeEnvVar)` expression is executed twice (in line 32 and 40). Should we consider moving that out of the `IsServiceAttributeConfigured` method so that it can be executed only once here?

src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
RohitRanjanMS: This is not user configurable. There’s an ongoing conversation about this in another PR -  [HostAssignmentContext.cs](https://github.com/Azure/azure-functions-host/pull/11527/files#diff-e7955cf28637e880e4ec5fd4fc24d709061c97f01a318f5012eaa599dca11150).

src/WebJobs.Script/Diagnostics/OpenTelemetry/FunctionsResourceDetector.cs
RohitRanjanMS: These are evaluated once to build the attributes list and then added to all the telemetry.

src/WebJobs.Script/Diagnostics/OpenTelemetry/ResourceSemConventions.cs
RohitRanjanMS: Nothing specific. I was just annoyed by the long name. Made it a bit short.

src/WebJobs.Script/Diagnostics/OpenTelemetry/FunctionsResourceDetector.cs
RohitRanjanMS: Removed check for OTEL_SERVICE_VERSION, now there's only one call to get `ResourceAttributeEnvVar`.

jviau (approved):

src/WebJobs.Script/Diagnostics/OpenTelemetry/FunctionsResourceDetector.cs
```diff
@@ -10,66 +12,131 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics.OpenTelemetry
...
         public Resource Detect()
         {
-            List<KeyValuePair<string, object>> attributeList = new(9);
             try
             {
-                string serviceName = Environment.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebsiteName);
-                string version = typeof(ScriptHost).Assembly.GetName().Version.ToString();
+                var attributes = new List<KeyValuePair<string, object>>(capacity: 10)
+                {
+                    new(ResourceSemanticConventions.AISDKPrefix, $"{OpenTelemetryConstants.SDKPrefix}:{AssemblyVersion}"),
```
jviau, line 24: I know you already have that, but interesting we emit `ai.sdk.prefix` for all exporters, including non azure monitor.

src/WebJobs.Script/Diagnostics/OpenTelemetry/FunctionsResourceDetector.cs
```diff
@@ -10,66 +12,131 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics.OpenTelemetry
...
+        }
+
+        private static bool IsResourceAttributeConfigured(string key, string? resourceAttributes)
+        {
+            if (resourceAttributes is not { Length: > 2 })
+            {
+                return false;
+            }
+
+            foreach (var segment in resourceAttributes.Split(','))
```
jviau: This will allocate a string per segment. You can do this with no allocations by converting to a span first and then walking through the span.

Not sure if we have access to https://learn.microsoft.com/en-us/dotnet/api/system.memoryextensions.split?view=net-10.0#system-memoryextensions-split-1(system-readonlyspan((-0))-0) or not

src/WebJobs.Script/Diagnostics/OpenTelemetry/FunctionsResourceDetector.cs
```diff
@@ -10,66 +12,131 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics.OpenTelemetry
...
+                {
+                    attributes.Add(new(ResourceSemanticConventions.ServiceName, azureWebsiteName ?? typeof(ScriptHost).Assembly.GetName().Name ?? "unknown"));
+                }
+
+                // Priority: OTEL_RESOURCE_ATTRIBUTES[service.version] > AssemblyVersion
+                // OTel decided to not have OTEL_SERVICE_VERSION, so we only check OTEL_RESOURCE_ATTRIBUTES
+                // https://github.com/open-telemetry/semantic-conventions/issues/2669
+                if (!IsResourceAttributeConfigured(ResourceSemanticConventions.ServiceVersion, resourceAttributes))
+                {
+                    attributes.Add(new(ResourceSemanticConventions.ServiceVersion, AssemblyVersion));
```
jviau, line 42: This will impact customers' ability to see what func host version the telemetry is from. Is that important to keep somewhere?

jviau (approved):

src/WebJobs.Script/Diagnostics/OpenTelemetry/FunctionsResourceDetector.cs
RohitRanjanMS: Yeah, I only need this for the AzureMonitorExporter to build the sdkVersion attribute, but it ends up being emitted by all exporters. Instrumentation is exporter‑agnostic.
I could add logic to check whether Azure Monitor is enabled, but if both exporters are active, the attribute would still go to both. That would just make things more confusing.

src/WebJobs.Script/Diagnostics/OpenTelemetry/FunctionsResourceDetector.cs
RohitRanjanMS: Updated the logic.

src/WebJobs.Script/Diagnostics/OpenTelemetry/FunctionsResourceDetector.cs
RohitRanjanMS: We have it in the sdkVersion attribute.

src/WebJobs.Script/Diagnostics/OpenTelemetry/FunctionsResourceDetector.cs
```diff
@@ -10,66 +12,145 @@ namespace Microsoft.Azure.WebJobs.Script.Diagnostics.OpenTelemetry
...
+            if (Environment.GetEnvironmentVariable(ResourceSemanticConventions.ServiceNameEnvVar) is { Length: > 0 })
+            {
+                return true;
+            }
+
+            // Fall back to checking OTEL_RESOURCE_ATTRIBUTES
+            return IsResourceAttributeConfigured(ResourceSemanticConventions.ServiceName, resourceAttributes);
+        }
+
+        private static bool IsResourceAttributeConfigured(string key, string? resourceAttributes)
```
jviau, line 117: Just want to make sure this has good test coverage.

Also add a TODO to replace implementation with `MemoryExtensions.Split` when we upgrade to net10.

jviau (approved):

satvu (approved):

src/WebJobs.Script/Diagnostics/OpenTelemetry/FunctionsResourceDetector.cs
RohitRanjanMS: yes, I have good test coverage. 
Added TODO

satvu (approved):

PR #11511, opened 2025-12-08
Refactor functions worker runtime retrieval to use `IWorkerRuntimeResolver` abstraction

### LINKED ISSUE #11482: Finalize changes to expose the configuration profile at initialization

kshyju:
This PR introduces a new abstraction, `IWorkerRuntimeResolver`, for retrieving the Functions worker runtime value. All application code paths that previously read the runtime directly from environment variables have been updated to use this abstraction instead.

There are 2 implementations of this.

#### ScriptHostWorkerRuntimeResolver
Provides the worker runtime for the current Script Host scope.

- Applies custom logic when a configuration profile is present.
- Falls back to reading the value from the environment variable when no special handling is required.
- Caches the resolved value and returns it on subsequent calls.

#### WebHostWorkerRuntimeResolverAdapter
A Web Host–level adapter that:

- Delegates to `ScriptHostWorkerRuntimeResolver` when it becomes available.
- Falls back to reading from the environment variable until the Script Host–specific resolver can be used.


src/WebJobs.Script/Host/HostFunctionMetadataProvider.cs
```diff
@@ -104,6 +104,7 @@ private FunctionMetadata ReadFunctionMetadata(string functionDirectory, IFileSys
 
                     JObject functionConfig = JObject.Parse(json);
 
+                    // FACAVAL|FWR: Dependency on parent scope
```
kshyju: Will follow up with Fabio to understand he had any specific changes here in mind.

src/WebJobs.Script/Workers/ScriptHostWorkerRuntimeResolver.cs
```diff
@@ -0,0 +1,49 @@
...
+            }
+
+            if (string.Equals(_scriptJobHostOptionsMonitor.CurrentValue.ConfigurationProfile, "mcp-custom-handler"))
+            {
+                return _resolvedWorkerRuntime = "custom";
+            }
+
+            if (string.Equals(_scriptJobHostOptionsMonitor.CurrentValue.ConfigurationProfile, "web-app-custom-handler"))
+            {
+                return _resolvedWorkerRuntime = "custom";
```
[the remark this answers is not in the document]
kshyju: Updated to use OrdinalIgnoreCase comparison.

src/WebJobs.Script.Grpc/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -95,7 +97,8 @@ public RpcFunctionInvocationDispatcher(IOptions<ScriptJobHostOptions> scriptHost
             _managedDependencyOptions = managedDependencyOptions ?? throw new ArgumentNullException(nameof(managedDependencyOptions));
             _logger = loggerFactory.CreateLogger<RpcFunctionInvocationDispatcher>();
             _rpcWorkerChannelFactory = rpcWorkerChannelFactory;
-            _workerRuntime = _environment.GetEnvironmentVariable(EnvironmentSettingNames.FunctionWorkerRuntime);
+            _workerRunimeResolver = workerRunimeResolver;
```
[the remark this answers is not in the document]
kshyju: Local assignment was not needed. Removed now.

test/WebJobs.Script.Tests/Workers/Rpc/RpcWorkerConfigTests.cs
```diff
@@ -686,13 +686,14 @@ private IEnumerable<RpcWorkerConfig> TestReadWorkerProviderFromConfig(IEnumerabl
                 var scriptSettingsManager = new ScriptSettingsManager(config);
                 var workerProfileManager = new Mock<IWorkerProfileManager>();
                 var testMetricLogger = new TestMetricsLogger();
+                var workerRuntimeResolver = new Mock<IWorkerRuntimeResolver>(MockBehavior.Strict);
```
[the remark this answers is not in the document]
kshyju: Removed.

src/WebJobs.Script.Grpc/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -95,7 +97,8 @@ public RpcFunctionInvocationDispatcher(IOptions<ScriptJobHostOptions> scriptHost
...
+            _workerRuntime = _workerRunimeResolver.GetWorkerRuntime();
```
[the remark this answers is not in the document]
kshyju: local variable is not needed anymore. Removed.

test/WebJobs.Script.Tests/WebJobsScriptHostServiceTests.cs
```diff
@@ -40,6 +41,7 @@ public class WebJobsScriptHostServiceTests
         private ILoggerFactory _loggerFactory;
         private Mock<IScriptWebHostEnvironment> _mockScriptWebHostEnvironment;
         private Mock<IEnvironment> _mockEnvironment;
+        private Mock<IWorkerRuntimeResolver> _mockWorkerRuntimeResolver;
```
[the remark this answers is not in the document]
kshyju: Fixed.

src/WebJobs.Script.WebHost/DependencyInjection/WebHostWorkerRuntimeResolverAdapter.cs
```diff
@@ -0,0 +1,124 @@
...
+                return jobHostResolver.GetWorkerRuntime(defaultValue);
+            }
+
+            // Fallback to read from environment when Job Host scoped resolver is not available yet
+            _environment ??= _rootProvider.GetRequiredService<IEnvironment>();
+
+            return _environment.GetEnvironmentVariableOrDefault(
+                EnvironmentSettingNames.FunctionWorkerRuntime,
+                defaultValue);
+        }
```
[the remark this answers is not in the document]
kshyju: This singleton is disposed only during application shutdown. At that point, `GetWorkerRuntime()` is not expected to be called. So checking disposed in GetWorkerRuntime is unnecessary.

src/WebJobs.Script/Host/HostFunctionMetadataProvider.cs
kshyju: Replaced env variable access with new abstraction.

src/WebJobs.Script.Grpc/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -78,7 +79,8 @@ public RpcFunctionInvocationDispatcher(IOptions<ScriptJobHostOptions> scriptHost
             IRpcFunctionInvocationDispatcherLoadBalancer functionDispatcherLoadBalancer,
             IOptions<WorkerConcurrencyOptions> workerConcurrencyOptions,
             IOptions<FunctionsHostingConfigOptions> hostingConfigOptions,
-            IHostMetrics hostMetrics)
+            IHostMetrics hostMetrics,
+            IWorkerRuntimeResolver workerRunimeResolver)
```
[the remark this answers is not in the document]
kshyju: Fixed typo.

src/WebJobs.Script.WebHost/DependencyInjection/WebHostWorkerRuntimeResolverAdapter.cs
```diff
@@ -0,0 +1,124 @@
...
+    /// Otherwise, falls back to environment based resolution.
+    /// </summary>
+    internal sealed class WebHostWorkerRuntimeResolverAdapter : IWorkerRuntimeResolver, IDisposable
+    {
+        private readonly IServiceProvider _rootProvider;
+        private readonly ILogger<WebHostWorkerRuntimeResolverAdapter> _logger;
+        private IWorkerRuntimeResolver _cachedHostResolver;
+        private IScriptHostManager _hostManager;
+        private IEnvironment _environment;
+        private bool _disposed;
```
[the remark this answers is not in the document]
kshyju: using of disposed is in a thread safe manner now.

src/WebJobs.Script/Workers/ScriptHostWorkerRuntimeResolver.cs
```diff
@@ -0,0 +1,50 @@
...
+                              StringComparison.OrdinalIgnoreCase)
+                || string.Equals(_scriptJobHostOptionsMonitor.CurrentValue.ConfigurationProfile,
+                                 HostConfigurationProfile.WebAppCustomHandlerProfile,
+                                 StringComparison.OrdinalIgnoreCase))
+            {
+                return _resolvedWorkerRuntime = "custom";
+            }
+
+            return _resolvedWorkerRuntime = _environment.GetEnvironmentVariableOrDefault(EnvironmentSettingNames.FunctionWorkerRuntime, defaultValue);
+        }
```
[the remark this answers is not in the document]
kshyju: _resolvedWorkerRuntime is thead safe now.

src/WebJobs.Script/Workers/ScriptHostWorkerRuntimeResolver.cs
```diff
@@ -0,0 +1,58 @@
...
+                || string.Equals(_scriptJobHostOptionsMonitor.CurrentValue.ConfigurationProfile,
+                                 HostConfigurationProfile.WebAppCustomHandlerProfile,
+                                 StringComparison.OrdinalIgnoreCase))
+            {
+                workerRuntime = "custom";
+            }
+            else
+            {
+                workerRuntime = _environment.GetEnvironmentVariableOrDefault(EnvironmentSettingNames.FunctionWorkerRuntime, defaultValue);
+            }
```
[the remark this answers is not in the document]
kshyju: No. I prefer the existing approach of style in this case. I find that to be equally readable.

src/WebJobs.Script.WebHost/DependencyInjection/WebHostWorkerRuntimeResolverAdapter.cs
```diff
@@ -0,0 +1,122 @@
...
+    /// If a Script Host scoped resolver is available, delegates to it.
+    /// Otherwise, falls back to environment based resolution.
+    /// </summary>
+    internal sealed class WebHostWorkerRuntimeResolverAdapter : IWorkerRuntimeResolver, IDisposable
+    {
+        private readonly IServiceProvider _rootProvider;
+        private readonly ILogger<WebHostWorkerRuntimeResolverAdapter> _logger;
+        private IWorkerRuntimeResolver _cachedHostResolver;
+        private IScriptHostManager _hostManager;
+        private IEnvironment _environment;
```
brettsam: We're generally trying to reduce our usage of `IEnvironment` -- can this use `IConfiguration` instead?

src/WebJobs.Script.WebHost/DependencyInjection/WebHostWorkerRuntimeResolverAdapter.cs
```diff
@@ -0,0 +1,122 @@
...
+                }
+            }
+
+            if (scriptHostResolver is not null)
+            {
+                return scriptHostResolver.GetWorkerRuntime(defaultValue);
+            }
+
+            // Fallback to read from environment when Job Host scoped resolver is not available yet
+            _environment ??= _rootProvider.GetRequiredService<IEnvironment>();
```
brettsam: Why would `_rootProvider` need to be called here? I thought the passed in `IEnvironment` already would have come from the root?

src/WebJobs.Script.WebHost/DependencyInjection/WebHostWorkerRuntimeResolverAdapter.cs
```diff
@@ -0,0 +1,122 @@
...
+        private void EnsureSubscribedToHostManagerStateChange()
+        {
+            // Fast-path: already subscribed.
+            var hostManager = _hostManager;
+            if (hostManager is not null)
+            {
+                return;
+            }
+
+            hostManager = _rootProvider.GetRequiredService<IScriptHostManager>();
```
brettsam, line 123: Same question about the root provider here -- or maybe I'm not quite understanding the scope of these...

src/WebJobs.Script.WebHost/DependencyInjection/WebHostWorkerRuntimeResolverAdapter.cs
```diff
@@ -0,0 +1,122 @@
...
+            EnsureSubscribedToHostManagerStateChange();
+
+            var scriptHostResolver = _cachedHostResolver;
+
+            if (scriptHostResolver is null)
+            {
+                var scriptHostWorkerResolver = _rootProvider.GetScriptHostServiceOrNull<IWorkerRuntimeResolver>();
+                if (scriptHostWorkerResolver is not null)
+                {
+                    _logger.ScriptHostWorkerResolverResolvedSuccessfully();
```
brettsam, line 53: How noisy is this? Want to make sure we're not flooding Kusto...

src/WebJobs.Script/Workers/ScriptHostWorkerRuntimeResolver.cs
```diff
@@ -0,0 +1,58 @@
...
+using System.Threading;
+using Microsoft.Azure.WebJobs.Script.Configuration;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.WebJobs.Script.Workers
+{
+    /// <summary>
+    /// Resolves the worker runtime for the current script host instance.
+    /// </summary>
+    internal sealed class ScriptHostWorkerRuntimeResolver : IWorkerRuntimeResolver
```
brettsam, line 17: Did you look into having this be part of an Options object instead of our own interface? On the surface it looks like it could work as we want to store/cache a simple string at the end of the day and a ConfigureOptions class could do all of that...

src/WebJobs.Script.WebHost/DependencyInjection/WebHostWorkerRuntimeResolverAdapter.cs
kshyju: Fixed.

brettsam:
@kshyju -- did you want to get these merges cleaned up and then we can continue the review? I'm okay with this approach now that I saw how complicated some of the IOptions<> dependencies can get...

test/WebJobs.Script.Tests/Workers/Rpc/WorkerConfigurationProviderBaseTests.cs
```diff
@@ -158,7 +167,7 @@ public void WorkerDescription_Skipped_When_Profile_Disables_Worker()
 
             var loggerFactory = GetTestLoggerFactory();
             var testScriptHostManager = new Mock<IScriptHostManager>();
-            var optionsMonitor = GetTestWorkerConfigurationResolverOptions(config, testEnvironment, testScriptHostManager.Object);
+            var optionsMonitor = GetTestWorkerConfigurationResolverOptions(config, workerRuntime: null, testEnvironment, testScriptHostManager.Object);
```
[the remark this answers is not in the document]
kshyju: This is a false positive. Since C# 7.2, named arguments are permitted in non-trailing positions as long as they are in the correct ordinal position ([spec reference](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-version-history#c-version-72)). In all of these call sites, `workerRuntime:` is the second argument matching the second parameter in the `GetTestWorkerConfigurationResolverOptions` signature, so positional arguments can legally follow it. The project builds without errors or warnings.

test/WebJobs.Script.Tests/Workers/Rpc/DynamicWorkerConfigurationProviderTests.cs
```diff
@@ -57,7 +55,12 @@ public void GetWorkerConfigs_MultiLanguageWorker_ReturnsExpectedConfigs(string r
 
             var hostingOptions = new FunctionsHostingConfigOptions();
             hostingOptions.Features.Add(RpcWorkerConstants.WorkersAvailableForDynamicResolution, "java|node");
-            var optionsMonitor = GetTestWorkerConfigurationResolverOptions(config, mockEnvironment.Object, testScriptHostManager.Object, new OptionsWrapper<FunctionsHostingConfigOptions>(hostingOptions));
+            var optionsMonitor = GetTestWorkerConfigurationResolverOptions(
+                config,
+                workerRuntime: null,
+                mockEnvironment.Object,
+                testScriptHostManager.Object,
+                new OptionsWrapper<FunctionsHostingConfigOptions>(hostingOptions));
```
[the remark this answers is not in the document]
kshyju: This is a false positive. Since C# 7.2, named arguments are permitted in non-trailing positions as long as they are in the correct ordinal position ([spec reference](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-version-history#c-version-72)). In all of these call sites, `workerRuntime:` is the second argument matching the second parameter in the `GetTestWorkerConfigurationResolverOptions` signature, so positional arguments can legally follow it. The project builds without errors or warnings.

kshyju:
> @kshyju -- did you want to get these merges cleaned up and then we can continue the review? I'm okay with this approach now that I saw how complicated some of the IOptions<> dependencies can get...

@brettsam Rebased and ready to review.

src/WebJobs.Script/Host/ScriptHost.cs
```diff
@@ -471,9 +477,9 @@ private void PreInitialize()
             InitializeFileSystem();
         }
 
-        internal static void ValidateFunctionsWorkerRuntime(IEnvironment environment, IOptions<FunctionsHostingConfigOptions> hostingConfigOptions, ILogger logger)
+        internal static void ValidateFunctionsWorkerRuntime(IWorkerRuntimeResolver workerRuntimeResolver, IOptions<FunctionsHostingConfigOptions> hostingConfigOptions, ILogger logger)
         {
-            if (string.IsNullOrEmpty(environment.GetFunctionsWorkerRuntime()))
+            if (string.IsNullOrEmpty(workerRuntimeResolver.GetWorkerRuntime()))
```
liliankasem, line 482: Can't this just use the global resolver (`_workerRuntimeResolver`)? Why do we need to pass it as a param?

src/WebJobs.Script/Workers/IWorkerRuntimeResolver.cs
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Script.Workers
+{
+    /// <summary>
+    /// Defines a mechanism for resolving the functions worker runtime value for the current context.
+    /// </summary>
+    public interface IWorkerRuntimeResolver
```
liliankasem, line 9: Could this be internal?

src/WebJobs.Script/Workers/ScriptHostWorkerRuntimeResolver.cs
```diff
@@ -0,0 +1,69 @@
...
+            _scriptJobHostOptionsMonitor = scriptJobHostOptionsMonitor;
+        }
+
+        public string GetWorkerRuntime(string defaultValue = null)
+        {
+            var cachedRuntime = _resolvedWorkerRuntime;
+            if (cachedRuntime is not null)
+            {
+                return cachedRuntime;
+            }
```
liliankasem, line 40: no need for extra allocation here?

```suggestion
            if (_resolvedWorkerRuntime is not null)
            {
                return _resolvedWorkerRuntime;
            }
```

src/WebJobs.Script/Workers/ScriptHostWorkerRuntimeResolver.cs
```diff
@@ -0,0 +1,69 @@
...
+            IOptionsMonitor<ScriptJobHostOptions> scriptJobHostOptionsMonitor)
+        {
+            ArgumentNullException.ThrowIfNull(environment);
+            ArgumentNullException.ThrowIfNull(scriptJobHostOptionsMonitor);
+
+            _environment = environment;
+            _scriptJobHostOptionsMonitor = scriptJobHostOptionsMonitor;
+        }
+
+        public string GetWorkerRuntime(string defaultValue = null)
```
liliankasem, line 34: nit: might be nice to have a convenience property i.e.

`public string WorkerRuntime => GetWorkerRuntime();`

src/WebJobs.Script/Config/HostConfigurationProfile.cs
```diff
@@ -29,6 +25,10 @@ public sealed class HostConfigurationProfile
             [ConfigurationPath.Combine(ConfigurationSectionNames.CustomHandler, "http", "routes", "0", "route")] = "{*route}"
         };
 
+        internal const string McpCustomHandlerProfile = "mcp-custom-handler";
+
+        internal const string WebAppCustomHandlerProfile = "web-app-custom-handler";
```
liliankasem, line 31: These were propbably private since only this class used them, instead of making them internal - does it make more sense to move these to a Constants file?

src/WebJobs.Script.Grpc/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
```diff
@@ -198,7 +199,7 @@ internal async Task InitializeWebhostLanguageWorkerChannel(IEnumerable<string> l
 
         internal async void ShutdownWebhostLanguageWorkerChannels()
         {
-            _logger.LogDebug("{workerRuntimeConstant}={value}. Will shutdown all the worker channels that started in placeholder mode", EnvironmentSettingNames.FunctionWorkerRuntime, _workerRuntime);
+            _logger.LogDebug("Worker runtime={value}. Will shutdown all the worker channels that started in placeholder mode", _workerRuntime);
```
liliankasem, line 202: Will this impact any detectors?

src/WebJobs.Script.WebHost/DependencyInjection/WebHostWorkerRuntimeResolverAdapter.cs
liliankasem: +1

src/WebJobs.Script/Workers/ScriptHostWorkerRuntimeResolver.cs
liliankasem, line 34: Do we want to consider introducing an enum here instead of using string? Since we are already refactoring, it might be nice to avoid all the string == const checks

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -82,10 +83,20 @@ public class WebJobsScriptHostService : IHostedService, IScriptHostManager, ISer
...
+            HostPerformanceManager hostPerformanceManager,
+            IOptions<HostHealthMonitorOptions> healthMonitorOptions,
+            IMetricsLogger metricsLogger,
+            IApplicationLifetime applicationLifetime,
+            IConfiguration config,
+            IScriptEventManager eventManager,
+            IHostMetrics hostMetrics,
+            IWorkerRuntimeResolver workerRuntimeResolver,
             IOptions<FunctionsHostingConfigOptions> hostingConfigOptions,
             WorkerConfigCacheInvalidator workerConfigCacheInvalidator)
```
liliankasem, line 101: Maybe out of scope for this PR, but is the workerConfigCacheInvalidator thread safe? A couple of suggestions came up from the AI review about this class - let me know if you think these concerns are valid:

---

#### Guard the WorkerConfigCacheInvalidator downcast with an assertion

In `WorkerConfigCacheInvalidator.InvalidateCache()`, the pattern match against `RefreshWorkerOptionsChangeTokenSource<T>` silently no-ops if a different IOptionsChangeTokenSource<T> is registered. Consider either:

1. Adding a `Debug.Assert()` or warning log when the downcast fails, so misconfiguration is detectable
2. Accepting `RefreshWorkerOptionsChangeTokenSource<T>` directly as the constructor parameter type instead of the interface, since the class explicitly depends on that concrete behavior (the xmldoc already says this, but the type system doesn't enforce it)

#### Make WorkerConfigCacheInvalidator thread-safety assumptions explicit

The `_firstRun` and `_postBuildInvalidationEnabled` fields are plain bool with no synchronization, relying on the caller holding the host start semaphore. This assumption is implicit. Options:

1. Add a code comment or xmldoc noting the expected threading model
2. Or use Interlocked operations on int flags (like the adapter's _disposed) for defense-in-depth, which is cheap and avoids a subtle bug if future callers bypass the semaphore

src/WebJobs.Script.WebHost/DependencyInjection/WebHostWorkerRuntimeResolverAdapter.cs
```diff
@@ -0,0 +1,143 @@
...
+        {
+            ArgumentNullException.ThrowIfNull(rootProvider);
+            ArgumentNullException.ThrowIfNull(environment);
+            ArgumentNullException.ThrowIfNull(logger);
+            _rootProvider = rootProvider;
+            _environment = environment;
+            _logger = logger;
+        }
+
+        public string GetWorkerRuntime(string defaultValue = null)
```
liliankasem, line 42: AI suggestion:

> The defaultValue parameter on GetWorkerRuntime(string defaultValue = null) prevents caching null results (since different callers pass different defaults). If instead the interface returned string? with no default parameter, callers would do resolver.GetWorkerRuntime() ?? "fallback" at the call site. This would let the resolver always cache its resolved value (including null = "no runtime configured"), improving performance on the pre-specialization path where every call currently re-reads the environment variable.

src/WebJobs.Script/Workers/ScriptHostWorkerRuntimeResolver.cs
liliankasem: I got the same AI suggestion for this - @kshyju did you look into IOptions instead?

>The .NET Options pattern already has built-in caching, change-token support, and DI integration. An alternative to IWorkerRuntimeResolver would be a WorkerRuntimeOptions class with a Runtime property, configured via IConfigureOptions<WorkerRuntimeOptions> (the setup reads the env var / profile). Consumers would inject IOptionsMonitor<WorkerRuntimeOptions> and get automatic change notification and caching from the framework. This would eliminate the hand-rolled Interlocked.CompareExchange caching and ActiveHostChanged subscription logic, relying instead on the same pattern already used for LanguageWorkerOptions.

liliankasem (commented):
initial review

src/WebJobs.Script.WebHost/DependencyInjection/WebHostWorkerRuntimeResolverAdapter.cs
kshyju: The current behavior to not cache the default value passed in is intentional. I have code comments in the method and a test to ensure that. This is to prevent cache poisoning. One caller can update the cache with the default value and another caller may get it before it gets  a chance to see /readthe actual underlying env value.

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
kshyju: Good suggestions. Like you said, it should be a separate PR addressing that.

src/WebJobs.Script/Workers/ScriptHostWorkerRuntimeResolver.cs
kshyju: Did you mean changing `_scriptJobHostOptionsMonitor.CurrentValue.ConfigurationProfile` to enum instead of string? That might need additional changes on how the Option is populated etc. Perhaps a future improvement if we think it adds any value. This PR is already big, and I prefer to avoid increasing the scope further.

src/WebJobs.Script.Grpc/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
kshyju: I am not aware of any. But not 100% sure about it.  Changing this log statement here makes it correct (with this new changes).  What is the best way to check this?

src/WebJobs.Script/Config/HostConfigurationProfile.cs
kshyju: _Agent response, which I agree._

They were made internal because ScriptHostWorkerRuntimeResolver now needs to compare against these profile names to determine the worker runtime . The constants are semantically tied to the profile concept — they define profile names and are used alongside HostConfigurationProfile.Get(), the profile static instances (McpCustomHandler, WebAppCustomHandler), and the SupportedValues string. Moving them to a generic Constants file would separate them from the type they describe. The existing comment at line 14 ("note: profile name consts are intentionally private...") already explains the original design intent — changing to internal still preserves the test-safety property since tests reference these constants too. Prefer keeping them in HostConfigurationProfile for cohesion.

src/WebJobs.Script/Workers/ScriptHostWorkerRuntimeResolver.cs
kshyju: We don't want anyone to directly use this class. All access should be via the new abstraction (`IWorkerRuntimeResolver.GetWorkerRuntime`). Is there a use case you have in mind where exposing this property is helpful?

src/WebJobs.Script/Workers/IWorkerRuntimeResolver.cs
kshyju: There are public class constructors to which this is being injected. So this interface has to be also public.

src/WebJobs.Script/Host/ScriptHost.cs
kshyju: The method is internal static so that it can be unit tested directly in ScriptHostTests.cs without constructing a full ScriptHost instance.

src/WebJobs.Script.WebHost/DependencyInjection/WebHostWorkerRuntimeResolverAdapter.cs
kshyju: My copilot CLI agent review flagged this as a major problem and I then ended up fixing it.

src/WebJobs.Script/Config/HostConfigurationProfile.cs
liliankasem: not a hill I want to die on :)

src/WebJobs.Script/Workers/ScriptHostWorkerRuntimeResolver.cs
liliankasem: I just meant that worker runtime should be an enum i.e.

```chsarp
    public enum WorkerRuntime
    {
        None,
        Dotnet,
        DotnetIsolated,
        Node,
        Python,
        Java,
        Powershell,
        Custom
    }
```

src/WebJobs.Script.Grpc/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
liliankasem: I don't know if there is an easy way - one approach is to open the detectors repo and search for the original log statement to see if it comes up at all

src/WebJobs.Script/Workers/ScriptHostWorkerRuntimeResolver.cs
liliankasem: Does not necessarily mean `_scriptJobHostOptionsMonitor.CurrentValue.ConfigurationProfile` needs to change, just what `GetWorkerRuntime` returns and how the consumers of this method check what the language is

src/WebJobs.Script/Workers/ScriptHostWorkerRuntimeResolver.cs
kshyju: The local variable copies only the reference (a pointer-sized value on the stack), not the string contents, so there is no heap allocation involved. The reason it exists is thread safety. `_resolvedWorkerRuntime` is a shared field . Without the local, the JIT/CPU could issue two separate reads of the field, one for the null check and one for the return, and those reads could observe different values. Capturing a single snapshot in a local guarantees the null check and the return operate on the same value, eliminating a TOCTOU (time-of-check-time-of-use) race. This is a standard lock-free pattern.

src/WebJobs.Script/Workers/ScriptHostWorkerRuntimeResolver.cs
kshyju: Yes, the IOptions approach was considered. The key challenge is that this resolver bridges two DI scopes. The WebHost singleton scope and the Script Host child scope, which the Options pattern doesn't handle well out of the box:

Cross-scope delegation: WebHostWorkerRuntimeResolverAdapter (WebHost singleton) needs to delegate to the Script Host-scoped resolver when available, falling back to environment variable reads when no host is running. IOptionsMonitor<T> doesn't have a concept of "delegate to a child scope's options when available."

Specialization timing: During placeholder mode, the env var isn't set yet. After specialization, it is. The adapter must re-read on each call until a value appears, then cache it. IConfigureOptions runs once at options creation time & it wouldn't naturally re-evaluate on each access.

The IWorkerRuntimeResolver abstraction gives us explicit control over these scope-bridging and lifecycle concerns. That said, if we find the hand-rolled caching becomes a maintenance burden, we could revisit. The ScriptHostWorkerRuntimeResolver (the Script Host-scoped one) is simpler and could potentially be an Options object, but keeping both implementations behind the same interface is cleaner for consumers.

@brettsam  agreed it is not a bad idea to use custom abstraction instead of Options here.

src/WebJobs.Script.WebHost/DependencyInjection/WebHostWorkerRuntimeResolverAdapter.cs
kshyju: Expect this to be logged once per script host instance.

test/WebJobs.Script.Tests/Workers/Rpc/WorkerConfigurationProviderBaseTests.cs
```diff
@@ -41,7 +41,7 @@ public void DefaultLanguageWorkersDir()
             var testScriptHostManager = new Mock<IScriptHostManager>();
             var testLoggerFactory = GetTestLoggerFactory();
 
-            var optionsMonitor = GetTestWorkerConfigurationResolverOptions(config, testEnvironment, testScriptHostManager.Object, null);
+            var optionsMonitor = GetTestWorkerConfigurationResolverOptions(config, workerRuntime: null, testEnvironment, testScriptHostManager.Object, null);
```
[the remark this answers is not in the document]
kshyju: false alarm.

src/WebJobs.Script/Workers/ScriptHostWorkerRuntimeResolver.cs
kshyju: Asked agent to prepare pros and cons and impact.

<blockquote>
  Impact of Switching Worker Runtime from string to enum

  What exists today

  7 known runtime values in RpcWorkerConstants:

   - dotnet, dotnet-isolated, node, java, powershell, python, custom

  These are compared via case-insensitive string.Equals in ~25+ locations across ~30 source files with ~300+ references to workerRuntime/_workerRuntime/WorkerRuntime.

  Pros

   1. Type safety — Eliminates misspelled runtime strings. Compiler catches invalid values.
   2. No more case-sensitivity bugs — Today there's inconsistency: some comparisons use OrdinalIgnoreCase, others InvariantCultureIgnoreCase, and one at RpcFunctionInvocationDispatcher:480 uses plain string.Equals with no StringComparison at all (an existing bug). An enum eliminates all of these.
   3. Cleaner branching — switch expressions on enum vs. chained string.Equals calls (e.g., WebHostRpcWorkerChannelManager:210-212 checks node/powershell/python in 3 separate string comparisons).
   4. Performance — Enum comparison is a single integer op vs. string comparison. Marginal, but it's on a hot path (resolver is called frequently).

  Cons / Risks

   1. Massive scope — ~300+ references across ~30 source files in src/, plus an equivalent or larger number in test files. This is not a small refactor.
   2. String↔Enum boundary — The runtime value originates as a string from the FUNCTIONS_WORKER_RUNTIME environment variable and flows through several string-based systems:
    - WorkerDescription.Language is a string property used throughout the worker config pipeline. WorkerConfigs are matched to the runtime via c.Description.Language.Equals(workerRuntime). Changing this to enum means touching the entire worker configuration resolution chain.
    - FunctionMetadata.Language is also a string. It's compared against the runtime in Utility.cs for function validation.
    - Logging — The runtime string is interpolated into 20+ log messages. Each would need .ToString() or a format change.
    - Dictionary keys — WebHostRpcWorkerChannelManager._workerChannels uses the runtime string as a dictionary key.
   3. Unknown/custom runtimes — The enum would need an Unknown or Other variant for unrecognized values from the environment variable. This loses information (the actual string) and requires special handling everywhere.
   4. Serialization concerns — The runtime string appears in configuration (host.json via HostJsonFileConfigurationSource:261), API responses (HostController.GetWorkerProcesses), and metrics/telemetry. All of these expect the string form.
   5. Worker profile system — WorkerDescriptionProfile has ProfileDescription.Language as a string that gets compared to the runtime. The profile system would need parallel changes.
   6. External contract — Worker descriptions loaded from worker.config.json files define Language as a string. The parsing pipeline would need enum conversion at the boundary.

  Bottom line

  The enum is a better domain model, but the runtime string is deeply embedded as a cross-cutting concern — it flows from environment variables through configuration, worker registration, function metadata, channel management, logging, and API responses. The conversion boundary (string → enum at input, enum →    
  string at output) would need to be introduced at multiple layers. This is a dedicated refactoring effort, not something to fold into the resolver PR.
</blockquote>

liliankasem: I think we should do it since we are refactoring anyway, but OK for this to be out of scope of this PR

src/WebJobs.Script/Workers/ScriptHostWorkerRuntimeResolver.cs
```diff
@@ -0,0 +1,69 @@
...
+    /// has been configured (e.g., via environment variables or specialization).
+    /// </summary>
+    internal sealed class ScriptHostWorkerRuntimeResolver : IWorkerRuntimeResolver
+    {
+        private readonly IEnvironment _environment;
+        private readonly IOptionsMonitor<ScriptJobHostOptions> _scriptJobHostOptionsMonitor;
+        private string _resolvedWorkerRuntime;
+
+        public ScriptHostWorkerRuntimeResolver(
+            IEnvironment environment,
```
brettsam: We've been talking about getting rid of `IEnvironment` and using `IConfiguration` everywhere -- can we take that step here and populate this from `IConfiguration` instead?

src/WebJobs.Script/Workers/ScriptHostWorkerRuntimeResolver.cs
```diff
@@ -0,0 +1,69 @@
...
+
+            string workerRuntime;
+            if (string.Equals(_scriptJobHostOptionsMonitor.CurrentValue.ConfigurationProfile,
+                              HostConfigurationProfile.McpCustomHandlerProfile,
+                              StringComparison.OrdinalIgnoreCase)
+                || string.Equals(_scriptJobHostOptionsMonitor.CurrentValue.ConfigurationProfile,
+                                 HostConfigurationProfile.WebAppCustomHandlerProfile,
+                                 StringComparison.OrdinalIgnoreCase))
+            {
+                workerRuntime = RpcWorkerConstants.CustomHandlerLanguageWorkerName;
```
brettsam: I don't love this being hardcoded in here -- it makes it tough to search through the codebase to see "what exactly does the mcp profile do?". If we do switch to `IConfiguration`, I wonder if it'd make more sense to have these profiles as "real" config providers and inject their settings into `IConfiguration` all in one spot?

I don't think it should block this, but I do see it as something we're trying to improve on as we change things.

src/WebJobs.Script.WebHost/DependencyInjection/WebHostWorkerRuntimeResolverAdapter.cs
brettsam: Want to bump this up -- can't we inject IScriptHostManager into this class and not have to do the service locator thing?

We may have talked about it already but I forget :-)

src/WebJobs.Script.WebHost/DependencyInjection/WebHostWorkerRuntimeResolverAdapter.cs
```diff
@@ -0,0 +1,143 @@
...
+            // a value is available. Caching the "not set" state would cause a race condition
+            // since we rely on OnActiveHostChanged to clear the cache, which fires after
+            // the value is already needed.
+            var cachedValue = _cachedEnvironmentValue;
+            if (cachedValue is not null)
+            {
+                return cachedValue;
+            }
+
+            var valueFromEnvironment = _environment.GetEnvironmentVariable(EnvironmentSettingNames.FunctionWorkerRuntime);
```
brettsam: Same question as elsewhere... can we pull this from `IConfiguration` and phase out `IEnvironment`?

brettsam (changes requested):

src/WebJobs.Script.WebHost/DependencyInjection/WebHostWorkerRuntimeResolverAdapter.cs
kshyju: Switched to read from IConfiguration.

src/WebJobs.Script/Workers/ScriptHostWorkerRuntimeResolver.cs
kshyju: Switched to read from IConfiguration.

src/WebJobs.Script/Workers/ScriptHostWorkerRuntimeResolver.cs
kshyju: Good suggestion. In the latest iteration, I removed the code to check profile information from this class and moved it to HostConfigurationProfile class where it will add to IConfiguration. (This is a new config entry called "ProfileWorkerRuntime" which will be used if present when FWR is requested, else fallback to our existing logic to read FUNCTIONS_WORKER_RUNTIME from IConfiguration).

src/WebJobs.Script.WebHost/DependencyInjection/WebHostWorkerRuntimeResolverAdapter.cs
```diff
@@ -0,0 +1,144 @@
...
+        private void EnsureSubscribedToHostManagerStateChange()
+        {
+            // Fast-path: already subscribed.
+            var hostManager = _hostManager;
+            if (hostManager is not null)
+            {
+                return;
+            }
+
+            hostManager = _rootProvider.GetRequiredService<IScriptHostManager>();
```
kshyju, line 123: This is intentional. We cannot DI to constructor as that would cause a circular dependency. 

   IWorkerRuntimeResolver (WebHostWorkerRuntimeResolverAdapter)
     └─ needs IScriptHostManager (WebJobsScriptHostService)
          └─ needs IWorkerRuntimeResolver  ← CIRCULAR

So we are doing a deferred resolution here.

src/WebJobs.Script/Workers/ScriptHostWorkerRuntimeResolver.cs
```diff
@@ -40,19 +39,13 @@ public string GetWorkerRuntime(string defaultValue = null)
...
-            }
-            else
+            // Check the worker runtime from the options first (set by configuration profiles),
+            // then fall back to the configuration entry (e.g., FUNCTIONS_WORKER_RUNTIME env var).
+            string workerRuntime = _scriptJobHostOptionsMonitor.CurrentValue.ProfileWorkerRuntime;
+
+            if (string.IsNullOrEmpty(workerRuntime))
             {
-                workerRuntime = _environment.GetEnvironmentVariable(EnvironmentSettingNames.FunctionWorkerRuntime);
+                workerRuntime = _configuration[EnvironmentSettingNames.FunctionWorkerRuntime];
```
brettsam, line 48: Just double-checking... if this isn't present does it return null or throw?

brettsam (approved):

src/WebJobs.Script/Workers/ScriptHostWorkerRuntimeResolver.cs
kshyju: It returns `null` if key is not present. Won't throw.

src/WebJobs.Script.Grpc/WorkerFunctionMetadataProvider.cs
```diff
@@ -54,7 +58,7 @@ public ImmutableDictionary<string, ImmutableArray<string>> FunctionErrors
 
         public async Task<FunctionMetadataResult> GetFunctionMetadataAsync(IEnumerable<RpcWorkerConfig> workerConfigs, bool forceRefresh)
         {
-            _workerRuntime = _environment.GetEnvironmentVariable(EnvironmentSettingNames.FunctionWorkerRuntime);
+            _workerRuntime = _workerRuntimeResolver.GetWorkerRuntime();
```
satvu, line 61: Just curious - why is `_workerRuntime` re-calculated multiple times in this file (instead of using the stored value in constructor line 51)? Do we expect the value to change?

satvu (approved):

src/WebJobs.Script.Grpc/WorkerFunctionMetadataProvider.cs
kshyju: WorkerFunctionMetadataProvider is created in webhost scope. This method gets called after other life cycle events of script host such as specialization and we want to get the latest data at that time.

src/WebJobs.Script.Grpc/Rpc/FunctionRegistration/RpcFunctionInvocationDispatcher.cs
kshyju: Checked for "Will shutdown all the worker channels" message in detector repo on main as of 3/18 and no entries found.

PR #11527, opened 2025-12-16
Log siteUpdateId during initialization

kamperiadis:

As part of the rolling updates feature, customers would like to track how their rolling update is progressing (i.e. how many v1 vs v2 instances they have up and running). In order to achieve this, we want to add a log during initialization that will indicate what version the worker is running on. In Legion, this systemVersion attribute is sent as a long representation of the MAX(configLastModifiedTime, contentLastModifiedTime) where configLastModifiedTime is the last timestamp when the site configuration was updated, and contentLastModifiedTime is the last timestamp when the code deployment was triggered for the app. 

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
@@ -364,7 +364,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics.Extensions
         public static void LogHostInitializationSettings(this ILogger logger, string originalFunctionWorkerRuntime, string functionWorkerRuntime,
             string originalFunctionWorkerRuntimeVersion, string functionsWorkerRuntimeVersion, string functionExtensionVersion, string hostDirectory,
             bool inStandbyMode, bool hasBeenSpecialized, bool usePlaceholderDotNetIsolated, string websiteSku, string featureFlags,
-            IDictionary<string, string> hostingConfig, string hisMode, bool adminIsolationEnabled)
+            IDictionary<string, string> hostingConfig, string hisMode, bool adminIsolationEnabled, string functionsSiteUpdateId)
         {
             // This is a dump of values for telemetry right now, but eventually we will refactor this
             // into a proper Options object
@@ -383,7 +383,8 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics.Extensions
                 FeatureFlags = featureFlags,
                 HostingConfig = hostingConfig,
                 HISMode = hisMode,
-                AdminIsolationEnabled = adminIsolationEnabled
+                AdminIsolationEnabled = adminIsolationEnabled,
+                FunctionsSiteUpdateId = functionsSiteUpdateId
             };
 
             var options = new JsonSerializerOptions { WriteIndented = true };
--- a/src/WebJobs.Script.WebHost/Models/HostAssignmentContext.cs
+++ b/src/WebJobs.Script.WebHost/Models/HostAssignmentContext.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
@@ -23,6 +23,9 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Models
 
         [JsonProperty("lastModifiedTime")]
         public DateTime LastModifiedTime { get; set; }
+
+        [JsonProperty("siteUpdateId")]
+        public long SiteUpdateId { get; set; }
 
         [JsonProperty("MSISpecializationPayload")]
         public MSIContext MSIContext { get; set; }
@@ -164,6 +167,8 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Models
                 }
                 environment.SetEnvironmentVariable(EnvironmentSettingNames.EasyAuthClientId, EasyAuthSettings.SiteAuthClientId);
             }
+
+            environment.SetEnvironmentVariable(EnvironmentSettingNames.FunctionsSiteUpdateId, SiteUpdateId.ToString());
         }
     }
 }
--- a/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
+++ b/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
@@ -816,6 +816,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
             var featureFlags = environment.GetEnvironmentVariable(AzureWebJobsFeatureFlags);
             var hostingConfigDict = hostingConfigOptions.Features;
             var adminIsolationEnabled = environment.IsAdminIsolationEnabled();
+            var functionsSiteUpdateId = environment.GetFunctionsSiteUpdateId();
 
             string hisMode = "Disabled";
             if (FeatureFlags.IsEnabled(ScriptConstants.FeatureFlagStrictHISModeEnabled))
@@ -832,7 +833,8 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
             }
 
             logger.LogHostInitializationSettings(originalFunctionsWorkerRuntime, functionWorkerRuntime, originalFunctionsWorkerRuntimeVersion, functionWorkerRuntimeVersion,
-                functionExtensionVersion, currentDirectory, inStandbyMode, hasBeenSpecialized, usePlaceholderDotNetIsolated, websiteSku, featureFlags, hostingConfigDict, hisMode, adminIsolationEnabled);
+                functionExtensionVersion, currentDirectory, inStandbyMode, hasBeenSpecialized, usePlaceholderDotNetIsolated, websiteSku, featureFlags, hostingConfigDict, hisMode, adminIsolationEnabled,
+                functionsSiteUpdateId);
         }
 
         private void OnHostHealthCheckTimer(object state)
--- a/src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
+++ b/src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
@@ -35,6 +35,7 @@ namespace Microsoft.Azure.WebJobs.Script
         public const string AppInsightsAgent = "APPLICATIONINSIGHTS_ENABLE_AGENT";
         public const string FunctionsExtensionVersion = "FUNCTIONS_EXTENSION_VERSION";
         public const string FunctionWorkerRuntime = "FUNCTIONS_WORKER_RUNTIME";
+        public const string FunctionsSiteUpdateId = "FUNCTIONS_SITE_UPDATE_ID";
         public const string WorkerProbingPaths = "WORKER_PROBING_PATHS";
         public const string ContainerName = "CONTAINER_NAME";
         public const string WebsitePodName = "WEBSITE_POD_NAME";
--- a/release_notes.md
+++ b/release_notes.md
@@ -6,3 +6,4 @@
 - Add JitTrace Files for v4.1046
 - Update Java Worker Version to [2.19.4](https://github.com/Azure/azure-functions-java-worker/releases/tag/2.19.4)
 - Add area & error code to script & web host health checks (#11552)
+- Log siteUpdateId during initialization to track rolling update progress (#11527)
--- a/src/WebJobs.Script/Environment/EnvironmentExtensions.cs
+++ b/src/WebJobs.Script/Environment/EnvironmentExtensions.cs
@@ -653,6 +653,11 @@ namespace Microsoft.Azure.WebJobs.Script
             return environment.GetEnvironmentVariableOrDefault(FunctionWorkerRuntime, string.Empty);
         }
 
+        public static string GetFunctionsSiteUpdateId(this IEnvironment environment)
+        {
+            return environment.GetEnvironmentVariableOrDefault(FunctionsSiteUpdateId, string.Empty);
+        }
+
         /// <summary>
         /// Gets a value indicating whether AzureFileShare should be mounted when specializing Linux Consumption workers.
         /// </summary>
--- a/test/WebJobs.Script.Tests/WebJobsScriptHostServiceTests.cs
+++ b/test/WebJobs.Script.Tests/WebJobsScriptHostServiceTests.cs
@@ -595,6 +595,7 @@ namespace Microsoft.Azure.WebJobs.Script.Tests
             string functionExtensionVersion = "~4";
             string websiteSku = "Dynamic";
             string featureFlags = "EnableWorkerIndexing";
+            string functionsSiteUpdateId = "1767996105";
 
             mockEnvironment.Setup(e => e.GetEnvironmentVariable(EnvironmentSettingNames.FunctionWorkerRuntime)).Returns(functionWorkerRuntime);
             mockEnvironment.Setup(e => e.GetEnvironmentVariable(RpcWorkerConstants.FunctionWorkerRuntimeVersionSettingName)).Returns(functionsWorkerRuntimeVersion);
@@ -603,6 +604,7 @@ namespace Microsoft.Azure.WebJobs.Script.Tests
             mockEnvironment.Setup(e => e.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebJobsFeatureFlags)).Returns(featureFlags);
             mockEnvironment.Setup(e => e.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebsiteUsePlaceholderDotNetIsolated)).Returns("1");
             mockEnvironment.Setup(e => e.GetEnvironmentVariable(EnvironmentSettingNames.FunctionsAdminIsolationEnabled)).Returns("1");
+            mockEnvironment.Setup(e => e.GetEnvironmentVariable(EnvironmentSettingNames.FunctionsSiteUpdateId)).Returns(functionsSiteUpdateId);
 
             mockScriptWebHostEnvironment.Setup(e => e.InStandbyMode).Returns(false);
 
@@ -630,6 +632,7 @@ namespace Microsoft.Azure.WebJobs.Script.Tests
             Assert.True(result.GetProperty("UsePlaceholderDotNetIsolated").GetBoolean());
             Assert.Equal(websiteSku, result.GetProperty("WebSiteSku").GetString());
             Assert.Equal(featureFlags, result.GetProperty("FeatureFlags").GetString());
+            Assert.Equal(functionsSiteUpdateId, result.GetProperty("FunctionsSiteUpdateId").GetString());
 
             var hostingConfig = result.GetProperty("HostingConfig");
             Assert.Equal("TestValue1", hostingConfig.GetProperty("TestFeature1").GetString());


jviau:
Can you expand on what `systemVersion` represents exactly? Is this the customer's app payload version or does it represent some platform components version? How was the naming `systemVersion` settled upon?

kamperiadis:
Thank you for your response @jviau `systemVersion` represents the customer's app payload version (i.e. the UUIDv7 version of MAX(latest_timestamp_site_config_update, latest_timestamp_ code_content_update). It is not to represent platform components version. The naming of `systemVersion` was chosen to mean a platform-managed version that is automatically incremented for the customer whenever there is a code or site config change. As opposed to `appVersion` (vision for the future), which will be a user-defined version. But I do think that the name `systemVersion` is a bit confusing so I am open to change the name.

src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
```diff
@@ -204,6 +204,12 @@ public static class ScriptHostServiceLoggerExtension
                 new EventId(531, nameof(RequestAborted)),
                 "The request was aborted by the client (requestId: '{mS_ActivityId}').");
 
+        private static readonly Action<ILogger, string, Exception> _logSystemVersion =
+            LoggerMessage.Define<string>(
+                LogLevel.Information,
+                new EventId(532, nameof(LogSystemVersion)),
+                "Host is running on version '{systemVersion}'.");
```
RohitRanjanMS: 'systemVersion' is not the host runtime version. We have to rephrase the message.

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -798,6 +798,10 @@ private void LogInitialization(IHost host, bool isOffline, int attemptCount, int
                 logger.InStandByMode(operationId);
             }
 
+            // Log system version for rolling update tracking
+            var systemVersion = _environment.GetSystemVersion();
```
RohitRanjanMS: This is only applicable for Flex, add a condition so that we don't emit empty systemVersion for non-Flex skus.

src/WebJobs.Script.WebHost/Models/HostAssignmentContext.cs
```diff
@@ -164,6 +167,12 @@ public void ApplyAppSettings(IEnvironment environment, ILogger logger)
                 }
                 environment.SetEnvironmentVariable(EnvironmentSettingNames.EasyAuthClientId, EasyAuthSettings.SiteAuthClientId);
             }
+
+            if (SystemVersion != Guid.Empty)
+            {
+                logger.LogInformation("ApplyAppSettings is adding systemVersion {SystemVersion}", SystemVersion);
```
RohitRanjanMS: Do we need this log?

RohitRanjanMS (changes requested):

src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
FinVamp1: Did we talk about if we could have this in its own ColumnName or just in the Message to avoid having to parse the FunctionsLogs?

src/WebJobs.Script.WebHost/Models/HostAssignmentContext.cs
kamperiadis: Removed as not strictly necessary

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
kamperiadis: Done

jviau:
Yeah, not sure I like `systemVersion`. It sounds too much like a platform version. Some ideas: `deploymentId`, `deploymentRevision`, `payloadId`, `payloadRevision`. `appPayloadId`, `appPayloadRevision`, `appContentsId`, or `appContentsRevision`. 

I like `deploymentId` (and documenting it is UUIDv7, so orderable). But I get that this represents deployment contents + site config. But it could still work. If there are objections, then maybe `appContentsId` or `appContentsRevision` then?

jviau:
Another question: will this ID also incorporate host version? As in, a new ID will be used when host version is updated. I could see this going either way.

kamperiadis:
> Yeah, not sure I like systemVersion. It sounds too much like a platform version. Some ideas: deploymentId, deploymentRevision, payloadId, payloadRevision. appPayloadId, appPayloadRevision, appContentsId, or appContentsRevision.

> I like deploymentId (and documenting it is UUIDv7, so orderable). But I get that this represents deployment contents + site config. But it could still work. If there are objections, then maybe appContentsId or appContentsRevision then?

These are good suggestions. Thank you @jviau. I've passed them along to Sam so he can weigh in on which option will best fit customers' expectations (for context, this will also become a column name in the requests table in App Insights, potentially under customDimensions). 

> Another question: will this ID also incorporate host version? As in, a new ID will be used when host version is updated. I could see this going either way.

It is not meant to be updated when the host version is updated. It is just meant to track the site_config+code_content that the instance is running on. So, unless the customer updated the app's site configs or deployed to the app, this value will remain the same even after a host version update.

src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
kamperiadis: > 'systemVersion' is not the host runtime version. We have to rephrase the message.

@RohitRanjanMS I updated the message. @FinVamp1, @im-samz can you confirm if this logging message looks good to you or if you have suggestions for the message? 

> Did we talk about if we could have this in its own ColumnName or just in the Message to avoid having to parse the FunctionsLogs?

@FinVamp1 Since this is meant to be a log (instead of a column in the logs), users will have to search for this specific message and then parse it in the logs. But aside from the log changes in this PR, Rohit will do separate work to have it be its own column in the requests table (potentially under customDimensions), and this will not require having to search and parse a specific message. To provide you more context on why we decided to go with this approach, we discussed that adding it as a column for all logs means additional cost for storage to keep this telemetry. So instead, we will only have this single log when the instance is initialized, and it will also be a dimension for the requests table since requests is another critical area where we think customers will want to see the version. TLDR: instead of adding it as it as a column for all logs, we have agreed on the critical areas where customers would want to see this information.

RohitRanjanMS:
> Yeah, not sure I like `systemVersion`. It sounds too much like a platform version. Some ideas: `deploymentId`, `deploymentRevision`, `payloadId`, `payloadRevision`. `appPayloadId`, `appPayloadRevision`, `appContentsId`, or `appContentsRevision`.
> 
> I like `deploymentId` (and documenting it is UUIDv7, so orderable). But I get that this represents deployment contents + site config. But it could still work. If there are objections, then maybe `appContentsId` or `appContentsRevision` then?

Let's use `deploymentId`. We do have deployment.id resource attribute - https://opentelemetry.io/docs/specs/semconv/registry/attributes/deployment/#deployment-id:~:text=staging%3B%20production-,deployment.id,-string

FinVamp1:
> > Yeah, not sure I like `systemVersion`. It sounds too much like a platform version. Some ideas: `deploymentId`, `deploymentRevision`, `payloadId`, `payloadRevision`. `appPayloadId`, `appPayloadRevision`, `appContentsId`, or `appContentsRevision`.
> > I like `deploymentId` (and documenting it is UUIDv7, so orderable). But I get that this represents deployment contents + site config. But it could still work. If there are objections, then maybe `appContentsId` or `appContentsRevision` then?
> 
> Let's use `deploymentId`. We do have deployment.id resource attribute - https://opentelemetry.io/docs/specs/semconv/registry/attributes/deployment/#deployment-id:~:text=staging%3B%20production-,deployment.id,-string

Kudu already has a deploymentId property which is a GUID for each unique deployment. Do we risk overloading the term?

jviau:
@FinVamp1 do you have some docs on the Kudu deployment ID? How is it exposed to customers today, if at all? Can that Kudu deployment ID be updated to use the same value as this PR for Flex apps? If we can unify these two IDs that would be awesome.

FinVamp1:
> @FinVamp1 do you have some docs on the Kudu deployment ID? How is it exposed to customers today, if at all? Can that Kudu deployment ID be updated to use the same value as this PR for Flex apps? If we can unify these two IDs that would be awesome.

Thanks @jviau the deploymentId isn't a GUID after all but a string. I know code isn't documentation but here's a sense of how it works. 

https://github.com/projectkudu/kudu/blob/7e7e310dee7e9e6488a87a20e99f2813b2f74808/Kudu.Core/Deployment/DeploymentManager.cs#L43

This deployment ID gets exposed to customers in a number of places.

1) Azure DevOps logs
2) Github Actions Logs
3) ApplicationInsights Logs 
4) Files on the Kudu file system for the other SKU's that have a Kudu site browse experience. 

This version value related to this PR is for both code and site config changes whereas the deploymentId has historically only been for code changes. So it might be considered a breaking change to re-purpose it here.

im-samz:
ACK'ing on this thread. I'm still reasoning over the names for our platform-managed and user-defined versions. The deployment commitments we have in 2026 all require customers to reason about the state of an app over time, so the vocabulary decisions we make will either make the deployment story feel coherent or permanently confusing.

Thinking out loud here: 

"version" in Functions is already an overloaded term...Given that the two proposed "versions" serve different purposes, I wonder if we should escape the use of "versions" altogether. Imagine hitting CTRL-F for "version" in our learn docs that explain how to use rolling updates with rollbacks; reasoning about different types of versions with only an adjective differentiating them might be prone to misreading.

My hunch so far is:
- platform-managed == `appRevisionId`
  - "app" to clarify scope: the entire app and not some component of it. From the customer perspective, this is their app, but from our perspective, this is their deployment or payload we are specializing with.
  - "revision" to clarify type: it feels more _neutral_ than "version" and gets us away from overloading the term.
  - "Id" to clarify format: UUIDv7s feel more like IDs imo
- user-defined == `appRelease`
  - "app" to clarify scope
  - "release" to clarify type: hints at the nature of each release being intentionally created
  - no ID because the the format _could_ be a freeform string

Other pairs:
- {appRevision; appRelease}
  - if we feel "Id" is unecessary and we want the terms to be parallel
- {appRevisionId; appReleaseId}
  - if we feel like this is more clear
- {appVersion; appRelease}
  - Version still fundamentally conveys a difference in state; Release makes it more obvious that they are customer instantiated
- {appGeneration; appRelease}
  - Generation is very neutral, even mechanical, doubling down on the fact that it is platform-managed. K8s uses this.

jviau:
> Thanks @jviau the deploymentId isn't a GUID after all but a string. I know code isn't documentation but here's a sense of how it works.

Guid vs string isn't that big of a deal. GUIDs are often represented as strings.

> https://github.com/projectkudu/kudu/blob/7e7e310dee7e9e6488a87a20e99f2813b2f74808/Kudu.Core/Deployment/DeploymentManager.cs#L43
> 
> This deployment ID gets exposed to customers in a number of places.
> 
> 1. Azure DevOps logs
> 2. Github Actions Logs
> 3. ApplicationInsights Logs
> 4. Files on the Kudu file system for the other SKU's that have a Kudu site browse experience.
> 
> This version value related to this PR is for both code and site config changes whereas the deploymentId has historically only been for code changes. So it might be considered a breaking change to re-purpose it here.

How does this show up in these logs? Do you have an exact log example you can share? And does it show up in Flex sku specifically?

jviau:
Another point on the site config changes influencing this value: will this detect changes when a config value is pulled in via key vault or azure app config?

FinVamp1:
> > Thanks @jviau the deploymentId isn't a GUID after all but a string. I know code isn't documentation but here's a sense of how it works.
> 
> Guid vs string isn't that big of a deal. GUIDs are often represented as strings.
> 
> > https://github.com/projectkudu/kudu/blob/7e7e310dee7e9e6488a87a20e99f2813b2f74808/Kudu.Core/Deployment/DeploymentManager.cs#L43
> > This deployment ID gets exposed to customers in a number of places.
> > 
> > 1. Azure DevOps logs
> > 2. Github Actions Logs
> > 3. ApplicationInsights Logs
> > 4. Files on the Kudu file system for the other SKU's that have a Kudu site browse experience.
> > 
> > This version value related to this PR is for both code and site config changes whereas the deploymentId has historically only been for code changes. So it might be considered a breaking change to re-purpose it here.
> 
> How does this show up in these logs? Do you have an exact log example you can share? And does it show up in Flex sku specifically?

Is there something specific you want to know about the Id? I have outlined examples below.

For Flex we log the Deployment details to ApplicationInsights if it is configured. 

image

For Azure DevOps here's an example of where it is seen 

image

with additional information if you have deployed with the DevOps Debug Flag turned on.

image

For DevOps we will also look the deployment logs to Application Insights if the resource that we deploy to has it enabled. 

For Linux Dedicated the Kudu logs look like this. 

image

In the Linux Dedicated NewUI it looks like this

image

For Flex with OneDeploy you would see something like this


Run Azure Functions Action
Package deployment using One Deploy initiated.
Deploy logs can be viewed at https://<FUNCTION_APP_NAME>.scm.azurewebsites.net/api/deployments/<DEPLOYMENT_ID>/log
Successfully deployed web package to Function App.


From the CLI you can get the Deployment list and ID's


# List all deployment records for a Function App
az functionapp log deployment list \
  --name <FUNCTION_APP_NAME> \
  --resource-group <RESOURCE_GROUP>

# Show logs for the latest deployment (or a specific one by ID)
az functionapp log deployment show \
  --name <FUNCTION_APP_NAME> \
  --resource-group <RESOURCE_GROUP> \
  --deployment-id <DEPLOYMENT_ID>

For Windows Dedicated here's how it looks in Deployment Center

image

And in Kudu

image

And in the Kudu File System

image

Windows Consumption looks the same as dedicated.

image

So the Kudu Logs will look the same for Windows Consumption.

Linux Consumption doesn't have a Kudu instance that persists but the deploymentId is still logged.

kamperiadis:
Thank you for your input @im-samz @FinVamp1 

@jviau What do you think of using `appRevisionId` or `appRevision` as per Sam's suggestion (or the other options that Sam proposed)? 

> Another point on the site config changes influencing this value: will this detect changes when a config value is pulled in via key vault or azure app config?

Whenever the siteLastModifiedTime is updated, we will update the site config timestamp that backs this value. My understanding is that if the customer updates that value in key vault or azure app config, we do not update siteLastModifiedTime on our end since we are just pulling the updated value from the same source that was previously configured. Is that correct? If so, then to answer your question, we would not detect changes from updating values in key vault or azure app config. But I would say that this is the intended behavior since we are only trying to track the updates made to the Site resource that cause your instances to be recycled.

jviau:
After @RohitRanjanMS shared that OTel link, I saw that `app` terminology in OTel is primarily used for client-side applications. Like a desktop or phone app. Maybe `siteRevision` or `siteRevisionId`?

I do really like `deploymentId` still, unfortunate that kudu already leverages it.

FinVamp1:
> After @RohitRanjanMS shared that OTel link, I saw that `app` terminology in OTel is primarily used for client-side applications. Like a desktop or phone app. Maybe `siteRevision` or `siteRevisionId`?
> 
> I do really like `deploymentId` still, unfortunate that kudu already leverages it.

Deployment is already an overloaded term in terms of Functions Deployment. Customers use it interchangeably to mean app create/delete, code deployment , site configuration changes and also v1 to v2 even if it's a new app. So hopefully we are in agreement that using deploymentId is out of scope for this discussion.

kamperiadis:
To keep this thread updated, @im-samz is still deciding if we can use `siteRevision` or `siteRevisionId`. He will add his insights on this once ready. Thanks everyone.

im-samz:
Hey team! I'm learning towards `stateId` now. What do you think?

OTel has an attribute [faas.version](https://opentelemetry.io/docs/specs/semconv/registry/attributes/faas/#faas-version); the equivalent in Lambda is **[version](https://docs.aws.amazon.com/lambda/latest/dg/configuration-versions.html?icmpid=docs_console_unmapped)** and in GCP it's **[revision](https://docs.cloud.google.com/run/docs/managing/revisions)**. Both of these are immutable snapshots of an app that can be audited and routed traffic to.

What we're introducing here isn't equivalent either to Lamba's versions or GCP's revisions; what we have here is just a unique identifier. Hence, I want to avoid the term "revision" or "version" since those terms already have connotations in the industry. In fact, I want to avoid introducing another concept at all. Moreover, we are introducing a user-defined version, so I want to reserve both the term and the OTel attribute for it: `versionId` and/or `versionName`. 

For this reason, I'm now learning towards `stateId` since it (1) doesn't introduce another concept and (2) is intuitive enough that someone new to Azure/Functions would grasp (for example, "`site`" feels Azure-y).

---

For a sanity check, this is essentially what we'd say in our documentation without the terms:
> Rolling updates replace your app incrementally, which can result in **multiple runtime states** running at the same time.
Each runtime state represents a distinct snapshot of your app’s code and configuration. During the update, traffic may be served by instances running different states until the rollout completes.
>
> To understand which state an instance is running, you can inspect **a platform-assigned identifier that distinguishes one state from another**. Instances with the same **identifier** are running the same code and configuration.
>
> If you need the ability to route traffic to a known state or roll back after an update, you can create **an immutable snapshot of your app** before starting the rollout. This snapshot can later be reactivated to restore the app to a previously known state.

This is what it would look like when using our terms:
> Rolling updates replace your app incrementally, which can result in **multiple runtime states** running at the same time.
Each runtime state represents a distinct snapshot of your app’s code and configuration. During the update, traffic may be served by instances running different states until the rollout completes.
>
> To understand which state an instance is running, you can inspect **its `stateId`**. Instances with the same **`stateId`** are running the same code and configuration.
>
> If you need the ability to route traffic to a known state or roll back after an update, you can create **a version** before starting the rollout. This snapshot can later be reactivated to restore the app to a previously known state.

src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
```diff
@@ -204,6 +204,12 @@ public static class ScriptHostServiceLoggerExtension
                 new EventId(531, nameof(RequestAborted)),
                 "The request was aborted by the client (requestId: '{mS_ActivityId}').");
 
+        private static readonly Action<ILogger, string, Exception> _logFunctionAppVersion =
+            LoggerMessage.Define<string>(
+                LogLevel.Information,
+                new EventId(532, nameof(LogFunctionAppVersion)),
+                "Running following version from app configuration update or code content deployment: '{functionAppVersion}'.");
```
im-samz:
```suggestion
                "Running app with code and config state '{functionAppVersion}'.");
```

im-samz: Framing this field less like a "version" and more like an "identifier".

jviau: I do worry that using terms like `config state` is a bit misleading. Customers may have _more_ config state than what we are including in the calculation here.

Will this identifier detect changes to app settings? How about when app settings point to a key vault or azure app config, will this know when those dereferenced values change?

Also, will customers assume this is the complete config state for their own worker process? Because it isn't, they can be loading config state from wherever, it's a black box to us.

RohitRanjanMS: Since this is structured logging, `functionAppVersion` will be recorded as a custom dimension. We need both a clear log message and the attribute name. How about `deployment.id` but scoping it to Azure Functions, for example `azure.functions.deployment.id`

FinVamp1: Can we agree on appRevisionId or do we need to use deploymentId for Otel reasons?

im-samz: @jviau - this identifier will detect changes to app settings. But if the app setting is pointing to a KV or App Config reference, and the key is rotated, there would be no update. The behavior would be similar to ACA's revisions, which makes no promise to create a new revision when the referenced value is changed - only when the value of the "secret" (e.g., app setting pointer to KV) is changed.

---

@FinVamp1 - `appRevisionId` or `revisionId` is a candidate imo only if we have plans to layer on a UX. For example, list my historical revisionIds. Otherwise, I feel like we will fail to meet expectations since a "revision" already has connotations in the industry (re: ACA Revisions; GCP Cloud Run Revisions).

im-samz: After spending (too long) thinking about this, this is my recommendation to close the loop.

**Key consideration**
This field shows up in startup logs and telemetry and will be used by customers to reason about runtime behavior. Once shipped, it becomes a semantic contract. We want to avoid introducing a name that implies guarantees, completeness, or future capabilities that the platform does not actually provide.

**Recommendation**
- Field name: `siteUpdateId`
- Startup log message: `"Running app for site update '{UUIDv7}'"`

**Why `siteUpdateId`?**
1. It frames the field as a diagnostic correlation id, not a first-class product concept. 
    - Introducing a concept like `revisionId` implies product capabilities like lifecycle, history, diffing, and restoring against the revision artifact - which we have not committed to, but would be expected.

2. It scopes the identifier to the platform operation that produced the runtime, without implying the app's full effective runtime configuration. 
    - I agree with @jviau that `stateId` can be initially misinterpreted to imply the latter. We could tighten the docs to explain ourselves, but first-impressions matter, and I don't want to confuse customers.

3. It aligns more with reality than using `deploymentId`, which is currently scoped to code-only site updates as @FinVamp1 has mentioned. 
    - I agree with @RohitRanjanMS and @jviau's sentiment that `deploymentId` suits the scenario, if it had not already been used. Fortunately for us, [we have documentation](https://learn.microsoft.com/en-us/azure/azure-functions/flex-consumption-site-updates?tabs=Bicep) to describe the **site update** concept; site updates are clearly positioned to be the parent operation to a code deployment or configuration change (app setting or site property).  

`siteUpdateId` does sound mechanical, but for the intention of it being a diagnostic correlation id, I think it does the job to scope the semantics without overpromising anything. Plus, I suspect customers will only find out about it after visiting our documentation describing what site updates are: [Site update strategies in Flex Consumption > Monitoring site updates](https://learn.microsoft.com/en-us/azure/azure-functions/flex-consumption-site-updates?tabs=Bicep#monitoring-site-updates).

**Future considerations**
Longer-term, we may introduce a first-class Revision concept once there is meaningful UX behind it (e.g., listing revisions, comparing changes, restoring a prior revision). We use "version" loosely in our docs, so introducing concepts to clean up our story does make sense. At the point we decide to do so, `siteUpdateId` can remain a low-level diagnostic artifact or be subsumed behind a higher-level revision model.

For now, `siteUpdateId` gives us precise, honest semantics that match the current product surface while leaving room to iterate cleanly in the future.

kamperiadis: Thank you @im-samz. 

@jviau, @RohitRanjanMS, @FinVamp1 - what do you think of Sam's recommendation? Are you ok moving forward with this? 

> Field name: siteUpdateId
> Startup log message: "Running app for site update '{UUIDv7}'"

FinVamp1 (approved):

src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
RohitRanjanMS: I’m fine with using siteUpdateId; the only issue is that it doesn’t follow the standard naming conventions. We should also scope it under the functions namespace so it’s clear where the attribute belongs. I know `azure.functions.site.update.id` is a bit long, but it does align with the OpenTelemetry guidelines and keeps our telemetry consistent and future‑proof.

kamperiadis: Thanks @RohitRanjanMS Can you please expand on how that name doesn't follow the standard naming conventions?

kamperiadis: Also, quick question just for my own curiosity - would `azure.functions.siteUpdateId` not align with OpenTelemetry guideline?

RohitRanjanMS: OpenTelemetry semantic conventions:
- all lowercase
- dot‑separated namespaces
- no underscores
- no camelCase

jviau: nit:
> no underscores

You do use underscores when a dot doesn't make sense. Primarily separating multi-words sections of a namespace. IE: `http.response.status_code` and not `http.response.status.code`, since 'status' is not a namespace.

As such, I would call it `azure.functions.site_update_id`. Unless we think `site` is worthy of its own namespace, then it would be `azure.functions.site.update_id`.

I could see `site` being a namespace. Or something to convey "these are Azure functions attributes related to the _customer_ payload/config (as opposed to platform)". `azure.functions.site.` or `azure.functions.app.` or any other suggestions?

kamperiadis: Thank you @jviau. @RohitRanjanMS what do you think of @jviau's comment? 

Also, scoping it back to this PR and putting aside the OpenTelemetry name that Rohit will use in a separate PR, are we ok with using `siteUpdateId` in this PR? Just looking to see if we can get this PR approved and merged if it looks good. Thank you.

RohitRanjanMS: It seems the original convention used underscores (_), but later it was changed to dots (.). For example, deployment attributes now follow `deployment.id`. 
However, I’m [adding ](https://github.com/Azure/azure-functions-host/pull/11506/files#diff-e75d1ba53d9a443454db31c59076865e8a5d2b400858a386559786e30d9a61d3:~:text=attributes.Add(new(ResourceSemConventions.SiteUpdateId%2C%20appVersion))%3B)this attribute to all telemetry emitted by the host—logs, requests, and exceptions—so this log is no longer needed.

jviau: @RohitRanjanMS it is subjective. In this case, they believe `deployment` is a namespace and will have multiple deployment-related attributes, as such they separate with `.`. But look at http spec: https://opentelemetry.io/docs/specs/semconv/registry/attributes/http/, there are multiple underscore separated attributes. I believe it is actually camel casing (or pascal) which is **not** used in otel attributes. so `azure.functions.siteUpdateId` would be wrong.

To me, `update` is definitely not its own namespace. So `update_id` is a given here. But it is open to discussion if `site` will be a namespace or not. So, will it be `azure.functions.site.update_id` or `azure.functions.site_update_id`? I think I like the former more (site.update_id).

Also I do want to say I like using "update id" here over "deployment id". I think that does better capture this is deployments + setting updates.

src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
```diff
@@ -35,7 +35,7 @@ public static class EnvironmentSettingNames
         public const string AppInsightsAgent = "APPLICATIONINSIGHTS_ENABLE_AGENT";
         public const string FunctionsExtensionVersion = "FUNCTIONS_EXTENSION_VERSION";
         public const string FunctionWorkerRuntime = "FUNCTIONS_WORKER_RUNTIME";
-        public const string FunctionAppVersion = "FUNCTION_APP_VERSION";
+        public const string SiteUpdateId = "SITE_UPDATE_ID";
```
jviau: I think we should start standardizing our environment names with namespaces, much like how we are discussing namespacing the otel attributes. This will reduce chance of collision with a customer's app setting.

`FUNCTIONS_SITE_UPDATE_ID` -- or some other namespace would be good.

src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
im-samz: @RohitRanjanMS - I can go either way on adding the log; thanks for getting the attribute into all those tables.
- Keeping it is a **nice-to-have** since it explains what the field is.
- However, I can see a case where we should keep startup logs to only what is necessary to mitigate noise. Do we have a precedent of the host emitting logs to call out attributes that are being logged?
  - @jviau, do you think the host team has a preference regarding this?

---
Re: is site its own namespace?
- Would making `site` or `app` its own namespace be redundant? I'm not familiar with all the OTell attributes, but I assume that the attributes that pertain to the app as a whole fall under the `azure.functions` namespace. 
- Colloquially, based on the language in our [docs](https://learn.microsoft.com/en-us/azure/azure-functions/flex-consumption-site-updates?tabs=Bicep), we've been using "site updates" (i.e., combined namspace) and "updates" (i.e., separate namespace) interchangeably.

jviau: @im-samz is there an existing startup log this can be added to? How about the log statement that was also just recently updated: https://github.com/Azure/azure-functions-host/pull/11547

> but I assume that the attributes that pertain to the app as a whole fall under the azure.functions namespace.

Yes, we will include `azure.functions.` namespace in front. But we can also namespace further underneath as it fits. My vote is for `azure.functions.site.update_id`. But I can be convinced otherwise. I don't think having it as a namespace will really have any negatives.

src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
kamperiadis: Fixed

src/WebJobs.Script.WebHost/Diagnostics/Extensions/ScriptHostServiceLoggerExtension.cs
kamperiadis: @jviau I added onto that log statement you linked instead of creating a new log statement. Thank you. Can you please review the PR again and approve if it all looks good to you? Thank you!

jviau (approved):
Approved from my end. But please wait for @RohitRanjanMS to approve as well so he can get a look. I think adding to the existing log statement will address Rohit's concern, but want to make sure.

The open discussion isn't relevant to this PR, but to a later resource update Rohit will be making.

RohitRanjanMS:
I’m aligned with this proposal. My only concern was ending up with the same site update ID added twice under different attribute names, but since we’ll rely only on the resource attribute, that won’t be an issue.

RohitRanjanMS (approved):

PR #11538, opened 2025-12-31
Reduce instances to 3 when generating JIT traces to prevent YAML size failures

Francisco-Gamino:

This PR optimizes the cold start pipeline to automatically use a reduced number of instances (3 instead of 20) when generating JIT traces, preventing pipeline failures caused by excessively large YAML files.

**Problem**
The pipeline currently uses 20 instances from `coldstart.yml` for all runs. When generating JIT traces for releases with `createJitTrace: true`, this creates very large YAML files that exceed size limitations and cause pipeline failures.

**Solution**
Introduced a variable `effectiveRunInstances` that automatically selects the appropriate instance count based on the `createJitTrace` parameter:
- When `createJitTrace: true` -> uses 3 instances (1,2,3) for JIT trace generation
- When `createJitTrace: false` -> uses 20 instances for comprehensive cold start data collection

### DIFF
--- a/eng/ci/host.coldstart.yml
+++ b/eng/ci/host.coldstart.yml
@@ -41,6 +41,10 @@ parameters:
   displayName: Create Jit trace (Ensure "Collect DotNet trace" is set to true)
   type: boolean
   default: false
+- name: useReducedInstancesForJitTrace
+  displayName: Use reduced instances (3) for JIT trace generation (reduces YAML size for releases)
+  type: boolean
+  default: false
 
 - name: collectPerfViewProfileOnWindows
   displayName: Collect PerfView Profile (on Windows)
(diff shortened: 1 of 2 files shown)


eng/ci/host.coldstart.yml
```diff
@@ -41,6 +41,10 @@ parameters:
   displayName: Create Jit trace (Ensure "Collect DotNet trace" is set to true)
   type: boolean
   default: false
+- name: useReducedInstancesForJitTrace
```
kshyju: I think we can remove this `useReducedInstancesForJitTrace` parameter and just check `eq(parameters.createJitTrace, true)` to set `effectiveRunInstances`

Francisco-Gamino: The code has been updated to use 3 instances to generate the JIT Trace file when the `createJitTrace` is set to `true`.

kshyju (approved):

amamounelsayed (approved):

RohitRanjanMS (approved):

PR #11582, opened 2026-01-26
improvements to LanguageWorkerOptions cache invalidation

### LINKED ISSUE #11535: Using custom worker arguments can cause worker process to start multiple times during specialization
Using custom arguments will incur a longer cold start b/c we cannot specialize. However, in some conditions (notably, when bundles are used), we will start the process twice causing longer cold start than required.

The issue seems to be:
1. We see args which make the app ineligible for specialization so we shut down the placeholder and issue restart request to JobHost.
2. Before the new JobHost starts... bundles need loaded... bundles need metadata... which starts up a new process and asks for metadata. But critically, at this point, there is no active JobHost, so [the new process doesn't see the args](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs#L193) (but it should... it just isn't looking in all the right places).
3. JobHost starts. This [triggers an "options reloaded" event](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.Grpc/Rpc/WebHostRpcWorkerChannelManager.cs#L55-L63), which causes the process to start again with the correct args.

[Internal query](https://dataexplorer.azure.com/clusters/wawsweu/databases/wawsprod?query=H4sIAAAAAAAAA32OQUvDMBiG7%2FsVYScHRr4mbdoNK3iYMHAy1p52%2B5Z%2BasUkI0ktA3%2B8KYMhHjy9l%2Fd5n%2FdpsDr2zoZn9xZm32x8J0%2Bs3WzXTfu43bGHmnUYKfaGbgSIgmeCg2yhXElYCXmXgSwrKA4Lhrb7xd3%2Fy6nEgRSVOiyuzvUX2bjzvUF%2FbiKa0wsaYnXN5iOOgZ%2B86zgawZUq51do7z5pY0NEqy9d4LzKNWWULTlUSZrrnPhxqTRXWhR5IRFeRT4tpMUP0vEibtPRMFlvWeMGrynlYKYvqel8R54dz3%2BqDIP%2BAfckGD5AAQAA)

brettsam:
Fixes #11535

There were bugs around exactly when we invalidate the `LanguageWorkerOptions` cache. This was causing multiple restarts in some scenarios.

We ultimately have two scenarios here:
- **using bundles** -> We start the worker very early -- while we're still building the `JobHost` -- in order to pull metadata (used to filter which triggers we need from bundles). Therefore, we need to invalidate `LanguageWorkerOptions` early -- before we start the worker. This is the only way to take App Settings into consideration. This is what the bug was hitting.
- **no bundles** -> We start the worker later and also support the JobHost supplying some `LanguageWorkerOptions` overrides. Therefore we need to invalidate `LanguageWorkerOptions` *after* we build the host, but before we start it.

The fix here introduces a `WorkerConfigCacheInvalidator` to make this behavior explicit. If we go down the bundles path, we know we're using bundles, so we invalidate early and do *not* invalidate later. And vice-versa.

There was also a bug where we'd registered two `LanguageWorkerOptionsSetup` services, meaning we were double-creating our `LanguageWorkerOptions` every time we needed them.

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
+++ b/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
@@ -394,7 +394,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
                         _ = Task.Run(() => deferredLogProvider.ProcessBufferedLogsAsync(selectedProviders));
                     }
                 }
-                _workerConfigCacheInvalidator.InvalidateCacheIfNotUsingBundles();
+                _workerConfigCacheInvalidator.InvalidateCachePostBuildIfEnabled();
 
                 var scriptHost = (ScriptHost)ActiveHost.Services.GetService<ScriptHost>();
                 if (scriptHost != null)
--- a/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
+++ b/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
@@ -20,15 +20,14 @@ using Microsoft.Azure.WebJobs.Host;
 using Microsoft.Azure.WebJobs.Logging;
 using Microsoft.Azure.WebJobs.Logging.ApplicationInsights;
 using Microsoft.Azure.WebJobs.Script.Config;
+using Microsoft.Azure.WebJobs.Script.Configuration;
 using Microsoft.Azure.WebJobs.Script.Diagnostics;
 using Microsoft.Azure.WebJobs.Script.Eventing;
 using Microsoft.Azure.WebJobs.Script.Metrics;
 using Microsoft.Azure.WebJobs.Script.Scale;
-using Microsoft.Azure.WebJobs.Script.WebHost.Configuration;
 using Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics;
 using Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics.Extensions;
 using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
-using Microsoft.Azure.WebJobs.Script.Workers.Rpc.Configuration;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
@@ -52,6 +51,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
         private readonly ILogger _logger;
         private readonly IEnvironment _environment;
         private readonly IMetricsLogger _metricsLogger;
+        private readonly WorkerConfigCacheInvalidator _workerConfigCacheInvalidator;
         private readonly HostPerformanceManager _performanceManager;
         private readonly IOptions<HostHealthMonitorOptions> _healthMonitorOptions;
         private readonly IConfiguration _config;
@@ -65,8 +65,6 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
         private readonly bool _originalStandbyModeValue;
         private readonly string _originalFunctionsWorkerRuntime;
         private readonly string _originalFunctionsWorkerRuntimeVersion;
-        private readonly IOptionsChangeTokenSource<WorkerConfigurationResolverOptions> _workerConfigResolverOptionsChangeTokenSource;
-        private readonly IOptionsChangeTokenSource<LanguageWorkerOptions> _languageWorkerOptionsChangeTokenSource;
 
         // we're only using this dictionary's keys so it acts as a "ConcurrentHashSet"
         private readonly ConcurrentDictionary<ScriptHostStartupOperation, byte> _activeStartupOperations = new();
@@ -89,8 +87,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
             HostPerformanceManager hostPerformanceManager, IOptions<HostHealthMonitorOptions> healthMonitorOptions,
             IMetricsLogger metricsLogger, IApplicationLifetime applicationLifetime, IConfiguration config, IScriptEventManager eventManager, IHostMetrics hostMetrics,
             IOptions<FunctionsHostingConfigOptions> hostingConfigOptions,
-            IOptionsChangeTokenSource<LanguageWorkerOptions> languageWorkerOptionsChangeTokenSource,
-            IOptionsChangeTokenSource<WorkerConfigurationResolverOptions> workerConfigResolverOptionsChangeTokenSource)
+            WorkerConfigCacheInvalidator workerConfigCacheInvalidator)
         {
             ArgumentNullException.ThrowIfNull(loggerFactory);
 
@@ -101,8 +98,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
             RegisterApplicationLifetimeEvents();
 
             _metricsLogger = metricsLogger;
-            _workerConfigResolverOptionsChangeTokenSource = workerConfigResolverOptionsChangeTokenSource ?? throw new ArgumentNullException(nameof(workerConfigResolverOptionsChangeTokenSource));
-            _languageWorkerOptionsChangeTokenSource = languageWorkerOptionsChangeTokenSource ?? throw new ArgumentNullException(nameof(languageWorkerOptionsChangeTokenSource));
+            _workerConfigCacheInvalidator = workerConfigCacheInvalidator ?? throw new ArgumentNullException(nameof(workerConfigCacheInvalidator));
             _applicationHostOptions = applicationHostOptions ?? throw new ArgumentNullException(nameof(applicationHostOptions));
             _scriptWebHostEnvironment = scriptWebHostEnvironment ?? throw new ArgumentNullException(nameof(scriptWebHostEnvironment));
             _scriptHostBuilder = scriptHostBuilder ?? throw new ArgumentNullException(nameof(scriptHostBuilder));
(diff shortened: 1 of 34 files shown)


src/WebJobs.Script/Config/WorkerConfigCacheInvalidator.cs
```diff
@@ -0,0 +1,63 @@
...
+    public void InvalidateCacheIfNotUsingBundles()
+    {
+        if (!_usingBundles)
+        {
+            InvalidateCache();
+        }
+
+        // Reset for future restarts
+        _usingBundles = false;
+    }
```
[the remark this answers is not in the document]
brettsam: this is called in a very deterministic way and thread-safety shouldn't be an issue

src/WebJobs.Script/Config/WorkerConfigCacheInvalidator.cs
```diff
@@ -0,0 +1,63 @@
...
+public class WorkerConfigCacheInvalidator
+{
+    private readonly IOptionsChangeTokenSource<WorkerConfigurationResolverOptions> _workerConfigResolverOptionsChangeTokenSource;
+    private readonly IOptionsChangeTokenSource<LanguageWorkerOptions> _languageWorkerOptionsChangeTokenSource;
+
+    private bool _usingBundles = false;
+    private bool _firstRun = true;
+
+    public WorkerConfigCacheInvalidator(
+        IOptionsChangeTokenSource<WorkerConfigurationResolverOptions> workerConfigResolverOptionsChangeTokenSource,
```
jviau, line 62: I am not a big fan of this 2-way coupling. For every options you want refreshed with this, you both need to register the `ConfigureOptionsWithChangeTokenSource` during setup **and** you need to add it to this service.

Is there a way we can make it only needing to do one of those? Preferably only needing to do the setup component, and then having this be able to discover all of them and refresh them all?

src/WebJobs.Script/Config/WorkerConfigCacheInvalidator.cs
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
+using Microsoft.Azure.WebJobs.Script.Workers.Rpc.Configuration;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.WebJobs.Script.Configuration;
+
+public class WorkerConfigCacheInvalidator
```
kshyju:
```suggestion
public sealed class WorkerConfigCacheInvalidator
```

src/WebJobs.Script/Config/WorkerConfigCacheInvalidator.cs
```diff
@@ -0,0 +1,63 @@
...
+    {
+        _usingBundles = true;
+
+        if (!_firstRun)
+        {
+            InvalidateCache();
+        }
+        else
+        {
+            _firstRun = false;
```
jviau: does this need to be outside of the `else` statement? This will never be set to `false` because the `InvalidateCache()` branch will always be chosen, never this branch.

src/WebJobs.Script/Config/WorkerConfigCacheInvalidator.cs
```diff
@@ -0,0 +1,63 @@
...
+        _languageWorkerOptionsChangeTokenSource = languageWorkerOptionsChangeTokenSource;
+    }
+
+    public void InvalidateCacheForBundles()
+    {
+        _usingBundles = true;
+
+        if (!_firstRun)
+        {
+            InvalidateCache();
```
jviau, line 99: Do we expect concurrent callers? If so, are we okay with duplicate `InvalidateCache()` calls? If not okay, putting this in a `lock` will help.

src/WebJobs.Script/Config/WorkerConfigCacheInvalidator.cs
brettsam: It was chosen on first run, then flipped to false. But I did remove the else branch. I didn't like it either.

src/WebJobs.Script/Config/WorkerConfigCacheInvalidator.cs
brettsam: This API itself isn't thread-safe but it's never being called concurrently. The only time this is used is during the ScriptHost building/startup inside of WebJobScriptHostService, which is all inside of a lock.

On top of that, multiple `InvalidatCache()` calls wouldn't be the end of the world -- we'd just have to re-build the cache resulting in some slight perf impact.

src/WebJobs.Script/Config/WorkerConfigCacheInvalidator.cs
brettsam: Yeah it's gross. The reason we couldn't use a traditional setup for Options invalidation is b/c we need this done in a strict order, which isn't guaranteed when triggering a ChangeTokenSource. 

Unfortunately the LanguageWorkerOptions depends on the WorkerConfigResolverOptions... so they have to be invalidated in reverse order or else you don't pick up the changes.

I think we really shouldn't even be using the Options pattern here with how complicated this relationship is. Part of my refactoring will address this in my feature branch. But because this is a targeted fix, I didn't want to go re-implementing all of this here.

I'm open to another pattern if you see one (that wouldn't require a bunch of changes?) -- is there anything you can think of that'd be smoother?

jviau: Ah the ordering problem does significantly complicate any decoupled approach. If your feature branch is addressing this more holistically, then I am okay with this for the interim. Mostly because I don't have a better suggestion 😆

brettsam: Thanks for thinking it through -- I was banging my head on trying to get something together for a targeted fix ASAP while I think through the bigger change. It's definitely becoming a warning sign to me when `IConfigureOptions<>` takes lots of dependencies, especially when you need to deal with cache invalidation. At that point, it seems like it's better to implement your own caching setup as it's no longer "simple". We have a couple places like this in the code (or where `IOptions<>` depends on another `IOptions<>`).

jviau: I for some reason though option dependency chaining had implicit reload support. Knowing you need to configure that all manually is definitely painful. I think that is solvable through some improved registration APIs.

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
```diff
@@ -398,16 +394,7 @@ private async Task UnsynchronizedStartHostCoreAsync(ScriptHostStartupOperation a
...
-                if (_workerConfigResolverOptionsChangeTokenSource is HostBuiltChangeTokenSource<WorkerConfigurationResolverOptions> { } hostBuiltChangeTokenResolverOptions)
-                {
-                    hostBuiltChangeTokenResolverOptions.TriggerChange();
-                }
-
-                if (_languageWorkerOptionsChangeTokenSource is HostBuiltChangeTokenSource<LanguageWorkerOptions> { } hostBuiltChangeTokenSource)
-                {
-                    hostBuiltChangeTokenSource.TriggerChange();
-                }
+                _workerConfigCacheInvalidator.InvalidateCachePostBuildIfEnabled();
```
jviau, line 397: Thinking about this a bit more: don't we have some kind of event that gets sent out whenever the script host reloads?

https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Host/IScriptHostManager.cs#L22

Instead of this class knowing about the invalidator, can the invalidator listen to that event and trigger the option reloads?

brettsam: That could work to remove `InvalidateCachePostBuildIfEnabled()`, yeah. It'd just happen automatically if we wire up to that event.

If we wanted to remove `InvalidateCacheForBundles()` we'd need to expose a new event that we raise at that specific point and then also expose through the ScriptHostManager ([like we do with Initializing today](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs#L415)).

Then the weird middle API `EnableInvalidationForNextBuild()` would also still be required. From what I could find we had no way today to see if bundles were actually being loaded and used (if not, we need to invalidate after the build step). We could introduce something new for this as well -- where externally we can see if that ScriptHost is actually using Bundles? The issue here is that you can specify bundles in your host.json... but if you're using dotnet-isolated (for example), we won't actually load them.

I mentioned elsewhere but this is already part of my planning for the compute separation (it affects how we wire everything up) so my goal is to re-think this and make it much more explicit about how things are running while still fixing this bug for our customer short-term.

jviau (approved):

src/WebJobs.Script.WebHost/WebJobsScriptHostService.cs
jviau: Can we just like, not have it be so hard? 😵‍💫

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
```diff
@@ -93,6 +96,9 @@ public async Task<IEnumerable<Type>> GetExtensionsStartupTypesAsync()
             ImmutableArray<FunctionMetadata> functionMetadataCollection = ImmutableArray<FunctionMetadata>.Empty;
             if (bundleConfigured)
             {
+                // refresh the cache immediately before we attempt to start a worker and get metadata for bundles
+                _workerConfigCacheInvalidator.InvalidateCacheForBundles();
+
```
kshyju, line 101: Could there still be race condition since we have no definite way of knowing that the effect of the `TriggerChange` is completed (ie: the `Configure` methods finished execution), before we proceed to lines below?

kshyju (approved):

src/WebJobs.Script/DependencyInjection/ScriptStartupTypeLocator.cs
kshyju: Never mind. I guess the .NET DI infrastructure ensure that that dependency is injected after the IConfigureOptions<T>.Configure method call is completed before injecting that dependency.

jviau (approved):

kshyju (approved):

PR #11601, opened 2026-02-10
improve timer trigger schedule validation for consumption plans

### LINKED ISSUE #11555: Timer expression checks need improving for all Consumption plans
We have [code that ensures a Timer on Windows Consumption has a 6-digit CRON expression](https://github.com/Azure/azure-functions-host/blob/69b013c04f1ec8e88f9d3cb04a3b2fca1df16f30/src/WebJobs.Script/Binding/CoreExtensionsScriptBindingProvider.cs#L66-L79) (and does not support other expressions, such as a TimeSpan).

The background here is that the ScaleController only supports CRON expressions, so we need to ensure that this is restricted. At the time this code was added, only 6-digit CRON expressions were supported, but now the ScaleController actually supports both 5 and 6 digit expressions.

There's a few problems with this restriction though:
1. This code should have also applied to Linux Consumption (CV1). But it never did, which means some customers are using 5-digit CRON expressions. This works fine today, though.
2. CV1 customers could use a TimeSpan and this check would pass, only to cause issues with scaling later.
3. The `IsWindowsConsumption()` check is actually invalid. It returns true for Flex Consumption as well. This means that during migrations from CV1 to Flex, some customers using 5-digit CRON expressions are now hitting errors on startup.

We should be able to fix this by:
1. Checking telemetry to see if any CV1 customers are using TimeSpans (as the following steps would break them). They'd technically already be broken in the ScaleController, but their timers would start properly when the host started up.
2. Relaxing the restriction and also allowing 5-digit expressions.
3. Applying the restriction to *all consumption plans*.

This should allow CV1 customers to migrate to Flex without hitting this error... and also relax the restriction on Windows Consumption users.

brettsam:
Resolves #11555 

Improve timer trigger schedule validation for consumption plans

   - Replace direct IEnvironment dependency with IOptions<TimerTriggerPlatformOptions>
   - Accept both 5 and 6-digit CRON expressions
   - Apply CRON-only validation to all consumption SKUs, not just Windows
   - Warn (instead of block) on non-CRON schedules for Linux Consumption (ACI/Legion)
   - Add IEnvironment.Platform property for OS detection
   
CRON Expression Support

   | Platform                        | Before               | After            |
   |---------------------------------|-----------------------|------------------|
   | Windows Consumption             | 6-digit only          | 5 and 6-digit    |
   | Flex Consumption                | 6-digit only¹         | 5 and 6-digit    |
   | Linux Consumption (ACI/Atlas)   | Any (no validation)   | 5 and 6-digit    |
   | Linux Consumption (Legion)      | Any (no validation)   | 5 and 6-digit    |

  Non-CRON Schedule Behavior (e.g. TimeSpan)

   | Platform                        | Before    | After    |
   |---------------------------------|-----------|----------|
   | Windows Consumption             | Error     | Error    |
   | Flex Consumption                | Error¹    | Error    |
   | Linux Consumption (ACI/Atlas)   | Allow     | Warn     |
   | Linux Consumption (Legion)      | Allow     | Warn     |

   ¹ Flex was incorrectly caught by the IsWindowsConsumption() check, which only looked at WEBSITE_SKU == "Dynamic" without verifying the OS.

### Pull request checklist

    * [x] Otherwise: I have added/updated all related diagnostic events and their documentation ([Documentation issue linked to PR](https://github.com/MicrosoftDocs/azure-docs-pr/pull/311654))

### DIFF
--- a/src/WebJobs.Script/Binding/CoreExtensionsScriptBindingProvider.cs
+++ b/src/WebJobs.Script/Binding/CoreExtensionsScriptBindingProvider.cs
@@ -3,9 +3,11 @@
 using System;
 using System.Collections.ObjectModel;
 using Microsoft.Azure.WebJobs.Host;
+using Microsoft.Azure.WebJobs.Script.Configuration;
+using Microsoft.Azure.WebJobs.Script.Diagnostics;
 using Microsoft.Azure.WebJobs.Script.Extensibility;
-using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
 using NCrontab;
 
 namespace Microsoft.Azure.WebJobs.Script.Binding
@@ -16,19 +18,19 @@ namespace Microsoft.Azure.WebJobs.Script.Binding
     internal class CoreExtensionsScriptBindingProvider : ScriptBindingProvider
     {
         private readonly INameResolver _nameResolver;
-        private readonly IEnvironment _environment;
+        private readonly TimerTriggerPlatformOptions _platformOptions;
 
-        public CoreExtensionsScriptBindingProvider(INameResolver nameResolver, IEnvironment environment, ILogger<CoreExtensionsScriptBindingProvider> logger)
+        public CoreExtensionsScriptBindingProvider(INameResolver nameResolver, IOptions<TimerTriggerPlatformOptions> platformOptions, ILogger<CoreExtensionsScriptBindingProvider> logger)
             : base(logger)
         {
             _nameResolver = nameResolver ?? throw new ArgumentNullException(nameof(nameResolver));
-            _environment = environment ?? throw new ArgumentNullException(nameof(environment));
+            _platformOptions = (platformOptions ?? throw new ArgumentNullException(nameof(platformOptions))).Value;
         }
 
         /// <inheritdoc/>
         public override bool TryCreate(ScriptBindingContext context, out ScriptBinding binding)
         {
-            if (context == null)
+            if (context is null)
             {
                 throw new ArgumentNullException(nameof(context));
             }
@@ -37,21 +39,23 @@ namespace Microsoft.Azure.WebJobs.Script.Binding
 
             if (string.Equals(context.Type, "timerTrigger", StringComparison.OrdinalIgnoreCase))
             {
-                binding = new TimerTriggerScriptBinding(_nameResolver, _environment, context);
+                binding = new TimerTriggerScriptBinding(_nameResolver, _platformOptions, Logger, context);
             }
 
-            return binding != null;
+            return binding is not null;
         }
 
         internal class TimerTriggerScriptBinding : ScriptBinding
         {
             private readonly INameResolver _nameResolver;
-            private readonly IEnvironment _environment;
+            private readonly TimerTriggerPlatformOptions _platformOptions;
+            private readonly ILogger _logger;
 
-            public TimerTriggerScriptBinding(INameResolver nameResolver, IEnvironment environment, ScriptBindingContext context) : base(context)
+            public TimerTriggerScriptBinding(INameResolver nameResolver, TimerTriggerPlatformOptions platformOptions, ILogger logger, ScriptBindingContext context) : base(context)
             {
                 _nameResolver = nameResolver ?? throw new ArgumentNullException(nameof(nameResolver));
-                _environment = environment ?? throw new ArgumentNullException(nameof(environment));
+                _platformOptions = platformOptions ?? throw new ArgumentNullException(nameof(platformOptions));
+                _logger = logger ?? throw new ArgumentNullException(nameof(logger));
             }
 
             public override Type DefaultType => typeof(TimerInfo);
@@ -63,18 +67,25 @@ namespace Microsoft.Azure.WebJobs.Script.Binding
                 string schedule = Context.GetMetadataValue<string>("schedule");
                 bool runOnStartup = Context.GetMetadataValue<bool>("runOnStartup");
                 bool useMonitor = Context.GetMetadataValue<bool>("useMonitor", true);
-                if (_environment.IsWindowsConsumption())
+
+                if (_platformOptions.NonCronScheduleBehavior is not NonCronScheduleBehavior.Allow)
                 {
                     // pre-resolve app setting specifiers
                     schedule = _nameResolver.ResolveWholeString(schedule);
 
-                    var options = new CrontabSchedule.ParseOptions()
+                    // Accept both 5-digit and 6-digit (with seconds) CRON expressions.
+                    bool isCron = CrontabSchedule.TryParse(schedule) is not null
+                        || CrontabSchedule.TryParse(schedule, new CrontabSchedule.ParseOptions { IncludingSeconds = true }) is not null;
+
+                    if (!isCron)
                     {
-                        IncludingSeconds = true
-                    };
-                    if (CrontabSchedule.TryParse(schedule, options) == null)
-                    {
-                        throw new ArgumentException(string.Format("'{0}' is not a valid CRON expression.", schedule));
+                        if (_platformOptions.NonCronScheduleBehavior is NonCronScheduleBehavior.Error)
+                        {
+                            throw new ArgumentException(string.Format("'{0}' is not a valid CRON expression.", schedule));
+                        }
+
+                        string message = $"The timer schedule '{schedule}' is not a CRON expression. Non-CRON expressions are not supported by the scale controller and may cause scaling issues. Please use a CRON expression instead. See {DiagnosticEventConstants.TimerConstantExpressionWarningHelpLink} for more information.";
+                        _logger.LogDiagnosticEventWarning(DiagnosticEventConstants.TimerConstantExpressionWarningErrorCode, message, DiagnosticEventConstants.TimerConstantExpressionWarningHelpLink, null);
                     }
                 }
 
--- a/src/WebJobs.Script/Binding/CoreExtensionsScriptBindingProvider.cs
+++ b/src/WebJobs.Script/Binding/CoreExtensionsScriptBindingProvider.cs
@@ -20,7 +20,10 @@ namespace Microsoft.Azure.WebJobs.Script.Binding
         private readonly INameResolver _nameResolver;
         private readonly TimerTriggerPlatformOptions _platformOptions;
 
-        public CoreExtensionsScriptBindingProvider(INameResolver nameResolver, IOptions<TimerTriggerPlatformOptions> platformOptions, ILogger<CoreExtensionsScriptBindingProvider> logger)
+        public CoreExtensionsScriptBindingProvider(
+            INameResolver nameResolver,
+            IOptions<TimerTriggerPlatformOptions> platformOptions,
+            ILogger<CoreExtensionsScriptBindingProvider> logger)
             : base(logger)
         {
             _nameResolver = nameResolver ?? throw new ArgumentNullException(nameof(nameResolver));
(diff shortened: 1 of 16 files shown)


brettsam:
documentation PR is at https://github.com/MicrosoftDocs/azure-docs-pr/pull/311654

src/WebJobs.Script/Binding/CoreExtensionsScriptBindingProvider.cs
```diff
@@ -16,19 +18,19 @@ namespace Microsoft.Azure.WebJobs.Script.Binding
     internal class CoreExtensionsScriptBindingProvider : ScriptBindingProvider
     {
         private readonly INameResolver _nameResolver;
-        private readonly IEnvironment _environment;
+        private readonly TimerTriggerPlatformOptions _platformOptions;
 
-        public CoreExtensionsScriptBindingProvider(INameResolver nameResolver, IEnvironment environment, ILogger<CoreExtensionsScriptBindingProvider> logger)
+        public CoreExtensionsScriptBindingProvider(INameResolver nameResolver, IOptions<TimerTriggerPlatformOptions> platformOptions, ILogger<CoreExtensionsScriptBindingProvider> logger)
```
kshyju: Could we wrap parameters here? One per line. I like the readability aspect of that.

src/WebJobs.Script/Config/TimerTriggerPlatformOptions.cs
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Script.Configuration;
+
+/// <summary>
+/// Options that describe platform-level capabilities for timer trigger bindings.
+/// </summary>
+public class TimerTriggerPlatformOptions
```
kshyju: Don't see anything extending it. So let's seal it?
```
public sealed class TimerTriggerPlatformOptions
```

kshyju: Should we implement `IOptionsFormatter` so this is logged at creation time? Once[ this change](https://github.com/Azure/azure-webjobs-sdk/issues/3178) is consumed by host, webhost level TOptions will be logged.

src/WebJobs.Script/Config/TimerTriggerPlatformOptionsSetup.cs
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.WebJobs.Script.Configuration;
+
+internal class TimerTriggerPlatformOptionsSetup : IConfigureOptions<TimerTriggerPlatformOptions>
```
kshyju: Mark `sealed`

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -204,7 +204,7 @@ public static bool IsLinuxAppServiceWithPersistentFileSystem(this IEnvironment e
         public static bool IsWindowsConsumption(this IEnvironment environment)
         {
             string value = environment.GetEnvironmentVariable(AzureWebsiteSku);
-            return string.Equals(value, ScriptConstants.DynamicSku, StringComparison.OrdinalIgnoreCase);
+            return environment.Platform == OSPlatform.Windows && string.Equals(value, ScriptConstants.DynamicSku, StringComparison.OrdinalIgnoreCase);
```
kshyju: Should we check the platform condition first and return early to avoid retrieving the environment variable all the time?
```
if (environment.Platform != OSPlatform.Windows)
{
    return false;
}

// now get env var value and compare
```

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -301,7 +301,7 @@ public static bool IsDynamicSku(this IEnvironment environment)
         /// <returns><see cref="true"/> if running in a Windows Azure managed hosting environment; otherwise, false.</returns>
         public static bool IsWindowsAzureManagedHosting(this IEnvironment environment)
         {
-            return environment.IsAppService() && RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
+            return environment.IsAppService() && environment.Platform == OSPlatform.Windows;
```
kshyju: Here also, swap the order. Check platform first and return early. Then `IsAppService` is retrieving the env variable in it.

src/WebJobs.Script/Environment/SystemEnvironment.cs
```diff
@@ -33,5 +36,27 @@ public void SetEnvironmentVariable(string name, string value)
...
             Environment.SetEnvironmentVariable(name, value);
         }
+
+        private static OSPlatform GetCurrentPlatform()
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                return OSPlatform.Windows;
+            }
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
```
kshyju: nit: Can remove the `else` from `else if`

src/WebJobs.Script/Environment/SystemEnvironment.cs
```diff
@@ -33,5 +36,27 @@ public void SetEnvironmentVariable(string name, string value)
         {
             Environment.SetEnvironmentVariable(name, value);
         }
+
+        private static OSPlatform GetCurrentPlatform()
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
kshyju: The `OperatingSystem.IsXYZ` method appears to be a little faster as it simply returns [compile time  boolean result](https://github.com/dotnet/runtime/blob/dc5fd7a8dce8309e4add8fd4bd5d8718f221b15a/src/libraries/System.Private.CoreLib/src/System/OperatingSystem.cs#L313) value (from the OS specific implementation). The IsOSPlatform does a[ string comparison](https://github.com/dotnet/runtime/blob/dc5fd7a8dce8309e4add8fd4bd5d8718f221b15a/src/libraries/System.Private.CoreLib/src/System/OperatingSystem.cs#L123).

Tiny improvement is visible on linux where `RuntimeInformation.IsOSPlatform` has to fail the Windows check before matching Linux, 
https://github.com/kshyju/Benchmarks/actions/runs/21890721558/job/63195728999#step:7:313

```
public static OSPlatform GetCurrentPlatform()
{
    if (OperatingSystem.IsWindows())
    {
        return OSPlatform.Windows;
    }

    if (OperatingSystem.IsLinux())
    {
        return OSPlatform.Linux;
    }

    if (OperatingSystem.IsMacOS())
    {
        return OSPlatform.OSX;
    }

    if (OperatingSystem.IsFreeBSD())
    {
        return OSPlatform.FreeBSD;
    }

    return OSPlatform.Create("Unknown");
}
```

src/WebJobs.Script/Binding/CoreExtensionsScriptBindingProvider.cs
```diff
@@ -63,18 +67,25 @@ public override Collection<Attribute> GetAttributes()
...
+                    if (!isCron)
                     {
-                        throw new ArgumentException(string.Format("'{0}' is not a valid CRON expression.", schedule));
+                        if (_platformOptions.NonCronScheduleBehavior is NonCronScheduleBehavior.Error)
+                        {
+                            throw new ArgumentException(string.Format("'{0}' is not a valid CRON expression.", schedule));
+                        }
+
+                        string message = $"The timer schedule '{schedule}' is not a CRON expression. Non-CRON expressions are not supported by the scale controller and may cause scaling issues. Please use a CRON expression instead. See {DiagnosticEventConstants.TimerConstantExpressionWarningHelpLink} for more information.";
+                        _logger.LogDiagnosticEventWarning(DiagnosticEventConstants.TimerConstantExpressionWarningErrorCode, message, DiagnosticEventConstants.TimerConstantExpressionWarningHelpLink, null);
```
kshyju, line 94: Should we log the diagnostic warning before the throw (in line 84) so those customers also get to see this warning in the portal?

src/WebJobs.Script/Environment/SystemEnvironment.cs
brettsam: Oh nice catch! And benchmark! Copilot did this for me. Bad copilot!

src/WebJobs.Script/Binding/CoreExtensionsScriptBindingProvider.cs
brettsam: Good idea. I was just maintaining current behavior with the exception but making this error clearer to others is a good idea also.

test/WebJobs.Script.Tests/Binding/CoreExtensionsScriptBindingProviderTests.cs
```diff
@@ -5,26 +5,41 @@
...
+        private readonly TestLoggerProvider _loggerProvider = new TestLoggerProvider();
+        private readonly ILogger _logger;
+
+        public CoreExtensionsScriptBindingProviderTests()
+        {
+            var loggerFactory = new LoggerFactory();
+            loggerFactory.AddProvider(_loggerProvider);
+            _logger = loggerFactory.CreateLogger<CoreExtensionsScriptBindingProvider>();
+        }
+
```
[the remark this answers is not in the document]
brettsam: we don't dispose loggerfactory in tests anywhere so not going to implement this one...

kshyju (approved):

src/WebJobs.Script/Binding/CoreExtensionsScriptBindingProvider.cs
```diff
@@ -37,21 +42,23 @@ public override bool TryCreate(ScriptBindingContext context, out ScriptBinding b
...
+            private readonly TimerTriggerPlatformOptions _platformOptions;
+            private readonly ILogger _logger;
 
-            public TimerTriggerScriptBinding(INameResolver nameResolver, IEnvironment environment, ScriptBindingContext context) : base(context)
+            public TimerTriggerScriptBinding(INameResolver nameResolver, TimerTriggerPlatformOptions platformOptions, ILogger logger, ScriptBindingContext context) : base(context)
             {
                 _nameResolver = nameResolver ?? throw new ArgumentNullException(nameof(nameResolver));
-                _environment = environment ?? throw new ArgumentNullException(nameof(environment));
+                _platformOptions = platformOptions ?? throw new ArgumentNullException(nameof(platformOptions));
+                _logger = logger ?? throw new ArgumentNullException(nameof(logger));
```
fabiocav, line 61: After we've updated the projects, we should be able to use `ArgumentNullException.ThrowIfNull` here.

src/WebJobs.Script/Binding/CoreExtensionsScriptBindingProvider.cs
```diff
@@ -63,18 +70,28 @@ public override Collection<Attribute> GetAttributes()
...
+                        || CrontabSchedule.TryParse(schedule, new CrontabSchedule.ParseOptions { IncludingSeconds = true }) is not null;
+
+                    if (!isCron)
                     {
-                        throw new ArgumentException(string.Format("'{0}' is not a valid CRON expression.", schedule));
+                        string message = $"The timer schedule '{schedule}' is not a CRON expression. Non-CRON expressions are not supported by the scale controller and may cause scaling issues. Please use a CRON expression instead. See {DiagnosticEventConstants.TimerConstantExpressionWarningHelpLink} for more information.";
+
+                        if (_platformOptions.NonCronScheduleBehavior is NonCronScheduleBehavior.Error)
+                        {
+                            var exception = new ArgumentException(string.Format("'{0}' is not a valid CRON expression.", schedule));
```
fabiocav, line 89: Can you replace `ArgumentException`  with `FunctionConfigurationException` for this?

src/WebJobs.Script/Diagnostics/DiagnosticEventConstants.cs
```diff
@@ -37,5 +37,8 @@ internal static class DiagnosticEventConstants
 
         public const string DeprecatedProxiesErrorCode = "AZFD0014";
         public const string DeprecatedProxiesHelpLink = "https://aka.ms/functions-deprecated-proxies";
+
+        public const string TimerConstantExpressionWarningErrorCode = "AZFD0015";
+        public const string TimerConstantExpressionWarningHelpLink = "https://aka.ms/functions-timer-cron";
```
fabiocav, line 42: Are we tracking the creation of this AKA.MS link anywhere?

fabiocav (changes requested):

src/WebJobs.Script/Binding/CoreExtensionsScriptBindingProvider.cs
brettsam: Done. I thought that may technically be a breaking change? Although at this part of the code I don't think it matters much.

src/WebJobs.Script/Diagnostics/DiagnosticEventConstants.cs
brettsam: It's created -- https://aka.ms/functions-timer-cron -- just doesn't point to anything as the doc isn't published yet.

liliankasem (approved):

fabiocav (approved):

PR #11609, opened 2026-02-17
Filter out dependency telemetry for localhost calls

### LINKED ISSUE #11602: Host leaks function/host key verbatim to Application Insights as dependency telemetry
- Timestamp: 2026-02-11T09:29:59.7474204Z
- Function App version: Dotnet-isolated - 10.0 (linux)
- Invocation ID: 7626f396-74bc-40e9-beae-e6e2e92fc5e7
- Region: West Europe

#### Repro steps

1. Perform an HTTP call to an HttpTrigger function, defined like this:
``` cs
public async Task<IActionResult> RunAsync(
    [HttpTrigger(AuthorizationLevel.Function, "get", "post")] HttpRequest req)
{
```
2. Observe logging in Application Insights resource
3. Notice a dependency telemetry is logged like this:
```http://localhost:39357/<function-path>?code=<the-actual-key>```

#### Expected behavior

The function host should never leak the function key to Application Insights or other logging/monitoring tools.
Instead, the secret value should be redacted like this:
```http://localhost:39357/<function-path>?code=REDACTED```

#### Actual behavior

The function or host key (whatever was used to authorize the request) is leaked to Application Insights.

#### Known workarounds

* Not use Azure functions
* Not use HttpTrigger
* Not use key-based authorization
* Not use isolated process model
* Not have security or compliance needs

#### Related information 

I developed my function in C# on .NET 10.
I was not able to work around this problem with an `ITelemetryInitializer`. I suspect this is because the telemetry is written by the host and not by my code, so there is nothing I can do about it.

Image

Image

Image

### PR DESCRIPTION
**Pull request by RohitRanjanMS**
This prevents host-to-worker HTTP proxy calls from appearing as dependency telemetry in Application Insights.

resolves #11602 

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script/Config/ScriptTelemetryProcessor.cs
+++ b/src/WebJobs.Script/Config/ScriptTelemetryProcessor.cs
@@ -21,8 +21,10 @@ namespace Microsoft.Azure.WebJobs.Script.Config
 
         public void Process(ITelemetry item)
         {
-            // Filter out dependency telemetry for localhost calls (host-to-worker interactions).
-            if (item is DependencyTelemetry dependencyTelemetry && IsLocalhostDependency(dependencyTelemetry))
+            // Filter out all outgoing HTTP dependency telemetry. In out-of-proc scenarios the host's
+            // only HTTP dependency is the loopback call to the worker.
+            if (item is DependencyTelemetry dep
+                && string.Equals(dep.Type, "Http", StringComparison.OrdinalIgnoreCase))
             {
                 return;
             }
@@ -35,22 +37,6 @@ namespace Microsoft.Azure.WebJobs.Script.Config
                 item = ToUserException(rpcException, item);
             }
             this.Next.Process(item);
-        }
-
-        private static bool IsLocalhostDependency(DependencyTelemetry dependencyTelemetry)
-        {
-            if (string.IsNullOrEmpty(dependencyTelemetry.Data))
-            {
-                return false;
-            }
-
-            if (string.Equals(dependencyTelemetry.Type, "Http", StringComparison.OrdinalIgnoreCase)
-                && Uri.TryCreate(dependencyTelemetry.Data, UriKind.Absolute, out var uri))
-            {
-                return uri.IsLoopback;
-            }
-
-            return false;
         }
 
         private ITelemetry ToUserException(RpcException rpcException, ITelemetry originalItem)
--- a/src/WebJobs.Script/Config/ScriptTelemetryProcessor.cs
+++ b/src/WebJobs.Script/Config/ScriptTelemetryProcessor.cs
@@ -44,7 +44,8 @@ namespace Microsoft.Azure.WebJobs.Script.Config
                 return false;
             }
 
-            if (string.Equals(dependencyTelemetry.Type, "Http", StringComparison.OrdinalIgnoreCase) && Uri.TryCreate(dependencyTelemetry.Data, UriKind.Absolute, out var uri))
+            if (string.Equals(dependencyTelemetry.Type, "Http", StringComparison.OrdinalIgnoreCase)
+                && Uri.TryCreate(dependencyTelemetry.Data, UriKind.Absolute, out var uri))
             {
                 return uri.IsLoopback;
             }
--- a/src/WebJobs.Script/Config/ScriptTelemetryProcessor.cs
+++ b/src/WebJobs.Script/Config/ScriptTelemetryProcessor.cs
@@ -21,6 +21,12 @@ namespace Microsoft.Azure.WebJobs.Script.Config
 
         public void Process(ITelemetry item)
         {
+            // Filter out dependency telemetry for localhost calls (host-to-worker interactions).
+            if (item is DependencyTelemetry dependencyTelemetry && IsLocalhostDependency(dependencyTelemetry))
+            {
+                return;
+            }
+
             // Only process if exception is thrown by user code (if IsUserException is true).
             if (item is ExceptionTelemetry exceptionTelemetry
                 && exceptionTelemetry?.Exception?.InnerException is RpcException rpcException
@@ -29,6 +35,21 @@ namespace Microsoft.Azure.WebJobs.Script.Config
                 item = ToUserException(rpcException, item);
             }
             this.Next.Process(item);
+        }
+
+        private static bool IsLocalhostDependency(DependencyTelemetry dependencyTelemetry)
+        {
+            if (string.IsNullOrEmpty(dependencyTelemetry.Data))
+            {
+                return false;
+            }
+
+            if (string.Equals(dependencyTelemetry.Type, "Http", StringComparison.OrdinalIgnoreCase) && Uri.TryCreate(dependencyTelemetry.Data, UriKind.Absolute, out var uri))
+            {
+                return uri.IsLoopback;
+            }
+
+            return false;
         }
 
         private ITelemetry ToUserException(RpcException rpcException, ITelemetry originalItem)
--- a/src/WebJobs.Script/Config/ScriptTelemetryProcessor.cs
+++ b/src/WebJobs.Script/Config/ScriptTelemetryProcessor.cs
@@ -25,8 +25,7 @@ namespace Microsoft.Azure.WebJobs.Script.Config
         {
             // Filter out HTTP dependency telemetry originating from the host's proxy calls to
             // out-of-proc workers.
-            if (item is DependencyTelemetry
-                && IsHostProxyForwarding.Value)
+            if (item is DependencyTelemetry && IsHostProxyForwarding.Value)
             {
                 return;
             }
--- a/src/WebJobs.Script/Config/ScriptTelemetryProcessor.cs
+++ b/src/WebJobs.Script/Config/ScriptTelemetryProcessor.cs
@@ -12,7 +12,7 @@ namespace Microsoft.Azure.WebJobs.Script.Config
 {
     internal class ScriptTelemetryProcessor : ITelemetryProcessor
     {
-        internal static readonly AsyncLocal<bool> IsHostProxyForwarding = new();
+        internal static readonly AsyncLocal<bool> SuppressDependencyTelemetry = new();
 
         public ScriptTelemetryProcessor(ITelemetryProcessor next)
         {
@@ -25,7 +25,7 @@ namespace Microsoft.Azure.WebJobs.Script.Config
         {
             // Filter out HTTP dependency telemetry originating from the host's proxy calls to
             // out-of-proc workers.
-            if (item is DependencyTelemetry && IsHostProxyForwarding.Value)
+            if (item is DependencyTelemetry && SuppressDependencyTelemetry.Value)
             {
                 return;
             }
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/WebJobs.Script/Config/ScriptTelemetryProcessor.cs [hunk 1]
```diff
@@ -21,6 +21,12 @@ public ScriptTelemetryProcessor(ITelemetryProcessor next)
 
         public void Process(ITelemetry item)
         {
+            // Filter out dependency telemetry for localhost calls (host-to-worker interactions).
+            if (item is DependencyTelemetry dependencyTelemetry && IsLocalhostDependency(dependencyTelemetry))
```
kshyju: Wondering this has any impact on other localhost communications? Ex: custom handler scenario or talking to any of the side cars we have (ex: Auth/Managed Identity). Will we lose those logs? What is the behavior if we fix this in the /ScriptInvocationRequestTransformer's [TransformRequestAsync](https://github.com/Azure/azure-functions-host/blob/e5e773d1453e361e5713e28b9214d7ecc310f8b8/src/WebJobs.Script/Http/ScriptInvocationRequestTransformer.cs#L20) method?

src/WebJobs.Script/Config/ScriptTelemetryProcessor.cs
```diff
@@ -31,6 +37,21 @@ public void Process(ITelemetry item)
...
         }
 
+        private static bool IsLocalhostDependency(DependencyTelemetry dependencyTelemetry)
+        {
+            if (string.IsNullOrEmpty(dependencyTelemetry.Data))
+            {
+                return false;
+            }
+
+            if (string.Equals(dependencyTelemetry.Type, "Http", StringComparison.OrdinalIgnoreCase) && Uri.TryCreate(dependencyTelemetry.Data, UriKind.Absolute, out var uri))
```
kshyju: nit: Wrap the long line to improve readability.

src/WebJobs.Script/Config/ScriptTelemetryProcessor.cs
RohitRanjanMS: Any localhost communication is just noise from the customer’s perspective. The same applies to dependencies generated by the host’s interactions with storage and other services. I could have redacted the query string, but I chose to remove it instead.
Also, for this particular dependency, there’s no corresponding server span from the worker, so the interaction is incomplete anyway — we only see the client span with no matching server span.

src/WebJobs.Script/Config/ScriptTelemetryProcessor.cs
```diff
@@ -31,6 +37,22 @@ public void Process(ITelemetry item)
             this.Next.Process(item);
         }
 
+        private static bool IsLocalhostDependency(DependencyTelemetry dependencyTelemetry)
```
soninaren: Should we have a switch / config in case we need to have this telemetry data?

soninaren (approved):
Looks good. Wondering if we need to have some override config to get this telemetry data.

fabiocav (dismissed):
@RohitRanjanMS have you identified why the URI was being logged the way it was? Was the issue limited to localhost calls? How can we ensure other calls won't be impacted?

src/WebJobs.Script/Config/ScriptTelemetryProcessor.cs
```diff
@@ -31,6 +37,22 @@ public void Process(ITelemetry item)
...
+        {
+            if (string.IsNullOrEmpty(dependencyTelemetry.Data))
+            {
+                return false;
+            }
+
+            if (string.Equals(dependencyTelemetry.Type, "Http", StringComparison.OrdinalIgnoreCase)
+                && Uri.TryCreate(dependencyTelemetry.Data, UriKind.Absolute, out var uri))
```
jviau: Is there a way to check for loopback without creating a URI? Avoiding that object allocation will be helpful in a hotpath like telemetry.

RohitRanjanMS:
@fabiocav ,

 There were two things I was investigating: why this dependency telemetry was appearing, and why its `sdkVersion` was set
   to `rdddsc` instead of the version we configure in our telemetry pipeline.

  **Why rdddsc as the sdkVersion:**
  The Application Insights SDK has an `HttpCoreDiagnosticSourceListener` that emits dependency telemetry for outgoing HTTP
   calls. This component has been enabled for at least 7-8 years (there was an App Insights repo migration, so I can't
  trace further back). When [HttpCoreDiagnosticSourceListener](https://github.com/microsoft/ApplicationInsights-dotnet/blob/develop/WEB/Src/DependencyCollector/DependencyCollector/HttpCoreDiagnosticSourceListener.cs#L93) creates a dependency, it stamps its own `sdkVersion` (`rdddsc:<version>`) on it. Our telemetry configuration only sets `sdkVersion` if it hasn't already been set — so this dependency retains the `rdddsc` prefix instead of picking up the version we configure for the rest of our telemetry.

  **Why we're seeing it now:**
  We have an `OperationFilteringTelemetryProcessor` that filters out internal host operations — outgoing request, queue polling, blob lease, and similar infrastructure calls. It does this by checking for the presence of a `Category` property: all
  host-internal operations happen outside the function invocation scope, so `WebJobsTelemetryInitializer` never assigns a
  Category to them. [OperationFilteringTelemetryProcessor](https://github.com/Azure/azure-webjobs-sdk/blob/b2c42816aeb9179ecd0eff9b10ea046e0c291391/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Processors/OperationFilteringTelemetryProcessor.cs#L26) drops any operation telemetry without a Category, which is why
  we don't normally see these internal dependencies.

  This particular dependency started appearing after we enabled ASP.NET Core integration. With
  `ConfigureFunctionsWebApplication()`, the host now makes an HTTP call to forward the request to the worker's Kestrel
  server. `HttpCoreDiagnosticSourceListener` captures this call and emits a dependency. Because the call happens during a
  function invocation, there is an active logger scope with a Category — so `WebJobsTelemetryInitializer` stamps it on the
   dependency, and `OperationFilteringTelemetryProcessor` can no longer filter it out.

src/WebJobs.Script/Config/ScriptTelemetryProcessor.cs
RohitRanjanMS: I couldn't find a way to avoid the URI allocation. I initially tried string comparison — checking if the URL contains
  `localhost` or `127.0.0.1`. It isn't the most elegant approach, but it avoids the allocation. I can switch it back to
  string comparison to avoid the allocation.

src/WebJobs.Script/Config/ScriptTelemetryProcessor.cs
RohitRanjanMS: No, this telemetry has no value for the customer. Details here -> https://github.com/Azure/azure-functions-host/pull/11609#discussion_r2818446441

src/WebJobs.Script/Config/ScriptTelemetryProcessor.cs
RohitRanjanMS: Another option is to simply disable HTTP dependency tracking. It was useful in the in‑proc model, but in the out‑of‑proc model there won’t be any dependencies that customers would care about.

RohitRanjanMS: Disabling dependency tracking would turn off all dependency telemetry, so instead I’m filtering out only HTTP dependencies in the telemetry processor. I’ve also removed the loopback check.

jviau (commented):
I think that filtering out http, and even http + loopback will also impact emulators during local development. This might be problematic for aspire integration or other local testing.

I think we will need to scope this to something specific. Can we add some context onto the HTTP request message when we forward it to the worker, and then identify it here and suppress the telemetry?

src/WebJobs.Script/Config/ScriptTelemetryProcessor.cs
```diff
@@ -12,6 +12,8 @@ namespace Microsoft.Azure.WebJobs.Script.Config
 {
     internal class ScriptTelemetryProcessor : ITelemetryProcessor
     {
+        internal static readonly AsyncLocal<bool> IsHostProxyForwarding = new();
```
jviau: Nit, recommend naming this more for what it does and not where it is used:

``` suggestion
        internal static readonly AsyncLocal<bool> SuppressDependency = new();
```

src/WebJobs.Script/Http/DefaultHttpProxyService.cs
```diff
@@ -103,7 +104,10 @@ public void StartForwarding(ScriptInvocationContext context, Uri httpUri)
             // This helps track failures/cancellations that should halt retrying the http request.
             httpContext.Items[ScriptConstants.HttpProxyScriptInvocationContext] = context;
 
+            ScriptTelemetryProcessor.IsHostProxyForwarding.Value = true;
             var forwardingTask = _httpForwarder.SendAsync(httpContext, httpUri.ToString(), _messageInvoker, _forwarderRequestConfig, _httpTransformer).AsTask();
+            ScriptTelemetryProcessor.IsHostProxyForwarding.Value = false;
```
jviau: Need to put this in a `finally` statement. Or wrap it in a disposable and have a helper method like:

``` CSharp
using var _ = ScriptTelemetryProcessor.SuppressDependencies();
```

src/WebJobs.Script/Http/DefaultHttpProxyService.cs
```diff
@@ -104,9 +104,16 @@ public void StartForwarding(ScriptInvocationContext context, Uri httpUri)
...
             httpContext.Items[ScriptConstants.HttpProxyScriptInvocationContext] = context;
 
-            ScriptTelemetryProcessor.IsHostProxyForwarding.Value = true;
-            var forwardingTask = _httpForwarder.SendAsync(httpContext, httpUri.ToString(), _messageInvoker, _forwarderRequestConfig, _httpTransformer).AsTask();
-            ScriptTelemetryProcessor.IsHostProxyForwarding.Value = false;
+            Task<ForwarderError> forwardingTask;
+            try
+            {
+                ScriptTelemetryProcessor.SuppressDependencyTelemetry.Value = true;
+                forwardingTask = _httpForwarder.SendAsync(httpContext, httpUri.ToString(), _messageInvoker, _forwarderRequestConfig, _httpTransformer).AsTask();
```
jviau: oh, we aren't waiting on the task here. Hmm there might be some complications with that. This code will be re-enabling dependency telemetry before the forwarding is complete. We may need to wrap this all in a helper:

``` CSharp
async Task ForwardAsync()
{
    try
    {
        await Task.Yield(); // think this will prevent the AsyncLocal from influencing the callers scope.
        ScriptTelemetryProcessor.SuppressDependencyTelemetry.Value = true;
        await _httpForwarder.SendAsync(httpContext, httpUri.ToString(), _messageInvoker, _forwarderRequestConfig, _httpTransformer);
    }
    finally
    {
        ScriptTelemetryProcessor.SuppressDependencyTelemetry.Value = false;
    }
}

context.Properties[ScriptConstants.HttpProxyTask] = ForwardAsync();
```

jviau (approved):

PR #11611, opened 2026-02-17
Refactor DiagnosticEventTableStorageRepository TableClient

### LINKED ISSUE #11198: Refactor DiagnosticEventTableStorageRepository TableClient
The `TableClient` property in `DiagnosticEventTableStorageRepository` is overloaded and doing too much. The TableClient property performs hidden side-effects, including I/O, state mutation, and service disabling, which violates .NET property conventions. It should be refactored into an explicit, thread-safe async initialization method to separate concerns and avoid unexpected behavior.

https://github.com/Azure/azure-functions-host/blob/051753227450ff71c4cb97a82f567da16b7691d0/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs#L61-L104

We should also add a test that ensures we never call the provider to create the client on the disposal path.

soninaren:
This pull request refactors the `DiagnosticEventTableStorageRepository` to improve thread safety, asynchronous initialization, and testability. The changes migrate from lock-based synchronization to `SemaphoreSlim`, introduce atomic operations for event handling, and update the public API to support async table client initialization. Several tests are also updated to use the new async methods.

**Repository refactoring and thread safety:**

* Replaces lock-based synchronization with a `SemaphoreSlim` for initializing the table client, ensuring thread-safe, asynchronous, and idempotent initialization (`DiagnosticEventTableStorageRepository.cs`). [[1]](diffhunk://#diff-76bee56cc44bd0539a1f7b4af8358cee5f5fee80b75fa772fb00ebdf30710a7fL35-R45) [[2]](diffhunk://#diff-76bee56cc44bd0539a1f7b4af8358cee5f5fee80b75fa772fb00ebdf30710a7fL61-R116) [[3]](diffhunk://#diff-76bee56cc44bd0539a1f7b4af8358cee5f5fee80b75fa772fb00ebdf30710a7fL101-R186)
* Uses `volatile` and atomic operations for `_tableClientInitialized` and `_isEnabled` to improve thread safety (`DiagnosticEventTableStorageRepository.cs`).
* Swaps the `_events` dictionary atomically using `Interlocked.Exchange` during flush to prevent race conditions (`DiagnosticEventTableStorageRepository.cs`).

**Async API and code modernization:**

* Adds async initialization and retrieval methods (`InitializeTableClientAsync`, `EnsureInitializedAsync`, `GetDiagnosticEventsTableAsync`), and updates internal logic to use these methods for better reliability and testability (`DiagnosticEventTableStorageRepository.cs`). [[1]](diffhunk://#diff-76bee56cc44bd0539a1f7b4af8358cee5f5fee80b75fa772fb00ebdf30710a7fL61-R116) [[2]](diffhunk://#diff-76bee56cc44bd0539a1f7b4af8358cee5f5fee80b75fa772fb00ebdf30710a7fL101-R186) [[3]](diffhunk://#diff-76bee56cc44bd0539a1f7b4af8358cee5f5fee80b75fa772fb00ebdf30710a7fL224-R280)

**Performance and correctness improvements:**

* Replaces lock-based event updates with `ConcurrentDictionary.AddOrUpdate` for atomic, thread-safe event aggregation (`DiagnosticEventTableStorageRepository.cs`).
* Ensures proper disposal of resources by disposing the semaphore and timer safely, and flushing logs on disposal (`DiagnosticEventTableStorageRepository.cs`).

**Test updates:**

* Refactors tests to use async versions of table retrieval and log flushing, and verifies correct service enable/disable behavior and error logging (`DiagnosticEventTableStorageRepositoryTests.cs`). [[1]](diffhunk://#diff-2c25f81b3f369634e157518921140171c64a7b77a8c392c74543cce300217166L168-R184) [[2]](diffhunk://#diff-2c25f81b3f369634e157518921140171c64a7b77a8c392c74543cce300217166L193-L201) [[3]](diffhunk://#diff-2c25f81b3f369634e157518921140171c64a7b77a8c392c74543cce300217166L215-R221) [[4]](diffhunk://#diff-2c25f81b3f369634e157518921140171c64a7b77a8c392c74543cce300217166L256-R262) [[5]](diffhunk://#diff-2c25f81b3f369634e157518921140171c64a7b77a8c392c74543cce300217166L351-R357)

**Documentation:**

* Adds a release note describing the refactor of `DiagnosticEventTableStorageRepository` and its `TableClient` logic (`release_notes.md`).

resolves https://github.com/Azure/azure-functions-host/issues/11198

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
@@ -33,9 +33,11 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
         private readonly ILogger<DiagnosticEventTableStorageRepository> _logger;
         private readonly IServiceProvider _serviceProvider;
         private readonly object _syncLock = new object();
+        private readonly SemaphoreSlim _initSemaphore = new SemaphoreSlim(1, 1);
 
         private ConcurrentDictionary<string, DiagnosticEvent> _events = new ConcurrentDictionary<string, DiagnosticEvent>();
         private TableServiceClient _tableClient;
+        private volatile bool _tableClientInitialized;
         private TableClient _diagnosticEventsTable;
         private string _hostId;
         private bool _disposed = false;
@@ -58,50 +60,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics
             IAzureTableStorageProvider azureTableStorageProvider, ILogger<DiagnosticEventTableStorageRepository> logger)
             : this(hostIdProvider, environment, scriptHost, azureTableStorageProvider, logger, LogFlushInterval) { }
 
-        internal TableServiceClient TableClient
-        {
-            get
-            {
-                if (_tableClient is null && !_environment.IsPlaceholderModeEnabled())
-                {
-                    if (!_azureTableStorageProvider.TryCreateHostingTableServiceClient(out _tableClient))
-                    {
-                        DisableService();
-                        Logger.ServiceDisabledFailedToCreateClient(_logger);
-                        return _tableClient;
-                    }
-
-                    try
-                    {
-                        // When using RBAC, we need "Storage Table Data Contributor" as we require to list, create and delete tables and query/insert/delete entities.
-                        // Testing permissions by listing tables, creating and deleting a test table.
-                        var testTable = _tableClient.GetTableClient($"{TableNamePrefix}Check");
-                        _ = TableStorageHelpers.TableExists(testTable, _tableClient);
-                        _ = testTable.CreateIfNotExists();
-                        _ = testTable.Delete();
-                    }
-                    catch (RequestFailedException rfe) when (rfe.Status == (int)HttpStatusCode.Conflict || rfe.ErrorCode == TableErrorCode.TableBeingDeleted)
-                    {
-                        // The table is being deleted or there could be a conflict for several instances initializing.
-                        // We can ignore this error as it is not a failure and we tested the permissions.
-                    }
-                    catch (RequestFailedException rfe) when (rfe.Status == (int)HttpStatusCode.Forbidden)
-                    {
-                        DisableService();
-                        Logger.ServiceDisabledUnauthorizedClient(_logger, rfe);
-                    }
-                    catch (Exception ex)
-                    {
-                        // We failed to connect to the table storage account. This could be due to a transient error or a configuration issue, such network issues.
-                        // We will disable the service.
-                        DisableService();
-                        Logger.ServiceDisabledUnableToConnectToStorage(_logger, ex);
-                    }
-                }
-
-                return _tableClient;
-            }
-        }
+        internal TableServiceClient TableClient => _tableClient;
 
         internal string HostId
         {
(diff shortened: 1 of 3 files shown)


src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -117,6 +76,68 @@ internal string HostId
...
+                    // We can ignore this error as it is not a failure and we tested the permissions.
+                }
+                catch (RequestFailedException rfe) when (rfe.Status == (int)HttpStatusCode.Forbidden)
+                {
+                    DisableService();
+                    Logger.ServiceDisabledUnauthorizedClient(_logger, rfe);
+                }
+                catch (Exception ex)
+                {
+                    // We failed to connect to the table storage account. This could be due to a transient error or a configuration issue, such network issues.
```
kshyju, line 130: Should we disable the service if there was a transient error to connect? Why not retry?

soninaren: This was actually part of existing code. But based on the comment it looks like it could be transient or permanent. we don't really know.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -117,6 +76,68 @@ internal string HostId
...
         internal ConcurrentDictionary<string, DiagnosticEvent> Events => _events;
 
+        internal async Task InitializeTableClientAsync()
+        {
+            if (_tableClientInitialized || _disposed)
+            {
+                return;
+            }
+
+            await _initSemaphore.WaitAsync();
```
liliankasem: Should we have this within the try catch and catch a `ObjectDisposedException`? 

If dispose is called while we are in init, this could throw

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -117,6 +76,68 @@ internal string HostId
...
+                {
+                    _tableClientInitialized = true;
+                }
+                _initSemaphore.Release();
+            }
+        }
+
         private void DisableService()
         {
             _isEnabled = false;
```
liliankasem, line 151: Should this be `volatile`? There could be a race with concurrent read/writes

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -126,7 +147,7 @@ private void DisableService()
 
         internal TableClient GetDiagnosticEventsTable(DateTime? now = null)
         {
```
liliankasem, line 157: This method will return null if _tableClient hasn’t been initialized yet. If a caller forgets to await InitializeTableClientAsync, the next call can hit “Unable to get table reference” and disable the service. 

Should we throw/assert when _tableClientInitialized is false?

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
```diff
@@ -272,10 +300,17 @@ internal virtual async Task FlushLogs(TableClient table = null)
...
 
-            // Assigning a new empty directory to reset the event count in the new duration window.
+            // Swap the events dictionary to reset the event count for the new flush window.
             // All existing events are logged to other logging pipelines already.
-            ConcurrentDictionary<string, DiagnosticEvent> tempDictionary = _events;
-            _events = new ConcurrentDictionary<string, DiagnosticEvent>();
+            // Use the same lock as WriteDiagnosticEvent to prevent race conditions when
+            // concurrent writes happen during the swap operation.
+            ConcurrentDictionary<string, DiagnosticEvent> tempDictionary;
+            lock (_syncLock)
```
RohitRanjanMS: We should be able to avoid lock and instead use `Interlocked.Exchange`

RohitRanjanMS (changes requested):

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
kshyju: We should consider adding the retry to be resilient. We shouldn't disable the service if it was a transient network connectivity issue.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
soninaren: @copilot can you make `_isEnabled` volatile.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
soninaren: I was debating that. This method is either called from tests or flushLogs that already calls `InitializeTableClientAsync`. This should have been a private method. I can add `InitializeTableClientAsync` so this method can be independently called.

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
soninaren: done

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
soninaren: Actually this will not disable the service. worst it can do is keep getting null size table client is not initialized.

soninaren: Have updated the initialization logic

src/WebJobs.Script.WebHost/Diagnostics/DiagnosticEventTableStorageRepository.cs
soninaren: I am a bit hesitant to implement retry logic here. We have seen a lot of CRI's because of noisy logs from diagnostic events.  Also, i think its a little out of scope of intended fix. Would like to gather some evidence of transient errors and open a separate issue for retry logic.

liliankasem (approved):

RohitRanjanMS (approved):

Issue #7710, opened 2021-09-28
Caching issue - function app staging slot keys appear in portal to change after swap slot performed


madelinegordon:
Caching issue - When a Function app with both a prod and staging slot performs a swap, the functions keys for the staging slot appear in the portal to adopt the same values as the prod keys, but in storage they do not change. (Not changing keys on swap is the expected behavior with default settings). 

#### Investigative information

This [IcM](https://portal.microsofticm.com/imp/v3/incidents/details/254306950/home) was filed

Please provide the following:

- Timestamp: 9-28-21 23:07:00 UTC
- Function App version: 3
- Function App name: SlotsSwapKey
- Function name(s) (as appropriate):
- Invocation ID: 
- Region: Central US

#### Repro steps

Provide the steps required to reproduce the problem:

1. Create a function app with 2 deployment slots.
2. Go to function app's storage account (name can be found in Configuration --> AzureWebJobsStorage)
3. In the storage account: containers --> azure-webjobs-secrets, then for each slot, host.json --> Edit --> set masterKey "encrypted" to false.
4. Note keys for each slot. (App keys)
5. Perform a slot swap in the function app, check keys for both prod and staging slot. 

#### Expected behavior

Provide a description of the expected behavior.

After step 5, the keys should have stayed the same. 

#### Actual behavior

Instead, the portal shows that the staging slot adopted the prod slot's keys. 

#### Known workarounds
Setting WEBSITE_FUNCTIONS_ARMCACHE_ENABLED=0 makes the issue go away. 

#### Related information 
screenshot of portal after a slot swap:
![keys](https://user-images.githubusercontent.com/50681653/135178357-a4cbe384-8879-4109-98fb-728684bcb07e.JPG)
arounds.

RobertPaulson90:
Any known fix to this? We use terraform to create create the initial deployment slot and az cli to perform the swap. For some very odd reason our DEV and QA environments have no problem and the production and staging slots share the same host keys, but for our prod environment, our API Management backend integration to the Function App breaks due to the key different between prod and staging slots. Super weird. `WEBSITE_FUNCTIONS_ARMCACHE_ENABLED=0` did not help, unfortunately.

satvu:
[...]

PR #11629, opened 2026-02-26
Fix stale hostname in ARM cache after slot swap

### LINKED ISSUE #9908: Slot Swap Feature - Send signal to function extension on hostname/routing swap completes.

Azure Logic Apps is introducing zero downtime deployment by taking advantage of slot support in Azure Functions.   Our workers depend on hostname and are initialized/registered during extension startup.    During a swap, hostname/routing is updated after startup under the covers without recycling the app.  The issue is once swap is complete, our workers reference the non-production hostname in a production slot. 

At startup, we get the endpoint by calling `context.GetWebhookHandler()`, see code in Repo: [WorkflowExtensionProvider.cs](https://msazure.visualstudio.com/One/_git/AzureUX-BPM?path=/src/functionExtensions/Flow.WebJobs.Extensions/Startup/WorkflowExtensionProvider.cs&version=GBmaster&_a=contents)

`WorkflowExtensionProvider` initialize classes that are used to build URLs displayed in the portal but also in other components, such as webhook and HTTP triggers, callback URLs and inputs/outputs URLs

An example of the issue is that in the portal, the end-user will see the staging hostname in production slot.  Below is one example.  

image

The ideal solution is for Azure Logic Apps to subscribe or depend on a signal from host runtime when hostname/routing is updated during a swap.  This ask will allow us to delay worker initialization/registration until hostname is updated.  In addition, it would be helpful to know that the startup was triggered by a swap.  This will help us to differentiate between, normal app recycle, and swap app recycle.  

We looked into synching the hostname after startup, however, there is no reliable path to perform the synch at the "right time".  The solution/workaround implemented in [HostNameProvider.cs](https://github.com/Azure/azure-functions-host/blob/c64e6c4ec447fdc091426576ed60c4e8f2f0cd44/src/WebJobs.Script/HostNameProvider.cs) is not a viable solution/workaround for us.

#### Additional context

Below is a screenshot of environment variables available at swap restart. 

image

pagariyaalok:

Partially addresses #9908 (see [Limitations](#limitations) below)
Related to #7710

**Summary:** Fix stale hostname persisting in ARM cache after slot swap, which causes wrong keys on `ListKeys` / Geo, and `UnresolvableHostName` errors when Logic Apps or other services invoke the function.

### Root Cause

After a slot swap, the Antares FE flips the hostname→site mapping, but worker processes are **not recycled**. `HostNameProvider.Value` retains the old (pre-swap) hostname in memory. When `FunctionsSyncManager` fires a background sync:

1. **`GetSyncTriggersPayload()`** builds `invoke_url_template` using the stale hostname (e.g., `https://myapp-staging.azurewebsites.net/api/func`)
2. The payload also includes **secrets/keys** from the current worker
3. **`BuildSetTriggersRequest()`** sends `POST https://{stale-hostname}/operations/settriggers` to the FE
4. The FE routes based on the hostname in the URL — which **now points to the other slot** after the swap
5. The ARM cache is updated **for the wrong site** with the wrong keys and wrong `invoke_url_template`

**Result:** `ListKeys` through ARM/Geo returns the wrong slot's keys. `invoke_url_template` points to the staging hostname, which may later be deleted — causing `UnresolvableHostName` errors for Logic Apps and other consumers that cached it.

### Fix

Add an event-based re-sync mechanism:

1. `HostNameProvider` now raises a `HostNameChanged` event when `Synchronize()` detects a hostname change via the `WAS_DEFAULT_HOSTNAME` header on an incoming HTTP request
2. `FunctionsSyncManager` subscribes to this event and triggers a **debounced (5-second)** background `TrySyncTriggersAsync()`
3. The debounce allows the hostname to stabilize after the swap before re-syncing
4. The re-sync sends the corrected payload to the correct FE endpoint, fixing the ARM cache

The `HostNameProvider.HostNameChanged` event also enables extensions (e.g., Logic Apps `WorkflowExtensionProvider`) to subscribe and re-register workers with the correct production hostname after a swap completes.

### Limitations

This PR **does not fully resolve #9908**. Specifically:

| Gap | Explanation |
|-----|-------------|
| **Non-HTTP apps** (timer, queue, Service Bus triggers only) | No HTTP request ever arrives after the swap → `Synchronize()` is never called → `HostNameChanged` never fires → ARM cache stays stale indefinitely |
| **Timing window** before first HTTP request | If a background sync fires before any request carries the new `WAS_DEFAULT_HOSTNAME` header, it still uses the stale hostname. Self-corrects once the first HTTP request arrives |
| **Very low-traffic apps** | Could be minutes or hours before the first HTTP request post-swap, extending the stale cache window |

A **complete** fix would require the swap infrastructure (Antares FE / site controller) to push a signal to the runtime when a swap completes, rather than relying on a header piggy-backed on the next HTTP request. That platform-level signal is what #9908 originally requested.

Related: ADO Task 4760001, ICM 21000000866148, ICM 654088427, ICM 510101424

### Pull request checklist

- [x] My changes **do not** require backporting to the `in-proc` branch
- [x] My changes do not require documentation changes
- [ ] My changes should not be added to the release notes for the next release
- [x] My changes do not require backporting to a previous version
- [x] My changes do not require diagnostic events changes
- [x] I have added all required tests (Unit tests, E2E tests)

### Additional information

**Files changed:**

| File | Description |
|------|-------------|
| `src/WebJobs.Script/HostNameProvider.cs` | Added `HostNameChanged` event, `OnHostNameChanged` method, fixed `string.Compare` → `string.Equals` with `StringComparison.OrdinalIgnoreCase` |
| `src/WebJobs.Script/HostNameChangedEventArgs.cs` | New sealed `HostNameChangedEventArgs` class (extracted per codebase convention) |
| `src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs` | Added event subscription with debounced background re-sync handler, improved `Dispose()` |
| `test/WebJobs.Script.Tests/HostNameProviderTests.cs` | 2 new unit tests for the event mechanism |
| `test/WebJobs.Script.Tests.Integration/Management/FunctionsSyncManagerTests.cs` | 3 new integration tests for hostname change sync behavior |

### DIFF
--- a/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
@@ -50,10 +50,6 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
         private const string ManagedKubernetesBuildServiceName = "k8se-build-service";
         private const string ManagedKubernetesBuildServiceNamespace = "k8se-system";
 
-        // Debounce delay for hostname change re-sync (in milliseconds).
-        // This prevents rapid re-syncs during slot swap hostname flip-flop.
-        private const int HostNameChangeDebounceDelayMs = 5000;
-
         private readonly Regex versionRegex = new Regex(@"Version=(?<majorversion>\d)\.\d\.\d");
 
         private readonly IOptionsMonitor<ScriptApplicationHostOptions> _applicationHostOptions;
@@ -69,11 +65,9 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
         private readonly IAzureBlobStorageProvider _azureBlobStorageProvider;
         private readonly IOptions<FunctionsHostingConfigOptions> _hostingConfigOptions;
         private readonly IScriptHostManager _scriptHostManager;
-        private readonly object _hostNameChangeLock = new object();
+        private readonly Func<Task> _syncTriggersAfterHostNameChange;
 
         private BlobClient _hashBlobClient;
-        private CancellationTokenSource _hostNameChangeDebounceCts;
-        private Task _hostNameChangeTask;
         private bool _disposed;
 
         public FunctionsSyncManager(IHostIdProvider hostIdProvider, IOptionsMonitor<ScriptApplicationHostOptions> applicationHostOptions, ILogger<FunctionsSyncManager> logger, IHttpClientFactory httpClientFactory, ISecretManagerProvider secretManagerProvider, IScriptWebHostEnvironment webHostEnvironment, IEnvironment environment, HostNameProvider hostNameProvider, IFunctionMetadataManager functionMetadataManager, IAzureBlobStorageProvider azureBlobStorageProvider, IOptions<FunctionsHostingConfigOptions> functionsHostingConfigOptions, IScriptHostManager scriptHostManager)
@@ -93,6 +87,8 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
 
             // Subscribe to hostname changes to trigger re-sync after slot swaps.
             // This ensures the ARM cache (invoke_url_template) is updated with the correct hostname.
+            Func<Task> syncAction = SyncTriggersAfterHostNameChangeAsync;
+            _syncTriggersAfterHostNameChange = syncAction.Debounce(milliseconds: 5000);
             _hostNameProvider.HostNameChanged += OnHostNameChanged;
         }
 
@@ -819,71 +815,58 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
         /// </summary>
         private void OnHostNameChanged(object sender, HostNameChangedEventArgs e)
         {
+            if (_disposed)
+            {
+                return;
+            }
+
             _logger.LogInformation("Hostname changed from '{PreviousHostName}' to '{NewHostName}'. Scheduling background sync triggers update.",
                 e.PreviousHostName, e.NewHostName);
 
-            // Cancel any pending debounced re-sync and start a new one.
-            // This ensures we wait for hostname to stabilize before re-syncing.
-            CancellationTokenSource cts;
-            lock (_hostNameChangeLock)
-            {
-                if (_disposed)
-                {
-                    return;
-                }
+            _ = _syncTriggersAfterHostNameChange();
+        }
 
-                _hostNameChangeDebounceCts?.Cancel();
-                _hostNameChangeDebounceCts?.Dispose();
-                _hostNameChangeDebounceCts = new CancellationTokenSource();
-                cts = _hostNameChangeDebounceCts;
+        /// <summary>
+        /// Executes the background sync triggers after hostname change has stabilized.
+        /// This method is wrapped by <see cref="FuncExtensions.Debounce"/> to prevent rapid re-syncs.
+        /// </summary>
+        private async Task SyncTriggersAfterHostNameChangeAsync()
+        {
+            if (_disposed)
+            {
+                return;
             }
 
-            _hostNameChangeTask = Task.Run(async () =>
+            _logger.LogInformation("Executing background sync triggers after hostname change stabilized to '{NewHostName}'.",
+                _hostNameProvider.Value);
+
+            try
             {
-                try
+                var result = await TrySyncTriggersAsync(isBackgroundSync: true);
+                if (result.Success)
                 {
-                    // Wait for debounce period to allow hostname to stabilize
-                    await Task.Delay(HostNameChangeDebounceDelayMs, cts.Token);
-
-                    _logger.LogInformation("Executing background sync triggers after hostname change stabilized to '{NewHostName}'.",
-                        _hostNameProvider.Value);
-
-                    var result = await TrySyncTriggersAsync(isBackgroundSync: true);
-                    if (result.Success)
-                    {
-                        _logger.LogInformation("Background sync triggers completed successfully after hostname change.");
-                    }
-                    else
-                    {
-                        _logger.LogWarning("Background sync triggers failed after hostname change: {Error}", result.Error);
-                    }
+                    _logger.LogInformation("Background sync triggers completed successfully after hostname change.");
                 }
-                catch (Exception ex) when (ex is TaskCanceledException or ObjectDisposedException)
+                else
                 {
-                    // Another hostname change occurred or Dispose was called; this sync was superseded
-                    _logger.LogDebug("Background sync triggers cancelled due to subsequent hostname change or disposal.");
+                    _logger.LogWarning("Background sync triggers failed after hostname change: {Error}", result.Error);
                 }
-                catch (Exception ex)
-                {
-                    _logger.LogError(ex, "Error during background sync triggers after hostname change.");
-                }
-            });
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex, "Error during background sync triggers after hostname change.");
+            }
         }
 
         public void Dispose()
         {
-            // Unsubscribe from hostname changes
-            _hostNameProvider.HostNameChanged -= OnHostNameChanged;
-
-            // Cancel and dispose of any pending debounced re-sync
-            lock (_hostNameChangeLock)
+            if (_disposed)
             {
-                _disposed = true;
-                _hostNameChangeDebounceCts?.Cancel();
-                _hostNameChangeDebounceCts?.Dispose();
-                _hostNameChangeDebounceCts = null;
+                return;
             }
 
+            _disposed = true;
+            _hostNameProvider.HostNameChanged -= OnHostNameChanged;
             _syncSemaphore.Dispose();
         }
 
--- a/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
@@ -809,7 +809,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
         }
 
         /// <summary>
-        /// Handles hostname changes (e.g., after slot swap) by triggering a debounced background re-sync.
+        /// Handles hostname changes (e.g., after slot swap) by triggering a debounced re-sync.
         /// This ensures the ARM cache (invoke_url_template) is updated with the correct hostname.
         /// The debounce prevents rapid re-syncs during hostname flip-flop that can occur during slot swaps.
         /// </summary>
@@ -820,16 +820,34 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
                 return;
             }
 
-            _logger.LogInformation("Hostname changed from '{PreviousHostName}' to '{NewHostName}'. Scheduling background sync triggers update.",
+            // During standby mode, IsSyncTriggersEnvironment returns false and SetTriggers would be a no-op.
+            // Skip scheduling to avoid misleading log messages.
+            if (_webHostEnvironment.InStandbyMode)
+            {
+                return;
+            }
+
+            _logger.LogInformation("Hostname changed from '{PreviousHostName}' to '{NewHostName}'. Scheduling sync triggers update.",
                 e.PreviousHostName, e.NewHostName);
 
             _ = _syncTriggersAfterHostNameChange();
         }
 
         /// <summary>
-        /// Executes the background sync triggers after hostname change has stabilized.
+        /// Executes a non-background (authoritative) sync triggers after hostname change has stabilized.
+        /// Uses <c>isBackgroundSync: false</c> to force a <c>SetTriggers</c> call without hash-based de-duplication,
+        /// ensuring the ARM cache (e.g., invoke_url_template) is updated even when the trigger payload is unchanged.
         /// This method is wrapped by <see cref="FuncExtensions.Debounce"/> to prevent rapid re-syncs.
         /// </summary>
+        /// <remarks>
+        /// Trade-off: Using <c>isBackgroundSync: false</c> means that if the site temporarily has no functions
+        /// (e.g., transient file system issue during swap), an empty payload will be sent to SetTriggers, which
+        /// could cause ARM to think the app has no triggers. This is an acceptable trade-off because:
+        /// 1. It is extremely unlikely for functions to disappear during a swap.
+        /// 2. The regular background sync timer will eventually re-sync with the correct payload.
+        /// 3. The alternative (using background sync) would bypass SetTriggers due to hash de-dup, defeating
+        ///    the purpose of this fix entirely.
+        /// </remarks>
         private async Task SyncTriggersAfterHostNameChangeAsync()
         {
             if (_disposed)
@@ -837,24 +855,24 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
                 return;
             }
 
-            _logger.LogInformation("Executing background sync triggers after hostname change stabilized to '{NewHostName}'.",
+            _logger.LogInformation("Executing sync triggers after hostname change stabilized to '{NewHostName}'.",
                 _hostNameProvider.Value);
 
             try
             {
-                var result = await TrySyncTriggersAsync(isBackgroundSync: true);
+                var result = await TrySyncTriggersAsync(isBackgroundSync: false);
                 if (result.Success)
                 {
-                    _logger.LogInformation("Background sync triggers completed successfully after hostname change.");
+                    _logger.LogInformation("Sync triggers completed successfully after hostname change.");
                 }
                 else
                 {
-                    _logger.LogWarning("Background sync triggers failed after hostname change: {Error}", result.Error);
+                    _logger.LogWarning("Sync triggers failed after hostname change: {Error}", result.Error);
                 }
             }
             catch (Exception ex)
             {
-                _logger.LogError(ex, "Error during background sync triggers after hostname change.");
+                _logger.LogError(ex, "Error during sync triggers after hostname change.");
             }
         }
 
--- a/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
@@ -50,6 +50,10 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
         private const string ManagedKubernetesBuildServiceName = "k8se-build-service";
         private const string ManagedKubernetesBuildServiceNamespace = "k8se-system";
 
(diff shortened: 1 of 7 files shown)


pagariyaalok:
## Additional context: Resolves #9908

This PR also "partially" addresses the feature request in #9908 — **"Slot Swap Feature - Send signal to function extension on hostname/routing swap completes."**

The new `HostNameProvider.HostNameChanged` event provides the signal that the Logic Apps team requested. The `WorkflowExtensionProvider` (and any other extension) can now subscribe to this event to re-register workers with the correct production hostname after a swap completes:

```csharp
hostNameProvider.HostNameChanged += (sender, args) =>
{
    // args.PreviousHostName — the old (now stale) hostname
    // args.NewHostName — the correct production hostname after swap
    // Re-register workers, refresh cached URLs, etc.
};
```

This eliminates the need for extensions to poll or rely on timing workarounds to detect hostname changes during slot swaps.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -800,8 +810,76 @@ public async Task<TriggersResult> GetTriggersAsync()
...
+                catch (TaskCanceledException)
+                {
+                    // Another hostname change occurred, this sync was superseded
+                    _logger.LogDebug("Background sync triggers cancelled due to subsequent hostname change.");
+                }
+                catch (Exception ex)
+                {
+                    _logger.LogError(ex, "Error during background sync triggers after hostname change.");
+                }
+            });
```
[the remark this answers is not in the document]
pagariyaalok: Done. Stored the task reference in _hostNameChangeTask field.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -800,8 +810,76 @@ public async Task<TriggersResult> GetTriggersAsync()
...
+                try
+                {
+                    // Wait for debounce period to allow hostname to stabilize
+                    await Task.Delay(HostNameChangeDebounceDelayMs, cts.Token);
+
+                    if (cts.Token.IsCancellationRequested)
+                    {
+                        return;
+                    }
+
```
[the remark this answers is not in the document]
pagariyaalok: Done. Removed the redundant cancellation check.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -800,8 +810,76 @@ public async Task<TriggersResult> GetTriggersAsync()
...
+                {
+                    _logger.LogError(ex, "Error during background sync triggers after hostname change.");
+                }
+            });
+        }
+
         public void Dispose()
         {
+            // Unsubscribe from hostname changes
+            _hostNameProvider.HostNameChanged -= OnHostNameChanged;
```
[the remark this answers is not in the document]
pagariyaalok: Done. Added _disposed flag with check in OnHostNameChanged to prevent orphaned tasks after Dispose.

src/WebJobs.Script/HostNameProvider.cs
```diff
@@ -56,14 +62,49 @@ public virtual void Synchronize(HttpRequest request, ILogger logger)
             if (!string.IsNullOrEmpty(hostNameHeaderValue) &&
                 string.Compare(Value, hostNameHeaderValue) != 0)
```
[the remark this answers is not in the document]
pagariyaalok: Done. Changed to string.Equals(Value, hostNameHeaderValue, StringComparison.OrdinalIgnoreCase).

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -800,8 +810,76 @@ public async Task<TriggersResult> GetTriggersAsync()
...
+            // Cancel any pending debounced re-sync and start a new one.
+            // This ensures we wait for hostname to stabilize before re-syncing.
+            lock (_hostNameChangeLock)
+            {
+                _hostNameChangeDebounceCts?.Cancel();
+                _hostNameChangeDebounceCts?.Dispose();
+                _hostNameChangeDebounceCts = new CancellationTokenSource();
+            }
+
+            var cts = _hostNameChangeDebounceCts;
```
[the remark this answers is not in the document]
pagariyaalok: Done. Moved cts capture inside the lock block as suggested.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -82,6 +88,10 @@ public FunctionsSyncManager(IHostIdProvider hostIdProvider, IOptionsMonitor<Scri
             _azureBlobStorageProvider = azureBlobStorageProvider;
             _hostingConfigOptions = functionsHostingConfigOptions;
             _scriptHostManager = scriptHostManager;
+
+            // Subscribe to hostname changes to trigger re-sync after slot swaps.
+            // This ensures the ARM cache (invoke_url_template) is updated with the correct hostname.
+            _hostNameProvider.HostNameChanged += OnHostNameChanged;
```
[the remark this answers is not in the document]
pagariyaalok: Done. Added 3 integration tests: HostNameChange_TriggersBackgroundSyncTriggers, HostNameChange_Debounces_MultipleChanges, and Dispose_CancelsHostNameChangeSyncAndUnsubscribes.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -800,8 +810,76 @@ public async Task<TriggersResult> GetTriggersAsync()
...
+                _hostNameChangeDebounceCts = new CancellationTokenSource();
+            }
+
+            var cts = _hostNameChangeDebounceCts;
+            _ = Task.Run(async () =>
+            {
+                try
+                {
+                    // Wait for debounce period to allow hostname to stabilize
+                    await Task.Delay(HostNameChangeDebounceDelayMs, cts.Token);
```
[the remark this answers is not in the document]
pagariyaalok: Done. Added ObjectDisposedException to the catch clause: catch (Exception ex) when (ex is TaskCanceledException or ObjectDisposedException).

pagariyaalok:
Hi team — all 6 CI checks (unit tests, SDL, CLA, dependency review) are green. 

Could you please trigger the **integration test pipeline** (`integration-tests.yml`) against this branch? The 3 new integration tests in `FunctionsSyncManagerTests.cs` (hostname change sync, debounce, dispose/cancel) pass locally but can't run in the public PR pipeline due to service connection / Key Vault requirements.

Thanks!

FinVamp1:
@pagariyaalok  let's see what Azure Functions Core thinks. I can see that you are honest about the limitations and this might help for the Logic App scenario but if no HTTP requests come in then we may not apply in all scenarios. WEBSITE_SWAP_WARMUP_PING_PATH should be called during Warmup of the Slot before the swap and may help with this.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -65,8 +69,12 @@ public class FunctionsSyncManager : IFunctionsSyncManager, IDisposable
         private readonly IAzureBlobStorageProvider _azureBlobStorageProvider;
         private readonly IOptions<FunctionsHostingConfigOptions> _hostingConfigOptions;
         private readonly IScriptHostManager _scriptHostManager;
+        private readonly object _hostNameChangeLock = new object();
 
         private BlobClient _hashBlobClient;
+        private CancellationTokenSource _hostNameChangeDebounceCts;
+        private Task _hostNameChangeTask;
```
fabiocav: The task is never being used (beyond assignment) and/or awaited on. We should be able to use a discard instead and not define a variable.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -800,8 +812,78 @@ public async Task<TriggersResult> GetTriggersAsync()
...
+                _hostNameChangeDebounceCts = new CancellationTokenSource();
+                cts = _hostNameChangeDebounceCts;
+            }
+
+            _hostNameChangeTask = Task.Run(async () =>
+            {
+                try
+                {
+                    // Wait for debounce period to allow hostname to stabilize
+                    await Task.Delay(HostNameChangeDebounceDelayMs, cts.Token);
```
fabiocav: We already have some common debouncing logic in the `FunctExtensions.Debounce` method. We should be able to use that instead.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -800,8 +812,78 @@ public async Task<TriggersResult> GetTriggersAsync()
...
+                catch (Exception ex)
+                {
+                    _logger.LogError(ex, "Error during background sync triggers after hostname change.");
+                }
+            });
+        }
+
         public void Dispose()
         {
+            // Unsubscribe from hostname changes
```
fabiocav: Use the newly introduced `_disposed` and return early.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -800,8 +812,78 @@ public async Task<TriggersResult> GetTriggersAsync()
...
+                try
+                {
+                    // Wait for debounce period to allow hostname to stabilize
+                    await Task.Delay(HostNameChangeDebounceDelayMs, cts.Token);
+
+                    _logger.LogInformation("Executing background sync triggers after hostname change stabilized to '{NewHostName}'.",
+                        _hostNameProvider.Value);
+
+                    var result = await TrySyncTriggersAsync(isBackgroundSync: true);
+                    if (result.Success)
```
fabiocav: nit: blank line before this condition.

test/WebJobs.Script.Tests.Integration/Management/FunctionsSyncManagerTests.cs
```diff
@@ -1305,6 +1306,99 @@ private IFileSystem CreateFileSystem(ScriptApplicationHostOptions hostOptions, s
...
 
+        [Fact]
+        public async Task HostNameChange_TriggersBackgroundSyncTriggers()
+        {
+            // Arrange - set up environment for sync triggers
+            using (var env = new TestScopedEnvironmentVariable(_vars))
+            {
+                ResetMockFileSystem();
+
+                // Act - simulate a hostname change (e.g., after slot swap)
```
fabiocav: nit: remove act / arrange / assert / verify comments in this file. Copilot should have caught that, as it is in the instructions :)

fabiocav (changes requested):

fabiocav:
Thanks @pagariyaalok ! This is a nice find!

fabiocav (commented):
Can you please also update the release notes with the change?

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
pagariyaalok: Good call — removed the `_hostNameChangeTask` field entirely and switched back to `_ = _syncTriggersAfterHostNameChange()`. The debounced method has a full try/catch internally, so fire-and-forget is safe here.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
pagariyaalok: Done — refactored to use `FuncExtensions.Debounce(milliseconds: 5000)` which simplified things significantly. Removed the custom lock/CTS/Task.Run debounce logic entirely.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
pagariyaalok: Done — `Dispose()` now uses the `_disposed` guard with early return, matching the pattern used elsewhere in the codebase.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
pagariyaalok: Fixed — added blank line before the condition.

test/WebJobs.Script.Tests.Integration/Management/FunctionsSyncManagerTests.cs
pagariyaalok: Ha, fair point! Removed all the Act/Arrange/Assert/Verify comments from the test file. Lesson learned 😄

pagariyaalok:
Done — added release notes entry in afb89a3:

```
- Fix stale hostname in ARM cache after slot swap by adding event-based re-sync when `HostNameProvider` detects a hostname change (#11629)
```

pagariyaalok:
@fabiocav All review comments addressed, release notes updated, and tests verified locally (8/8 unit tests pass, 34/38 integration tests pass — the 4 failures are pre-existing on `dev`). Ready for your re-review when you get a chance. Thanks!

cjaliaga (commented):
The issue with running this as a background trigger sync (`isBackgroundSync: true`) is that it’s intentionally conservative/optimistic: it uses a stored payload hash to de-dup calls and can legitimately short-circuit (no-op) **if the hash of the payload is unchanged** compared to what’s stored in the Storage Account (it can also skip empty payloads). That’s great for periodic, best-effort consistency, but it’s **not ideal when the goal is to force a `SetTriggers` to remediate a broken state**, because we may never reach `SetTriggers` if the payload is unchanged.

A non-background sync, on the other hand, is an on-demand, authoritative push: it **always calls `SetTriggers`** (no hash-based de-dup / no skipping), which makes it the better fit for “force re-apply triggers” scenarios. The trade-off is that it may end up syncing an **empty payload**, and then a subsequent background sync might not fix things because **only the background path computes/persists the hash** (not a typical situation, but it can happen).

More info: https://github.com/Azure/azure-functions-host/issues/11330

fabiocav:
@cjaliaga yes, that behavior is by design. Are you proposing a change to this to force a state sync? The host won't have any awareness of the ARM state, so it would need to force a refresh on slot swap operations. 

@pagariyaalok did you have a chance to test the specialization flow? Based on the logic, this new event will also be triggered in those scenarios. Asking because, if that is the case, the event alone cannot be used a a signal to force a refresh, as it would happen on every scale out operation.

pagariyaalok:
@fabiocav Great callout — you're right that the original implementation would have fired on every specialization/scale-out, which is definitely not what we want.

I've updated the logic so the `HostNameChanged` event is only raised when the previous hostname was already set (non-null). During specialization, `_hostName` is null and `Value` either returns null or falls back to the environment variable — so `previousHostName` ends up null/empty, and we skip raising the event. This means:

- **Specialization/scale-out**: first request sets hostname from null → real value → **no event fired**
- **Slot swap**: hostname changes from one real value to another → **event fires, triggers sync**

I also added a standby mode guard in `OnHostNameChanged` — if we're still in standby, we skip scheduling entirely since `IsSyncTriggersEnvironment` would reject it anyway.

Per @cjaliaga's feedback, I also switched to `isBackgroundSync: false` so that `SetTriggers` is always called (no hash de-dup), which is what we need to force the ARM cache refresh after a swap.

Changes are in the latest push (31b8a58). Let me know if you'd like me to test anything specific around the specialization path.

Testing the latest changes.

pagariyaalok:
@cjaliaga Thanks for the detailed explanation — that really helped clarify the difference between the two sync paths. You're absolutely right that background sync would just no-op here since the payload hash wouldn't change after a swap (same functions, same bindings — only the hostname embedded in `invoke_url_template` differs).

I've switched to `isBackgroundSync: false` so we always call `SetTriggers` and force the ARM cache update. I'm aware of the trade-off you mentioned about potentially syncing an empty payload, but that scenario is pretty unlikely during a swap and the regular background sync would eventually correct it anyway. I've documented this trade-off in the code comments as well.

Thanks again for pointing this out!

fabiocav:
@pagariyaalok as this is a pretty sensitive area, so I want to make sure we're validating what was stated above:

- About the following:
>  _hostName is null and Value either returns null
Did you validate this on a placeholder site? I want to make sure that is indeed true and we don't have `WEBSITE_HOSTNAME` set in that environment. If we do, even if `_hostName` was currently `null`, it would introduce an issue where an unrelated change could perform that initialization and change the behavior of your check.

Either way, the current guard needs to be resilient to _hostName being non-null at the time of the first Synchronize call.

- Can you confirm the exact scenario you are capturing? With the flow you're describing, the production slot should not be impacted, but the slot would be updated with production state.
  - Swaps are supposed to ping the site, exercising the synchronization flow for the host. Is there a platform bug causing issues with that? Have we traced end-to-end to see whether something is failing?
  - If a ping is happening, background syncs should pick up the correct, updated hostname.
  - If the site doesn't have storage, the platform logic that pings the site is also supposed to be calling synctriggers as part of a swap. Have you identified whether that is happening?
  - The slot is the one that ends up with the invalid state, correct? I guess the main site could potentially be missing keys for newly introduced functions, but shouldn't have invalid keys for the existing one. 
  
 
 Would you be able to share the sequence of events leading to the issue including the platform components?

mathewc:
@pagariyaalok The logic performed by Controller post swap is here: https://msazure.visualstudio.com/One/_git/AAPT-Antares-Websites?path=/src/Hosting/AdministrationService/Microsoft.Web.Hosting.Administration.WebCommon/Controllers/WebCloudController.cs&version=GBdev&line=18413&lineEnd=18420&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents.

The platform does ping both slots with the new hostname before issuing a sync triggers on both slots. So after the swap is fully complete, everything should be consistent. Are you talking about an intermediate inconsistent state before these operations complete, where the staging slot (with production settings) does a background sync before it gets the platform ping to update its hostname? That might be because the staging slot is in an intermediate state where it's _transitioning_ into the production slot - it has prod settings + code, but stage hostname still because swap isn't complete. If a background sync happens at this time, background sync will use the prod storage cached sync hash to determine whether a sync is needed, and it will because the payload will be different because the site/slot name vars are for the slot.

What I need to help review this and help determine the right fix is a Kusto query for a repro as per https://github.com/Azure/azure-functions-host/issues/7710 (or whatever repro you want me to focus on) so we can trace all the requests in Controller/FE/Host to figure out the exact sequence of events, when things are inconsistent, and when they're corrected. You can share the Kusto query repro on Teams and we can analyze together.

pagariyaalok:
@mathewc, @fabiocav  - I replied to the internal thread with query details. Thank you !

fabiocav:
Closing this as this PR is now stale. We can reopen if still needed.

PR #11655, opened 2026-03-13
[Compute Separation] M2: ConnectedWorkerChannel + channel manager + invocation dispatcher

### LINKED ISSUE #11643: [Compute Separation] M2: ConnectedWorkerChannel + channel manager + invocation dispatcher
New channel type for externally-connected workers. Inherits `WorkerChannelBase` — no process management. The gRPC stream IS the worker lifecycle.

**Dependencies**: M1 (#11641)

### Deliverables

| File | Purpose |
|---|---|
| `ExternalWorkers/ConnectedWorkerChannel.cs` | Channel subclass — no `IWorkerProcess`, no process spawn/terminate |
| `ExternalWorkers/ConnectedWorkerChannelFactory.cs` | Creates `ConnectedWorkerChannel` instances |
| `ExternalWorkers/IConnectedWorkerChannelManager.cs` | Interface with `AddChannel`, `GetChannels`, `WaitForChannelAsync`, `ShutdownChannelAsync` |
| `ExternalWorkers/ConnectedWorkerChannelManager.cs` | Implementation with `TaskCompletionSource`-based `WaitForChannelAsync` |
| `ExternalWorkers/ConnectedWorkerInvocationDispatcher.cs` | `IFunctionInvocationDispatcher` that routes to connected channels, no restart logic |

### Acceptance Criteria

- [ ] `ConnectedWorkerChannel` can be constructed and processes the gRPC protocol identically to `GrpcWorkerChannel` (minus process lifecycle)
- [ ] `WorkerProcess` returns null, `WorkerProcessId` returns -1
- [ ] `ConnectedWorkerChannelManager.WaitForChannelAsync()` blocks until a channel is added, with configurable timeout
- [ ] `ConnectedWorkerInvocationDispatcher` routes invocations to ready channels
- [ ] `RestartWorkerAsync` returns `false` (no-op for external workers)
- [ ] Unit tests for all new classes

brettsam:
## Summary

New types for the separated compute model — channel, manager, factory, and dispatcher for externally-connected workers. All additive, no DI wiring (that comes in M4).

Resolves #11643

## New Files

| File | Purpose |
|---|---|
| `ExternalWorkers/ConnectedWorkerChannel.cs` | Inherits `WorkerChannelBase`, no process management — gRPC stream IS the lifecycle |
| `ExternalWorkers/ConnectedWorkerChannelFactory.cs` | DI factory for creating channels |
| `ExternalWorkers/IConnectedWorkerChannelManager.cs` | Interface with `WaitForChannelAsync` |
| `ExternalWorkers/ConnectedWorkerChannelManager.cs` | `ConcurrentDictionary` + `TaskCompletionSource` blocking |
| `ExternalWorkers/ConnectedWorkerInvocationDispatcher.cs` | Routes invocations to connected channels, no restart logic |
| `Workers/ExternalWorkers/ConnectedWorkerChannelManagerTests.cs` | 8 unit tests |
| `Workers/ExternalWorkers/ConnectedWorkerInvocationDispatcherTests.cs` | 5 unit tests |

## Testing

- Full solution builds with 0 errors
- 85/85 GrpcWorkerChannel tests pass (unchanged)
- 13/13 new ExternalWorkers tests pass
- 3,228/3,230 full unit suite (2 pre-existing CPU-count failures)

### DIFF
--- a/src/WebJobs.Script.Grpc/ExternalWorkers/ConnectedWorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/ExternalWorkers/ConnectedWorkerChannel.cs
@@ -1,0 +1,64 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Script.Config;
+using Microsoft.Azure.WebJobs.Script.Diagnostics;
+using Microsoft.Azure.WebJobs.Script.Eventing;
+using Microsoft.Azure.WebJobs.Script.Http;
+using Microsoft.Azure.WebJobs.Script.Workers;
+using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
+using Microsoft.Azure.WebJobs.Script.Workers.SharedMemoryDataTransfer;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.WebJobs.Script.Grpc.ExternalWorkers
+{
+    /// <summary>
+    /// A worker channel for a worker that connected inbound — the host did not spawn a process.
+    /// Lifecycle: the gRPC stream IS the lifecycle; disconnection → WorkerErrorEvent.
+    /// </summary>
+    internal sealed class ConnectedWorkerChannel : WorkerChannelBase
+    {
+        internal ConnectedWorkerChannel(
+            string workerId,
+            IScriptEventManager eventManager,
+            IScriptHostManager hostManager,
+            RpcWorkerConfig workerConfig,
+            ILogger logger,
+            IMetricsLogger metricsLogger,
+            IEnvironment environment,
+            IOptionsMonitor<ScriptApplicationHostOptions> applicationHostOptions,
+            ISharedMemoryManager sharedMemoryManager,
+            IOptions<WorkerConcurrencyOptions> workerConcurrencyOptions,
+            IOptions<FunctionsHostingConfigOptions> hostingConfigOptions,
+            IHttpProxyService httpProxyService)
+            : base(workerId, eventManager, hostManager, workerConfig, logger, metricsLogger,
+                   attemptCount: 0, environment, applicationHostOptions, sharedMemoryManager,
+                   workerConcurrencyOptions, hostingConfigOptions, httpProxyService)
+        {
+        }
+
+        /// <inheritdoc/>
+        public override IWorkerProcess WorkerProcess => null;
+
+        /// <inheritdoc/>
+        public override Task StartWorkerProcessAsync(CancellationToken cancellationToken)
+        {
+            BeginInboundProcessing(startStreamTimeout: TimeSpan.FromSeconds(30));
+            return Task.CompletedTask;
+        }
+
+        /// <inheritdoc/>
+        protected override void OnDisposing()
+        {
+        }
+
+        /// <inheritdoc/>
+        protected override void DisposeWorkerResources()
+        {
+        }
+    }
+}
--- a/src/WebJobs.Script.Grpc/ExternalWorkers/ConnectedWorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/ExternalWorkers/ConnectedWorkerChannel.cs
@@ -20,7 +20,7 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc.ExternalWorkers
     /// A worker channel for a worker that connected inbound — the host did not spawn a process.
     /// Lifecycle: the gRPC stream IS the lifecycle; disconnection → WorkerErrorEvent.
     /// </summary>
-    internal sealed class ConnectedWorkerChannel : WorkerChannelBase
+    internal sealed class ConnectedWorkerChannel : WorkerChannel
     {
         internal ConnectedWorkerChannel(
             string workerId,
@@ -49,11 +49,6 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc.ExternalWorkers
         {
             BeginInboundProcessing(startStreamTimeout: TimeSpan.FromSeconds(30));
             return Task.CompletedTask;
-        }
-
-        /// <inheritdoc/>
-        protected override void OnDisposing()
-        {
         }
 
         /// <inheritdoc/>
--- a/src/WebJobs.Script.Grpc/ExternalWorkers/ConnectedWorkerChannelFactory.cs
+++ b/src/WebJobs.Script.Grpc/ExternalWorkers/ConnectedWorkerChannelFactory.cs
@@ -1,0 +1,69 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Azure.WebJobs.Script.Config;
+using Microsoft.Azure.WebJobs.Script.Diagnostics;
+using Microsoft.Azure.WebJobs.Script.Eventing;
+using Microsoft.Azure.WebJobs.Script.Http;
+using Microsoft.Azure.WebJobs.Script.Workers;
+using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
+using Microsoft.Azure.WebJobs.Script.Workers.SharedMemoryDataTransfer;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.WebJobs.Script.Grpc.ExternalWorkers
+{
+    /// <summary>
+    /// Factory for creating <see cref="ConnectedWorkerChannel"/> instances.
+    /// </summary>
+    internal class ConnectedWorkerChannelFactory
+    {
+        private readonly IScriptEventManager _eventManager;
+        private readonly IScriptHostManager _hostManager;
+        private readonly IEnvironment _environment;
+        private readonly IOptionsMonitor<ScriptApplicationHostOptions> _applicationHostOptions;
+        private readonly ISharedMemoryManager _sharedMemoryManager;
+        private readonly IOptions<WorkerConcurrencyOptions> _workerConcurrencyOptions;
+        private readonly IOptions<FunctionsHostingConfigOptions> _hostingConfigOptions;
+        private readonly IHttpProxyService _httpProxyService;
+        private readonly ILoggerFactory _loggerFactory;
+        private readonly IMetricsLogger _metricsLogger;
+
+        public ConnectedWorkerChannelFactory(
+            IScriptEventManager eventManager,
+            IScriptHostManager hostManager,
+            IEnvironment environment,
+            IOptionsMonitor<ScriptApplicationHostOptions> applicationHostOptions,
+            ISharedMemoryManager sharedMemoryManager,
+            IOptions<WorkerConcurrencyOptions> workerConcurrencyOptions,
+            IOptions<FunctionsHostingConfigOptions> hostingConfigOptions,
+            IHttpProxyService httpProxyService,
+            ILoggerFactory loggerFactory,
+            IMetricsLogger metricsLogger)
+        {
+            _eventManager = eventManager;
+            _hostManager = hostManager;
+            _environment = environment;
+            _applicationHostOptions = applicationHostOptions;
+            _sharedMemoryManager = sharedMemoryManager;
+            _workerConcurrencyOptions = workerConcurrencyOptions;
+            _hostingConfigOptions = hostingConfigOptions;
+            _httpProxyService = httpProxyService;
+            _loggerFactory = loggerFactory;
+            _metricsLogger = metricsLogger;
+        }
+
+        /// <summary>
+        /// Creates a new <see cref="ConnectedWorkerChannel"/> for an externally-connected worker.
+        /// </summary>
+        public ConnectedWorkerChannel Create(string workerId, RpcWorkerConfig workerConfig)
+        {
+            var logger = _loggerFactory.CreateLogger($"Worker.{workerId}");
+            return new ConnectedWorkerChannel(
+                workerId, _eventManager, _hostManager, workerConfig,
+                logger, _metricsLogger, _environment, _applicationHostOptions,
+                _sharedMemoryManager, _workerConcurrencyOptions,
+                _hostingConfigOptions, _httpProxyService);
+        }
+    }
+}
--- a/src/WebJobs.Script.Grpc/ExternalWorkers/ConnectedWorkerChannelManager.cs
+++ b/src/WebJobs.Script.Grpc/ExternalWorkers/ConnectedWorkerChannelManager.cs
@@ -1,0 +1,71 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
+
+namespace Microsoft.Azure.WebJobs.Script.Grpc.ExternalWorkers
+{
+    /// <summary>
+    /// Manages <see cref="ConnectedWorkerChannel"/> instances for externally-connected workers.
+    /// Thread-safe. Supports blocking waits for the first available channel.
+    /// </summary>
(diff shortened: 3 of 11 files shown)


src/WebJobs.Script.Grpc/FunctionInvocationDispatcher.cs
```diff
@@ -0,0 +1,149 @@
...
+            var channels = await GetReadyChannelsAsync(invocationContext: null);
+            var drainTasks = channels.Select(c => c.DrainInvocationsAsync()).ToList();
+
+            if (drainTasks.Count > 0)
+            {
+                Task completedTask = await Task.WhenAny(
+                    Task.WhenAll(drainTasks),
+                    Task.Delay(ShutdownTimeout));
+
+                if (completedTask is not Task<Task>)
```
liliankasem: Flagged by AI: this makes sense to me

> Task.Delay() returns Task (not Task<Task>), and Task.WhenAll() also returns Task. The is Task<Task> check will never be true

The suggestion being made is to check against the actual timeout task i.e.

```csharp
Task drainAll = Task.WhenAll(drainTasks);
Task timeoutTask = Task.Delay(ShutdownTimeout);
Task completedTask = await Task.WhenAny(drainAll, timeoutTask);

if (completedTask == timeoutTask)
{
    Logger.LogDebug("Draining invocations from worker channels timed out during shutdown");
}
else
{
    Logger.LogDebug("Draining invocations from worker channels completed during shutdown");
}
```

liliankasem: Something else noted is 

> Task.WhenAny doesn't observe exceptions from the losing task. If Task.WhenAll(drainTasks) fails after the timeout wins, that exception will go unobserved. Depending on your app, you may want to attach a continuation to swallow or log it

src/WebJobs.Script.Grpc/ExternalWorkers/ConnectedWorkerChannel.cs
```diff
@@ -0,0 +1,59 @@
...
+        {
+        }
+
+        /// <inheritdoc/>
+        public override IWorkerProcess WorkerProcess => null;
+
+        /// <inheritdoc/>
+        public override Task StartWorkerProcessAsync(CancellationToken cancellationToken)
+        {
+            BeginInboundProcessing(startStreamTimeout: TimeSpan.FromSeconds(30));
```
liliankasem, line 50: Do we want this value to be configurable?

src/WebJobs.Script.Grpc/ExternalWorkers/ConnectedWorkerChannelManager.cs
```diff
@@ -0,0 +1,71 @@
...
+namespace Microsoft.Azure.WebJobs.Script.Grpc.ExternalWorkers
+{
+    /// <summary>
+    /// Manages <see cref="ConnectedWorkerChannel"/> instances for externally-connected workers.
+    /// Thread-safe. Supports blocking waits for the first available channel.
+    /// </summary>
+    internal class ConnectedWorkerChannelManager : IConnectedWorkerChannelManager
+    {
+        private readonly ConcurrentDictionary<string, ConnectedWorkerChannel> _channels = new();
+        private TaskCompletionSource<IRpcWorkerChannel> _firstChannelReady = new();
```
liliankasem: AI review

#### ConnectedWorkerChannelManager._firstChannelReady is never reset

Once the TaskCompletionSource is completed by the first AddChannel call, WaitForChannelAsync will always return that first channel from the TCS — even if that channel was later shut down via ShutdownChannelAsync. The FirstOrDefault fast path checks IsChannelReadyForInvocations(), but the TCS fallback does not. After shutdown of the only channel, a new WaitForChannelAsync call would return the disposed channel from the TCS.

Consider resetting _firstChannelReady in ShutdownChannelAsync when the dictionary becomes empty, or validating the returned channel is still registered.

src/WebJobs.Script.Grpc/ExternalWorkers/ConnectedWorkerInvocationDispatcher.cs
```diff
@@ -0,0 +1,138 @@
...
+        /// <inheritdoc/>
+        public override void PreShutdown()
+        {
+        }
+
+        /// <inheritdoc/>
+        public override void Dispose()
+        {
+            State = FunctionInvocationDispatcherState.Disposing;
+            State = FunctionInvocationDispatcherState.Disposed;
```
liliankasem, line 125: Any reason we're not going straight to Disposed if no work is happening in between these states?

src/WebJobs.Script.Grpc/FunctionInvocationDispatcher.cs
```diff
@@ -0,0 +1,149 @@
...
+        /// Gets the timeout for draining invocations during shutdown.
+        /// Subclasses can override to use a configured value.
+        /// </summary>
+        protected virtual TimeSpan ShutdownTimeout => DefaultShutdownTimeout;
+
+        /// <inheritdoc/>
+        public virtual FunctionInvocationDispatcherState State { get; set; }
+
+        /// <inheritdoc/>
+        public virtual int ErrorEventsThreshold { get; set; }
```
liliankasem: Do both of these need a public setter?

src/WebJobs.Script.Grpc/ExternalWorkers/ConnectedWorkerChannelFactory.cs
```diff
@@ -0,0 +1,69 @@
...
+using Microsoft.Azure.WebJobs.Script.Workers.SharedMemoryDataTransfer;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.WebJobs.Script.Grpc.ExternalWorkers
+{
+    /// <summary>
+    /// Factory for creating <see cref="ConnectedWorkerChannel"/> instances.
+    /// </summary>
+    internal class ConnectedWorkerChannelFactory
```
liliankasem, line 19: Do we want an interface for this for testability?

src/WebJobs.Script.Grpc/FunctionInvocationDispatcher.cs
brettsam: I've just modified this to have this look exactly like it does in dev branch... that was my original intent but the AI took some liberties that I've now adjusted back.

src/WebJobs.Script.Grpc/ExternalWorkers/ConnectedWorkerInvocationDispatcher.cs
brettsam: Main reason was that I wanted to make sure that we don't miss this other state should we add something in between them. Also, I'm trying to think through state transitions... and it feels weird to skip one even though no one is listening. There's not much harm here so I'm going to leave it.

src/WebJobs.Script.Grpc/ExternalWorkers/ConnectedWorkerChannel.cs
brettsam: Tracked in #11664. In the separated compute model there is no worker.config on the host side, so this will likely become a platform-level configuration knob. Will address in a follow-up.

src/WebJobs.Script.Grpc/ExternalWorkers/ConnectedWorkerChannelFactory.cs
brettsam: We may eventually when the wire-up happens in a later milestone but I'll leave it off for now.

src/WebJobs.Script.Grpc/ExternalWorkers/ConnectedWorkerChannelManager.cs
brettsam: Yeah I actually purposefully ignored this one b/c I didn't think it mattered. Turns out it was important and it's been refactored and tests were added. Also added locks for potential races.

src/WebJobs.Script.Grpc/FunctionInvocationDispatcher.cs
brettsam: Good catch — these were `private set` on `dev` and got widened to `public set` to support `virtual` overrides. Changed to `protected set` to maintain encapsulation.

src/WebJobs.Script.Grpc/ExternalWorkers/ConnectedWorkerInvocationDispatcher.cs
```diff
@@ -0,0 +1,138 @@
...
+
+        /// <inheritdoc/>
+        public override void PreShutdown()
+        {
+        }
+
+        /// <inheritdoc/>
+        public override void Dispose()
+        {
+            State = FunctionInvocationDispatcherState.Disposing;
```
liliankasem, line 124: AI review raised by multiple models

The `State` getter is overridden to compute state dynamically from channel readiness, but `Dispose()` writes to the base class auto-property backing field via the inherited setter. Since the getter never reads that backing field, the `State = Disposing` and `State = Disposed` assignments are dead code. After `Dispose()`, `State` returns `Initializing` (because channels are gone), never `Disposed`.

The base class declares `public virtual FunctionInvocationDispatcherState State { get; protected set; }`. The subclass overrides only the getter. The inherited setter writes to the auto-property backing field, which the overridden getter ignores entirely.

src/WebJobs.Script.Grpc/ExternalWorkers/ConnectedWorkerChannelManager.cs
```diff
@@ -0,0 +1,90 @@
...
+
+            using var cts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+            cts.CancelAfter(timeout);
+
+            try
+            {
+                await _channelAvailable.Task.WaitAsync(cts.Token);
+
+                return _channels.Values.FirstOrDefault(c => c.IsChannelReadyForInvocations())
+                    ?? _channels.Values.First();
```
liliankasem: this could throw InvalidOperationException

```suggestion
                    ?? _channels.Values.FirstOrDefault();
```

src/WebJobs.Script.Grpc/ExternalWorkers/ConnectedWorkerChannelManager.cs
```diff
@@ -0,0 +1,90 @@
...
+            {
+                return ready;
+            }
+
+            using var cts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+            cts.CancelAfter(timeout);
+
+            try
+            {
+                await _channelAvailable.Task.WaitAsync(cts.Token);
```
liliankasem: Shouldn't this also use _channelAvailableLock?

src/WebJobs.Script.Grpc/ExternalWorkers/ConnectedWorkerInvocationDispatcher.cs
```diff
@@ -0,0 +1,138 @@
...
+                    .Any(c => c.IsChannelReadyForInvocations());
+
+                return anyReady
+                    ? FunctionInvocationDispatcherState.Initialized
+                    : FunctionInvocationDispatcherState.Initializing;
+            }
+        }
+
+        /// <inheritdoc/>
+        public override int ErrorEventsThreshold => int.MaxValue;
```
liliankasem: AI review:

The external dispatcher sets `ErrorEventsThreshold` to `int.MaxValue`, but `WorkerFunctionInvoker` uses `(_functionDispatcher.ErrorEventsThreshold + 1) * timeoutSeconds` to decide how long to wait for the dispatcher to become ready. With `int.MaxValue`, that arithmetic overflows to a negative value, so the "wait for worker readiness" path becomes an immediate timeout and calls `StopApplication()` on the first invocation that arrives before any connected worker is ready.

**Evidence:**
- `WorkerFunctionInvoker` computes the wait using `(_functionDispatcher.ErrorEventsThreshold + 1) * ...` (`WorkerFunctionInvoker.cs:126-137`)
- `Utility.DelayAsync(int timeoutSeconds, ...)` converts that negative timeout to a `TimeSpan` that exits immediately (`Utility.cs:193-210`)

src/WebJobs.Script.Grpc/ExternalWorkers/ConnectedWorkerInvocationDispatcher.cs
brettsam: Simplified this greatly. No need to override getter at all here and removed virtual properties.

src/WebJobs.Script.Grpc/ExternalWorkers/ConnectedWorkerChannelManager.cs
brettsam: This whole method got revisited as some reviews I did also called out races. Now it'll wait until a timeout and check for null, etc. Will never return a null channel but may timeout. We have a "hardening" milestone that will likely do some more here in the future but this should work for now.

src/WebJobs.Script.Grpc/ExternalWorkers/ConnectedWorkerInvocationDispatcher.cs
brettsam: Just set this to 3 for now... it may become irrelevant in a future milestone.

src/WebJobs.Script.Grpc/ExternalWorkers/ConnectedWorkerChannelManager.cs
brettsam: Fixed — the entire method is now a retry loop that reads and resets the TCS inside the lock. No more stale reads.

src/WebJobs.Script.Grpc/ExternalWorkers/ConnectedWorkerChannelManager.cs
```diff
@@ -0,0 +1,99 @@
...
+                    Task waitTask;
+                    lock (_channelAvailableLock)
+                    {
+                        var channel = _channels.Values.FirstOrDefault();
+                        if (channel is not null)
+                        {
+                            return channel;
+                        }
+
+                        _channelAvailable = new TaskCompletionSource();
```
mathewc: If two callers enter WaitForChannelAsync concurrently when no channels exist, each caller creates a new TaskCompletionSource, overwriting the previous one (_channelAvailable). The first caller captures a waitTask from a TCS that gets orphaned — it will never be signaled and can only exit via timeout.

// Caller A locks, creates TCS-A, captures waitTask-A, releases lock
// Caller B locks, creates TCS-B (overwrites TCS-A), captures waitTask-B, releases lock
// AddChannel signals TCS-B → Caller B wakes up; Caller A blocks forever

Fix: Either (a) don't replace the TCS if it's already unsignaled — just reuse it, or (b) use a SemaphoreSlim / Channel that supports multiple waiters, or (c) document single-waiter semantics and enforce it.

src/WebJobs.Script.Grpc/ExternalWorkers/ConnectedWorkerInvocationDispatcher.cs
```diff
@@ -0,0 +1,127 @@
...
+            _channelManager = channelManager ?? throw new ArgumentNullException(nameof(channelManager));
+            State = FunctionInvocationDispatcherState.Default;
+            ErrorEventsThreshold = 3;
+        }
+
+        /// <inheritdoc/>
+        protected override Task<IEnumerable<IRpcWorkerChannel>> GetReadyChannelsAsync(ScriptInvocationContext invocationContext)
+        {
+            IEnumerable<IRpcWorkerChannel> channels = _channelManager.GetChannels().Values
+                .Where(c => c.IsChannelReadyForInvocations());
```
mathewc: Should you ToList here to avoid multiple enumerations by callers?

src/WebJobs.Script.Grpc/ExternalWorkers/ConnectedWorkerChannelManager.cs
brettsam: That line snuck in from somewhere. There's really no scenario where the tcs should be re-created here (i ran through it with AI also) -- so just removed this line. With the add/remove/wait operations all under lock, they should all get a proper view of this and do the right thing now. Added a test to ensure one doesn't hang in the future also.

src/WebJobs.Script.Grpc/ExternalWorkers/ConnectedWorkerInvocationDispatcher.cs
brettsam: Yep. Fixed.

mathewc (approved):

liliankasem (approved):

PR #11656, opened 2026-03-16
Add check for worker tag propagation to improve performance

satvu:

Updates a `null` check both null / empty mapfield to avoid execution context sync/change when no worker tags are propagated.

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
@@ -16,7 +16,6 @@ using System.Threading.Tasks;
 using System.Threading.Tasks.Dataflow;
 using Google.Protobuf.Collections;
 using Google.Protobuf.WellKnownTypes;
-using Microsoft.Azure.AppService.Proxy.Common.Extensions;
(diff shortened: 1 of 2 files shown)


kshyju (approved):
Do we need any test updates?

satvu:
> Do we need any test updates?

We could add a test to make sure [some step] doesn't happen if there's an empty/null worker tag collection but I didn't think this really required that coverage/check.

src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs
```diff
@@ -16,6 +16,7 @@
 using System.Threading.Tasks.Dataflow;
 using Google.Protobuf.Collections;
 using Google.Protobuf.WellKnownTypes;
+using Microsoft.Azure.AppService.Proxy.Common.Extensions;
```
RohitRanjanMS: This using `Microsoft.Azure.AppService.Proxy.Common.Extensions` looks unnecessary — MapField<string, string>already has a native .Count property via IDictionary<TKey, TValue>.

RohitRanjanMS (changes requested):

RohitRanjanMS (approved):

PR #11658, opened 2026-03-16
feat: auto-handle HEAD requests on GET triggers and return 405 for non-GET/HEAD routes

### LINKED ISSUE #11625: HTTP GET triggers should auto-handle HEAD requests (or return clearer error)
## Summary

When an Azure Function has an HTTP trigger configured for GET only, HEAD requests return a generic 404 from the routing layer rather than being auto-handled or returning a clear "method not allowed" error. This causes significant debugging confusion.

## Problem

A function defined as:

```csharp
[Function("RedirectUrl")]
public async Task<IActionResult> RedirectUrl(
    [HttpTrigger(AuthorizationLevel.Anonymous, "get", Route = "r/{shortCode}")] HttpRequest req,
    string shortCode)
```

**Behavior:**
- `GET /api/r/abc123` → **302 Found** ✅ (works)
- `HEAD /api/r/abc123` → **404 Not Found** ❌ (confusing)

The 404 comes from the routing layer before the function is invoked, with no indication that HEAD is unsupported.

## Why This Matters

1. **Common testing pattern**: Developers frequently use `curl -I` (HEAD) to test redirects and check response headers without downloading the body
2. **Debugging time waste**: I spent ~20 minutes debugging why redirects "weren't working" before realizing HEAD vs GET was the issue
3. **HTTP spec expectation**: RFC 7231 states "The HEAD method is identical to GET except that the server MUST NOT send a message body" - many developers expect GET endpoints to auto-handle HEAD

### Option A: Auto-handle HEAD for GET triggers (Preferred)
When a function accepts GET, automatically handle HEAD by:
1. Invoking the function with the HEAD request
2. Returning headers only (strip body)

This matches behavior of most web frameworks (ASP.NET Core, Express, Flask, etc.)

### Option B: Return 405 Method Not Allowed
If HEAD is not supported, return:
```
HTTP/1.1 405 Method Not Allowed
Allow: GET, POST
```

This clearly indicates the method isn't supported and which methods are.

### Option C: Better documentation
At minimum, document that HEAD requests require explicit `"head"` in the trigger:
```csharp
[HttpTrigger(AuthorizationLevel.Anonymous, "get", "head", Route = "...")]
```

## Environment

- Azure Functions Flex Consumption
- .NET 10 isolated worker (net10.0)
- Functions Runtime: 4.1046.x

## Reproduction

1. Create function with `[HttpTrigger(..., "get", ...)]`
2. Deploy to Azure
3. `curl -I https://<func>.azurewebsites.net/api/<route>` → 404
4. `curl https://<func>.azurewebsites.net/api/<route>` → 200/302 (works)

The inconsistency between HEAD and GET is unexpected and causes debugging confusion.

MO2k4:
Fixes #11625

## Summary

- **GET triggers now auto-handle HEAD requests** — `HEAD /api/myFunc` no longer returns a misleading 404; it executes the GET function normally and returns the same status/headers with no body.
- **Non-GET/non-HEAD routes return 405** — `HEAD /api/postOnlyFunc` now returns `405 Method Not Allowed` with an `Allow` header listing the supported methods, instead of 404.

## Architecture (three-layer fix, host-side only)

1. **`WebScriptHostHttpRoutesManager`** — at route registration time, `"head"` is appended to the HTTP method constraint whenever `"get"` is present; a HEAD-only shadow route with a sentinel name (`$head_not_allowed:<METHODS>`) is registered for functions that accept neither GET nor HEAD.
2. **`ScriptRouteHandler`** — detects the sentinel name prefix, sets `IHeadNotAllowedFeature` on the request context, and returns immediately.
3. **`FunctionInvocationMiddleware`** — after `_next(context)` returns (which includes routing), checks for `IHeadNotAllowedFeature` to return 405 + `Allow` header; for HEAD requests that reach function execution, swaps `Response.Body` with `Stream.Null` to suppress the body while preserving status and headers.

> **Note:** The `IHeadNotAllowedFeature` check must occur *after* `_next(context)` because `ScriptRouteHandler` runs later in the ASP.NET pipeline than the middleware.

## Test Plan

- [x] Unit tests for route helper methods (BuildConstraintMethods, ShouldRegisterHeadNotAllowedRoute, BuildHeadNotAllowedSentinelName) — 23 tests
- [x] Unit tests for sentinel detection in ScriptRouteHandler — 4 tests
- [x] Unit tests for FunctionInvocationMiddleware (405 response, HEAD body suppression, GET body preservation) — 4 tests
- [x] **End-to-end integration tests** against real TestServer + Azurite — 4 tests:
  - `HEAD` on GET-only function (`HttpTrigger-Redirect`) → same status as GET, empty body
  - `HEAD` on POST-only function (`HttpTrigger-Dynamic`) → 405 + `Allow: POST`
  - `HEAD` on non-existent route → 404
  - `GET` on GET-only function → body preserved (regression check)
- [x] Build verified — solution builds with 0 errors
- [x] All 31 tests pass

### DIFF
--- a/src/WebJobs.Script.WebHost/Features/IHeadNotAllowedFeature.cs
+++ b/src/WebJobs.Script.WebHost/Features/IHeadNotAllowedFeature.cs
@@ -1,0 +1,19 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Features;
+
+internal interface IHeadNotAllowedFeature
+{
+    string AllowedMethods { get; }
+}
+
+internal sealed class HeadNotAllowedFeature : IHeadNotAllowedFeature
+{
+    public HeadNotAllowedFeature(string allowedMethods)
+    {
+        AllowedMethods = allowedMethods;
+    }
+
+    public string AllowedMethods { get; }
+}
--- a/src/WebJobs.Script.WebHost/Middleware/FunctionInvocationMiddleware.cs
+++ b/src/WebJobs.Script.WebHost/Middleware/FunctionInvocationMiddleware.cs
@@ -34,6 +34,15 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Middleware
         // TODO: Confirm that only HttpTrigger requests would flow through here
         public async Task Invoke(HttpContext context)
         {
+            // HEAD request on a route that doesn't support HEAD/GET → 405 Method Not Allowed.
+            var headNotAllowed = context.Features.Get<IHeadNotAllowedFeature>();
+            if (headNotAllowed is not null)
+            {
+                context.Response.StatusCode = StatusCodes.Status405MethodNotAllowed;
+                context.Response.Headers.Allow = headNotAllowed.AllowedMethods;
+                return;
+            }
+
             if (_next != null)
             {
                 await _next(context);
--- a/src/WebJobs.Script.WebHost/Middleware/FunctionInvocationMiddleware.cs
+++ b/src/WebJobs.Script.WebHost/Middleware/FunctionInvocationMiddleware.cs
@@ -3,6 +3,7 @@
 
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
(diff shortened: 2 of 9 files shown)


satvu:
Hi @MO2k4, we've flagged the issue as a breaking change, so unfortunately, we cannot take any PRs for this at the moment without a deeper internal review of its priority.

MO2k4:
@satvu ok and thanks for the update 😊

PR #11662, opened 2026-03-18
Add Golang Worker NuGet package reference

ahmedmuhsin:
## Changes

- Add `eng/build/Workers.Golang.props` with PackageReference to `Microsoft.Azure.Functions.GolangWorker`
- Auto-imported by the existing glob in `Workers.props`        

## Related PRs
- azure-functions-golang-worker: https://github.com/Azure/azure-functions-golang-worker/pull/15
- AAPT-Antares-Functions-Docker: ahmedmuhsin/add-golang-dockerfiles

### DIFF
--- a/eng/build/Workers.Golang.props
+++ b/eng/build/Workers.Golang.props
@@ -1,0 +1,7 @@
+<Project>
+
+  <ItemGroup>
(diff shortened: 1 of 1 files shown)


fabiocav:
@ahmedmuhsin for new workers, particularly since this would be scoped to Flex only, we can pull them directly into the image, so there's no need to add that as a host dependency.

PR #11577, opened 2026-01-22
Implement app capabilities feature

satvu:
### Issue describing the changes in this PR

resolves #11574 

- Example branch of webjobs extension this works e2e with [here](https://github.com/Azure/azure-webjobs-sdk-extensions/compare/dev...satvu/prototype-capabilities-registration).
- Example branch of dotnet worker that works e2e with this [here](https://github.com/Azure/azure-functions-dotnet-worker/compare/main...satvu/prototype-capabilities)

### Pull request checklist

**IMPORTANT**: Currently, changes must be backported to the `in-proc` branch to be included in Core Tools and non-Flex deployments.

* [x] Backporting to the `in-proc` branch is not required
    * Otherwise: Link to backporting PR
* [ ] My changes **do not** require documentation changes
    * [x] Otherwise: Documentation issue linked to PR
* [ ] My changes **should not** be added to the release notes for the next release
    * [ ] Otherwise: I've added my notes to `release_notes.md`
* [x] My changes **do not** need to be backported to a previous version
    * [ ] Otherwise: Backport tracked by issue/PR #issue_or_pr
* [x] My changes **do not** require diagnostic events changes
    * Otherwise: I have added/updated all related diagnostic events and their documentation (Documentation issue linked to PR)
* [x] I have added all required tests (Unit tests, E2E tests)

### Additional Comments

The two new endpoints added (`admin/capabilities` and `admin/capabilities/{name}`) can be used by partners or customers to quickly query for an app's capabilities/features and related values (endpoints, names, etc.). Capabilities include values that can be set by extensions, language workers, customers, and the host.

One example: At the ARM level, the `admin/capabilities` endpoint can be called and the values can be stored in ARG for later purposes (such as querying/filtering apps for a dashboard).

#### Related PRs:

- Protobuf PR: https://github.com/Azure/azure-functions-language-worker-protobuf/pull/100
- Abstractions PR: TODO

PR #11682, opened 2026-04-01
Add `AppCapabilitiesOptions` to `WebJobs.Script.Abstractions`

satvu:

related to #11577 - this includes just the changes for `WebJobs.Script.Abstractions` which has its own package/release steps.

### Pull request checklist


src/WebJobs.Script.Abstractions/Directory.Version.props
```diff
@@ -1,6 +1,6 @@
 <Project>
   <PropertyGroup>
-    <VersionPrefix>1.0.4</VersionPrefix>
+    <VersionPrefix>1.0.5</VersionPrefix>
```
liliankasem: Shouldn't this be `1.1.0` since we are adding a new public type?

src/WebJobs.Script.Abstractions/AppCapabilities/AppCapabilitiesOptions.cs
```diff
@@ -0,0 +1,112 @@
...
+
+namespace Microsoft.Azure.WebJobs.Script.AppCapabilities
+{
+    /// <summary>
+    /// Represents a collection of application capabilities, allowing for flexible access and management of capability
+    /// values as key-value pairs.
+    /// </summary>
+    /// <remarks>The keys in the dictionary are case-insensitive, enabling developers to access capability
+    /// values without concern for key casing.</remarks>
+    public sealed class AppCapabilitiesOptions : IDictionary<string, string>
```
liliankasem, line 19: Does this need to be public?

src/WebJobs.Script.Abstractions/AppCapabilities/AppCapabilitiesOptions.cs
```diff
@@ -0,0 +1,112 @@
...
+            return Capabilities.Remove(item);
+        }
+
+        /// <inheritdoc />
+        public bool TryGetValue(string key, out string value)
+        {
+            return Capabilities.TryGetValue(key, out value!);
+        }
+    }
+}
```
liliankasem: missing new line end of file

src/WebJobs.Script.Abstractions/AppCapabilities/AppCapabilitiesOptions.cs
```diff
@@ -0,0 +1,112 @@
...
+        /// <inheritdoc />
+        public bool Remove(KeyValuePair<string, string> item)
+        {
+            return Capabilities.Remove(item);
+        }
+
+        /// <inheritdoc />
+        public bool TryGetValue(string key, out string value)
+        {
+            return Capabilities.TryGetValue(key, out value!);
```
liliankasem: The signature out string value (out string value) tells callers that this value will never be null, but when the key doesn't exist, Dictionary.TryGetValue sets value = null. The `!` you used here on value is just suppressing the compiler warning. You could do this instead

```
 public bool TryGetValue(string key, [MaybeNullWhen(false)] out string value)
 {
     return Capabilities.TryGetValue(key, out value);
 }
```

src/WebJobs.Script.Abstractions/AppCapabilities/AppCapabilitiesOptions.cs
satvu: Yes, for use with WebJobs extensions.

src/WebJobs.Script.Abstractions/AppCapabilities/AppCapabilitiesOptions.cs
satvu: This package targets `netstandard2.0` and `[MaybeNullWhen(false)]` is only available for .NET6 and higher. Let me try a workaround for this ...

Edit: discussing offline as adding the attribute plus subsequent changes (`string?`) strays away from implementing `IDictionary` exactly

liliankasem: I think you can stick to the IDictionary contract and still have safeguards in place by having a null check instead of suppressing it

satvu: The suppression pattern is in multiple `dotnet` repos: https://github.com/dotnet/sdk/blob/3ff71030f075942d8c768280552be5ba69a37464/src/Microsoft.CodeAnalysis.NetAnalyzers/src/Utilities/FlowAnalysis/FlowAnalysis/Framework/DataFlow/DictionaryAnalysisData.cs#L139

including `sdk`, `roslyn`, etc so I think we can follow their lead here.

liliankasem (approved):

src/WebJobs.Script.Abstractions/release_notes.md
```diff
@@ -4,6 +4,6 @@
 - My change description (#PR/#issue)
 -->
 
-### Microsoft.Azure.WebJobs.Script.Abstractions 1.0.4
+### Microsoft.Azure.WebJobs.Script.Abstractions 1.0.5
```
kshyju: Is the version number here correct?

src/WebJobs.Script.Abstractions/AppCapabilities/MaybeNullWhenAttribute.cs
```diff
@@ -0,0 +1,34 @@
...
+#if NETSTANDARD2_0
+
+namespace System.Diagnostics.CodeAnalysis
+{
+    /// <summary>
+    /// Specifies that when a method returns <see cref="ReturnValue"/>,
+    /// the parameter may be null even if the corresponding type disallows it.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Parameter, Inherited = false)]
+    internal sealed class MaybeNullWhenAttribute : Attribute
```
kshyju, line 13: Do we really need the polyfill implementation of this attribute?

src/WebJobs.Script.Abstractions/AppCapabilities/AppCapabilitiesOptions.cs
```diff
@@ -0,0 +1,115 @@
...
+
+        /// <inheritdoc />
+        public bool Remove(KeyValuePair<string, string> item)
+        {
+            return Capabilities.Remove(item);
+        }
+
+        /// <inheritdoc />
+#pragma warning disable CS8767 // Nullability of reference types in type of parameter doesn't match implicitly implemented member (possibly because of nullability attributes).
+        public bool TryGetValue(string key, [MaybeNullWhen(false)] out string? value)
```
kshyju: So this `public` method uses an `internal` attribute. Is that okay/confusing to do? Do we really need this attribute decoration? I am wondering instead of adding the polyfill implementation, what if we conditionally include the attribute decoration part when the code is running in a supported version?

src/WebJobs.Script.Abstractions/AppCapabilities/AppCapabilitiesOptions.cs
```diff
@@ -0,0 +1,115 @@
...
+
+namespace Microsoft.Azure.WebJobs.Script.AppCapabilities
+{
+    /// <summary>
+    /// Represents a collection of application capabilities, allowing for flexible access and management of capability
+    /// values as key-value pairs.
+    /// </summary>
+    /// <remarks>The keys in the dictionary are case-insensitive, enabling developers to access capability
+    /// values without concern for key casing.</remarks>
+    public sealed class AppCapabilitiesOptions : IDictionary<string, string>
```
kshyju, line 19: Wondering we should unseal this as we are exposing this now for others to use.

satvu: There aren't currently any extensibility scenarios for this so I think it's fine to leave this as more restrictive (for now) and if there's a need in the future we can loosen this restriction.

src/WebJobs.Script.Abstractions/AppCapabilities/AppCapabilitiesOptions.cs
jviau: This is a common polyfill pattern. It is for the roslyn compiler more than anything else. This csproj only targets netstandard2.0 and this is a compile time (not runtime) operation. So, no dynamic detection possibility here.

src/WebJobs.Script.Abstractions/AppCapabilities/AppCapabilitiesOptions.cs
jviau: Yeah, `sealed` is perfectly fine for public types. In fact, that is when it makes _most_ sense as you want to communicate "don't bother inheriting this" to consumers.

src/WebJobs.Script.Abstractions/AppCapabilities/MaybeNullWhenAttribute.cs
jviau: While I am all for polyfils and encourage it for netstandard2.0, I would actually remove this _specific_ usage. The reason is that it is only used for the `TryGet` implementation of `IDictionary`. But in netstandard2.0 version of `IDictionary`, the interface itself does not have nullability on its types. Hence the compiler warning when adding `?` to `string?` in the method declaration.

src/WebJobs.Script.Abstractions/AppCapabilities/AppCapabilitiesOptions.cs
jviau: You have a mismatch of generic type and implementation, which is probably the source of your issue.

Your interface declaration is `string`, but you have `string?` here. Either add `?` to the declaration:

``` csharp
: IDictionary<string, string?>
```

or remove `?` here.

jviau (changes requested):

jviau (approved):

liliankasem (approved):

PR #11690, opened 2026-04-03
Implement notification pattern for sync trigger operations

kamperiadis:

Implement a new notification pattern for Flex Consumption sync trigger operations

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
@@ -92,7 +92,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
             }
         }
 
-        public async Task<TriggersOperationResult> TrySyncTriggersAsync(bool isBackgroundSync = false)
+        public virtual async Task<TriggersOperationResult> TrySyncTriggersAsync(bool isBackgroundSync = false)
         {
             var result = new TriggersOperationResult
             {
@@ -304,7 +304,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
             return _hashBlobClient;
         }
 
-        private async Task<SyncTriggersPayload> GetSyncTriggersPayload()
+        protected async Task<SyncTriggersPayload> GetSyncTriggersPayload()
         {
             PrepareSyncTriggers();
 
--- a/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
@@ -92,7 +92,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
             }
         }
 
-        public virtual async Task<TriggersOperationResult> TrySyncTriggersAsync(bool isBackgroundSync = false)
+        public async Task<TriggersOperationResult> TrySyncTriggersAsync(bool isBackgroundSync = false)
         {
             var result = new TriggersOperationResult
             {
@@ -304,7 +304,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
             return _hashBlobClient;
         }
 
-        protected async Task<SyncTriggersPayload> GetSyncTriggersPayload()
+        private async Task<SyncTriggersPayload> GetSyncTriggersPayload()
         {
             PrepareSyncTriggers();
 
@@ -730,7 +730,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
 
         // This function will call POST https://{app}.azurewebsites.net/operation/settriggers with the content
         // of triggers. It'll verify app ownership using a site token valid for 5 minutes. It should be plenty.
-        private async Task<(bool Success, string ErrorMessage)> SetTriggersAsync(string content)
+        protected virtual async Task<(bool Success, string ErrorMessage)> SetTriggersAsync(string content)
         {
             // sanitize the content before logging
             var sanitizedContent = JToken.Parse(content);
--- a/src/WebJobs.Script.WebHost/Management/MeshServiceClient.cs
+++ b/src/WebJobs.Script.WebHost/Management/MeshServiceClient.cs
@@ -143,14 +143,13 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
             response.EnsureSuccessStatusCode();
         }
 
-        public async Task NotifyTriggersChanged(string contentHash)
+        public async Task NotifyTriggersChanged()
         {
             _logger.LogDebug("Posting triggers changed notification to appserver.");
 
             var response = await SendAsync(
             [
-                KeyValuePair.Create(Operation, "update-triggers-timestamp"),
-                KeyValuePair.Create("content-hash", contentHash)
+                KeyValuePair.Create(Operation, "update-triggers-timestamp")
             ]);
 
             response.EnsureSuccessStatusCode();
--- a/src/WebJobs.Script.WebHost/Management/MeshServiceClient.cs
+++ b/src/WebJobs.Script.WebHost/Management/MeshServiceClient.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
@@ -140,6 +140,19 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
             };
 
             var response = await SendAsync(operation);
+            response.EnsureSuccessStatusCode();
+        }
+
+        public async Task NotifyTriggersChanged(string contentHash)
+        {
+            _logger.LogDebug("Posting triggers changed notification to appserver.");
+
+            var response = await SendAsync(
+            [
+                KeyValuePair.Create(Operation, "update-triggers-timestamp"),
+                KeyValuePair.Create("content-hash", contentHash)
+            ]);
+
             response.EnsureSuccessStatusCode();
         }
 
--- a/src/WebJobs.Script.WebHost/Management/SignalBasedFunctionsSyncManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/SignalBasedFunctionsSyncManager.cs
@@ -3,8 +3,6 @@
 
 using System;
 using System.Net.Http;
-using System.Security.Cryptography;
-using System.Text;
 using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Host.Executors;
 using Microsoft.Azure.WebJobs.Host.Storage;
@@ -40,13 +38,11 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
         }
 
         // Overrides the default settriggers HTTP call to instead notify the mesh service
-        // with a content hash of the triggers payload
         protected override async Task<(bool Success, string ErrorMessage)> SetTriggersAsync(string content)
         {
             try
             {
-                string contentHash = ComputeContentHash(content);
-                await _meshServiceClient.NotifyTriggersChanged(contentHash);
+                await _meshServiceClient.NotifyTriggersChanged();
 
                 return (true, null);
             }
@@ -57,13 +53,6 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
 
                 return (false, message);
             }
-        }
-
-        private static string ComputeContentHash(string content)
-        {
-            byte[] hash = SHA256.HashData(Encoding.UTF8.GetBytes(content));
-
-            return Convert.ToHexStringLower(hash);
         }
     }
 }
--- a/src/WebJobs.Script.WebHost/Management/SignalBasedFunctionsSyncManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/SignalBasedFunctionsSyncManager.cs
@@ -1,0 +1,85 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Net.Http;
+using System.Security.Cryptography;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host.Executors;
+using Microsoft.Azure.WebJobs.Host.Storage;
+using Microsoft.Azure.WebJobs.Script.Config;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
+{
+    public class SignalBasedFunctionsSyncManager : FunctionsSyncManager
+    {
+        private readonly IMeshServiceClient _meshServiceClient;
+        private readonly ILogger _logger;
+        private readonly IScriptWebHostEnvironment _webHostEnvironment;
+        private readonly IEnvironment _environment;
+
+        public SignalBasedFunctionsSyncManager(
+            IHostIdProvider hostIdProvider,
+            IOptionsMonitor<ScriptApplicationHostOptions> applicationHostOptions,
+            ILogger<SignalBasedFunctionsSyncManager> logger,
+            IHttpClientFactory httpClientFactory,
+            ISecretManagerProvider secretManagerProvider,
+            IScriptWebHostEnvironment webHostEnvironment,
+            IEnvironment environment,
+            HostNameProvider hostNameProvider,
+            IFunctionMetadataManager functionMetadataManager,
+            IAzureBlobStorageProvider azureBlobStorageProvider,
+            IOptions<FunctionsHostingConfigOptions> functionsHostingConfigOptions,
+            IScriptHostManager scriptHostManager,
+            IMeshServiceClient meshServiceClient)
+            : base(hostIdProvider, applicationHostOptions, logger, httpClientFactory, secretManagerProvider, webHostEnvironment, environment, hostNameProvider, functionMetadataManager, azureBlobStorageProvider, functionsHostingConfigOptions, scriptHostManager)
+        {
+            _meshServiceClient = meshServiceClient;
+            _logger = logger;
+            _webHostEnvironment = webHostEnvironment;
+            _environment = environment;
+        }
+
+        public override async Task<TriggersOperationResult> TrySyncTriggersAsync(bool isBackgroundSync = false)
+        {
+            var result = new TriggersOperationResult
+            {
+                Success = true
+            };
+
+            if (!IsSyncTriggersEnvironment(_webHostEnvironment, _environment))
+            {
+                result.Success = false;
+                result.Error = "Invalid environment for SyncTriggers operation.";
+                _logger.LogWarning(result.Error);
+
+                return result;
+            }
+
+            try
+            {
+                var payload = await GetSyncTriggersPayload();
+                string contentHash = ComputeContentHash(payload.Content);
+                await _meshServiceClient.NotifyTriggersChanged(contentHash);
+            }
+            catch (Exception ex)
+            {
+                result.Success = false;
+                result.Error = "Failed to notify triggers changed.";
+                _logger.LogError(ex, result.Error);
+            }
+
+            return result;
+        }
+
+        private static string ComputeContentHash(string content)
+        {
+            byte[] hash = SHA256.HashData(Encoding.UTF8.GetBytes(content));
+
+            return Convert.ToHexStringLower(hash);
+        }
+    }
+}
--- a/src/WebJobs.Script.WebHost/Management/SignalBasedFunctionsSyncManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/SignalBasedFunctionsSyncManager.cs
@@ -18,8 +18,6 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
     {
         private readonly IMeshServiceClient _meshServiceClient;
         private readonly ILogger _logger;
-        private readonly IScriptWebHostEnvironment _webHostEnvironment;
-        private readonly IEnvironment _environment;
 
         public SignalBasedFunctionsSyncManager(
             IHostIdProvider hostIdProvider,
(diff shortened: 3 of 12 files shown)


src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
```diff
@@ -40,6 +40,7 @@ public static class EnvironmentSettingNames
         public const string ContainerName = "CONTAINER_NAME";
         public const string WebsitePodName = "WEBSITE_POD_NAME";
         public const string LegionServiceHost = "LEGION_SERVICE_HOST";
+        public const string TriggersSyncSignalEnabled = "TRIGGERS_SYNC_SIGNAL_ENABLED";
```
mathewc: I didn't see mention of this env setting in any design docs. When will this mode be enabled - is it tied to FDM or ATOL? Depending on your answer, I'm wondering if instead of per feature settings like this you don't need just a single var that indicates we're in that mode - that can then be used by other features you need to make conditional.

kamperiadis: This is not tied to FDM since this doesn't impact the deployment flow but rather the non-deployment sync trigger operations (e.g. key rotation, customer sync triggers call). This is more ATOL-related and we already have env settings for ATOL detection, but the intention for this env setting is to ensure that the platform has the implementation to support this.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -107,6 +109,22 @@ public async Task<TriggersOperationResult> TrySyncTriggersAsync(bool isBackgroun
                 return result;
             }
 
+            if (_environment.IsTriggersSyncSignalEnabled())
```
jviau: Can we move this enable/disable check to mesh client.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -107,6 +109,22 @@ public async Task<TriggersOperationResult> TrySyncTriggersAsync(bool isBackgroun
...
+            {
+                try
+                {
+                    await _meshServiceClient.NotifyTriggersChanged();
```
jviau: I'd prefer a better decoupling between sync manager and specific environment behaviors. Can we use a general event callback and have mesh client listen to that event?

Or some other decoupling instead of having the manager know to call mesh client in these circumstances.

src/WebJobs.Script.WebHost/Management/MeshServiceClient.cs
```diff
@@ -139,6 +139,18 @@ private async Task PublishActivities(IEnumerable<ContainerFunctionExecutionActiv
             response.EnsureSuccessStatusCode();
         }
 
+        public async Task NotifyTriggersChanged()
+        {
+            _logger.LogDebug("Posting triggers changed notification to appserver.");
+
+            var response = await SendAsync(new[]
```
jviau: use collection expressions:

```suggestion
            var response = await SendAsync([ KeyValuePair.Create(Operation, "update-triggers-timestamp") ]);
```

src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
jviau: We need to start being consistent in prefixing our env variables so they are effectively namespaced, thus avoiding collisions with any customer-defined variables. Can we prefix this with `FUNCTIONS_`?

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
mathewc: I was contemplating a similar comment, but hadn't made it yet because for better or worse, IMeshServiceClient is being used by other services in a similar way. I was thinking what event approaches might look like. Another option would be a custom IFunctionsSyncManager implementation that gets used in this mode, but that'd require a class hierarchy to be added such that the new impl. inherits behavior and only overrides a protected SetTriggersAsync for example to send its notification.

src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
mathewc: So the behavior is tied to ATOL then - I.e. on stamps with ATOL enabled, you'll be injecting this into the container context always? Then my comment still applies - shouldn't we have a single env var that indicates we're running in the ATOL mode. What ATOL detection do we have already in host? I couldn't find any.

kamperiadis: For ATOL detection, we need to check `IsFlexConsumptionSku() && IsConsumptionOnLegion()` but checking if we are in ATOL mode wouldn't really help here since I am concerned that we'll release the host with a version of Legion that might not yet have the backing changes (e.g. delays in reviews for Legion PR).  And then we are left in a state where the host makes the notification assuming that the platform is handling the triggers when in reality that is not the case.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
kamperiadis: I personally favor the custom IFunctionsSyncManager implementation over the event callback approach just because in this case, there would be a single subscriber listening for this event and if there were an error we might want to return that to the caller. 

To build on what you mentioned @mathewc, wouldn't we want to override all of TrySyncTriggersAsync? If we only override SetTriggersAsync, we would still be generating the triggers payload which would be essentially thrown away.

src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
kamperiadis: Done

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
kamperiadis: Just refactored to create a new SignalBasedFunctionsSyncManager.cs

src/WebJobs.Script.WebHost/Management/SignalBasedFunctionsSyncManager.cs
jviau, line 1: Sorry to go back and forth, but I don't think the derived type is a good idea. It makes for a more challenging long-term maintenance. If we need to add another conditional sync behavior, where do we add it? This type? Base class? What if we have a combination of conditions? Do we need a derived type _per_ combination?

If you don't want to extract an event pattern, at the very minimum add the environment check condition to mesh service and have the sync manager call `NotifyTriggersChanged` unconditionally (and it just no-ops when it is not the correct environment).

kamperiadis: I am not sure that I can move the `_environment.IsTriggersSyncSignalEnabled()` check to `NotifyTriggersChanged()` since I want to return early and skip the rest of the `TrySyncTriggersAsync` method if we are in that `FUNCTIONS_TRIGGERS_SYNC_SIGNAL_ENABLED` mode. 

The rest of the logic in `TrySyncTriggersAsync` (i.e. generating triggers payload, calling settriggers) would just be thrown away work for this notification path. Eventually, the platform will receive the notification and fetch the triggers payload but for this method, we shouldn't run anything else other than notify that the triggers have been updated.

src/WebJobs.Script.WebHost/Management/SignalBasedFunctionsSyncManager.cs
```diff
@@ -0,0 +1,74 @@
...
+            try
+            {
+                await _meshServiceClient.NotifyTriggersChanged();
+            }
+            catch (Exception ex)
+            {
+                result.Success = false;
+                result.Error = "Failed to notify triggers changed.";
+                _logger.LogError(ex, result.Error);
+            }
```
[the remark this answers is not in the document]
kamperiadis: The semaphore is not really necessary for the SignalBasedFunctionsSyncManager since there is no shared mutable state that we need to sync. We are just making POST requests to notify of triggers updates.

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
```diff
@@ -144,7 +144,15 @@ public static void AddWebJobsScriptHost(this IServiceCollection services, IConfi
...
             // Management services
-            services.AddSingleton<IFunctionsSyncManager, FunctionsSyncManager>();
+            if (SystemEnvironment.Instance.IsTriggersSyncSignalEnabled())
+            {
+                services.AddSingleton<IFunctionsSyncManager, SignalBasedFunctionsSyncManager>();
+            }
+            else
+            {
+                services.AddSingleton<IFunctionsSyncManager, FunctionsSyncManager>();
+            }
```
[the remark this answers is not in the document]
kamperiadis: This is not applicable since env setting will only be set by the platform specifically for Flex Consumption on Legion.

test/WebJobs.Script.Tests.Integration/Management/MeshServiceClientTests.cs
```diff
@@ -261,6 +261,30 @@ private static bool ContainsActivities(HttpRequestMessage request, params Contai
...
+            {
+                StatusCode = HttpStatusCode.OK
+            });
+
+            await _meshServiceClient.NotifyTriggersChanged();
+
+            _handlerMock.Protected().Verify<Task<HttpResponseMessage>>("SendAsync", Times.Once(),
+                ItExpr.Is<HttpRequestMessage>(r => IsNotifyTriggersChangedRequest(r)),
+                ItExpr.IsAny<CancellationToken>());
+        }
```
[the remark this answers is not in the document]
kamperiadis: Fixed

test/WebJobs.Script.Tests.Integration/Management/SignalBasedFunctionsSyncManagerTests.cs
```diff
@@ -0,0 +1,101 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Net.Http;
```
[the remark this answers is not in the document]
kamperiadis: This is not unused

src/WebJobs.Script.WebHost/Management/SignalBasedFunctionsSyncManager.cs
jviau: ahh I see. I was missing that it skipped the rest of the sync action. Okay the derived class approach is fine with me. Thanks for explaining!

jviau (approved):

kamperiadis:
@jviau @mathewc Thank you so much for all your feedback. I refactored the changes based on our last discussion:
1. We want to also have background sync triggers be handled via this notification pattern
2. We do not want to attempt to fetch the triggers payload and update the db if there have been no updates. For this reason, we might want to send the hash so that in the platform we can determine if there were no updates    
3. We want to reduce blob dependency if possible

Can you please take another look? Thank you so much.

src/WebJobs.Script.WebHost/Management/SignalBasedFunctionsSyncManager.cs
```diff
@@ -0,0 +1,85 @@
...
+            IMeshServiceClient meshServiceClient)
+            : base(hostIdProvider, applicationHostOptions, logger, httpClientFactory, secretManagerProvider, webHostEnvironment, environment, hostNameProvider, functionMetadataManager, azureBlobStorageProvider, functionsHostingConfigOptions, scriptHostManager)
+        {
+            _meshServiceClient = meshServiceClient;
+            _logger = logger;
+            _webHostEnvironment = webHostEnvironment;
+            _environment = environment;
+        }
+
+        public override async Task<TriggersOperationResult> TrySyncTriggersAsync(bool isBackgroundSync = false)
```
fabiocav: Is this the right method to override? There's some amount of duplication happening here and maintenance moving forward will also have to account for this derived type.

If the difference between the two implementations is purely the way the platform is notified of the change, can we be more granular about what we override (e.g., either override `SetTriggersAsync` or introduce a method for this purpose)?

src/WebJobs.Script.WebHost/Management/SignalBasedFunctionsSyncManager.cs
```diff
@@ -0,0 +1,85 @@
...
+                result.Success = false;
+                result.Error = "Invalid environment for SyncTriggers operation.";
+                _logger.LogWarning(result.Error);
+
+                return result;
+            }
+
+            try
+            {
+                var payload = await GetSyncTriggersPayload();
```
fabiocav: There's a problem with this implementation (part of that would be addressed with the suggestion I made above). `GetSyncTriggersPayload` is guarded in the base type, but it's called without any synchronization here. The call to that method operates on shared mutable state, and removing that guard will lead to problems. One clear example is the call chain into `GetSyncTriggersPayload()` → `PrepareSyncTriggers()` → 
`_secretManagerProvider.Current.ClearCache()`.

src/WebJobs.Script.WebHost/Management/SignalBasedFunctionsSyncManager.cs
```diff
@@ -0,0 +1,85 @@
...
+                string contentHash = ComputeContentHash(payload.Content);
+                await _meshServiceClient.NotifyTriggersChanged(contentHash);
```
fabiocav: Are we delegating change validation to the platform? The `NotifyTriggersChanged` method name imply a change, but the current implementation will be calling this for every instance and relying on the platform to determine whether there has been a change. Where is the platform persisting the hash for comparison?

I realize you wanted to avoid blob dependencies, but I'm concerned this is leading to additional requirements already handled by the base implementation. If you were to preserve the current behavior and simply change the notification mechanism, that would address problems like the unnecessary notifications coming out of the host.

fabiocav (changes requested):

src/WebJobs.Script.WebHost/Management/SignalBasedFunctionsSyncManager.cs
kamperiadis: Thank you for your feedback @fabiocav 

I have updated the class so that we only override the `SetTriggersAsync` method instead of overriding `TrySyncTriggersAsync`

kamperiadis: It might be wasteful that we are generating the triggers payload when we are not really using it for anything other than just sending the notification. 

Question - in the design meeting, it was mentioned that we might want to send the hash to Legion for dedupe purposes. But wouldn't it be considered expensive to generate the triggers payload and calculate the hash to send it over in the notification?

fabiocav: >Question - in the design meeting, it was mentioned that we might want to send the hash to Legion for dedupe purposes. But wouldn't it be considered expensive to generate the triggers payload and calculate the hash to send it over in the notification?

It's unnecessary work, and it leads to other inefficiencies from the platform side as well. I don't think we should be sending this hash, given the notification semantics.

jviau: @kamperiadis I had a chat with @fabiocav and he makes a good point to leave this for later with this recent change. We are already doing all the blob storage checks before we get into your new set triggers handler.

Hopefully it won't be hard to change the notification payload at a later date if we want to remove storage dependency in the future.

fabiocav: BTW, the feedback was based on the original implementation that was already building the payload and calculating the hash. 

One thing you can also do here is pass the hash, when available, into the `SetTriggersAsync` call

fabiocav: But I question using that hash at this point.

jviau: Given how we may need more significant changes to this signal in the future for multi-revision, and we don't have a path to removing the storage dependency just yet, my original motivation for adding this is no longer applicable. So I agree we can remove it.

kamperiadis: Thank you @jviau and @fabiocav. There is one edge case that I thought of that would be caught by sending the hash. Let's say the customer took an action where we rely on the background sync triggers to catch the update in the triggers payload. 

1.  Host startup triggered the background sync triggers call
2. The host notified the platform that there is an update to the triggers payload by sending the notification and then we updated the blob with the new hash
3. The platform received the notification, but by the time it attempted to call GET admin/host/triggers, the worker was deleted
4. Because the hash was already updated in the blob, in the next worker startup, we will not reattempt to notify the platform to fetch the triggers. We will only reattempt to get the triggers payload the next time we go through this flow for a non-background-sync-triggers scenario (e.g. key rotation, customer-facing sync triggers call)

Sending the hash in the notification would mitigate this in that the platform would compare this hash with what is in the db and reattempt to get the triggers payload. 

But at the same time, my understanding is that we consider the background sync triggers as a best-effort attempt to have the triggers payload updated in the db (please correct me if I'm wrong) so in that case, we might not want to always send the hash just to mitigate the scenario I described above, especially since the probability of it happening is very small. What are your thoughts on this?

kamperiadis: Also, I was giving it more thought, we only do the hash check when `isBackgroundSync == true`, so for key rotation, customer-facing sync call, etc., for the SignalBasedFunctionsSyncManager class, it is wasteful to be generating the triggers payload when we are relying on the platform to fetch it later and this payload is just thrown away. So wouldn't it be better if I did the following in SignalBasedFunctionsSyncManager:

```

public override async Task<TriggersOperationResult> TrySyncTriggersAsync(bool isBackgroundSync = false)
        {
            if (isBackgroundSync)
            {
                return await base.TrySyncTriggersAsync(isBackgroundSync);
            }

            var result = new TriggersOperationResult
            {
                Success = true
            };

            if (!IsSyncTriggersEnvironment(_webHostEnvironment, _environment))
            {
                result.Success = false;
                result.Error = "Invalid environment for SyncTriggers operation.";
                _logger.LogWarning(result.Error);

                return result;
            }

            try
            {
                await _meshServiceClient.NotifyTriggersChanged();
            }
            catch (Exception ex)
            {
                string message = "Failed to notify triggers changed.";
                _logger.LogError(ex, message);
                result.Success = false;
                result.Error = message;
            }

            return result;
        }

protected override async Task<(bool Success, string ErrorMessage)> SetTriggersAsync(string content)
        {
            try
            {
                await _meshServiceClient.NotifyTriggersChanged();

                return (true, null);
            }
            catch (Exception ex)
            {
                string message = "Failed to notify triggers changed.";
                _logger.LogError(ex, message);

                return (false, message);
            }
        }
```

kamperiadis: Basically, for the background sync triggers flow, we just follow the existing logic in FunctionsSyncManager (i.e. generating triggers payload, comparing the hash with what is in blob storage, etc.) except that for SetTriggersAsync, we will be doing the notification to the platform. 

And for non-background sync triggers flows, we will just be doing the notification to the platform. 

What do you think?

jviau: >  it is wasteful to be generating the triggers payload when we are relying on the platform to fetch it later and this payload is just thrown away

I wouldn't index too much on this. The overall cost of this should be negligible in the grand scale of an instance's lifetime. Plus, it is on a background thread, so not impacting cold start.

If we ignore that optimization as a goal, does it affect your approach here?

test/WebJobs.Script.Tests.Integration/Management/SignalBasedFunctionsSyncManagerTests.cs
```diff
@@ -0,0 +1,148 @@
...
+using Newtonsoft.Json.Linq;
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Microsoft.Azure.WebJobs.Script.Tests.Managment
```
mathewc, line 22: Spelling error Management

src/WebJobs.Script/Environment/EnvironmentSettingNames.cs
```diff
@@ -40,6 +40,7 @@ public static class EnvironmentSettingNames
         public const string ContainerName = "CONTAINER_NAME";
         public const string WebsitePodName = "WEBSITE_POD_NAME";
         public const string LegionServiceHost = "LEGION_SERVICE_HOST";
+        public const string FunctionsTriggersSyncSignalEnabled = "FUNCTIONS_TRIGGERS_SYNC_SIGNAL_ENABLED";
```
mathewc, line 43: What do you think of a setting named FUNCTIONS_SYNC_TRIGGERS_MODE. The default value would be "FrontEnd" when not specified, and for ATOL you'd use "Platform". We'd map this to a TriggerSyncMode enum and use that for parsing.

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
```diff
@@ -144,7 +144,15 @@ public static void AddWebJobsScriptHost(this IServiceCollection services, IConfi
             });
 
             // Management services
-            services.AddSingleton<IFunctionsSyncManager, FunctionsSyncManager>();
+            if (SystemEnvironment.Instance.IsTriggersSyncSignalEnabled())
```
mathewc, line 147: If you like the mode suggestion I make below, this'd become `Instance.GetTrigerSyncMode() == TriggerSyncMode.AppServer`

src/WebJobs.Script.WebHost/Management/MeshServiceClient.cs
```diff
@@ -143,6 +143,18 @@ private async Task PublishActivities(IEnumerable<ContainerFunctionExecutionActiv
             response.EnsureSuccessStatusCode();
         }
 
+        public async Task NotifyTriggersChanged()
+        {
+            _logger.LogDebug("Posting triggers changed notification to appserver.");
+
+            var response = await SendAsync(
+            [
+                KeyValuePair.Create(Operation, "update-triggers-timestamp")
```
mathewc, line 152: Wondering if you shouldn't generalize this operation name to something like "update-triggers". That way if for whatever reason you want to pass any additional info along on the operation it makes sense.

test/WebJobs.Script.Tests.Integration/Management/MeshServiceClientTests.cs
```diff
@@ -261,6 +261,30 @@ private static bool ContainsActivities(HttpRequestMessage request, params Contai
             return expectedActivities.Aggregate(matchesOperation, (current, activity) => current && formData["content"].Contains(activity.FunctionName));
         }
 
+        private static bool IsNotifyTriggersChangedRequest(HttpRequestMessage request)
+        {
+            var formData = request.Content.ReadAsFormDataAsync().Result;
```
mathewc, line 266: I realize the rest of the methods in this test are using the sync over async pattern, but we should clean this up and make it truly async (ask @fabiocav :))

src/WebJobs.Script.WebHost/Management/MeshServiceClient.cs
```diff
@@ -143,6 +143,18 @@ private async Task PublishActivities(IEnumerable<ContainerFunctionExecutionActiv
             response.EnsureSuccessStatusCode();
         }
 
+        public async Task NotifyTriggersChanged()
+        {
+            _logger.LogDebug("Posting triggers changed notification to appserver.");
```
mathewc, line 148: I don't think we should log "appserver" - its not really a concept that exists currently in this codebase. Instead we could just be opaque and say "Performing platform triggers changed notification."

patelchandni:
Will be handled as part of https://github.com/Azure/azure-functions-host/pull/11813

PR #11694, opened 2026-04-06
[in-proc] Cap bundles for net8 in-proc

jviau:

resolves #issue_for_this_pr

### Pull request checklist

* [ ] I have added all required tests (Unit tests, E2E tests) -- TODO

### Additional information

This PR does two primary changes:

1. Adds a cap to net8 in-proc bundles of 4.32.0. This is in addition to the existing net6 cap.
2. ~Addresses a bug with hosting config bundle caps which will _always_ override any user configured max value. We only want to set a _lower_ max, not a higher max.~ -- reverted this change. We can separately discuss if this is desired behavior or not and address in another PR if so.

#### Additional Comments

I noticed an existing issue where if the user-defined version range and the allowed range has 0 overlap, then we will produce an invalid version range of `(HIGHER_VALUE, LOWER_VALUE)`, which will never download a bundle. This logic still will never download a bundle, but it gets the '(0.0.0, 0.0.0)` version range instead. Unsure if we want to address that issue. We could log a warning or error when it occurs.

### DIFF
--- a/src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
+++ b/src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
@@ -20,6 +20,13 @@ namespace Microsoft.Azure.WebJobs.Script.ExtensionBundle
 {
     public class ExtensionBundleManager : IExtensionBundleManager
     {
+        private static readonly IReadOnlyDictionary<int, VersionRange> _allowedBundleVersionRanges =
+            new Dictionary<int, VersionRange>()
+            {
+                [6] = new VersionRange(new NuGetVersion("4.2.0"), true, new NuGetVersion("4.22.0"), true),
+                [8] = new VersionRange(new NuGetVersion("4.2.0"), true, new NuGetVersion("4.32.0"), true),
+            };
+
         private readonly IEnvironment _environment;
         private readonly ExtensionBundleOptions _options;
         private readonly FunctionsHostingConfigOptions _configOption;
@@ -253,16 +260,7 @@ namespace Microsoft.Azure.WebJobs.Script.ExtensionBundle
         internal string FindBestVersionMatch(VersionRange versionRange, IEnumerable<string> versions, string bundleId, FunctionsHostingConfigOptions configOption)
         {
             int dotnetVersion = typeof(string).Assembly.GetName().Version.Major;
-
-            // Limit the version range for v4.x bundles on .NET 6 to be between 4.2.0 and 4.22.0
-            // Return original version range if not .NET 6 or not v4.x bundle
-            var effectiveVersionRange = GetAdjustedVersion(dotnetVersion, versionRange, bundleId);
-
-            // Check if there is a max version configured in hosting config for the current dotnet version and bundle major version
-            if (TryGetMaxBundleVersionFromHostingConfig(bundleId, versionRange.MinVersion.Major, dotnetVersion, configOption, out NuGetVersion maxBundleVersion))
-            {
-                effectiveVersionRange = new VersionRange(effectiveVersionRange.MinVersion, effectiveVersionRange.IsMinInclusive, maxBundleVersion, true);
-            }
+            var effectiveVersionRange = GetAdjustedVersion(dotnetVersion, versionRange, bundleId, configOption);
 
             var bundleVersions = versions.Select(p =>
             {
@@ -280,29 +278,30 @@ namespace Microsoft.Azure.WebJobs.Script.ExtensionBundle
             return matchingVersion?.ToString();
         }
 
-        // Applies bundle version limits for .NET 6, using the default bundle if referencing a v4 major bundle version.
+        // Applies bundle version limits for apps based on .NET version and hosting config, using the default bundle if referencing a v4 major bundle version.
         // Does not apply version limit in placeholder mode. This will prevent bundle download in placeholder mode as long as a matching bundle is available.
-        private VersionRange GetAdjustedVersion(int dotnetVersion, VersionRange versionRange, string bundleId)
+        private VersionRange GetAdjustedVersion(int dotnetVersion, VersionRange versionRange, string bundleId, FunctionsHostingConfigOptions configOption)
         {
             if (_environment.IsPlaceholderModeEnabled()
-                || dotnetVersion != 6
                 || !string.Equals(bundleId, ScriptConstants.DefaultExtensionBundleId, StringComparison.OrdinalIgnoreCase)
                 || versionRange.MinVersion.Major != ScriptConstants.ExtensionBundleV4MajorVersion)
             {
                 return versionRange;
             }
 
-            var effectiveMinVersion = new NuGetVersion(ScriptConstants.Net6MinimumV4BundleVersion);
-            var effectiveMaxVersion = new NuGetVersion(ScriptConstants.Net6MaximumV4BundleVersion);
+            VersionRange configRange = null;
+            if (TryGetMaxBundleVersionFromHostingConfig(bundleId, versionRange.MinVersion.Major, dotnetVersion, configOption, out NuGetVersion maxBundleVersion))
+            {
+                configRange = new VersionRange(versionRange.MinVersion, versionRange.IsMinInclusive, maxBundleVersion, true);
+            }
 
-            var minVersion = versionRange.MinVersion < effectiveMinVersion ? effectiveMinVersion : versionRange.MinVersion;
-            var maxVersion = versionRange.MaxVersion > effectiveMaxVersion ? effectiveMaxVersion : versionRange.MaxVersion;
+            if (_allowedBundleVersionRanges.TryGetValue(dotnetVersion, out VersionRange allowedRange))
+            {
+                versionRange = configRange is null
+                    ? VersionRange.CommonSubSet([versionRange, allowedRange])
+                    : VersionRange.CommonSubSet([versionRange, allowedRange, configRange]);
+            }
 
-            // Use the original inclusion values if it is within the min or max range.
-            bool isMinInclusive = versionRange.MinVersion > effectiveMinVersion && versionRange.IsMinInclusive;
-            bool isMaxInclusive = versionRange.MaxVersion < effectiveMaxVersion && versionRange.IsMaxInclusive;
-
-            versionRange = new VersionRange(minVersion, isMinInclusive, maxVersion, isMaxInclusive);
             return versionRange;
         }
 
--- a/src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
+++ b/src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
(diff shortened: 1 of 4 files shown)


src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -20,6 +20,13 @@ namespace Microsoft.Azure.WebJobs.Script.ExtensionBundle
 {
     public class ExtensionBundleManager : IExtensionBundleManager
     {
+        private static readonly IReadOnlyDictionary<int, VersionRange> _allowedBundleVersionRanges =
+            new Dictionary<int, VersionRange>()
```
manvkaur: Can we also cap Preview bundle to 4.44.0 ?

Preview hasn't bumped Identity packages yet.

soninaren: @manvkaur We may have to start putting host version compatibility information in the bundle. Specifically bundle.json.

soninaren: It would be easier to pursue that change with preview bundles since they are download only.

manvkaur: Let's sync on the expected spec/design for it, I can do the bundle side change for it.

jviau: It could be possible depending on how 'preview' is selected by the user. Is there a way for this code to easily differentiate if the feed they are using is preview or release?

manvkaur: Yes, this class has code which determines the bundle Id as well.

soninaren: We have official bundles ids, Prod, Preview, experimental. We also have logic app bundles - Prod bundle. But if possible i would want to avoid host having knowledge about bundle and its release trains.

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -280,29 +278,34 @@ internal string FindBestVersionMatch(VersionRange versionRange, IEnumerable<stri
...
+            VersionRange configRange = null;
+            if (TryGetMaxBundleVersionFromHostingConfig(bundleId, versionRange.MinVersion.Major, dotnetVersion, configOption, out NuGetVersion maxBundleVersion))
+            {
+                configRange = new VersionRange(versionRange.MinVersion, versionRange.IsMinInclusive, maxBundleVersion, true);
+            }
 
-            // Use the original inclusion values if it is within the min or max range.
-            bool isMinInclusive = versionRange.MinVersion > effectiveMinVersion && versionRange.IsMinInclusive;
-            bool isMaxInclusive = versionRange.MaxVersion < effectiveMaxVersion && versionRange.IsMaxInclusive;
+            if (_allowedBundleVersionRanges.TryGetValue(dotnetVersion, out VersionRange allowedRange))
```
soninaren: This change makes is so we that we cannot override the bundle version if it is out of allowed range. We might need to do that via hosting config. Should we maintain the config hierarchy here?

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
fabiocav: Agreed for the preview, we should be using the mechanism we discussed for the feature dependency and compatibility check.

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
jviau: Right okay. So we want hosting config to be able to override the max value of the built-in value?

soninaren: yeah. in case we need to patch bundles.

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -280,29 +278,34 @@ internal string FindBestVersionMatch(VersionRange versionRange, IEnumerable<stri
...
+            {
+                versionRange = configRange is null
+                    ? VersionRange.CommonSubSet([versionRange, allowedRange])
+                    : VersionRange.CommonSubSet([versionRange, allowedRange, configRange]);
+            }
+            else
+            {
+                versionRange = VersionRange.CommonSubSet([versionRange, configRange]);
```
fabiocav: `configRange` could be `null` here. In practice, this will not be an issue with the current target, but a potential bug if they change.

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -20,6 +20,13 @@ namespace Microsoft.Azure.WebJobs.Script.ExtensionBundle
 {
     public class ExtensionBundleManager : IExtensionBundleManager
     {
+        private static readonly IReadOnlyDictionary<int, VersionRange> _allowedBundleVersionRanges =
+            new Dictionary<int, VersionRange>()
+            {
+                [6] = new VersionRange(new NuGetVersion("4.2.0"), false, new NuGetVersion("4.22.0"), false),
+                [8] = new VersionRange(null, false, new NuGetVersion("4.33.0"), false),
```
fabiocav: Can we add some tests to validate the cap?

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
manvkaur: cap should be anything strictly less than 4.33.0, in case we need to do bundle patch, we can do 4.32.x

jviau: Made the change.

@manvkaur, that is what I have - it is an exclusive upper bound of 4.33.0

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
jviau: Refactored this area.

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
jviau: Yes, I will be adding those. Also have broken tests to address.

jviau: Added new test cases and updated some existing tests.

test/WebJobs.Script.Tests/ExtensionBundle/ExtensionBundleManagerTests.cs
```diff
@@ -410,13 +412,13 @@ public void LimitMaxVersion(string versionRange, string hostConfigVersion, strin
...
         [Theory]
 #if NET8_0
-        // No host-config max (full list available, no caps applied on .NET 8)
-        [InlineData(ScriptConstants.LatestPlatformChannelNameUpper, "[4.*, 5.0.0)", null, "4.23.0")]
-        [InlineData("latest", "[4.*, 5.0.0)", null, "4.23.0")]
-        [InlineData(ScriptConstants.StandardPlatformChannelNameUpper, "[4.*, 5.0.0)", null, "4.22.1")]
-        [InlineData("standard", "[4.*, 5.0.0)", null, "4.22.1")]
-        [InlineData(ScriptConstants.ExtendedPlatformChannelNameUpper, "[4.*, 5.0.0)", null, "4.22.1")]
-        [InlineData("extended", "[4.*, 5.0.0)", null, "4.22.1")]
+        // No host-config, built in cap of 4.32.0.
```
manvkaur:
```suggestion
        // No host-config, built in cap of 4.32.*.
```

soninaren (approved):

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -280,43 +280,38 @@ internal string FindBestVersionMatch(VersionRange versionRange, IEnumerable<stri
...
+            BundleRangeId id = (dotnetVersion, versionRange.MinVersion.Major);
+            if (_allowedBundleVersionRanges.TryGetValue(id, out VersionRange allowedRange))
+            {
+                versionRange = VersionRange.CommonSubSet([allowedRange, versionRange]);
+            }
 
-            // Use the original inclusion values if it is within the min or max range.
-            bool isMinInclusive = versionRange.MinVersion > effectiveMinVersion && versionRange.IsMinInclusive;
-            bool isMaxInclusive = versionRange.MaxVersion < effectiveMaxVersion && versionRange.IsMaxInclusive;
+            if (TryGetMaxBundleVersionFromHostingConfig(id, configOption, out NuGetVersion maxBundleVersion))
```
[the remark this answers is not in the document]
jviau: Intersection was considered and the PR description calls that out. I am intentionally **not** doing this to align with behavior today. If we want to do intersection, we can discuss and make that change separately.

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
```diff
@@ -16,10 +16,19 @@
...
+
 namespace Microsoft.Azure.WebJobs.Script.ExtensionBundle
 {
     public class ExtensionBundleManager : IExtensionBundleManager
     {
+        private static readonly IReadOnlyDictionary<BundleRangeId, VersionRange> _allowedBundleVersionRanges =
+            new Dictionary<BundleRangeId, VersionRange>()
+            {
+                [(6, 4)] = new VersionRange(new NuGetVersion("4.2.0"), false, new NuGetVersion("4.22.0"), false),
+                [(8, 4)] = new VersionRange(null, false, new NuGetVersion("4.33.0"), false),
```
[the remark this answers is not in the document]
jviau: There is no meaningful difference here. They are still statics one way or another.

test/WebJobs.Script.Tests/ExtensionBundle/ExtensionBundleManagerTests.cs
```diff
@@ -383,10 +383,13 @@ public async Task GetExtensionBundle_CannotReachZipEndpoint_ReturnsFalseAsync()
         [Theory]
         [InlineData("[3.*, 4.0.0)", "3.19.0", "3.19.0")]
         [InlineData("[4.*, 5.0.0)", "4.3.0", "4.3.0")]
+        [InlineData("[4.*, 4.3.0)", "4.30.0", "4.23.0")] // Existing behavior. We will override the customer's max range.
 #if NET6_0
         [InlineData("[4.*, 5.0.0)", null, "4.3.0")]
+        [InlineData("[4.*, 4.1.0)", null, null)] // No overlap. Bundle will fail to resolve.
```
jviau, line 389: This test case was added to prove the no overlap scenario results in `null` bundle in current and new changes.

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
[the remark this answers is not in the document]
liliankasem: It would be worth adding a code comment here to make it clear to maintainers i.e. `Hosting config intentionally overrides the built-in cap to allow patching.`

liliankasem (approved):

src/WebJobs.Script/ExtensionBundle/ExtensionBundleManager.cs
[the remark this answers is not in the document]
jviau: I'll leave that comment to a result of our discussion about if we change this or not.

PR #11697, opened 2026-04-07
Restrict GET admin/host/triggers to platform claim

kamperiadis:

Restrict GET admin/host/triggers API so that it can only be called internally by the platform

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.WebHost/Controllers/HostController.cs
+++ b/src/WebJobs.Script.WebHost/Controllers/HostController.cs
@@ -21,6 +21,7 @@ using Microsoft.Azure.WebJobs.Script.Scale;
 using Microsoft.Azure.WebJobs.Script.WebHost.Filters;
 using Microsoft.Azure.WebJobs.Script.WebHost.Management;
 using Microsoft.Azure.WebJobs.Script.WebHost.Models;
+using Microsoft.Azure.WebJobs.Script.WebHost.Security.Authentication;
 using Microsoft.Azure.WebJobs.Script.WebHost.Security.Authorization.Policies;
 using Microsoft.Azure.WebJobs.Script.Workers;
 using Microsoft.Azure.WebJobs.Script.Workers.Rpc;
@@ -414,20 +415,15 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Controllers
         {
             _metricsLogger.LogEvent(MetricEventNames.GetTriggersInvoked);
 
-            // This endpoint is only available in validation mode
-            if (_environment.IsInValidationMode())
+            if (!User.IsFuncPlatform())
             {
-                var result = await _functionsSyncManager.GetTriggersAsync();
-
-                // GetTriggersAsync() does not swallow exceptions, so any failures will still be a 500.
-                // The only result.Success == false we need to consider is when the environment
-                // does not support sync triggers.
-                return result.Success
-                    ? Ok(result)
-                    : StatusCode(StatusCodes.Status403Forbidden, new { status = result.Error });
+                return Forbid();
             }
 
-            return StatusCode(StatusCodes.Status403Forbidden, new { status = "GetTriggers operation is not supported in for this instance." });
+            var result = await _functionsSyncManager.GetTriggersAsync();
+            return result.Success
+                ? Ok(result)
+                : StatusCode(StatusCodes.Status500InternalServerError, new { status = result.Error });
         }
 
         [HttpPost]
--- a/src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
+++ b/src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
@@ -74,7 +74,7 @@ namespace Microsoft.Extensions.DependencyInjection
 
                     if (string.Equals(c.SecurityToken.Issuer, ScriptConstants.LegionCoreUri, StringComparison.OrdinalIgnoreCase))
                     {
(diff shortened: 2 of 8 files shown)


src/WebJobs.Script.WebHost/Security/Authentication/Jwt/ScriptJwtBearerExtensions.cs
```diff
@@ -75,6 +75,7 @@ private static void ConfigureOptions(JwtBearerOptions options)
                     if (string.Equals(c.SecurityToken.Issuer, ScriptConstants.LegionCoreUri, StringComparison.OrdinalIgnoreCase))
                     {
                         claims.Add(new Claim(SecurityConstants.AssignUnencryptedClaimType, "true"));
+                        claims.Add(new Claim(SecurityConstants.LegionInfraClaimType, "true"));
```
jviau: The pattern already started here is to have claims for the kind of action to be performed, not who you are. We either should

1. Continue with that per-action claim pattern, so something like `GetTriggersClaimType`.
2. Or, if we want to avoid needing a claim-per-action, we can think of a generic claim that describes the _role_ and not the _identity_. Think something like "admin" claim, but for these scenarios. Could we introduce something like `func-platform` (`FuncPlatformClaimType`)? That is less specific about it being legion infra, and a more broad claim. Would then consolidate the `AssignUnencryptedClaimType` claim type here as well.

kamperiadis: Thank you @jviau. I went with approach 2.

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -414,20 +415,15 @@ public async Task<IActionResult> GetTriggers()
...
-                    ? Ok(result)
-                    : StatusCode(StatusCodes.Status403Forbidden, new { status = result.Error });
+                return Forbid();
             }
 
-            return StatusCode(StatusCodes.Status403Forbidden, new { status = "GetTriggers operation is not supported in for this instance." });
+            var result = await _functionsSyncManager.GetTriggersAsync();
+            return result.Success
+                ? Ok(result)
+                : StatusCode(StatusCodes.Status500InternalServerError, new { status = result.Error });
```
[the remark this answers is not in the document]
kamperiadis: We don't really expect `result.Success == false` since Legion will only call this for a running pod so IsSyncTriggersEnvironment should always evaluate to true. So if we have `result.Success == false`, then we should be returning a 500 since something unexpectedly went wrong. I would not use 403 or 400 since the failure is not related to permissions or a bad request.

src/WebJobs.Script.WebHost/Security/Authentication/SecurityConstants.cs
```diff
@@ -21,9 +21,6 @@ public class SecurityConstants
         /// </summary>
         public const string InvokeClaimType = "http://schemas.microsoft.com/2017/07/functions/claims/invoke";
```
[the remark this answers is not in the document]
kamperiadis: Done

src/WebJobs.Script.WebHost/Controllers/HostController.cs
```diff
@@ -414,20 +415,15 @@ public async Task<IActionResult> GetTriggers()
         {
             _metricsLogger.LogEvent(MetricEventNames.GetTriggersInvoked);
 
-            // This endpoint is only available in validation mode
-            if (_environment.IsInValidationMode())
+            if (!User.HasClaim(SecurityConstants.FuncPlatformClaimType, "true"))
```
jviau: Can we add an extension method so we can call `User.IsFuncPlatform()`?

kamperiadis: Good suggestion. Done.

jviau (approved):

eliaslopezgt (approved):

src/WebJobs.Script.WebHost/Security/Authentication/SecurityConstants.cs
```diff
@@ -22,8 +22,9 @@ public class SecurityConstants
...
 
         /// <summary>
-        /// Claim indicating whether a principal is authorized to invoke assign with an unencrypted payload.
+        /// Claim indicating if the principal has the role to perform platform-only operations
+        /// such as unencrypted assignment and triggers retrieval.
         /// </summary>
-        public const string AssignUnencryptedClaimType = "http://schemas.microsoft.com/2017/07/functions/claims/assign-unencrypted";
+        public const string FuncPlatformClaimType = "http://schemas.microsoft.com/2017/07/functions/claims/func-platform";
```
mathewc, line 28: Isn't this a breaking change at this point?

mathewc (changes requested):

src/WebJobs.Script.WebHost/Security/Authentication/SecurityConstants.cs
jviau: No, no one outside of the host ever knew about this claim. This claim was both added by the host as part of AuthN and then consumed by the host in AuthZ.

mathewc (approved):

jviau (approved):

patelchandni:
Hey @mathewc @jviau - I just rebased this branch, so PR needs re-approval. Thanks!

mathewc (approved):

mathewc (approved):

mathewc (approved):

PR #11714, opened 2026-04-14
[Compute Separation] Specialization flow: StandbyManager skips host restart for external workers, /link triggers ScriptHost startup

brettsam:
- **`StandbyManager` skips host lifecycle for external workers** — During `/assign`, env vars and config are applied but `SpecializeAsync`, `RestartHostAsync`, and `DelayUntilHostReady` are skipped. The `ScriptHost` is started later by the `/link` call.
- **`RpcInitializationService` not registered in external worker mode** — No inbound gRPC server or local worker channel pre-warming when workers are external.
- **First-worker failure recovery** — If the first worker's `StartAsync` fails, subsequent workers can retry instead of being permanently blocked by a faulted `TaskCompletionSource`.
- **`DisposeAsync` ordering fix** — `ReaderWriterLockSlim` is now disposed after worker cleanup to prevent `ObjectDisposedException` during shutdown races.
- **Aspire harness updated for placeholder mode** — Adds `WEBSITE_PLACEHOLDER_MODE`, `WEBSITE_SKU`, `CONTAINER_ENCRYPTION_KEY`, and `MESH_INIT_URI` env vars for local specialization testing.
- **`.http` file supports `/assign`** — Pre-encrypted specialization payload using a well-known dev encryption key for manual testing.

### DIFF
--- a/src/WebJobs.Script.Grpc/ExternalWorkers/WorkerConnectionService.cs
+++ b/src/WebJobs.Script.Grpc/ExternalWorkers/WorkerConnectionService.cs
@@ -37,8 +37,8 @@ internal sealed class WorkerConnectionService : IHostedService, IWorkerConnectio
     private readonly ILogger _logger;
 
     private readonly ConcurrentDictionary<string, WorkerConnection> _workers = new();
-    private readonly TaskCompletionSource _scriptHostStarted = new(TaskCreationOptions.RunContinuationsAsynchronously);
     private readonly ReaderWriterLockSlim _lifecycleLock = new();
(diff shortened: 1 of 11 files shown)


src/WebJobs.Script.Grpc/ExternalWorkers/WorkerConnectionService.cs
fabiocav: This here could potentially hang indefinitely. We should pass the cancellation token into this call:

``` csharp
await _scriptHostStarted.Task.WaitAsync(cancellationToken);
```

PR #11721, opened 2026-04-16
Shutdown all invocations when scripthost restarts

### LINKED ISSUE #11502: Worker shutdown does not always fail current executions
In #10936 we added the capability of terminating in-flight executions when the language worker is restarted due to a timeout. We should extend that capability to cover other scenarios where the language worker is shutdown, such restarting the `ScriptHost` due to a resume operation.

satvu:

resolves #11502 - ScriptHost restart does not properly "fail" all in-flight invocations. It simply shuts down and discards of the invocations, resulting in partners like durable thinking these invocations are still actively executing when they are not. 

### Pull request checklist

### Additional Information 

- Whenever there is a call to dispose a worker channel, we should call `Shutdown(exception?)` first to make sure all invocations are failed/handled so they are properly compelted.


src/WebJobs.Script.Grpc/Rpc/JobHostRpcWorkerChannelManager.cs
```diff
@@ -72,6 +72,7 @@ public void ShutdownChannels()
                         if (removedChannels.TryRemove(removedChannel.Id, out IRpcWorkerChannel _))
                         {
                             _logger.LogDebug("Disposing language worker channel with id:{workerId}", removedChannel.Id);
+                            removedChannel.Shutdown(null);
```
satvu: All of the new calls to `Shutdown` pass in `null` (no exception) since none is provided at this level. 

This will result in `FunctionTimeoutAbortException` being set as the reason for failure [here](https://github.com/Azure/azure-functions-host/blob/6bf029c09523a645c7c9c661eb27803f63fcb45c/src/WebJobs.Script.Grpc/Channel/WorkerChannel.cs#L1549). This isn't entirely accurate since the case it was previously used for was scripthost/jobhost stays up but worker channel is disposed. If extensions are using this exception for only that scenario (not a hard failure, resend request again), do we need to come up with a new exception? Or is this behavior ok as is?

src/WebJobs.Script.Grpc/Rpc/JobHostRpcWorkerChannelManager.cs
```diff
@@ -49,7 +49,15 @@ public Task<bool> ShutdownChannelIfExistsAsync(string channelId, Exception worke
...
                         _logger.LogDebug("Disposing language worker channel with id:{workerId}", id);
-                        rpcChannel.Shutdown(workerException);
+
+                        try
+                        {
+                            rpcChannel.Shutdown(workerException);
+                        }
+                        catch (Exception ex)
+                        {
+                            _logger.LogError(ex, "Error failing executions on shutdown for language worker channel with id:{workerId}", id);
```
kshyju, line 59: Previously an exception in the Shutdown call would have been bubbled up to the caller. Now we are catching the exception and preventing the bubble up and executing the code below. Is that control flow change behavior intentional?

src/WebJobs.Script.Grpc/Rpc/JobHostRpcWorkerChannelManager.cs
```diff
@@ -72,6 +80,16 @@ public void ShutdownChannels()
                         if (removedChannels.TryRemove(removedChannel.Id, out IRpcWorkerChannel _))
                         {
                             _logger.LogDebug("Disposing language worker channel with id:{workerId}", removedChannel.Id);
+
+                            try
+                            {
+                                removedChannel.Shutdown(null);
```
kshyju, line 86: So, this `ShutDown `call is the actual fix? Might be a good idea to update PR description with a bit more details of the issue and how this fixes it to help reviewers.

satvu: Yes, updated the description with a one liner of the GH issue.

src/WebJobs.Script.Grpc/Rpc/JobHostRpcWorkerChannelManager.cs
satvu: Made this change based on a GH copilot review comment - we still want the channel to be disposed (line 62) so throwing here may prevent that.

Any errors in the `Shutdown` call is most likely a failure to "fail"/complete the execution which puts us in the same position as we are without this fix. I don't think any failure here is worth preventing channel disposal.

PR #11723, opened 2026-04-17
[Compute Separation] Align WorkerProxy management APIs; wire tests to CI

brettsam:
### Route changes
- `GET /ready` → `GET /admin/worker/ready`
- `POST /drain` → `POST /admin/worker/drain`
- `POST /instanceState` → `POST /admin/infra/instanceState`

### Readiness fix
- Changed readiness check from `>= ReadyForRequest` to `== ReadyForRequest` — the old comparison incorrectly reported draining/deleted workers as ready

### `instanceState` response redesign
- Replaced transition-based model (`fromPodStatus`/`toPodStatus`, `changeFlags`) with the flat `FunctionsWorkerPod` schema: `{ functionsContainerType, podName, revision, state: { podStatus, runtimePodName, functionGroupName, isAlwaysReady, replacementPolicy } }`
- Response field renamed from `revisionId` to `revision`; poll request field also `revision`

### Drain API
- Request body now requires `{ "reason": "..." }` with `DrainReason` enum: `IdleScaleIn`, `RuntimeStopping`, `ReplaceWorkerKeepRuntime`, `OrphanCleanup`
- First-reason-wins semantics — subsequent drain calls are idempotent
- Reason maps to sticky `ReplacementPolicy` (`NoReplacement` or `SameRuntimeRefill`)
- `ReplaceWorkerKeepRuntime` yields `SameRuntimeRefill` only when `functionGroupName == "http"`
- Runtime-initiated gRPC drain now also routes through `AcceptDrain(RuntimeStopping)` for consistent `replacementPolicy` population
- gRPC drain message only sent on first accepted drain (not on retries)

### Assign payload
- Added optional fields: `FunctionAppName`, `FunctionAppId`, `FunctionGroupName`, `IsAlwaysReady`
- `FunctionGroupName` and `IsAlwaysReady` are stored and published in `instanceState`

### Enum/model cleanup
- Removed `DrainCompleted` (never observable — was immediately followed by `MarkForDeletion`)
- Renamed `MarkForDeletion` → `MarkedForDeletion`
- Removed `WorkerPodHealthStatus` (dead state — only ever set to `Healthy`, never `Unhealthy`)

### Configuration
- Moved `podName` resolution from `Environment.GetEnvironmentVariable` inside `WorkerPodStateManager` to `RelayOptions` (constructor-injected, testable)
- Added `--pod-name` CLI arg with fallback: `COMPUTERNAME` → `POD_NAME` → hostname

### CI
- Wired compute separation E2E tests into the integration test pipeline under group `ExternalWorkerEndToEndTests`

### DIFF
--- a/src/Functions.WorkerProxy/Program.cs
+++ b/src/Functions.WorkerProxy/Program.cs
@@ -73,12 +73,7 @@ app.MapGrpcService<FunctionRpcRelay>();
 var stateManager = app.Services.GetRequiredService<WorkerPodStateManager>();
 var relay = app.Services.GetRequiredService<FunctionRpcRelay>();
 
-app.MapGet("/admin/worker/ready", () =>
-{
-    return stateManager.CurrentStatus == WorkerPodStatus.ReadyForRequest
-        ? Results.Ok()
-        : Results.StatusCode(503);
-});
+app.MapGet("/admin/worker/ready", () => ManagementApiHandlers.HandleReady(stateManager));
 
 // Worker specialization. NNA calls this after /admin/worker/ready succeeds with the
 // app settings payload. The worker proxy drives the full init + specialization +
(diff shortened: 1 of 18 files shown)


src/Functions.WorkerProxy/Program.cs
```diff
@@ -114,30 +122,48 @@
...
+    }
+
+    if (drainRequest is null)
+    {
+        return Results.BadRequest("Request body with 'reason' is required.");
+    }
+
+    if (stateManager.AcceptDrain(drainRequest.Reason))
+    {
+        await relay.SendDrainRequestToRuntimeAsync();
```
fabiocav, line 144: `AcceptDrain` persists `_drainReason` and transitions the pod to `Draining` before `SendDrainRequestToRuntimeAsync()` is awaited in the handler. If the gRPC write fails
(channel not yet connected, cancellation, transient I/O), the proxy now reports `Draining` but the runtime was never told. This may ultimately result in the worker being permanently stuck.

src/Functions.WorkerProxy/Program.cs
```diff
@@ -114,30 +122,48 @@
...
 
-app.MapPost("/drain", async () =>
+app.MapPost("/admin/worker/drain", async (HttpContext ctx, CancellationToken cancellationToken) =>
 {
-    stateManager.UpdatePodStatus(WorkerPodStatus.Draining);
-    await relay.SendDrainRequestToRuntimeAsync();
+    WorkerDrainRequest? drainRequest;
+    try
+    {
+        drainRequest = await ctx.Request.ReadFromJsonAsync(WorkerProxyJsonContext.Default.WorkerDrainRequest, cancellationToken);
```
fabiocav, line 130: Can you run a test on this where the body is `{}` or has an invalid property name for the reason? Because the reason is non-nullable, you'll end up with the default `IleScaleIn`. We should make that nullable and perform the appropriate checks.

PR #11725, opened 2026-04-17
[Compute Separation] Adding runtime state tracking and publishing

mathewc:
Adding runtime worker/slot tracking and publishing as per agreed upon platform contracts.

### DIFF
--- a/src/WebJobs.Script.WebHost/ContainerManagement/RuntimeStatePublisher.cs
+++ b/src/WebJobs.Script.WebHost/ContainerManagement/RuntimeStatePublisher.cs
@@ -1,0 +1,147 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Script.WebHost.Management;
+using Microsoft.Azure.WebJobs.Script.Workers;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.ContainerManagement
+{
+    /// <summary>
+    /// Hosted service that forwards <see cref="IRuntimeStateManager"/> state
+    /// changes to the mesh service via <see cref="IMeshServiceClient.PublishRuntimeState"/>.
+    /// </summary>
+    /// <remarks>
+    /// Change notifications are debounced over a short window so bursts (e.g.
+    /// several <c>OnWorkerAdded</c> calls during initial link, or many
+    /// acquire/release operations in quick succession) coalesce into a single
+    /// mesh publish carrying the latest snapshot. Only registered when
+    /// compute separation is enabled.
+    /// </remarks>
+    internal sealed class RuntimeStatePublisher : IHostedService, IDisposable
+    {
+        private const int PublishDebounceMs = 500;
+
+        private readonly IRuntimeStateManager _stateManager;
+        private readonly IMeshServiceClient _meshServiceClient;
+        private readonly ILogger<RuntimeStatePublisher> _logger;
+        private readonly Timer _timer;
+
+        private int _publishPending;
+        private int _publishInFlight;
+        private bool _disposed;
+
+        public RuntimeStatePublisher(
+            IRuntimeStateManager stateManager,
+            IMeshServiceClient meshServiceClient,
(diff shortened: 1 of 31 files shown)


src/WebJobs.Script.WebHost/ContainerManagement/RuntimeStatePublisher.cs
```diff
@@ -0,0 +1,110 @@
...
+                {
+                    // Raced with Dispose — drop the publish.
+                    Interlocked.Exchange(ref _publishPending, 0);
+                }
+            }
+        }
+
+        private async void OnPublishTimer(object state)
+        {
+            Interlocked.Exchange(ref _publishPending, 0);
```
fabiocav: Gate this call to avoid concurrent executions?

src/WebJobs.Script.WebHost/Controllers/InstanceController.cs
```diff
@@ -122,64 +120,32 @@ public IActionResult GetHttpHealthStatus()
...
         /// <example>
         /// <code>
-        /// POST /admin/instance/stop
+        /// GET /admin/instance/state
         /// </code>
         /// </example>
-        [HttpPost]
-        [Route("admin/instance/stop")]
+        [HttpGet]
+        [Route("admin/instance/state")]
```
mathewc, line 137: admin/instance/stop is moving to admin/host/stop

This new admin/instance/state API we can decide on - I added it here to give the platform a way to query for the worker/slot info. We might decide we don't need this.

src/WebJobs.Script.WebHost/Controllers/WorkerController.cs
```diff
@@ -33,14 +33,14 @@ public WorkerController(
     }
 
     /// <summary>
-    /// Links a worker pod to this runtime. Initiates an outbound gRPC connection
+    /// Assigns an external worker to this runtime. Initiates an outbound gRPC connection
     /// to the worker proxy and returns immediately. The connection and handshake
```
fabiocav, line 37: We have discussed waiting on the connection handshake to complete (ensure we can at least connect to the worker proxy) before continuing and returning a 200. I believe the doc was updated to reflect that.

src/WebJobs.Script.WebHost/Controllers/RequestSlotsController.cs
```diff
@@ -0,0 +1,106 @@
...
+        /// <summary>
+        /// Reserves request slots against the runtime's shared pool.
+        /// </summary>
+        /// <remarks>
+        /// Grants may be partial: the response's <see cref="RequestSlotsLeaseResponse.AcquiredSlotCount"/>
+        /// indicates the number of slots actually reserved, which can be anywhere
+        /// from zero to <see cref="RequestSlotsLeaseRequest.Count"/>.
+        /// The caller is responsible for releasing whatever was granted.
+        /// </remarks>
+        [HttpPut]
```
fabiocav: Discussed offline: POST

src/WebJobs.Script.WebHost/Controllers/WorkerController.cs
```diff
@@ -33,14 +33,14 @@ public WorkerController(
...
     /// complete in the background.
     /// </summary>
-    [HttpPost]
-    [Route("admin/workers/link")]
+    [HttpPut]
```
fabiocav, line 40: Discussed offline. This should PUT to `admin/workers/{id}`

src/WebJobs.Script/Workers/RuntimeState.cs
```diff
@@ -0,0 +1,46 @@
...
+
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.WebJobs.Script.Workers;
+
+/// <summary>
+/// Immutable snapshot of runtime-wide state relevant to the App Server:
+/// the cap on linked workers and request-slot accounting.
+/// Published to the mesh service as <c>publish-runtime-state</c> and
+/// returned by the <c>GET /admin/workers/status</c> admin API.
```
fabiocav: This endpoint is not correct anymore, right?

src/WebJobs.Script.WebHost/Controllers/WorkerController.cs
mathewc: Let's handle this in a follow up, since it'll require a bunch of test changes. Also @brettsam may be looking at this already - If not, I can handle in a follow up.

brettsam (approved):

PR #11731, opened 2026-04-20
[Compute Separation] improving docker images

brettsam:
### Runtime image (`WebJobs.Script.WebHost/Dockerfile`)
- **2.48 GB → ~317 MB**
- Added `ExcludeWorkers` MSBuild property to skip restoring bundled language workers (Python 960 MB, PowerShell 348 MB, dotnet-isolated 98 MB, Java 60 MB) — they're unused in compute separation mode where all execution happens on remote worker pods
- Also uses `PublishWorkers=false` to exclude workers from publish output
- Switched base from `aspnet:10.0` (full Ubuntu) to `aspnet:10.0-noble-chiseled-extra` (distroless), saving ~120 MB on the base layer
- Empty `workers/` directory preserved for host startup probe

### WorkerProxy image (`Functions.WorkerProxy/Dockerfile`)
- **93.9 MB → ~46 MB**
- Copy only the native AOT binary into the final image, excluding the 36 MB `.dbg` debug symbols file that was inadvertently included via `COPY /app/publish .`

### Build infrastructure (`Workers.props`)
- New `ExcludeWorkers` property gates the `Import` of per-language worker `.props` files, skipping NuGet restore of all worker packages when set to `true`
- Existing `PublishWorkers` property continues to control publish-time inclusion

### DIFF
--- a/eng/build/Workers.props
+++ b/eng/build/Workers.props
@@ -1,7 +1,7 @@
 <Project>
 
   <!-- Individual workers are in their own props file. -->
-  <Import Project="$(MSBuildThisFileDirectory)Workers.*.props" />
+  <Import Project="$(MSBuildThisFileDirectory)Workers.*.props" Condition="'$(ExcludeWorkers)' != 'true'" />
 
   <!-- Remove all worker items from the ReadyToRun publish list -->
   <Target Name="ExcludeWorkers" AfterTargets="ComputeFilesToPublish" BeforeTargets="ResolveReadyToRunCompilers" Condition="'$(PublishReadyToRun)' == 'true'">
--- a/eng/build/Workers.props
+++ b/eng/build/Workers.props
@@ -1,7 +1,7 @@
 <Project>
 
-  <!-- Individual workers are in their own props file. -->
-  <Import Project="$(MSBuildThisFileDirectory)Workers.*.props" Condition="'$(ExcludeWorkers)' != 'true'" />
+  <!-- Individual workers are in their own props file. Skip restore entirely when SkipWorkerPackages is set. -->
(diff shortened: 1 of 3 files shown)


mathewc (approved):

eng/build/Workers.props
```diff
@@ -1,7 +1,7 @@
 <Project>
 
   <!-- Individual workers are in their own props file. -->
-  <Import Project="$(MSBuildThisFileDirectory)Workers.*.props" />
```
fabiocav, line 4: This property has the same name as the target below, which might lead to confusion.

eng/build/Workers.props
fabiocav: Should we also evaluate the new property here for clarity/consistency?

src/WebJobs.Script.WebHost/Dockerfile
```diff
@@ -4,17 +4,16 @@ FROM mcr.microsoft.com/dotnet/sdk:10.0 AS build
 WORKDIR /repo
 COPY . .
 RUN dotnet publish src/WebJobs.Script.WebHost/WebJobs.Script.WebHost.csproj \
-    -c Release -o /app/publish
+    -c Release -o /app/publish \
+    -p:ExcludeWorkers=true -p:PublishWorkers=false \
+    && mkdir -p /app/publish/workers
 
-FROM mcr.microsoft.com/dotnet/aspnet:10.0
+FROM mcr.microsoft.com/dotnet/aspnet:10.0-noble-chiseled-extra
```
fabiocav: One thing to validate is whether the host will be OK in all the flows where it runs different processes with this image, as it would fail if it launches the shell at any point.

brettsam: I've reverted it but not entirely sure when the host would ever launch a process now? Plus this WebHost image is really only for local dev work so prod wouldn't be affected by this but will be safe anyway.

PR #11736, opened 2026-04-22
[Compute Separation] Changes to WorkerProxy: inbound and outbound http ports

sidkri:
Changes the WorkerProxy component
- Changed the default management http port to 80 to align with platform and existing functions worker
- Language worker picks a port dynamically on startup to receive http function executions and communicates this over GRPC but WorkerProxy was using a configure port.  Updated WorkerProxy to use the port specified over GRPC.

### Pull request checklist

### DIFF
--- a/src/Functions.WorkerProxy/Dockerfile
+++ b/src/Functions.WorkerProxy/Dockerfile
@@ -10,5 +10,5 @@ RUN dotnet publish src/Functions.WorkerProxy/Functions.WorkerProxy.csproj \
 FROM mcr.microsoft.com/dotnet/runtime-deps:10.0-noble-chiseled
 WORKDIR /app
 COPY --from=build /app/publish/Microsoft.Azure.Functions.WorkerProxy .
-EXPOSE 50051 50052 50053 50054
+EXPOSE 80 50051 50052 50053
 ENTRYPOINT ["./Microsoft.Azure.Functions.WorkerProxy"]
--- a/src/Functions.WorkerProxy/FunctionRpcRelay.cs
+++ b/src/Functions.WorkerProxy/FunctionRpcRelay.cs
@@ -75,7 +75,7 @@ internal sealed class FunctionRpcRelay : FunctionRpc.FunctionRpcBase
     // HTTP endpoint advertised by the worker in its WorkerInitResponse "HttpUri" capability.
     // Populated once per worker lifetime (during /assign); read by the HTTP forwarding
     // middleware to route invocations to the worker's dynamically-chosen port.
-    private volatile string? _workerHttpEndpoint;
+    internal volatile string? _workerHttpEndpoint;
 
     /// <summary>
     /// HTTP endpoint advertised by the worker via the <c>HttpUri</c> capability in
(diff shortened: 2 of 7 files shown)


src/Functions.WorkerProxy/Dockerfile
```diff
@@ -10,5 +10,5 @@ RUN dotnet publish src/Functions.WorkerProxy/Functions.WorkerProxy.csproj \
 FROM mcr.microsoft.com/dotnet/runtime-deps:10.0-noble-chiseled
 WORKDIR /app
 COPY --from=build /app/publish/Microsoft.Azure.Functions.WorkerProxy .
-EXPOSE 50051 50052 50053 50054
+EXPOSE 80 50051 50052 50053
```
brettsam, line 13: Is this required? My agent tells me this doesn't do much and you can still bind to port 80 and then pass 80 as -`-management-port` and it should work as expected.

src/Functions.WorkerProxy/FunctionRpcRelay.cs
```diff
@@ -353,9 +364,23 @@ private void RewriteHttpUri(StreamingMessage message)
...
-        // Always set HttpUri to the proxy's endpoint. The worker may or may not have
-        // reported its own HttpUri — in container mode it often doesn't because
-        // its HTTP listener address isn't known at init time.
+        // Capture the worker's advertised HttpUri (dynamically-chosen port reported by
+        // the isolated worker SDK's HttpUriProvider) before we overwrite it. This is what
+        // the HTTP forwarding middleware will use as the YARP destination for invocations.
+        if (message.WorkerInitResponse.Capabilities.TryGetValue("HttpUri", out var advertisedUri)
+            && !string.IsNullOrWhiteSpace(advertisedUri))
+        {
+            _workerHttpEndpoint = advertisedUri;
```
brettsam, line 380: Yep, definitely a bug. I just realized all our e2e tests used a hardcoded port which is why they worked. I can follow up with test hardening after this across e2e if you don't want to tackle it here. But this does unblock you quickly.

brettsam (approved):

src/Functions.WorkerProxy/Program.cs
```diff
@@ -53,9 +57,22 @@
     if (ctx.Connection.LocalPort == httpProxyPort)
     {
         var logger = ctx.RequestServices.GetRequiredService<ILogger<FunctionRpcRelay>>();
+        var relayInstance = ctx.RequestServices.GetRequiredService<FunctionRpcRelay>();
+
+        // Prefer the HttpUri the worker advertised via gRPC WorkerInitResponse
+        // (dynamic port). Fall back to the CLI/env override when provided (dev harness).
```
brettsam: We need to flip this so that we fwd to the specified port in the args over whatever the app replied with for local dev reasons:
- docker desktop cannot expose dynamic ports between containers so we need to just fwd to some known endpoint
- i had to add a "hack" to the [sample app we run for testing](https://github.com/Azure/azure-functions-host/blob/feature/compute_separation/tools/ComputeSeparation/SampleIsolatedApp/Program.cs#L9-L20) so that it will listen on this specific port
- we'll never do any of this in production. `--worker-http-endpoint` is for local dev only.
- ideally we'll figure out a way around this for local dev as well and not need it.

But in summary -- I tested this and we need the arg to take preference over what the worker's grpc reported.

brettsam (changes requested):
Just realized that this does break local dev via containers.

src/Functions.WorkerProxy/Dockerfile
brettsam: Leaving it since it doesn't matter much...

src/Functions.WorkerProxy/Program.cs
brettsam: Fixed in latest iteration

PR #11757, opened 2026-04-29
Ensure wwwroot directory exists on new slot and app creation w/ networking restrictions

### LINKED ISSUE #11654: New slot creation OR network restrictions can throw error where no wwwroot is found
### Issue 

When the default directory structure is not created for an app, the customer gets errors like: _System.ComponentModel.Win32Exception : An error occurred trying to start process 'node' with working directory 'C:\home\site\wwwroot'. The directory name is invalid._

For the main site, this is typically done by the control plane _unless_ it is blocked by networking restrictions. For slots, since control plane does not create the necessary directory structure on creation, any new slot will hit this error.

In the past, retries would result in the directory creation, but changes [here](https://github.com/Azure/azure-functions-host/pull/10937/changes#diff-a419f494ef14a26a0e437c7c6a8a3385b111e5517a6198036c7c839e106ca6a6) to fix a separate bug have exposed this one.

### Slots repro

- Create a premium Windows app (used Node).
- Create a new slot.
- New slot immediately displays error.

This also reportedly occurred on Linux and for python apps.

### Networking Restrictions on Main Site Repro

- See 21000000953877:
  - EP app pointing to storage account w/ networking restrictions (deny public access)

### Mitigation

Once an app is deployed the slot/site, the error is resolved as the act of deployment results in the necessary directory structure.

satvu:

resolves #11654

### Pull request checklist


src/WebJobs.Script.Grpc/ProcessManagement/DefaultWorkerProcessFactory.cs
```diff
@@ -40,6 +41,24 @@ public virtual Process CreateWorkerProcess(WorkerContext context)
             {
                 throw new ArgumentNullException(nameof(context.Arguments.ExecutablePath));
             }
+
+            // Ensure the working directory exists before starting the process.
+            // This handles cases where the directory structure hasn't been created yet,
+            // such as new deployment slots or apps with network restrictions blocking
+            // the control plane from initializing the default directory structure.
+            if (!string.IsNullOrEmpty(context.WorkingDirectory) && !Directory.Exists(context.WorkingDirectory))
```
satvu: Will `context.WorkingDirectory` be reliable? Worried about a potential scenario like this value being different for placeholder and a real app deployment, leading to issues where `wwwroot` isn't created where it's expected.

Either way I think this would mitigate it for some scenarios since the value `context.WorkingDirectory` is what's used in line 69 below to start the worker process.

src/WebJobs.Script.Grpc/ProcessManagement/DefaultWorkerProcessFactory.cs
```diff
@@ -40,6 +41,24 @@ public virtual Process CreateWorkerProcess(WorkerContext context)
...
+            {
+                try
+                {
+                    _logger.LogInformation("Working directory '{workingDirectory}' does not exist. Creating it.", context.WorkingDirectory);
+                    Directory.CreateDirectory(context.WorkingDirectory);
+                }
+                catch (Exception ex)
+                {
+                    _logger.LogWarning(ex, "Failed to create working directory '{workingDirectory}'. The worker process may fail to start.", context.WorkingDirectory);
```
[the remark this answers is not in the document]
satvu: I think the current failure behavior (diagnostic error saying no directory is found) is sufficient and this log adds more information upon investigation.

src/WebJobs.Script.Grpc/ProcessManagement/DefaultWorkerProcessFactory.cs
jviau: Ah yeah during specialization we might not necessarily be starting a new process but specializing an existing process. You can track down how we do this by looking for when we send the grpc "FunctionsEnvironmentReloadRequest" (or something like that) to the worker.

But I _think_ any scenario with slots doesn't have placeholders, so it might be an irrelevant concern.

satvu: Reviewed the code - when a process is being specialized, this is not called (as this creates a new process) so this won't be a concern for slots (or other scenarios). This bug also affects non-slots scenarios such as when there are networking restrictions preventing azure files from creating the directory.

jviau (approved):

jviau:
Make sure to update release notes.

jviau (approved):

soninaren (approved):

PR #11760, opened 2026-05-01
Implementing support for WebHook ARM support opt-in

mathewc:
Changes to allow extensions that support webhooks to be more intentional about whether their APIs are intended to be used by ARM clients.

Supporting SDK PR: https://github.com/Azure/azure-webjobs-sdk/pull/3201. Once we have the packages for that, this PR will be updated to bump the SDK nuget version.


src/WebJobs.Script.WebHost/Filters/ArmExtensionResourceFilter.cs
```diff
@@ -30,24 +38,97 @@ public async Task OnResourceExecutionAsync(ResourceExecutingContext context, Res
...
+
+                    string rawConfig = context.HttpContext.RequestServices
+                        .GetService<IOptionsMonitor<FunctionsHostingConfigOptions>>()?.CurrentValue?.ArmWebhookOptInEnabled;
+
+                    if (IsArmWebhookOptInEnabled(rawConfig, out IReadOnlySet<string> exclusions))
+                    {
+                        string extensionName = context.RouteData.Values.TryGetValue("extensionName", out object value)
+                            ? value as string
+                            : null;
+                        IScriptWebHookProvider provider = context.HttpContext.RequestServices.GetService<IScriptWebHookProvider>();
```
kshyju: Can we move this inside the `IsArmAllowedForExtension` method? (closer to it's usage). Looks like it is needed only in the last line.

src/WebJobs.Script/Config/FunctionsHostingConfigOptions.cs
```diff
@@ -206,6 +206,31 @@ internal bool IsDotNetInProcDisabled
             }
         }
 
+        /// <summary>
+        /// Gets or sets a '|'-delimited list of extension names to exclude
+        /// from enforcement. A single <c>"|"</c> enables enforcement with no exclusions.
+        /// Enforcement is enabled if and only if this property returns a non-empty value.
+        /// </summary>
+        internal string ArmWebhookOptInEnabled
```
kshyju: The `Enabled` suffix sounds like this should be a property which returns a boolean value. If this is returning a string of excluded extensions (from the strict check). Is there a more readable name we could use?

src/WebJobs.Script.WebHost/Filters/ArmExtensionResourceFilter.cs
```diff
@@ -30,24 +38,97 @@ public async Task OnResourceExecutionAsync(ResourceExecutingContext context, Res
...
+            if (string.IsNullOrEmpty(extensionName))
+            {
+                return false;
+            }
+
+            if (exclusions is not null && exclusions.Contains(extensionName))
+            {
+                return true;
+            }
+
```
kshyju, line 140: Shouldn't you be returning false if  `exclusions` is `null`? (Based on the comment " A single <c>"|"</c> enables enforcement with no exclusions") in `ArmWebhookOptInEnabled` property?

src/WebJobs.Script.WebHost/Filters/ArmExtensionResourceFilter.cs
```diff
@@ -30,24 +38,97 @@ public async Task OnResourceExecutionAsync(ResourceExecutingContext context, Res
...
+                    {
+                        string extensionName = context.RouteData.Values.TryGetValue("extensionName", out object value)
+                            ? value as string
+                            : null;
+                        IScriptWebHookProvider provider = context.HttpContext.RequestServices.GetService<IScriptWebHookProvider>();
+
+                        // Determine if target extension allows ARM requests.
+                        if (!IsArmAllowedForExtension(extensionName, exclusions, provider))
+                        {
+                            context.HttpContext.Response.StatusCode = StatusCodes.Status403Forbidden;
```
kshyju: Does this mean that we can only enable this feature flag after ALL supported extensions finishes consider/review adopting the use of this new attribute? Otherwise, it will break existing extensions. rite?

src/WebJobs.Script.WebHost/Filters/ArmExtensionResourceFilter.cs
mathewc: I think it makes sense to leave this as is. I'd rather keep the IsArmAllowedForExtension method a simple decision function, rather than also having it do service location.

src/WebJobs.Script.WebHost/Filters/ArmExtensionResourceFilter.cs
mathewc: Correct.

src/WebJobs.Script.WebHost/Filters/ArmExtensionResourceFilter.cs
mathewc: If we have an exclusion list, we short circuit the extension check below if we have a match. If we DONT have an exclusion list (null), then ARM is allowed for the extension IFF the extension has opted in.

src/WebJobs.Script/Config/FunctionsHostingConfigOptions.cs
mathewc: Yes, the naming of this has been difficult, because I've chosen to use a single config for both enablement as well as an exclusion list. Having two distinct settings would make the naming simpler, but config rollout harder.

Renaming it ArmWebhookOptInEnforcement and keeping the same semantics.

src/WebJobs.Script.WebHost/WebHooks/IScriptWebHookProvider.cs
```diff
@@ -13,5 +13,15 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
...
+
+        /// <summary>
+        /// Returns a value indicating whether the named extension has opted-in to receiving
+        /// requests forwarded via the ARM hostruntime extensions bridge. Extensions opt-in by
+        /// applying <c>SupportsArmRequestsAttribute</c> to their <see cref="IExtensionConfigProvider"/>
+        /// implementation.
+        /// </summary>
+        /// <param name="name">The extension name as registered with the provider.</param>
+        /// <returns><c>true</c> if the extension has opted-in; otherwise <c>false</c>.</returns>
+        bool IsArmAllowed(string name);
```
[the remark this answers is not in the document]
mathewc: These interfaces are only used internally, even though public. This assembly isn't published as a nuget package for external users.

NuGet.config
```diff
@@ -1,12 +1,12 @@
...
+<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
     <clear />
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
     <add key="Microsoft.Azure.Functions.PowerShellWorker" value="https://azfunc.pkgs.visualstudio.com/e6a70c92-4128-439f-8012-382fe78d6396/_packaging/Microsoft.Azure.Functions.PowerShellWorker/nuget/v3/index.json" />
     <add key="AzureFunctions@internalrelease" value="https://azfunc.pkgs.visualstudio.com/e6a70c92-4128-439f-8012-382fe78d6396/_packaging/AzureFunctions%40internalrelease/nuget/v3/index.json" />
     <add key="AzureFunctionsRelease" value="https://azfunc.pkgs.visualstudio.com/e6a70c92-4128-439f-8012-382fe78d6396/_packaging/AzureFunctionsRelease/nuget/v3/index.json" />
-    <add key="AzureFunctionsPreRelease" value="https://azfunc.pkgs.visualstudio.com/e6a70c92-4128-439f-8012-382fe78d6396/_packaging/AzureFunctionsPreRelease/nuget/v3/index.json" />
+    <add key="pre-release" value="https://pkgs.dev.azure.com/azfunc/public/_packaging/pre-release/nuget/v3/index.json" />
```
satvu, line 9: Is this change to allow referencing the unreleased sdk? Wondering b/c this is a replacement instead of something additive.

mathewc: @jviau  told me that this is this is the new pre-release feed. I don't think we were using the other one anymore. This feed is where our release pipelines publish to. I think we're good here since CI builds with this.

jviau: Yeah, I think if CI works, we are good. We want to remove all these "AzureFunctions" feeds from "e6a70c92-4128-439f-8012-382fe78d6396" anyways.

src/WebJobs.Script.WebHost/Filters/ArmExtensionResourceFilter.cs
```diff
@@ -1,53 +1,144 @@
...
+            var request = context.HttpContext.Request;
+            var controllerActionDescriptor = context.ActionDescriptor as ControllerActionDescriptor;
+            if (IsExtensionWebHookAction(controllerActionDescriptor))
+            {
+                // Stamp the HttpContext so downstream consumers (e.g. the extension webhook
+                // handler) can easily detect that the request was forwarded via the ARM
+                // extensions bridge and apply ARM-specific behavior.
+                context.HttpContext.Items[ScriptConstants.AzureFunctionsArmWebhookRequestKey] = true;
+
+                string rawConfig = context.HttpContext.RequestServices
```
jviau, line 62: Can this be DI'd instead of pulled out of the services?

fabiocav: +1

src/WebJobs.Script.WebHost/Filters/ArmExtensionResourceFilter.cs
```diff
@@ -1,53 +1,144 @@
...
+                string.Equals(method.Name, nameof(HostController.ExtensionWebHookHandler), StringComparison.Ordinal);
+        }
+
+        // Determines whether ARM webhook opt-in enforcement is enabled, based on the raw
+        // ArmWebhookOptInEnforcement hosting config value. Returns false when the value is null
+        // or empty. When enabled, returns true and populates <paramref name="exclusions"/>
+        // with the parsed '|'-delimited exemption set, or null when there are no exemptions.
+        internal static bool IsArmWebhookOptInEnforced(string rawConfig, out IReadOnlySet<string> exclusions)
+        {
+            if (string.IsNullOrEmpty(rawConfig))
```
fabiocav, line 115: Should we switch this to `IsNullOrWhiteSpace` instead?

fabiocav (approved):
A couple of nit comments, but flagging as approved.

src/WebJobs.Script.WebHost/Filters/ArmExtensionResourceFilter.cs
mathewc: I did it this way because this config check is temporary. Eventually it becomes the default and config is removed, so I want to avoid having to make another set of changes to the ctor and tests for something temporary.

src/WebJobs.Script.WebHost/Filters/ArmExtensionResourceFilter.cs
mathewc: I think I'll leave as is. I get your point, but I think I want it to be that if any value is set, enforcement is enabled.

kshyju (approved):

PR #11793, opened 2026-05-27
fix(config): load connstrings directly into config

jviau:
.NET10 has a breaking change where they started to check for more connection string formats and loading them into 'ConnectionStrings:' section. This PR loads them directly into the configuration on-top of the .NET10 ConnectionStrings section behavior.

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script/Config/ScriptEnvironmentVariablesConfigurationSource.cs
+++ b/src/WebJobs.Script/Config/ScriptEnvironmentVariablesConfigurationSource.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using System.Collections;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Configuration.EnvironmentVariables;
 
@@ -11,14 +12,74 @@ namespace Microsoft.Azure.WebJobs.Script
     /// This source exists to replace the default <see cref="EnvironmentVariablesConfigurationSource"/> to allow
     /// us to override caching behavior.
     /// </summary>
-    public class ScriptEnvironmentVariablesConfigurationSource : IConfigurationSource
+    /// <param name="liveEnvironmentLoading">
+    /// If true, the provider will use the live environment variables when fetching values.
+    /// If false, it will cache environment variables one time on load.
+    /// </param>
+    /// <remarks>
+    /// Live environment fetching is the existing behavior we use in the WebHost so that post-specialization
+    /// changes to the environment are reflected in the configuration.
+    /// </remarks>
+    public class ScriptEnvironmentVariablesConfigurationSource(bool liveEnvironmentLoading = true) : IConfigurationSource
     {
         public IConfigurationProvider Build(IConfigurationBuilder builder)
         {
-            return new WebHostEnvironmentVariablesProvider();
+            return liveEnvironmentLoading
+                ? new WebHostEnvironmentVariablesProvider() : new CompatEnvironmentVariablesProvider();
         }
 
-        private class WebHostEnvironmentVariablesProvider : EnvironmentVariablesConfigurationProvider
+        private class CompatEnvironmentVariablesProvider : EnvironmentVariablesConfigurationProvider
+        {
+            // Values taken from https://github.com/dotnet/runtime/blob/46e21499a8f230988fac59b694af407a86e0f46c/src/libraries/Microsoft.Extensions.Configuration.EnvironmentVariables/src/EnvironmentVariablesConfigurationProvider.cs#L18
+            // These should be kept in sync with any values added AFTER .NET8
+            private const string PostgreSqlServerPrefix = "POSTGRESQLCONNSTR_";
+            private const string ApiHubPrefix = "APIHUBCONNSTR_";
+            private const string DocDbPrefix = "DOCDBCONNSTR_";
+            private const string EventHubPrefix = "EVENTHUBCONNSTR_";
+            private const string NotificationHubPrefix = "NOTIFICATIONHUBCONNSTR_";
+            private const string RedisCachePrefix = "REDISCACHECONNSTR_";
+            private const string ServiceBusPrefix = "SERVICEBUSCONNSTR_";
+
+            public CompatEnvironmentVariablesProvider() : base()
+            {
+            }
+
+            public override void Load()
+            {
+                base.Load();
+
+                static string Normalize(string key)
+                {
+                    return key.Replace("__", ConfigurationPath.KeyDelimiter);
+                }
+
+                static bool IsPrefixedString(string key)
+                {
+                    return key.StartsWith(PostgreSqlServerPrefix, StringComparison.OrdinalIgnoreCase) ||
+                           key.StartsWith(ApiHubPrefix, StringComparison.OrdinalIgnoreCase) ||
+                           key.StartsWith(DocDbPrefix, StringComparison.OrdinalIgnoreCase) ||
+                           key.StartsWith(EventHubPrefix, StringComparison.OrdinalIgnoreCase) ||
+                           key.StartsWith(NotificationHubPrefix, StringComparison.OrdinalIgnoreCase) ||
+                           key.StartsWith(RedisCachePrefix, StringComparison.OrdinalIgnoreCase) ||
+                           key.StartsWith(ServiceBusPrefix, StringComparison.OrdinalIgnoreCase);
+                }
+
+                // .NET 10 changed to special case some prefixes, inserting it into the "ConnectionStrings" section.
+                // For backwards compatibility, we still want to include these in the configuration.
+                foreach (DictionaryEntry kvp in Environment.GetEnvironmentVariables())
+                {
+                    if (kvp.Key is string key && kvp.Value is string value)
+                    {
+                        if (IsPrefixedString(key))
+                        {
+                            Data[Normalize(key)] = value;
+                        }
+                    }
+                }
+            }
+        }
+
+        private class WebHostEnvironmentVariablesProvider : CompatEnvironmentVariablesProvider
         {
             public WebHostEnvironmentVariablesProvider() : base()
             {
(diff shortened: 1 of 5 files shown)


src/WebJobs.Script/Config/ScriptEnvironmentVariablesConfigurationSource.cs
```diff
@@ -36,6 +45,48 @@ public override void Set(string key, string value)
...
+                            Data[NormalizeKey(key)] = value;
+                        }
+                        else if (key.StartsWith(DocDbPrefix, StringComparison.OrdinalIgnoreCase))
+                        {
+                            Data[NormalizeKey(key)] = value;
+                        }
+                        else if (key.StartsWith(EventHubPrefix, StringComparison.OrdinalIgnoreCase))
+                        {
+                            Data[NormalizeKey(key)] = value;
+                        }
```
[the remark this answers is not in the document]
jviau: No, this is the opposite of what we want. We need to go from `__` to `:` when mapping to `Data`.

mathewc:
.NET change referenced above: https://github.com/dotnet/runtime/pull/116037

fabiocav (approved):

src/WebJobs.Script/Config/ScriptEnvironmentVariablesConfigurationSource.cs
```diff
@@ -36,6 +45,53 @@ public override void Set(string key, string value)
...
+            public override void Load()
+            {
+                base.Load();
+
+                static string Normalize(string key)
+                {
+                    return key.Replace("__", ConfigurationPath.KeyDelimiter);
+                }
+
+                // .NET 10 changed to special case some prefixes, inserting it into the "ConnectionStrings" section.
```
mathewc, line 67: Unless the plan is for this code to be temporary pending a regression fix from .NET, we should probably put a link to the file https://github.com/dotnet/runtime/blob/46e21499a8f230988fac59b694af407a86e0f46c/src/libraries/Microsoft.Extensions.Configuration.EnvironmentVariables/src/EnvironmentVariablesConfigurationProvider.cs#L18 indicating that these need to be kept in sync?

mathewc (approved):

src/WebJobs.Script.WebHost/Program.cs
```diff
@@ -94,7 +94,7 @@ public static IHostBuilder CreateHostBuilder(string[] args = null)
                                 config.Sources.Remove(envVarsSource);
                             }
 
-                            config.Add(new ScriptEnvironmentVariablesConfigurationSource());
+                            config.Add(new ScriptEnvironmentVariablesConfigurationSource(liveEnvironmentLoading: true));
```
mathewc: Can you explain why of all the calls to this ctor in the codebase, only this one is specifying true for this param?

jviau: that is the existing behavior we have today. Having `true` will get you the `WebHostEnvironmentVariablesProvider`, which is what this line of code always had.

src/WebJobs.Script/Config/ScriptEnvironmentVariablesConfigurationSource.cs
```diff
@@ -11,14 +12,79 @@ namespace Microsoft.Azure.WebJobs.Script
     /// This source exists to replace the default <see cref="EnvironmentVariablesConfigurationSource"/> to allow
     /// us to override caching behavior.
     /// </summary>
-    public class ScriptEnvironmentVariablesConfigurationSource : IConfigurationSource
+    public class ScriptEnvironmentVariablesConfigurationSource(bool liveEnvironmentLoading = true) : IConfigurationSource
     {
         public IConfigurationProvider Build(IConfigurationBuilder builder)
         {
-            return new WebHostEnvironmentVariablesProvider();
+            return liveEnvironmentLoading
```
mathewc, line 27: A comment on liveEnvironmentLoading would be helpful here - why it exists, under what circumstances we need to vary this behavior, etc.

From my read of the code, you're just saying at the WebHost level, you ALWAYS want to go to the raw environment, not pulling from Data. Why is that?

test/Resources/TestProjects/WebJobsStartupTests/Function1.cs
```diff
@@ -119,7 +119,7 @@ private static bool ValidateConfig(IConfiguration _config)
                     root.Providers.ElementAt(i++) is ChainedConfigurationProvider &&
                     root.Providers.ElementAt(i++) is MemoryConfigurationProvider && // FUNCTIONS_WORKER_RUNTIME surfaced from environment.
                     root.Providers.ElementAt(i++) is JsonConfigurationProvider &&
-                    root.Providers.ElementAt(i++) is EnvironmentVariablesConfigurationProvider &&
+                    root.Providers.ElementAt(i++).GetType().Name.StartsWith("CompatEnvironmentVariablesConfigurationProvider") &&
```
mathewc: The class name is CompatEnvironmentVariablesProvider so this will always be false? Rather than a string compare, cant you do a Type level check?

test/WebJobs.Script.Tests/Config/ScriptEnvironmentVariablesConfigurationSourceTests.cs
```diff
@@ -0,0 +1,120 @@
...
+                // The base provider still surfaces arbitrary env vars, so the value should be
+                // resolvable through TryGet; the assertion confirms our override doesn't break that.
+                config[prefixed].Should().Be(_value);
+            }
+        }
+
+        private static IConfigurationRoot BuildConfiguration()
+        {
+            var builder = new ConfigurationBuilder();
+            builder.Add(new ScriptEnvironmentVariablesConfigurationSource(liveEnvironmentLoading: false));
```
mathewc: Are you testing the liveEnvironmentLoading=true path at all?

jviau: No, that code path is not important to this test scenario. We can expand tests later.

test/Resources/TestProjects/WebJobsStartupTests/Function1.cs
jviau: Might explain the test failures...

src/WebJobs.Script/Config/ScriptEnvironmentVariablesConfigurationSource.cs
```diff
@@ -11,14 +12,79 @@ namespace Microsoft.Azure.WebJobs.Script
...
+                    return key.Replace("__", ConfigurationPath.KeyDelimiter);
+                }
+
+                // .NET 10 changed to special case some prefixes, inserting it into the "ConnectionStrings" section.
+                // For backwards compatibility, we still want to include these in the configuration.
+                foreach (DictionaryEntry kvp in Environment.GetEnvironmentVariables())
+                {
+                    if (kvp.Key is string key && kvp.Value is string value)
+                    {
+                        if (key.StartsWith(PostgreSqlServerPrefix, StringComparison.OrdinalIgnoreCase))
```
mathewc: nit: since the logic is the same for all these prefixes, rather than duplicating it 7 times you could simplify:

```
if (IsPrefixedConnectionString(key))
{
  Data[Normalize(key)] = value;
}
```

src/WebJobs.Script/Config/ScriptEnvironmentVariablesConfigurationSource.cs
jviau: > Why is that?

That is existing behavior. I am intentionally leaving it in place. My guess it is for specialization - so new env vars from specialization are reflected in config.

jviau: Added comments

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -142,6 +144,16 @@ public static IHostBuilder AddScriptHost(this IHostBuilder builder,
             // Allow FunctionsStartup to add configuration after all other configuration is registered.
             builder.ConfigureAppConfiguration((context, configBuilder) =>
             {
+                // Replace the default environment variables source with a .NET8 back-compat version.
+                for (int i = 0; i < configBuilder.Sources.Count; i++)
+                {
+                    if (configBuilder.Sources[i] is EnvironmentVariablesConfigurationSource)
+                    {
+                        configBuilder.Sources[i] = new ScriptEnvironmentVariablesConfigurationSource(liveEnvironmentLoading: false);
+                        return;
```
mathewc: I think you want a break here, not return? Hopefully some test is failing because of this.

fabiocav: Indeed. As-is, this will skip the remaining configuration logic.

fabiocav (changes requested):

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
jviau: Fixed. Yeah, that was not supposed to be there at all. Not using a `break`, I think it is safer to replace all of them. There is only 1 in my testing, but better safe here. Was a tab-completion zealousness that added the `return;`

test/Resources/TestProjects/WebJobsStartupTests/Function1.cs
```diff
@@ -119,7 +119,7 @@ private static bool ValidateConfig(IConfiguration _config)
                     root.Providers.ElementAt(i++) is ChainedConfigurationProvider &&
                     root.Providers.ElementAt(i++) is MemoryConfigurationProvider && // FUNCTIONS_WORKER_RUNTIME surfaced from environment.
                     root.Providers.ElementAt(i++) is JsonConfigurationProvider &&
-                    root.Providers.ElementAt(i++) is EnvironmentVariablesConfigurationProvider &&
+                    root.Providers.ElementAt(i++).GetType().Name.StartsWith("CompatEnvironmentVariablesProvider") &&
```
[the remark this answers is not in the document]
jviau: Not really needed.

fabiocav (approved):

PR #11799, opened 2026-06-02
Only register HostnameFixupMiddleware on front-end-routed SKUs

fabiocav:
HostnameFixupMiddleware reads the `WAS_DEFAULT_HOSTNAME` request header to update the cached host name. The header is populated by the Antares front end, so the middleware is only meaningful on SKUs that route through it.

WebJobsApplicationBuilderExtension now skips the registration on hosting environments where front-end routing is not guaranteed. The decision is based on env vars that are set in the placeholder image and stable for the lifetime of the process. To support the registration check, the FrontEndRoutingEnsured extension is promoted from private to internal.


jviau (approved):

brettsam (approved):

fabiocav:
@copilot resolve the merge conflicts in this pull request

mathewc (approved):

PR #11800, opened 2026-06-03
Removing unnecessary service registrations for Flex.

mathewc:
We've identified that the events LinuxContainerActivityPublisher is publishing to the platform currently are only needed for Linux Consumption CV1 (ACI and Legion), not Flex. So removing this service registration as an optimization.

### DIFF
--- a/src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
+++ b/src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
@@ -296,9 +296,11 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
             });
         }
 
-        private static void AddLinuxContainerServices(this IServiceCollection services)
+        internal static void AddLinuxContainerServices(this IServiceCollection services, IEnvironment environment = null)
         {
-            if (SystemEnvironment.Instance.IsLinuxConsumptionOnLegion())
+            environment ??= SystemEnvironment.Instance;
+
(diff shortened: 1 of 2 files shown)


src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
```diff
@@ -373,7 +375,7 @@ private static void AddLinuxContainerServices(this IServiceCollection services)
             services.AddSingleton<LinuxContainerActivityPublisher>(s =>
             {
                 var environment = s.GetService<IEnvironment>();
-                if (environment.IsAnyLinuxConsumption())
+                if (environment.IsLinuxConsumptionOnAtlas())
```
yojagad: If this is going to be a hotfix, would it be safer to exclude FlexConsumption alone for this change and then follow up with excluding cv1 on legion as follow up ?

Tonewall: Slight change, we need to enable this for LinuxConsumption on Atlas and LinuxConsumption on Legion (only disable this for FlexConsumption)

brettsam (approved):

Tonewall (approved):
lgtm

yojagad (approved):

src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs
mathewc: Not going to be a hotfix.

PR #11805, opened 2026-06-03
Build coldstart artifacts once per OS and simplify the pipeline

kshyju:
## Summary
 The coldstart benchmark pipeline fans out to many run instances (20 per app, across 2 OSes). Previously **every** instance restored from the Azure Artifacts feed and ran its own `dotnet publish` of the host and the benchmark app, which caused **occasional HTTP 429 throttling** of the feed.
 
 This PR builds the host + benchmark apps **once per OS**, publishes them as pipeline artifacts, and has the run instances download those artifacts instead of restoring. This both fixes the feed throttling and **reduces overall pipeline run time**, since each run instance no longer repeats the build/restore - it just downloads prebuilt binaries. It also de-duplicates the pipeline YAML.
 
 ## Changes
 
 ### Build once, fan out (fixes feed throttling, cuts run time)
 - **New `eng/ci/templates/official/jobs/publish-coldstart-artifacts.yml`** - builds the host and each included benchmark app a single time per OS and publishes them as `coldstart_host_<OS>` / `coldstart_app_<OS>_<App>` artifacts.
   - .NET apps are built from a copy of `perf/Apps` **outside the repo** so repo-root MSBuild props/targets (central package management, signing, StyleCop, ...) don't apply, mirroring the existing per-instance JIT path (avoids `NU1008`).
   - For JIT trace generation, the .NET apps are built with the `AUTH_FUNCTION` constant, matching the SLA site.
   - Output is `tar.gz`'d so the Linux execute bit survives pipeline-artifact upload/download.
 - **New `eng/ci/templates/steps/acquire-coldstart-binaries.yml`** - downloads and extracts the prebuilt host + app for a run instance.
 - **`run-coldstart.yml`** - removed the per-instance host/app `dotnet publish` and `npm install`/copy steps; it now consumes the prebuilt artifacts via the new step template. This removes redundant build work from every run instance, shortening the `Run` stages. Added an `appType` parameter (`dotnet`/`node`) so the job branches on app type instead of hard-coded app names.
 
 ### Readability / de-duplication
 - `host.coldstart.yml` - collapsed the near-identical `Publish`, `Run`, and `Calculate` stages for Windows and Linux into `${{ each os in split('Windows,Linux', ',') }}` loops. PerfView parameters are injected only for Windows (its profiling steps run `perfview.exe`).
 - `coldstart.yml` variables - reduced `jitReleaseInstances` from `1,2,3` to `1` (one instance is sufficient for JIT trace generation and keeps generated YAML smaller).
 
 ## Behavior
 No intended change to the benchmark methodology or collected results - apps and host are built with the same configuration, just once per OS instead of per instance. Stage names (`PublishWindows`, `RunWindows`, `CalculateWindows`, and their Linux counterparts) are preserved, so cross-stage `dependsOn` wiring (including the `Merge*JitTraceFiles` stages) is unchanged.
 
 ## Testing
 - Validated by running the `host.coldstart` pipeline in ADO. 
 - Confirmed the publish stages produce the host/app artifacts and run instances consume them without per-instance feed restores.
 
 ## Risk / rollback
 - Pipeline-only change (`eng/ci/**`); no product code affected.
 - The publish job runs on `1es-pool-azfunc-benchmarking`, sharing the benchmarking pool with run jobs; watch pool capacity on large fan-outs.
 - Revert is a straight revert of this commit

### DIFF
--- a/eng/ci/templates/official/jobs/publish-coldstart-artifacts.yml
+++ b/eng/ci/templates/official/jobs/publish-coldstart-artifacts.yml
@@ -1,0 +1,129 @@
+# Builds the coldstart output (host + benchmark apps) ONCE per OS and publishes it as
+# pipeline artifacts, so the fan-out instances download instead of each running 'dotnet
+# publish' (which caused occasional throttling of the Azure Artifacts feed with HTTP 429).
+# Output is tar.gz'd because pipeline artifacts drop the Unix execute bit the host/worker
+# executables need on Linux (tar ships on the Windows runner image too, so the archive/extract
+# steps are identical on both OSes).
+
+parameters:
+- name: os
+  type: string
+  default: Windows
+  values:
+    - Windows
+    - Linux
+- name: apps
+  type: object
+  default: []
+- name: createJitTrace
+  type: boolean
+  default: false
+
+jobs:
+- job: PublishColdStartArtifacts_${{ parameters.os }}
+  displayName: Publish coldstart artifacts (${{ parameters.os }})
+
+  pool:
+    name: 1es-pool-azfunc-benchmarking
+    ${{ if eq(parameters.os, 'Linux') }}:
+      image: 1es-ubuntu-22.04-benchmark-runner-vanilla
+      os: linux
+    ${{ else }}:
+      image: 1es-windows-2022-benchmark-runner-vanilla
+      os: windows
+
+  variables:
+    ${{ if eq(parameters.os, 'Linux') }}:
+      publishRid: linux-x64
+    ${{ else }}:
+      publishRid: win-x64
+    hostOutputPath: $(Build.BinariesDirectory)/Published/HostRuntime
+    artifactsStagingPath: $(Build.ArtifactStagingDirectory)/coldstart
+    perfAppsStagingPath: $(Agent.TempDirectory)/perfapps
+    # JIT trace generation collects from an auth-enabled app (as the SLA site does), so the
+    # .NET apps must be built with the AUTH_FUNCTION constant in that case.
+    ${{ if eq(parameters.createJitTrace, true) }}:
+      benchmarkAppBuildConstants: -p:DefineConstants=AUTH_FUNCTION
+    ${{ else }}:
+      benchmarkAppBuildConstants: ''
+    # Restore pulls every worker's packages (python/node/java/...), a large NuGet cache; keep it
+    # off the OS disk by pointing NUGET_PACKAGES at the Agent.TempDirectory work/resource disk.
+    NUGET_PACKAGES: $(Agent.TempDirectory)/.nuget/packages
+
+  templateContext:
+    outputParentDirectory: $(artifactsStagingPath)
+    outputs:
+    - output: pipelineArtifact
+      displayName: Publish host artifact
+      path: $(artifactsStagingPath)/host
+      artifact: coldstart_host_${{ parameters.os }}
+    - ${{ each app in parameters.apps }}:
+      - ${{ if eq(app.include, true) }}:
+        - output: pipelineArtifact
+          displayName: Publish ${{ app.name }} app artifact
+          path: $(artifactsStagingPath)/app_${{ app.name }}
+          artifact: coldstart_app_${{ parameters.os }}_${{ app.name }}
+
+  steps:
+  - template: /eng/ci/templates/install-dotnet.yml@self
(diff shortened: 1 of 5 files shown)


eng/ci/templates/official/jobs/run-coldstart.yml
```diff
@@ -11,6 +11,9 @@ parameters:
   values:
     - Windows
     - Linux
+- name: appType
+  type: string
+  default: dotnet
```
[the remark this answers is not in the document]
kshyju: Valid feedack.  Fixed.

eng/ci/templates/official/jobs/publish-coldstart-artifacts.yml
```diff
@@ -0,0 +1,129 @@
+# Builds the coldstart output (host + benchmark apps) ONCE per OS and publishes it as
+# pipeline artifacts, so the fan-out instances download instead of each running 'dotnet
+# publish' (which caused occasional throttling of the Azure Artifacts feed with HTTP 429).
+# Output is tar.gz'd because pipeline artifacts drop the Unix execute bit the host/worker
+# executables need on Linux (tar ships on the Windows runner image too, so the archive/extract
+# steps are identical on both OSes).
+
+parameters:
+- name: os
```
jviau, line 9: Tiny nitpick: I personally would have this be `runtime` and take in values of `win-x64` and `linux-x64`. The pipeline is called "publish" and so to me it is more intuitive to describe _what_ you want published (the runtime) and not the _how_ you want it published (using a specific OS).

eng/ci/templates/official/jobs/publish-coldstart-artifacts.yml
```diff
@@ -0,0 +1,129 @@
...
+    perfAppsStagingPath: $(Agent.TempDirectory)/perfapps
+    # JIT trace generation collects from an auth-enabled app (as the SLA site does), so the
+    # .NET apps must be built with the AUTH_FUNCTION constant in that case.
+    ${{ if eq(parameters.createJitTrace, true) }}:
+      benchmarkAppBuildConstants: -p:DefineConstants=AUTH_FUNCTION
+    ${{ else }}:
+      benchmarkAppBuildConstants: ''
+    # Restore pulls every worker's packages (python/node/java/...), a large NuGet cache; keep it
+    # off the OS disk by pointing NUGET_PACKAGES at the Agent.TempDirectory work/resource disk.
+    NUGET_PACKAGES: $(Agent.TempDirectory)/.nuget/packages
```
jviau: With 1ES hosted images, the temp and work directory are the same drive by default. Did you customize the images in this pool at all to have different drives?

eng/ci/templates/official/jobs/publish-coldstart-artifacts.yml
```diff
@@ -0,0 +1,129 @@
...
+        - pwsh: |
+            New-Item -ItemType Directory -Force -Path "$(artifactsStagingPath)/app_${{ app.name }}" | Out-Null
+            tar -czf "$(artifactsStagingPath)/app_${{ app.name }}/app.tar.gz" -C "$(Build.BinariesDirectory)/Published/${{ app.name }}" .
+          displayName: Archive ${{ app.name }}
+      # Node app: 'npm install' in place, then archive the app folder (incl. node_modules).
+      - ${{ if eq(app.type, 'node') }}:
+        - task: UseNode@1
+          displayName: Install Node.js for ${{ app.name }}
+          inputs:
+            version: '18.x'
```
jviau: We should probably be using an in-service node version. Or will this impact the benchmarks?

jviau (approved):

brettsam (approved):

eng/ci/templates/official/jobs/publish-coldstart-artifacts.yml
kshyju: Good call. I switched to latest stable. All we care about exercising the host code which handles this worker (the .net worker has some special code path. so we want to use a non dotnet worker for jit trace gen)

eng/ci/templates/official/jobs/publish-coldstart-artifacts.yml
kshyju: Yea, this is not needed. Cleanedup

satvu (approved):

PR #11807, opened 2026-06-04
[INVESTIGATION - DO NOT MERGE] Capture deadlock dump from TestFunctionHost.Dispose

brettsam:
**This PR is for diagnostic investigation only - DO NOT MERGE.**

## Goal

Reproduce the sync-over-async deadlock that occurs when `TestFunctionHost.Dispose()` stops the underlying `IHost`, and capture a full process dump in CI so we can identify the actual blocking primitive.

Prior sessions narrowed the workaround to `CSharpPrecompiledTestFixture.DisposeAsync()` (PR #11795 base branch), which avoids the deadlock by stopping the WebHost from an async path before `TestFunctionHost.Dispose()` runs. That ships, but it leaves the underlying production bug unidentified - other fixtures hit the same shape (e.g. the `ProxyEndToEndTests.TestFixture` `FileMonitoringService` crash uncovered last build).

## Changes in this PR

1. **`TestFunctionHost.Dispose()`** - call `_webHost.StopAsync().GetAwaiter().GetResult(); _webHost.Dispose();` before the existing `HttpClient` / timer cleanup. This is the sync-over-async path that has historically deadlocked.
2. **`CSharpPrecompiledTestFixture.DisposeAsync()`** - removed the narrower workaround so we don't double-stop. The base `EndToEndTestFixture.DisposeAsync` calls `Host?.Dispose()` which now stops the WebHost itself.
3. **`eng/ci/templates/official/jobs/run-non-e2e-tests.yml`** -
   - Added `--blame-hang-dump-type full --blame-hang-timeout 5m` to the per-iteration `dotnet test` invocation. When the testhost goes silent for 5 minutes (which a sync-over-async deadlock will cause), VSTest captures a full dump.
   - Added `templateContext.outputs.pipelineArtifact` so `$(Agent.TempDirectory)\flake-repro` (containing both the per-iter trx/log AND the testhost dumps) is uploaded as a build artifact even on failure (`condition: succeededOrFailed()`).

## How to use

1. Wait for the CI run on this PR. With `parallel: 4` x `iterations: 5` = 20 testhost invocations per build, the deadlock is expected to land at least once.
2. From the failed build's **Artifacts** page, download the `flake-repro-*` zip(s). Each contains:
   - `iteration-N.log` / `iteration-N.trx` - the per-iter VSTest output
   - `<machine>_<random>\<testhost>.dmp` (only present when the hang fired)
   - `Sequence_*.xml` - which test was running when the hang was detected
3. Open the `.dmp` in Visual Studio (`File > Open > File...` then **Debug Managed Memory**), or analyze with `dotnet-dump analyze` / WinDbg + SOS.
4. Inspect the thread stuck in `IHost.StopAsync` -> hosted-service `StopAsync` -> the actual blocker. Prior hypothesis: `Microsoft.Azure.WebJobs.Script.Grpc.JobObjectRegistry.Close()` awaiting `ProcessWaitingForTermination.Task.Wait()` unbounded.

## Companion PR

Investigation parent: #11795 (`brettsam/repro-externalstartup-flake`) - that PR contains the narrower fix that will actually ship. This PR is the diagnostic harness only.

### DIFF
--- a/eng/ci/templates/official/jobs/run-integration-tests.yml
+++ b/eng/ci/templates/official/jobs/run-integration-tests.yml
@@ -154,14 +154,82 @@ jobs:
       arguments: '--filter "Group=RawAssemblyEndToEndTests" $(test_args)'
       projects: $(test_projects)
 
-  - task: DotNetCoreCLI@2
+  - task: PowerShell@2
     displayName: Samples end to end tests
     condition: succeededOrFailed()
     inputs:
-      command: test
+      targetType: inline
+      pwsh: true
+      script: |
+        $dumpDir = Join-Path $env:BUILD_ARTIFACTSTAGINGDIRECTORY "SamplesHangDumps"
+        New-Item -ItemType Directory -Path $dumpDir -Force | Out-Null
+
+        # Install dotnet-dump globally for hang diagnostics
+        dotnet tool install --global dotnet-dump 2>$null
+        $env:PATH = [System.Environment]::GetEnvironmentVariable("PATH","User") + ";" + $env:PATH
+
+        # Start the test run in the background
+        $testProc = Start-Process -FilePath "dotnet" `
+          -ArgumentList "test","test/WebJobs.Script.Tests.Integration/WebJobs.Script.Tests.Integration.csproj",`
+            "--filter","Group=SamplesEndToEndTests","-c","release","--no-build",`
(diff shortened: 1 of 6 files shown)


brettsam:
This work was folded into several other PRs. Closing.

PR #11810, opened 2026-06-09
Call shutdown before ScriptHost disposal

### LINKED ISSUE #11502: Worker shutdown does not always fail current executions

satvu:

resolves #11502

**Summary:** Calls Shutdown() before Dispose() on worker channels to ensure in-flight invocations are properly failed before the channel is disposed.

**Changes:**
- In both JobHostRpcWorkerChannelManager and WebHostRpcWorkerChannelManager, added Shutdown() calls prior to Dispose() in all channel teardown paths (shutdown, dispose, specialization).
- Wrapped each Shutdown() call in a try/catch to log errors without blocking disposal.
- Added corresponding unit tests for the new behavior.

### Pull request checklist


src/WebJobs.Script.Grpc/Rpc/WebHostRpcWorkerChannelManager.cs
```diff
@@ -321,9 +339,18 @@ internal void ScheduleShutdownStandbyChannels()
                                         else
                                         {
                                             IRpcWorkerChannel workerChannel = channelTask.Result;
-                                            if (workerChannel != null)
+                                            if (workerChannel is not null)
                                             {
-                                                (channelTask.Result as IDisposable)?.Dispose();
+                                                try
+                                                {
+                                                    workerChannel.Shutdown(null);
```
satvu: When `null` is passed in, customer is still informed through the default exception for this scenario that the worker channel is shutting down. See `Shutdown()` implementation [here](https://github.com/Azure/azure-functions-host/blob/455594d5c25191c644a0ced055e08f4c0434f2ff/src/WebJobs.Script.Grpc/Channel/WorkerChannel.cs#L1557).

src/WebJobs.Script.Grpc/Rpc/JobHostRpcWorkerChannelManager.cs
```diff
@@ -49,7 +49,15 @@ public Task<bool> ShutdownChannelIfExistsAsync(string channelId, Exception worke
                     {
                         string id = rpcChannel.Id;
                         _logger.LogDebug("Disposing language worker channel with id:{workerId}", id);
-                        rpcChannel.Shutdown(workerException);
+
+                        try
```
jviau: Consider adding a helper or extension method to simplify this code duplication:

``` CSharp
channel.ShutdownAndDispose(exception, logger);


public static void ShutdownAndDispose(this WorkerChannel channel, Exception ex, ILogger logger)
{
    ArgumentNullException.ThrowIfNull(channel);
    try
    {
        channel.Shutdown(ex);
        (channel as IDisposable)?.Dispose();
    }
    catch (Exception ex)
    {
        logger.LogError(ex, "Error failing executions on shutdown for language worker channel with id:{workerId}", channel.Id);
    }
}
```

jviau (approved):

test/WebJobs.Script.Tests/Workers/Rpc/JobHostRpcWorkerChannelManagerTests.cs
```diff
@@ -208,5 +209,110 @@ public void TestGetChannels_WithLanguage()
...
+
+            _workerChannelJava1.Verify(wc => wc.Shutdown(null), Times.Once);
+            _workerChannelJava2.Verify(wc => wc.Shutdown(null), Times.Once);
+            _workerChannelJs1.Verify(wc => wc.Shutdown(null), Times.Once);
+            _workerChannelJs2.Verify(wc => wc.Shutdown(null), Times.Once);
+            Assert.Equal(0, _jobHostRpcWorkerChannelManager.GetChannels().Count());
+        }
     }
-}
+}
```
kshyju: nit: add a line break at the end of the file if you push another iteration.

kshyju (commented):
@satvu My Agent review flagged this:

>_Ordering is never asserted. No MockSequence/InSequence anywhere — tests only check that both Shutdown and Dispose were each called Times.Once. But order is the entire fix: if someone reordered to Dispose-then-Shutdown, the bug returns (Dispose drains invocations before Shutdown can fault them) and every test still passes. This is the most important missing assertion. Cheap to close with a shared MockSequence or a callback recording call order_.

kshyju (approved):

PR #11813, opened 2026-06-10
Add additive platform notification for sync triggers

alrod:

Adds an opt-in additive platform notification for sync triggers. When the new `FUNCTIONS_SYNC_TRIGGERS_MODE` environment variable is set to `Platform`, the host fires a best-effort notification (`operation=update-triggers`) to the in-pod mesh server after a successful sync triggers operation, so the platform can asynchronously fetch and persist the triggers payload via its own pipeline.

### Iteration 1 scope (coordinated across repos)

- **This PR (azure-functions-host):** best-effort `NotifyTriggersChanged()` to the mesh server after a successful `SetTriggersAsync`. The call is awaited so any failure is observed and logged, but swallowed so the front-end sync result is unaffected.
- **Companion PR #11697 (azure-functions-host):** relaxes `GET /admin/host/triggers` to a platform JWT claim so the platform can pull the triggers payload from any pod.
- Companion platform-side changes track the notification and asynchronously fetch the triggers payload from the host.

All of the above are required for end-to-end behavior. Until activated, this PR is a no-op.

### Behavior

| `FUNCTIONS_SYNC_TRIGGERS_MODE` | Behavior |
|---|---|
| unset / `FrontEnd` (default) | Unchanged. Existing front-end sync (POST `/operations/settriggers`), blob-hash commit, background sync, key rotation, and customer sync flows all behave exactly as today. |
| `Platform` | Front-end sync runs first (unchanged). On success, an additional `NotifyTriggersChanged()` is awaited. Failures are logged as a warning and swallowed -- the operation result is determined entirely by the existing front-end sync. |

### Design notes

- **Additive, not a replacement.** The existing `SetTriggersAsync` body and `TrySyncTriggersAsync` flow are unchanged. The only new behavior is a single call to `NotifyPlatformIfEnabledAsync()` after a successful sync, which is a no-op when the mode is `FrontEnd`.
- **Backward-compatible constructor.** `FunctionsSyncManager` gains one optional `IMeshServiceClient` constructor parameter (defaults to `NullMeshServiceClient.Instance`). Existing call sites continue to compile and behave identically.
- **No DI fork.** A single `IFunctionsSyncManager` registration; mode selection happens inside the manager.
- **Internal mode enum.** `TriggerSyncMode` is `internal` -- not committed to the host's public API surface.
- **Network-safe default.** When the registered `IMeshServiceClient` is `NullMeshServiceClient` (e.g., dev/test or non-Linux-Consumption environments), the notification is a no-op even with the flag enabled, so misconfiguration cannot cause harm.

### Out of scope (future iterations)

- Transactional guarantees between the notification and platform-side persistence (e.g., synchronous ack from the platform before committing the blob hash).
- Removing the front-end sync once the platform has its own authoritative store.
- Removing the background sync at host startup.
- Stale-blob-hash recovery scenarios.

### Testing

- New unit tests: `FunctionsSyncManagerPlatformNotificationTests` (3 tests covering mode-on/mode-off/notification-failure-swallowed paths).
- New `MeshServiceClientTests` (2 tests covering the new `NotifyTriggersChanged` operation, written truly async with no sync-over-async pattern).
- Existing `FunctionsSyncManagerTests` and `WebFunctionsManagerTests` updated to stub the new env var (strict-mock setups).
- Verified zero new regressions vs. clean `dev` baseline (same 4 pre-existing test failures requiring local infrastructure).

### Pull request checklist

> IMPORTANT: Currently, changes must be backported to the `in-proc` branch to be included in Core Tools and non-Flex deployments.

- [x] Backporting to the `in-proc` branch is not required (platform infrastructure for Flex Consumption)
- [x] My changes do not require documentation changes
- [x] My changes should be added to the release notes for the next release
- [x] My changes do not need to be backported to a previous version
- [x] My changes do not require diagnostic events changes
- [x] I have added all required tests (Unit tests)


src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -146,6 +150,14 @@ public async Task<TriggersOperationResult> TrySyncTriggersAsync(bool isBackgroun
...
                         await UpdateHashAsync(hashBlobClient, newHash);
                     }
+                    if (success)
+                    {
+                        // Iteration 1: when FUNCTIONS_SYNC_TRIGGERS_MODE=Platform, fire a best-effort
+                        // notification to the in-pod mesh server so the platform can asynchronously
+                        // fetch and persist the triggers payload. No-op when mode is the default
+                        // (FrontEnd). Failures are swallowed -- the front-end sync already succeeded.
+                        await NotifyPlatformIfEnabledAsync();
+                    }
```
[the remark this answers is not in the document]
alrod: Good catch — fixed in d0d0df9. "fire-and-forget" was misleading because the call is awaited. Renamed throughout to "best-effort" (XML docs + PR body) so the latency profile is honest. The mesh server is on localhost:6060 so added latency is minimal in practice, and any failure is logged and swallowed so the front-end sync result is never affected.

src/WebJobs.Script.WebHost/Management/IMeshServiceClient.cs
```diff
@@ -21,5 +21,7 @@ public interface IMeshServiceClient
         Task NotifyHealthEvent(ContainerHealthEventType healthEventType, Type source, string details);
 
         Task CreateBindMount(string sourcePath, string targetPath);
+
+        Task NotifyTriggersChanged();
```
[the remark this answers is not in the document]
alrod: Acknowledged but skipping. `IMeshServiceClient` is platform-internal in practice — consumed by the host's `WebJobs.Script.WebHost` assembly, not shipped as a contract for external implementers. Both implementers in this repo (`MeshServiceClient`, `NullMeshServiceClient`) provide explicit impls. Following the precedent from #11690 which added `NotifyHealthEvent` the same way. Adding a single DIM here would also make the interface internally inconsistent (every other member is abstract).

src/WebJobs.Script/Environment/EnvironmentExtensions.cs
```diff
@@ -371,6 +371,24 @@ public static bool IsLinuxConsumptionOnLegion(this IEnvironment environment)
             return IsConsumptionOnLegion(environment) && environment.WebsiteSkuIsDynamic();
         }
 
+        /// <summary>
+        /// Gets the configured <see cref="TriggerSyncMode"/> for this host instance.
+        /// The mode is read once at startup via the <see cref="FunctionsSyncTriggersMode"/> environment variable.
+        /// Defaults to <see cref="TriggerSyncMode.FrontEnd"/> when the variable is unset or not a recognized value.
+        /// </summary>
```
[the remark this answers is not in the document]
alrod: Fixed in d0d0df9 — reworded the summary to describe what the extension actually does (reads and parses on each call), removing the misleading "read once at startup" wording. In 5a37886 also dropped a follow-up verbose `<remarks>` block about caller-side caching — that's already obvious from the implementation pattern in `FunctionsSyncManager`.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -289,6 +297,36 @@ internal async Task UpdateHashAsync(BlobClient hashBlobClient, string hash)
...
+        /// notification to the in-pod mesh server so the platform can asynchronously fetch and persist
+        /// the triggers payload. No-op when mode is the default <see cref="TriggerSyncMode.FrontEnd"/>.
+        /// Exceptions are logged and swallowed so the caller's sync triggers result is never affected.
+        /// </summary>
+        /// <remarks>
+        /// Exposed as <see langword="internal"/> for direct unit testing.
+        /// </remarks>
+        internal async Task NotifyPlatformIfEnabledAsync()
+        {
+            if (_triggerSyncMode != TriggerSyncMode.Platform)
```
mathewc: The enum made sense on the other PR because there it was an either or mode - FrontEnd OR Platform. In this PR it's an additive thing, so I think it makes sense to simplify to just using a single env var, e.g. FUNCTIONS_NOTIFY_PLATFORM_ON_SYNC

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -289,6 +297,36 @@ internal async Task UpdateHashAsync(BlobClient hashBlobClient, string hash)
             }
         }
 
+        /// <summary>
+        /// Iteration-1 additive notification. When <see cref="TriggerSyncMode.Platform"/> is enabled
```
mathewc: I'd remove this "Iteration-1 additive notification." - seems like unnecessary copilot verbiage. Also you could consider simplifying the description to remove all the platform details, and just say something like "Performs a best-effort notification to the platform when triggers are synchronized.".

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -289,6 +297,36 @@ internal async Task UpdateHashAsync(BlobClient hashBlobClient, string hash)
...
+            if (_triggerSyncMode != TriggerSyncMode.Platform)
+            {
+                return;
+            }
+
+            try
+            {
+                await _meshServiceClient.NotifyTriggersChanged();
+            }
+            catch (Exception ex)
```
mathewc: use the `catch (Exception exc) when (!exc.IsFatal())` pattern that we use elsewhere to avoid swallowing fatal exceptions

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -289,6 +297,36 @@ internal async Task UpdateHashAsync(BlobClient hashBlobClient, string hash)
...
+            {
+                // Best-effort. The platform will catch up on the next sync triggers operation
```
mathewc: Let's remove all these comments and simplify the warning ("Failed to notify platform of triggers change." suffices) - I don't think we need all these external details here.

test/WebJobs.Script.Tests.Integration/Management/FunctionsSyncManagerPlatformNotificationTests.cs
```diff
@@ -0,0 +1,100 @@
...
+using Microsoft.Azure.WebJobs.Script.WebHost.Management;
+using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.Extensions.Options;
+using Moq;
+using Xunit;
+
+namespace Microsoft.Azure.WebJobs.Script.Tests.Integration.Management
+{
+    /// <summary>
+    /// Tests for the iteration-1 additive behavior in <see cref="FunctionsSyncManager"/>:
```
mathewc: Same comment about these copilot "iteration-1" comments

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
alrod: Done in edf7213. Removed the `TriggerSyncMode` enum and replaced it with a single boolean env var `FUNCTIONS_NOTIFY_PLATFORM_ON_SYNC` (read via `IEnvironment.ShouldNotifyPlatformOnSync()`). The notification path is now a simple `if (!_notifyPlatformOnSync) { return; }` guard.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
alrod: Done in edf7213. Simplified the summary to "Performs a best-effort notification to the platform when triggers are synchronized." and dropped the iteration-1 verbiage.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
alrod: Done in edf7213. Switched to `catch (Exception exc) when (!exc.IsFatal())`.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
alrod: Done in edf7213. Removed the comments; the warning is now just "Failed to notify platform of triggers change."

test/WebJobs.Script.Tests.Integration/Management/FunctionsSyncManagerPlatformNotificationTests.cs
alrod: Done in edf7213. Removed the iteration-1 wording from the test class doc.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -146,6 +150,10 @@ public async Task<TriggersOperationResult> TrySyncTriggersAsync(bool isBackgroun
                     {
                         await UpdateHashAsync(hashBlobClient, newHash);
                     }
+                    if (success)
+                    {
+                        await NotifyPlatformIfEnabledAsync();
```
mathewc, line 155: As it stands, it looks like you don't have any tests that verify that on a sync, mesh is actually called? I.e. a test that would fail if this line were removed. For reference, there are existing tests like TrySyncTriggers_PostsExpectedContent that use a mock http client to verify we call the FE operation.

E.g. a Theory test `TrySyncTriggers_NotifiesPlatform_WhenEnabled(bool functionsNotifyPlatformOnSync)` that tests both enabled/disabled

alrod: Good call. Added `TrySyncTriggers_NotifiesPlatform_WhenEnabled(bool functionsNotifyPlatformOnSync)` in 5f0bd3b — it runs the full `TrySyncTriggersAsync` flow with a mock `IMeshServiceClient` and verifies `NotifyTriggersChanged()` is called `Times.Once` when enabled and `Times.Never` when disabled. Confirmed it fails if the notification call is removed from `TrySyncTriggersAsync`.

mathewc (approved):

soninaren (approved):

PR #11816, opened 2026-06-10
fix: avoid health check auth for non-expanded requests

### LINKED ISSUE #11787: Intermittent BlobServiceClient creation issue
## Summary

When `AzureWebJobsStorage` is configured for identity-based access using `AzureWebJobsStorage__blobServiceUri`, the host can fail to create the `BlobServiceClient` during early startup with:

```
Could not create BlobServiceClient. Exception: System.InvalidOperationException: Unable to find matching constructor while trying to create an instance of BlobServiceClient. Expected one of the follow sets of configuration parameters:
 1. connectionString
 2. serviceUri
 3. serviceUri, credential:accountKey
 ...
```

The root cause is a race between `HostAzureBlobStorageProvider` consumers and `ActiveHostConfigurationSource` loading the script host's `IConfiguration`. Setting `AzureWebJobsStorage__accountName` (instead of, or in addition to, `__blobServiceUri`) avoids the failure and is a known workaround.

## Affected source files

- https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/StorageProvider/HostAzureBlobStorageProvider.cs
- https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/StorageProvider/BlobServiceClientProvider.cs
- https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/StorageProvider/StorageServiceUriOptions.cs
- https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Config/ActiveHostConfigurationSource.cs

## Root cause analysis

`HostAzureBlobStorageProvider` builds its `IConfiguration` by layering `ActiveHostConfigurationSource` on top of the WebHost-level `IConfiguration`:

```csharp
Configuration = new ConfigurationBuilder()
    .Add(new ActiveHostConfigurationSource(scriptHostManager))
    .AddConfiguration(configuration)
    .Build();
```

`ActiveHostConfigurationSource.Load()` only populates its data when the script host's `IConfiguration` is resolvable from `IScriptHostManager.Services`:

```csharp
if ((_scriptHostManager as IServiceProvider)?.GetService(typeof(IConfiguration)) is IConfigurationRoot activeHostConfiguration)
{
    // populate Data
}
```

Before the script host is fully initialized, this provider is empty. If a caller invokes `TryCreateBlobServiceClientFromConnection` during that window, `BlobServiceClientProvider.CreateClient` runs:

```csharp
var serviceUri = configuration.Get<StorageServiceUriOptions>().GetBlobServiceUri();
if (serviceUri != null)
{
    return new BlobServiceClient(serviceUri, tokenCredential, options);
}
return base.CreateClient(configuration, tokenCredential, options);
```

Because the merged configuration does not yet contain `blobServiceUri`, `StorageServiceUriOptions.GetBlobServiceUri()` returns `null` and the code falls through to `AzureComponentFactory.CreateClient`. That factory does not recognize `blobServiceUri` - it only understands `connectionString`, `serviceUri`, or `accountName` - so it throws the "Unable to find matching constructor" exception above.

## Why __accountName works as a workaround

`AzureWebJobsStorage__accountName` is recognized by both code paths:

1. `StorageServiceUriOptions.AccountName` constructs `https://{accountName}.blob.core.windows.net`.
2. `AzureComponentFactory.CreateClient` fallback also accepts `accountName` natively.

`AzureWebJobsStorage__blobServiceUri` is only recognized by path (1). When path (1) fails because of the race, there is no fallback that understands the key. This is why customers who set `__accountName` don't hit the failure even when the same race condition occurs.

## Repro

A unit test reproducing the race is on the `config-bug-repro` branch:

- Branch: https://github.com/Azure/azure-functions-host/tree/config-bug-repro
- Test file: https://github.com/Azure/azure-functions-host/blob/config-bug-repro/test/WebJobs.Script.Tests/Storage/HostAzureBlobStorageProviderTests.cs

The test uses an `IScriptHostManager` stub whose `Services.GetService(typeof(IConfiguration))` returns `null` until a signal method is called, then raises `ActiveHostChanged`. It asserts:

1. `TryCreateBlobServiceClient_BeforeActiveHostLoaded_FailsThenSucceedsAfterActiveHostChanged` - with `__blobServiceUri` only in the active host configuration, `TryCreateBlobServiceClientFromConnection` fails before the active host is loaded and succeeds after `ActiveHostChanged` fires.
2. `TryCreateBlobServiceClient_AccountNameInWebHostConfig_SucceedsEvenBeforeActiveHostLoaded` - documents that `__accountName` succeeds regardless of the race because of the `AzureComponentFactory` fallback path.

## Expected behavior

`HostAzureBlobStorageProvider` should reliably honor `AzureWebJobsStorage__blobServiceUri` regardless of whether the script host's `IConfiguration` has been merged via `ActiveHostConfigurationSource` at the time of the call.

## Suggested fix directions

- Ensure the WebHost-level `IConfiguration` passed into `HostAzureBlobStorageProvider` already contains identity-based settings (e.g., via the environment variables provider), so the active host merge is not on the critical path.
- Make `BlobServiceClientProvider.CreateClient` defer or retry when the configuration section exists but contains no usable URI / account / connection string.
- Eagerly trigger `ActiveHostConfigurationSource.Load()` (or block on first access) so callers see a fully merged view.

## Workaround

Set `AzureWebJobsStorage__accountName` to the storage account name. For sovereign clouds or custom DNS where `__accountName` alone isn't sufficient, set `__accountName` alongside `__blobServiceUri`; the presence of `__accountName` will satisfy the `AzureComponentFactory` fallback even when the race occurs.eComponentFactory` fallback even when the race occurs.

jviau:

resolves #11787 

Alternative approach to mitigate #11787 with a more minimal change. Health checks are ultimately what trigger secret evaluation earlier in the specialization flow as part of AuthN/AuthZ flow. This change will just make non-expanded health checks unauthenticated. These calls don't need auth as all they return is HTTP 200 or 500, and a JSON payload of "healthy", "degraded" or "unhealthy" - no possibility of sensitive information.

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.WebHost/Diagnostics/HealthChecks/HealthCheckHelpers.cs
+++ b/src/WebJobs.Script.WebHost/Diagnostics/HealthChecks/HealthCheckHelpers.cs
@@ -1,0 +1,24 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics.HealthChecks
+{
+    public static class HealthCheckHelpers
+    {
+        /// <summary>
+        /// Determines if the health check request is for an expanded view.
+        /// </summary>
+        /// <param name="request">The request to check.</param>
+        /// <returns>true if the request is for an expanded view; otherwise, false.</returns>
+        public static bool IsExpandedHealthCheck(HttpRequest request)
+        {
+            ArgumentNullException.ThrowIfNull(request);
+            return request.Query.TryGetValue("expand", out StringValues value)
+                && bool.TryParse(value, out bool expand) && expand;
+        }
+    }
+}
--- a/test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SamplesEndToEndTests_CSharp.cs
+++ b/test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SamplesEndToEndTests_CSharp.cs
@@ -367,6 +367,21 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.EndToEnd
         }
 
         [Theory]
+        [InlineData("/admin/health")]
+        [InlineData("/admin/health/live")]
+        [InlineData("/admin/health/ready")]
+        public async Task HealthCheck_NoToken_Succeeds(string uri)
+        {
+            // token specified as bearer token
+            HttpRequestMessage request = new(HttpMethod.Get, uri);
+            HttpResponseMessage response = await _fixture.Host.HttpClient.SendAsync(request);
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+
+            string body = await response.Content.ReadAsStringAsync();
+            Assert.Equal("{\"status\":\"Healthy\"}", body);
+        }
+
+        [Theory]
         [InlineData("/admin/health?expand=true", null)]
         [InlineData("/admin/health/live?expand=true", HealthCheckTags.Liveness)]
         [InlineData("/admin/health/ready?expand=true", HealthCheckTags.Readiness)]
@@ -404,9 +419,9 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.EndToEnd
         }
 
         [Theory]
-        [InlineData("/admin/health")]
-        [InlineData("/admin/health/live")]
-        [InlineData("/admin/health/ready")]
+        [InlineData("/admin/health?expand=true")]
+        [InlineData("/admin/health/live?expand=true")]
+        [InlineData("/admin/health/ready?expand=true")]
         public async Task HealthCheck_NoAdminToken_Fail(string uri)
         {
             // token specified as bearer token
--- a/release_notes.md
+++ b/release_notes.md
@@ -5,3 +5,4 @@
 -->
 
 - fix: address config reload concurrent read/write race (#11815)
+- fix: avoid health checks triggering secret-manager too early (#11816)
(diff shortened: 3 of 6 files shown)


src/WebJobs.Script.WebHost/Diagnostics/HealthChecks/HealthCheckHelpers.cs
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics.HealthChecks
+{
+    public static class HealthCheckHelpers
```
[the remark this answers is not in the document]
jviau: This is not a publicly consumed assembly.

anandagopal6 (approved):

src/WebJobs.Script.WebHost/Diagnostics/HealthChecks/HealthCheckHelpers.cs
```diff
@@ -0,0 +1,24 @@
...
+        /// <param name="request">The request to check.</param>
+        /// <returns>true if the request is for an expanded view; otherwise, false.</returns>
+        public static bool IsExpandedHealthCheck(HttpRequest request)
+        {
+            ArgumentNullException.ThrowIfNull(request);
+            return request.Query.TryGetValue("expand", out StringValues value)
+                && bool.TryParse(value, out bool expand) && expand;
+        }
+    }
+}
```
satvu: nit: add trailing newline to follow repo convention

satvu (approved):

satvu (approved):

liliankasem (approved):

jviau:

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SamplesEndToEndTests_CSharp.cs
```diff
@@ -404,9 +419,9 @@ public async Task HealthCheck_AdminToken_ExpandSucceeds(string uri, string tag)
         }
 
         [Theory]
-        [InlineData("/admin/health")]
-        [InlineData("/admin/health/live")]
-        [InlineData("/admin/health/ready")]
+        [InlineData("/admin/health?expand=true")]
```
mathewc, line 422: The test name needs to be updated since it no longer fails - e.g. HealthCheck_NoAdminToken_ReturnsMinimal

PR #11829, opened 2026-06-15
copy controller test apps to temp folder to prevent races

brettsam:
### The bug

Multiple test fixtures (`ControllerScenarioTestFixture`, fixtures derived from `EndToEndTestFixture`) point the host's `ScriptPath` at a shared source directory under `sample\csharp` and `TestScripts\`. When that directory contains a CSX function with a `function.proj` (e.g. `HttpTrigger-Compat`), the host kicks off `dotnet restore` as a background process during indexing. When the restore exits, the [`Process.Exited` handler in `PackageManager.cs`](https://github.com/Azure/azure-functions-host/blob/455594d5c25191c644a0ced055e08f4c0434f2ff/src/WebJobs.Script/Description/DotNet/PackageManager.cs#L83-L100) does an unguarded `File.Copy` into the shared script folder.

If the next fixture has already started and is reading that same `project.assets.json`, the copy throws `IOException` on a threadpool thread, which crashes the test host and aborts the run.

### The fix

This is a **targeted test-side fix**: each affected fixture now copies its script root into a unique `%TEMP%` directory before the host starts, so fixtures no longer share writable state. The copy/cleanup logic was already in `EndToEndTestFixture`; it's extracted into a small `TemporaryScriptRoot` helper and reused by `ControllerScenarioTestFixture`.

We deliberately are not changing `PackageManager`. The unguarded handler is real, but it's in legacy CSX-restore code that's been in production for years and is risky to modify just to satisfy a test setup. Isolating the fixtures is the smaller, lower-risk change and fully eliminates the race we're seeing in CI.

Also sets `ScriptJobHostOptions.Functions = ["HttpTrigger"]` so the host only indexes the single function these tests actually exercise — this avoids the CSX restore entirely and drops the suite runtime down.

### Pull request checklist

### DIFF
--- a/test/WebJobs.Script.Tests.Integration/Controllers/ControllerScenarioTestFixture.cs
+++ b/test/WebJobs.Script.Tests.Integration/Controllers/ControllerScenarioTestFixture.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 using System;
 using System.IO;
@@ -8,6 +8,7 @@ using System.Web.Http;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.TestHost;
 using Microsoft.Azure.WebJobs.Script.Config;
+using Microsoft.Azure.WebJobs.Script.Tests.Integration.Fixtures;
 using Microsoft.Azure.WebJobs.Script.WebHost;
 using Microsoft.Azure.WebJobs.Script.WebHost.Diagnostics;
 using Microsoft.Extensions.DependencyInjection;
@@ -20,6 +21,7 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.Controllers
     {
         private ScriptSettingsManager _settingsManager;
         private HttpConfiguration _config;
+        private TemporaryScriptRoot _scriptRoot;
 
         public ScriptApplicationHostOptions HostOptions { get; private set; }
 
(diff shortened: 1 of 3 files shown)


test/WebJobs.Script.Tests.Integration/Controllers/ControllerScenarioTestFixture.cs
```diff
@@ -63,16 +70,21 @@ public virtual async Task InitializeAsync()
...
                         o.LogPath = Path.Combine(Path.GetTempPath(), @"Functions", "ControllerScenarioTests");
                         o.SecretsPath = Path.Combine(Path.GetTempPath(), @"FunctionsTests\Secrets");
                         o.HasParentScope = true;
 
                         HostOptions = o;
                     });
+
+                    services.PostConfigure<ScriptJobHostOptions>(o =>
+                    {
+                        o.Functions = ["HttpTrigger"];
```
jviau, line 66: This is net-new, what prompted this change?

brettsam: See last comment in description:

> Also sets ScriptJobHostOptions.Functions = ["HttpTrigger"] so the host only indexes the single function these tests actually exercise — this avoids the CSX restore entirely and drops the suite runtime down.

... turns out these tests has no need to index, restore, etc, any of the other 20 samples in this app. They only check anything for this one function. So restricting the functions we load to only theses should speed things up and prevent runaway restores anyway.

... at least that's the theory and they all passed locally. We'll see if it works in CI.

test/WebJobs.Script.Tests.Integration/Controllers/ControllerScenarioTestFixture.cs
```diff
@@ -86,7 +98,7 @@ public virtual async Task InitializeAsync()
         }
 
         public Task DisposeAsync()
-        { 
+        {
             return Task.CompletedTask;
         }
```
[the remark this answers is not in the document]
brettsam: dispose was running... but i did remove it for clarity and to remove the sync-over-async stopasync() call. Everything is now in DisposeAsync.

jviau (approved):

kshyju (approved):

PR #11831, opened 2026-06-15
Only cache positive SecretsEnabled results

mathewc:
Addresses startup time race condition issues described in https://github.com/Azure/azure-functions-host/issues/11787. Specifically, we recently saw a case where a configuration race condition during specialization lead to the host caching a failed result for SecretsEnabled, and the host didn't recover.

### DIFF
--- a/src/WebJobs.Script.WebHost/Security/KeyManagement/DefaultSecretManagerProvider.cs
+++ b/src/WebJobs.Script.WebHost/Security/KeyManagement/DefaultSecretManagerProvider.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
@@ -25,7 +25,7 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
         private readonly StartupContextProvider _startupContextProvider;
         private readonly IAzureBlobStorageProvider _azureBlobStorageProvider;
         private Lazy<ISecretManager> _secretManagerLazy;
-        private Lazy<bool> _secretsEnabledLazy;
+        private volatile bool _secretsEnabled;
 
         public DefaultSecretManagerProvider(IOptionsMonitor<ScriptApplicationHostOptions> options, IHostIdProvider hostIdProvider, IEnvironment environment,
             ILoggerFactory loggerFactory, IMetricsLogger metricsLogger, HostNameProvider hostNameProvider, StartupContextProvider startupContextProvider,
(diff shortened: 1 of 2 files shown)


src/WebJobs.Script.WebHost/Security/KeyManagement/DefaultSecretManagerProvider.cs
```diff
@@ -56,19 +55,29 @@ public bool SecretsEnabled
...
                 }
-                return _secretsEnabledLazy.Value;
+
+                // Only cache positive results, to ensure we're responsive to changes in configuration that enable secrets.
+                // Re-evaluating on each call until we can confirm secrets are enabled allows the host to recover automatically
+                // once the underlying configuration becomes available.
+                bool secretsEnabled = GetSecretsEnabled();
+                if (secretsEnabled)
+                {
+                    _secretsEnabled = true;
```
kshyju, line 69: If thread A (reader) evaluates `GetSecretsEnabled() == true`, and is about to cache it
and thread B (`ResetSecretManager()` via options change) sets `_secretsEnabled = false` and swaps `_secretManagerLazy` and thread A then runs `_secretsEnabled = true` Wouldn't thread B's reset get lost here?, leaving the flag stuck to true?

jviau (commented):
I have two concerns overall:

1. An existing problem, but also not addressed in this PR. This class has a structural split brain problem. The two existing independent lazys (with this PR: bool and a lazy), they evaluate the same config but at different points in time, meaning they can have inconsistent results if a caller does not always guard their access to `Current` with `SecretsEnabled`.

Can we make a change where it is one source of truth? Can `SecretsEnabled` property resolve the actual `.Current` lazy, and then check it is non-null?

2. Not caching the failure can lead to an increase in exceptions logged (and maybe even request latency) if we now continually check for the secrets enablement when they are legitimately disabled. I guess first is answering, is it ever a valid scenario to have secrets disabled? The code seems to suggest we are trying to account for that. 

I think we may want to cache the failure once ActiveHost is fully loaded? And thus we can trust we have the full config source.

mathewc:
@jviau Thanks. For the first concern, this is actually by design. For better or worse, a long time ago I introduced an explicit guard pattern in https://github.com/Azure/azure-functions-host/pull/7931 - so SecretsEnabled can't use Current, because it's a guard for that.

For #2, yes secret-less is a valid scenario (e.g. static web apps), and that scenario is also referenced in the PR linked above. So your point on the problem with not caching false is actually a problem - for secret-less scenarios we don't want to allow the logs/handled errors to continue to happen for the lifetime of the host.

I think I'm going to reevaluate the plan for this fix. This avoidance of caching is just working around the fundamental problem that we're trying to initialize things before configuration reload is complete (race conditions). I think the right approach may be to move to handling the IConfiguration change token as the trigger for reset. That way we're sure configuration is done reloading before we reset.

mathewc:
On further reflection I'm going to close this PR, since the changes proposed here are just working around the issues described in https://github.com/Azure/azure-functions-host/pull/11811. DefaultSecretManagerProvider is already operating correctly - it resets itself on specialization, and at that time all config should be available and reliable. So the right fix isn't in this class, but will come as part of the config changes @jviau mentioned.

PR #11839, opened 2026-06-18
isolating and improving timeout tests

brettsam:
~~The `TimeoutTest_UsingToken_CSharp` would sometimes fail -- and it seemed to be whenever the CI machine was reporting health issues (low-memory). Analysis suggests that the tight while loop this test used may have caused it to fail in such conditions by consuming a thread while waiting for the test scenario to run.~~

~~Changing the behavior in the test function to be more cpu-friendly by reacting directly to the token rather than spinning. Also swapping the other test scenario to use a `Task` rather than blocking a thread.~~

~~Not 100% guaranteed that this will fix the issue, but it seems like the right direction.~~

~~It's entirely possible that the low-memory conditions are caused by other tests leaking somewhere, but that's analysis I'll be doing elsewhere.~~

Edited:
It turns out that the 3 timeout tests were leaking massive memory. During some runs they'd use upwards of 2.4gb. Dump analysis showed tens of thousands of CancellationToken registrations that appeared to come from configuration somewhere. 

The one big oddity with these tests was that they wired up both the *production* configuration and the *test* configurations. It's possible that something in there was causing some cyclical change token registration causing chaos.

In reality, these tests are testing *in-proc cancellation* and likely don't even impact anything we do today which is strictly out-of-proc. So rather than chase it down to the exact culprit, I:
- removed the production registration from the tests (memory usage plummetted)
- kept the optimization from before with the test method itself to avoid spinning
- separated the timeout tests into their own process

I've run this several times, both with and without memory diagnostics running and it seems stable now.

### Pull request checklist

### DIFF
--- a/eng/ci/templates/official/jobs/run-non-e2e-tests.yml
+++ b/eng/ci/templates/official/jobs/run-non-e2e-tests.yml
@@ -112,6 +112,17 @@ jobs:
       arguments: '--filter "Category!=E2E&Group=NonE2E_WebHost" -c release --no-build'
       projects: $(test_projects)
     env:
+      AzureWebJobsStorage: $(Storage) 
+
+  - task: DotNetCoreCLI@2
+    displayName: 'Timeout tests'
+    condition: succeededOrFailed()
+    inputs:
+      command: test
+      testRunTitle: 'Timeout'
+      arguments: '--filter "Category!=E2E&Group=NonE2E_Timeout" -c release --no-build'
+      projects: $(test_projects)
+    env:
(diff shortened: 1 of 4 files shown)


cjaliaga (approved):

soninaren (approved):

brettsam:
ok that clearly didn't work -- adding diagnostics to see if we can catch it and determine where mem is leaking. i'll re-ping when it's sready for another review

kshyju (approved):

PR #11840, opened 2026-06-18
Refactor to check full path in secrets repository

satvu:

Adds a check to validate file path in secrets repository file.

### Pull request checklist

    * [] Otherwise: I've added my notes to `release_notes.md`


src/WebJobs.Script.WebHost/Security/KeyManagement/FileSystemSecretsRepository.cs
```diff
@@ -162,8 +162,18 @@ private string GetSecretsFilePath(ScriptSecretsType secretsType, string function
 
         private string GetFunctionSecretsFilePath(string functionName)
         {
+            ArgumentException.ThrowIfNullOrEmpty(functionName);
+
             string secretFileName = string.Format(CultureInfo.InvariantCulture, "{0}.json", functionName);
-            return Path.Combine(_secretsPath, secretFileName);
+            string fullPath = Path.GetFullPath(Path.Combine(_secretsPath, secretFileName));
+            string fullSecretsPath = Path.GetFullPath(_secretsPath + Path.DirectorySeparatorChar);
```
kshyju: Should we use `Path.Combine` instead of `+` ? What if `_secretsPath` ends with the separator char already?

satvu: Updated to handle existing separator char but using `Path.Combine` wouldn't work if we had `Path.Combine("/secrets", "/")` (returns just `"/"`), and we want to check if we're inside the secrets folder and need the separator char.

src/WebJobs.Script.WebHost/Security/KeyManagement/FileSystemSecretsRepository.cs
```diff
@@ -162,8 +162,18 @@ private string GetSecretsFilePath(ScriptSecretsType secretsType, string function
 
         private string GetFunctionSecretsFilePath(string functionName)
         {
+            ArgumentException.ThrowIfNullOrEmpty(functionName);
+
             string secretFileName = string.Format(CultureInfo.InvariantCulture, "{0}.json", functionName);
-            return Path.Combine(_secretsPath, secretFileName);
+            string fullPath = Path.GetFullPath(Path.Combine(_secretsPath, secretFileName));
+            string fullSecretsPath = Path.TrimEndingDirectorySeparator(Path.GetFullPath(_secretsPath)) + Path.DirectorySeparatorChar;
```
kshyju: Should this be precomputed (in constructor) and stored in a local variable? looks like _secretsPath is immutable.

src/WebJobs.Script.WebHost/Security/KeyManagement/FileSystemSecretsRepository.cs
```diff
@@ -162,8 +162,18 @@ private string GetSecretsFilePath(ScriptSecretsType secretsType, string function
...
+
+            if (!fullPath.StartsWith(fullSecretsPath, StringComparison.OrdinalIgnoreCase))
+            {
+                throw new InvalidOperationException($"Invalid function name '{functionName}'. The resolved path is outside the secrets directory.");
```
kshyju: How is this exception handled in the caller side? Just want to make sure it does not cause any unexpected behavior, since this is a new behavior (even in a non happy path)

kshyju (approved):
Approving as it mostly looks good to me. Good to get clarity on the 2 comments.

src/WebJobs.Script.WebHost/Security/KeyManagement/FileSystemSecretsRepository.cs
satvu: Here's the call chain where the exception bubbles up:

```
HTTP Request (e.g., GET /admin/functions/{name}/keys/{keyName})
  └─ KeysController.Get / Post / Delete
       └─ GetFunctionKeys(functionName)
            └─ SecretManager.GetFunctionSecretsAsync(functionName)
                 └─ LoadSecretsAsync<FunctionSecrets>(functionName)
                      └─ _repository.ReadAsync(type, keyScope)
                           └─ GetSecretsFilePath(type, functionName)
                                └─ GetFunctionSecretsFilePath(functionName)  ← throws here
```

The exception is **unhandled** in this chain — no `try/catch` wraps it between `GetFunctionSecretsFilePath` and the controller action. It will propagate up to the ASP.NET Core middleware pipeline, which returns a **500 Internal Server Error** to the caller.

This is acceptable since a path-traversal attempt is a malicious/invalid input that shouldn't reach this code under normal operation (and `IsFunction()` in the controller already validates the function exists). If you'd prefer a **400 Bad Request**, you could catch `InvalidOperationException` in `SecretManager` or the controller and return `BadRequest()` instead.

release_notes.md
```diff
@@ -8,3 +8,4 @@
 - fix: avoid health checks triggering secret-manager too early (#11816)
 - Restrict GET admin/host/triggers to platform claim (#11697)
 - fix: `SetProcessCountToNumberOfCpuCores` silently overriding `MaxProcessCount` on high core count machines (#11842)
+- Fixed a bug where `SetProcessCountToNumberOfCpuCores` would silently override `MaxProcessCount` on high core count machines, spawning more worker processes than the configured maximum. `MaxProcessCount` is now respected as a hard ceiling. (#11842)
```
liliankasem: dupe of line 10

src/WebJobs.Script.WebHost/Security/KeyManagement/FileSystemSecretsRepository.cs
```diff
@@ -162,8 +164,17 @@ private string GetSecretsFilePath(ScriptSecretsType secretsType, string function
...
         {
+            ArgumentException.ThrowIfNullOrEmpty(functionName);
+
             string secretFileName = string.Format(CultureInfo.InvariantCulture, "{0}.json", functionName);
-            return Path.Combine(_secretsPath, secretFileName);
+            string fullPath = Path.GetFullPath(Path.Combine(_secretsPath, secretFileName));
+
+            if (!fullPath.StartsWith(_normalizedSecretsPath, StringComparison.OrdinalIgnoreCase))
+            {
+                throw new InvalidOperationException($"Invalid function name '{functionName}'. The resolved path is outside the secrets directory.");
```
liliankasem: ArgumentException makes more sense here than invalid operation (it's bad input, not bad state)

src/WebJobs.Script.WebHost/Security/KeyManagement/FileSystemSecretsRepository.cs
```diff
@@ -162,8 +164,17 @@ private string GetSecretsFilePath(ScriptSecretsType secretsType, string function
 
         private string GetFunctionSecretsFilePath(string functionName)
         {
+            ArgumentException.ThrowIfNullOrEmpty(functionName);
+
             string secretFileName = string.Format(CultureInfo.InvariantCulture, "{0}.json", functionName);
-            return Path.Combine(_secretsPath, secretFileName);
+            string fullPath = Path.GetFullPath(Path.Combine(_secretsPath, secretFileName));
```
liliankasem: Do we want to use the `_normalizedSecretsPath` here instead?

`string fullPath = Path.GetFullPath(Path.Combine(_normalizedSecretsPath, secretFileName));`

liliankasem (approved):

kshyju (approved):

PR #11856, opened 2026-06-26
Modifying Admin Isolation rules for Kudu sync triggers operations

mathewc:
Currently when Admin Isolation is enabled, SCM (Kudu) invoked Sync Trigger operations are rejected on a technicality since they go through the Antares Front End. From an operations point of view they are platform internal requests and should be allowed.

MSFT Internal: The full design doc for this feature can be found on https://eng.ms/ - search for "**Admin Isolation: Kudu Sync Triggers Fix**". This PR implements "Option A".

### DIFF
--- a/test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SamplesEndToEndTests_CSharp.cs
+++ b/test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SamplesEndToEndTests_CSharp.cs
@@ -268,6 +268,8 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.EndToEnd
         {
             var environment = this._fixture.Host.WebHostServices.GetService<IEnvironment>();
             string originalWebsiteInstanceId = environment.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebsiteInstanceId);
+            string originalWebsiteSku = environment.GetEnvironmentVariable(EnvironmentSettingNames.AzureWebsiteSku);
+            string originalAdminIsolationEnabled = environment.GetEnvironmentVariable(EnvironmentSettingNames.FunctionsAdminIsolationEnabled);
 
             environment.SetEnvironmentVariable(EnvironmentSettingNames.AzureWebsiteSku, sku);
             environment.SetEnvironmentVariable(EnvironmentSettingNames.AzureWebsiteInstanceId, websiteInstanceId);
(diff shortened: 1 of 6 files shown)


test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SamplesEndToEndTests_CSharp.cs
```diff
@@ -310,6 +310,90 @@ public async Task AdminIsolation_ReturnsExpectedStatus(string uri, string sku, s
...
+
+                using (var httpClient = _fixture.Host.CreateHttpClient())
+                {
+                    HttpResponseMessage response = await httpClient.SendAsync(request);
+                    Assert.Equal(expectedStatus, response.StatusCode);
+                }
+            }
+            finally
+            {
+                environment.SetEnvironmentVariable(EnvironmentSettingNames.FunctionsAdminIsolationEnabled, null);
```
fabiocav: This test's cleanup only restores `AzureWebsiteInstanceId`. `FunctionsAdminIsolationEnabled` is nulled in finally rather than restored (and nulled even when the test never set it), and `AzureWebsiteSku` is never reset. Since the class uses IClassFixture, the IEnvironment is shared across all tests, so these values leak past the test and could affect siblings that rely on the ambient SKU or isolation flag. It passes today only because the fixture starts with them unset. This mirrors the existing AdminIsolation_ReturnsExpectedStatus, so it isn't a new, but let's fix it here.

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SamplesEndToEndTests_CSharp.cs
```diff
@@ -310,6 +310,90 @@ public async Task AdminIsolation_ReturnsExpectedStatus(string uri, string sku, s
...
+        // Existing isolation allowances are unaffected: platform-internal callers still succeed.
+        [InlineData(true, true, false, SyncTriggersAuthMode.MasterKey, HttpStatusCode.OK)]
+        // Existing isolation allowances are unaffected: AppService-internal (Front End bypassed) callers still succeed.
+        [InlineData(true, false, true, SyncTriggersAuthMode.MasterKey, HttpStatusCode.OK)]
+        // With isolation disabled, behavior is unchanged from today for both master key and SCM token callers.
+        [InlineData(false, false, false, SyncTriggersAuthMode.MasterKey, HttpStatusCode.OK)]
+        [InlineData(false, false, false, SyncTriggersAuthMode.ScmSiteToken, HttpStatusCode.OK)]
+        public async Task SyncTriggers_AdminIsolation_ReturnsExpectedStatus(bool enableIsolation, bool isPlatformInternal, bool bypassFE, SyncTriggersAuthMode authMode, HttpStatusCode expectedStatus)
+        {
+            var environment = this._fixture.Host.WebHostServices.GetService<IEnvironment>();
```
fabiocav: nit: remove `this.`

mathewc: Fixed in both places.

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/SamplesEndToEndTests_CSharp.cs
mathewc: Fixed both tests

test/WebJobs.Script.Tests/Security/Authorization/AuthorizationOptionsExtensionsTests.cs
```diff
@@ -0,0 +1,115 @@
...
+            {
+                // The presence of this header indicates the request was routed through the Front End,
+                // i.e. it is not an internal (Front End bypassing) request.
+                httpContext.Request.Headers[ScriptConstants.AntaresLogIdHeaderName] = Guid.NewGuid().ToString();
+            }
+
+            var actionDescriptor = new ControllerActionDescriptor
+            {
+                MethodInfo = typeof(HostController).GetMethod(actionName)
+            };
```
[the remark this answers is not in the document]
mathewc: Fixed

test/WebJobs.Script.Tests/Security/Authorization/AuthorizationOptionsExtensionsTests.cs
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        private static AssertionRequirement GetAdminAuthLevelAssertion()
+        {
+            var options = new AuthorizationOptions();
+            options.AddScriptPolicies();
+
+            AuthorizationPolicy policy = options.GetPolicy(PolicyNames.AdminAuthLevel);
+            return policy.Requirements.OfType<AssertionRequirement>().Single();
+        }
```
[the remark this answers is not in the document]
mathewc: If the policy is removed, the PolicyNames member will be removed and this won't compile. Won't fix.

fabiocav (approved):

PR #11874, opened 2026-07-16
preventing SyncTriggers on placeholder hosts

### LINKED ISSUE #11873: Apps get stuck showing only the placeholder `WarmUp` function after a deploy (SyncTriggers race during specialization)
A function app can end up with the platform/portal showing **only the placeholder `WarmUp` function** instead of the customer's real functions. The **host itself is healthy and running the real functions** — the problem is that the *trigger metadata* published to the platform (via SyncTriggers) is stuck on the placeholder `WarmUp` payload.

Root cause is a **race between a deployment's SyncTriggers call and a specialization that times out**. During the ~30s specialization-timeout window, an inbound (foreground) SyncTriggers request is served by the still-placeholder host and publishes the `WarmUp` payload. Because foreground SyncTriggers **bypass the hash blob**, the platform never self-heals, even though the host recovers a few seconds later and runs the real functions locally.

Related: #10169, https://github.com/Azure/functions-action/issues/245

## The mechanism

A deploy does two things effectively simultaneously:

1. **Restarts the app**, forcing a cold start through the placeholder → specialize path (worker `FunctionEnvironmentReloadRequest`, guarded by the 30s `EnvironmentReloadTimeout`).
2. **Fires SyncTriggers** (Kudu post-deploy hook + the client tool, often 2–3× per deploy).

When the first env-reload times out, the SyncTriggers request that is parked/inbound during that window is completed by the placeholder host, which publishes the placeholder `WarmUp` trigger payload.

### Per-pod timeline (observed, repeated on essentially every pod)

```
T+0s    Starting language worker channel specialization      (placeholder env-reload begins)
T+30s   Specialization failed. Shutting down.                (30s EnvironmentReloadTimeout hit)
        -> StandbyManager fault handler calls StopApplication()
        -> the parked, inbound /admin/host/synctriggers completes during graceful drain
        -> WarmUp SetTriggers published (FOREGROUND sync)
T+31s   Starting language worker channel specialization      (immediate retry — worker now warm)
T+31s   Completed language worker channel specialization
        Restart requested. Reason 'Host specialization.'
T+33s   Job host started                                     (REAL functions now running locally)
```

The host recovers within ~3s and runs the real functions — but the last **published** trigger payload is `WarmUp`, and nothing corrects it.

## Why it doesn't self-heal.

`FunctionsSyncManager.TrySyncTriggersAsync` only reads/writes the hash blob at `synctriggers/{hostId}/last` when `isBackgroundSync == true`:

- **Foreground** syncs (inbound `/admin/host/synctriggers`, which is what the bad `WarmUp` publish is) **publish but never touch the hash blob**.
- **Background** syncs check the hash and skip publishing when `Last == Current`.

Sequence that wedges the app:

1. A prior healthy **background** sync writes the *real-functions* hash into the blob.
2. The bad **foreground** `WarmUp` publish overwrites the live trigger payload but **leaves the hash blob unchanged** (still the real-functions hash).
3. The host recovers, recomputes the real-functions hash, compares to the blob → **they match** → background sync concludes "nothing changed" and **suppresses the corrective re-publish**.

Result: the live platform view stays on `WarmUp` until an **external forced sync** (portal Refresh, ARM `host/default/sync`, or another deploy) happens to re-publish.

## Impact

- Portal and any ARM/trigger-metadata consumer show only `WarmUp`; customer functions appear "missing."
- Event-driven scaling (scale controller reads trigger metadata) is degraded — the platform sees only the placeholder trigger.
- The running host is actually healthy, so HTTP invocations against the real functions may still work, which makes the symptom confusing and hard to diagnose.
- Self-heals only on the next forced/background-detected change; can persist indefinitely otherwise.

## Suggested fixes (for discussion)

1. **Never publish placeholder/`WarmUp` triggers.** Guard SetTriggers so a host that is in placeholder / not-yet-specialized (or is shutting down due to specialization failure) refuses to publish trigger metadata, or publishes a no-op instead of the `WarmUp` payload. This closes the window directly.
2. **Make foreground syncs update the hash blob** (or invalidate it) so a subsequent background sync detects the mismatch and re-publishes the correct payload — restoring self-heal.
3. **Reconcile after successful specialization.** When the host successfully specializes and loads real functions, force a corrective SyncTriggers (bypassing the hash short-circuit) so the last-published payload always reflects the specialized host.
4. **Address the deterministic 30s first-attempt env-reload timeout** on the Flex dotnet-isolated placeholder path (why does the first `FunctionEnvironmentReloadRequest` reliably exceed 30s, then succeed instantly on retry?).

Any one of (1)–(3) prevents the stuck state; (4) reduces how often the window opens.

brettsam:
Fixes #11873

Certain flows (namely worker specialization failures) could result in a SyncTriggers call being made against a placeholder host. This would result in a Warmup function being sent to the platform and could cause issues with scaling and metadata.

This change checks whether the current host is a placeholder and if so, skip the sync. Previous behavior checked if the *environment* was a placeholder but did not validate the active host.

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
+++ b/src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
@@ -111,6 +111,18 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Management
                 return result;
             }
 
+            // The active host may still be the placeholder/standby host (e.g. mid-specialization,
+            // after placeholder mode has been cleared but before RestartHostAsync swaps in the
+            // specialized host).
+            if (Utility.TryGetHostService(_scriptHostManager, out IOptions<ScriptJobHostOptions> jobHostOptions)
+                && jobHostOptions.Value.IsStandbyConfiguration)
+            {
+                result.Success = false;
+                result.Error = "Active host is in standby configuration. Skipping SyncTriggers to avoid publishing placeholder triggers.";
+                _logger.LogWarning(result.Error);
+                return result;
+            }
+
             try
             {
                 await _syncSemaphore.WaitAsync();
--- a/release_notes.md
+++ b/release_notes.md
@@ -5,6 +5,7 @@
 -->
 
 - Fixed a WebHost worker channel shutdown race that could hang host teardown when a language worker was still initializing. (#11853)
+- Fixed a race where a SyncTriggers request during specialization could publish placeholder (`WarmUp`) triggers. (#11874)
 - Update Python Worker Version to [4.45.1](https://github.com/Azure/azure-functions-python-worker/releases/tag/4.45.1) (#11860)
 - Update Microsoft.Azure.AppService.Middleware to 1.5.11 (#11866)
 - Update PS 7.4 worker to [v4.0.5212](https://github.com/Azure/azure-functions-powershell-worker/releases/tag/v4.0.5212) (#11858)
--- a/test/WebJobs.Script.Tests.Integration/Management/FunctionsSyncManagerTests.cs
+++ b/test/WebJobs.Script.Tests.Integration/Management/FunctionsSyncManagerTests.cs
@@ -412,9 +412,10 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.Managment
                 Assert.Equal(0, _mockHttpHandler.RequestCount);
                 Assert.Equal(0, _contentBuilder.Length);
 
-                // The operation is a successful no-op, not an error.
-                Assert.True(result.Success);
-                Assert.Null(result.Error);
+                // The standby guard blocks the publish and reports failure, matching the existing
+                // standby/placeholder skip contract (surfaced as a 500 by the admin/host/synctriggers endpoint).
+                Assert.False(result.Success);
+                Assert.Equal("Active host is in standby configuration. Skipping SyncTriggers to avoid publishing placeholder triggers.", result.Error);
             }
         }
 
--- a/test/WebJobs.Script.Tests.Integration/Management/FunctionsSyncManagerTests.cs
+++ b/test/WebJobs.Script.Tests.Integration/Management/FunctionsSyncManagerTests.cs
@@ -330,6 +330,95 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.Managment
         }
 
         [Fact]
+        public async Task TrySyncTriggers_PlaceholderWarmupPayload_DoesNotPublish()
+        {
+            // Repro for the SyncTriggers race in azure-functions-host#10169.
+            //
+            // When a specialization times out (the placeholder env-reload exceeds its 30s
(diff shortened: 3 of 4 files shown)


src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -111,6 +111,18 @@ public async Task<TriggersOperationResult> TrySyncTriggersAsync(bool isBackgroun
...
+            // after placeholder mode has been cleared but before RestartHostAsync swaps in the
+            // specialized host).
+            if (Utility.TryGetHostService(_scriptHostManager, out IOptions<ScriptJobHostOptions> jobHostOptions)
+                && jobHostOptions.Value.IsStandbyConfiguration)
+            {
+                result.Success = false;
+                result.Error = "Active host is in standby configuration. Skipping SyncTriggers to avoid publishing placeholder triggers.";
+                _logger.LogWarning(result.Error);
+                return result;
+            }
```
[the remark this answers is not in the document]
brettsam: I made this behavior match the other "still in placeholder mode" behavior immediately above this check. I just hadn't update the test. That is now done.

jviau (approved):

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
```diff
@@ -111,6 +111,18 @@ public async Task<TriggersOperationResult> TrySyncTriggersAsync(bool isBackgroun
                 return result;
             }
 
+            // The active host may still be the placeholder/standby host (e.g. mid-specialization,
+            // after placeholder mode has been cleared but before RestartHostAsync swaps in the
+            // specialized host).
+            if (Utility.TryGetHostService(_scriptHostManager, out IOptions<ScriptJobHostOptions> jobHostOptions)
```
fabiocav, line 117: Looking at the code, I think there is still a narrow window where `TrySyncTriggersAsync` can be called without requiring a running host. During a host restart, resolving the active JobHost services may fail while the metadata manager still holds cached metadata from the placeholder host. In that case, this guard is bypassed and the placeholder function metadata could still be published.

This change significantly reduces the exposure, but we should flag the remaining gap.

We could follow up with things like clearing cached metadata during host restarts or recording which JobHost/configuration produced it, so we could detect and ensure it's never part of sync triggers, but ultimately, the fact that function metadata is not scoped to the JobHost is really problematic.

I don’t consider this a blocker because the change already protects against the common scenario, but wanted to flag the gap.

fabiocav (approved):
Approving, but left a comment and I think a follow up is required.

src/WebJobs.Script.WebHost/Management/FunctionsSyncManager.cs
brettsam: Created #11877

PR #11883, opened 2026-07-21
fail env reload fast on worker exit during specialization

### LINKED ISSUE #11876: Worker crash during `FunctionEnvironmentReload` (specialization) blocks for the full 30s timeout instead of failing fast
### Summary
When a language worker process crashes while handling a `FunctionEnvironmentReloadRequest` (i.e. during specialization), the host does **not** detect the crash for the purpose of the reload. Instead, `SpecializeAsync` waits the full `EnvironmentReloadTimeout` (default 30s) before failing with a `TimeoutException`, even though the worker process died within milliseconds. This adds a ~30s stall to specialization/cold start whenever the worker crashes on reload.

### Environment
- Component: `Microsoft.Azure.WebJobs.Script.Grpc` (WebHost worker channel / specialization)
- Applies to placeholder → specialize (env reload) path on Linux/Windows.

### Expected behavior
If the worker process exits while a `FunctionEnvironmentReload` is pending, specialization should fail fast (surface the worker exit as the failure cause) rather than waiting for the reload timeout to elapse.

### Actual behavior
Specialization blocks for the full `EnvironmentReloadTimeout` (30s) and then fails with a generic `TimeoutException`, masking the real cause (worker crash) and delaying recovery/host restart.

### Root cause
`SpecializeAsync` awaits the reload completion source, which is only completed on the gRPC response or on the reload timeout — a worker process crash never completes it.

Trace (paths from `dev`):

1. `WebHostRpcWorkerChannelManager.SpecializeAsync()` awaits `rpcWorkerChannel.SendFunctionEnvironmentReloadRequest()`.
   - `WebHostRpcWorkerChannelManager.cs` → `envReloadRequestResultSuccessful = await rpcWorkerChannel.SendFunctionEnvironmentReloadRequest();`
2. `WorkerChannel.SendFunctionEnvironmentReloadRequest()` returns `_reloadTask.Task`.
   - `WorkerChannel.cs:710` → `return _reloadTask.Task;`
3. `_reloadTask` (a `TaskCompletionSource<bool>`, `WorkerChannel.cs:85`) is completed in exactly two places:
   - **Success:** on receiving `FunctionEnvironmentReloadResponse` → `SetResult` (`WorkerChannel.cs:499/503`).
   - **Timeout:** the `PendingItem` registered for the response (`WorkerChannel.cs:698-699`) uses `CancellationTokenSource.CancelAfter(EnvironmentReloadTimeout)`. On expiry, `PendingItem.OnTimeout` throws `TimeoutException` → fault handler `HandleWorkerEnvReloadError` → `_reloadTask.SetException` (`WorkerChannel.cs:1367`).
4. A worker crash flows through a **completely separate** path that never touches `_reloadTask`:
   - `WorkerProcess.OnProcessExited` (`WorkerProcess.cs:219`) → `RpcWorkerProcess.HandleWorkerProcessExitError` → `PublishNoThrow(new WorkerErrorEvent(...))` (`RpcWorkerProcess.cs:84-86`).
   - `WorkerErrorEvent` is consumed by `RpcFunctionInvocationDispatcher.WorkerError` (`RpcFunctionInvocationDispatcher.cs:476`) for dispatcher-level restart handling.
   - Nothing in this path completes/faults `_reloadTask`, so the pending reload is orphaned until the 30s timer fires.

### Proposed fix
When the worker process exits while a reload is still pending, fault the reload completion source so specialization fails fast:
- Have the `WorkerChannel` observe its own process exit / `WorkerErrorEvent`, and if `_reloadTask` is still pending, call `_reloadTask.TrySetException(...)` (e.g. wrapping the `WorkerProcessExitException`).
- Use `TrySetException`/`TrySetResult` (not `SetException`/`SetResult`) throughout the reload-completion paths to avoid races with the timeout/response callbacks completing the TCS first.

### Impact
- Adds ~30s to specialization/cold start on every worker crash-during-reload.
- Masks the real failure cause (`TimeoutException` instead of the worker exit), making diagnosis harder.

### Notes
- `EnvironmentReloadTimeout` default is 30s (`WorkerProcessCountOptions.cs:48`).
- Related deterministic-timeout behavior: #10169. (This issue is the "worker actually crashed" variant of that timeout window.)

brettsam:
fixes https://github.com/Azure/azure-functions-host/issues/11876

If an app crashed during env reload, we'd sit and wait 30 seconds for the timeout before continuing. This change fails that Task quickly now so that recovery happens quicker.

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.Grpc/Channel/WorkerChannel.cs
+++ b/src/WebJobs.Script.Grpc/Channel/WorkerChannel.cs
@@ -7,6 +7,7 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using System.Reactive.Linq;
 using System.Text;
 using System.Threading;
 using System.Threading.Channels;
@@ -83,6 +84,7 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
         private IDisposable _startLatencyMetric;
         private IEnumerable<FunctionMetadata> _functions;
         private TaskCompletionSource<bool> _reloadTask = new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously);
+        private volatile bool _reloadRequestSent;
         private TaskCompletionSource<List<RawFunctionMetadata>> _functionsIndexingTask = new TaskCompletionSource<List<RawFunctionMetadata>>(TaskCreationOptions.RunContinuationsAsynchronously);
         private TimeSpan _functionLoadTimeout = TimeSpan.FromMinutes(1);
         private bool _isSharedMemoryDataTransferEnabled;
@@ -152,6 +154,7 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
             _workerInvocationFailedMetric = string.Format(MetricEventNames.WorkerInvokeFailed, Id);
 
             LoadScriptJobHostOptions(_scriptHostManager as IServiceProvider);
+            _eventSubscriptions.Add(_eventManager.OfType<WorkerErrorEvent>().Subscribe(OnWorkerError));
         }
 
         protected virtual int WorkerProcessId => -1;
@@ -496,12 +499,14 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
                 {
                     _workerChannelLogger.LogWarning(reloadEnvironmentVariablesException, reloadEnvironmentVariablesException.Message);
                 }
-                _reloadTask.SetResult(false);
+                _reloadTask.TrySetResult(false);
             }
             else
             {
-                _reloadTask.SetResult(true);
+                _reloadTask.TrySetResult(true);
             }
+
+            _reloadRequestSent = false;
             latencyEvent.Dispose();
         }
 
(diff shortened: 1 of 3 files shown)


src/WebJobs.Script.Grpc/Channel/WorkerChannel.cs
```diff
@@ -689,6 +690,7 @@ internal FunctionLoadRequestCollection GetFunctionLoadRequestCollection(IEnumera
 
         public Task<bool> SendFunctionEnvironmentReloadRequest()
         {
+            _reloadTask = new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously);
```
jviau: How much confidence do you have with setting this field here? I wonder if there is any concern for us swapping the `_reloadTask`, and then one of the other methods which operates on `_reloadTask` is called but for the wrong "reload". As in our signals get crossed, we end up canceling the reload task for the wrong reload.

kshyju (approved):

src/WebJobs.Script.Grpc/Channel/WorkerChannel.cs
brettsam: I undid this. part in latest commit It's not supported today, so I really shouldn't be introducing something new like this.

jviau (approved):

PR #11886, opened 2026-07-21
Add function name validation to KeysController.IsFunction

satvu:
## Summary

Adds an early `Utility.IsValidFunctionName()` check at the top of `KeysController.IsFunction()` to reject invalid function names before any filesystem access occurs.

## Motivation

The `IsFunction` method is the primary gate for all function-scoped key management endpoints (`GET`/`POST`/`PUT`/`DELETE` on `/admin/functions/{name}/keys/...`). Previously, it would perform `Path.Combine` with the raw user-supplied function name and attempt to read a `function.json` from that path before checking the function metadata manager.

## Changes

- `KeysController.IsFunction()`: Added `Utility.IsValidFunctionName(functionName)` check that returns `false` early if the name doesn't match the existing regex `^[a-z][a-z0-9_\-]{0,127}$(?<!^host$)`.

## Testing

Existing tests continue to pass. The validation regex is already well-tested in `UtilityTests.IsValidFunctionNameTests`.

### DIFF
--- a/src/WebJobs.Script.WebHost/Controllers/KeysController.cs
+++ b/src/WebJobs.Script.WebHost/Controllers/KeysController.cs
@@ -285,6 +285,11 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost.Controllers
 
         private bool IsFunction(string functionName)
         {
+            if (!Utility.IsValidFunctionName(functionName))
+            {
+                return false;
+            }
+
             string functionPath = Path.Combine(_applicationOptions.Value.ScriptPath, functionName);
             if (Utility.TryReadFunctionConfig(functionPath, out _, _fileSystem))
             {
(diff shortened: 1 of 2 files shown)


jviau (approved):

src/WebJobs.Script.WebHost/Controllers/KeysController.cs
```diff
@@ -285,6 +285,11 @@ internal bool IsBuiltInSystemKeyName(string keyName)
 
         private bool IsFunction(string functionName)
         {
+            if (string.IsNullOrEmpty(functionName) || !Utility.IsValidFunctionName(functionName))
+            {
+                return false;
```
kshyju, line 290: Is it possible that existing function apps with invalid function name now behave differently (return 404 for the /keys controller call) after this change goes live? (I am not 100% sure it's possible to run an app with invalid function name). If that is the case, this would be a breaking change.

kshyju: @copilot Can you investigate?

kshyju (approved):

PR #11902, opened 2026-07-30
Fix for RPC initialization issue #11741

### LINKED ISSUE #11741: AppOffline GRPC Initialization Bug
During a Web Deploy presence of the AppOffline file can cause the WebHost to skip GRPC initialization, and even after the file is removed, the WebHost stays in that state, leading to worker connection failures. See **ICM #781438451** for more details.

RpcInitializationService is registered as a WebHost-level IManagedHostedService. Its StartAsync() is called once at WebHost startup. During MSDeploy deployment:

1) The AspNetCoreGrpcServer singleton is constructed via DI, selecting port 53591 via GetUnusedTcpPort()
2) RpcInitializationService.StartAsync() runs, sees app_offline.htm, and returns without calling _rpcServer.StartAsync() ? the Kestrel gRPC server never binds port 53591
3) MSDeploy removes app_offline.htm ? the host detects the deletion and restarts the ScriptHost
4) The ScriptHost restart transitions the host from Offline to Default and launches worker processes with --port 53591
5) But RpcInitializationService.StartAsync() is NOT re-invoked on ScriptHost restart (it is a WebHost service, not a ScriptHost service)
6) Workers connect() to port 53591 where nothing is listening ? SocketException 10013 ? unhandled crash ? exit code 0xE0434352
7) The host retries ~30 times over ~17 minutes, all failing identically on the same port, until it exhausts the retry threshold and recycles the entire process

The RpcInitializationService needs to monitor for when the app transitions to online and initialize itself.

mathewc:
Resolves https://github.com/Azure/azure-functions-host/issues/11741

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script.Grpc/Server/AspNetCoreGrpcServer.cs
+++ b/src/WebJobs.Script.Grpc/Server/AspNetCoreGrpcServer.cs
@@ -57,9 +57,9 @@ namespace Microsoft.Azure.WebJobs.Script.Grpc
             _logger.LogDebug($"Started {nameof(AspNetCoreGrpcServer)} on {address}.");
         }
 
-        public Task ShutdownAsync() => _grpcHost.StopAsync();
+        public Task ShutdownAsync() => _grpcHost?.StopAsync() ?? Task.CompletedTask;
 
-        public Task KillAsync() => _grpcHost.StopAsync();
+        public Task KillAsync() => _grpcHost?.StopAsync() ?? Task.CompletedTask;
 
         protected async ValueTask DisposeAsync(bool disposing)
         {
--- a/test/WebJobs.Script.Tests/Workers/Rpc/RpcInitializationServiceTests.cs
+++ b/test/WebJobs.Script.Tests/Workers/Rpc/RpcInitializationServiceTests.cs
@@ -44,13 +44,20 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.Workers.Rpc
         }
 
         [Fact]
-        public async Task RpcInitializationService_AppOffline()
+        public async Task RpcInitializationService_AppOffline_InitializesRpcServerButNotChannels()
         {
             IRpcServer testRpcServer = new TestRpcServer();
             var mockEnvironment = new Mock<IEnvironment>();
+
+            // Placeholder mode with a non-inproc runtime ensures InitializeChannelsAsync would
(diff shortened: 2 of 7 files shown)


src/WebJobs.Script.Grpc/Server/AspNetCoreGrpcServer.cs
```diff
@@ -57,9 +57,9 @@ public async Task StartAsync()
             _logger.LogDebug($"Started {nameof(AspNetCoreGrpcServer)} on {address}.");
         }
 
-        public Task ShutdownAsync() => _grpcHost.StopAsync();
+        public Task ShutdownAsync() => _grpcHost?.StopAsync() ?? Task.CompletedTask;
```
mathewc, line 60: Preexisting issue in this class - I was attempting to run my new regression test without the RpcInitializationService fix, and it was failing with a null ref here. With this guard, the regression test fails as expected - the host never reaches a Running state.

test/WebJobs.Script.Tests/Workers/Rpc/RpcInitializationServiceTests.cs
```diff
@@ -44,7 +44,7 @@ public RpcInitializationServiceTests()
         }
 
         [Fact]
-        public async Task RpcInitializationService_AppOffline()
+        public async Task RpcInitializationService_AppOffline_InitializesRpcServerButNotChannels()
```
fabiocav, line 47: This test leaves `AzureWebsitePlaceholderMode` unset, so `InitializeChannelsAsync()` returns before calling the channel manager even if the new app-offline guard is removed, so the `Times.Never` assertion does not protect the "but not channels" behavior named here. 

Configure placeholder mode (and Java runtime?) before constructing `RpcInitializationService`, then deleting the guard would call InitializeChannelAsync and make this regression test fail.

The issue predates the PR, but the test name change doesn't align with the behavior.

fabiocav (approved):
Flagging as approved, but there's a test issue that would be good to address.

test/WebJobs.Script.Tests/Workers/Rpc/RpcInitializationServiceTests.cs
mathewc: Fixed

fabiocav (approved):

brettsam (approved):

PR #11904, opened 2026-07-31
Never re-mint a system key that already exists in the secrets store

### LINKED ISSUE #11834: First EventGrid delivery fails due to cold-start race in Flex Consumption
During host startup, there's a brief window where HTTP traffic is accepted but the eventgrid_extension system key hasn't load from storage yet. Requests in that window get rejected. Retries succeed once the key is loaded. VNet integration adds additional latency to cold starts, which widens this window.

The solution should wait until the system key was loaded and not showing failures within EventGrid monitoring. **But even more important for me is that this behavior is documented.**

With "Always Ready" instances the problem was solved. Switching to managed-identity delivery on the Event Grid subscription would solve the problem as well, because it removes the system key dependency entirely.

#### Additional context
Having this feature in place or the limitation at least documented would help Azure user. For example for a solution, which reacts to an event, which occurs only several times a year scaling down to zero and not using Always Ready would be nice.

zahalzel2:

No public issue - found while investigating an internal incident where Event Grid deliveries to a Functions app started returning 401 after a host restart. Happy to open a tracking issue if the team prefers one.

Related but distinct: #11834 reports EventGrid 401s during Flex Consumption cold start. That failure is transient - the reporter notes retries succeed once the key finishes loading. This one is permanent, because the key is overwritten rather than merely not-yet-loaded, so retries never recover. **This PR does not fix #11834.**

### What's wrong

`DefaultScriptWebHookProvider` looked up an extension's system key in the in-memory host-secrets cache. On Flex Consumption, specialization seeds that cache from a startup context whose `SystemKeys` can be incomplete. When it is: the lookup misses, the provider generates a replacement, and `AddOrUpdateFunctionSecretAsync` blindly upserts it - overwriting the key the subscribing service already holds.

The webhook URL then carries a key the subscription doesn't have, so every delivery 401s. It is not self-healing: the host has no way to notify Event Grid, and the subscription stores the full URL, so recovery requires re-creating the subscription by hand.

The root cause is that a **present-but-incomplete** cache was treated as authoritative. A stale "no" was indistinguishable from a real one, and only one of those should authorize a write.

### The fix

Move get-or-create into `SecretManager`, the only component that can tell them apart:

- `GetOrCreateSystemKeyAsync` returns a cached hit unchanged. On a miss it invalidates only the snapshot it missed on (`Interlocked.CompareExchange`, so a concurrent reload isn't discarded), then performs one authoritative repository read before concluding the key doesn't exist.
- Creation is gated on `Created`/`Updated`. Any other result means nothing was persisted, and returning the value would mint a webhook URL carrying a key that authorizes nothing.
- `DefaultScriptWebHookProvider` now only derives the key name.

`AddOrUpdateFunctionSecretAsync` is unchanged - `KeysController` still depends on its result codes for HTTP status mapping.

### Known limitation

The create path is still a last-writer-wins upsert, so a key persisted concurrently by another instance can be replaced. That window only opens during true first creation - not on restarts, scale-out, or specialization. Closing it needs an ETag/CAS write no `ISecretsRepository` exposes today. The interface documents this rather than over-promising.

### Tests

On the miss path: zero `WriteAsync`/`WriteSnapshotAsync`, exactly one `ReadAsync`, and the invalidated cache heals. A second test proves a cached hit never touches the repository. A third seeds a host document missing only the target key and asserts exactly one write whose persisted ciphertext decrypts to the returned value. `DefaultScriptWebHookProviderTests` verifies the exact derived key name.

Mutation-checked: disabling the invalidation fails with `Expected invocation on the mock should never have been performed, but was 1 times: WriteAsync` - the incident itself, caught by assertion rather than inferred from a returned value.

### Pull request checklist

    * **Maintainer input wanted.** The trigger is Flex-specific (specialization seeding a partial cache) but the code path is shared. I don't know whether in-proc can reach the same state, so I'm not asserting either way.
    * This is customer-visible, so it should be - the `release_notes.md` entry is included in this PR.
    * **Maintainer input wanted** - depends on which released versions carry the same path.
    * No events added or changed. The existing `Host secret ... Updated` event simply stops firing spuriously.

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -4,3 +4,4 @@
 - My change description (#PR)
 -->
 - Update Python Worker Version to [4.45.1](https://github.com/Azure/azure-functions-python-worker/releases/tag/4.45.1) (#11860)
+- Fixed extension system keys being overwritten on host restart when the startup context supplies an incomplete secrets snapshot, which broke Event Grid and other webhook deliveries with 401s (#11904)
--- a/release_notes.md
+++ b/release_notes.md
@@ -3,6 +3,4 @@
 <!-- Please add your release notes in the following format:
 - My change description (#PR)
 -->
-- Update Python Worker Version to [4.45.1](https://github.com/Azure/azure-functions-python-worker/releases/tag/4.45.1) (#11860)
 - Fixed extension system keys being overwritten on host restart when the startup context supplies an incomplete secrets snapshot, which broke Event Grid and other webhook deliveries with 401s (#11904)
--->
--- a/src/WebJobs.Script.WebHost/Security/KeyManagement/ISecretManager.cs
+++ b/src/WebJobs.Script.WebHost/Security/KeyManagement/ISecretManager.cs
@@ -70,6 +70,16 @@ namespace Microsoft.Azure.WebJobs.Script.WebHost
         Task PurgeOldSecretsAsync(string rootScriptPath, ILogger logger);
 
         /// <summary>
+        /// Returns the value of the specified system key, creating and persisting one only after an
+        /// authoritative read shows none exists. Unlike <see cref="AddOrUpdateFunctionSecretAsync"/>
+        /// it will not replace a key it can observe; the create path remains a last-writer-wins
+        /// upsert, so a key persisted concurrently by another host instance can still be replaced.
+        /// </summary>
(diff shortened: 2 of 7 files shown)


zahalzel2:
@microsoft-github-policy-service agree company="Microsoft"

PR #11914, opened 2026-08-06
ci: onboard in-proc restores to Central Feed Service

jviau:
## Summary

- route NuGet and npm dependency restores through Central Feed Service
- centralize NuGet authentication in the .NET setup template
- update Az.Storage lease handling for hosted build agents

Backports #11903 and #11909 to the `in-proc` branch.

### DIFF
--- a/eng/ci/templates/install-dotnet.yml
+++ b/eng/ci/templates/install-dotnet.yml
@@ -17,3 +17,5 @@ steps:
   inputs:
     packageType: sdk
     useGlobalJson: true
+
+- task: NuGetAuthenticate@1
--- a/test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/EndToEndTestFixture.cs
+++ b/test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/EndToEndTestFixture.cs
@@ -126,7 +126,7 @@ namespace Microsoft.Azure.WebJobs.Script.Tests
             var extensionsToInstall = GetExtensionsToInstall();
             if (extensionsToInstall != null && extensionsToInstall.Length > 0)
             {
-                TestFunctionHost.WriteNugetPackageSources(_copiedRootPath, "https://azfunc.pkgs.visualstudio.com/e6a70c92-4128-439f-8012-382fe78d6396/_packaging/AzureFunctionsTempStaging/nuget/v3/index.json", "https://api.nuget.org/v3/index.json");
+                TestFunctionHost.WriteNugetPackageSources(_copiedRootPath, "https://pkgs.dev.azure.com/azfunc/public/_packaging/upstream-public/nuget/v3/index.json");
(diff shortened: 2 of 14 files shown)


satvu (approved):

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/EndToEndTestFixture.cs
```diff
@@ -126,7 +126,7 @@ string GetDestPath(int counter)
             var extensionsToInstall = GetExtensionsToInstall();
             if (extensionsToInstall != null && extensionsToInstall.Length > 0)
             {
-                TestFunctionHost.WriteNugetPackageSources(_copiedRootPath, "https://azfunc.pkgs.visualstudio.com/e6a70c92-4128-439f-8012-382fe78d6396/_packaging/AzureFunctionsTempStaging/nuget/v3/index.json", "https://api.nuget.org/v3/index.json");
+                TestFunctionHost.WriteNugetPackageSources(_copiedRootPath, "https://pkgs.dev.azure.com/azfunc/public/_packaging/upstream-public/nuget/v3/index.json");
```
fabiocav, line 129: This only updates the `ExtensionsManager` restore. The C# sample tests also restore a copied `function.proj` from `%TEMP%` Do we need to backport #11911 here (or add the same user-level NuGet config setup) so all restores go through CFS?

eng/ci/templates/install-dotnet.yml
```diff
@@ -17,3 +17,5 @@ steps:
   inputs:
     packageType: sdk
     useGlobalJson: true
+
+- task: NuGetAuthenticate@1
```
fabiocav, line 21: This only authenticates jobs that use `install-dotnet.yml`. Do we need to add `NuGetAuthenticate@1` to `BuildArtifactsLinux` as it installs the SDK directly and then runs `dotnet restore` ?

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/EndToEndTestFixture.cs
jviau: Yes, I will be backporting those. I was going to do another round of backports after this, but I can add to this PR if you want

eng/ci/templates/install-dotnet.yml
jviau: Good catch. I added `NuGetAuthenticate@1` to `BuildArtifactsLinux` immediately before its restore, since that job bypasses `install-dotnet.yml` and needs credentials for the authenticated feeds.

test/WebJobs.Script.Tests.Integration/WebHostEndToEnd/EndToEndTestFixture.cs
jviau: ported that PR

kshyju (approved):

PR #11916, opened 2026-08-10
extracting new Azure.Functions.Rpc.Server assembly in preparation for compute separation

brettsam:
High-level this is moving our server-based Grpc functionality (including worker process management) into a new assembly called `Azure.Fucntions.Rpc.Server`. The base `WebJobs.Grpc` assembly continues to hold common functionality that we will use in a follow-up to introduce the "compute separation" functionality in an assembly called `Azure.Functions.Rpc.Client`. 

Details are below, but only a few interfaces had to change to separate out some process-management that was baked into our lowest-level interfaces.

## Summary

This PR extracts the existing host-managed Function RPC server and worker-process topology from `Microsoft.Azure.WebJobs.Script.Grpc` into a new `Azure.Functions.Rpc.Server` assembly.

The existing Grpc assembly remains the shared FunctionRpc contract, channel protocol, and ScriptHost invocation core. The implementation is intentionally dominated by repository-aware file moves; moved types retain their existing namespaces and behavior.

## Assembly layout

Arrows point from consumer to dependency:

```text
Microsoft.Azure.WebJobs.Script.WebHost
    ├──> Azure.Functions.Rpc.Server             (this PR)
    │       ├──> Microsoft.Azure.WebJobs.Script.Grpc
    │       └──> Microsoft.Azure.WebJobs.Script
    └──> Microsoft.Azure.WebJobs.Script.Grpc
                └──> Microsoft.Azure.WebJobs.Script

Azure.Functions.Rpc.Client                      (future PR)
    └──> Microsoft.Azure.WebJobs.Script.Grpc
```

### `Microsoft.Azure.WebJobs.Script.Grpc` — shared core

Retains:

- The protobuf source and sole generated FunctionRpc contract; generation remains server-only in this PR.
- The internal `WorkerChannel` protocol engine, capabilities, message dispatch, events, conversions, and shared-memory helpers.
- Process-neutral channel contracts and RPC channel state.
- Dispatcher contracts/state, `WorkerFunctionInvoker`, descriptor providers, lifecycle services, and shutdown coordination.
- HTTP/custom-handler configuration models and shared `AddRpcScriptHostCoreServices` registration.

### `Azure.Functions.Rpc.Server` — standard host-managed topology

Owns:

- The inbound ASP.NET Core gRPC server and `FunctionRpcService`.
- Worker process management and process-backed gRPC/HTTP channels.
- RPC channel managers, worker process factories, initialization, and load balancing.
- Standard dispatchers, concurrency management, metadata discovery, and worker managers.
- Concrete HTTP/custom-handler execution.
- `AddScriptGrpc`, `AddRpcScriptHostServices`, and `AddCommonRpcServices`.

### `Azure.Functions.Rpc.Client` — future outbound topology

A future PR will add the outbound client topology, including `Grpc.Net.Client`, connection/stream management, client channels, and a client dispatcher. It will reference the shared Grpc assembly, not Server.

### Channel contracts are split by responsibility

- `IWorkerChannel` remains in Grpc and is now process-neutral: it exposes only `Id` and `GetWorkerStatusAsync`.
- The former broad `IRpcWorkerChannel` contract is renamed to `IFunctionRpcChannel` and remains in Grpc with its protocol members unchanged.
- Server introduces `IWorkerProcessChannel`, which adds `WorkerProcess` and `StartWorkerProcessAsync`.
- Server's `IRpcWorkerChannel` composes `IFunctionRpcChannel` and `IWorkerProcessChannel`.
- `IHttpWorkerChannel` moves to Server and inherits `IWorkerProcessChannel`.

`WorkerChannel` now implements `IFunctionRpcChannel` and no longer requires every future channel topology to expose or launch a local process. It retains `WorkerProcessId`, resource-disposal hooks, and the complete protocol/disposal implementation.

`GrpcWorkerChannel` implements the Server composite interface. Its process-start and disposal bodies are unchanged; only the now-unnecessary `override` modifiers were removed from the process members.

### DI registration movement is minimized

Registration methods that reference Server-owned concrete types must move to Server. To keep this mechanical:

- `AddScriptGrpc` moves with its existing class, namespace, method name, and call sites unchanged.
- The original `RpcServiceCollectionExtensions` class moves to Server with `AddRpcScriptHostServices` and `AddCommonRpcServices` unchanged.
- Only the newly separated shared registrations receive a new Grpc-owned class/method: `RpcScriptHostCoreServiceCollectionExtensions.AddRpcScriptHostCoreServices`.
- Existing call sites and registration ordering remain unchanged.

### Package ownership follows implementation ownership

- Server now directly owns `Grpc.AspNetCore`, `Mono.Posix.NETStandard`, and `Microsoft.Azure.WebJobs.Rpc.Core`.
- Grpc directly owns `Google.Protobuf`, `Grpc.Core.Api`, and private `Grpc.Tools`, because it owns protobuf generation.
- `Grpc.Net.Client` is not introduced in this PR.

### Temporary visibility preserves a mechanical extraction

`Microsoft.Azure.WebJobs.Script` and `Microsoft.Azure.WebJobs.Script.Grpc` temporarily grant Server internal access. This avoids promoting broad implementation details to public API or introducing a large host-adapter abstraction during the move. Existing test-equivalent friend access is mirrored for moved internal types.

### Build and runtime wiring includes the new assembly

- WebHost directly references Server and Grpc.
- Server is added to the solution and official package build.
- Runtime assembly policies, generated runtime metadata, and the WebHost deps baseline include `Azure.Functions.Rpc.Server`.
- Package direction is `Server -> Grpc`, with no `Grpc -> Server` dependency.

## Why `Azure.Functions.Rpc.Server`?

`Microsoft.Azure.WebJobs.*` is an existing legacy product identity. New assemblies should not introduce additional WebJobs-named components when the functionality is part of the Azure Functions runtime rather than the WebJobs programming model.

`Azure.Functions.Rpc.Server`:

- Describes the assembly by its topology role: the inbound server and local worker-process implementation.
- Establishes a consistent future family with `Azure.Functions.Rpc.Client`.
- Avoids expanding the legacy WebJobs assembly/package naming surface.
- Makes the Server/Client dependency boundary visible at the assembly level.

Moved classes intentionally retain their existing `Microsoft.Azure.WebJobs.Script.*` namespaces for now. Renaming them in the same PR would obscure the assembly-boundary review, reduce rename detection, and create unnecessary source churn. The new package, assembly, and root namespace use `Azure.Functions.Rpc.Server`; class namespace alignment can be handled separately if desired.

## Scope and behavior

This PR does **not** add Client, outbound transport, composition, channel handles, type forwarding, or a new Host/Worker product. The protobuf contract stays in Grpc, and standard WebHost remains the only executable topology.

No worker startup, invocation, shutdown, or disposal behavior is intended to change.

### Pull request checklist


src/WebJobs.Script.Grpc/IWorkerChannel.cs
```diff
@@ -10,10 +9,6 @@ internal interface IWorkerChannel
     {
         string Id { get; }
 
-        IWorkerProcess WorkerProcess { get; }
```
brettsam, line 13: These move to new `IWorkerProcessChannel` so that process management isn't a core part of channels anymore. That interface is used where needed in the Server proj.

src/WebJobs.Script/Properties/AssemblyInfo.cs
```diff
@@ -12,6 +12,7 @@
 [assembly: Guid("1dc670cd-f42f-4d8f-97bd-0e1aa8221094")]
 
 [assembly: InternalsVisibleTo("Microsoft.Azure.WebJobs.Script.Benchmarks")]
+[assembly: InternalsVisibleTo("Azure.Functions.Rpc.Server")]
```
brettsam, line 15: I wanted to avoid this but unfortunately there's so much still internal that would require a disruptive refactor to properly expose. We can revisit, but for the purposes of moving this forward I'm continuing to share (this stuff was already InternalsVisibleTo with the Grpc assembly so nothing new is being opened up here).

src/Functions.Rpc.Server/IWorkerProcessChannel.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.Script.Workers
```
kshyju, line 7: Should we consider removing webjobs from the namespace (based on what you shared about the context of how the new project was named)  and potentially use the Root namespace for new types?

src/Functions.Rpc.Server/Functions.Rpc.Server.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net10.0</TargetFramework>
+    <PackageId>Azure.Functions.Rpc.Server</PackageId>
+    <AssemblyName>Azure.Functions.Rpc.Server</AssemblyName>
+    <RootNamespace>Azure.Functions.Rpc.Server</RootNamespace>
```
kshyju, line 7: Do we need to use `Microsoft.` prefix in front of these? or that was not used intentionally?

src/Functions.Rpc.Server/Functions.Rpc.Server.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net10.0</TargetFramework>
+    <PackageId>Azure.Functions.Rpc.Server</PackageId>
```
kshyju, line 5: Do we need the PackageId? Are you planning to publish this package?

kshyju (approved):

src/Functions.Rpc.Server/IWorkerProcessChannel.cs
brettsam: It's worth discussing and I think we can follow-up with that if we decide it's worth it. But that would have caused a ripple effect of a ton more changes here that I thought made the review more difficult, so I decided to leave it out for now, even though the assembly itself doesn't match anymore.

src/Functions.Rpc.Server/Functions.Rpc.Server.csproj
brettsam: Both @fabiocav and @jviau preferred leaving off `Microsoft`, so I went with that :-)

src/Functions.Rpc.Server/Functions.Rpc.Server.csproj
brettsam: No, this won't be published. But I noticed that PackageId is set on all our csproj's (whether we publish them or not), so it was attempting to match those.

jviau: We do actually publish our other projects for core-tools to consume. That is why they have PackageId.

brettsam: Ah! I thought they were all bundled inside the WebHost package. Good to know.

src/Functions.Rpc.Server/Functions.Rpc.Server.csproj
```diff
@@ -0,0 +1,21 @@
...
+    <AssemblyName>Azure.Functions.Rpc.Server</AssemblyName>
```
jviau, line 6: I think you only need `AssemblyName`. `PackageId` and `RootNamespace` should default to `AssemblyName` when unset.

jviau (approved):

src/Functions.Rpc.Server/Functions.Rpc.Server.csproj
brettsam: Ah ok. I think other csproj's explicitly listed them so I'll keep this as-is for now. We can come back and clean up if desired later.

PR #11919, opened 2026-08-11
fixing host disposal test

brettsam:
Fixing two flaky tests.

### DisposesScriptHost
Did some investigation and the issue was that it was verifying that `RestartAsync()` would dispose of the current host instance. However, it was mocking only a single host. That meant that it was possible for the disposal to occur before a new host started. That could result in an exception if anything tried to access the host as the mocked ServiceProvider was disposed.

This changes it to use two mocked hosts and an `OrderedScriptHostBuilder` (something that already existed) as well as some coordination with a Semaphore and a Task. This controls the ordering and ensures that we correctly replace the host with a valid one as well as verify the original host was disposed.

### LogsAreFlushedOnTimer
Should be a simple timing isisue. This one was hard-coded to wait 100ms. Changed to use `TestHelper.Await()`

### Pull request checklist

### DIFF
--- a/test/WebJobs.Script.Tests/Diagnostics/FileWriterTests.cs
+++ b/test/WebJobs.Script.Tests/Diagnostics/FileWriterTests.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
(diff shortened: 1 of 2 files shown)


test/WebJobs.Script.Tests/WebJobsScriptHostServiceTests.cs
```diff
@@ -405,59 +405,70 @@ public async Task DisposedHost_ServicesNotExposed()
         public async Task DisposesScriptHost()
         {
             var metricsLogger = new TestMetricsLogger();
+            using var disposedSemaphore = new SemaphoreSlim(0, 1);
```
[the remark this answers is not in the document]
brettsam: throwing an exception on a failure path is fine for this test -- we'll leave this as-is.

test/WebJobs.Script.Tests/Diagnostics/FileWriterTests.cs
```diff
@@ -195,15 +195,21 @@ public async Task LogsAreFlushedOnTimer()
...
+
+                string logFile = directory.EnumerateFiles().First().FullName;
+                string[] fileLines = File.ReadAllLines(logFile);
+                return fileLines.Length;
+            }
 
-            string logFile = directory.EnumerateFiles().First().FullName;
-            string[] fileLines = File.ReadAllLines(logFile);
-            Assert.Equal(numLogs, fileLines.Length);
+            await TestHelpers.Await(() => GetLogLineCount() == numLogs, timeout: 5000);
```
jviau, line 212: I wonder if we should use `>=`? Do we expect there to be more logs than we assert and we are trying to just gate it when we reach that threshold?

I worry that using `==` will risk the number of logs going past the expected value between checks.

brettsam: Since this is testing the filewriter -- we're only writing 100 lines so we expect that exact number. Even seeing more should be an error.  The sporadic failures I saw was when we'd check before the final flush had occurred so it'd be like 99 or something.

jviau (approved):

satvu (approved):

PR #11921, opened 2026-08-12
introducing WorkerProxy project for compute separation

brettsam:
The WorkerProxy is a lightweight sidecar that provides a stable boundary between the Functions runtime and language worker. It allows us to evolve how the host connects to and manages workers while preserving the existing worker protocol, so existing language workers can participate without topology-specific changes. This only comes into play for "separated compute" scenarios in Flex consumption where language workers run on separate instances from the runtime.

This PR adds the initial project scaffolding, Native AOT container, readiness endpoint, tests, and CI smoke coverage. Relay, lifecycle, HTTP forwarding, and authentication behavior will follow in subsequent changes.

### Pull request checklist


Azure.Functions.Host.sln
```diff
@@ -0,0 +1,56 @@
+﻿
```
brettsam: Note this is a new solution that will eventually contain all the projects and tests for the "compute separation" model of running Functions. We're going to name it Azure.Functions.Host as it's going to be the "future" way we run things and we didn't want to give it some one-off name.

Right now it has nothing but WorkerProxy in it, but we will add to this in future commits as we wire all the incoming projects together.

test/Functions.WorkerProxy.Tests/Configuration/WorkerProxyOptionsTests.cs
```diff
@@ -0,0 +1,133 @@
...
+    [Theory]
+    [InlineData("invalid")]
+    [InlineData(" ")]
+    public void Options_Throw_WhenValueIsNotAnInteger(string value)
+    {
+        using EnvironmentVariableScope managementPort =
+            new(WorkerProxyOptions.ManagementPortConfigurationKey, value: null);
+
+        Assert.Throws<InvalidOperationException>(
+            () => WorkerProxyApplication.Build(["--management-port", value]));
```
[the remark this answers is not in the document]
brettsam: `Build()` intentionally exercises the application composition boundary. Kestrel options depend on `IOptions<WorkerProxyOptions>` and access `.Value` while the server is constructed, so binding and validation are deterministically triggered here; `ValidateOnStart()` also preserves startup validation.

src/Functions.WorkerProxy/Program.cs
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Threading.Tasks;
```
kshyju: nit: remove unneeded using statement.

test/Functions.WorkerProxy.Tests/EnvironmentVariableScope.cs
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    public void Dispose()
+    {
+        Environment.SetEnvironmentVariable(_name, _previousValue);
+    }
+}
+
+[CollectionDefinition(nameof(EnvironmentVariableCollection), DisableParallelization = true)]
+public sealed class EnvironmentVariableCollection
```
kshyju: nit. Empty class definition can be simplified by adding ; at the end.
 `public sealed class EnvironmentVariableCollection;`

Azure.Functions.Host.sln
```diff
@@ -0,0 +1,56 @@
...
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 17
+VisualStudioVersion = 17.0.31903.59
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{827E0CD3-B72D-47B6-A68D-7590B98EB39B}"
```
kshyju: Should we use the new .slnx format? That will keep this smaller and more readable.
See example in the MAF Durable extension repo - https://github.com/microsoft/agent-framework-durable-extension/blob/main/dotnet/agent-framework-durable-extension.slnx

eng/ci/templates/jobs/run-worker-proxy-tests.yml
```diff
@@ -0,0 +1,87 @@
...
+    imageContext: $(Agent.TempDirectory)/workerproxy-image
+    localImage: functions-worker-proxy-ci:latest
+    publishDirectory: $(Agent.TempDirectory)/workerproxy-image/.workerproxy-publish
+    # This solution will expand as the remaining compute-separation projects are introduced.
+    solution: Azure.Functions.Host.sln
+
+  steps:
+  - template: /eng/ci/templates/install-dotnet.yml@self
+
+  - task: DotNetCoreCLI@2
```
kshyju, line 28: Should we consider enabling `dotnet format` on build. I like how it keeps the code (format/style) consistent with .editorconfig. 

EX: https://github.com/microsoft/agent-framework-durable-extension/blob/ad941eff53617840c0a046498be36d0b3871329b/dotnet/Directory.Build.targets#L9

kshyju (approved):

eng/ci/templates/jobs/run-worker-proxy-tests.yml
brettsam: we can do that in a follow-up -- i like the idea

src/Functions.WorkerProxy/WorkerProxyApplication.cs
```diff
@@ -0,0 +1,98 @@
...
+        });
+    }
+
+    internal static WebApplication Build(Action<ConfigurationManager> configureExternalConfiguration)
+    {
+        ArgumentNullException.ThrowIfNull(configureExternalConfiguration);
+
+        WebApplicationBuilder builder = WebApplication.CreateSlimBuilder();
+
+        builder.Configuration.Sources.Clear();
```
jviau: Curious why the stripping of config and adding our own? We don't want `ASPNETCORE_` and `DOTNET_` env prefixes?

I wonder how this will work at runtime, because the `CreateSlimBuilder` is going to bootstrap that config, potentially consume some of it immediately, then return and we remove this config. Will we be in a world where we consumed config then removed it later?

.gitignore
```diff
@@ -35,6 +35,7 @@ bld/
 [Ll]og/
 [Ll]ogs/
 [Oo]ut/
+.workerproxy-publish/
```
jviau: Publish output should go to `/out/pub/` by default. What is causing publish output needing a new folder to be excluded?

jviau:
dotnet sdk has support for containerization via `dotnet publish`. https://learn.microsoft.com/en-us/dotnet/core/containers/sdk-publish. 1ES doesn't support this as far as I am aware, but maybe we can still use it for local development?

eng/ci/templates/jobs/run-worker-proxy-tests.yml
```diff
@@ -0,0 +1,87 @@
...
+
+  pool:
+    name: ${{ parameters.linuxPoolName }}
+    image: ${{ parameters.linuxPoolImage }}
+    os: linux
+
+  variables:
+    imageContext: $(Agent.TempDirectory)/workerproxy-image
+    localImage: functions-worker-proxy-ci:latest
+    publishDirectory: $(Agent.TempDirectory)/workerproxy-image/.workerproxy-publish
```
jviau: Consider using `$(Build.ArtifactStagingDirectory)`

src/Functions.WorkerProxy/WorkerProxyApplication.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        builder.Services
+            .AddOptions<KestrelServerOptions>()
+            .Configure<IOptions<WorkerProxyOptions>, IConfiguration>(
+                (kestrelOptions, workerProxyOptions, configuration) =>
+            {
+                kestrelOptions.Configure(
+                    configuration.GetSection(KestrelConfigurationSection),
+                    reloadOnChange: false);
+                kestrelOptions.ListenAnyIP(workerProxyOptions.Value.ManagementPort, listenOptions =>
```
jviau: Is there an existing aspnetcore parameter we can use?

.gitignore
brettsam: i think this was a remnant of a previous change -- removing it

brettsam: actually i was wrong. the ci and dockerfile chose this custom path for some reason. i've reverted it to use the standard output now

src/Functions.WorkerProxy/WorkerProxyApplication.cs
brettsam: good point. i think that was inadvertently done for testing purposes -- revisiting this now

brettsam: Dramatically simplified things. No more Options, just using built-in env vars for this first commit. Using WebApplicationFactory<> for tests.

src/Functions.WorkerProxy/WorkerProxyApplication.cs
brettsam: yeah, adjusted a bunch of stuff to simplify this now.

brettsam:
> dotnet sdk has support for containerization via `dotnet publish`. https://learn.microsoft.com/en-us/dotnet/core/containers/sdk-publish. 1ES doesn't support this as far as I am aware, but maybe we can still use it for local development?

Yeah, we can address that when we add the local dev scenarios w/ aspire, etc. This is just a preliminary checkin to get the structure here and some (very) basic validation.

jviau (approved):

PR #11927, opened 2026-08-17
feat(health): scope storage check to script host

### LINKED ISSUE #11636: WebJobsStorageHealthCheck causing warnings (unhealthy container) when a AzureWebJobsStorage is not configured
I have a  single Azure Function running in a container in AKS. The function is a Service Bus Trigger and doesn't require any connection to an Azure Storage account. Today I needed to rebuild my images, which probably has a newer version of the Azure Function Host, because suddenly I get health check errors like:

> [Tag=''] Process reporting unhealthy: Unhealthy. Health check entries are {"azure.functions.web_host.lifecycle":{"status":"Healthy","description":null},"azure.functions.script_host.lifecycle":{"status":"Healthy","description":null},"azure.functions.webjobs.storage":{"status":"Unhealthy","description":"Unable to create client for AzureWebJobsStorage"}}

The issue is cause by the the new health checks in the Azure Function Host. More specific in [WebJobsStorageHealthCheck ](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Diagnostics/HealthChecks/WebJobsStorageHealthCheck.cs). 

In my case I don't have AzureWebJobsStorage configured, because it's not a durable function and I don't need any Azure Storage Account. 

C# Binding code.
```csharp
public async Task RunAsync(
    [ServiceBusTrigger(ServiceBusTopicNames.Events, ServiceBusSubscriptionNames.Audit, Connection = "AttestServiceBusConnectionString")] string sbMessage,
    string label,
    string correlationId)
```

Docker image being used: mcr.microsoft.com/azure-functions/dotnet-isolated:4-dotnet-isolated8.0

#### Repro steps

Just create a Azure Function using the ServiceBusTrigger attribute, and only configure the required settings for the ServiceBusTrigger (so don't configure AzureWebJobsStorage)

jviau:

resolves #11636

### Pull request checklist

    * Otherwise: Backport PR to follow

### Additional information

Moves `WebJobsStorageHealthCheck` into the ScriptHost service scope. Placeholder hosts skip JobHost health-check registration entirely. For active hosts, storage registration is deferred through `HealthCheckServiceOptions` so it evaluates the final DI-provided `IConfiguration` and registers only when `AzureWebJobsStorage` has a non-empty scalar or hierarchical configuration value.

The health check no longer depends on WebHost-only state, supports both synchronous and asynchronous disposal to match ScriptHost lifetime management, and has end-to-end coverage for configured storage and placeholder hosts.


src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -395,6 +396,13 @@ public static IHostBuilder AddScriptHostCore(this IHostBuilder builder, ScriptAp
                 services.AddSingleton<IHostOptionsProvider, HostOptionsProvider>();
                 services.AddSingleton<IInstanceServicesProviderFactory, ScriptInstanceServicesProviderFactory>();
                 services.AddSingleton<IWorkerRuntimeResolver, ScriptHostWorkerRuntimeResolver>();
+
+                IEnvironment environment =
+                    applicationHostOptions.RootServiceProvider.GetRequiredService<IEnvironment>();
+                if (!environment.IsPlaceholderModeEnabled())
+                {
+                    services.AddHealthChecks().AddJobHostScopedHealthChecks();
```
brettsam, line 402: my review agent pointed out that now AddHealthChecks() is being called at both WebHost and JobHost levels... which could mean that IHealthCheckPublisher from the root would flow to the child, and result in duplicate telemetry.

it suggests adding `IHealthCheckPublisher` to `ChildContainerIgnoredTypes`

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
```diff
@@ -395,6 +396,13 @@ public static IHostBuilder AddScriptHostCore(this IHostBuilder builder, ScriptAp
                 services.AddSingleton<IHostOptionsProvider, HostOptionsProvider>();
                 services.AddSingleton<IInstanceServicesProviderFactory, ScriptInstanceServicesProviderFactory>();
                 services.AddSingleton<IWorkerRuntimeResolver, ScriptHostWorkerRuntimeResolver>();
+
+                IEnvironment environment =
+                    applicationHostOptions.RootServiceProvider.GetRequiredService<IEnvironment>();
+                if (!environment.IsPlaceholderModeEnabled())
```
brettsam: instead of checking this on environment (which is global and can change mid-specialization), use `applicationHostOptions.IsStandbyConfiguration`. It's rare, but we've seen races with this elsewhere.

brettsam (changes requested):

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
jviau: SO much better. Why did I not notice that was available.

src/WebJobs.Script/ScriptHostBuilderExtensions.cs
jviau: Good catch! Updating to prevent that.

PR #11940, opened 2026-08-24
initial Rpc.Client project

brettsam:
Introduces the Azure.Functions.Rpc.Client project, which will be the home to the compute separation logic (channel, channel registry, grpc client logic, etc). This commit adds the new project, new tests, and introduces the RpcClientConnection, which is responsible for connecting to the worker proxy when given a Uri by the platform. This will be used by the new client worker channel in the next PR.

### Pull request checklist


Azure.Functions.Host.slnx
```diff
@@ -5,9 +5,15 @@
     <Platform Name="x86" />
   </Configurations>
   <Folder Name="/src/">
+    <Project Path="src/Functions.Rpc.Client/Functions.Rpc.Client.csproj" />
+    <Project Path="src/Functions.Rpc.Server/Functions.Rpc.Server.csproj" />
     <Project Path="src/Functions.WorkerProxy/Functions.WorkerProxy.csproj" />
+    <Project Path="src/WebJobs.Script/WebJobs.Script.csproj" />
+    <Project Path="src/WebJobs.Script.Grpc/WebJobs.Script.Grpc.csproj" />
+    <Project Path="src/WebJobs.Script.WebHost/WebJobs.Script.WebHost.csproj" />
```
brettsam, line 13: Note that none of this is usable yet as we have no DI wire-up available until a later milestone. But this is getting the various projects into place.

src/Functions.Rpc.Client/RpcClientConnection.cs
jviau, line 1: This file is fairly large, can we break it up?

1. Can use partial classes to move things like `TerminalState` to their own file.
2. Maybe we can create a general `Channel<TIn, TOut>` implementation to wrap a `AsyncDuplexStreamingCall<TIn, TOut>`, and extract that to a separate class. With this separation then tests and everything can work off of a `Channel` - no need to create mock gRPC streams, beyond testing the wrapper itself.

test/Functions.Rpc.Client.Tests/ControlledDuplexStream.cs
```diff
@@ -0,0 +1,154 @@
...
+using System.Collections.Concurrent;
+using System.Runtime.ExceptionServices;
+using System.Threading;
+using System.Threading.Channels;
+using System.Threading.Tasks;
+using Grpc.Core;
+
+namespace Azure.Functions.Rpc.Client.Tests;
+
+internal sealed class ControlledDuplexStream<T>
```
jviau: Can you add a description to this class explaining its purpose?

src/Functions.Rpc.Client/IDuplexCall.cs
```diff
@@ -0,0 +1,38 @@
...
+/// Represents the request and response streams of one duplex call.
+/// </summary>
+/// <remarks>
+/// <see cref="IAsyncDisposable.DisposeAsync"/> aborts the call and must unblock concurrent
+/// <see cref="WriteAsync"/> and <see cref="ReadAllAsync"/> operations. Graceful request-stream half-close is not part of
+/// this raw transport contract.
+/// </remarks>
+/// <typeparam name="TRequest">The request type.</typeparam>
+/// <typeparam name="TResponse">The response type.</typeparam>
+internal interface IDuplexCall<in TRequest, out TResponse> : IAsyncDisposable
```
jviau: oh, I was meant to inherit `Channel` itself and not our own abstraction.

``` csharp
using Grpc.Core;
using System.Threading.Channels;

public sealed class GrpcDuplexChannel<T> : Channel<T>, IAsyncDisposable
{
    private readonly AsyncDuplexStreamingCall<T, T> _call;
    private readonly Channel<T> _incoming;
    private readonly Channel<T> _outgoing;
    private readonly CancellationTokenSource _shutdown = new();
    private readonly Task _readPump;
    private readonly Task _writePump;
    private int _disposed;

    public GrpcDuplexChannel(AsyncDuplexStreamingCall<T, T> call)
    {
        _call = call ?? throw new ArgumentNullException(nameof(call));

        _incoming = System.Threading.Channels.Channel.CreateUnbounded<T>(
            new UnboundedChannelOptions
            {
                SingleReader = false,
                SingleWriter = true
            });

        _outgoing = System.Threading.Channels.Channel.CreateUnbounded<T>(
            new UnboundedChannelOptions
            {
                SingleReader = true,
                SingleWriter = false
            });

        _readPump = PumpResponsesAsync();
        _writePump = PumpRequestsAsync();
    }

    public override ChannelReader<T> Reader => _incoming.Reader;

    public override ChannelWriter<T> Writer => _outgoing.Writer;

    private async Task PumpResponsesAsync()
    {
        Exception? error = null;

        try
        {
            while (await _call.ResponseStream.MoveNext(_shutdown.Token) .ConfigureAwait(false))
            {
                await _incoming.Writer.WriteAsync(_call.ResponseStream.Current, _shutdown.Token).ConfigureAwait(false);
            }
        }
        catch (OperationCanceledException) when (_shutdown.IsCancellationRequested)
        {
        }
        catch (Exception exception)
        {
            error = exception;
        }
        finally
        {
            _incoming.Writer.TryComplete(error);
            _outgoing.Writer.TryComplete(error);

            if (error is not null)
            {
                _shutdown.Cancel();
            }
        }
    }

    private async Task PumpRequestsAsync()
    {
        Exception? error = null;

        try
        {
            await foreach (T item in _outgoing.Reader.ReadAllAsync(_shutdown.Token))
            {
                await _call.RequestStream.WriteAsync(item).ConfigureAwait(false);
            }

            await _call.RequestStream.CompleteAsync().ConfigureAwait(false);
        }
        catch (OperationCanceledException) when (_shutdown.IsCancellationRequested)
        {
        }
        catch (Exception exception)
        {
            error = exception;
        }
        finally
        {
            if (error is not null)
            {
                _incoming.Writer.TryComplete(error);
                _outgoing.Writer.TryComplete(error);
                _shutdown.Cancel();
            }
        }
    }

    public async ValueTask DisposeAsync()
    {
        if (Interlocked.Exchange(ref _disposed, 1) != 0)
        {
            return;
        }

        _outgoing.Writer.TryComplete();
        _shutdown.Cancel();
        _call.Dispose();

        await Task.WhenAll(_readPump, _writePump).ConfigureAwait(false);
        _shutdown.Dispose();
    }
}
```

brettsam: omg i totally misread your initial proposal. yeah this makes sense; refactoring now. It's not really that different, i kinda just invented my own new interface with slightly different behavior :-)

src/Functions.Rpc.Client/RpcClientConnection.cs
```diff
@@ -20,7 +20,7 @@ namespace Azure.Functions.Rpc.Client;
 internal sealed partial class RpcClientConnection : IAsyncDisposable
 {
     private readonly Task _completion;
-    private readonly IDuplexCall<StreamingMessage, StreamingMessage> _duplexCall;
+    private readonly Channel<StreamingMessage> _duplexChannel;
     private readonly Channel<StreamingMessage> _inbound;
```
jviau: Do we need the `_inbound` and `_outbound` in this type anymore? The `_duplexChannel` already has the channels to buffer inbound and outbound messages.

If you need to tweak some stuff like SingleReader, add an options object for the `GrpcDuplexChannel` to configure that at construction.

src/Functions.Rpc.Client/FunctionRpcDuplexChannelFactory.cs
jviau: Should the factory be the constructing and owning the channel? For gRPC it is common to keep the channel lifetime longer than any individual call, and re-use it between calls to the same host.

src/Functions.Rpc.Client/FunctionRpcDuplexStreamFactory.cs
```diff
@@ -77,17 +77,17 @@ public async Task<IDuplexCall<StreamingMessage, StreamingMessage>> ConnectAsync(
             var grpcCall = client.EventStream(cancellationToken: callLifetimeSource.Token);
```
jviau: Why do we not use the passed in `cancellationToken`?

src/Functions.Rpc.Client/FunctionRpcDuplexChannelFactory.cs
jviau: Is `ConnectAsync` needed? Or will establishing the stream below initiate the connection itself? Or is the connection only established when the first message is sent on the stream?

src/Functions.Rpc.Client/FunctionRpcDuplexChannelFactory.cs
```diff
@@ -12,9 +13,9 @@
 namespace Azure.Functions.Rpc.Client;
 
 /// <summary>
-/// Creates outbound FunctionRpc duplex calls.
+/// Creates outbound FunctionRpc duplex channels.
 /// </summary>
-internal sealed class FunctionRpcDuplexCallFactory : IDuplexCallFactory<StreamingMessage, StreamingMessage>
+internal sealed class FunctionRpcDuplexChannelFactory : IDuplexChannelFactory<StreamingMessage>
```
jviau, line 16: I would prefer we have a factory for producing the `FunctionRpc.FunctionRpcClient`, and then callers can open the event stream themselves. We can then add an extension method for wrapping as a channel. End use would be something like:

``` CSharp
IRpcClientFactory factory;

FunctionRpcClient client = factory.Create(<uri>); // factory will create and cache `GrpcChannel`, so they are re-used between `Create` calls for the same URI

Channel<StreamingMessage> channel = client.EventStream(cancellation).AsChannel(new GrpcChannelOptions()); // Fill out GrpcChannelOptions as you need.
```

This does affect some of the cleanup on failure to connect. Not sure how crucial that all is though. Could make it a `ValueTask<FunctionRpcClient> CreateAsync(..)` signature, so the first creation performs a `ConnectAsync` on the channel to verify we can connect before caching it.

src/Functions.Rpc.Client/FunctionRpcDuplexChannelFactory.cs
jviau: Do we want a separate handler per connection? I thought you wanted to re-use handlers in .NET

src/Functions.Rpc.Client/GrpcDuplexChannel.cs
```diff
@@ -0,0 +1,218 @@
...
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="GrpcDuplexChannel{T}"/> class.
+    /// </summary>
+    /// <param name="call">The SDK duplex call.</param>
+    /// <param name="callLifetimeSource">The cancellation source used to create <paramref name="call"/>.</param>
+    /// <param name="ownedResource">The underlying connection resource.</param>
+    /// <param name="logger">The logger used for secondary cleanup failures.</param>
+    internal GrpcDuplexChannel(AsyncDuplexStreamingCall<T, T> call, CancellationTokenSource callLifetimeSource,
+        IDisposable ownedResource, ILogger logger)
```
jviau: Is the logging valuable here? Can we rely on the owner of this channel to log thrown exceptions (if they care).

src/Functions.Rpc.Client/RpcClientConnection.cs
jviau, line 1: How much value does this abstraction add over `Channel<T>`? If this is to combine worker ID and maybe some reconnect on fault semantics, that might be useful. That would entail it holding the RPC client directly (which honestly seems fine to me).

But also if there are benefits to this abstraction that will be clear in later PRs that is fine with me.

src/Functions.Rpc.Client/RpcClientConnection.cs
brettsam: Nope. No idea how I missed this being there still.

src/Functions.Rpc.Client/FunctionRpcDuplexChannelFactory.cs
brettsam: `EventStream()` returns an `AsyncDuplexStreamingCall` immediately. It does not give us an awaitable indication that DNS, TCP, TLS, and HTTP/2 setup succeeded; without `channel.ConnectAsync`, the factory could return a seemingly valid connection and only surface an unreachable endpoint from the first read/write pump.

Calling `ConnectAsync(cancellationToken)` ensures:
- the endpoint reaches gRPC `Ready` before the factory returns;
- the caller’s short-lived token bounds initial network establishment;
- connection refusal/cancellation does not return a partially initialized `RpcClientConnection`;
- the separate `callLifetimeSource` can then own the long-lived `EventStream` independently.

src/Functions.Rpc.Client/FunctionRpcDuplexStreamFactory.cs
brettsam: that's for the short call to ConnectAsync() -- this token is for us to cancel the long-lived duplex connection

src/Functions.Rpc.Client/FunctionRpcDuplexChannelFactory.cs
brettsam: are you suggesting one (global) handler that's shared for every connection to every endpoint? i actually wasn't aware that this would be beneficial... we certainly can come back to this, but I'll leave it like this for now...

agent's answer was:

A shared `SocketsHttpHandler` across different endpoints is valid:
- It maintains separate connection pools per endpoint/authority.
- `GrpcChannel`s must use `DisposeHttpClient = false`.
- A DI singleton—not a static—owns and disposes the handler at app shutdown.

However, different worker endpoints cannot share TCP/HTTP2 connections. Sharing the handler only saves handler infrastructure; it does not provide meaningful connection reuse across endpoints.

src/Functions.Rpc.Client/FunctionRpcDuplexChannelFactory.cs
brettsam: My goal was to make the Channel<StreamingMessage> the test seam, hence the ChannelFactory being the replace-able class. We have a TestDuplexChannelFactory and now we can write our tests w/o needing any grpc stuff at all.

I'm not sure that we have a lot of value in caching GrpcChannel in our scenario -- we'll most often have one long-lived connection to a single endpoint, used by a single Client. I don't expect us to have multiple clients created against that one uri -- are you thinking we might? If the platform asks us to connect to the same uri again, I'd imagine we'd return something like a 409 back?

For now I'm going to skip this, but we can certainly adapt in a follow-up.

src/Functions.Rpc.Client/RpcClientConnection.cs
brettsam: Yeah, none, really. I'm removing it and the Channel will be our primary usage...

brettsam: That being said, there are a couple more layers here for full wire-up, so once we see those we may revisit... but I think I have it all laid out now where this Connection thing is unnecessary.

src/Functions.Rpc.Client/GrpcDuplexStream.cs
```diff
@@ -0,0 +1,220 @@
...
+/// Adapts an SDK <see cref="AsyncDuplexStreamingCall{TRequest, TResponse}"/> to a bidirectional
+/// <see cref="Channel{T}"/>.
+/// </summary>
+/// <remarks>
+/// Writes complete when a request is admitted to the outgoing queue, not when it reaches the peer.
+/// Completing <see cref="Channel{T}.Writer"/> gracefully completes the gRPC request stream after queued requests drain.
+/// The stream supports one response reader and multiple concurrent request writers.
+/// </remarks>
+/// <typeparam name="T">The message type used in both directions.</typeparam>
+internal sealed class GrpcDuplexStream<T> : Channel<T>, IAsyncDisposable
```
brettsam: Note that I'd originally had this called `GrpcDuplexChannel<T>` b/c it's a Channel... but we have too many dang things called "channel". And since it's meant to wrap an `AsyncDuplexStreamingCall` in prod, I chose "stream".

jviau (approved):
I have a preference for naming it `GrpcDuplexChannel` over `Stream`. I know a lot of stuff is named channel, but this inherits `Channel`, which is very applicable to name it as such. Channel vs Stream also has different connotations:

Channel is bidirectional discrete payloads.
Stream is typically one direction and is raw data that becomes a single payload.

jviau (approved):

src/Functions.Rpc.Client/RpcClientFactory.cs
```diff
@@ -0,0 +1,357 @@
...
+            }
+
+            TryCleanup(() => candidate?.Dispose(), endpoint, "dispose the unused gRPC channel");
+
+            // Disposal waits for all cold-path operations to leave this block before releasing the connection gates and
+            // shutdown source. The final operation completes that shared wait while holding the count's lock.
+            lock (_syncLock)
+            {
+                _activeOperations--;
+                if (Interlocked.CompareExchange(ref _disposeState, 0, 0) != 0 && _activeOperations == 0)
```
kshyju: GHCP flagged:

Nit: `_disposeState` is only ever touched under `_syncLock` — the write in `DisposeAsync` and all three read sites — so the `Interlocked.CompareExchange(ref _disposeState, 0, 0)` reads are redundant. `Monitor` already provides the barrier.

src/Functions.Rpc.Client/RpcClientFactory.cs
```diff
@@ -0,0 +1,357 @@
...
+{
+    /// <summary>
+    /// Gets the maximum time allowed for an individual socket connection attempt.
+    /// </summary>
+    internal static readonly TimeSpan DefaultConnectTimeout = TimeSpan.FromSeconds(5);
+
+    /// <summary>
+    /// Gets the idle interval before an HTTP/2 keepalive ping is sent.
+    /// </summary>
+    internal static readonly TimeSpan DefaultKeepAlivePingDelay = TimeSpan.FromSeconds(30);
```
kshyju: Do we need these as `internal`? `private` good?

src/Functions.Rpc.Client/FunctionRpcDuplexChannelFactory.cs
```diff
@@ -0,0 +1,62 @@
...
+        _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+    }
+
+    /// <inheritdoc />
+    public async Task<Channel<StreamingMessage>> ConnectAsync(Uri endpoint, CancellationToken cancellationToken = default)
+    {
+        FunctionRpc.FunctionRpcClient client = await _clientFactory.CreateAsync(endpoint, cancellationToken);
+        cancellationToken.ThrowIfCancellationRequested();
+
+        var call = client.EventStream();
```
kshyju, line 39: Should we pass `cancellationToken` to the `EventStream` method?

src/Functions.Rpc.Client/GrpcDuplexChannel.cs
```diff
@@ -0,0 +1,191 @@
...
+            error = exception;
+        }
+        finally
+        {
+            _outgoing.Writer.TryComplete(error);
+            _incoming.Writer.TryComplete(error);
+
+            if (error is not null)
+            {
+                _shutdownSource.Cancel();
```
kshyju: Cancel has an async version.

kshyju (approved):
Should we use `System.Threading.Lock` instead of `object` for the different lock uses cases? That seems like the modern API for this use case.

src/Functions.Rpc.Client/RpcClientFactory.cs
brettsam: ah yeah -- refactored this and missed it. Fixed.

src/Functions.Rpc.Client/FunctionRpcDuplexChannelFactory.cs
brettsam: added a short comment why we don't pass it there -- here's explanation:

For the `EventStream(cancellationToken)` comment, we should **not** pass the existing token. It only bounds `CreateAsync`/channel readiness; passing it to `EventStream` would let later cancellation unexpectedly terminate the long-lived call. `GrpcDuplexChannel.DisposeAsync()` owns call cancellation through `AsyncDuplexStreamingCall.Dispose()`.

kshyju (approved):

PR #11603, opened 2026-02-11
Suppress noise from telemetry

RohitRanjanMS:
### Issue describing the changes in this PR

resolves #11594 
### Pull request checklist

**IMPORTANT**: Currently, changes must be backported to the `in-proc` branch to be included in Core Tools and non-Flex deployments.

* [x] Backporting to the `in-proc` branch is not required
    * Otherwise: Link to backporting PR
* [x] My changes **do not** require documentation changes
    * [ ] Otherwise: Documentation issue linked to PR
* [ ] My changes **should not** be added to the release notes for the next release
    * [x] Otherwise: I've added my notes to `release_notes.md`
* [x] My changes **do not** need to be backported to a previous version
    * [ ] Otherwise: Backport tracked by issue/PR #issue_or_pr
* [x] My changes **do not** require diagnostic events changes
    * Otherwise: I have added/updated all related diagnostic events and their documentation (Documentation issue linked to PR)
* [x] I have added all required tests (Unit tests, E2E tests)

### Additional information

Additional PR information

PR #11947, opened 2026-08-27
Suppress polling logs by event name

RohitRanjanMS:

Follow-up to #11603.

Moves FunctionsLogs noise suppression into `SystemLogger`, where category, level, and EventName are available together. The standard `ILoggingBuilder.AddFilter` callback only receives category and level, so it cannot distinguish individual events — which is why the previous approach had to drop every Debug/Trace log in a category.

Suppression is driven by a single declarative table. An empty event list suppresses the whole category; otherwise only the listed EventNames are suppressed:

| Category | Suppressed events |
| --- | --- |
| `Host.Executor` | all (unchanged from #11603) |
| `Microsoft.Azure.WebJobs.EventHubs.EventHubProducerClientImpl` | all (unchanged from #11603) |
| `Microsoft.Azure.WebJobs.Extensions.Storage.Common.Listeners.QueueListener` | `GetMessages`, `BackoffDelay` |
| `Microsoft.Azure.WebJobs.Host.Queues.Listeners.QueueListener` | `GetMessages`, `BackoffDelay` |
| `Microsoft.Azure.WebJobs.Extensions.Storage.Blobs.Listeners.BlobListener` | `PollBlobContainer`, `BlobAlreadyProcessed`, `BlobDoesNotMatchPattern` |

Adding a category or event requires only a new line in that table.

Compared to the original approach, this preserves queue/blob listener start/stop lifecycle logs, `HandlingStorageException`, `BlobMessageEnqueued`, `ContainerDoesNotExist`, `FunctionNotFound`, `BlobHasNoETag`, and the other low-volume diagnostics that share these categories.

### Telemetry data

A 30-minute production `FunctionsLogs` sample contained ~992.5M Debug/Trace rows. The queue rules cover 13.52% and the blob rules 4.76%; together with the existing Event Hub producer suppression the targeted rules cover ~20.6% of Debug/Trace volume, while the retained blob diagnostics are 3–4 orders of magnitude smaller (for example `BlobMessageEnqueued` at ~118K).

Event Hub `PartitionProcessor` (8.87%) and Kafka (2.40%) are intentionally **not** suppressed. Their Debug logs carry no EventName, so they cannot be filtered without discarding partition lifecycle, checkpoint, batching, and native-initialization diagnostics. Targeting those requires the extensions to emit stable named EventIds first.

### Scope and safety

- Applies only to the internal `SystemLogger` sink, so customer-configured providers (Application Insights, OpenTelemetry, console) continue to honor `host.json`.
- Only affects Debug/Trace; Information and above are untouched.
- Diagnostic mode bypasses suppression entirely.
- The suppression check runs before the message formatter, so suppressed logs are not formatted.

### Pull request checklist

### DIFF
--- a/src/WebJobs.Script/Extensions/ScriptLoggingBuilderExtensions.cs
+++ b/src/WebJobs.Script/Extensions/ScriptLoggingBuilderExtensions.cs
@@ -23,8 +23,12 @@ namespace Microsoft.Extensions.Logging
         // suppressed to reduce noise in the FunctionsLogs table.
         private static readonly HashSet<string> _suppressedCategories = new(StringComparer.Ordinal)
         {
+            "Host.Triggers.Kafka",
+            "Microsoft.Azure.WebJobs.EventHubs.Listeners.EventHubListener.PartitionProcessor",
             "Microsoft.Azure.WebJobs.Extensions.Storage.Common.Listeners.QueueListener",
+            "Microsoft.Azure.WebJobs.Extensions.Storage.Blobs.Listeners.BlobListener",
             "Microsoft.Azure.WebJobs.EventHubs.EventHubProducerClientImpl",
+            "Microsoft.Azure.WebJobs.Host.Queues.Listeners.QueueListener",
             "Host.Executor"
         };
 
--- a/test/WebJobs.Script.Tests/Extensions/ScriptLoggingBuilderExtensionsTests.cs
+++ b/test/WebJobs.Script.Tests/Extensions/ScriptLoggingBuilderExtensionsTests.cs
@@ -62,6 +62,14 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.Extensions
         [InlineData("Microsoft.Azure.WebJobs.EventHubs.EventHubProducerClientImpl", LogLevel.Trace)]
         [InlineData("Host.Executor", LogLevel.Debug)]
         [InlineData("Host.Executor", LogLevel.Trace)]
+        [InlineData("Microsoft.Azure.WebJobs.EventHubs.Listeners.EventHubListener.PartitionProcessor", LogLevel.Debug)]
+        [InlineData("Microsoft.Azure.WebJobs.EventHubs.Listeners.EventHubListener.PartitionProcessor", LogLevel.Trace)]
+        [InlineData("Microsoft.Azure.WebJobs.Extensions.Storage.Blobs.Listeners.BlobListener", LogLevel.Debug)]
+        [InlineData("Microsoft.Azure.WebJobs.Extensions.Storage.Blobs.Listeners.BlobListener", LogLevel.Trace)]
+        [InlineData("Host.Triggers.Kafka", LogLevel.Debug)]
+        [InlineData("Host.Triggers.Kafka", LogLevel.Trace)]
+        [InlineData("Microsoft.Azure.WebJobs.Host.Queues.Listeners.QueueListener", LogLevel.Debug)]
+        [InlineData("Microsoft.Azure.WebJobs.Host.Queues.Listeners.QueueListener", LogLevel.Trace)]
         public void Filter_SuppressedCategory_DebugAndTrace_ReturnsFalse(string category, LogLevel level)
         {
             Assert.False(ScriptLoggingBuilderExtensions.Filter(category, level, LogLevel.Trace));
@@ -74,6 +82,14 @@ namespace Microsoft.Azure.WebJobs.Script.Tests.Extensions
         [InlineData("Microsoft.Azure.WebJobs.EventHubs.EventHubProducerClientImpl", LogLevel.Error)]
         [InlineData("Host.Executor", LogLevel.Information)]
         [InlineData("Host.Executor", LogLevel.Critical)]
+        [InlineData("Microsoft.Azure.WebJobs.EventHubs.Listeners.EventHubListener.PartitionProcessor", LogLevel.Information)]
+        [InlineData("Microsoft.Azure.WebJobs.EventHubs.Listeners.EventHubListener.PartitionProcessor", LogLevel.Warning)]
+        [InlineData("Microsoft.Azure.WebJobs.Extensions.Storage.Blobs.Listeners.BlobListener", LogLevel.Information)]
+        [InlineData("Microsoft.Azure.WebJobs.Extensions.Storage.Blobs.Listeners.BlobListener", LogLevel.Error)]
+        [InlineData("Host.Triggers.Kafka", LogLevel.Information)]
+        [InlineData("Host.Triggers.Kafka", LogLevel.Critical)]
+        [InlineData("Microsoft.Azure.WebJobs.Host.Queues.Listeners.QueueListener", LogLevel.Information)]
+        [InlineData("Microsoft.Azure.WebJobs.Host.Queues.Listeners.QueueListener", LogLevel.Warning)]
         public void Filter_SuppressedCategory_InformationAndAbove_ReturnsTrue(string category, LogLevel level)
         {
             Assert.True(ScriptLoggingBuilderExtensions.Filter(category, level, LogLevel.Trace));


src/WebJobs.Script/Extensions/ScriptLoggingBuilderExtensions.cs
```diff
@@ -23,8 +23,12 @@ public static class ScriptLoggingBuilderExtensions
         // suppressed to reduce noise in the FunctionsLogs table.
         private static readonly HashSet<string> _suppressedCategories = new(StringComparer.Ordinal)
         {
+            "Host.Triggers.Kafka",
+            "Microsoft.Azure.WebJobs.EventHubs.Listeners.EventHubListener.PartitionProcessor",
             "Microsoft.Azure.WebJobs.Extensions.Storage.Common.Listeners.QueueListener",
+            "Microsoft.Azure.WebJobs.Extensions.Storage.Blobs.Listeners.BlobListener",
             "Microsoft.Azure.WebJobs.EventHubs.EventHubProducerClientImpl",
+            "Microsoft.Azure.WebJobs.Host.Queues.Listeners.QueueListener",
```
brettsam, line 31: Did you double-check what we'll lose from this? Is it just the polling logs? We should still have the option to do a { `Category`, `EventName` } pair for these if there's something we want to be more targeted with.

brettsam (approved):

src/WebJobs.Script/Extensions/ScriptLoggingBuilderExtensions.cs
mathewc: Yeah, I think the category/eventname approach would be safer. All of the listener start/stop lifecycle logs appear to be logged with this category, e.g. https://github.com/Azure/azure-sdk-for-net/blob/ca61f202024f99d33f4693fe3176d54092f41205/sdk/storage/Microsoft.Azure.WebJobs.Extensions.Storage.Common/src/Shared/Queues/QueueListener.cs#L173. We don't want to lose those.

mathewc, line 31: Ah subtle point here - this source you're adding Microsoft.Azure.WebJobs.Host.Queues.Listeners.QueueListener is for older versions of the extension - the current QueueListener in newer storage extension is in namespace Microsoft.Azure.WebJobs.Extensions.Storage.Common.Listeners.QueueListener. The code you're modifying here already filters Microsoft.Azure.WebJobs.Extensions.Storage.Common.Listeners.QueueListener - has that been released yet? We don't want to be filtering all these QueueListener debug logs out

src/WebJobs.Script/Extensions/ScriptLoggingBuilderExtensions.cs
```diff
@@ -23,8 +23,12 @@ public static class ScriptLoggingBuilderExtensions
         // suppressed to reduce noise in the FunctionsLogs table.
         private static readonly HashSet<string> _suppressedCategories = new(StringComparer.Ordinal)
         {
+            "Host.Triggers.Kafka",
+            "Microsoft.Azure.WebJobs.EventHubs.Listeners.EventHubListener.PartitionProcessor",
             "Microsoft.Azure.WebJobs.Extensions.Storage.Common.Listeners.QueueListener",
+            "Microsoft.Azure.WebJobs.Extensions.Storage.Blobs.Listeners.BlobListener",
```
mathewc, line 29: The logger for this listener is here https://github.com/Azure/azure-sdk-for-net/blob/ca61f202024f99d33f4693fe3176d54092f41205/sdk/storage/Microsoft.Azure.WebJobs.Extensions.Storage.Blobs/src/Listeners/BlobListener.cs#L35 and appears only to be used to log listener start/stop lifecycle events. Those are important and also aren't high volume, so why do we want to suppress them?

Surveying prod logs, I see the following event names being logged using this source: "BlobAlreadyProcessed", "PollBlobContainer", "BlobDoesNotMatchPattern", "BlobMessageEnqueued", "ContainerDoesNotExist", "InitializedScanInfo", "ScanBlobLogs", "FunctionNotFound", "BlobHasNoETag"

Some of those are clearly noisy, but some are useful for debugging and infrequent. I worry that simply throwing all of these away is going to hamper our ability to investigate issues.

src/WebJobs.Script/Extensions/ScriptLoggingBuilderExtensions.cs
```diff
@@ -23,8 +23,12 @@ public static class ScriptLoggingBuilderExtensions
         // suppressed to reduce noise in the FunctionsLogs table.
         private static readonly HashSet<string> _suppressedCategories = new(StringComparer.Ordinal)
         {
+            "Host.Triggers.Kafka",
```
mathewc, line 26: @TsuyoshiUshio to review what types of logs we emit using this category and whether anything important would be lost

src/WebJobs.Script/Extensions/ScriptLoggingBuilderExtensions.cs
```diff
@@ -23,8 +23,12 @@ public static class ScriptLoggingBuilderExtensions
...
+            "Microsoft.Azure.WebJobs.EventHubs.Listeners.EventHubListener.PartitionProcessor",
```
mathewc, line 27: I notice that for EventHub, you're not suppressing Microsoft.Azure.WebJobs.EventHubs.Listeners.EventHubListener. If we're inconsistent across extensions wrt. suppression of their listener start/stop lifecycle events as we will be with these changes, I wonder if that will cause confusion - we'll no longer know which extensions we can count on these logs for in investigations.

RohitRanjanMS: Event Hub PartitionProcessor category suppression has been removed. Its Debug events are currently unnamed and include lifecycle, batching, checkpoint, and timing diagnostics, so they should remain until the extension exposes stable EventNames for any proven noise targets.

src/WebJobs.Script/Extensions/ScriptLoggingBuilderExtensions.cs
RohitRanjanMS: Updated the implementation to filter inside `SystemLogger.Log`, where Category, EventName, and LogLevel are all available. Queue suppression is now limited to the named `GetMessages` and `BackoffDelay` events, so listener lifecycle and `HandlingStorageException` logs are preserved. Diagnostic mode bypasses suppression.

src/WebJobs.Script/Extensions/ScriptLoggingBuilderExtensions.cs
RohitRanjanMS: Agreed. Category-wide Blob suppression has been removed. The updated implementation suppresses only the named `PollBlobContainer` event below Information; lifecycle, initialization, enqueue, pattern, ETag, missing-function, and missing-blob diagnostics remain available.

src/WebJobs.Script/Extensions/ScriptLoggingBuilderExtensions.cs
RohitRanjanMS: Kafka category suppression has been removed from this PR. Its Debug events currently have no stable EventNames, so targeted filtering is not possible without an extension change or a dedicated librdkafka category.

src/WebJobs.Script/Extensions/ScriptLoggingBuilderExtensions.cs
RohitRanjanMS: The legacy category is now handled with the same narrow named-event rules as the current category: only `GetMessages` and `BackoffDelay` below Information are suppressed. Other Debug events are not filtered.

src/WebJobs.Script/Extensions/ScriptLoggingBuilderExtensions.cs
RohitRanjanMS: Follow-up after querying a 30-minute production `FunctionsLogs` window: `BlobDoesNotMatchPattern` produced ~35.6M Debug rows and `BlobAlreadyProcessed` ~11.2M, versus ~118K for `BlobMessageEnqueued` and fewer than 100K for each diagnostic/error event. I added those two high-volume named events to the targeted rules. This suppression applies only to internal `SystemLogger`/FunctionsLogs, only below Information, and is bypassed in diagnostic mode; customer-configured providers can still ingest them through `host.json`. The remaining Blob lifecycle and diagnostic events are preserved.

PR #11949, opened 2026-08-28
adding WorkerProxy relay

brettsam:
Adds a bidirectional `FunctionRpc` relay to WorkerProxy.

WorkerProxy now hosts separate HTTP/2 endpoints for the Host runtime and language worker. The relay pairs one stream from each side into a session and forwards opaque `StreamingMessage` values in both directions without changing the existing wire protocol.

## Relay behavior

- Either side may connect first; messages remain ordered until the peer connects.
- Each direction has one channel reader, one channel writer, and one writer to the gRPC response stream.
- Duplicate active streams are rejected with `AlreadyExists`; reconnects during teardown receive transient `Unavailable`.
- The first peer close, cancellation, transport failure, or application shutdown terminates the entire session and stops both sides.
- A replacement session is created only after both previous attachments release, preventing streams from different sessions from being paired.
- Listener startup, shutdown, cleanup, and disposal are serialized through an explicit lifecycle state machine.

## Deferred work

This change intentionally does not interpret or modify FunctionRpc messages. Later changes will add worker assignment and readiness, Host configuration injection, worker HTTP endpoint rewriting and forwarding, higher-level reconnect/link policy, management APIs and authentication, drain behavior, diagnostics, and production Host/platform integration.

### Pull request checklist


src/Functions.WorkerProxy/FunctionRpcRelay.Session.cs
```diff
@@ -0,0 +1,350 @@
...
+namespace Azure.Functions.WorkerProxy;
+
+internal sealed partial class FunctionRpcRelay
+{
+    /// <summary>
+    /// Owns the queues, forwarding tasks, and terminal state for one runtime/worker stream pair.
+    /// </summary>
+    private sealed class FunctionRpcRelaySession
+    {
+        private readonly object _stateLock = new();
```
kshyju: Should we use `System.Threading.Lock` type instead of object? That seems like the modern way/API to do this.

src/Functions.WorkerProxy/FunctionRpcRelay.Session.cs
```diff
@@ -0,0 +1,350 @@
...
+            Task writeTask = WriteOutboundAsync(source, responseStream, linkedSource.Token);
+            Task firstCompletedTask = await Task.WhenAny(readTask, writeTask);
+
+            if (!_completion.Task.IsCompleted)
+            {
+                FunctionRpcRelayTerminalState terminalState = await ClassifyCompletionAsync(side, firstCompletedTask, cancellationToken);
+                TryTerminate(terminalState);
+            }
+
+            linkedSource.Cancel();
```
kshyju: Should we use the async version of Cancel? 
```
await linkedSource.CancelAsync();
```

src/Functions.WorkerProxy/FunctionRpcRelayOptions.cs
```diff
@@ -0,0 +1,96 @@
...
+            throw new InvalidOperationException($"Configuration values '{RuntimeGrpcPortKey}' and " +
+                $"'{WorkerGrpcPortKey}' must identify different ports.");
+        }
+
+        return new FunctionRpcRelayOptions(runtimeGrpcPort, workerGrpcPort);
+    }
+
+    private static int ReadInt32(IConfiguration configuration, string key, int defaultValue, int minimumValue, int maximumValue)
+    {
+        string environmentKey = key.Replace('-', '_').ToUpperInvariant();
```
kshyju: Do we need the ToUpper call? I think keys are case insensitive as per docs: https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/fundamentals/configuration/index.md?utm_source=chatgpt.com#configuration-keys-and-values

kshyju: Creating this string (enveironmentKey) can be moved down, closer to usage, so we don't allocate if `configuration[key]` is present/returns a value.

src/Functions.WorkerProxy/FunctionRpcRelayServer.cs
```diff
@@ -0,0 +1,294 @@
...
+                case LifecycleState.Stopped:
+                case LifecycleState.Disposed:
+                    return;
+                default:
+                    throw new InvalidOperationException($"Unknown FunctionRpc relay lifecycle state '{_state}'.");
+            }
+
+            try
+            {
+                await StopCoreAsync(cancellationToken);
```
kshyju: Do we need any debug logging here so we know framework called StopAsync)?

src/Functions.WorkerProxy/FunctionRpcRelayServer.cs
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    private async Task StopCoreAsync(CancellationToken cancellationToken)
+    {
+        Task relayStopTask = _relay.StopAsync(cancellationToken);
+        Task applicationStopTask = StopApplicationsAsync(_runtimeApplication, _workerApplication, cancellationToken);
+        await Task.WhenAll(relayStopTask, applicationStopTask);
+    }
+
+    private WebApplication BuildEndpointApplication(int port, FunctionRpcRelaySide side)
```
kshyju: __Asked GHCP about this method and it came up with below.__

------------------------------------------------
### Notes on `BuildEndpointApplication`

Poked at this method a bit and ran some measurements. Four things, roughly in order of value.

#### 1. `CreateSlimBuilder` → `CreateEmptyBuilder`

Slim isn't actually minimal — it still loads env vars, command line, `appsettings.json`, logging providers, and `ConsoleLifetime`. The endpoint apps need none of it: their only inputs are `port` and `side`, and both are passed as arguments.

```csharp
WebApplicationBuilder builder = WebApplication.CreateEmptyBuilder(new WebApplicationOptions());
builder.WebHost.UseKestrelCore();
builder.WebHost.ConfigureKestrel(...);   // unchanged
builder.Services.AddRoutingCore();
builder.Services.AddGrpc(...);           // unchanged
// ... singletons unchanged

WebApplication application = builder.Build();
application.UseRouting();                // now explicit
application.MapGrpcService<FunctionRpcRelayService>();
```

I built both variants and pushed real bidirectional gRPC streams through them to make sure this isn't just a startup-time illusion:

| variant | startup | heap | alloc | working set | bidi stream |
|---|---|---|---|---|---|
| `CreateSlimBuilder` | 241 / 166 ms | 497 KB | 823 KB | 42 MB | ok |
| `CreateEmptyBuilder` | 110 / 116 ms | 436 KB | 621 KB | 38 MB | ok |

~40% faster startup, ~25% fewer allocations, ~4 MB less working set — and that's per app, so double it for two. `UseKestrelCore` and `AddRoutingCore` are also the AOT-friendly variants, which matters since this project has the AOT analyzers enabled. Bonus: it drops `ConsoleLifetime` from the endpoint apps, so we stop carrying three sets of SIGTERM/Ctrl+C handlers in one process.

#### 2. The endpoint apps inherit ambient env config

The Dockerfile sets `ASPNETCORE_HTTP_PORTS=80`, and `CreateSlimBuilder()` picks that up in both endpoint apps. Reproduced it:

```
warn: Microsoft.AspNetCore.Server.Kestrel[0]
      Overriding address(es) 'http://*:18099'. Binding to endpoints
      defined via IConfiguration and/or UseKestrel() instead.
```

Not a bug today — `ListenAnyIP` wins, the address collection is cleared and repopulated, `Urls` stays correct. But the relay endpoints are silently entangled with ambient configuration, and the warning that would tell us so is suppressed by `ClearProviders()`. `CreateEmptyBuilder` drops the config sources entirely, so this stops being something we have to reason about.

Related: `GetAddress` uses `Urls.SingleOrDefault()`, which throws `"Sequence contains more than one element"` — not the friendly "listener has not started" message — if anything ever contributes a second address. A `Single`/match-by-port would fail more clearly.

#### 3. `ClearProviders()` blinds the endpoint apps

Same repro as above shows it swallowing a legitimate Kestrel warning; it also swallows gRPC and transport-level errors inside those apps. The relay's own logging survives only because `_logger` comes from the outer container.

Since the container doesn't dispose instances registered via `AddSingleton(instance)`, we can just share the parent pipeline instead:

```csharp
builder.Services.AddSingleton(_loggerFactory);
```

Endpoint-app diagnostics then flow into the main logging pipeline, and we skip building a second one.

#### 4. `ListenAnyIP` vs. what the existing server does

`AspNetCoreGrpcHostBuilder` binds `IPAddress.Parse(WorkerConstants.HostName)` (`127.0.0.1`), with the comment *"This server is exclusively between host and worker. All clients are trusted."*

Here we're on `ListenAnyIP` for both ports. The runtime side needs that under compute separation, but does the worker side? If the worker is a sidecar in the same pod, `ListenLocalhost` would keep the less-trusted party off the network surface. Worth deciding deliberately, since neither port has auth yet.

kshyju (commented):
Few comments, as I am reading the PR to understand the flow (with GHCP helping me to walk me through the flow). Will continue to review today.

src/Functions.WorkerProxy/FunctionRpcRelay.cs
```diff
@@ -0,0 +1,189 @@
...
+/// Owns the active pair of runtime-facing and worker-facing FunctionRpc streams.
+/// </summary>
+/// <remarks>
+/// A relay session admits one stream per <see cref="FunctionRpcRelaySide"/>. The first peer close,
+/// cancellation, stream failure, or application shutdown terminates the whole session. A replacement
+/// session is created only after both attachments from the previous session have released.
+/// </remarks>
+internal sealed partial class FunctionRpcRelay : IAsyncDisposable
+{
+    private readonly object _syncLock = new();
```
jviau: Use `System.Threading.Lock`

src/Functions.WorkerProxy/FunctionRpcRelay.cs
```diff
@@ -0,0 +1,189 @@
...
+    private FunctionRpcRelaySession? _currentSession;
+    private FunctionRpcRelayTerminalState? _lastTerminalState;
+    private long _nextSessionId;
+    private bool _stopping;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="FunctionRpcRelay"/> class.
+    /// </summary>
+    /// <param name="logger">The logger used for terminal and secondary stream failures.</param>
+    public FunctionRpcRelay(ILogger<FunctionRpcRelay> logger)
```
jviau: style nit: prefer primary ctors when possible

src/Functions.WorkerProxy/FunctionRpcRelay.cs
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    /// <summary>
+    /// Gets the terminal state of the most recently completed relay session.
+    /// </summary>
+    internal FunctionRpcRelayTerminalState? LastTerminalState
+    {
+        get
+        {
+            lock (_syncLock)
```
jviau: This lock doesn't seem necessary. A reference read is atomic in C#. And the lock will be released by the time this property is actually used by the caller, so it is not protecting anything.

test/Functions.WorkerProxy.Tests/FunctionRpcRelayTests.cs
```diff
@@ -0,0 +1,477 @@
...
+    private static RelayClient CreateClient(Uri address, CancellationToken cancellationToken)
+    {
+        UriBuilder normalizedAddress = new(address);
+        if (IPAddress.TryParse(address.Host.Trim('[', ']'), out IPAddress? ipAddress)
+            && (ipAddress.Equals(IPAddress.Any) || ipAddress.Equals(IPAddress.IPv6Any)))
+        {
+            normalizedAddress.Host = IPAddress.Loopback.ToString();
+        }
+
+        GrpcChannel channel = GrpcChannel.ForAddress(normalizedAddress.Uri);
```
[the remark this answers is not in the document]
brettsam: not needed for .net 10

src/Functions.WorkerProxy/FunctionRpcRelay.cs
```diff
@@ -0,0 +1,189 @@
...
+            if (attachResult != FunctionRpcRelayAttachResult.Attached)
+            {
+                FunctionRpcRelayAttachmentFailure failure = attachResult switch
+                {
+                    FunctionRpcRelayAttachResult.Duplicate => FunctionRpcRelayAttachmentFailure.Duplicate,
+                    FunctionRpcRelayAttachResult.Terminated =>
+                        FunctionRpcRelayAttachmentFailure.PreviousSessionTearingDown,
+                    _ => throw new InvalidOperationException($"Unexpected attach result '{attachResult}'.")
+                };
+                string message = failure == FunctionRpcRelayAttachmentFailure.Duplicate
```
jviau: small style nit: have the switch expression return a tuple

```csharp
(FunctionRpcRelayAttachmentFailure failure, string message) = attachResult switch
{
 ....
}
```

src/Functions.WorkerProxy/FunctionRpcRelay.cs
```diff
@@ -0,0 +1,189 @@
...
+
+        return RunAttachmentAsync(session, side, requestStream, responseStream, cancellationToken);
+    }
+
+    /// <summary>
+    /// Prevents new attachments and terminates the active relay session.
+    /// </summary>
+    /// <param name="cancellationToken">A token that bounds the wait for all attachments to release.</param>
+    /// <returns>A task that completes when the active session has been cleared.</returns>
+    public async Task StopAsync(CancellationToken cancellationToken)
```
jviau: Not sure how impactful this will be, but two concurrent callers here will not both experience the same `StopAsync`. One caller will perform the actual stop, and the other will short circuit at `if (session is null)` and return before the stop is actually done.

To make them consistent you will need a `TaskCompletionSource` to track the stop, and have all callers join the stop by awaiting that TCS.

src/Functions.WorkerProxy/FunctionRpcRelay.cs
```diff
@@ -0,0 +1,189 @@
...
+            {
+                ClearCurrentSessionLocked();
+            }
+        }
+    }
+
+    private void ClearCurrentSessionLocked()
+    {
+        FunctionRpcRelaySession session = _currentSession ?? throw new InvalidOperationException("There is no current relay session to clear.");
+        _lastTerminalState = session.Completion.GetAwaiter().GetResult();
```
jviau: Can we avoid this sync-over-async?

src/Functions.WorkerProxy/FunctionRpcRelay.Session.cs
```diff
@@ -0,0 +1,352 @@
...
+
+        /// <summary>
+        /// Records application shutdown as the terminal state for this session.
+        /// </summary>
+        public void RequestShutdown()
+        {
+            TryTerminate(new FunctionRpcRelayTerminalState(Id, FunctionRpcRelayTerminationReason.Shutdown, Side: null, Exception: null));
+        }
+
+        private static Channel<StreamingMessage> CreateQueue()
```
jviau:
```suggestion
        private static Channel<StreamingMessage> CreateChannel()
```

src/Functions.WorkerProxy/FunctionRpcRelay.Session.cs
```diff
@@ -0,0 +1,352 @@
...
+            }
+
+            await linkedSource.CancelAsync();
+            // Reads can remain blocked until the transport releases them, so observe that task
+            // out of band. The write must finish before gRPC writes trailers to the same response.
+            _ = ObserveStreamOperationAfterTerminationAsync(side, "read", readTask);
+            await ObserveStreamOperationAfterTerminationAsync(side, "write", writeTask);
+
+            FunctionRpcRelayTerminalState finalState = await _completion.Task;
+            throw new FunctionRpcRelayTerminatedException(finalState);
```
jviau: Does this method always throw? I don't see a regular `return` anywhere.

src/Functions.WorkerProxy/FunctionRpcRelay.Session.cs
```diff
@@ -0,0 +1,352 @@
...
+                FunctionRpcRelayTerminationReason.Faulted => terminalState.Exception,
+                FunctionRpcRelayTerminationReason.Canceled => terminalState.Exception,
+                _ => null
+            };
+            _toRuntime.Writer.TryComplete(completionException);
+            _toWorker.Writer.TryComplete(completionException);
+
+            if (terminalState.Reason == FunctionRpcRelayTerminationReason.Faulted)
+            {
+                _logger.LogWarning(terminalState.Exception, "FunctionRpc relay session {SessionId} faulted on the {Side} side.",
```
jviau: nit: use logger source gen

src/Functions.WorkerProxy/WorkerProxyApplication.cs
```diff
@@ -4,21 +4,35 @@
...
     public static WebApplication Build(string[] args)
     {
         ArgumentNullException.ThrowIfNull(args);
 
         // Endpoint selection intentionally follows standard ASP.NET Core configuration and
         // precedence. For example, URLS overrides HTTP_PORTS when both are present.
-        WebApplicationBuilder builder =
-            WebApplication.CreateSlimBuilder(new WebApplicationOptions { Args = args });
+        WebApplicationBuilder builder = WebApplication.CreateSlimBuilder(new WebApplicationOptions { Args = args });
+        builder.Services.AddSingleton(static services => FunctionRpcRelayOptions.FromConfiguration(services.GetRequiredService<IConfiguration>()));
```
jviau: Why not use `IOptions`? This avoids needing all the manual `IConfiguration` binding,

src/Functions.WorkerProxy/FunctionRpcRelayOptions.cs
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    /// <summary>
+    /// Reads one integer option, returning its default when the option is absent and rejecting values outside the supported range.
+    /// </summary>
+    /// <remarks>
+    /// The original key supports command-line configuration (for example, <c>--runtime-grpc-port</c>).
+    /// The underscore form supports the equivalent environment variable, such as <c>RUNTIME_GRPC_PORT</c>.
+    /// </remarks>
+    private static int ReadInt32(IConfiguration configuration, string key, int defaultValue, int minimumValue, int maximumValue)
```
jviau: Can you not use `IConfiguration.Bind`?

src/Functions.WorkerProxy/FunctionRpcRelayServer.cs
```diff
@@ -0,0 +1,337 @@
...
+    private LifecycleState _state;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="FunctionRpcRelayServer"/> class.
+    /// </summary>
+    /// <param name="relay">The singleton relay shared by both endpoint applications.</param>
+    /// <param name="options">The listener configuration.</param>
+    /// <param name="loggerFactory">The parent logging pipeline shared by both endpoint applications.</param>
+    /// <param name="logger">The listener lifecycle logger.</param>
+    public FunctionRpcRelayServer(FunctionRpcRelay relay, FunctionRpcRelayOptions options, ILoggerFactory loggerFactory,
```
jviau: Can probably remove the `ILogger<T>` since you already have `ILoggerFactory`.

src/Functions.WorkerProxy/FunctionRpcRelayServer.cs
```diff
@@ -0,0 +1,337 @@
...
+    private async Task StopCoreAsync(CancellationToken cancellationToken)
+    {
+        Task relayStopTask = _relay.StopAsync(cancellationToken);
+        Task applicationStopTask = StopApplicationsAsync(_runtimeApplication, _workerApplication, cancellationToken);
+        await Task.WhenAll(relayStopTask, applicationStopTask);
+    }
+
+    private WebApplication BuildEndpointApplication(int port, FunctionRpcRelaySide side)
+    {
+        WebApplicationBuilder builder = WebApplication.CreateEmptyBuilder(new WebApplicationOptions());
```
jviau: Why does this build a separate nested WebApp instead of using the outer WebApp?

We can have the outer webapp listen to both ports, and then make the endpoints conditionally available based on the local port the caller has used.

src/Functions.WorkerProxy/FunctionRpcRelayService.cs
```diff
@@ -0,0 +1,82 @@
...
+{
+    private readonly FunctionRpcRelay _relay;
+    private readonly FunctionRpcRelayEndpoint _endpoint;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="FunctionRpcRelayService"/> class.
+    /// </summary>
+    /// <param name="relay">The shared FunctionRpc relay.</param>
+    /// <param name="endpoint">The immutable side owned by this endpoint application.</param>
+    public FunctionRpcRelayService(FunctionRpcRelay relay, FunctionRpcRelayEndpoint endpoint)
```
jviau: If we want to consolidate to a single service container (as I mentioned in the other comment), we can use a keyed service with the `FunctionRpcRelaySide` enum as the key.

src/Functions.WorkerProxy/FunctionRpcRelayService.cs
```diff
@@ -0,0 +1,82 @@
...
+        try
+        {
+            await _relay.AttachAsync(_endpoint.Side, requestStream, responseStream, context.CancellationToken);
+        }
+        catch (FunctionRpcRelayAttachmentException exception)
+        {
+            StatusCode statusCode = exception.Failure switch
+            {
+                FunctionRpcRelayAttachmentFailure.Duplicate => StatusCode.AlreadyExists,
+                FunctionRpcRelayAttachmentFailure.PreviousSessionTearingDown => StatusCode.Unavailable,
```
jviau, line 42: I am curious about the "PreviousSessionTearingDown" behavior. Is this saying there is a window where we expect the relay to be unavailable, and callers need to know to wait for a period before establishing the event stream in these cases?

How often do we expect this to happen in the real world?

jviau:
Main concerns / things I'd like to see:

1. Use `IOptions` instead of manual construction and `IConfiguration` parsing
2. Consolidate to registering both gRPC endpoints on the same root web host, condition the behavior based on LocalPort from the HTTP context.
3. Use Keyed services, if necessary, for resolving the correct "side"-specific services

src/Functions.WorkerProxy/FunctionRpcRelayService.cs
brettsam: Yes, that's correct. We don't expect it to happen often. I went back-and-forth on this for a bit -- but I think that matches closest with what we have today in that if one side goes down, we effectively have to "start over". 

Once full e2e is ready, we may want to revisit this behavior and it's possible there's a better mechanism, esp if we see more disconnects than we expect.

There's a "resilience" milestone built into all of this work, and that's where bashing on something like this would come in.

src/Functions.WorkerProxy/FunctionRpcRelayServer.cs
brettsam: That was the most straightforward approach I could come up with :-) -- but your idea is good. Refactoring now. That will address a bunch of your comments when it comes  in.

src/Functions.WorkerProxy/FunctionRpcRelay.Session.cs
brettsam: Yeah, it was using the exception to teardown the grpc service and report final state. I'm changing this as I don't like it either...

src/Functions.WorkerProxy/FunctionRpcRelayService.cs
```diff
@@ -4,41 +4,38 @@
...
-    /// Initializes a new instance of the <see cref="FunctionRpcRelayService"/> class.
-    /// </summary>
-    /// <param name="relay">The shared FunctionRpc relay.</param>
-    /// <param name="endpoint">The immutable side owned by this endpoint application.</param>
-    public FunctionRpcRelayService(FunctionRpcRelay relay, FunctionRpcRelayEndpoint endpoint)
-    {
-        _relay = relay;
-        _endpoint = endpoint;
-    }
+    private readonly FunctionRpcRelay _relay = relay;
```
jviau: With primary ctors you don't need the fields unless you plan on doing arg validation or something. You can access the ctor params directly throughout the class.

My rec: either add arg validation here or remove the fields.

jviau (approved):

src/Functions.WorkerProxy/FunctionRpcRelayService.cs
brettsam: somehow i missed a bunch of these -- fixing now.

jviau (approved):

kshyju (approved):

PR #11950, opened 2026-08-28
removing GrpcEvent envelopes

brettsam:
With compute separation, we'll need a way to pass our `Channel<StreamingMessage>` directly into the `WorkerChannel` so we can re-use all of the existing logic. Today, that relies on an odd channel registration via the `IScriptEventManager` and uses `InboundGrpcEvent`  and `OutboundGrpcEvent` to do its message types. These are just wrappers around `StreamingMessage`.

I don't want to make these `GrpcEvent` types public and consume them in the new Rpc.Client assembly, so instead, I'm removing them and we'll use `StreamingMessage` directly. 

That had a bonus... perf gain. Now that we're not instantiating wrapper types on *every* invocation. The numbers below are for a write and read pair of operations. Small gains, but drops 160 bytes of allocation to 0 and halves the (already small) time:

|                         Method |      Mean |    Error |   StdDev |    Median |       Min |       Max | Ratio |  Gen 0 | Allocated |
|------------------------------- |----------:|---------:|---------:|----------:|----------:|----------:|------:|-------:|----------:|
|         LegacyEnvelopeChannels | 102.49 ns | 1.706 ns | 1.595 ns | 102.21 ns | 100.09 ns | 105.27 ns |  1.00 | 0.0091 |     160 B |
| DirectStreamingMessageChannels |  47.72 ns | 0.204 ns | 0.170 ns |  47.67 ns |  47.58 ns |  48.15 ns |  0.47 |      - |         - |

This change is mostly mechanical -- and what is touching a lot of the files. There's some other things here as well:

1. To simplify the requirements, I'm introducing an abstract base class `DuplexChannel<T>`  that gives us something to easily pass into the `WorkerChannel` with the endpoints it needs. We'll update the Rpc.Cliient channel to derive from this so it can be used here in a follow-up. This is now what's stored in the `IScriptEventManager` for the `RpcServer` to use when it gets a connection as well. It has a special method it calls to get the *other end* of the channels so things pump through.
2. `WorkerChannel` is now `IAsyncDisposable` because it has to dispose the passed-in `DuplexChannel<T>`. It's not required now but it will be. It always owned the disposal of the channels that it was using, but it wasn't entirely clear that this is what was happening... now it's more clear. 

### Deferred change
~~I want to remove the way we store these channels in `IScriptEventManager`, but that gets bigger and don't want to bog down this already big PR (even though it's mostly search/repplace).~~

*Note I ended up adding this deferred change. ^*

### Pull request checklist


src/WebJobs.Script.Grpc/Channel/WorkerChannel.cs
```diff
@@ -1263,9 +1268,9 @@ internal void SendCloseSharedMemoryResourcesForInvocationRequest(IList<string> o
             });
         }
 
-        internal void Log(GrpcEvent msg)
+        internal void Log(StreamingMessage msg)
         {
-            var rpcLog = msg.Message.RpcLog;
+            var rpcLog = msg.RpcLog;
```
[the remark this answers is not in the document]
brettsam: that's fine; we'll be re-generating.

src/Functions.Rpc.Server/Channel/GrpcWorkerChannelFactory.cs
```diff
@@ -58,21 +59,34 @@ public IRpcWorkerChannel Create(string scriptRootPath, string runtime, IMetricsL
...
+                return CreateInternal(workerId, channel, _eventManager, _hostManager,
+                    languageWorkerConfig, rpcWorkerProcess, workerLogger, metricsLogger, attemptCount, _environment, _applicationHostOptions,
+                    _sharedMemoryManager, _workerConcurrencyOptions, _hostingConfigOptions, _appCapabilitiesStore, _httpProxyService);
+            }
+            catch
+            {
+                // The standard server channel always completes disposal synchronously.
+                channel.DisposeAsync().GetAwaiter().GetResult();
+                throw;
+            }
```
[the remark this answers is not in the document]
brettsam: This cleanup never happened before. In fact, I'm going to back out the try/catch so we don't add any nrew behavior with this PR as it's mostly supposed to be mechanical. Will clean up this in a follow-up PR.

src/WebJobs.Script.Grpc/Channel/DuplexChannel.cs
```diff
@@ -0,0 +1,37 @@
...
+
+    /// <summary>
+    /// Asynchronously releases the channel and its owned resources.
+    /// </summary>
+    /// <returns>A task representing the asynchronous disposal operation.</returns>
+    public ValueTask DisposeAsync()
+    {
+        lock (_disposeLock)
+        {
+            _disposeTask ??= DisposeAsyncCore();
```
jviau: You could avoid all of the async callers blocking on the `lock` by checking if `_disposeTask` is non-null before entering the lock, and returning it then if so.


```csharp
if (_disposeTask is { } t)
{
    return new(t);
}

lock (_disposeLock)
{
    _disposeTask ??= DisposeAsyncCore();
    return new(_disposeTask);
}
```

src/WebJobs.Script.Grpc/Channel/DuplexChannel.cs
```diff
@@ -0,0 +1,37 @@
...
+
+namespace Microsoft.Azure.WebJobs.Script.Grpc;
+
+/// <summary>
+/// Represents an owned bidirectional message channel.
+/// </summary>
+/// <typeparam name="T">The transported message type.</typeparam>
+public abstract class DuplexChannel<T> : Channel<T>, IAsyncDisposable
+{
+    private readonly object _disposeLock = new object();
```
jviau: Use System.Threading.Lock

src/WebJobs.Script.Grpc/Channel/DuplexChannel.cs
```diff
@@ -0,0 +1,37 @@
...
+            _disposeTask ??= DisposeAsyncCore();
+            return new ValueTask(_disposeTask);
+        }
+    }
+
+    /// <summary>
+    /// Asynchronously releases resources owned by the derived channel.
+    /// </summary>
+    /// <returns>A task representing the asynchronous disposal operation.</returns>
+    protected abstract Task DisposeAsyncCore();
```
jviau: Recommend using `ValueTask` here and not `Task`.

jviau: nit: keep `Async` at the end
```suggestion
    protected abstract Task DisposeCoreAsync();
```

src/WebJobs.Script.Grpc/Channel/ServerDuplexChannelExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+namespace Microsoft.Azure.WebJobs.Script.Grpc;
+
+internal static class ServerDuplexChannelExtensions
+{
+    internal static ServerDuplexChannel AddServerDuplexChannel(this IScriptEventManager eventManager, string workerId)
+    {
+        var channel = new ServerDuplexChannel();
+        if (!eventManager.TryAddWorkerState(workerId, channel))
+        {
+            channel.DisposeAsync().GetAwaiter().GetResult();
```
jviau: All of the disposals of this new type are synchronous. Should it just implement `IDisposable`?

Actually, this type doesn't even _need_ disposal. `Channel<T>` is not disposable. Completing the writer is not a disposal contract: it is to inform writers you cannot write anymore. In this method there is no writer yet. So just letting the channel be garbage collected (and no explicit disposal) is just fine,

src/WebJobs.Script.Grpc/Channel/ServerDuplexChannelExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        eventManager.TryRemoveWorkerState(workerId, out ServerDuplexChannel removedChannel);
+        removedChannel?.DisposeAsync().GetAwaiter().GetResult();
+        if (!ReferenceEquals(removedChannel, channel))
+        {
+            channel.DisposeAsync().GetAwaiter().GetResult();
+        }
+
+        throw new InvalidOperationException("Could not retrieve server duplex channel for worker ID: " + workerId);
```
jviau: This seems odd for an `Add*` method to throw saying "could not __retrieve__...". Should this say it could not _add_?

src/WebJobs.Script.Grpc/Channel/ServerDuplexChannelExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+    internal static ServerDuplexChannel AddServerDuplexChannel(this IScriptEventManager eventManager, string workerId)
+    {
+        var channel = new ServerDuplexChannel();
+        if (!eventManager.TryAddWorkerState(workerId, channel))
+        {
+            channel.DisposeAsync().GetAwaiter().GetResult();
+            throw new ArgumentException("Duplicate worker id: " + workerId, nameof(workerId));
+        }
+
+        if (eventManager.TryGetWorkerState(workerId, out ServerDuplexChannel registeredChannel))
```
jviau: What scenario are these extra checks guarding against? It looks like it is trying to ensure expected state, but these assertions only hold up for this exact point in time. The underlying `eventManager` could be modified immediately after this and these assertions no longer hold.

Worker state is a `ConcurrentDictionary` in the end. I think the best option would be to expose `GetOrAdd` semantics and let concurrent dictionary handle this consistency guarantees.

src/WebJobs.Script.Grpc/Channel/ServerDuplexChannelExtensions.cs
brettsam: In Client this will need to be AsyncDisposable. GrpcDuplexChannel will derive from this so we can pass it in here directly.

src/WebJobs.Script.Grpc/Channel/DuplexChannel.cs
brettsam: DisposeAsyncCore is the recommended name: https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync#explore-disposeasync-and-disposeasynccore-methods

src/WebJobs.Script.Grpc/Channel/ServerDuplexChannelExtensions.cs
brettsam: well, i was trying not to make too many changes (like removing the workerstate concept from IScriptEventManager), but your last two comments makes me think it may be clearer if i did that anyway right now... let me follow up with a change that should improve this.

src/WebJobs.Script.Grpc/Channel/ServerDuplexChannelExtensions.cs
brettsam: latest commit removes this...

brettsam:
My latest commit effectively does the "Deferred change" that is in the PR description because I think it's clearer our intent.

I replaced the generic worker-state storage on `IScriptEventManager` with a dedicated `ServerDuplexChannelRegistry`. The event manager APIs were only supporting this RPC plumbing, and adapting them further would make an unrelated pub/sub abstraction responsible for channel registration and lifetime.

The registry now provides atomic worker-ID registration and lookup, exposes only borrowed endpoints to `FunctionRpcService`, and returns an owned channel registration to `WorkerChannel`. Disposing that registration unregisters it and closes the underlying channel. This makes ownership explicit and removes the add-then-read consistency logic discussed here.

My next PR will build on this by having the client GrpcDuplexChannel derive from `DuplexChannel<T>` and pass itself directly into the `WorkerChannel`. It has more substantial cleanup to do on `DisposeAsync()`

src/WebJobs.Script.Grpc/Channel/FunctionRpcServiceEndpoints.cs
```diff
@@ -0,0 +1,19 @@
...
+
+/// <summary>
+/// Provides the borrowed channel endpoints used by <c>FunctionRpcService</c> to bridge a worker's gRPC stream.
+/// </summary>
+/// <param name="HostToWorkerReader">Reads host messages that should be written to the worker.</param>
+/// <param name="WorkerToHostWriter">Writes messages received from the worker for host processing.</param>
+/// <remarks>
+/// This value does not own the channel lifetime. The associated <see cref="ServerDuplexChannel"/> owns both endpoints.
+/// </remarks>
+internal readonly record struct FunctionRpcServiceEndpoints(
```
kshyju: nit: CP thinks `FunctionRpcChannelEndpoints` is more readable name for this type.

src/WebJobs.Script/Eventing/IScriptEventManager.cs
```diff
@@ -8,11 +8,5 @@ namespace Microsoft.Azure.WebJobs.Script.Eventing
     public interface IScriptEventManager : IObservable<ScriptEvent>
     {
         void Publish(ScriptEvent scriptEvent);
-
-        bool TryAddWorkerState<T>(string workerId, T state);
```
kshyju, line 12: Do we have any other implementation of this outside this repo? (ex: webjobs or so..)? 
If it's only inside our repo, wonder we should mark it as internal?

kshyju (approved):

src/WebJobs.Script.Grpc/Channel/FunctionRpcServiceEndpoints.cs
brettsam: Fixed.

src/WebJobs.Script/Eventing/IScriptEventManager.cs
brettsam: Can't be internal b/c it's used in a bunch of public constructors. I'd love to clean that up at some point (and move away from this thing completely), but for now, we need it.

The WorkerState stuff was never used anywhere but for this one, thing -- stashing channels to retrieve later. Making it a concrete Registry means we can get rid of this behavior.

src/Functions.Rpc.Server/Channel/ServerDuplexChannelRegistry.cs
```diff
@@ -0,0 +1,92 @@
...
+    private readonly ConcurrentDictionary<string, Registration> _registrations = new(StringComparer.Ordinal);
+
+    /// <summary>
+    /// Creates and registers a channel whose ownership transfers to the caller.
+    /// </summary>
+    /// <param name="workerId">The worker identifier used to resolve the service endpoints.</param>
+    /// <returns>
+    /// The registered channel. Disposing it unregisters the worker and disposes the underlying server channel.
+    /// </returns>
+    internal DuplexChannel<StreamingMessage> CreateRegisteredChannel(string workerId)
```
jviau: Could we have get-or-add semantics? It would be preferrable to throwing if two callers create the same channel.

Speaking of which: is it expect for two callers to create the same workerId? Or is it meant that only ever 1 caller can hold onto a channel at a time. In which case, we should use the "Lease" terminology around here somehow. That implies assigned ownership.

src/WebJobs.Script/Eventing/IScriptEventManager.cs
jviau: These assemblies are not treated as libraries. We do not publicly distribute them. Marking types as `public` vs `internal` is no real difference.

The Script.Abstractions assembly is the exception: we _do_ publish that one as a library.

jviau (approved):

src/Functions.Rpc.Server/Channel/ServerDuplexChannelRegistry.cs
brettsam: We don't want GetOrAdd for this (or at least we've never had it). If a second worker with the same id tries to create channels, something has gone wrong and we should throw. [The behavior today is to throw](https://github.com/Azure/azure-functions-host/blob/a055cb66b8805f6d78f6c75b9921491c3f878956/src/WebJobs.Script.Grpc/Eventing/GrpcEventExtensions.cs#L48-L49) so I wanted to maintain that. But in reality, I don't think this ever happens.

Lease is a better name, I like that. I'll make that change.

kshyju (approved):

Issue #11984, opened 2026-09-02
[Compute Separation][HT7] Add the behavior-neutral standard composition seam


brettsam:
## Roadmap mapping

| Field | Value |
|---|---|
| Roadmap ID | `HT7` |
| Lane | Host |
| BYOC milestone | M1 |
| Parent | #11965 |
| Depends on | `HT1` / #11968 |
| Blocks | `HT8a` |
| Parallel with | HT4/HT5/HT6 and WorkerProxy work |
| Pull request | #11988 |

## Why this PR exists

Server-backed and Client-backed workers must remain separate product compositions over the shared FunctionRpc core. The standard WebHost must continue to reference Rpc.Server and must never acquire an Rpc.Client, compute-product, or WorkerProxy reference.

Today worker registrations are spread across WebHost and Rpc.Server and sometimes rely on `TryAdd`. Adding Client services on top would make worker selection depend on registration order, which the GA plan forbids. This PR introduces an explicit worker-composition seam, preserves the Server graph, and adds the dedicated Linux Host executable shell now so later PRs populate an already-visible Client path instead of creating another entrypoint.

## Source material

- Roadmap HT7; incremental roadmap standard composition milestone.
- `ga-plan.md`: standard WebHost, standard lifecycle, nested DI ownership, and prohibition on last-registration-wins.
- Static-entrypoint proposal in the compute-separation architecture walkthrough.
- Current `WebHostServiceCollectionExtensions.cs`, `WebScriptHostBuilderExtension.cs`, `Functions.Rpc.Server/GrpcServiceCollectionsExtensions.cs`, and `Hosting/RpcServiceCollectionExtensions.cs`.
- Existing dependency, startup, and lifecycle tests.

## Current state

Root registrations include `AddScriptGrpc`, `AddCommonRpcServices`, `WebJobsScriptHostService`, and its hosted-service/manager aliases. ScriptHost-child registrations come from `AddRpcScriptHostServices`. Some implementations use `TryAdd`. The only executable Host is the standard WebHost.

## Proposed implementation

1. Add `IWorkerComposition` with explicit root-WebHost and child-ScriptHost methods.
2. Add `ServerWorkerComposition` and move today's Server registration calls into it without changing order, lifetime, implementation kind, startup, or shutdown behavior.
3. Add `SelectedWorkerComposition` so the same explicitly selected composition configures the root and every restarted child ScriptHost; unrelated `IWorkerComposition` DI registrations cannot replace it.
[...]

PR #11988, opened 2026-09-02
adding composition extensibility for host

brettsam:
Completes #11984

Adds an explicit worker-composition seam so worker architecture is selected statically by each executable rather than through runtime flags or DI replacement ordering.

- Adds `IWorkerComposition`.
- Adds the existing Server-backed behavior as `ServerWorkerComposition`.
- Adds an intentionally incomplete `ClientWorkerComposition`.
- Moves process initialization, host construction, startup, shutdown, and disposal into `FunctionsHost`.
- Adds the Linux-only `Azure.Functions.Host` executable.
- Keeps Rpc.Client and WorkerProxy out of the standard WebHost dependency graph.

## Why

Rpc.Server and Rpc.Client should remain sibling implementations over the shared gRPC contracts. The existing WebHost must continue using Rpc.Server without referencing incomplete Client functionality.

Each executable now selects its worker composition explicitly:

- `Microsoft.Azure.WebJobs.Script.WebHost` selects `ServerWorkerComposition`.
- `Azure.Functions.Host` selects `ClientWorkerComposition`.

This avoids:

- runtime topology flags
- last-registration-wins service replacement
- accidentally shipping Rpc.Client in the standard WebHost
- mixing Server and Client worker lifecycles in one process

`SelectedWorkerComposition` preserves the root selection for subsequent child ScriptHost builds and restarts. It uses an internal DI key so unrelated `IWorkerComposition` registrations cannot silently replace the selected worker architecture.

## Design

`IWorkerComposition` controls the worker-specific registration points:

- root WebHost worker services, MVC application parts, and ScriptHost activation
- child ScriptHost worker services

`ServerWorkerComposition` retains the current behavior:

- inbound FunctionRpc server
- local worker process and channel management
- process-backed metadata
- standard invocation dispatch
- automatic ScriptHost startup

`FunctionsHost` now owns the shared executable flow:

- process initialization
- host construction and configuration
- startup
- shutdown handling
- host disposal

Both product entrypoints call `FunctionsHost.RunAsync` with their selected composition.

## Azure.Functions.Host

Adds a Linux-only `Azure.Functions.Host` executable to make the future Client-backed flow explicit.

Its `ClientWorkerComposition` currently throws `NotImplementedException`. This is intentional: the assembly and static composition path exist, but no incomplete Client behavior is registered or runnable.

The Linux build:

- supports only `linux-x64`
- uses the ASP.NET shared framework
- preserves the existing runtime and ReadyToRun settings
- publishes one extensionless `Azure.Functions.Host` apphost
- excludes `web.config` and referenced standard WebHost entrypoint artifacts

## Deliberate limitations

This PR does not implement the Client-backed runtime.

`Azure.Functions.Host` currently references WebHost and therefore receives Rpc.Server transitively. A future `WebHost.Core` extraction can remove that dependency, but that larger assembly/package split is not required for this behavior-neutral seam.

Future work will add Client-backed metadata, dispatch, deferred ScriptHost startup, configuration delivery, lifecycle management, and compute-specific controllers exclusively through `Azure.Functions.Host`.

JIT/PreJIT changes are also deferred until the Client composition becomes runnable.

### Pull request checklist

### DIFF
--- a/src/Functions.Host/Functions.Host.csproj
+++ b/src/Functions.Host/Functions.Host.csproj
@@ -1,0 +1,56 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="$(WorkersProps)" />
+
+  <PropertyGroup>
+    <TargetFramework>net10.0</TargetFramework>
+    <OutputType>Exe</OutputType>
+    <RuntimeIdentifiers>linux-x64</RuntimeIdentifiers>
+    <PackageId>Azure.Functions.Host</PackageId>
+    <AssemblyName>Azure.Functions.Host</AssemblyName>
+    <RootNamespace>Azure.Functions.Host</RootNamespace>
+    <Nullable>enable</Nullable>
+    <IsPackable>false</IsPackable>
+    <ServerGarbageCollection>true</ServerGarbageCollection>
+    <TieredCompilation>false</TieredCompilation>
+    <UserRuntimeConfig>$(MSBuildThisFileDirectory)..\WebJobs.Script.WebHost\runtimeconfig.template.json</UserRuntimeConfig>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
+  </ItemGroup>
+
+  <PropertyGroup Condition="'$(RuntimeIdentifier)' != ''">
+    <PublishReadyToRun>true</PublishReadyToRun>
+    <PublishReadyToRunUseCrossgen2>true</PublishReadyToRunUseCrossgen2>
+    <PublishReadyToRunComposite>true</PublishReadyToRunComposite>
+    <PublishReadyToRunEmitSymbols>true</PublishReadyToRunEmitSymbols>
+    <PublishReadyToRunShowWarnings>true</PublishReadyToRunShowWarnings>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <!-- This interim reference makes the shared FunctionsHost entrypoint explicit. A later project split removes the
+         transitive Rpc.Server implementation from the compute publish graph. -->
+    <ProjectReference Include="..\WebJobs.Script.WebHost\WebJobs.Script.WebHost.csproj" />
+  </ItemGroup>
+
+  <!-- WebHost is referenced for its shared implementation, not as a second executable or IIS artifact. -->
+  <Target Name="RemoveStandardWebHostEntrypointFromPublish" AfterTargets="ComputeFilesToPublish">
+    <ItemGroup>
+      <ResolvedFileToPublish
+        Remove="@(ResolvedFileToPublish)"
+        Condition="'%(RelativePath)' == 'Microsoft.Azure.WebJobs.Script.WebHost' Or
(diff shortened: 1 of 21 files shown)


src/Functions.Host/Functions.Host.csproj
```diff
@@ -0,0 +1,56 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="$(WorkersProps)" />
+
+  <PropertyGroup>
+    <TargetFramework>net10.0</TargetFramework>
+    <OutputType>Exe</OutputType>
+    <RuntimeIdentifiers>linux-x64</RuntimeIdentifiers>
```
jviau, line 7: Will this impact running locally on windows machines?

src/Functions.Host/Program.cs
jviau, line 1: Consider using the minimal entry point `Program.cs`

``` csharp
// Copyright (c) .NET Foundation. All rights reserved.

using Microsoft.Azure.WebJobs.Script.WebHost;

await FunctionsHost.RunAsync(args, ClientWorkerComposition.Instance);
```

src/WebJobs.Script.WebHost/FunctionsHost.cs
```diff
@@ -0,0 +1,265 @@
...
+        ArgumentNullException.ThrowIfNull(composition);
+
+        InitializeProcess();
+        IHost host = BuildHost(args, composition);
+        return RunHostAsync(host);
+    }
+
+    internal static IHost BuildHost(string[] args, IWorkerComposition composition)
+    {
+        return CreateHostBuilder(args, composition)
```
jviau: Can be less aggressive with newlines

```suggestion
        return CreateHostBuilder(args, composition).Build();
```

src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
```diff
@@ -37,6 +39,37 @@ public static class WebScriptHostBuilderExtension
         public static IHostBuilder AddWebScriptHost(this IHostBuilder builder, IServiceProvider rootServiceProvider, IServiceCollection rootServices,
            ScriptApplicationHostOptions webHostOptions, Action<IWebJobsBuilder> configureWebJobs = null)
         {
+            IWorkerComposition composition = rootServiceProvider?.GetService<SelectedWorkerComposition>()?.Value
```
jviau: Is the `SelectedWorkerComposition` wrapper needed? Or could `IWorkerComposition` be directly registered?

jviau (approved):

src/WebJobs.Script.WebHost/WebScriptHostBuilderExtension.cs
brettsam: I thought the same thing. The agent was nervous that something could register an IWorkerComposition in ScriptHost and thus have a mismatch between Web and ScriptHost... capturing it like this in WebHost makes that harder.

I could go either way. It obviously confused both of us, so maybe just remove it?

src/Functions.Host/Functions.Host.csproj
brettsam: I was able to run from VS with this. I think it's just for publishing?

kshyju (approved):

PR #11990, opened 2026-09-02
refactor(worker-proxy): route endpoints by listener port

jviau:
### Issue describing the changes in this PR

N/A — stack 1/2 for the WorkerProxy HTTP forwarding work.

### Pull request checklist

**IMPORTANT**: Currently, changes must be backported to the `in-proc` branch to be included in Core Tools and non-Flex deployments.

* [x] Backporting to the `in-proc` branch is not required
    * Otherwise: Link to backporting PR
* [x] My changes **do not** require documentation changes
    * [ ] Otherwise: Documentation issue linked to PR
* [x] My changes **should not** be added to the release notes for the next release
    * [ ] Otherwise: I've added my notes to `release_notes.md`
* [x] My changes **do not** need to be backported to a previous version
    * [ ] Otherwise: Backport tracked by issue/PR #issue_or_pr
* [x] My changes **do not** require diagnostic events changes
    * Otherwise: I have added/updated all related diagnostic events and their documentation (Documentation issue linked to PR)
* [x] I have added all required tests (Unit tests, E2E tests)

### Additional information

This bottom PR refactors `WorkerProxyApplication` to select management and FunctionRpc pipelines by listener port using `MapWhen`:

- maps `GET /admin/instance/ready` only on the management pipeline
- maps `FunctionRpcRelayService` through one pipeline shared by both existing HTTP/2 RPC listeners
- returns 404 for unmatched traffic and removes redundant inline listener guards

This PR contains no HTTP proxy changes and does not add YARP, HTTP forwarding/readiness infrastructure, new listener ports, or Docker port configuration.

Validation:

- `dotnet test test\Functions.WorkerProxy.Tests\Functions.WorkerProxy.Tests.csproj --configuration Release --warnaserror --no-restore --nologo` (27 passed)
- Windows x64 Native AOT publish of `src\Functions.WorkerProxy\Functions.WorkerProxy.csproj` with warnings as errors

PR #11991, opened 2026-09-02
feat(worker-proxy): add HTTP proxy listener

jviau:

N/A. This is stack **2/2** and depends on #11990.

### Pull request checklist

### Additional information

Adds a dedicated WorkerProxy HTTP/1 listener on port 28080, including Docker exposure and launch configuration. Non-admin HTTP traffic is forwarded to the worker through YARP direct forwarding with destination resolution and TCP readiness probing. Management remains readiness-only, every `/admin/*` request is isolated from forwarding, and `RuntimeGrpcPort` and `WorkerGrpcPort` remain in the shared gRPC-only pipeline.

Validation:

* `dotnet test test\Functions.WorkerProxy.Tests\Functions.WorkerProxy.Tests.csproj --configuration Release --no-restore -p:TreatWarningsAsErrors=true` — 50 passed
* `dotnet publish src\Functions.WorkerProxy\Functions.WorkerProxy.csproj --configuration Release --runtime win-x64 --no-restore -p:TreatWarningsAsErrors=true` — Native AOT publish succeeded


src/Functions.WorkerProxy/Http/WorkerHttpForwardingMiddleware.cs
```diff
@@ -0,0 +1,41 @@
...
+    WorkerHttpForwarder forwarder)
+{
+    /// <summary>
+    /// Forwards a request to the configured worker HTTP endpoint.
+    /// </summary>
+    public async Task InvokeAsync(HttpContext context)
+    {
+        System.Uri? destination = WorkerHttpDestinationResolver.Resolve(
+            options.Value.WorkerHttpEndpoint, advertisedEndpoint: null);
+
```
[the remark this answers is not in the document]
jviau: WorkerProxy's FunctionRpc relay does not expose the worker-advertised HTTP URI yet. I added a TODO in `e1cf8733b` at this call site; until that relay surface is available, `WorkerHttpEndpoint` is the supported explicit destination.

src/Functions.WorkerProxy/Http/WorkerEndpointReadinessProbe.cs
```diff
@@ -0,0 +1,185 @@
...
+        Stopwatch stopwatch = Stopwatch.StartNew();
+        int attempts = 0;
+        Exception? lastError = null;
+
+        IPAddress[] addresses;
+        if (string.Equals(destination.Host, "localhost", StringComparison.OrdinalIgnoreCase))
+        {
+            addresses = [IPAddress.Loopback];
+        }
+        else
```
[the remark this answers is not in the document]
jviau: Fixed in `e1cf8733b`. `localhost` readiness probing now tries both IPv4 and IPv6 loopback addresses, and the new test verifies an IPv6-only loopback listener is detected.

test/Functions.WorkerProxy.Tests/Functions.WorkerProxy.Tests.csproj
```diff
@@ -23,6 +23,7 @@
     <PackageReference Include="Microsoft.NET.Test.Sdk" PrivateAssets="all" />
     <PackageReference Include="xunit" />
     <PackageReference Include="xunit.runner.visualstudio" PrivateAssets="all" />
+    <PackageReference Include="Yarp.ReverseProxy" VersionOverride="2.3.0" PrivateAssets="all" />
```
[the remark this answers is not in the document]
jviau: Keeping this reference intentionally. This stack requires YARP 2.3.0 scoped project overrides in both the product and test projects to match the source implementation; `PrivateAssets="all"` keeps the test-only direct pin from flowing transitively.

src/Functions.WorkerProxy/Http/WorkerHttpForwardingMiddleware.cs
```diff
@@ -0,0 +1,43 @@
...
+        {
+            context.Response.StatusCode = StatusCodes.Status503ServiceUnavailable;
+            return;
+        }
+
+        ForwarderError error = await forwarder.ForwardAsync(context, destination);
+        if (error is not ForwarderError.None && !context.RequestAborted.IsCancellationRequested
+            && !context.Response.HasStarted)
+        {
+            context.Response.StatusCode = StatusCodes.Status502BadGateway;
```
brettsam, line 40: is this masking whatever the actual underlying error may be?

src/Functions.WorkerProxy/WorkerProxyOptions.cs
```diff
@@ -27,4 +27,14 @@ internal sealed class WorkerProxyOptions
...
+
+    /// <summary>
+    /// Gets or sets the runtime-facing HTTP/1 forwarding listener port.
+    /// </summary>
+    public int HttpPort { get; set; } = 28080;
+
+    /// <summary>
+    /// Gets or sets an optional explicit worker HTTP destination.
+    /// </summary>
+    public string? WorkerHttpEndpoint { get; set; }
```
brettsam: Not sure we'll ever know the WorkerHttpEndpoint at startup -- the worker will tell us where it's listening via its capabilities in grpc like it does today. So we'll need to intercept it and set it somewhere... 

Don't think this has to be all resolved in this change, but just wanted to raise that.

src/Functions.WorkerProxy/Http/WorkerEndpointReadinessProbe.cs
```diff
@@ -0,0 +1,182 @@
...
+    /// Waits for a destination to accept a TCP connection within the readiness deadline.
+    /// </summary>
+    /// <param name="destination">The destination to probe.</param>
+    /// <param name="cancellationToken">The caller cancellation token.</param>
+    /// <returns><see langword="true"/> when ready; otherwise, <see langword="false"/>.</returns>
+    public async ValueTask<bool> WaitForReadyAsync(Uri destination, CancellationToken cancellationToken)
+    {
+        ArgumentNullException.ThrowIfNull(destination);
+
+        if (_readyDestinations.ContainsKey(destination))
```
brettsam, line 55: If a worker crashes or restarts for some reason and re-connects, we'll need to clear that destination, right? 

I'm fine with deferring that, but wanted to call it out.

src/Functions.WorkerProxy/Http/WorkerEndpointReadinessProbe.cs
```diff
@@ -0,0 +1,182 @@
...
+        return _readyDestinations.ContainsKey(destination);
+    }
+
+    /// <summary>
+    /// Waits for a destination to accept a TCP connection within the readiness deadline.
+    /// </summary>
+    /// <param name="destination">The destination to probe.</param>
+    /// <param name="cancellationToken">The caller cancellation token.</param>
+    /// <returns><see langword="true"/> when ready; otherwise, <see langword="false"/>.</returns>
+    public async ValueTask<bool> WaitForReadyAsync(Uri destination, CancellationToken cancellationToken)
```
brettsam: If the worker is the one telling us its http uri to connect to (b/c it's already up and listening), do we have to be probing for ready? It's nice to ensure that things are actually running, but not sure if in practice we'll be in a situation where the worker isn't already listening when we find out the uri we need to forward to.

src/Functions.WorkerProxy/Http/WorkerHttpForwardingMiddleware.cs
```diff
@@ -0,0 +1,43 @@
...
+    IOptions<WorkerProxyOptions> options,
+    WorkerEndpointReadinessProbe readinessProbe,
+    WorkerHttpForwarder forwarder)
+{
+    /// <summary>
+    /// Forwards a request to the configured worker HTTP endpoint.
+    /// </summary>
+    public async Task InvokeAsync(HttpContext context)
+    {
+        // TODO: Supply the worker-advertised HTTP endpoint once FunctionRpc exposes it.
```
brettsam, line 25: Thoughts on how this will be exposed here? Is that in scope for this change, or should we file a quick follow-up if you're not able to get to it before you're gone?

brettsam (approved):
Some comments, but I'm good getting this in if you don't think they need addressing right now.

src/Functions.WorkerProxy/Http/WorkerHttpForwardingMiddleware.cs
jviau: I think it would be basically what we have in the feature branch. This depends on the gRPC init handshake with the worker to get the capabilities from it.

src/Functions.WorkerProxy/Http/WorkerHttpForwardingMiddleware.cs
jviau: 502 is the right error to use here - it indicates a failure in the proxy (and not the end service). But I think it could use some better telemetry to help us diagnose when this is an issue. I will add another PR for that.

src/Functions.WorkerProxy/Http/WorkerEndpointReadinessProbe.cs
jviau: good point. I'll work on that in another PR

kshyju (approved):
GHCP flagged this:

>HTTP forwarding is capped at Kestrel’s default 30,000,000-byte body limit, which breaks larger requests and surfaces as `502`

jviau:
> GHCP flagged this:
> 
> > HTTP forwarding is capped at Kestrel’s default 30,000,000-byte body limit, which breaks larger requests and surfaces as `502`

will follow up in another PR

PR #11964, opened 2026-09-01
introducing RpcClientWorkerChannel

brettsam:
Adds the client-backed `RpcClientWorkerChannel` used when the Functions runtime connects outbound to a worker endpoint supplied through `/link`. It reuses the existing shared `WorkerChannel` protocol while replacing the local-process/server channel with an outbound FunctionRpc client channel.

This PR adds the underlying channel and factories. The `/link` endpoint, channel registry, and dispatcher wiring will be added in later PRs.

Note there's a bunch of added doc comments here b/c WorkerChannel is now public to allow us to derive from it. That's a bunch of the changes you'll see...

## Changes

- Adds `RpcClientWorkerChannel : WorkerChannel` in `Azure.Functions.Rpc.Client`.
- Adds an internal, DI-ready `RpcClientWorkerChannelFactory`.
- Makes `WorkerChannel` publicly derivable while keeping server-only hooks internal.
- Changes `GrpcDuplexChannel<T>` to derive from shared `DuplexChannel<T>` and use its idempotent disposal implementation.
- Updates duplex-channel factories to return `DuplexChannel<T>`, making ownership explicit.
- Makes repeated and concurrent `StartAsync` calls share one initialization attempt.
- Keeps caller cancellation scoped to that caller’s wait.
- Preserves channel exceptions observed before initialization.
- Treats unexpected clean channel closure before initialization as startup failure.
- Preserves explicit disposal as cancellation rather than unexpected failure.
- Keeps the Client → Grpc/Script dependency direction; Grpc, Server, and WebHost do not reference Client.

**IMPORTANT**: Currently, changes must be backported to the `in-proc` branch to be included in Core Tools and non-Flex deployments.

* [x] Backporting to the `in-proc` branch is not required
    * Otherwise: Link to backporting PR
* [x] My changes **do not** require documentation changes
    * [ ] Otherwise: Documentation issue linked to PR
* [x] My changes **should not** be added to the release notes for the next release
    * [ ] Otherwise: I've added my notes to `release_notes.md`
* [x] My changes **do not** need to be backported to a previous version
    * [ ] Otherwise: Backport tracked by issue/PR #issue_or_pr
* [x] My changes **do not** require diagnostic events changes
    * Otherwise: I have added/updated all related diagnostic events and their documentation (Documentation issue linked to PR)
* [x] I have added all required tests (Unit tests, E2E tests)

Closes #11970

PR #11994, opened 2026-09-03
adding ChannelRegistry

### LINKED ISSUE #11979: [Compute Separation][HT4] Add the Client channel registry and link primitive
## Roadmap mapping

| Field | Value |
|---|---|
| Roadmap ID | `HT4` |
| Lane | Host |
| BYOC milestone | M1 |
| Parent | #11965 |
| Depends on | `HT3` / #11970 |
| Blocks | `HT5`, `HT6`, `HT9` |
| Pull request | One PR |

## Why this PR exists

HT3 provides one manually constructed client-backed channel. Nothing owns multiple outbound channel lifetimes, records a linked worker, waits for the first ready channel, or disposes a failed channel. This PR adds one behavior: Client-owned channel lifetime keyed by worker ID. Keeping it separate makes review about concurrency and ownership rather than platform HTTP contracts.

## Source material

- `Functions-NoPlaceholders@brettsam/ga_plan/docs/host/ga/compute-separation-pr-roadmap.md`, Milestone 4.
- `docs/host/ga/ga-plan.md`, Rpc.Client registry/lifecycle and nested-DI ownership.
- `docs/host/ga/roadmap.md`, HT4.
- Current `src/Functions.Rpc.Client` plus HT3 PR #11964.
- Reference-only prototype `ConnectedWorkerChannelManager` and `WorkerConnectionService`; do not port their placement, event coupling, or public platform shape.

## Current state

`RpcClientFactory` and duplex factories exist; HT3 adds `RpcClientWorkerChannel` and factory. Still missing: deterministic duplicate/relink behavior, ready filtering, cancellation-aware first-ready wait, automatic failure removal, and one shutdown owner.

## Proposed implementation

1. Add a narrow Client-owned registry contract with `LinkAsync`, `UnlinkAsync`, ready lookup/enumeration, `WaitForFirstReadyAsync`, and `IAsyncDisposable`.
2. Add transport-only `WorkerLinkContext`: worker ID, gRPC URI, optional connection material, and the worker config needed by the channel factory. No pod/platform DTO or HTTP endpoint.
3. Implement `WorkerChannelRegistry` with an ordinal worker-ID dictionary and per-worker serialization. Different workers may link concurrently; link/unlink for one ID must serialize. Use `System.Threading.Lock`, `SemaphoreSlim`, and/or `Interlocked` deliberately—never an `object` lock.
4. `LinkAsync`: connect duplex transport, create/start `RpcClientWorkerChannel`, publish the entry only after successful initialization. On any failure remove the entry, dispose every owned resource exactly once, and surface a typed result/exception.
5. Reject duplicate healthy links deterministically. Relink succeeds only after failed/terminal entry removal.
6. Subscribe to channel completion and remove/dispose that entry exactly once without disturbing other channels.
7. Implement ready filtering and cancellation-aware multi-waiter first-ready signaling without polling.
8. `DisposeAsync` stops new links, releases waiters, and disposes every owned channel exactly once; repeated disposal is safe.
9. Add a test-only Client registration helper only. Product composition waits for HT8a.

### Ownership and lifecycle

The registry solely owns channels it creates. `RpcClientFactory` owns cached gRPC clients/channels and outlives the registry. The registry is intended as a compute root-container singleton surviving ScriptHost restarts; consumers borrow channels and never dispose them.

Resolve in review whether ready-channel APIs expose public `WorkerChannel`, a narrow Client abstraction, or a revised shared interface. Do not casually widen product friend access; record any narrow seam and pre-GA disposition.

## Tests and evidence

Under `Functions.Rpc.Client.Tests`:

- concurrent links for different IDs and serialized link/unlink for one ID;
- duplicate link rejected, relink after failure succeeds;
- failed/completed transport removes and disposes only that entry;
- ready filtering excludes uninitialized/failed/disposed channels;
- multiple first-ready waiters complete and cancellation leaks nothing;
- shutdown and repeated disposal dispose each channel once and reject new links;
- Rpc.Client still has no Server reference.

Post HT14 dependency, standard service-graph, standard publish/deps, startup, and topology-flag evidence.

## Explicit exclusions

No HTTP/admin endpoint or platform DTO, dispatcher, metadata provider, composition, ScriptHost startup, configuration/host.json, capacity/slots/state, retry/reconnect/drain/eviction policy, auth/security, WorkerProxy reference, multi-worker scheduling, billing, extension gRPC, or assembly streaming.

## Acceptance criteria

- Several fake workers link concurrently with deterministic lifecycle.
- Duplicate/relink rules and failure cleanup are explicit.
- First-ready wait is cancellable and non-polling.
- One channel failure leaves others healthy.
- Shutdown disposes every owned channel once.
- Registry remains test-only and compute separation is not yet functional end to end.

## PR stop condition

Stop when registry, transport link input, ready/wait primitives, concurrency/failure/disposal tests, and dependency guards exist in Rpc.Client but remain reachable only from tests. Do not add dispatcher, metadata, endpoint, composition, startup, or retry policy.

brettsam:
Adds the Client-owned worker channel registry and explicit link lifecycle for compute separation. This builds on the client-backed `WorkerChannel` from #11964 without wiring Client services into production Host composition.

### Changes

- Adds `WorkerChannelRegistry`, keyed by worker ID.
- Adds `LinkAsync`, which connects the outbound transport and returns after `StartStream` and a successful `WorkerInitResponse`.
- Supports concurrent links for different workers while serializing link, unlink, completion cleanup, and shutdown for the same worker.
- Rejects duplicate links deterministically and allows relink after cleanup completes.
- Adds initialized-channel lookup, enumeration, and cancellation-aware waiting for the first initialized channel.
- Automatically removes and disposes channels when their outbound transport completes or fails.
- Adds idempotent unlink and registry shutdown with exactly-once channel disposal.
- Simplifies `RpcClientWorkerChannelFactory` to create the fixed M1 worker configuration internally.
- Adds a test-only Client service-registration helper. No production composition consumes it yet.

An initialized channel has completed the FunctionRpc initialization handshake, but it is not yet invocation-ready. Metadata retrieval, invocation-buffer setup, function loading, ScriptHost startup, HTTP/admin endpoints, retry policy, and capacity publication remain separate follow-up work.

Resolves #11979

### Pull request checklist


jviau (approved):

src/Functions.Rpc.Client/WorkerChannelRegistry.cs
```diff
@@ -0,0 +1,481 @@
...
+
+    public WorkerChannelRegistry(IDuplexChannelFactory<StreamingMessage> duplexChannelFactory,
+        IRpcClientWorkerChannelFactory channelFactory, ILogger<WorkerChannelRegistry> logger)
+    {
+        _duplexChannelFactory = duplexChannelFactory ?? throw new ArgumentNullException(nameof(duplexChannelFactory));
+        _channelFactory = channelFactory ?? throw new ArgumentNullException(nameof(channelFactory));
+        _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+    }
+
+    private bool IsDisposed => Interlocked.CompareExchange(ref _disposed, 0, 0) != 0;
```
jviau: Why is this performing a compare exchange? This doesn't seem right. Shouldn't it just be a read?

src/Functions.Rpc.Client/WorkerChannelRegistry.cs
```diff
@@ -0,0 +1,481 @@
...
+        {
+            if (IsDisposed)
+            {
+                return [];
+            }
+
+            return _slots.Values
+                .Where(slot => slot.Channel is not null)
+                .Select(slot => (WorkerChannel)slot.Channel)
+                .ToArray();
```
jviau: tiny nit: use collection initializer to let the compiler determine the best way to realize the right type:

``` csharp
return [.. _slots.Values.Where(slot => slot.Channel is not null).Select(slot => (WorkerChannel)slot.Channel)];
```

src/Functions.Rpc.Client/WorkerChannelRegistry.cs
```diff
@@ -0,0 +1,481 @@
...
+    {
+        if (Interlocked.CompareExchange(ref _disposed, 1, 0) == 0)
+        {
+            _ = CompleteDisposalAsync();
+        }
+
+        return new ValueTask(_disposeCompletion.Task);
+    }
+
+    private static Exception CombineExceptions(Exception currentException, Exception nextException)
```
jviau: nit: use extension methods for common helpers like this.

``` csharp
public static class ExceptionExtensions
{
    extension (AggregateException)
    {
        public static Exception Combine(Exception first, Exception second)
        {
        }
    }
}
```

src/Functions.Rpc.Client/WorkerChannelRegistry.cs
```diff
@@ -0,0 +1,481 @@
...
+        => currentException is null ? nextException : new AggregateException(currentException, nextException);
+
+    private static async Task<Exception> CaptureCleanupFailureAsync(Exception currentException, IAsyncDisposable disposable)
```
jviau: I think I have seen this in a couple of your PRs now. Could also be an extension method off of `IAsyncDisposable`

``` csharp
public static class DisposableExtensions
{
    extension (IAsyncDisposable disposable)
    {
        public async ValueTask<Exception> CleanupAsync(Exception ex)
        {
            ...
        }
    }
}
```

src/Functions.Rpc.Client/WorkerChannelRegistry.cs
```diff
@@ -0,0 +1,481 @@
...
+        cancellationToken.ThrowIfCancellationRequested();
+
+        WorkerSlot slot = ReserveLinkSlot(workerId);
+        bool gateEntered = false;
+
+        try
+        {
+            using CancellationTokenSource operationSource =
+                CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, _shutdownSource.Token);
+            await slot.Gate.WaitAsync(operationSource.Token);
```
jviau: I like to add extension methods for semaphores:

``` csharp
public static class LockExtensions
{
    extension (SemaphoreSlim semaphore)
    {
        public async ValueTask<SemaphoreLock> LockAsync(CancellationToken cancellation = default)
        {
            ArgumentNullException.ThrowIfNull(semaphore);
            await semaphore.WaitAsync(cancellation).ConfigureAwait(false);
            return new(semaphore);
        }
    }
}

/// <summary>
/// Allows for releasing semaphores via a using statement.
/// IMPORTANT!: this is NOT safe for multiple disposal. 
/// </summary>
public readonly struct SemaphoreLock(SemaphoreSlim semaphore) : IDisposable
{
    public void Dispose()
    {
        semaphore?.Release();
    }
}
```

src/Functions.Rpc.Client/WorkerChannelRegistry.cs
```diff
@@ -0,0 +1,481 @@
...
+            }
+
+            if (removedChannel is null)
+            {
+                return;
+            }
+
+            if (terminalException is null)
+            {
+                _logger.LogDebug("FunctionRpc channel for worker {WorkerId} completed and was removed.", workerId);
```
jviau: nit: consider logger source gen for all logs

src/Functions.Rpc.Client/WorkerChannelRegistry.cs
```diff
@@ -0,0 +1,481 @@
...
+            }
+            else
+            {
+                _logger.LogWarning(terminalException, "FunctionRpc channel for worker {WorkerId} failed and was removed.", workerId);
+            }
+
+            try
+            {
+                await removedChannel.DisposeAsync();
```
jviau: It is typically expected a disposal NEVER throws. Do we have cases where we will throw?

jviau (commented):
The concurrent link + unlink leading to an orphaned slot that copilot review seems like a valid concern

src/Functions.Rpc.Client/WorkerChannelRegistry.cs
brettsam: updated other callers of this as well...

src/Functions.Rpc.Client/Registry/WorkerChannelRegistry.cs
```diff
@@ -0,0 +1,481 @@
...
+            throw new ObjectDisposedException(GetType().FullName);
+        }
+        finally
+        {
+            if (gateEntered)
+            {
+                slot.Gate.Release();
+            }
+
+            RemoveEmptySlot(workerId, slot);
```
kshyju, line 104: GHCP review flagged this:

> `slot.Gate.Release()` happens **before** `RemoveEmptySlot(...)`. A waiter (`UnlinkAsync` or completion monitor) can acquire the gate, set `slot.Channel = null`, and begin disposal; then the original `LinkAsync` removes the slot key while cleanup is still running. That allows a new `LinkAsync` for the same worker ID before prior cleanup completes, breaking the intended same-worker serialization guarantee.

Check for `UnlinkAsync` method as wel.

src/Functions.Rpc.Client/WorkerChannelRegistry.cs
```diff
@@ -0,0 +1,481 @@
...
+    private RpcClientWorkerChannel DetachChannelLocked(WorkerSlot slot, RpcClientWorkerChannel expectedChannel)
+    {
+        RpcClientWorkerChannel channel = slot.Channel;
+        if (channel is null || (expectedChannel is not null && !ReferenceEquals(channel, expectedChannel)))
+        {
+            return null;
+        }
+
+        slot.Channel = null;
+        if (!IsDisposed && !_slots.Values.Any(slot => slot.Channel is not null))
```
kshyju: GHCP suggestion:

> **Non-blocking perf suggestion:** can we track initialized channels as explicit O(1) state (for example `_initializedCount` + cached first channel, or a linked list) instead of scanning `_slots.Values` in `DetachChannelLocked` / `WaitForFirstInitializedAsync`?
> This would remove repeated O(N) work under `_stateLock`, reduce lock hold time during link/unlink churn, and keep the `WaitForFirstInitializedAsync` fast path O(1), without changing semantics.

kshyju: Consider if it does not impact readability. Keeping more than one collection and keeping them in sync could go wrong in many ways, causing race conditions, if not properly done.

src/Functions.Rpc.Client/WorkerChannelRegistry.cs
```diff
@@ -0,0 +1,481 @@
...
+    public IReadOnlyList<WorkerChannel> GetInitializedChannels()
+    {
+        lock (_stateLock)
+        {
+            if (IsDisposed)
+            {
+                return [];
+            }
+
+            return _slots.Values
```
kshyju: Assuming this method may be called on hot path, We could consider to a manual loop like below to save WhereEnumerableIterator and WhereSelectEnumerableIterator allocations.

```
public IReadOnlyList<WorkerChannel> GetInitializedChannels()
{
    lock (_stateLock)
    {
        if (IsDisposed || _slots.Count == 0)
        {
            return [];
        }

        var buffer = new WorkerChannel[_slots.Count];
        var count = 0;

        foreach (var slot in _slots.Values)
        {
            if (slot.Channel is { } channel)
            {
                buffer[count++] = channel;
            }
        }

        if (count == 0)
        {
            return [];
        }

        return count == buffer.Length
            ? buffer
            : buffer.AsSpan(0, count).ToArray();
    }
}
```

kshyju (approved):

src/Functions.Rpc.Client/WorkerChannelRegistry.cs
kshyju: @copilot Does the above also prevent boxing compared to the original version in the PR?

src/Functions.Rpc.Client/Registry/WorkerChannelRegistry.cs
brettsam: yep that was mentioned above -- a slight refactor has fixed this now. coming in next commit.

src/Functions.Rpc.Client/WorkerChannelRegistry.cs
brettsam: I purposefully kept this simple for now to not over-optimize. I'm not sure this will be a super-hot path. But it's a good point and something we should come back to when we get some real data.

src/Functions.Rpc.Client/WorkerChannelRegistry.cs
brettsam: I was optimizing for simplicity right now. I'm not sure we'll have perf impact at all here, but we can always come back to it. The initial implementation did have things like this and it was tough to follow so I want to make sure it works first, then we'll come back to make it fast :-)

src/Functions.Rpc.Client/WorkerChannelRegistry.cs
brettsam: That's funny, it's what I'd originally had but I backed it out for readability. But I do want to start using these initializers in new projects, though, so I'll re-add it back.

src/Functions.Rpc.Client/WorkerChannelRegistry.cs
brettsam: it's just paranoia at this point... i'll leave this for now but come back and clean all this up to ensure we don't throw anywhere.

brettsam:
Note I moved some of the files here into subfolders, so there's a bunch of renames on top of the other changes... the project was growing so wanted some organization.

kshyju (approved):

PR #12015, opened 2026-09-10
Update System.Memory.Data to 10.0.12

fabiocav:

No linked issue.

Update the centrally managed `System.Memory.Data` reference from `8.0.1` to `10.0.12`, the latest stable 10.x release at the time of this change, to align with the host's `net10.0` target framework. Removing the central pin alone would still resolve `8.0.1` through the current Azure SDK dependencies.

### Pull request checklist

- [ ] Backporting to the `in-proc` branch is not required
- [x] My changes **do not** require documentation changes
- [ ] My changes **should not** be added to the release notes for the next release
- [ ] My changes **do not** need to be backported to a previous version
- [x] My changes **do not** require diagnostic events changes
- [x] I have added all required tests (Unit tests, E2E tests)

Backport and release-note disposition are left for maintainer review; no backport is included in this PR.

### Additional information

- Refresh the generated WebHost dependency baseline to use `System.Memory.Data`'s `net10.0` asset. Regeneration also captures the current first-party host version metadata (`4.1054` to `4.1055`), which is excluded from the dependency comparison.
- Add a regression test confirming that an assembly request for the previously used `System.Memory.Data, Version=8.0.0.1` unifies to the current host assembly in the default load context.
- The intended assembly-version change is `8.0.0.1` to `10.0.0.12`; the public key token remains unchanged.

**Compatibility and validation**

| Validation | Result |
|---|---|
| Dependency baseline, runtime assembly loading, and BinaryData binding/RPC conversion | All 13 targeted tests passed |
| Microsoft ApiCompat: actual 8.0.1 `net8.0` DLL versus 10.0.12 `net10.0` DLL | No breaking API changes, including attribute and parameter-name rules; full reference resolution, no suppressions, zero warnings/errors |
| Consumer compiled against 8.0.1 and run unchanged against both DLLs on .NET 10 | All 53 compatibility assertions passed with each version |
| Decompiled implementation comparison | Four additive `FromFile`/`FromFileAsync` overloads; no concrete breaking behavior identified in existing implementations |

The compatibility evidence covers the exercised API and runtime scenarios, not an exhaustive guarantee that every possible behavior is identical.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -88,7 +88,7 @@
     <PackageVersion Include="System.IO.Abstractions" Version="2.1.0.227" />
     <PackageVersion Include="System.IO.FileSystem.AccessControl" Version="5.0.0" />
     <PackageVersion Include="System.IO.FileSystem.Primitives" Version="4.3.0" />
-    <PackageVersion Include="System.Memory.Data" Version="8.0.1" />
+    <PackageVersion Include="System.Memory.Data" Version="10.0.12" />
     <PackageVersion Include="System.Net.Http" Version="4.3.4" />
     <PackageVersion Include="System.Net.NameResolution" Version="4.3.0" />
     <PackageVersion Include="System.Private.Uri" Version="4.3.2" />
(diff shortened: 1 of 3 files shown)


mathewc (approved):

brettsam (approved):
These are always nerve-wracking... are we sure there's no subtle breaking changes like we've seen in the past? Not sure we can ever be 100% confident, though...

